PR #767, opened 2019-04-24
Research on token distribution and minimum stake

eth-r:
To model realistic conditions, we need an estimate of token distribution and an initial minimum stake example value. Using empirical data from other projects, we can conclude that KEEP is likely to follow a power law where the top `X` addresses hold at least `total / (10 * X)` tokens each. Setting the minimum stake at `total / 100,000` would thus be likely to result in roughly 10,000 addresses holding at least the minimum.

### DIFF
--- a/docs/random-beacon/staking/distribution.adoc
+++ b/docs/random-beacon/staking/distribution.adoc
@@ -1,0 +1,71 @@
+= Token distribution
+
+Token distribution at existing projects
+can be empirically observed to follow a power law:
+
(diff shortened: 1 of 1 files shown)


mhluongo (approved):
It'd be nice to see ranges in suggestions but looks good to me. Leaving it to someone else to merge

lionakhnazarov:
Closing as stale. 2019 research document on token distribution/minimum stake; retained for reference but not an actionable PR.

PR #768, opened 2019-04-24
Research on models of operating

eth-r:
The main possibilities are "leveraged" and "fully backed" operation. Analyze
both and their vulnerability to various attacks and problems, investigate
mitigations.

(Could probably use a bit more polish especially at the end.)


mhluongo (commented):
Interesting that you don't consider the mixed leverage model that we're currently building (leveraged work token, fully backed secondary bonds)

lionakhnazarov:
Closing as stale. 2019 research document on operating models; retained for reference but not an actionable PR.

PR #1274, opened 2020-01-14
Validation of signature shares

### LINKED ISSUE #1075: Validation of signature shares
If signature shares aren't verified, broadcasting invalid shares can serve as a DOS attack vector.

If signature shares are verified before reconstructing the threshold signature, and invalid signature shares rejected, the major DOS attack vector is closed.

This issue aims at implementing signature shares validation. Only valid shares should be used to reconstruct a signature.

lukasz-zimnoch:
Closes #1075

**Summary**
- Added calculation of group public key shares at the end of GJKR (phase 12) and passing the result to the `ThresholdSigner` instance.
- Validation of incoming shares using BLS verification


pdyraga:
@lukasz-zimnoch Can you please update it with the recent `master` changes?

pkg/beacon/relay/dkg/dkg.go
```diff
@@ -19,6 +20,8 @@ import (
 
 var logger = log.Logger("keep-dkg")
 
+const publicationDelayBlocks = 20
```
pdyraga, line 23: I'd prefer having this in the state machine for `gjkr` as we are executing calculations on `gjkr` coproducts. And the value of this delay reflects the complexity of those calculations.

A possible idea how to get some of this code out of `dkg.go`:

In `states.go`, at `Initiate` of `combinationState` we can do:
```
cs.member.CombineGroupPublicKey()
cs.member.ComputeGroupPublicKeyShares()
```

`ComputeGroupPublicKeyShares` triggers a goroutine computing the shares asynchronously. Once the result is ready, the goroutine unblocks a `Cond` in `gjkr.Result`:

```
package gjkr

type Result struct {
  // (...)
  
  groupPublicKeySharesReady sync.Cond
  groupPublicKeyShares map[group.MemberIndex]*bn256.G2
}

func (r *Result) GroupPublicKeyShares() map[group.MemberIndex]*bn256.G2 {
  // wait for cond to be fulfilled and return groupPublicKeyShares
}
```

Then, in `dkg.go` we:
- get rid of `publicationDelayBlocks` and just increase the time for `combinationState` 
- proceed as before, remove.`groupPublicKeySharesChan` construction
- when we return the result we do `gjkrResult.GroupPublicKeyShares()` and this is possibly blocking

I think it's a little more elegant as the calculation of the shares is not leaking out of `gjkr`- we just may need to wait for `GroupPublicKeyShares()`. I was also thinking about using a channel instead of `sync.Cond` but this is problematic as the value can be read only once.

pkg/beacon/relay/entry/states.go
```diff
@@ -125,6 +125,12 @@ func (scs *signatureCompleteState) Initiate(ctx context.Context) error {
 		scs.MemberIndex(),
 	)
 
+	previousEntryG1 := new(bn256.G1)
+	_, err := previousEntryG1.Unmarshal(scs.previousEntry)
+	if err != nil {
+		return err
+	}
```
pdyraga, line 132: What if we do this in `entry.SignAndSubmit` and pass `bn256.G1` as `previousEntry` to `signingStateBase`? 

`previousEntry` is currently represented as `byte[]` and is used in two places:
- `sss.signer.CalculateSignatureShare` where it's unmarshalled to G1 before passing to `bls.SignG1`
- in this function where it is also unmarshalled to G1 before doing anything else.

pkg/beacon/relay/entry/states.go
```diff
@@ -136,12 +142,31 @@ func (scs *signatureCompleteState) Initiate(ctx context.Context) error {
...
-				"[member:%v] accepting signature share from member [%v]",
-				scs.MemberIndex(),
-				message.senderID,
-			)
-			seenShares[message.senderID] = share
+			publicKey, ok := scs.signer.GroupPublicKeyShares()[message.senderID]
+			if !ok {
+				logger.Warningf(
+					"[member:%v] could not validate signature share from "+
+						"member [%v] because their public key was not found",
```
pdyraga, line 149: `because their public key was not found` or `because their public key share was not found`?

We can also shorten it a little:
```
[member:%v] could not validate signature share from member [%v]; public key share not found
```

pkg/beacon/relay/entry/states.go
```diff
@@ -136,12 +142,31 @@ func (scs *signatureCompleteState) Initiate(ctx context.Context) error {
 				err,
 			)
 		} else {
-			logger.Debugf(
-				"[member:%v] accepting signature share from member [%v]",
-				scs.MemberIndex(),
-				message.senderID,
-			)
-			seenShares[message.senderID] = share
+			publicKey, ok := scs.signer.GroupPublicKeyShares()[message.senderID]
```
pdyraga, line 145: `publicKey` or `publicKeyShare`?

pkg/beacon/relay/entry/states.go
```diff
@@ -136,12 +142,31 @@ func (scs *signatureCompleteState) Initiate(ctx context.Context) error {
 				err,
```
pdyraga, line 142: We should also refactor this case: if a share could not be unmarshalled or is not valid, we log a warning and continue. We currently reject invalid `bn256.G1` shares but still let the malicious member break the entire protocol by broadcasting invalid public key share `[]byte`. This can lead to an attack when a member breaks signing for the entire group and reports entry timeout to get a tattletale reward.

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -906,27 +906,37 @@ func (sm *SharingMember) isValidMemberPublicKeySharePointsMessage(
 // What, using elliptic curve, is the same as:
 // G * s_ji == Σ ( A_j[k] * (i^k) ) for `k` in `[0..T]`
 func (sm *SharingMember) isShareValidAgainstPublicKeySharePoints(
-	senderID group.MemberIndex,
+	receiverID group.MemberIndex,
```
pdyraga, line 909: I think I understand why we did this rename but would like to clarify. It is not about a *message receiver*, it is about a *share receiver*, that is, for which party `shareS` was generated for. Is that correct?
- In phase 8 (`VerifyPublicKeySharePoints`), we use this function passing message receiver (`sm.ID`)
- In phase 9 (`ResolvePublicKeySharePointsAccusationsMessages`), we use this function passing message sender (`message.senderID`)

If my understanding is correct, maybe we should rename `receiverID` to `shareReceiverID` and make it clear in the documentation. For example:

```
// isShareValidAgainstPublicKeySharePoints verifies if public key share points
// are valid for passed share S generated for the specific member, denoted as
// a share receiver.
```
?

pkg/beacon/relay/registry/marshalling_test.go
```diff
@@ -17,6 +17,9 @@ func TestMembershipRoundtrip(t *testing.T) {
 		group.MemberIndex(2),
 		new(bn256.G2).ScalarBaseMult(big.NewInt(10)),
 		big.NewInt(1),
+		map[group.MemberIndex]*bn256.G2{
+			group.MemberIndex(1): new(bn256.G2).ScalarBaseMult(big.NewInt(10)),
```
pdyraga, line 21: Can we add one more entry to this map? In all real-world cases, we should have > 1 entry in this map.

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -1655,8 +1668,49 @@ func (rm *CombiningMember) CombineGroupPublicKey() {
...
+	groupPublicKeyShares := make(map[group.MemberIndex]*bn256.G2)
+
+	// Calculate group public key shares for all other operating members.
+	for _, receiverID := range rm.group.OperatingMemberIDs() {
+		if receiverID == rm.ID {
+			continue
+		}
+
+		// Calculate first public key share for given receiver based on
+		// current member public key share points.
```
pdyraga, line 1690:
```
// Calculate the first public key share for the given receiver based on
// the current member public key share points.
```

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -1655,8 +1668,49 @@ func (rm *CombiningMember) CombineGroupPublicKey() {
...
+		sum := rm.publicKeyShare(receiverID, rm.publicKeySharePoints)
+
+		// Iterate through the `QUAL` set and calculate subsequent
+		// public key share for given receiver based on...
```
pdyraga, line 1694: `the given`

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -1655,8 +1668,49 @@ func (rm *CombiningMember) CombineGroupPublicKey() {
...
+		for senderID := range rm.receivedQualifiedSharesS {
+			// ...received and valid other members public key share points...
```
pdyraga, line 1696: I'd say `...received and valid member's public key share points` - I think it better explains what are we doing here.

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -1655,8 +1668,49 @@ func (rm *CombiningMember) CombineGroupPublicKey() {
...
+				// take their reconstructed share and recover the public key share.
+			} else {
+				for _, shares := range rm.revealedMisbehavedMembersShares {
+					if shares.misbehavedMemberID == senderID {
+						publicKeyShare := new(bn256.G2).ScalarBaseMult(shares.peerSharesS[receiverID])
+						sum = new(bn256.G2).Add(sum, publicKeyShare)
+					}
+				}
+			}
+		}
```
pdyraga, line 1710: What do you think about renaming:
- `receiverID` -> `operatingMemberID`
- `senderID` -> `qualifiedMemberID`
?

pkg/beacon/relay/gjkr/protocol.go
pdyraga: ...or replace `the given receiver` with `the given operating member`, per [this](https://github.com/keep-network/keep-core/pull/1274#discussion_r401466167) comment.

pkg/beacon/relay/gjkr/protocol.go
nkuba: >  share receiver, that is, for which party shareS was generated for

We used to call them `peerID`, right?

pdyraga: I am not sure, I see `peerID` is used only in `reconstructIndividualPrivateKeys`

pdyraga:
Can we implement an integration test stressing signature share validation? We can easily intercept a signature share message from one or more members and insert there either incorrect bytes (not deserializing to G1) or valid G1 bytes but with a share that does not pass the validation.

pkg/beacon/relay/gjkr/protocol.go
lukasz-zimnoch: Yeah, agree with Piotr. `peerID` is a network concept used by libp2p and in terms of gjkr we should rather use `member`. Regarding `isShareValidAgainstPublicKeySharePoints` param I think `shareReceiverID` is the best proposition.

pkg/beacon/relay/dkg/dkg.go
lukasz-zimnoch: Excellent idea but there is one problem. We must pass the result of `ComputeGroupPublicKeyShares` which works in a separate goroutine to `gjkr.Result` so we can't avoid an intermediate channel. Also `sync.Cond` is for cases when we have multiple goroutines waiting for an async result. In our case, a simple channel seems to be the best option. Apart from that, I agree we should push those computations inside GJKR. I will propose a solution inspired by this comment but without `sync.Cond` usage.

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -1655,8 +1668,49 @@ func (rm *CombiningMember) CombineGroupPublicKey() {
...
+
+// groupPublicKeyShares combines group public key shares for each
+// individual member in the group. Those group public key shares are
+// needed to perform the verification of relay entry signature shares coming
+// from given group member.
+func (rm *CombiningMember) groupPublicKeyShares() map[group.MemberIndex]*bn256.G2 {
+	groupPublicKeyShares := make(map[group.MemberIndex]*bn256.G2)
+
+	// Calculate group public key shares for all other operating members.
+	for _, receiverID := range rm.group.OperatingMemberIDs() {
```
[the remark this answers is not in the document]
lukasz-zimnoch: As I mentioned above `peer` is rather a network concept. I will rename `receiverID` to `operatingMemberID` as Piotr suggested above.

pkg/beacon/relay/entry/states.go
lukasz-zimnoch: Good point! Thanks!

pkg/beacon/relay/entry/states.go
pdyraga: Ah, we are not returning here, my bad! Let's just change it to `Warningf`.

lukasz-zimnoch:
@pdyraga, @nkuba ready for the next shot!

pkg/beacon/relay/integration_test.go
```diff
@@ -165,6 +168,66 @@ func TestInactiveMemberPublicKeySharesReconstructionAndSigning(t *testing.T) {
 	}
 }
 
+func TestInvalidSignatureSharesExists(t *testing.T) {
```
pdyraga, line 171: Maybe `TestsSigningWithInvalidSignatureShare`? I think this one better fits the rest of the names we have in integration tests.

pkg/beacon/relay/entry/message.go
```diff
@@ -11,6 +11,13 @@ type SignatureShareMessage struct {
 	shareBytes []byte
 }
 
+func NewSignatureShareMessage(
+	senderID group.MemberIndex,
+	shareBytes []byte,
+) *SignatureShareMessage {
+	return &SignatureShareMessage{senderID, shareBytes}
+}
```
pdyraga, line 19: Instead of adding it here, we should use the trick with `export_test.go`. See `pkg/beacon/relay/gjkr/export_test.go`.

pkg/beacon/relay/integration_test.go
```diff
@@ -165,6 +168,66 @@ func TestInactiveMemberPublicKeySharesReconstructionAndSigning(t *testing.T) {
...
+		// Member 1 sends shares which could not be unmarshalled as a G1 point.
+		if ok && signatureShareMessage.SenderID() == group.MemberIndex(1) {
+			return entry.NewSignatureShareMessage(
+				signatureShareMessage.SenderID(),
+				[]byte{0, 1},
+			)
+		}
+
+		// Member 2 sends a proper G1 point which is invalid in terms of
+		// current relay entry request.
```
pdyraga, line 186: `the current`

pkg/beacon/relay/entry/message.go
lukasz-zimnoch: We can't use this trick. `integration_tests.go` are in `relay` package and `SignatureShareMessage` in `entry` package.

pkg/beacon/relay/integration_test.go
```diff
@@ -165,6 +168,66 @@ func TestInactiveMemberPublicKeySharesReconstructionAndSigning(t *testing.T) {
 	}
 }
 
+func TestsSigningWithInvalidSignatureShare(t *testing.T) {
```
pdyraga, line 171:
```
17:10:02.967 WARNI keep-entry: [member:2] failed to unmarshal signature share from member [1]: [bn256: not enough data] states.go:128
17:10:02.967 DEBUG keep-entry: [member:7] accepting signature share from member [5] states.go:147
17:10:02.967 WARNI keep-entry: [member:7] failed to unmarshal signature share from member [1]: [bn256: not enough data] states.go:128
(...)
17:10:02.995 WARNI keep-entry: [member:3] rejecting invalid signature share from member [2] states.go:154
17:10:03.009 WARNI keep-entry: [member:9] rejecting invalid signature share from member [2] states.go:154
(...)
```

🙌

pkg/beacon/relay/entry/message.go
pdyraga: Hmm... right. I was thinking about moving this integration test to `entry` but I see it's painful. Not worth the effort at this point.

pkg/beacon/relay/gjkr/member.go
```diff
@@ -197,6 +200,8 @@ type CombiningMember struct {
 	// Group public key calculated from individual public keys of all group members.
 	// Denoted as `Y` across the protocol specification.
 	groupPublicKey *bn256.G2
+
+	groupPublicKeySharesChannel chan map[group.MemberIndex]*bn256.G2
```
pdyraga, line 204: It's worth explaining the magic behind it. For example:
```
// Group public key shares calculated for each QUAL group member. 
// Public key shares calculation is time-expensive so we do it in an async
// manner and publish the result to this channel, once ready.
```

pkg/beacon/relay/node.go
```diff
@@ -121,6 +121,11 @@ func (n *Node) JoinGroupIfEligible(
 				if err != nil {
 					logger.Errorf("failed to register a group: [%v]", err)
 				}
+
+				logger.Debugf(
```
pdyraga, line 125: I'd escalate it to `Info`

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -1643,20 +1657,69 @@ func pow(id group.MemberIndex, y int) *big.Int {
...
+								shares.peerSharesS[operatingMemberID],
+							)
+							sum = new(bn256.G2).Add(sum, publicKeyShare)
+						}
+					}
+				}
+			}
+
+			groupPublicKeyShares[operatingMemberID] = sum
+		}
```
pdyraga, line 1721: I think it's worth wrapping the code in this goroutine with an info log with a classic `[member:x] ...` pattern informing when do we start computations and when they are completed.

pdyraga:
All nitpicks, let's address in a separate PR. No need to block here.

pkg/beacon/relay/integration_test.go
lukasz-zimnoch: Addressed in #1547.

pkg/beacon/relay/gjkr/member.go
lukasz-zimnoch: Addressed in #1547.

pkg/beacon/relay/node.go
lukasz-zimnoch: Addressed in #1547.

pkg/beacon/relay/gjkr/protocol.go
lukasz-zimnoch: Addressed in #1547.

PR #1307, opened 2020-01-27
Validate declared group member index against group selection result

### LINKED ISSUE #484: Validate DKG member index against group selection result
The group selection protocol outcome is an array of values uniquely identifying clients selected to the group. Index of a client in that array is an index used for DKG protocol. During DKG protocol execution we need to validate those indexes - check if all members we cooperate with, use the correct ones. If not, the message from the given sender should be rejected.

### LINKED ISSUE #834: Validate thresholdsig member index against group selection result
The group selection protocol outcome is an array of values uniquely identifying clients selected to the group. Index of a client in that array is an index used for DKG protocol. During DKG protocol execution we need to validate those indexes - check if all members we cooperate with, use the correct ones. If not, the message from the given sender should be rejected. This is covered by #484. We need to do the same for threshold signature protocol when we sign relay entry. Messages with member indexes already seen are rejected. We need to make sure they were sent by eligible participants.

pdyraga:
Closes #484 
Closes #834 

This work is still in progress.

> The group selection protocol outcome is an array of values uniquely identifying clients selected to the group. Index of a client in that array is an index used for DKG protocol. During DKG protocol execution we need to validate those indexes - check if all members we cooperate with, use the correct ones. If not, the message from the given sender should be rejected.

The main component is `group.MembershipValidator` implemented by `groupselection.MembershipValidator` and passed to DKG executor. For every received message it checks the public key of the sender and its declared index against group selection result - ordered array of staker addresses. This component replaced `createGroupMemberFilter` in `node.go` as it offers the same functionality.

TODO:
- [ ] validation of sender index during relay entry signing
- [ ] this validation - although seems to be pretty lightweight - causes problems in commitment verification state; a lot of members is marked as inactive
- [ ] we are currently dropping message - should we consider disqualifying member?


lukasz-zimnoch:
Closed in favor of #1534

lukasz-zimnoch:
Closed in favor of #1534

PR #1408, opened 2020-03-03
Version Tether: Continuous packaging and versioning

Shadowfiend:
This PR represents an adjustment to how we manage cross-repository
versioning, package publishing, and dependency package and contract
references. Here are the major changes:
- `package.json` for Solidity contracts is set to `<version>-pre`, with no
  numeric suffix.
- A new Circle job, `publish_npm_package`, runs on master merge. This
  This job updates the package version to `-pre.<number>`, where `<number>`
  represents a new, unpublished version.
- Contract artifacts from the contract migration are copied to an `artifacts/`
  subdirectory and included in the package.
- The package is published to the main npm repository (not the GitHub
  package repository).
- We trigger a downstream build (placeholder, didn't get a chance to hook this
  up yet, plus we should wait until this build strategy propagates to the other
  repositories).

This PR includes tweaking the imports and dependency declaration on the
token dashboard. The token dashboard `package.json` references a dependency
range that includes all `-pre` versions of `keep-core`, and the docker build
now does an explicit `npm upgrade` to fetch the latest version of `keep-core`
as part of the build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -190,7 +190,11 @@ jobs:
             preid=$(echo $version | sed -e s/^.*-\\\([^.]*\\\).*$/\\1/)
 
             # Find the latest published package version matching this preid.
-            latest_version=$(npm show -json "$name@^$version" version | jq --raw-output ". | sort | .[-1]")
+            # Note that in jq, we wrap the result in an array and then flatten;
+            # this is because npm show json contains a single string if there
+            # is only one matching version, or an array if there are multiple,
+            # and we want to look at an array always.
+            latest_version=$(npm show -json "$name@^$version" version | jq --raw-output "[.] | flatten | sort | .[-1]")
             latest_version=${latest_version:-$version}
             if [ -z $latest_version ]; then
               echo "Latest version calculation failed. Resolved info:"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,7 +179,7 @@ jobs:
           at: /tmp/keep-client
       - checkout
       - run:
-          name: Copy contract artifacts
+          name: Bump and publish npm package
           working_directory: ~/project/solidity
           command: |
             set -x
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -332,9 +332,23 @@ workflows:
   solidity:
     jobs:
       - test_solidity
+  build-and-test:
+    jobs:
+      - build_client_and_test_go:
+          filters:
+            branches:
+              ignore: master
+          context: keep-dev
+      - build_token_dashboard_dapp:
+          filters:
+            branches:
+              ignore: master
   build-test-migrate-publish-keep-dev:
     jobs:
       - build_client_and_test_go:
+          filters:
+            branches:
+              only: master
           context: keep-dev
       - build_initcontainer:
           filters:
@@ -367,10 +381,10 @@ workflows:
             - build_initcontainer
             - migrate_contracts
       - trigger_downstream_builds:
-          context: keep-dev
           filters:
             branches:
               only: master
+          context: keep-dev
           requires:
             - publish_npm_package
       - publish_contract_data:
@@ -381,11 +395,6 @@ workflows:
           requires:
             - build_client_and_test_go
             - migrate_contracts
-      # When building the token dashboard for a PR, don't wait on anything.
-      - build_token_dashboard_dapp:
-          filters:
-            branches:
-              ignore: master
       # When building the token dashboard for a master merge, wait for npm
       # package publish
       - build_token_dashboard_dapp:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -222,11 +222,6 @@ jobs:
           docker_layer_caching: true
       - checkout
       - run:
-          name: Copy contract artifacts
-          working_directory: ~/project/solidity/dashboard
-          command: |
-            ./scripts/copy-contracts.sh /tmp/keep-client/contracts
-      - run:
           name: Run Docker build
           working_directory: ~/project/solidity/dashboard
           command: |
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -172,19 +172,65 @@ jobs:
...
-            ./scripts/copy-contracts.sh /tmp/keep-client/contracts
+            set -x
+            mkdir -p artifacts
+            cp -r /tmp/keep-client/contracts/* artifacts/
+            name=$(jq --raw-output .name package.json)
+            version=$(jq --raw-output .version package.json)
+            preid=$(echo $version | sed -e s/^.*-\\\([^.]*\\\).*$/\\1/)
+
+            # Find the latest published package version matching this preid.
+            latest_version=$(npm show -json "$name@^$version" version | jq --raw-output ". | sort | .[-1]")
```
sthompson22: I think this requires the return to be an `array` or it will fail.  Do we know when there's only a single published package for a particular version, if the return is an `array` or `string`?

sthompson22: I think this will return `empty` if we increment the minor version, e.g.  going from `0.12.0-pre` to `0.13.0-pre`, causing the subsequent `latest_version` check to fail.

In the test workflow [here](https://app.circleci.com/pipelines/github/keep-network/keep-core/6089/workflows/33505a29-33df-4c51-be4a-2c332a0d9245/jobs/47320) We're incrementing a minor package version `0.12.0-pre.1` who's minor version matches the result of `version` in this script.

Just want to make sure this behaves as expected when we are publishing `x.x.x-pre.0`

.circleci/config.yml
```diff
@@ -172,19 +172,65 @@ jobs:
...
+              echo "$name@$version ; preid $preid"
+              exit 1
+            fi
+
+            # Update package.json with the latest published package version matching this
+            # preid to prepare for bumping.
+            echo $(jq -M ".version=\"${latest_version}\"" package.json) > package.json
+
+            # Bump without doing any git work. Versioning is a build-time action for us.
+            # Consider including commit id? Would be +<commit id>.
```
sthompson22: I like that idea.

.circleci/config.yml
```diff
@@ -172,19 +172,65 @@ jobs:
...
+            npm version prerelease --preid=$preid --no-git-tag-version
+
+            # Fix resolved dependency versions.
+            npm update
+
+            # Publish to npm.
+            echo //registry.npmjs.org/:_authToken=${NPM_TOKEN} > .npmrc
+            npm publish --access=public
+
+  trigger_downstream_builds:
```
sthompson22: Yessssssss

.circleci/config.yml
```diff
@@ -172,19 +172,65 @@ jobs:
...
+
+            # Bump without doing any git work. Versioning is a build-time action for us.
+            # Consider including commit id? Would be +<commit id>.
+            npm version prerelease --preid=$preid --no-git-tag-version
+
+            # Fix resolved dependency versions.
+            npm update
+
+            # Publish to npm.
+            echo //registry.npmjs.org/:_authToken=${NPM_TOKEN} > .npmrc
```
sthompson22: We're storing `NPM_TOKEN` in a Circle Context?  Will this token be the same across tbtc and keep-ecdsa?  If we require different tokens and this is in a Circle context we'll need to supplement the token name.

.circleci/config.yml
```diff
@@ -172,19 +172,65 @@ jobs:
...
+  publish_npm_package:
     executor: docker-node
     steps:
       - attach_workspace:
           at: /tmp/keep-client
-      - setup_remote_docker:
-          docker_layer_caching: true
       - checkout
       - run:
           name: Copy contract artifacts
```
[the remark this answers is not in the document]
Shadowfiend: … Derp. Yeah, fixing.

.circleci/config.yml
Shadowfiend: That's a great question regarding what `npm` gives us if there's a single result. I would hope still an array, but it seems very likely we'll just get a string.

If there's *no* result, this will leave `latest_version` empty, but we default it in the next line:
```
            latest_version=${latest_version:-$version}
```

This means it will end up being set to `0.13.0-pre`. This is as designed, because when we later to do `npm version prerelease`, it will get bumped to `-pre.0`.

.circleci/config.yml
Shadowfiend: The token should be the same across repos, yeah. It's for a `thesis-heimdall` account on npm, at least for now.

I don't love that `thesis-heimdall` can deploy with the token even if we're releasing a final version, but… Need to cross that bridge separately.

.circleci/config.yml
Shadowfiend: < sigh > yep, we get a string when there's a single result, even for `-json`. I know exactly how this decision was made, and I disagree with it so vehemently it ain't even funny lol.

Give me a minute with `jq`.

sthompson22: > but we default it in the next line:

Shnikies, totally missed it, thanks for clarifying.

Shadowfiend: Got it. Going to wrap it in an array and then flatten 🙃

.circleci/config.yml
Shadowfiend: Let's leave this for future work? I also want to pull this into a script or orb we can reuse across repos, but want to get it out and working first.

.circleci/config.yml
sthompson22: weeeeeee

sthompson22:
Waiting on green.

.circleci/config.yml
Shadowfiend: How did you miss the obvious `${...:-default}` bash syntax?

I can't believe this is how you default things in bash.

PR #1417, opened 2020-03-04
Add IAM viewer role to keep-dev and keep-test, and add kborg to both as viewer

kb0rg:
While reviewing #1411 - and the associated flowdock threads (linked in that PR) describing the process, I got the impression there were a lot more steps than are visible via commits in that PR, and that having some visibility into the GCP console for those environments would be helpful.

This adds the iam "viewer" role to both `keep-dev` and `keep-test` and adds me as a viewer in each of those environments.

### DIFF
--- a/infrastructure/terraform/keep-dev/iam.tf
+++ b/infrastructure/terraform/keep-dev/iam.tf
@@ -5,6 +5,13 @@ module "iam_members_editor" {
   members = "${var.editor_iam_members}"
 }
 
+module "iam_members_viewer" {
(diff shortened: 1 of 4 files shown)


kb0rg:
@sthompson22 - I've been wondering...

Since you gave me more global GCP access, I know that could technically add myself in the console. 

But theoretically, without that access, I shouldn't be able to terraform-add myself, should I? Even if you gave me VPN access, wouldn't I still need IAM credentials already in place to be able to access the project/ run terraform?

kb0rg:
closing, no longer necessary

PR #1462, opened 2020-03-16
Member should not leave the group if it agrees with group public key

### LINKED ISSUE #1325: Member should not leave the group if it agrees with group public key
Right now during DKG member leaves the group if it does not agree with full DKG result, including misbehaving members.

It may happen that some malicious group member intentionally waits with sending a message so that it can't propagate to all group members. In that case, some part of the group will agree on the result and some other part of the group will not agree on the result.

If member does not agree fully with the misbehaving array it should not support this result by signing it but it should not leave the group if this result is supported by the majority and the public key of the group is the same.

lukasz-zimnoch:
Closes #1325 

Members stay in the group if they agree with the group public key registered on chain and are not considered as misbehaved by the group. This way there is more active signers while processing a relay entry request.


pkg/beacon/relay/dkg/dkg.go
```diff
@@ -71,3 +118,57 @@ func ExecuteDKG(
 		groupPrivateKeyShare: gjkrResult.GroupPrivateKeyShare,
 	}, nil
 }
+
+func waitForDkgResultEvent(
+	dkgResultChannel chan *event.DKGResultSubmission,
+	startPublicationBlockHeight uint64,
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah but you must consider the `SigningStateBlocks` while waiting. If you don't consider those `6` blocks you will timeout `6` block too early if DKG is published by members from the end of the queue. I can rename `SigningStateBlocks` to `PublicationBlocks` and treat this as an aggregated duration of all publication states. This way if a new step is added we won't be forced to modify `dkg.go`.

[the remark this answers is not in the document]
pdyraga: On a happy path, we shouldn't exit earlier because `Publish` does not return before it's ready to publish or before it knows there is no chance to publish the result. On the other hand, if we add another state before the publication, my idea will no longer work. 

So let's maybe rename `SigningStateBlocks` to `PrePublicationBlocks`? `PublicationBlocks` suggests it's about how long it takes to publish. And let's document that `PrePublicationBlocks` is the total number of blocks it takes to execute all the required work to get ready for the result publication or to decide to skip the publication because there are not enough supporters for that particular result.

pkg/beacon/relay/dkg/dkg_test.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func TestDecideMemberFate_HappyPath(t *testing.T) {
+	setup()
+
+	go func() {
+		// DKG result must arrive before the timeout block.
+		// Timeout block equals to 21 because:
+		// - startPublicationBlockHeight = 0
+		// - signingStateBlocks = 6
```
pdyraga: `s/signingStateBlocks/prePublicationBlocks` or just drop this comment because we are not testing `timeout - 1` corner case here.

pkg/beacon/relay/dkg/dkg_test.go
```diff
@@ -0,0 +1,157 @@
...
+		t.Errorf("Error should not be nil")
+	}
+
+	expectedError := fmt.Sprintf(
+		"[member:%v] could not stay in the group because "+
+			"member do not support the same group public key",
+		playerIndex,
+	)
+	if err != nil && err.Error() != expectedError {
+		t.Errorf("Error [%v] should be [%v]", err, expectedError)
```
pdyraga: For consistency with the rest of tests we should use the following pattern:
```
t.Errorf(
  "unexpected error\nexpected: %v\nactual:   %v\n",
  test.expectedError,
  err,
)
```

pkg/beacon/relay/dkg/dkg_test.go
```diff
@@ -0,0 +1,157 @@
...
+	startPublicationBlockHeight uint64
+	localChain                  chain.Handle
+	blockCounter                chain.BlockCounter
+)
+
+func setup() {
+	playerIndex = group.MemberIndex(1)
+	groupPublicKey = new(bn256.G2).ScalarBaseMult(big.NewInt(10))
+	gjkrResult = &gjkr.Result{GroupPublicKey: groupPublicKey}
+	dkgResultChannel = make(chan *event.DKGResultSubmission)
```
pdyraga: We can add `1` element buffer here and tests can be simplified by removing the goroutine and block mining:
```
dkgResultChannel = make(chan *event.DKGResultSubmission, 1)
```
and later:
```
func TestDecideMemberFate_HappyPath(t *testing.T) {
	setup()

	dkgResultChannel <- &event.DKGResultSubmission{
		(...)
	}

	err := decideMemberFate(
		(...)
	)

	(...)
}
```

pkg/beacon/relay/dkg/dkg.go
```diff
@@ -71,3 +112,82 @@ func ExecuteDKG(
...
+			playerIndex,
+		)
+	}
+
+	// If member is considered as misbehaved, it could not stay in the group.
+	for _, misbehaved := range dkgResultEvent.Misbehaved {
+		if playerIndex == misbehaved {
+			return fmt.Errorf(
+				"[member:%v] could not stay in the group because "+
+					"member is considered as misbehaved",
```
pdyraga: `is considered as misbehaving`

pkg/beacon/relay/dkg/dkg.go
```diff
@@ -71,3 +112,82 @@ func ExecuteDKG(
 		groupPrivateKeyShare: gjkrResult.GroupPrivateKeyShare,
 	}, nil
 }
+
+// decideMemberFate decides what the member will do in case it failed
+// publishing its DKG result. Member can stay in the group if it
+// supports the same group public key as the one registered on-chain and
+// the member is not considered as misbehaved by the group.
```
pdyraga: `... considered as misbehaving ...`

pdyraga:
Took me a while to test it but it works as advertised 👍

PR #1482, opened 2020-03-19
Reassign grantee addresses in token grants

eth-r:
Unlocked grant tokens are withdrawn to the grantee address. Token grants can last for a number of years, during which the grantees' custody solutions may need to change.

This can be accommodated without forcing grantees to set up separate smart contract instances as the grantee addresses, by supporting grantee address reassignment. Approval from the grant manager is required, to ensure that grantee reassignment doesn't render grants excessively liquid, and to improve security.

Without grantee reassignment an adversary who learns the secret key of a grantee may chain transactions to withdraw unlocked tokens to an address controlled exclusively by the adversary. With grantee reassignment, grantee addresses that are suspected to be compromised can be reassigned to protect the remaining locked tokens with the cooperation of the grant manager. However, a malicious grant manager alone can not reassign grantee addresses.

---

The current state of this PR implements the basic reassignment functionality and its associated ACL. Not all interactions have tests written for them yet, but the fundamental grant functionality should be fine as is.

However, a variety of state records aren't updated with the reassignment, leaving dApps relying on them in an inconsistent state. The token grant contract needs to be refactored to unbreak these. Also, the grant contract is close to running out of gas on deployment, and contract size needs to be reduced if further functionality is to be added.


pdyraga:
Having #1514, do we still need this one?

eth-r:
No, the `ManagedGrant` solution #1514 is building towards supersedes the method in this one.

Issue #1490, opened 2020-03-06
Stake undelegation time for longer deposits


pdyraga:
Stake delegation initialization (maturation) period [is currently set](https://github.com/keep-network/keep-core/blob/master/contracts/solidity/migrations/2_deploy_contracts.js#L16-L17) to `50000` blocks (roughly one week) and stake undelegation period is set to `800000` blocks (roughly 3 months).

This is fine for the bacon but might not be enough for 6-month deposits in ECDSA keeps.

Stake undelegation time should be longer than the lifetime of the random beacon group but also longer than the lifetime of ECDSA keep.

It looks like it probably makes sense to set the maximum lifetime of a keep and to increase stake undelegation period to that time plus some safety margin.

@mhluongo @eth-r Interested in your thoughts on this issue.

PR #1492, opened 2020-03-20
Add utility for uint256 sets

eth-r:
`UintSet` behaves like an array with unique members, but also supports O(1) membership tests and removals. Not currently suitable for in-memory use due to the mapping backend.


pdyraga:
Where do we need it? 🤔

eth-r:
Originally the sets were for grantee reassignment in the token grant contract, but it's a generally useful building block and I'm pretty sure a version of it can be used for the locks of #1490

pdyraga:
Having `LockUtils.LockSet` in #1519, do we still need this PR?

PR #1497, opened 2020-03-20
Altbn128 optimization

eth-r:
Depends on #1496 

Reduces memory allocations and special-cases some common patterns to consume less gas. Somehow makes `g2Decompress` tests consume 2 seconds and 2M gas instead of 30 seconds and 3.5M gas. Slightly reduces the level of how much the slowness of running native solidity tests is pissing me off.


pdyraga:
#1496 is merged, can we please rebase and un-draft this one?

pdyraga:
@eth-r I merged master here. Is it ready for review now or do you want to add anything more here?

eth-r:
There's a bunch of further cleanup to be done, as you've already noticed and commented on

solidity/contracts/cryptography/AltBn128.sol
```diff
@@ -407,32 +482,65 @@ library AltBn128 {
      * provided point isn't on the curve.
      */
     function scalarMultiply(G1Point memory p_1, uint256 scalar) internal view returns (G1Point memory) {
-        uint256[3] memory arg;
-        arg[0] = p_1.x;
-        arg[1] = p_1.y;
-        arg[2] = scalar;
-        uint256[2] memory p_2;
+        (uint256 x, uint256 y) = _scalarMultiply(p_1.x, p_1.y, scalar);
```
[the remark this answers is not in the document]
eth-r: Latest:
```
testScalarMultiply (409ms, 640843 gas)
```

pdyraga:
> There's a bunch of further cleanup to be done, as you've already noticed and commented on

I am happy to finish the review of this PR as-is and merge. This is already an improvement, we can add more in a separate PR. This will make it the process faster as I can review smaller diffs.

solidity/contracts/cryptography/AltBn128.sol
```diff
@@ -306,23 +318,39 @@ library AltBn128 {
...
+            mstore(add(arg, 0x60), mload(add(b, 0x20)))
             // 0x60 is the ECADD precompile address
             if iszero(staticcall(not(0), 0x06, arg, 0x80, c, 0x40)) {
                 revert(0, 0)
             }
         }
+    }
 
-        return G1Point(c[0], c[1]);
+    function _g1Add(uint256 ax, uint256 ay, uint256 bx, uint256 by)
```
pdyraga: Do we need this function? I think it's not used anywhere.

solidity/contracts/cryptography/AltBn128.sol
```diff
@@ -335,6 +363,12 @@ library AltBn128 {
         );
     }
 
+    function _gfP2Add(uint256 ax, uint256 ay, uint256 bx, uint256 by)
```
pdyraga: Why do we add `_` suffix for some functions? Solidity allows for function name overloading. I did a test and renamed all `_gfP2*` to `gfP2*` and gas cost of all tests in `TestAltBn128` is the same. Also, the execution time of tests is not any different from what I see.

eth-r: The prefix is to denote stack-only versions of the functions. They're semi not intended for external use because of the more annoying API, but they save gas when used inside loops in the library.

solidity/contracts/cryptography/AltBn128.sol
eth-r: Neither `g1Add` nor `_g1Add` is used in the library, but I made both the memory-allocating and the stack-only versions available in the API. This makes the API a bit crowded but I'm not sure what can be done to improve it, except maybe making the stack-only functions `private` and only used for optimization inside the library.

PR #1498, opened 2020-03-21
keep-test: Update participant bundles

sthompson22:
This is to support deploying tag `v0.11.0`

- Downloaded latest contracts for staking
- Updated participant config files to point at latest contracts
- Moved `dekrypt` bundle to `blockdaemon`, see commit.
- Removed `formatAmount` function in `stake-accounts.js` in favor of an explicit stake amount, see commit.

---- Bonus

- Added Stake Capital bundle
- Updated the quickstart guide.


infrastructure/eth-networks/scripts/stake-accounts.js
```diff
@@ -74,18 +74,17 @@ async function stakeOperatorAccount(operatorAddress, contractOwnerAddress) {
 
   console.log(`Staking 2000000 KEEP tokens on operator account ${operatorAddress}`);
 
+  // Current minimumStake value configured on the TokenStaking contract.
+  let stakeAmount = web3.utils.numberToHex('200000000000000000000000')
```
pdyraga: Does this script represent what's currently deployed to Ropsten?

`TokenStaking` contract defines `minimumStake = 200000 * 1e18;` and we were previously staking `formatAmount(20000000, 18)` KEEPs. This amount lets to generate `20000000 / 200000 = 100` tickets and this is what we can observe in Ropsten clients logs:
```
generated [2] tickets for initial submission phase and [98] tickets for reactive submission phase
```

I think we now have two problems here:
- `console.log` was and still is out-of-date:

    ```
    2000000  <- console.log
    20000000 <- formatAmount
    200000   <- numberToHex, 200000000000000000000000 / (10^18)
    ```

- The number of tokens declared in `numberToHex` will be enough to generate only `1` ticket. This is fine only when we have at least 64 clients in the network and they are all participating in group selection:

    ```
    200000000000000000000000 = 200000 * 1e18 <- minimumStake
    200000000000000000000000 <- value in numberToHex
    ```

I think we should assume each client has `100 * minimum_stake` for test purposes. If so, we need to stake 
`100 * 200000 * 1e18 = 20000000000000000000000000`

sthompson22: Hmm I did use this for staking providers accounts with the current bundle.  I used the `keep-client ethereum` cli to call `has-mimimum-stake` on one of their accounts after running this script as you see it here and it returned `true`.  I assumed it was fine.  @pdyraga, Going to check again.

sthompson22: Ahh I see what you're saying now.  I can update.

sthompson22:
@pdyraga I've adjusted the stake amount, this works just fine.  Additionally, I've added a Stake Capital bundle to this PR.  I needed the work here to make that go.

infrastructure/eth-networks/scripts/stake-accounts.js
```diff
@@ -72,20 +73,19 @@ async function stakeOperatorAccount(operatorAddress, contractOwnerAddress) {
     Buffer.from(contractOwnerAddress.substr(2), 'hex') // authorizer
   ]).toString('hex');;
 
-  console.log(`Staking 2000000 KEEP tokens on operator account ${operatorAddress}`);
+  console.log(`Staking 20000000000000000000000000 KEEP tokens on operator account ${operatorAddress}`);
```
pdyraga: `20000000` keeps, since `20000000000000000000000000 / 10^18 = 20000000`. `1 KEEP` has `18` decimal places.

sthompson22: Fixed, thanks @pdyraga

PR #1499, opened 2020-03-22
Minimum stake schedule

### LINKED ISSUE #1368: Minimum stake parameterization and schedule
I propose we start with a minimum stake of 100k KEEP. Over the following 2 years after launch, I'd like to see the minimum stake lowered periodically using a uniform stepwise function, eventually ending at 10k. Thinking 10 steps or fewer, so we can talk about each step with the community and prepare for the discrete jump.

I chose this start value based on our initial token distribution, with the goal of making sure the p2p network isn't overwhelmed early on, giving us time to prepare for more nodes.

I believe we should hardcode this schedule, rather than it being governed.

Interested to hear your thoughts @eth-r

ngrinkevich:
Closes #1368 

Set initial minimum stake of 100k and lower periodically over 2 years via 10 steps eventually ending minimum stake at 10k.


Shadowfiend:
Do we need to store the minimum stake used during group creation so slashing is based on that? Or are we okay with group slashing operating on the *current* minimum stake rather than the minimum stake used when the group was created?

eth-r:
Current minimum stake should be fine.

pdyraga:
> Current minimum stake should be fine.

Agreed. ☝️  Especially that in case of a relay entry timeout we are slashing for the *current* misbehaviour. Older groups could be exposed to a more severe punishment than those created later even though they all have the same chance of being selected.

It is getting more complex for leaked group public key but given that relay entry timeout has a higher probability, I think it's fine to leave this case.

solidity/test/TestTokenStake.js
```diff
@@ -61,6 +63,25 @@ contract('TokenStaking', function(accounts) {
     );
   }
 
+  it("should get correct minimum stake based on the minimum stake schedule", async () => {
```
pdyraga: Instead of updating the minimum stake schedule, can we go with the original values and split this single test into multiple `it`s where we'll cover all corner cases?
```
describe("minimumStake()", async () => {
  it("returns 100k when the schedule starts", ...
  it("returns 100k just before the first step ends", ...
  it("returns 50k when the second step starts"...
  ...
}
```

We don't need to cover all steps but having the first two, the last two and something in between could be helpful.

solidity/contracts/TokenStaking.sol
```diff
@@ -20,7 +20,12 @@ contract TokenStaking is StakeDelegatable {
 
     // Minimum amount of KEEP that allows sMPC cluster client to participate in
     // the Keep network. Expressed as number with 18-decimal places.
-    uint256 public minimumStake = 200000 * 1e18;
+    // Initial minimum stake is higher than the final and lowered periodically based
+    // on the amount of steps and the length of the minimum stake schedule in blocks.
+    uint256 public minimumStakeScheduleStart;
+    uint256 public minimumStakeSchedule = 86400 * 365 * 2; // 2 years in seconds (seconds per day * days in a year * years)
+    uint256 public minimumStakeSteps = 10;
+    uint256 public minimumStakeBase = 10000 * 1e18;
```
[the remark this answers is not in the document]
eth-r: The base is 10,000 and it gets multiplied by the number of remaining steps, so `10 * 10000` first, then stepping down to `9 * 10000`, etc

ngrinkevich:
@pdyraga @eth-r ready!

solidity/test/TestMinimumStake.js
```diff
@@ -0,0 +1,74 @@
...
+    })
+
+    it("returns max value right before the next schedule step", async () => {
+      let minimumStakeScheduleStart = await stakingContract.minimumStakeScheduleStart();
+      let timeForStepOne = minimumStakeScheduleStart.add(minimumStakeSchedule.div(minimumStakeSteps))
+      // Rounding timestamp jump to 1 minute less (looks like increaseTime() can occasionally add extra seconds)
+      await increaseTime(timeForStepOne.toNumber() - await latestTime() - duration.minutes(1))
+      expect(await stakingContract.minimumStake()).to.eq.BN(
+        minimumStakeBase.mul(minimumStakeSteps),
+        "Unexpected minimum stake amount"
```
eth-r: I think it would be clearer if we had this like `expect(await stakingContract.minimumStake()).to.eq.BN(keepTokens(100000))` (where `keepTokens(n) = n.mul(1e18)`). Currently this is basically reimplementing the exact same code it's testing.

pdyraga: Agreed, we should hardcode the expected values for this test.

ngrinkevich: Done! @pdyraga @eth-r pls check

solidity/contracts/TokenStaking.sol
eth-r: The minimum stake schedule could be packed into a single storage field with the following:
```
uint64 public minimumStakeScheduleStart;
uint64 public minimumStakeSchedule = 86400 * 365 * 2;
uint32 public minimumStakeSteps = 10;
uint96 public minimumStakeBase = 10000 * 1e18;
```

ngrinkevich: since we're going to store it only once and then it's read only, extra gas for unpacking probably not worth it?

eth-r: With read-only we can still save 2400 gas from the `SREAD`s at the cost of a couple dozen gas for unpacking. Not really critical but can be nice.

pdyraga:
This part is still missing:

> An additional requirement is to capture the minimum stake when group selection starts. This is to ensure the same staking weight divisor is applied for all member candidates participating.

pdyraga:
We can do it in a separate PR but until we do, we can't close #1368.

ngrinkevich:
>An additional requirement is to capture the minimum stake when group selection starts. This is to ensure the same staking weight divisor is applied for all member candidates participating.

oops sorry missed that, will do asap

solidity/test/TestMinimumStake.js
```diff
@@ -0,0 +1,75 @@
...
+    it("returns max value when the schedule starts", async () => {
+      expect(await stakingContract.minimumStake()).to.eq.BN(
+        web3.utils.toBN(100000).mul(keepDecimals),
+        "Unexpected minimum stake amount"
+      );
+    })
+
+    it("returns max value right before the next schedule step", async () => {
+      let minimumStakeScheduleStart = await stakingContract.minimumStakeScheduleStart();
+      let timeForStepOne = minimumStakeScheduleStart.add(minimumStakeSchedule.div(minimumStakeSteps))
```
pdyraga: What do you think about extracting `stepDuration` to a separate variable? I think it's easier to follow:
```
let minimumStakeScheduleStart = await stakingContract.minimumStakeScheduleStart();
let stepDuration = minimumStakeSchedule.div(minimumStakeSteps)
let timeForStepOne = minimumStakeScheduleStart.add(stepDuration)
```

solidity/test/TestMinimumStake.js
```diff
@@ -0,0 +1,75 @@
...
+
+  describe("minimumStake", async () => {
+    it("returns max value when the schedule starts", async () => {
+      expect(await stakingContract.minimumStake()).to.eq.BN(
+        web3.utils.toBN(100000).mul(keepDecimals),
+        "Unexpected minimum stake amount"
+      );
+    })
+
+    it("returns max value right before the next schedule step", async () => {
```
pdyraga: Can we add one more test that asserts the value right after the first step ends?

solidity/contracts/TokenStaking.sol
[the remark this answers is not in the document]
ngrinkevich: oh yes good idea

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -61,6 +61,10 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
     // TODO: replace with a secure authorization protocol (addressed in RFC 11).
     TokenStaking internal stakingContract;
 
+    // Captures the minimum stake when group selection starts. This is to ensure the
+    // same staking weight divisor is applied for all member candidates participating.
+    uint256 internal groupSelectionMinimumStake;
```
pdyraga: What do you think about putting this value in `GroupSelection.Storage groupSelection`? 
We already have there `ticketSubmissionStartBlock` and `inProgress`

solidity/contracts/TokenStaking.sol
```diff
@@ -63,6 +68,22 @@ contract TokenStaking is StakeDelegatable {
...
+    /**
+     * @notice Returns minimum amount of KEEP that allows sMPC cluster client to
+     * participate in the Keep network. Expressed as number with 18-decimal places.
+     * Initial minimum stake is higher than the final and lowered periodically based
+     * on the amount of steps and the length of the minimum stake schedule in seconds.
+     */
+    function minimumStake() public view returns (uint256) {
+        if (block.timestamp < minimumStakeScheduleStart.add(minimumStakeSchedule)) {
+            uint256 currentStep = minimumStakeSteps.mul(
+                block.timestamp.sub(minimumStakeScheduleStart)).div(minimumStakeSchedule);
```
pdyraga: Nitpick but this formatting is weird 🤔 Can we e.g.:
```
uint256 currentStep = minimumStakeSteps.mul(
                block.timestamp.sub(minimumStakeScheduleStart)
).div(minimumStakeSchedule);
```

solidity/test/random_beacon_operator/TestSlashing.js
```diff
@@ -129,7 +129,7 @@ contract('KeepRandomBeaconOperator/Slashing', function(accounts) {
   it("should seize no more than available operator's amount", async () => {
     let tattletaleBalanceBeforeSeizing = await token.balanceOf(tattletale)
     
-    let amountToSeize = largeStake.add(web3.utils.toBN(100)); // 400000000000000000000100
+    let amountToSeize = largeStake.add(web3.utils.toBN(100));
```
pdyraga: Can we add `// 200000000000000000000100` comment back here? It helps to understand where does the equation for expected tattletale reward below.

Issue #1502, opened 2020-03-23
Adjust start.sh script to behave the same as the one from keep-ecdsa


pdyraga:
`start.sh` script from `keep-ecdsa` presents all available config files and lets to choose one by number. This approach is much more convenient than having to type full path as in the case of `keep-core` `start.sh`.

We should update `keep-core` `start.sh` to behave the same way as the one from `keep-ecdsa` as well as update developer's documentation in this repo to mention `install.sh` and `start.sh` scripts. They are too convenient to hide them.

```
$ ./scripts/start.sh
Enter ethereum accounts password [password]:
Enter path to keep-ecdsa config files directory [/Users/piotr/go/src/github.com/keep-network/keep-ecdsa/configs]:

Select client config file:
1) config.1.toml
2) config.2.toml
3) config.3.toml
4) config.toml
1

Client config file: "/Users/piotr/go/src/github.com/keep-network/keep-ecdsa/configs/config.1.toml"

Select log level [info]:
1) info
2) debug
3) custom...
2
Log level: "debug"

Starting keep-ecdsa client...
```

PR #1508, opened 2020-03-23
Displaying available config files to run a client

dimpar:
Closes https://github.com/keep-network/keep-core/issues/1502

In this PR we are updating `start.sh` script to show all the available config `.toml` files with options to select one.

Updating docs with a simple instruction how to run the scripts.


dimpar:
Tested by running the scripts and selecting a confg file from the list.

README.adoc
```diff
@@ -68,3 +68,38 @@ The Keep Go client.
 
 It runs the Keep network’s random beacon and participates in related
 computations.
+
+== Quick installation
```
pdyraga: I think this one should be a part of `docs/development/README`

README.adoc
```diff
@@ -68,3 +68,38 @@ The Keep Go client.
...
+
+To quickly install and start a single client use the installation
```
pdyraga: It is not necessarily about a single client. One can start as many clients as they want with those scripts.

README.adoc
```diff
@@ -68,3 +68,38 @@ The Keep Go client.
...
+
+....
+./scripts/start.sh
+....
+
+The `+start.sh+` script will ask you to provide:
+
+* ethereum password
+* path to `keep-core` config files directory
+* select a config `.toml` file for your client
```
pdyraga:
```
... will ask you to:
- provide ...
- provide ...
- select ...
```

Right now, we have `to provide select` for the last bullet.

scripts/start.sh
```diff
@@ -4,14 +4,35 @@ set -e
...
-CONFIG_FILE_PATH_DEFAULT=$(realpath -m $(dirname $0)/../config.toml)
+CONFIG_DIR_PATH_DEFAULT=$(realpath -m $(dirname $0)/../)
 
 # Read user inputs.
 read -p "Enter ethereum accounts password [$KEEP_ETHEREUM_PASSWORD_DEFAULT]: " ethereum_password
 KEEP_ETHEREUM_PASSWORD=${ethereum_password:-$KEEP_ETHEREUM_PASSWORD_DEFAULT}
 
-read -p "Enter path to keep-core client config [$CONFIG_FILE_PATH_DEFAULT]: " config_file_path
-CONFIG_FILE_PATH=${config_file_path:-$CONFIG_FILE_PATH_DEFAULT}
+read -p "Enter path to keep-core config files directory [$CONFIG_DIR_PATH_DEFAULT]: " config_dir_path
```
pdyraga: I think we need to do the same in `install.sh` and keep config files in `configs` directory by default just like we do for `keep-ecdsa`. This approach is much more convenient. A single `./install.sh` modifies all the configuration files, and then we can start any number of clients with `./start.sh` without any additional work. 

The current workflow is weird: we first modify a single config file and then select this config file again when starting.

dimpar: okay, done

README.adoc
dimpar: changed

README.adoc
dimpar: Hm, not sure if a `single client` means _only 1 client_ in this context. I would interpret it as a single client at a time, cause the script won't run multiple clients after execution. 

Anyway, I removed "single" just to avoid any confusion.

README.adoc
dimpar: Okay, moved to `docs/development/README`

docs/development/README.adoc
```diff
@@ -28,6 +28,41 @@ go get -u github.com/gogo/protobuf/protoc-gen-gogoslick
 Run `+go generate ./.../+` from `keep-core` and if everything is fine, start the
 Keep client with `go run main.go`.
```
pdyraga: What do you think about moving `Setting up local Keep network` part between `Building` and `Quick Installation`? 

The developer needs to have their accounts set up and Ethereum client running to execute `Quick Installation` and this part is covered by `Setting up local Keep network`.

scripts/install.sh
```diff
@@ -39,9 +39,12 @@ KEEP_CORE_SOL_ARTIFACTS_PATH=$(realpath $KEEP_CORE_SOL_PATH/build/contracts)
 printf "${LOG_START}Initializing contracts...${LOG_END}"
 truffle exec scripts/delegate-tokens.js --network local
 
-printf "${LOG_START}Updating keep-core client config...${LOG_END}"
-KEEP_CORE_CONFIG_FILE_PATH=$KEEP_CORE_CONFIG_FILE_PATH \
-    truffle exec scripts/lcl-client-config.js --network local
+config_files=($KEEP_CORE_CONFIG_DIR_PATH/*.toml)
+printf "${LOG_START}Updating keep-core client configs...${LOG_END}"
+for f in "${config_files[@]}"; do
+    KEEP_CORE_CONFIG_FILE_PATH="/${f##/}" \
```
pdyraga: What's the advantage of this approach over the one in `keep-ecdsa`:
```
for CONFIG_FILE in $KEEP_ECDSA_CONFIG_DIR_PATH/*.toml
do
  KEEP_ECDSA_CONFIG_FILE_PATH=$CONFIG_FILE \
```
?
🤔

dimpar: No particular reason. I didn't notice we have similar loop at `keep-ecdsa/initialize.sh` and just implemented my own `for-each`.

Refactored to the one we currently have in `keep-ecdsa`.

docs/development/README.adoc
dimpar: Sure, makes sense

scripts/install.sh
```diff
@@ -3,22 +3,22 @@ set -e
 
 # Dafault inputs.
 KEEP_ETHEREUM_PASSWORD_DEFAULT="password"
-CONFIG_FILE_PATH_DEFAULT=$(realpath -m $(dirname $0)/../config.toml)
+CONFIG_DIR_PATH_DEFAULT=$(realpath -m $(dirname $0)/../configs)
```
Shadowfiend: I would love it if we could avoid `realpath`—it relies on a newer version of bash than comes with macOS by default, so it requires folks to immediately install something else in the default environment we support :/

Rather than try to guess at what directory a script is invoked from (this is still error-prone, it turns out, because shells), thoughts on just assuming it's invoked as `scripts/install.sh` as the instructions note? I've moved a couple of other scripts to this approach, though they're more utility scripts than day-to-day dev scripts.

dimpar: I just found that `realpath` was [committed](https://github.com/coreutils/coreutils/commit/77ea441f79aa115f79b47d9c1fc9c0004c5c7111) 8 years ago as part of GNU coreutils and thinking that shells _should_ include it by now but I'm far from being an expert on what includes what :)

> thoughts on just assuming it's invoked as `scripts/install.sh` as the instructions note?

Sorry, I think I might need a bit of clarification here. Instead of:
```
CONFIG_DIR_PATH_DEFAULT=$(realpath -m $(dirname $0)/../configs)
```
should we replace it with:
```
CONFIG_DIR_PATH_DEFAULT="../configs"
```
?

If not, can you pls point to your script where you recently made a similar change?

Shadowfiend: Yep, exactly that!

macOS ships an old version of bash due to a license shift (GPL v2 -> v3), basically. The default shell for new users as of Catalina and going forward is zsh (partly) for this reason.

dimpar: I've changed the scripts, ready for another shot.

PR #1509, opened 2020-03-23
CircleCi: Disable Tenderly push

sthompson22:
Tenderly push is presumably failing due to imports that have the same name but different paths:

```
@openzeppelin/contracts-ethereum-package/contracts/GSN/Context.sol
openzeppelin-solidity/contracts/GSN/Context.sol
```
We're not 100% sure this is the issue, or the only issue.  We're also not sure why this worked, for a build or two.  For now we're disabling the push to get migrations passing in Circle again.


sthompson22:
Closing, we've implemented a work around that doesn't fail the build if tenderly push fails [here](https://github.com/keep-network/keep-core/pulls?q=is%3Apr+is%3Aclosed)

PR #1510, opened 2020-03-24
Optimized ticket submission protocol

### LINKED ISSUE #1431: Optimized ticket submission protocol for keep clients
The current protocol for ticket submission works as follows:

There is a ticket submission timeout of 12 blocks. The entire ticket submission window is divided into two phases: initial ticket submission (6 blocks) and reactive ticket submission (6 blocks). 

During the initial ticket submission, only tickets with a value below the natural threshold are submitted to the chain.

During the reactive ticket submission, all other staker's tickets are submitted but never more than what's been already submitted in the initial submission phase. Reactive ticket submission is skipped if during the initial ticket submission there was enough tickets submitted to form a group. Those tickets could be submitted by any stakers participating in a new group selection.

For example:
- client A submitted `5` tickets during initial submission phase
- client B submitted `4` tickets during initial submission phase
- client A submitted `64 - (5+4)` tickets during reactive submission phase
- client B submitted `64 - (5+4)` tickets during reactive submission phase

Obviously, this is less than optimal:
- in the reactive submission phase clients submit too many tickets
- although tickets are validated on-chain, off-chain clients do not check the value of tickets submitted in the initial submission phase by comparing them to the tickets they planned to submit in the reactive submission phase; it may happen that some smart client submits 64 tickets in the initial submission phase and all other clients will blindly trust those tickets are better than the ones they have and decide not to participate in the reactive submission phase.

lukasz-zimnoch:
Closes #1431

Here we introduce some optimizations to ticket submission protocol according to the [leading-zeros-based schedule](https://github.com/keep-network/keep-ecdsa/blob/106d7b800adbe1ca9d61b2197c5e04cf8d4697e5/docs/ticket-submission-strategy.adoc).

To minimize the submitter's cost by minimizing the number of redundant tickets that are not selected into the group, tickets are submitted in `N` rounds, each round taking `3` blocks.
As the basic principle, the number of leading zeros in the ticket value is subtracted from the number of rounds to determine the round the ticket should be submitted in:
- in round `0`, tickets with `N` or more leading zeros are submitted
- in round `1`, tickets with `N-1` or more leading zeros are submitted
(...)
- in round `N`, tickets with no leading zeros are submitted.

In each round, group member candidate needs to monitor tickets submitted by other candidates and compare them against tickets of the candidate not yet submitted to determine if continuing with ticket submission still makes sense.

After the last round, there is a `6` blocks mining lag allowing all outstanding ticket submissions to have a higher chance of being mined before the deadline.

During the work, one bug was found and fixed. It was related to `intValue` method (`ticket.go`) which returned the whole byte slice instead of only the first 8 bytes.


solidity/contracts/libraries/operator/GroupSelection.sol
```diff
@@ -404,4 +418,21 @@ library GroupSelection {
             delete self.candidate[self.tickets[i]];
         }
     }
+
+    function reactiveTicketSubmissionRounds(
+        uint256 possibleOperatorsCount,
+        uint256 groupSize
+    ) public view returns (uint256) {
```
pdyraga, line 425: Should this one be public or `internal`?

solidity/contracts/libraries/operator/GroupSelection.sol
```diff
@@ -94,6 +97,17 @@ library GroupSelection {
...
+        // Calculate possible operators count by dividing total token supply
+        // by current minimum stake.
+        uint256 possibleOperatorsCount = uint256(10**27).div(self.minimumStake);
+
+        // ticketSubmissionTimeout consists of:
+        // - one initial ticket submission round
+        // - N reactive ticket submission rounds
+        // - one additional waiting round
+        self.ticketSubmissionTimeout = self.ticketSubmissionRoundDuration
+            .mul(reactiveTicketSubmissionRounds(possibleOperatorsCount, self.groupSize).add(2));
```
pdyraga, line 110: Worth considering putting this code in a function. `internal view` functions are cheap, e.g.
```
function calculateTicketSubmissionTimeout() internal view returns (uint256) {
  // Calculate possible operators count by dividing total token supply
  // by current minimum stake.
  uint256 possibleOperatorsCount = uint256(tokenSupply).div(self.minimumStake);
  
  // ticketSubmissionTimeout consists of:
  // - one initial ticket submission round
  // - N reactive ticket submission rounds
  // - one additional waiting round
  uint256 reactiveRounds = reactiveTicketSubmissionRounds(
    possibleOperatorsCount, 
    self.groupSize
  )
  return self.ticketSubmissionRoundDuration.mul(reactiveRounds.add(2));
}
```

solidity/contracts/libraries/operator/GroupSelection.sol
```diff
@@ -404,4 +418,21 @@ library GroupSelection {
             delete self.candidate[self.tickets[i]];
         }
     }
+
+    function reactiveTicketSubmissionRounds(
```
pdyraga, line 422: We should add documentation explaining the magic we need to do inside.

solidity/contracts/libraries/operator/GroupSelection.sol
```diff
@@ -404,4 +418,21 @@ library GroupSelection {
...
+            operatorsExponent++;
+        }
+
+        uint256 groupsExponent = 1;
+        while(2**groupsExponent < groupSize){
+            groupsExponent++;
+        }
+
+        return operatorsExponent.sub(groupsExponent);
+    }
```
pdyraga, line 437: The maximum number of virtual stakers after the launch is `1e9 / 1e5 = 1e4`. The operator exponent is `2^13`. The group exponent is `2^5` so the reactive ticket submission rounds count is `2^13 - 2^5 = 8160`. Assuming a block time of 10 seconds, it's about 2 hours.

The maximum number of virtual stakers after 2 years is `1e9 / 1e4 = 1e5`. The operator exponent is `2^16`. The group exponent is `2^5` so the reactive ticket submission rounds count is `2^16 - 2^5 = 65504`. Assuming a block time of 10 seconds, it's about 18 hours.

Is this correct? 

I'd greatly appreciate @eth-r look on the on-chain part here.

Another question - do we need to have all those calculations here? Can't we hardcode the number of rounds to `10k` minimum stake case? This will simplify the on-chain code and we are probably not losing that much given that we'll eventually be at that point.

eth-r: The round count is logarithms so it's `13 - 5 = 8`. We should hardcode the values in a constant; if we upgrade the beacon operator contract we can compile in new values, so there is no real need for them to be storage.

solidity/contracts/libraries/operator/GroupSelection.sol
eth-r: Might as well make all of this `constant`. We also need to add the mining lag at the end, to ensure that submissions will be reliably mined even in the worst-case scenario.

solidity/contracts/libraries/operator/GroupSelection.sol
pdyraga: Yeah, they should be logarithms but I don't see logarithms in the code, that's why I was so confused. 

`13 - 5 = 8` is the launch case but `16 - 5 = 11` is the two-years case. It's just `3` rounds more. Let's just have a constant number of reactive rounds of `11` (plus initial phase and mining lag at the end) and if we don't use all rounds for a post-launch case, we'll treat `11-8=3` as an additional mining lag buffer at the end. 

What do you think?

solidity/contracts/libraries/operator/GroupSelection.sol
pdyraga: If we decide to hardcode the number of rounds as a constant, please ignore this comment.

solidity/contracts/libraries/operator/GroupSelection.sol
eth-r: The logarithm is in: `while(2**groupsExponent < groupSize)`

Every time `groupsExponent` is incremented, the value of `2**groupsExponent` doubles.

And yes, we should have a constant 11 rounds plus the mining lag. We don't need the initial phase separately from the zero round.

pdyraga: > The logarithm is in: `while(2**groupsExponent < groupSize)`
> 
> Every time `groupsExponent` is incremented, the value of `2**groupsExponent` doubles.

Ah, of course.

> And yes, we should have a constant 11 rounds plus the mining lag. We don't need the initial phase separately from the zero round.

Not sure if I follow this one. Don't we need the `< natural_threshold` round?

eth-r: The zero round is the natural threshold round; operators with 11 or more leading zeros all submit in the zeroth round.

pdyraga: Synced with @eth-r out-of-band. The summary is:

We have `v=11` rounds. It is a constant, no need to calculate it in the contract. We start indexing rounds from `0` and we do:
- `round 0`: submit tickets with `11` or more leading zeros
- `round 1`: submit tickets with `10` leading zeros
- `round 2`: submit tickets with `9` leading zeros
...
- round `11`: submit tickets with `0` leading zeros

(we submit in the given round only if it still makes sense)

We no longer need the initial submission phase and we no longer calculate the natural threshold. We submit tickets based on the leading zeros and that's all. 

Round 11 is different from the other rounds; its duration has to be the entire mining lag even if other rounds are shorter. We have 11 rounds of `roundDuration` and then a post-round of `miningLag`. `miningLag = 6, roundDuration = 3`

pkg/beacon/relay/config/config.go
```diff
@@ -9,9 +9,6 @@ type Chain struct {
 	// HonestThreshold is the minimum number of active participants behaving
 	// according to the protocol needed to generate a new relay entry.
 	HonestThreshold int
-	// TicketSubmissionTimeout is the duration (in blocks) the staker has to
-	// submit any tickets to candidate to a new group.
-	TicketSubmissionTimeout uint64
```
pdyraga, line 14: We can keep the number of rounds, round duration, and mining lag as constants (I wonder if it's possible to keep them in a library... 🤔) and reference them from config. This should be cheaper and consume less bytecode on the contract-side.

solidity/contracts/libraries/operator/GroupSelection.sol
```diff
@@ -94,6 +97,17 @@ library GroupSelection {
         self.inProgress = true;
         self.seed = _seed;
         self.ticketSubmissionStartBlock = block.number;
+
+        // Calculate possible operators count by dividing total token supply
+        // by current minimum stake.
+        uint256 possibleOperatorsCount = uint256(10**27).div(self.minimumStake);
```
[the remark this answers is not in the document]
lukasz-zimnoch: Not applicable now as we removed the dynamic submission timeout.

solidity/contracts/libraries/operator/GroupSelection.sol
lukasz-zimnoch: Not applicable now as we removed the dynamic submission timeout.

solidity/contracts/libraries/operator/GroupSelection.sol
lukasz-zimnoch: Not applicable now as we removed the dynamic submission timeout.

solidity/contracts/libraries/operator/GroupSelection.sol
lukasz-zimnoch: Not applicable now as we removed the dynamic submission timeout.

solidity/contracts/libraries/operator/GroupSelection.sol
lukasz-zimnoch: Refactored the code according to the above comments. But, regarding the on-chain part, I've just increased the submission timeout constant and keep the contract not aware of rounds. I've done that because the contract doesn't need to know about them, so we can avoid problems with contract size if we add some additional fields (even if we put new fields in `GroupSelection` library we must add some getters in operator contract).

pkg/beacon/relay/config/config.go
lukasz-zimnoch: `TicketSubmissionTimeout` will not be removed from `config.go`. Regarding other constants see [my comment above](https://github.com/keep-network/keep-core/pull/1510#discussion_r402240710).

solidity/contracts/libraries/operator/GroupSelection.sol
pdyraga: Sounds like a good move given that it's the client who is interested in the optimization, not the chain.

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -173,7 +173,12 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
         groups.stakingContract = TokenStaking(_stakingContract);
         groups.groupActiveTime = TokenStaking(_stakingContract).undelegationPeriod();
 
-        groupSelection.ticketSubmissionTimeout = 12;
+        // ticketSubmissionTimeout is calculated as follow:
+        // - there is 11 submission rounds
+        // - submission round duration is 6 blocks
```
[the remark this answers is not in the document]
lukasz-zimnoch: I think it's to short for Ropsten.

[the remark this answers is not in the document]
pdyraga: Can we start e.g. 10 clients against Ropsten and do some experiments around those values? We should see what are the mining lags and how many redundant tickets are submitted.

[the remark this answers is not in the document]
lukasz-zimnoch: Already done that with two clients and observed some problems when rounds had 3 blocks because ticket submission transactions were mined with bigger lag. But OK,  I will check again with 10 clients now.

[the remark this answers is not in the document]
eth-r: If those times are fine for mainnet, does it being too short for ropsten cause any other problems apart from redudant tickets?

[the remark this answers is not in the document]
pdyraga: > If those times are fine for mainnet, does it being too short for ropsten cause any other problems apart from redudant tickets?

The thing is that we are not sure if those numbers are fine for mainnet. And we can't experiment on mainnet because of some other issues. We need to make sure it works well on Ropsten and add some safety margin.

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -27,21 +32,6 @@ type Result struct {
...
-// gas expenditure.
-//
-// During the reactive ticket submission, all other staker's tickets are
-// submitted. Reactive ticket submission is skipped if during the initial
-// ticket submission there was enough tickets submitted to a chain to form
-// a group. Those tickets could be submitted by any stakers participating in
-// a new group selection.
-//
 // The function never submits more tickets than the group size.
 func CandidateToNewGroup(
```
pdyraga, line 36: It's worth adding documentation explaining how ticket submission works. We can borrow [the one](https://github.com/keep-network/keep-core/pull/1510/files#r402287010) from the chain.

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -56,223 +46,214 @@ func CandidateToNewGroup(
...
-	// tickets than the group size.
-	var numberOfTicketsToSubmit int
-	if len(initialSubmissionTickets) > chainConfig.GroupSize {
-		numberOfTicketsToSubmit = chainConfig.GroupSize
-	} else {
-		numberOfTicketsToSubmit = len(initialSubmissionTickets)
+	quitTicketSubmission := make(chan struct{})
+
+	ticketSubmissionRounds := (chainConfig.TicketSubmissionTimeout /
+		ticketSubmissionRoundDuration) - 2
```
pdyraga, line 92: Why `-2`? Is it a mining lag? If so, let's make it a `const` next to `ticketSubmissionRoundDuration`.

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -56,223 +46,214 @@ func CandidateToNewGroup(
...
 	return startTicketSubmission(
-		initialSubmissionTickets,
-		reactiveSubmissionTickets,
+		tickets,
 		relayChain,
 		blockCounter,
 		chainConfig,
 		startBlockHeight,
 		onGroupSelected,
+		ticketSubmissionRoundDuration,
```
pdyraga, line 69: This is a constant, we don't need to pass it as a parameter.

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -56,223 +46,214 @@ func CandidateToNewGroup(
...
 ) error {
-	initialSubmissionTimeout, err := blockCounter.BlockHeightWaiter(
-		startBlockHeight + chainConfig.TicketSubmissionTimeout/2,
-	)
-	if err != nil {
-		return err
-	}
-
-	reactiveSubmissionTimeout, err := blockCounter.BlockHeightWaiter(
+	ticketSubmissionTimeoutChannel, err := blockCounter.BlockHeightWaiter(
```
pdyraga, line 82: Why don't we use it as `quitTicketSubmission` and avoid creating additional channel?

pkg/beacon/relay/groupselection/groupselection.go
pdyraga: Also, isn't the number of rounds `(timeout - miningLag) / roundDuration`?

solidity/contracts/KeepRandomBeaconOperator.sol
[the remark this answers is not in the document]
eth-r: Makes sense. We could make it 4/8 for 52 blocks, or 5/5 for 60 blocks.

pkg/beacon/relay/groupselection/groupselection.go
lukasz-zimnoch: For test purposes, we need a smaller number in order to avoid long test execution.

pkg/beacon/relay/groupselection/groupselection.go
lukasz-zimnoch: We must close the channel in order to terminate all submission goroutines not just pass a value which will be received only by one goroutine. I think we should avoid closing `ticketSubmissionTimeoutChannel` because we are on the receiver side here and as a rule of thumb in Go a receiver should never close the channel.

pkg/beacon/relay/groupselection/groupselection.go
lukasz-zimnoch: > Why -2? Is it a mining lag? If so, let's make it a const next to ticketSubmissionRoundDuration.

Yeah this is a mining lag which consists of two rounds, I will extract it to a constant.

> Also, isn't the number of rounds (timeout - miningLag) / roundDuration?

This is basically the same.

solidity/contracts/KeepRandomBeaconOperator.sol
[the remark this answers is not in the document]
lukasz-zimnoch: @pdyraga, made some tests on Ropsten with 10 clients and round duration of 3 blocks (with 6 blocks mining lag at the end). Here are the results:

```
operator 1: 6 members / 9 tickets submitted
operator 2: 7 members / 9 tickets submitted
operator 3: 6 members / 9 tickets submitted
operator 4: 4 members / 7 tickets submitted
operator 5: 5 members / 8 tickets submitted
operator 6: 9 members / 10 tickets submitted
operator 7: 5 members / 8 tickets submitted
operator 8: 6 members / 9 tickets submitted
operator 9: 7 members / 10 tickets submitted
operator 10: 9 members / 12 tickets submitted
```

I didn't see any problems this time, so maybe we can use the round duration of 3 blocks and if we observe problems in the future we can always change it to 6 blocks. What do you think?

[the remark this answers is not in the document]
eth-r: Sounds reasonable, although if we get problems I'd increment the round duration from 3 -> 4 -> 5 -> 6 until problems go away, instead of jumping straight to 6; the round duration doesn't need to be the mining lag divided by an integer.

pkg/beacon/relay/groupselection/groupselection.go
eth-r: It would probably be better to calculate it as `(timeout - miningLag) / roundDuration` to cover the case where `roundDuration != miningLag / 2`

solidity/contracts/KeepRandomBeaconOperator.sol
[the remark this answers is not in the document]
pdyraga: Yep, starting with `3` makes sense especially that our result publication block step is also set to 3.

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -173,7 +173,12 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
...
+        // - a mining lag (2 rounds) is added to the end
+        // So: (11 * 6) + (2 * 6) = 66 + 12 = 78
```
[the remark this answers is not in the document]
lukasz-zimnoch: Sounds good! Thanks!

solidity/contracts/KeepRandomBeaconOperator.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Ok, let's do it!

pkg/beacon/relay/groupselection/groupselection.go
lukasz-zimnoch: Ok, this is no longer an argument as we changed the round to `3` blocks. I will change.

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -16,6 +17,14 @@ import (
 
 var logger = log.Logger("keep-groupselection")
 
+const (
+	// Duration of one ticket submission round in blocks.
+	ticketSubmissionRoundDuration = 3
+
+	// Duration of the mining lag in blocks.
+	miningLag = 6
+)
```
pdyraga, line 26: Since `chainConfig.TicketSubmissionTimeout` is supplied as a parameter from the outside, we should add some sanity check:
```
// Recommended parameters all clients should use to minimize their expenses.
// It is not a must to obey but it is nice and polite. And being nice to others
// helps in reducing own costs because other clients should respect the same
// protocol.
const (
	// the number of blocks one round takes
	roundDuration = uint64(3)
	// the delay in blocks after all rounds complete to ensure all transactions
	// are mined
	miningLag     = uint64(6)
)

// calculateRoundsCount takes the on-chain ticket submission timeout
// and calculates the number of rounds for ticket submission. If it is not 
// possible to use the recommended round duration and mining lag because the 
// supplied timeout is too short, function returns an error.
func calculateRoundsCount(submissionTimeout uint64) (uint64, error) {
	if submissionTimeout-miningLag <= roundDuration {
		return 0, fmt.Errorf("submission timeout is too short")
	}

	return (submissionTimeout - miningLag) / roundDuration, nil		
}
```

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -16,6 +17,14 @@ import (
 
 var logger = log.Logger("keep-groupselection")
 
+const (
+	// Duration of one ticket submission round in blocks.
+	ticketSubmissionRoundDuration = 3
```
pdyraga, line 22: Let's simplify it to just `roundDuration`.

pkg/beacon/relay/groupselection/groupselection.go
pdyraga: Not worth doing it now but I'll drop the idea in the comment in case we'll be interested in longer round durations in the future.
```
type roundParameters struct {
	// the total number of rounds
	roundCount uint64
	// the number of blocks one round takes
	roundDuration uint64
	// the delay in blocks after all rounds complete to ensure all transactions
	// are mined
	miningLag uint64
}

// Recommended parameters all clients should use to minimize their expenses.
// It is not a must to obey but it is nice and polite. And being nice to others
// helps in reducing own costs because other clients should respect the same
// protocol.
const (
	defaultRoundDuration = uint64(3)
	defaultMiningLag     = uint64(6)
)

// calculateRoundCount takes the on-chain ticket submission timeout
// and calculates the expected round parameters. If it is not possible to use
// 
func calculateRoundParams(submissionTimeout uint64) (*roundParameters, error) {
	if submissionTimeout-defaultMiningLag <= defaultRoundDuration {
		return nil, fmt.Errorf("submission timeout is too short")
	}

	return &roundParameters{
		roundCount:    (submissionTimeout - defaultMiningLag) / defaultRoundDuration,
		roundDuration: defaultRoundDuration,
		miningLag:     defaultMiningLag,
	}, nil
}
```

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -87,192 +92,188 @@ func CandidateToNewGroup(
 }
 
 func startTicketSubmission(
```
pdyraga, line 94: Let's maybe rename it to `submitTickets`. `startTicketSubmission` suggests there is some other function that will end the ticket submission.

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -87,192 +92,188 @@ func CandidateToNewGroup(
...
-	initialSubmissionTickets []*ticket,
-	reactiveSubmissionTickets []*ticket,
+	tickets []*ticket,
 	relayChain relaychain.GroupSelectionInterface,
 	blockCounter chain.BlockCounter,
 	chainConfig *config.Chain,
 	startBlockHeight uint64,
 	onGroupSelected func(*Result),
```
pdyraga, line 100: Since this function blocks until ticket submission is over, we don't need to pass `onGroupSelected` as a parameter and execute this function in `CandidateToNewGroup` instead. I think it's much cleaner approach:
```
func CandidateToNewGroup(
  // (...)
) error {
  tickets, err := generateTickets(...)
  // (...)
  err = submitTickets(...)
  // (...)
  
  selectedStakers, err := relayChain.GetSelectedParticipants()
  if err != nil {
    return fmt.Errorf(
      "could not fetch selected participants after submission timeout [%v]",
      err,
    )
  }

  go onGroupSelected(&Result{
    SelectedStakers:        selectedStakers,
    GroupSelectionEndBlock: ticketSubmissionEndBlockHeight,
  })
}
 
```

pdyraga: It makes even more sense if we consider how we should handle errors from `err = submitTickets(...)`. I don't think we should `return` but just `logger.Errorf` and continue with the execution of `CandidateToNewGroup`. See that in case of e.g. `WaitForBlockHeight` in round 4 of ticket submission, there is a chance tickets from rounds 1-3 qualified to a group. It does make sense to stop `submitTickets` processing on any error but we should still wait for ticket submission timeout in `CandidateToNewGroup` and execute `GetSelectedParticipants`:

```
err = submitTickets(...)
if err != nil {
  logger.Errorf(...). // <--- !!!
}

// wait till the end of the ticket submission in case submitTickets failed in the middle
// and there is still a chance we qualified to a group
ticketSubmissionTimeoutChannel, err := blockCounter.BlockHeightWaiter(
  startBlockHeight + chainConfig.TicketSubmissionTimeout,
)
if err != nil {
  return err
}
  
selectedStakers, err := relayChain.GetSelectedParticipants()
if err != nil {
    return fmt.Errorf(
      "could not fetch selected participants after submission timeout [%v]",
      err,
    )
}
```

What do you think about it?

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -24,25 +33,28 @@ type Result struct {
...
+// CandidateToNewGroup attempts to generate and submit tickets for the
+// staker to join a new group.
 //
-// During the initial ticket submission, only tickets with a value below the
-// natural threshold are submitted to the chain. Those tickets have the highest
-// chance of being selected to the group and this way we minimize staker's
-// gas expenditure.
+// To minimize the submitter's cost by minimizing the number of redundant
+// tickets that are not selected into the group, tickets are submitted in
+// 11 rounds, each round taking 3 blocks.
```
pdyraga, line 41: Actually... we compute the number of rounds. We should say `tickets are submitted in N rounds` and later:
```
// - in round 0, tickets with N or more leading zeros are submitted
// - in round 1, tickets with N-1 or more leading zeros are submitted
// (...)
// - in round N, tickets with no leading zeros are submitted.
```

pkg/beacon/relay/groupselection/groupselection.go
pdyraga: The thing is that I don't think we need more than one goroutine now. The reason for triggering `submitTickets` in a separate goroutine is from ancient times. We were submitting a massive number of tickets, not even limited by a group size from what I remember. There is nothing expensive happening in `submitTickets` function, I think we can get rid of `quit <-chan struct{}` parameter from there and just submit tickets in a loop. 

It makes even more sense to do it this way so that we avoid potential nonce duplicate problems when two rounds overlap.

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -87,192 +92,188 @@ func CandidateToNewGroup(
...
+		roundStartBlock := startBlockHeight + roundStartDelay
+		roundLeadingZeros := ticketSubmissionRounds - roundIndex
+
+		logger.Infof(
+			"ticket submission round [%v] will start at "+
+				"block [%v] and cover tickets with [%v] leading zeros",
+			roundIndex,
+			roundStartBlock,
+			roundLeadingZeros,
+		)
```
pdyraga, line 125: 👌

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -87,192 +92,188 @@ func CandidateToNewGroup(
...
+		if err != nil {
+			return err
+		}
+
+		logger.Infof(
+			"ticket submission round [%v] submitting "+
+				"[%v] tickets",
+			roundIndex,
+			len(candidateTickets),
+		)
```
pdyraga, line 148: Perfect. We should log all of it on info level 👌

pkg/beacon/relay/groupselection/groupselection.go
pdyraga: Honestly, if we move to just a loop and get rid of a goroutine I am not sure if we need `ticketSubmissionTimeoutChannel` and `closeChannel` in this function at all. At maximum, we'll submit 64 tickets, all in the last round. And we have 3 + 6 = 9 blocks for it. Even for a local environment, this should be more than enough.

pkg/beacon/relay/groupselection/ticket_submission.go
```diff
@@ -41,7 +41,7 @@ func submitTickets(
 
 func toChainTicket(ticket *ticket) (*relaychain.Ticket, error) {
 	return &relaychain.Ticket{
-		Value: ticket.intValue(),
+		Value: new(big.Int).SetBytes(ticket.value[:]),
```
pdyraga, line 44: Why don't we use `intValue` here? I think we are fine with it now as you fixed the logic there.

lukasz-zimnoch: Nope, the `Value` of `relaychain.Ticket` must contain the whole `value` array of `groupselection.ticket` not just the first `8` bytes returned by `intValue`

pkg/beacon/relay/groupselection/groupselection.go
lukasz-zimnoch: Ok, I checked if nothing wrong happens when we remove the goroutine and everything seems to be good. Regarding `ticketSubmissionTimeoutChannel` we need it in order to prevent calling `relayChain.GetSelectedParticipants()` too early.  Though, I agree with [this comment](https://github.com/keep-network/keep-core/pull/1510#discussion_r402879191) and I will refactor this in the proposed way.

pkg/beacon/relay/groupselection/groupselection_test.go
```diff
@@ -14,180 +18,385 @@ import (
...
+			}
+
+			if !reflect.DeepEqual(test.expectedChainTickets, chainTickets) {
+				t.Fatalf(
+					"unexpected chain tickets\n"+
+						"expected: [%v]\nactual:   [%v]",
+					test.expectedChainTickets,
+					chainTickets,
+				)
+			}
```
pdyraga, line 398: I think we should get rid of this part as we are testing `toChainTicket`, `SubmitTicket` and `GetSubmittedTicktets` behaviour here. Checking which tickets and at which round were returned by `roundCandidateTickets` should be enough for this test.

pkg/beacon/relay/groupselection/groupselection.go
lukasz-zimnoch: Good point! Changed this as you proposed.

pkg/beacon/relay/groupselection/groupselection_test.go
pdyraga: In the previous approach, we were testing the entire `startTicketSubmission`. That's why we have the entire `stubGroupInterface` here. I think we have two options:

1. Apply the comment above and write additional unit test stressing `startTicketSubmission` with just one case making sure tickets are submitted in correct rounds. We can have just 2-3 rounds for this test.
2. Make this test stress the entire `startTicketSubmission`. We'll probably have to [parameterize round duration](https://github.com/keep-network/keep-core/pull/1510#discussion_r402868814) to do not keep this test running forever. 

I am personally in favor of the first approach as it seems simpler. You are the expert here so you decide though :)

pkg/beacon/relay/groupselection/ticket.go
```diff
@@ -79,6 +80,11 @@ func calculateTicketValue(
 	return keccak256Hash, nil
 }
 
+// Returns 8 starting bytes of ticket value as a big integer.
```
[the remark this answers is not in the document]
pdyraga: No, wait. You are right here. I need to go through this code one more time.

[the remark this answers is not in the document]
pdyraga: This is very misleading. On-chain we speak about value as the first 8 bytes. Off-chain, value is all 32 bytes. This code is right, the naming is wrong. But it's not something for this PR.

A kind request: can we copy the on-chain explanation of value bytes to `ticket` struct to avoid further confusion?

```
type ticket struct {
  // first 8 bytes: result of keccak256 cryptography hash
  // function on the combination of the group selection seed (previous
  // beacon output), staker-specific value (address) and virtual staker index.
  // next 20 bytes: a staker-specific value which is the address of the staker.
  // next 4 bytes: virtual staker index; a number within a range of 1 to staker's weight;
  value [32]byte // W_k

  proof *proof   // proof_k = Proof(Q_j, vs)
}
```

pkg/beacon/relay/groupselection/ticket_submission.go
pdyraga: You are right, resolving.

pkg/beacon/relay/groupselection/groupselection.go
lukasz-zimnoch: Excellent idea! Will change as you proposed.

pkg/beacon/relay/groupselection/ticket.go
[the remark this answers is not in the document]
lukasz-zimnoch: Sure, but the comment is not 100% right with the actual meaning of first `8` bytes, I think it should be:
```
type ticket struct {
	// Result of keccak256 cryptographic hash function on the combination of:
	// - group selection seed (previous beacon output)
	// - staker-specific value (address)
	// - virtual staker index;
	// From the chain's perspective it is composed of:
	// - first 8 bytes: ticket value
	// - next 20 bytes: a staker-specific value which is the address of the staker
	// - next 4 bytes: virtual staker index; a number within a range of 1 to staker's weight;
	value [32]byte // W_k

	proof *proof   // proof_k = Proof(Q_j, vs)
}
```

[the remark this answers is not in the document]
pdyraga: Nope. My first comment was correct. Please look at `packTicket` function in `ethereum.go`. We should be fine with 8-bytes value here.

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -87,192 +92,188 @@ func CandidateToNewGroup(
...
-						"entering reactive submission phase with [%v] "+
-						"additional tickets",
-					ticketsCount,
-					numberOfTicketsToSubmit,
-				)
+// roundCandidateTickets returns tickets which should be submitted in
+// given ticket submission round.
+func roundCandidateTickets(
+	relayChain relaychain.GroupSelectionInterface,
+	memberTickets []*ticket,
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, they need to be sorted. Actually they are returned sorted from `generateTickets` so I want to avoid sorting them twice. I will drop a comment.

pkg/beacon/relay/groupselection/ticket.go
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah but in this case we must change `ticket` struct and their `value` should hold only the first `8` bytes of the hash returned by `calculateTicketValue`. Let's address this problem in a separate pull request.

pkg/beacon/relay/groupselection/groupselection.go
```diff
@@ -87,192 +92,188 @@ func CandidateToNewGroup(
...
+		} else {
+			// If submitted tickets count is equal to the group
+			// size the candidate ticket can be added only if
+			// it is smaller than the highest submitted ticket.
+			// Note that, maximum length of submitted tickets slice
+			// will be exceeded and will be trimmed in next
+			// iteration.
+			highestSubmittedTicket := submittedTickets[len(submittedTickets)-1]
+			if candidateTicketValue < highestSubmittedTicket {
+				submittedTickets = append(
```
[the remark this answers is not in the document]
lukasz-zimnoch: Good point!

lukasz-zimnoch:
@pdyraga, all comments addressed! Ready for the next round.

pkg/beacon/relay/groupselection/groupselection_test.go
lukasz-zimnoch: Ok, done!

PR #1511, opened 2020-03-24
keep-dev: dev.*.network hosted zones

sthompson22:
Here we provide two public domains for use in the keep-dev environment
if we need them.  It may be that we want to expose applications to the
internet from our dev env for testing or sharing.  The immediate need
here was to expose the tbtc dapp with certificate so some testing could
be done.  That uses the dev.tbtc.network domain.  I went ahead and added
dev.keep.network while I was in here.

These domains are sub-domains which are NS aliased to their respective
parents on Cloudflare, keep.network and tbtc.network.


sthompson22:
This is deployed to the keep-dev env.

Issue #1501, opened 2020-03-23
Represent initialization and undelegation period in TokenStaking contract as timestamps


pdyraga:
As per #1490, customer promise for deposits is expressed in months (timestamps) and token staking periods are represented in blocks. We should update the initialization and undelegation period to use timestamps as well as make sure the rest of the code in `TokenStaking` contract is aligned with timestamps. 

As a part of this issue, we will have to modify KEEP token dashboard.

PR #1512, opened 2020-03-24
Convert initialization and undelegation parameters to timestamps

ngrinkevich:
Refs #1501

Updating for consistency reasons, fwiw initialization and undelegation params don't bring security risks of miner manipulation in this case.

PR #1513, opened 2020-03-25
 Represent initialization and undelegation period in token dashboard dapp as timestamps

r-czajkowski:
Depends on: #1512 
Closes: #1501


pdyraga:
#1512 has been merged to master

PR #1515, opened 2020-03-26
keep-test: cleanup internal ETH network configs

sthompson22:
We've been reducing the number of moving parts between our keep-dev and keep-test environments for some weeks now.  The primary motivator here is to reduce operational overhead while still providing the necessary resources for development and testing.

Part of that reduction came in the form of running an internal ETH network only in our keep-dev environment.  keep-test is now exclusively for public testnet deployments.  The internal ETH network that was running in keep-test was spun down a week or so ago, I just forgot to cleanup the config files.


sthompson22:
@nkuba This should be an easy one, just config cleanup.

Issue #1486, opened 2020-03-19
Typo in 'release tokens' on Grants


liz-shinn:
– Should read "Ready to _Release_" but let's change to just "Available"
– This will keep it from breaking onto 2nd line
– Style all violet + same type treatment: Work Sans, Medium, 14pt

![screenshot](https://user-images.githubusercontent.com/57226633/77113650-9aa15f80-6a01-11ea-932d-e48ad9e147c0.png)

Example:
<img width="484" alt="Image 2020-03-19 at 4 54 22 PM" src="https://user-images.githubusercontent.com/57226633/77114151-6a0df580-6a02-11ea-842a-9c35d8e56afb.png">

Issue #1390, opened 2020-02-26
Grant is getting unlocked, not vested


pdyraga:
To avoid legal complications we need to use a term of `unlocking` instead of `vesting`. This issue aims at renaming Solidity functions, fields etc. as well as updating the dApp.

PR #1516, opened 2020-03-26
Grant is getting unlocked not vested

r-czajkowski:
Closes: #1390
Closes: #1486 

In this PR the `vesting` term has been updated to `unlocking`.


solidity/dashboard/src/components/TokenGrantVestingSchedule.jsx
```diff
@@ -1,62 +0,0 @@
-import React from 'react'
```
pdyraga: It is unclear to me why do we remove this file. Is there any *unlocking* replacement or do we no longer need it?

r-czajkowski: This component is no longer needed.

PR #1518, opened 2020-03-27
keep-test: More ETH accounts

sthompson22:
In prep for some upcoming system tests we're adding 5 more ETH accounts to the pool of available accounts.

In addition I caveman'd a bash script to setup accounts via the `geth` cli.  I didn't feel like doing this manually, and I know we're going to need more in the future.  It's basic basic..does the trick but doesn't guard against most error conditions.

I've already updated the touched ConfigMaps.

### DIFF
--- a/infrastructure/eth-networks/scripts/create-eth-accounts.sh
+++ b/infrastructure/eth-networks/scripts/create-eth-accounts.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# Requires geth https://github.com/ethereum/go-ethereum/wiki/Installation-Instructions-for-Mac
+# Requres bip39-cli https://www.npmjs.com/package/bip39-cli
+
+# Please don't use this for mainnet accounts, it's for testing only.
+
+HELP="Usage: ./$(basename $0) -n <number of accounts> \ne.g. ./$(basename $0) -n 5"
+
+while getopts ":n:" opt; do
+  case $opt in
+    n  ) NUMBER_OF_ACCOUNTS=$OPTARG;;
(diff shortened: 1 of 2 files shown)


infrastructure/eth-networks/scripts/create-eth-accounts.sh
```diff
@@ -0,0 +1,39 @@
...
+
+for ((ACCOUNT_ORDINAL=0; $ACCOUNT_ORDINAL<$NUMBER_OF_ACCOUNTS; ACCOUNT_ORDINAL++))
+do
+  echo "=====Account $ACCOUNT_ORDINAL====="
+
+  ACCOUNT_PASSWORD=$(bip39-cli generate)
+
+  echo $ACCOUNT_PASSWORD > $PASSWORD_FILE
+
+  ACCOUNT=$(geth account new --keystore ./ --password $PASSWORD_FILE)
```
nkuba: Maybe we could replace `--keystore ./` with `--keystore ./keystore` so we don't end up with all the keys in this main directory next to other scripts?

nkuba:
Left one [non-blocking comment](https://github.com/keep-network/keep-core/pull/1518#discussion_r400790403).

infrastructure/eth-networks/scripts/create-eth-accounts.sh
sthompson22: Yeah I think the intent here, which wasn't clearly expressed is that none of the outputs of this script will be checked in directly.  We're just using the produced material to fill out the appropriate secrets and configmaps on the Kube side.

Next step here would be to extend the script into doing that filling out for us, and trashing the outputs.

PR #1519, opened 2020-03-29
Implement stake lock to an operator

ngrinkevich:
Closes: #1490 

Operator contracts can add locks to operator's stake to override the normal undelegation period.


eth-r:
@pdyraga ready for review

@ngrinkevich would you be able to take care of making ECDSA keeps use the locks and delegated authority?

solidity/test/utils/TestLockUtils.js
```diff
@@ -0,0 +1,64 @@
+import {createSnapshot, restoreSnapshot} from '../helpers/snapshot';
+
+const BN = web3.utils.BN
+const chai = require('chai')
+chai.use(require('bn-chai')(BN))
+const expect = chai.expect
+
+const LockStub = artifacts.require('./stubs/LockStub.sol');
+
+contract('LockStub', (accounts) => {
```
pdyraga, line 10: We should have `contract('LockUtils'` here. We are testing the library, the fact we need to do it through a stub contract is just an implementation detail.

solidity/test/utils/TestLockUtils.js
```diff
@@ -0,0 +1,64 @@
...
+    await locks.publicSetLock(alice, 1);
+    await locks.publicSetLock(bob, 2);
+    await locks.publicSetLock(carol, 3);
+    let creatorsPre = await locks.publicEnumerateCreators();
+    expect(creatorsPre[1]).to.equal(bob);
+
+    await locks.publicReleaseLock(bob);
+    let creatorsPost = await locks.publicEnumerateCreators();
+    expect(creatorsPost[1]).to.equal(carol);
+  })
```
pdyraga, line 63: Worth adding a test case for releasing when there is only one lock.

solidity/test/token_stake/TestTokenStakeLock.js
```diff
@@ -0,0 +1,247 @@
...
+      await undelegate(operator)
+      await expectThrowWithMessage(
+        stakingContract.recoverStake(operator),
+        "Can not recover locked stake"
+      )
+
+      await expectThrowWithMessage(
+        stakingContract.unlockStake(operator),
+        "Not authorized"
+      )
```
pdyraga, line 151: Should this check belong to this test?

solidity/test/token_stake/TestTokenStakeLock.js
```diff
@@ -0,0 +1,247 @@
...
+
+      // disable operator contract with panic button
+      await registry.disableOperatorContract(operatorContract)
+
+      await stakingContract.recoverStake(operator, {from: operator})
+      // ok, no revert
+    })
+
+    it("should be able to reduce the duration of existing locks", async () => {
+      await stakingContract.lockStake(operator, undelegationPeriod, {from: operatorContract})
```
pdyraga, line 185: Why `undelegationPeriod` and not e.g. `duration.minutes(5)`? I think we wanted to use just some lower value and using `undelegationPeriod` suggests there is something more going on here.

solidity/contracts/utils/LockUtils.sol
```diff
@@ -0,0 +1,77 @@
...
+
+    /// @notice Remove the lock of `creator`.
+    /// If no lock present, do nothing.
+    function releaseLock(
+        LockSet storage self,
+        address _creator
+    ) internal {
+        uint256 positionPlusOne = self.positions[_creator];
+        if (positionPlusOne != 0) {
+            uint256 memberCount = self.locks.length;
```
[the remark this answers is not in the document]
eth-r: That was a remnant of `SetUtils` variable naming, fixed in latest

ngrinkevich:
>@ngrinkevich would you be able to take care of making ECDSA keeps use the locks and delegated authority?

sure!

solidity/test/token_stake/TestTokenStakeLock.js
eth-r: Addressed in latest

solidity/test/token_stake/TestTokenStakeLock.js
eth-r: Removed in latest

solidity/contracts/TokenStaking.sol
```diff
@@ -259,6 +269,117 @@ contract TokenStaking is StakeDelegatable {
...
+    }
+
+    /**
+     * @dev Unlocks given operator stake. Only previously authorized operator
+     * contract can unlock the stake.
+     * @param operator Operator address.
+     */
+    function unlockStake(
+        address operator
+    ) public {
```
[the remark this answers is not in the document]
eth-r: Unlocking stake can't harm the operator or interfere with other operator contracts. Realistically we could get away with no ACL here whatsoever, but `isAuthorizedForOperator` implies that the contract must have been able to place a lock on the operator sometime in the past, thus sanity-checking out cases where someone other than an operator contract tries to unlock stake. We don't check if a lock is present because it might have been gotten removed with `releaseExpiredLock`.

eth-r:
Remaining questions:

- do we want to emit events when stake is locked/unlocked?
- do we want `activeStake` to return 0 when the stake is released (and thus could be recovered) even if it has not been recovered yet, to match the behavior of `slash/seize` which don't permit punishments when stake is released?
- if one operator contract is setting very long locks, unlucky operators may get stuck with a particularly long waiting period before their stake is released, yet unable to make use of their stake because they need to maintain their status as `undelegating` to prevent the contract from setting even longer locks; do we want to permit operators to announce "I want to start undelegating wrt this specific contract, but not others"?

I'm happy to postpone the last question for V2 and just tell operators to suck it up in V1, but the other two are fairly simple to address if desired.

pdyraga:
Let's do events, `activeStake` change, and update tests after moving to OpenZeppelin test engine and I am good to merge.

solidity/contracts/TokenStaking.sol
```diff
@@ -52,6 +52,9 @@ contract TokenStaking is StakeDelegatable {
     event RecoveredStake(address operator, uint256 recoveredAt);
     event TokensSlashed(address indexed operator, uint256 amount);
     event TokensSeized(address indexed operator, uint256 amount);
+    event StakeLocked(address indexed operator, address by, uint256 until);
```
pdyraga, line 55: Should we rename `by` to `lockCreator`?

solidity/contracts/TokenStaking.sol
```diff
@@ -52,6 +52,9 @@ contract TokenStaking is StakeDelegatable {
...
+    event LockReleased(address indexed operator, address by);
```
pdyraga, line 56: Can we rename `by` to `releasedBy`? `by` alone is a little confusing.

solidity/contracts/TokenStaking.sol
```diff
@@ -589,7 +589,13 @@ contract TokenStaking is StakeDelegatable {
 
         bool isActive = _isInitialized(operatorParams);
 
-        if (isAuthorized && isActive) {
+        bool stakeReleased = _isStakeReleased(
```
pdyraga, line 592: We should update the documentation to this function. Right now it says nothing about the fact the stake can't be released. We also need to update the documentation in `KeepRandomBeaconOperator.sol` for `hasMinimumStake` function.

eth-r: The documentation for `hasMinimumStake` in the operator contract doesn't really need updating because it doesn't refer to the specific meaning of active stake. Staking contract `activeStake` documentation is updated now.

PR #1530, opened 2020-03-30
Migration Fixes: Land 3 fixes to the Circle migration script

Shadowfiend:
This fixes migrations, which are currently broken, and maybe also gives them a small speed boost, without removing tenderly push from the equation.


sthompson22:
> Noting that this has been tested in a different branch.

https://github.com/keep-network/keep-core/tree/version-tether this branch 😬

PR #1531, opened 2020-03-30
Fix failing tests

ngrinkevich:
A couple of adjustments to help with `increaseTimeTo()` inconsistency.

### DIFF
--- a/solidity/test/token_stake/TestTokenStake.js
+++ b/solidity/test/token_stake/TestTokenStake.js
@@ -801,7 +801,7 @@ contract('TokenStaking', function(accounts) {
         {from: ownerOne}
       );
 
-      await increaseTimeTo(createdAt + initializationPeriod + delegationTime + 1)
+      await increaseTimeTo(createdAt + initializationPeriod + delegationTime + timeRoundMargin)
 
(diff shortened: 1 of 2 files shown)


ngrinkevich:
@pdyraga ready!

solidity/test/token_stake/TestTokenStake.js
```diff
@@ -801,7 +801,7 @@ contract('TokenStaking', function(accounts) {
         {from: ownerOne}
       );
 
-      await increaseTimeTo(createdAt + initializationPeriod + delegationTime + 1)
+      await increaseTimeTo(createdAt + initializationPeriod + delegationTime + timeRoundMargin)
```
pdyraga: Hmm... this is weird, shouldn't margin apply to `-` cases only?

ngrinkevich: yeah it's weird especially if you can't replicate it locally 😅 My only guess here I think if it was a normal transaction next (not .call like we have) then +1 would work reliably

ngrinkevich:
closing this, https://github.com/keep-network/keep-core/pull/1563 has a proper fix

Issue #1527, opened 2020-03-30
Service contract callbacks can be abused to call into other contracts


pdyraga:
`KeepRandomBeaconServiceImplV1` allows senders to specify an arbitrary method and contract that will receive a callback once the beacon generates a relay entry. Once an operator contract receives the relay entry, it calls `executeCallback`. Arbitrary callbacks can be used to force the service contract to execute many functions within the keep contract system. Currently, the `KeepRandomBeaconOperator` includes an `onlyServiceContract` modifier on some functions. The functions it protects cannot be targeted by the aforementioned service contract callbacks due to Solidity’s `CALLDATASIZE` checking. However, the presence of the modifier suggests that the service contract is expected to be a permissioned actor within some contracts.

We should:
1. Sick to a constant callback method signature, rather than allowing users to submit an arbitrary string.
2. Consider disallowing arbitrary callback destinations. Instead, rely on contracts making requests directly, and default the callback destination to `msg.sender`. Ensure the sender is not an EOA.

PR #1532, opened 2020-03-30
__beaconCallback(uint256)

pdyraga:
Closes: #1527 

See https://github.com/keep-network/keep-common/pull/22
See https://github.com/keep-network/keep-ecdsa/pull/339

Instead of allowing the customer to pass callback signature of their choice, we now expect that the callback contract delcares public `__beaconCallback(uint256)` function.

Previously, `KeepRandomBeaconServiceImplV1` allowed senders to specify an arbitrary method that would receive a callback once the beacon generated a relay entry. Arbitrary callbacks could be used to force the service contract to execute many functions within the keep contract system. For example, functions with `onlyServiceContract` modifier in the operator contract.

We still allow specifying an address of the callback contract. This could be beneficial in a situations where one contract pays for a random number for another contract.

### DIFF
--- a/solidity/contracts/KeepRandomBeaconServiceImplV1.sol
+++ b/solidity/contracts/KeepRandomBeaconServiceImplV1.sol
@@ -70,7 +70,6 @@ contract KeepRandomBeaconServiceImplV1 is DelayedWithdrawal, ReentrancyGuard, IR
 
     struct Callback {
         address callbackContract;
-        string callbackMethod;
         uint256 callbackFee;
         uint256 callbackGas;
         address payable surplusRecipient;
@@ -129,7 +128,7 @@ contract KeepRandomBeaconServiceImplV1 is DelayedWithdrawal, ReentrancyGuard, IR
         _pendingWithdrawal = 0;
         _previousEntry = _beaconSeed;
         _registry = registry;
-        _baseCallbackGas = 18789;
+        _baseCallbackGas = 10961;
 
         // Initialize DelayedWithdrawal owner.
         initialize(msg.sender);
@@ -221,15 +220,15 @@ contract KeepRandomBeaconServiceImplV1 is DelayedWithdrawal, ReentrancyGuard, IR
      * @return An uint256 representing uniquely generated entry Id.
      */
     function requestRelayEntry() public payable returns (uint256) {
-        return requestRelayEntry(address(0), "", 0);
+        return requestRelayEntry(address(0), 0);
     }
 
     /**
-     * @dev Creates a request to generate a new relay entry, which will include
-     * a random number (by signing the previous entry's random number).
-     * @param callbackContract Callback contract address. Callback is called once a new relay entry has been generated.
-     * @param callbackMethod Callback contract method signature. String representation of your method with a single
-     * uint256 input parameter i.e. "relayEntryCallback(uint256)".
+     * @dev Creates a request to generate a new relay entry (random number).
+     * @param callbackContract Callback contract address. Callback is called
+     * once a new relay entry has been generated. Callback contract must
+     * declare public `__beaconCallback(uint256)` function that is going to be
+     * executed with the result, once ready.
      * @param callbackGas Gas required for the callback.
      * The customer needs to ensure they provide a sufficient callback gas
      * to cover the gas fee of executing the callback. Any surplus is returned
@@ -239,7 +238,6 @@ contract KeepRandomBeaconServiceImplV1 is DelayedWithdrawal, ReentrancyGuard, IR
      */
     function requestRelayEntry(
         address callbackContract,
-        string memory callbackMethod,
         uint256 callbackGas
     ) public nonReentrant payable returns (uint256) {
         require(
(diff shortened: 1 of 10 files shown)


pdyraga:
Tested locally with a relay request without a callback and against `keep-ecdsa` as updated in https://github.com/keep-network/keep-ecdsa/pull/339; everything works as expected.

solidity/contracts/KeepRandomBeaconServiceImplV1.sol
```diff
@@ -129,7 +128,7 @@ contract KeepRandomBeaconServiceImplV1 is DelayedWithdrawal, ReentrancyGuard, IR
         _pendingWithdrawal = 0;
         _previousEntry = _beaconSeed;
         _registry = registry;
-        _baseCallbackGas = 18789;
+        _baseCallbackGas = 10961;
```
ngrinkevich: always nice to see discounts 😄☝️

solidity/contracts/examples/CallbackContract.sol
```diff
@@ -9,19 +9,12 @@ contract CallbackContract {
 
     uint256 internal _lastEntry;
 
-    /**
-     * @dev Example of a callback method. Method signature can be
-     * calculated as bytes4(keccak256("callback(uint256)")
-    */
-    function callback(uint256 requestResponse)
+    function __beaconCallback(uint256 entry)
         public
```
dimpar: do we want to have `public` in a new line?

pdyraga: This is how our Solidity linters work in `keep-ecdsa`, see [here](https://github.com/keep-network/keep-ecdsa/blob/master/solidity/contracts/BondedECDSAKeepFactory.sol). I don't think we ported this work to `keep-core` already but we'll be doing it soon. And then we'll have less work here.

dimpar: hah.. I see that it is kinda random in `keep-ecdsa`. I guess wip :) anyway, not biggy, we can leave it as is then

pdyraga: > hah.. I see that it is kinda random in keep-ecdsa

Yep. We should all have pre-commit hooks installed with the same linting configuration but we want to make it enforced on the build level.

solidity/scripts/request-relay-entry-with-callback.js
```diff
@@ -1,10 +1,9 @@
-const crypto = require("crypto")
 const KeepRandomBeaconServiceImplV1 = artifacts.require("KeepRandomBeaconServiceImplV1.sol");
 const KeepRandomBeaconService = artifacts.require('KeepRandomBeaconService.sol');
 
 // Example usage:
-// truffle exec ./scripts/request-relay-entry-with-callback.js yourContractAddress "callbackMethodName" callbackGas
-// truffle exec ./scripts/request-relay-entry-with-callback.js 0x9F57C01059057d821c6b4B04A4598322661C934F "callback(uint256)" 20000
+// truffle exec ./scripts/request-relay-entry-with-callback.js yourContractAddress callbackGas
+// truffle exec ./scripts/request-relay-entry-with-callback.js 0x9F57C01059057d821c6b4B04A4598322661C934F 20000
```
dimpar: I think this won't work, the number of params has changed. I've changed the script to this:
```
let entryFeeEstimate = await contractInstance.entryFeeEstimate(process.argv[5]);
    let tx = await contractInstance.methods['requestRelayEntry(address,uint256)'](
      process.argv[4],
      process.argv[5],
      {value: entryFeeEstimate}
    )
```
and then the request went trough

pdyraga: Good catch! Fixed in https://github.com/keep-network/keep-core/pull/1532/commits/767add64f9002e5e736ed13a9f5d2e957e908646

dimpar: we also need to change:
`await contractInstance.entryFeeEstimate(process.argv[6])` to
`await contractInstance.entryFeeEstimate(process.argv[5])`

pdyraga: 🤦‍♂ Done.

dimpar:
Are we sure that  `20000` of `callbackGas` is enough? I tried it a couple of times and then called `CallbackContract::lastEntry()` and the number didn't change. Only when I added gas `0` at the end (`200000`), then a new entry was set in the callback contract.

pdyraga:
> Are we sure that `20000` of `callbackGas` is enough? I tried it a couple of times and then called `CallbackContract::lastEntry()` and the number didn't change. Only when I added gas (`200000`), then a new entry was set in the callback contract. `200000` is probably too much, but just wanted to do a quick validation.

Depends on the callback. We don't hardcode `20000` anywhere in contracts. In `keep-ecdsa` we need [`41830`](https://github.com/keep-network/keep-ecdsa/blob/master/solidity/contracts/BondedECDSAKeepFactory.sol#L74) in the worst case. Each callback should be first checked with `estimateGas`.

dimpar:
maybe it would be good to add in `@dev` in `CallbackCotract.sol` that it will require `30k`?

anyway, to sum up:
tested with our `CallbackCotract.sol` deployed. It required `30k` of callback gas. Relay entry was generated and set. Looks good!

PR #1534, opened 2020-03-31
Validate declared group member index

### LINKED ISSUE #484: Validate DKG member index against group selection result

### LINKED ISSUE #834: Validate thresholdsig member index against group selection result

lukasz-zimnoch:
Closes #484
Closes #834

The group selection protocol outcome is an array of values uniquely identifying clients selected to the group. The index of a client in that array is an index used for DKG protocol. During DKG protocol execution we need to validate those indexes - check if all members we cooperate with, use the correct ones. If not, the message from the given sender should be rejected. Regarding relay entry signing, there is no need to validate member indexes explicitly as validation of signature shares introduced by #1274 covers it implicitly.


pdyraga:
Looks good. Can we add an integration test ensuring member who used incorrect sender id is removed from the final group?

lukasz-zimnoch:
> Looks good. Can we add an integration test ensuring member who used incorrect sender id is removed from the final group?

It's not that simple. Locally we are operating on a single local network provider so all members are coupled with the same public key and just changing the sender of a message won't do the trick.

pdyraga:
> > Looks good. Can we add an integration test ensuring member who used incorrect sender id is removed from the final group?
> 
> It's not that simple. Locally we are operating on a single local network provider so all members are coupled with the same public key and just changing the sender of a message won't do the trick.

We can probably set a sender ID to 65, it should do the trick.

lukasz-zimnoch:
> > > Looks good. Can we add an integration test ensuring member who used incorrect sender id is removed from the final group?
> > 
> > 
> > It's not that simple. Locally we are operating on a single local network provider so all members are coupled with the same public key and just changing the sender of a message won't do the trick.
> 
> We can probably set a sender ID to 65, it should do the trick.

Ok, sounds reasonable. I pushed a new test scenario.

pdyraga:
Tested with 9 clients locally, executed DKG 5 times, I had 64 supporting signatures on the result every single time.

PR #1535, opened 2020-03-31
Add delegated authorization functionality

eth-r:
Refs: #1490 

In #1519 individual keeps can't ask the factory to place locks on operators on their behalf, because only one lock per operator contract is permitted. To get around this, keep factories need a way to delegate their authority to keeps. The method in this PR lets keeps claim delegated authorization, which is verified by calling the factory. After authorization has been delegated, the keep's authorization status matches that of the factory, covering approval and panic button status in the registry and individual operators' authorization.

### DIFF
--- a/solidity/test/token_stake/TestDelegatedAuthority.js
+++ b/solidity/test/token_stake/TestDelegatedAuthority.js
@@ -86,6 +86,18 @@ contract("TokenStaking/DelegatedAuthority", async (accounts) => {
     return stakingContract.isAuthorizedForOperator(operator, operatorContract);
   }
 
+  async function authorize(operatorContract) {
+    stakingContract.authorizeOperatorContract(
+      operator,
+      operatorContract.address,
+      {from: authorizer}
+    );
+  }
+
+  async function disable(operatorContract) {
+    registry.disableOperatorContract(operatorContract.address);
+  }
+
   describe("claimDelegatedAuthority", async () => {
     it("lets contracts claim delegated authority", async () => {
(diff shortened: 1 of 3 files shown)


solidity/test/token_stake/TestDelegatedAuthority.js
```diff
@@ -0,0 +1,265 @@
...
+
+const BN = web3.utils.BN
+const chai = require('chai')
+chai.use(require('bn-chai')(BN))
+const expect = chai.expect
+
+// Depending on test network increaseTimeTo can be inconsistent and add
+// extra time. As a workaround we subtract timeRoundMargin in all cases
+// that test times before initialization/undelegation periods end.
+const timeRoundMargin = duration.minutes(1)
```
pdyraga, line 14: Not used in this file.

solidity/test/token_stake/TestDelegatedAuthority.js
```diff
@@ -0,0 +1,265 @@
+import increaseTime, {duration, increaseTimeTo} from '../helpers/increaseTime';
+import latestTime from '../helpers/latestTime';
```
pdyraga, line 2:
```
import increaseTimeTo from '../helpers/increaseTime';
```
should be enough here

pdyraga:
There are two outstanding parts:
- Claiming the authority by a keep in `keep-ecdsa` and implementing locks
- We can now clean up `keep-ecdsa` so that a keep does not have to ask the factory to do the slashing.

@eth-r maybe it's worth involving @ngrinkevich in some of this work?

Issue #1506, opened 2020-03-23
Hardware wallet integration


pdyraga:
Provide a way to use hardware wallets with KEEP token dashboard.

liz-shinn:
Dropping slides from Figma into here! I researched some dApps (mainly Oasis) to see how they were handling this. I've currently included MetaMask, Ledger (Live and Legacy), Coinbase, and Trezor. 

MetaMask: Should link to the download page in an external window if it's not installed and have a web3 error
Coinbase: Will show a QR code, link to download Coinbase Wallet 
Ledger: Shows Ledger Live and Ledger Legacy and a 'connect' button that I assume looks for something plugged into the USB drive?
Trezor: Links out to an external wizard

liz-shinn:
The far left modal window shows all the options supported, and the other two are links from when you click either Ledger or Coinbase, respectively.

![Image 2020-03-30 at 4 18 27 PM](https://user-images.githubusercontent.com/57226633/77957843-1f159d00-72a2-11ea-8b6c-6ab7dfdc2500.png)

liz-shinn:
Full screen for context (you can see there's an error message that's saying "Please connect a wallet" in the top lefthand corner).

![Image 2020-03-30 at 4 21 38 PM](https://user-images.githubusercontent.com/57226633/77958120-91867d00-72a2-11ea-98ce-0b5c4d5a5977.png)
![Image 2020-03-30 at 4 22 02 PM](https://user-images.githubusercontent.com/57226633/77958150-9ba87b80-72a2-11ea-9a09-a09c032545de.png)

liz-shinn:
I think that the modal should show initially, though, and not require the user to click on the "Please connect to a wallet" link. 

I'm thinking this link would be there for when the user closes out of this modal but still needs to link a wallet.

liz-shinn:
The copy has changed in Figma for most of these, btw. Also, addition of images to the hardware wallet modal flows:

<img width="409" alt="Image 2020-04-07 at 3 22 07 PM" src="https://user-images.githubusercontent.com/57226633/78710668-bb175800-78e3-11ea-90eb-42f1433cf236.png">
<img width="424" alt="Image 2020-04-07 at 3 22 01 PM" src="https://user-images.githubusercontent.com/57226633/78710674-beaadf00-78e3-11ea-9571-0abbec816d8a.png">

PR #1538, opened 2020-04-01
 Hardware wallet integration- dapp

r-czajkowski:
Refs: #1506

This PR adds support for the Ledger and Trezor hardware wallet based on 0x package. 0x and ledgerjs API's not fully supporting chainId's greater than 255 we had to implement a walkaround for signing transactions.

<details>
Zrzut ekranu 2020-04-15 o 14 35 08

obraz

trezor-login

  
</details>


r-czajkowski:
trezor-first-tx

Shadowfiend:
😮

solidity/dashboard/src/connectors/ledger.js
```diff
@@ -0,0 +1,77 @@
...
+      tx.raw[8] = Buffer.from([])
+      const result = await this._ledgerClientIfExists.signTransaction(
+        fullDerivationPath,
+        tx.serialize().toString('hex')
+      )
+
+      let v = result.v
+      const rv = parseInt(v, 16)
+      let cv = this.chainId * 2 + 35
+      if (rv !== cv && (rv & cv) !== rv) {
```
liamzebedee, line 57: `(rv & cv) !== rv` - could you explain this to me?

I had something different, checking if the ledger `v` was even/odd and using that to add the sign bit.

r-czajkowski: I have to refactor this snippet of code and add a comment, it's quite confusing. I'm going to apply [your approach](https://github.com/keep-network/tbtc-dapp/pull/205/files#diff-86a78d68537ecad0a90a50fc3100341dR60-R75).

So as you mentioned [here](https://github.com/keep-network/tbtc-dapp/pull/205/files#diff-86a78d68537ecad0a90a50fc3100341dR56) the transport layer only returns the lower 2 bytes. For `chainId = 1101` a transport layer returns => `v = be` because `2238 == 08be`:
* `rv = parseInt(v, 16) = 190`
*  `rv & cv => 190 & 2237 => 188` and `190 & 2238 => 190`
*  `(rv & cv) !== rv)` with this operation I check if have to add `1` according to the eip155 algorithm.

solidity/dashboard/src/connectors/abstract.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    setProvider = (provider) => {
+      this.provider = provider
+    }
+
+    getProvider = () => {
+      return this.provider
+    }
+}
```
lukasz-zimnoch, line 34: Can we rename this file to `abstract-connector.js`?

solidity/dashboard/src/connectors/ledger.js
```diff
@@ -0,0 +1,81 @@
...
+
+export class LedgerProvider extends AbstractHardwareWalletConnector {
+  constructor() {
+    super(new CustomLedgerSubprovider(getChainId()))
+  }
+}
+
+const ledgerEthereumClientFactoryAsync = async () => {
+  const ledgerConnection = await TransportU2F.create()
+  ledgerConnection.setExchangeTimeout(100000)
```
lukasz-zimnoch, line 17: Can we extract this timeout to a constant?

solidity/dashboard/src/connectors/utils.js
```diff
@@ -0,0 +1,27 @@
...
+// EIP-155 https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md
+// v = CHAIN_ID * 2 + 35 => CHAIN_ID = (v - 35) / 2
+export const getChainIdFromV = (vInHex) => {
+  const vIntValue = parseInt(vInHex, 16)
+  const chainId = Math.floor((vIntValue - 35) / 2)
+  return chainId < 0 ? 0 : chainId
+}
+
+export const getChainId = () => {
+  return process.env.CHAIN_ID || 1337
```
lukasz-zimnoch, line 22: Worth to drop a comment on why we return `1337` as a default value.

solidity/dashboard/src/connectors/utils.js
```diff
@@ -0,0 +1,27 @@
+import Common from 'ethereumjs-common'
+import { Transaction as EthereumTx } from 'ethereumjs-tx'
+
+export const getEthereumTxObj = (txData, chainId) => {
+  const customCommon = Common.forCustomChain('mainnet', {
+    name: 'keep-dev',
```
lukasz-zimnoch, line 6: What's the purpose of this param?

pdyraga:
How are we doing here? I'd love to ship it as soon as possible as this will block staking next week.

pdyraga:
How are we doing here? This is blocking staking so let's treat this one as a top priority next to managed grant changes.

r-czajkowski:
PR is ready for the next round. Also, I think several folks should test it with Trezor/Ledger.

mhluongo:
I don't have a Trezor on hand (it's been years) but happy to test with a Ledger. Is there a URL pre-deploy or should I run it locally?

pdyraga:
It has to be tested locally, I believe the steps from #1566 are up-to-date.

dimpar:
1. Delegation status won't auto-update after token delegation.

2. In Operations, "Undelegate" button is enabled even though we have 0 Keeps

r-czajkowski:
>     1. Delegation status won't auto-update after token delegation.
> 
> 
> Screenshot 2020-05-05 at 10 45 34
> 
>     2. In Operations, "Undelegate" button is enabled even though we have 0 Keeps

1. I think we should open a new issue for that.

2. We already have an opened issue for this  #1719

dimpar:
1) [issue](https://github.com/keep-network/keep-core/issues/1750)
2) ok!

PR #1539, opened 2020-04-01
You Can Quote Me On That: Add quotes to quickstart volume and env parameters

Shadowfiend:
In some cases previously-set variables can contain spaces, so if these
parameters aren't wrapped in quotes you can get unexpected and
hard-to-track-down docker errors.


leopoldjoy:
Encountered this problem ourselves, thanks @Shadowfiend! 👍

PR #1541, opened 2020-04-02
Please actually README

mhluongo:
Rework the README to better introduce the project and docs to a newcomer.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -51,10 +51,8 @@ computations.
 
 === Client Developers
 
-Client developers will be most interested in the link:go/[reference Keep Go
-client], as well as the link:docs/rfc/[RFCs],
-link:/docs/development-process.adoc[detailed development process], and the
-link:/docs/random-beacon/[random beacon yellowpaper].
+Client developers will be most interested in the link:./main.go[reference Keep
+Go client], as well as the link:docs/rfc/[RFCs] and repo directory structure 👇
 
 ==== Directory structure
 
@@ -80,9 +78,9 @@ keep-core/
     relay/
       relay.go, *.go
 ```
-<1> While Keep only uses Solidity at the moment, the directory structure allows
-    for other contract languages.
-<2> Keep subcommands are implemented here, though they should be minimal and
+<1> While the Keep network only uses Solidity at the moment, the directory
+    structure allows for other contract languages.
+<2> Keep client subcommands are implemented here, though they should be minimal and
     deal solely with user interaction. The meat of the commands should exist in
     a package fit for the appropriate purpose.
 <3> All additional packages live in `pkg/`.
--- a/README.adoc
+++ b/README.adoc
@@ -27,7 +27,7 @@ applications, and new tools for custodial applications — from Solidity. This
 capability is used heavily by https://tbtc.network/[tBTC].
 
 To learn more about ECDSA keeps, check out
-[keep-ecdsa](https://github.com/keep-network/keep-ecdsa)
+https://github.com/keep-network/keep-ecdsa[keep-ecdsa].
 
 == Getting Started
 
--- a/README.adoc
+++ b/README.adoc
@@ -42,12 +42,12 @@ https://github.com/keep-network/keep-core/issues[is appreciated!]
 dApp developers will be most interested in the smart contracts exposing Keep's
 on-chain facilities.
 
-The contracts can be found in the link:solidity/[`solidity/`] directory. They
-can be used to request miner-resistant random numbers, as well as handle
-creating and managing keeps, generating new secrets, and requesting signatures.
-
-It runs the Keep network’s random beacon and participates in related
-computations.
+The core contracts can be found in the link:solidity/[`solidity/`] directory.
+They can be used to request
+link:solidity/contracts/IRandomBeacon.sol[miner-resistant random numbers], as
+well as creating and managing keeps. To generate new ECDSA key material and
+request signatures, the contracts can be found in
+link:https://github.com/keep-network/keep-ecdsa/blob/master/solidity/contracts/api/IBondedECDSAKeep.sol[`keep-ecdsa`].
 
 === Client Developers
 
--- a/README.adoc
+++ b/README.adoc
@@ -55,7 +55,8 @@ link:https://github.com/keep-network/keep-ecdsa/blob/master/solidity/contracts/a
 === Client Developers
 
 Client developers will be most interested in the link:./main.go[reference Keep
-Go client], as well as the link:docs/rfc/[RFCs] and repo directory structure 👇
+Go client] and link:./CONTRIBUTING.adoc[CONTRIBUTORS file], as well as the
+link:docs/rfc/[RFCs] and repo directory structure 👇
 
 ==== Directory structure
 
--- a/README.adoc
+++ b/README.adoc
@@ -14,8 +14,11 @@ containers for private storage and computation that can be opened, closed, and
 managed by smart contracts autonomously.
 
 Keeps are maintained by stakers, actors who run nodes and have skin in the
-game. When a new keep is opened, the requisite number of stakers are chosen
-via a BLS-based random beacon to maintain the keep.
+game, and collect fees for operating the network. When a new keep is opened,
+the requisite number of stakers are chosen via a
+link:https://blog.keep.network/whats-in-a-beacon-12c34b0bc078[BLS-based random beacon]
+to maintain the keep, using a process called
+link:https://en.wikipedia.org/wiki/Sortition[sortition].
 
 The first type of keep launching with the network is the `BondedECDSAKeep`,
 allowing smart contracts to generate private keys and sign messages without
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -1,15 +1,62 @@
...
+
+It runs the Keep network’s random beacon and participates in related
+computations.
+
+=== Client Developers
+
+Client developers will be most interested in the link:go/[reference Keep Go
+client], as well as the link:docs/rfc/[RFCs],
+link:/docs/development-process.adoc[detailed development process], and the
+link:/docs/random-beacon/[random beacon yellowpaper].
```
pdyraga, line 54: We are introducing a concept of a keep in the first paragraph but `keep-core` client does not implement it. We should somehow make it clear and refer to `keep-ecdsa`.

README.adoc
```diff
@@ -1,15 +1,62 @@
...
+can be used to request miner-resistant random numbers, as well as handle
+creating and managing keeps, generating new secrets, and requesting signatures.
+
+It runs the Keep network’s random beacon and participates in related
+computations.
+
+=== Client Developers
+
+Client developers will be most interested in the link:go/[reference Keep Go
+client], as well as the link:docs/rfc/[RFCs],
```
Shadowfiend, line 52: Also `go/` is not now nor has it ever been a thing. The directory structure is explained below :wink:

Shadowfiend: > nor has it ever

Turns out [this isn't true](https://github.com/keep-network/keep-core/tree/9635be6899ae2e76bdeb668ea591c73df3288e98), for those keeping track :grin: Had a gnawing sense of overcommitment there.

Shadowfiend: I think the move here is pointing folks to the directory structure at the end of the doc + the RFCs.

README.adoc
mhluongo: Good call, I'll link it.

README.adoc
mhluongo: Heh the `go/` came from a very very old version of master (my fork) then rebased. Thanks, will fix

README.adoc
```diff
@@ -1,15 +1,62 @@
...
+dApp developers will be most interested in the smart contracts exposing Keep's
+on-chain facilities.
+
+The contracts can be found in the link:solidity/[`solidity/`] directory. They
+can be used to request miner-resistant random numbers, as well as handle
+creating and managing keeps, generating new secrets, and requesting signatures.
+
+It runs the Keep network’s random beacon and participates in related
+computations.
+
```
pdyraga, line 48: I think it makes sense to point them to:
- https://github.com/keep-network/keep-core/blob/master/solidity/contracts/IRandomBeacon.sol
- https://github.com/keep-network/keep-ecdsa/tree/master/solidity/contracts/api

These are facades for beacon and ECDSA keeps, I think it's a good starting point for a journey through the Solidity codebase.

README.adoc
```diff
@@ -1,15 +1,62 @@
...
+=== Client Developers
+
+Client developers will be most interested in the link:go/[reference Keep Go
+client], as well as the link:docs/rfc/[RFCs],
+link:/docs/development-process.adoc[detailed development process], and the
```
pdyraga, line 53: I am not sure if it makes sense to point to this document here. It's our internal process, it doesn't make sense for external contributors.

README.adoc
Shadowfiend: Omg so old 😂

README.adoc
```diff
@@ -1,15 +1,62 @@
...
 
-== Directory structure
+The network offers application developers **keeps**, small off-chain data
+containers for private storage and computation that can be opened, closed, and
+managed by smart contracts autonomously.
 
-The directory structure used in the Keep repository is very similar to that used
-in other Go projects:
+Keeps are maintained by stakers, actors who run nodes and have skin in the
+game. When a new keep is opened, the requisite number of stakers are chosen
```
liamzebedee, line 17: Keep for Dummies Q's:

 - Why do they stake?
 - What's a random beacon?
 - How are they chosen using a beacon?
 - Why are they chosen via a beacon?

README.adoc
```diff
@@ -1,15 +1,62 @@
...
+via a BLS-based random beacon to maintain the keep.
+
+The first type of keep launching with the network is the `BondedECDSAKeep`,
```
liamzebedee, line 20: "first type of Keep" I like the phrasing here, sets up the mental model for the AESKeep, etc.

Link out to keep-ecdsa here? Or perhaps below in dApp Developers.

README.adoc
```diff
@@ -1,15 +1,62 @@
...
+https://github.com/keep-network/keep-core/issues[is appreciated!]
+
+=== dApp Developers
+
+dApp developers will be most interested in the smart contracts exposing Keep's
+on-chain facilities.
+
+The contracts can be found in the link:solidity/[`solidity/`] directory. They
+can be used to request miner-resistant random numbers, as well as handle
+creating and managing keeps, generating new secrets, and requesting signatures.
```
liamzebedee, line 44: "and requesting signatures" - I'm confused if you're referring to Keep ECDSA or the Beacon in this context.

README.adoc
liamzebedee: 👍 I like the idea of linking the two most prominent API's that devs will be building on here - IRandomBeacon and IBondedECDSAKeep.

README.adoc
mhluongo: Yeah it was sort of depressing to realize :sweat_smile:

README.adoc
mhluongo: ECDSA - the "signatures" part of the beacon is an implementation detail

README.adoc
mhluongo: Porque no los dos?

mhluongo:
I think I've responded to all feedback- this is ready for another look.

README.adoc
```diff
@@ -14,15 +14,21 @@ containers for private storage and computation that can be opened, closed, and
...
+link:https://en.wikipedia.org/wiki/Sortition[sortition].
 
 The first type of keep launching with the network is the `BondedECDSAKeep`,
 allowing smart contracts to generate private keys and sign messages without
 endangering key material. ECDSA keeps mean decentralized signing, cross-chain
 applications, and new tools for custodial applications — from Solidity. This
 capability is used heavily by https://tbtc.network/[tBTC].
 
+To learn more about ECDSA keeps, check out
+[keep-ecdsa](https://github.com/keep-network/keep-ecdsa)
```
Shadowfiend, line 30: Roses are red
Violets are blue
Periods are nice
And so is using them to end sentences 😘

Shadowfiend: (Also this is asciidoc, not markdown. `https://github.com/keep-network/keep-ecdsa[keep-ecdsa]`)

README.adoc
```diff
@@ -39,19 +45,17 @@ https://github.com/keep-network/keep-core/issues[is appreciated!]
...
-creating and managing keeps, generating new secrets, and requesting signatures.
-
-It runs the Keep network’s random beacon and participates in related
-computations.
+The core contracts can be found in the link:solidity/[`solidity/`] directory.
+They can be used to request
+link:solidity/contracts/IRandomBeacon.sol[miner-resistant random numbers], as
+well as creating and managing keeps. To generate new ECDSA key material and
+request signatures, the contracts can be found in
+link:https://github.com/keep-network/keep-ecdsa/blob/master/solidity/contracts/api/IBondedECDSAKeep.sol[`keep-ecdsa`].
```
Shadowfiend, line 53: Hmmm… I feel like you're either linking to the contracts, so solidity dir in keep-ecdsa, or to the interface for Bonded ECDSA keeps, in which case it's worth saying as much.

Really I'm not convinced leaving the interface names out yet linking directly to the interfaces is a great idea. Not a blocking note, just something to think about.

mhluongo: Including the interface name in an introductory README seems like overkill. In my mind this doc is about convincing people to dig deeper + acting as an index of our work, rather than familiarizing people with specific code details immediately.

That said, if we have a killer code snippet for dApp devs that'd be a nice alternative to either

mhluongo: Will give it some more thought :thinking:

mhluongo: I guess what I really want is a link to a

* "how to use the random beacon" doc, or alternatively "random beacon Solidity API docs"
* "how to open a keep and do something with it" doc, or alternatively "all ECDSA keep Solidity API docs"

Shadowfiend: ^^ agreed. For the next pass :)

README.adoc
```diff
@@ -55,7 +55,8 @@ link:https://github.com/keep-network/keep-ecdsa/blob/master/solidity/contracts/a
 === Client Developers
 
 Client developers will be most interested in the link:./main.go[reference Keep
-Go client], as well as the link:docs/rfc/[RFCs] and repo directory structure 👇
+Go client] and link:./CONTRIBUTING.adoc[CONTRIBUTORS file], as well as the
```
Shadowfiend, line 58: I mean I'm going to merge this but………………

CONTRIBUTING.adoc
CONTRIBUTORS file

😝😁😆

mhluongo: I think it looks *great* <unclear lack of punctuation>

PR #1543, opened 2020-04-02
keep-dev Auto-deploy beacon clients + keep-dev/test log config update

sthompson22:
Here we have Keel monitor image publishing for random beacon clients.  If a fresh image is published, Keel will pick it up and deploy for each beacon client running in keep-dev.

We're still running `genesis` manually, so some human will still be required before the beacon is usable post deploy.

We also need to watch InitContainer behavior since all 5 clients use the same `purse` for funding, may run into nonce issues.  As long as things naturally clear up in a matter of minutes, I'm ok with it.

------

We're currently using Stackdriver for centralized logging, parsing logs in their current state is an exercise in frustration.  Setting `nocolor` here does away with some of the extra characters we were dealing with for cleaner log lines.


pdyraga:
I see we have two approvals and I am the last one missing... -_- 

Just to double-check before I hit the button: is there anything I missed why we haven't merged it yet? It all looks solid.

PR #1546, opened 2020-04-03
Grant staking policy- dapp update

### LINKED ISSUE #1478: Allow to delegate revocable grant
It is needed that the revocable grant can be staked. The network may need to take a hit when the owner of the revoked grant starts misbehaving. Doing the token distribution right should protect the network.

r-czajkowski:
Closes #1478

There are two staking policies deployed as contracts:
* `GuaranteedMinimumStakingPolicy`
* `PermissiveStakingPolicy`

The first one is applied to all revocable grants and the second one to all non-revocable grants. Grant policy address is passed in `_extraData` in `receiveApproval` of `TokenGrant`.


pdyraga:
image


Something is off:
1. The validation message is probably missing `X` value in `less than X or`
2. `KEEP Available` is not rendering the amount

Also, I think we should replace `KEEP Available` with `KEEP available to stake` as it's now possible not entire KEEP grant can be staked for a revocable grant (cc @liz-shinn)

Issue #1479, opened 2020-03-19
Incorrect gas estimation for the first relay entry submitter


pdyraga:
It often happens that member 1 trying to submit relay entry fails with out of gas. Member 2 or 3 is able to submit transaction successfully with just a little difference in TX fee.

Example:

member 1: https://ropsten.etherscan.io/tx/0x125b72cdc4b49647a249604cd6691019459b805423e3719adfaef70202a867cd
member 2: https://ropsten.etherscan.io/tx/0x15b62650ae9c1ddcc175d065fead4980f82eaccfdd617dce6299e79ed0e2c4c2 (out of gas problem on internal TX is a separate issue)

member 1: https://ropsten.etherscan.io/tx/0xaff348bde2f6dcce2feb6bc06674bdc75fbdf3076e27ecf1b4f905a8ce0db1d6

member 2: https://ropsten.etherscan.io/tx/0xfd83458c9badc6631de2f68e8330b86d623bb2a15301a0060012381eed24b828 (out of gas problem on internal TX is a separate issue)

From contract https://ropsten.etherscan.io/txs?a=0x1f5fecabf1cf3732d889e0bbccdcf5d5120391e4

PR #1550, opened 2020-04-03
Incorrect gas estimation for the first relay entry submitter

dimpar:
Closes https://github.com/keep-network/keep-core/issues/1479

In this PR we need to increase gas limit for the relay entry submission.  The reason behind it, is that sometimes we have cases in `KeepRandomBeaconOperator::getDelayFactor()` when we assign `block.number` to `entryReceivedBlock`. The original estimation does not take it into account and we run out of gas when the first member tries to submit a relay entry. The difference is about 15% between the original estimate and what is actually needed.


pkg/chain/ethereum/ethereum.go
```diff
@@ -237,7 +244,42 @@ func (ec *ethereumChain) SubmitRelayEntry(
...
+	msg := ethereum.CallMsg{
+		From: from,
+		To:   toAddress,
+		Data: packed,
+	}
+
+	gasEstimate, err := ec.client.EstimateGas(context.TODO(), msg)
+	if err != nil {
+		logger.Errorf("failed to estimate gas for relay entry [%v]", err)
+	}
```
pdyraga, line 274: All of it should perhaps be a generated code inside `contract.KeepRandomBeaconOperator`. We can generate `*EstimateGas` function for each transactional function we generate in `keep-commons`. Then, we'd have `RelayEntryGasEstimate`.

@Shadowfiend What do you think about it?

pkg/chain/ethereum/ethereum.go
```diff
@@ -237,7 +244,42 @@ func (ec *ethereumChain) SubmitRelayEntry(
...
+	gasEstimateWithMargin := float64(gasEstimate) * float64(1.2) // 20% more than original
```
pdyraga, line 275: `float64(1.2)` could be a `gasMultiplier` in `ethutil.TransactionOptions{`. If `gasMultiplier` is supplied, the contract executes `*EstimateGas` and calculates the final value. Then, we can reduce this code to just:
```
_, err = ec.keepRandomBeaconOperatorContract.RelayEntry(
  entry,
  ethutil.TransactionOptions{
    GasMultiplier: float64(1.2),
  },
)
```

What do you think?

pdyraga:
@dimpar From the PR description I understand that this line is problematic:
```
uint256 entryReceivedBlock = block.number <= submissionStartBlock ? submissionStartBlock:block.number;
```

Does it mean `block.number` requires ~1700 more gas? Does it add any additional opcodes? Looking at the Ropsten contract it is about ~1700 gas units less than required (e.g. 238,068 vs 239,760)

dimpar:
@pdyraga 

I've did some more testing on `Geth` to see what the `gas.used` looks like:
1. In our current implementation:

|  #      | Estimated and Gas used for a failed transaction  | Estimated gas for a successful transaction | Gas Used for a successful transaction | Gas diff |
| - |:-----:| -----:| ------:| ------:|
| 1     | 251781 | 283387 | 268387 | 16606 |
| 2    | 236781 | 253387 | 238387 | 1606   |

I'm not sure why we have the diff of `15k` between `#1` and `#2` relay request. Maybe it caches smth after the first attempt or optimizes it somehow. From what I found these are the opcodes that are used for `block.number` (but I didn't dive very deep) and `SSTORE` is the highest. [gas cost from 04.2017](https://docs.google.com/spreadsheets/d/1n6mRqkBz3iWcOlRem_mO09GtSKEKrAsfO7Frgx18pNU/edit#gid=0). Or [Appendix G](https://ethereum.github.io/yellowpaper/paper.pdf)
```
NUMBER
PUSH1
SSTORE
```

2. Tested when: `uint256 entryReceivedBlock =  submissionStartBlock;`
average received gas estimate: 236755
average gasUsed: 221755

3. Tested when: `uint256 entryReceivedBlock = block.number;`
- First relay request was right after genesis used the most: 
received gas estimate: 283339
gasUsed: 268339
- Next relay request on average:
received gas estimate: 253339
gasUsed: 238339

Again, based on 3) we probably reuse the existing data that lowers consecutive calls for relay entry submission.

pkg/chain/ethereum/ethereum.go
dimpar: agree, but we'd also need to pass `GasLimit` along with `GasMultiplier` either case:

```
_, err = ec.keepRandomBeaconOperatorContract.RelayEntry(
  entry,
  ethutil.TransactionOptions{
    GasLimit: 1234,
    GasMultiplier: float64(1.2),
  },
)
```

pdyraga: Why the gas limit? For relay entry we can't set a gas limit because of a callback.

dimpar: I mean the gas estimated earlier that would be assigned to `GasLimit` just like in this PR.

From what I see if we don't provide `GasLimit` in TransactionOptions then we estimate gas downstream in `go-ethereum/accounts/abi/bind/base.go::L222` before sending transaction. Unless I miss smth, then passing just `GasMultiplier` without `GasLimit` won't work.

pdyraga: If `GasMultiplier` is provided, the generated contract code need to do the gas estimation part, that's my understanding. We can do easier option first, though. Expose `*GasEstimate` functions and instead of `GasMultipler`, call `RelayEntryGasEstimate` here, multiply manually, and pass the resulting value as `GasLimit`.

dimpar: > If GasMultiplier is provided, the generated contract code need to do the gas estimation part, that's my understanding.

Ah.. I see, then in ex. `options.go::Apply()` we would need to call the estimation function and set `GasLimit` multiplied by ex. `1.2`? 

> We can do easier option first, though.
K, I'll go with that first.

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/ipfs/go-datastore v0.1.1
 	github.com/ipfs/go-log v0.0.1
 	github.com/keep-network/go-libp2p-bootstrap v0.0.0-20190611114437-e92bd71e8199
-	github.com/keep-network/keep-common v0.1.1-0.20200330215727-8ca95ff9de56
+	github.com/keep-network/keep-common v0.1.1-0.20200407084635-41258118ebd7
```
dimpar, line 17: NOTE! this is not a master branch dependency. Need to merge [keep common PR](https://github.com/keep-network/keep-common/pull/24) first and then update this version.

pkg/chain/ethereum/ethereum.go
dimpar: @pdyraga ready for another look

pkg/chain/ethereum/ethereum.go
dimpar: [PR](https://github.com/keep-network/keep-common/pulls) in `keep-commons`

go.mod
pdyraga: Done in https://github.com/keep-network/keep-core/pull/1550/commits/447538e8a9de2affae013251f513e7d9a552b11f

dimpar:
The keep-common version looks good! Let's merge

PR #1557, opened 2020-04-06
Improve off-chain tickets value consistency

lukasz-zimnoch:
Refs #1431

During the review of #1510, we agreed that the ticket value representation on the client side is not consistent and should be improved. Here we introduce some improvements which should make the code cleaner and more consistent with the on-chain part.

### DIFF
--- a/pkg/beacon/relay/chain/group_ticket.go
+++ b/pkg/beacon/relay/chain/group_ticket.go
@@ -10,11 +10,6 @@ type Ticket struct {
 	Proof *TicketProof
 }
 
-// IntValue returns ticket value as a big integer.
-func (t *Ticket) IntValue() *big.Int {
-	return new(big.Int).SetBytes(t.Value[:])
-}
-
 // TicketProof represents group selection ticket proof as seen on-chain.
 type TicketProof struct {
 	StakerValue        *big.Int
--- a/pkg/beacon/relay/chain/group_ticket.go
+++ b/pkg/beacon/relay/chain/group_ticket.go
@@ -2,19 +2,17 @@ package chain
 
 import (
 	"math/big"
-
-	"github.com/ethereum/go-ethereum/common"
 )
 
 // Ticket represents group selection ticket as seen on-chain.
 type Ticket struct {
-	Value *big.Int // W_k
+	Value [8]byte // W_k
 	Proof *TicketProof
 }
 
-// ValueBytes returns ticket value as a byte slice.
-func (t *Ticket) ValueBytes() []byte {
-	return common.LeftPadBytes(t.Value.Bytes(), 8)[:8]
+// IntValue returns ticket value as a big integer.
+func (t *Ticket) IntValue() *big.Int {
+	return new(big.Int).SetBytes(t.Value[:])
 }
 
 // TicketProof represents group selection ticket proof as seen on-chain.
--- a/pkg/beacon/relay/chain/group_ticket.go
+++ b/pkg/beacon/relay/chain/group_ticket.go
@@ -1,11 +1,20 @@
 package chain
 
-import "math/big"
+import (
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+)
 
 // Ticket represents group selection ticket as seen on-chain.
 type Ticket struct {
 	Value *big.Int // W_k
 	Proof *TicketProof
+}
+
+// ValueBytes returns ticket value as a byte slice.
+func (t *Ticket) ValueBytes() []byte {
+	return common.LeftPadBytes(t.Value.Bytes(), 8)[:8]
 }
 
 // TicketProof represents group selection ticket proof as seen on-chain.
--- a/pkg/beacon/relay/groupselection/groupselection_test.go
+++ b/pkg/beacon/relay/groupselection/groupselection_test.go
@@ -2,6 +2,7 @@ package groupselection
 
 import (
 	"encoding/binary"
+	"math/big"
 	"reflect"
 	"sort"
 	"testing"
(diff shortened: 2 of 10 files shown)


pkg/beacon/relay/chain/group_ticket.go
```diff
@@ -1,13 +1,22 @@
...
 // Ticket represents group selection ticket as seen on-chain.
 type Ticket struct {
 	Value *big.Int // W_k
 	Proof *TicketProof
 }
 
+// ValueBytes returns ticket value as a byte slice.
+func (t *Ticket) ValueBytes() []byte {
+	return common.LeftPadBytes(t.Value.Bytes(), 8)[:8]
+}
```
pdyraga, line 18: Instead of this function, why don't we just replace `Value *big.Int` with `Value [8]byte`? 

The conversion to and from `*big.Int` scares me as we may run into problems with byte ordering.

lukasz-zimnoch: Yeah, you are right. Good point!

pdyraga: Can we remove this function? If `local.go` uses this conversion, let's have it in `local.go` where there is no risk with endianness problems.

pkg/beacon/relay/groupselection/ticket.go
```diff
@@ -47,42 +47,43 @@ func newTicket(
...
 
-	copy(keccak256Hash[:], crypto.Keccak256(combinedValue[:]))
+	copy(ticketValue[:], crypto.Keccak256(combinedValue[:])[:8])
 
-	return keccak256Hash, nil
+	return ticketValue, nil
 }
 
-// Returns 8 starting bytes of ticket value as a big integer.
+// Returns ticket value as a big integer.
```
pdyraga, line 84: Let's be very specific here.
```
// Returns ticket value as a big integer.
// Ticket value bytes are interpreted as a big-endian unsigned integer.
```

pkg/beacon/relay/groupselection/ticket.go
```diff
@@ -14,8 +14,8 @@ import (
 // used to determine whether a given virtual staker is eligible for the group P
 // (the lowest N tickets will be chosen) and a proof of the validity of the value
```
pdyraga, line 15: Let's add this information:
```
// Ticket value bytes should be interpreted as a big-endian unsigned integer.
```

pkg/beacon/relay/chain/group_ticket.go
pdyraga: This function is used only in `local.go`. I think it makes sense to do the conversion there and remove it from here. Especially that bytes->uint256 conversion is done on-chain. This will work as an additional safety guard for byte endianness. Let's be specific about it in `local.go`.

```
sort.SliceStable(c.tickets, func(i, j int) bool {
  // ticket value bytes are interpreted as a big-endian unsigned integers
  iValue := new(big.Int).SetBytes(c.tickets[i].Value[:])
  jValue := new(big.Int).SetBytes(c.tickets[j].Value[:])

  return iValue.Cmp(jValue) == -1
})
```

and later:
```
_ = promise.Fulfill(&event.GroupTicketSubmission{
  TicketValue: new(big.Int).SetBytes(ticket.Value[:]),
  BlockNumber: c.simulatedHeight,
})
```

And the same in `groupselection_test.go`:
```
	sort.SliceStable(stg.submittedTickets, func(i, j int) bool {
		// ticket value bytes are interpreted as a big-endian unsigned integers
		iValue := new(big.Int).SetBytes(stg.submittedTickets[i].Value[:])
		jValue := new(big.Int).SetBytes(stg.submittedTickets[j].Value[:])

		return iValue.Cmp(jValue) == -1
	})

	// (...)

	_ = promise.Fulfill(&event.GroupTicketSubmission{
		TicketValue: new(big.Int).SetBytes(ticket.Value[:]),
		BlockNumber: 222,
	})
```

and finally `ticket_submission_test.go`:
```
TicketValue: new(big.Int).SetBytes(t.Value[:]),
```

My point here is that we shouldn't have this function here, it's used only for tests. And tests should ensure the input matches the specification (big-endian, unsigned integer)

pkg/beacon/relay/groupselection/groupselection_test.go
```diff
@@ -347,8 +345,7 @@ func (stg *stubGroupInterface) GetSubmittedTickets() ([]uint64, error) {
 	tickets := make([]uint64, len(stg.submittedTickets))
 
 	for i := range tickets {
-		valueBytes := common.LeftPadBytes(stg.submittedTickets[i].Value.Bytes(), 32)
-		tickets[i] = binary.BigEndian.Uint64(valueBytes)
+		tickets[i] = binary.BigEndian.Uint64(stg.submittedTickets[i].Value[:])
```
pdyraga, line 348: Let's add a comment:
```
// ticket bytes should be interpreted as a big-endian unsigned integers
```

pkg/chain/ethereum/ethereum_test.go
```diff
@@ -155,11 +155,11 @@ func TestPackTicket(t *testing.T) {
 		return bigInt
 	}
 
-	ticketValue := toBigInt("77475267169740498967948014258679832639111923451618263020575217281118610489031")
+	ticketValue := [8]byte{255, 255, 255, 255, 255, 255, 255, 255}
 	stakerValue := toBigInt("471938313681866282067432403796053736964016932944")
```
pdyraga, line 159: Let's get rid of `ticketValue` and `stakerValue` from here and inline them in test cases below.

pkg/beacon/relay/chain/group_ticket.go
lukasz-zimnoch: I added this method just for convenience but you're right. I will replace as you suggested.

lukasz-zimnoch:
@pdyraga all comments addressed! Ready for another shot!

PR #1559, opened 2020-04-06
Remove redundant CircleCI step

ngrinkevich:
We use openzeppelin test environment with its own version of ganache.


pdyraga:
`04:08` for the entire `test_solidity` job 🚀

PR #1560, opened 2020-04-06
Quicker relay entries

### LINKED ISSUE #1391: Quicker relay entries
We don't need to wait for 3 blocks for signature share phase to complete and then submit an entry to the chain. As soon as we have a t+1 valid signature shares we can continue.

lukasz-zimnoch:
Refs #1391 

This pull request removes the state machine for relay entry signing. Now, each signer keeps exchanging messages as long as it's needed (until the relay entry timeout at max) but no longer than it takes to receive `t + 1` valid shares. This way we reduce the relay entry execution time and make them quicker.

The pattern of a state machine works fine for DKG where we want to wait for a fixed amount of time and if someone is late, that's their loss. Relay entry signing is different because it's everyone's loss - every group member is slashed for a timeout so each of them is interested in submitting the result as quickly as possible.

Although, we still have a fixed result publication block step which is used to determine the moment when individual members become eligible to submit the result.


pkg/beacon/relay/entry/entry.go
```diff
@@ -33,26 +39,195 @@ func SignAndSubmit(
...
+	// timeout block is reached.
+	for len(receivedShares) < honestThreshold {
+		select {
+		case netMessage := <-receiveChannel:
+			message, ok := netMessage.Payload().(*SignatureShareMessage)
+			if !ok || group.IsMessageFromSelf(signer.MemberID(), message) {
+				continue
+			}
+
+			share, err := extractShare(
```
pdyraga, line 99: `extractAndValidateShare`? It took me a moment to find where the validation is 😆

pkg/beacon/relay/entry/entry.go
```diff
@@ -33,26 +39,195 @@ func SignAndSubmit(
...
+			"could not unmarshal signature share: [%v]",
+			err,
+		)
+	}
+
+	publicKeyShare, ok := groupPublicKeyShares[message.senderID]
+	if !ok {
+		return nil, fmt.Errorf(
+			"could not validate signature share; " +
+				"group public key share not found",
```
pdyraga, line 196: `group public key share for sender not found`? This can help avoid confusion about whose share it is.

pkg/beacon/relay/entry/entry.go
```diff
@@ -33,26 +39,195 @@ func SignAndSubmit(
...
+	selfShare := signer.CalculateSignatureShare(previousEntry)
+
+	go broadcastShare(ctx, signer.MemberID(), selfShare, channel)
+
+	receiveChannel := make(chan net.Message, 64)
+	channel.Recv(ctx, func(netMessage net.Message) {
+		receiveChannel <- netMessage
+	})
+
+	receivedShares := map[group.MemberIndex]*bn256.G1{
```
pdyraga, line 83: `recivedValidShares` perhaps?

pkg/beacon/relay/entry/entry.go
```diff
@@ -33,26 +39,195 @@ func SignAndSubmit(
...
-			relayChain:      relayChain,
-			blockCounter:    blockCounter,
-			signer:          signer,
-			previousEntry:   previousEntry,
-			honestThreshold: honestThreshold,
-		},
-		signingStartBlockHeight: startBlockHeight,
+	selfShare := signer.CalculateSignatureShare(previousEntry)
+
+	go broadcastShare(ctx, signer.MemberID(), selfShare, channel)
```
pdyraga, line 76: I see know how much that broadcast channel + ctx refactor was a good idea :heart:

pkg/beacon/relay/entry/submission.go
```diff
@@ -26,6 +26,8 @@ func (res *relayEntrySubmitter) submitRelayEntry(
 	newEntry []byte,
 	groupPublicKey []byte,
 	startBlockHeight uint64,
+	relayEntrySubmittedChannel <-chan uint64,
+	relayEntryTimeoutChannel <-chan uint64,
```
pdyraga, line 30: I think we should revert the change in this file. The messages from those channels can be read only one time. Either here or in `submitRelayEntry` function calling this one. 

I think we should be fine with the previous approach with just `onSubmittedResultChan`. When we are here, there is no really a possibility for a timeout. Submission is a non-blocking call, we submit to the pool, we don't wait for the transaction to be mined.

pkg/chain/local/local.go
```diff
@@ -306,6 +306,7 @@ func ConnectWithKey(
 			TicketSubmissionTimeout:    6,
 			ResultPublicationBlockStep: 3,
 			MinimumStake:               minimumStake,
+			RelayEntryTimeout:          12,
```
pdyraga, line 309: Why 12? It depends on `groupSize`, right? Where do we need it?

pkg/beacon/relay/entry/entry.go
lukasz-zimnoch: Indeed :)

pkg/beacon/relay/entry/submission.go
lukasz-zimnoch: Yep, but it is not a problem in the context we use those channels here. If any success/timeout signal arrives when we are in the message loop we will exit before reaching the submitter. If something arrives after the message loop, the submitter will nicely handle it.

pdyraga: Ah, it makes sense. Can we add a comment about it to `entry.go` where we call `submitter.submitRelayEntry(` ?

pdyraga:
@lukasz-zimnoch Please also merge the recent `master` in here.

pkg/chain/local/local.go
lukasz-zimnoch: Ah right, I changed.

pdyraga:
It works very well although we should remove `relayEntryGenerationTime` from operator contract or set it to `0` at least. Entry is immediately generated so we should be able to submit it right away:
```
17:56:20.735  INFO keep-beaco: new relay entry requested at block [1197] from group [0x0927826...8e] using previous entry [0x09a3c...05077e5e] beacon.go:66
17:56:21.910  INFO keep-entry: [member:1] submitting relay entry [0x07fb14...9a] on behalf of group [0x09278...8e] at block [1197] submission.go:59
17:56:25.315  INFO keep-entry: [member:50] leaving submitter; relay entry submitted by other member at block [1198] submission.go:82
```

I'll work on it in a separate PR, there is no need for blocking this one.

Issue #1529, opened 2020-03-30
Reassigning token grantee


pdyraga:
In case the KEEP token grantee has been compromised there should be a way to reassign the grantee to a new address.

PR #1562, opened 2020-04-06
Managed grants

eth-r:
Refs: #1529

Managed grants permit reassignment of grantee addresses, by having the smart contract act as the grantee towards the grant contract while letting the specified grantee issue instructions through it. 

Managed grants can be created with the `ManagedGrantFactory`, using `approveAndCall` to add funds to a grant pool and then calling `createGrant` which returns the address of the managed grant.


solidity/contracts/ManagedGrantFactory.sol
```diff
@@ -0,0 +1,96 @@
...
+        tokenGrant = TokenGrant(_tokenGrant);
+        nonRevocableStakingPolicy = GrantStakingPolicy(_nonRevocableStakingPolicy);
+        revocableStakingPolicy = GrantStakingPolicy(_revocableStakingPolicy);
+    }
+
+    function receiveApproval(
+        address _from,
+        uint256 _amount,
+        address _token,
+        bytes memory _extraData
```
pdyraga, line 39: I think we should mirror the API of `TokenGrant`and `createGrant` in the same transaction as `receiveApproval` using `_extraData` to pass all the details.

It seems more convenient - we can create grants from dApp without setting up the factory first and we don't need to think about functionalities like withdrawal from the factory. It's also consistent with what we have in `TokenGrant.sol` and cheaper.

solidity/contracts/ManagedGrantFactory.sol
```diff
@@ -0,0 +1,96 @@
...
+    ) public {
+        require(KeepToken(_token) == token, "Invalid token contract");
+        require(_amount <= token.balanceOf(_from), "Insufficient sender balance");
```
pdyraga, line 42: The balance check can be removed: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v2.4.0/contracts/token/ERC20/ERC20.sol#L158

We can also eliminate it from `TokenGrant.sol`. This was noted during the audit and somehow we missed to address it.

solidity/test/token_grant/TestManagedGrantFactory.js
```diff
@@ -0,0 +1,150 @@
...
+  let token, registry, tokenGrant, staking;
+  let permissivePolicy, minimumPolicy;
+  let minimumStake, grantAmount;
+
+  const grantCreator = accounts[0],
+        grantee = accounts[2],
+        operator = accounts[3],
+        beneficiary = accounts[4],
+        authorizer = accounts[5],
+        newGrantee = accounts[6],
```
pdyraga, line 34: They are not used

solidity/test/token_grant/TestManagedGrantFactory.js
```diff
@@ -0,0 +1,150 @@
...
+        unrelatedAddress = accounts[7];
+
+  let grantId, grantStart, returnedId;
```
pdyraga, line 37: `grantId` and `returnedId` are not used

solidity/test/token_grant/TestManagedGrantFactory.js
```diff
@@ -0,0 +1,150 @@
...
+const TokenStaking = contract.fromArtifact('TokenStaking');
+const TokenGrant = contract.fromArtifact('TokenGrant');
+const Registry = contract.fromArtifact("Registry");
+const PermissiveStakingPolicy = contract.fromArtifact("PermissiveStakingPolicy");
+const GuaranteedMinimumStakingPolicy = contract.fromArtifact("GuaranteedMinimumStakingPolicy");
+
+const ManagedGrant = contract.fromArtifact('ManagedGrant');
+const ManagedGrantFactory = contract.fromArtifact('ManagedGrantFactory');
+
+const nullAddress = '0x0000000000000000000000000000000000000000';
```
pdyraga, line 21: Not used.

solidity/test/token_grant/TestManagedGrantFactory.js
```diff
@@ -0,0 +1,150 @@
...
+
+  const grantUnlockingDuration = time.duration.days(60);
+  const grantCliff = time.duration.days(10);
+
+  const initializationPeriod = time.duration.minutes(10);
+  const undelegationPeriod = time.duration.minutes(30);
+
+  let factory;
+
+  let stakeFromManagedGrant;
```
pdyraga, line 47: Not used

solidity/contracts/ManagedGrant.sol
```diff
@@ -0,0 +1,108 @@
...
+            "Reassignment already requested"
+        );
+        require(_newGrantee != address(0), "Invalid new grantee address");
+        require(_newGrantee != grantee, "Unchanged new grantee address");
+
+        requestedNewGrantee = _newGrantee;
+    }
+
+    function confirmGranteeReassignment(address _newGrantee) public onlyManager {
+        address _requestedNewGrantee = requestedNewGrantee;
```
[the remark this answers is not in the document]
eth-r: It's a slight optimization, but probably not worth the lines of code.

solidity/contracts/ManagedGrantFactory.sol
eth-r: If we do this I'd like to request a change to the `TokenGrant` API (and other applicable APIs like `TokenStaking`) so that it could use [`abi.decode`](https://solidity.readthedocs.io/en/v0.5.10/units-and-global-variables.html#abi-encoding-and-decoding-functions) directly instead of manually de/constructing the data.

If the API here is identical to `TokenGrant.receiveApproval` we still can't pass `_extraData` straight through, as we need to get the managed grant address etc. and then change the relevant fields. `abi.decode` adds some slight overhead by requiring unnecessary empty bytes to be transmitted, but these functions aren't in any hot or even warm paths, and the resulting code would be significantly simpler.

eth-r: Another reason we can't really mirror the API is that we need to return the `ManagedGrant` address and `approveAndCall` only returns `bool`. We'd need to return the address in an event which then necessitates complex and somewhat fragile workarounds if another contract wants to perform those actions (like how the `ManagedGrantFactory` itself uses `numGrants()` to get the ID of the next grant, which I hate but can't get around).

One possible workaround would be to have _both_ the specific function which requires a setup and returns useful values, _and_ `receiveApproval` which can be called through the token but only gives info in events. In this case `receiveApproval` would be a relatively thin wrapper around `createGrant/stakeTokens/etc` and both use-cases would be accommodated.

eth-r: I made a separate PR for the API at #1592

ngrinkevich:
@eth-r FYI if you merge master it should fix all those inconsistent solidity test failures

solidity/contracts/ManagedGrantFactory.sol
pdyraga: Yeah, I don't like `approve` because of the re-approval attack but I have no better idea on how to solve it than what you proposed.

solidity/contracts/TokenGrant.sol
```diff
@@ -222,22 +222,19 @@ contract TokenGrant {
     function receiveApproval(address _from, uint256 _amount, address _token, bytes memory _extraData) public {
         require(ERC20Burnable(_token) == token, "Token contract must be the same one linked to this contract.");
         require(_amount <= token.balanceOf(_from), "Sender must have enough amount.");
-        require(_extraData.length == 137, "Invalid extra data length.");
```
pdyraga, line 225: Having #1592 in `master` we shouldn't see this change here if we merge in the recent `master` here.

pdyraga:
@ngrinkevich Just a heads up, this and #1592 will require changes to KEEP token dashboard.

solidity/contracts/TokenGrant.sol
eth-r: I merged it into `managed-grant`, not `master`

pdyraga: Ah, I missed this one. Cool.

solidity/test/token_grant/TestManagedGrantFactory.js
```diff
@@ -108,6 +108,25 @@ describe('TokenGrant/ManagedGrantFactory', () => {
...
+        factory.address, grantAmount, extraData, {from: grantCreator}
+      );
+      let event = (await factory.getPastEvents())[0];
+      expect(event.args['grantee']).to.equal(grantee);
+      let managedGrantAddress = event.args['grantAddress'];
+      let managedGrant = await ManagedGrant.at(managedGrantAddress);
+      let grantId = await managedGrant.grantId();
+      expect(await tokenGrant.availableToStake(grantId)).to.eq.BN(grantAmount);
+      expect(await managedGrant.grantee()).to.equal(grantee);
+      expect(await managedGrant.grantManager()).to.equal(grantCreator);
```
pdyraga, line 127: Do you think it makes sense to assert cliff, duration, and start parameter of the created grant? They are all `uint256`, it's easy to swap them and not notice.

pdyraga:
Three outstanding comments on my side and I think we should be good to go.

eth-r:
@pdyraga last comments addressed and one more interface change to the token grants so that we can assign a different grant manager than the sender of the tokens, and thus don't need to duplicate grant management code.

pdyraga:
@ngrinkevich Your turn!

PR #1564, opened 2020-04-07
Solid Ice: Bump solidity to 0.5.17 and freeze version

### LINKED ISSUE #1552: Pin solidity version to a final version for mainnet
Let's take the latest 0.5 version of Solidity and pin to exactly that version for mainnet release.

Shadowfiend:
Does exactly what it says, and no more :)

Closes #1552 .


solidity/package.json
```diff
@@ -45,7 +45,7 @@
     "ethereumjs-abi": "^0.6.5",
     "mocha": "^7.1.1",
     "mocha-junit-reporter": "^1.23.1",
-    "solc": "^0.5.7",
+    "solc": "^0.5.17",
```
pdyraga, line 48: Don't we want it to be `"solc": "0.5.17"`?

Shadowfiend: Yeah probably, though incidentally it doesn't matter because we explicitly set `solc` in `truffle-config.json`.

pdyraga:
Compiled locally to double-check. Tests are passing, client works as expected against the contracts.

PR #1569, opened 2020-04-09
Removed on-chain relay entry generation time

pdyraga:
Closes: #1391

Having #1560 merged, we are now submitting relay entry as soon as we are able to compute it. 
We need `t+1` valid shares and it shouldn't take more than a couple of seconds to get them. ⚡  

We no longer need relay entry generation time in the contract. The entry should be generated by clients almost instantly.


solidity/test/random_beacon_operator/TestPricingRewards.js
```diff
@@ -46,8 +46,7 @@ describe('KeepRandomBeaconOperator/PricingRewards', function() {
     let entryFeeEstimate = await serviceContract.entryFeeEstimate(0);
     await serviceContract.methods['requestRelayEntry()']({value: entryFeeEstimate, from: accounts[0]});
 
-    let relayEntryGenerationTime = await operatorContract.relayEntryGenerationTime();
-    await time.advanceBlockTo(relayEntryGenerationTime.addn(1).addn(await web3.eth.getBlockNumber()));
+    await time.advanceBlockTo((await web3.eth.getBlockNumber()) + 1);
```
ngrinkevich, line 49: I think `advanceBlockTo` wants BN by default so the shortest would be something like this :

```
await time.advanceBlockTo(web3.utils.toBN(await web3.eth.getBlockNumber()).addn(1));
```

solidity/test/random_beacon_operator/TestPricingRewards.js
```diff
@@ -59,8 +58,7 @@ describe('KeepRandomBeaconOperator/PricingRewards', function() {
     let entryFeeEstimate = await serviceContract.entryFeeEstimate(0);
     await serviceContract.methods['requestRelayEntry()']({value: entryFeeEstimate, from: accounts[0]});
 
-    let relayEntryGenerationTime = await operatorContract.relayEntryGenerationTime();
-    await time.advanceBlockTo(relayEntryGenerationTime.addn(2).addn(await web3.eth.getBlockNumber()));
+    await time.advanceBlockTo(await web3.eth.getBlockNumber() + 2);
```
ngrinkevich, line 61: same as above i.e. :
```
await time.advanceBlockTo(web3.utils.toBN(await web3.eth.getBlockNumber()).addn(2));
```

pdyraga:
@ngrinkevich Thanks for the review, the code is ready for another chance!

Issue #1551, opened 2020-04-04
Add Solidity, JS, and Go linting to CI


Shadowfiend:
We're doing this in pre-commit hooks in most cases right now, but adding it to CI allows for contributors who may not install pre-commit hooks to not sneak through poor code.

See also #20.

dimpar:
After briefly talking to @pdyraga , I'll take Solidity linting.

Shadowfiend:
You can probably directly forklift what I did for keep-ecdsa in [this commit](https://github.com/keep-network/keep-ecdsa/pull/368/commits/6c79487df7dfc6fa72d0f72f9afc768b73f5fa96). Still working on the Go side though.

dimpar:
Thanks Antonio, [PR](https://github.com/keep-network/keep-core/pull/1573) for `keep-core` is ready.

lionakhnazarov:
Closing as stale. Legacy infra/CI/local-dev task that no longer matches current tooling (Yarn 4, updated workflows, v2 stack) on main.

PR #1573, opened 2020-04-10
Solidity linting and testing in GitHub actions

dimpar:
Refs: https://github.com/keep-network/keep-core/issues/1551

In this PR we are adding Solidity linting and testing to GitHub workflow actions.

Based on [this commit](https://github.com/keep-network/keep-ecdsa/pull/368)


.github/workflows/contracts.yml
```diff
@@ -0,0 +1,72 @@
...
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      - name: Install dependencies
+        working-directory: ./solidity
+        run: |
+          npm ci
+          npm install ganache-cli@6.9.1
```
ngrinkevich, line 42: we switched to openzeppelin test enviroment recently so we don't need ganache-cli anymore

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Run tests
+        working-directory: ./solidity
+        run: |
+          node_modules/.bin/ganache-cli &
```
ngrinkevich, line 46: no need for this line as well pls see comment above, just having `npm run test` should be enough

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,72 @@
+name: Solidity
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - "solidity/contracts/**"
+      - "solidity/package.json"
+      - "solidity/package-lock.json"
```
Shadowfiend, line 10: Heh, we accidentally self-identified an issue here: these paths should include this file as well, otherwise the job won't run on changes to the job.

.github/workflows/contracts.yml
dimpar: thanks for the heads up! removed

.github/workflows/contracts.yml
dimpar: removed

.github/workflows/contracts.yml
dimpar: right, added this file for existing events

dimpar:
@Shadowfiend not sure how to remove `ci/circleci: test_solidity` check, since GH workflow replaces it. Any hint?

Shadowfiend:
I got you.

PR #1576, opened 2020-04-10
Run Random Beacon Docs: section about staking in KEEP token dashboard

pdyraga:
Extracted and slightly reworked some parts from https://github.com/keep-network/keep-core/pull/1455 to cover KEEP token dashboard terms, staking, and authorizations.

### DIFF
--- a/docs/run-random-beacon.adoc
+++ b/docs/run-random-beacon.adoc
@@ -6,6 +6,46 @@
 :numbered:
 toc::[]
 
+== Staking
+
+=== Terminology
+
+address:: Hexadecimal string consisting of 40 characters prefixed with "0x" uniquely identifying Ethereum account; 
+derived from ECDSA public key of the party. Example address: `0xb2560a01e4b8b5cb0ac549fa39c7ae255d80e943`.
+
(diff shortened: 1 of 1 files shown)


docs/run-random-beacon.adoc
```diff
@@ -6,6 +6,46 @@
 :numbered:
 toc::[]
 
+== Staking
+
+=== Terminology
```
Shadowfiend, line 11: I would actually move this to the end <_<

pdyraga: On it.

Shadowfiend: Eh, I actually wonder if this whole section shouldn't be its own doc linked from here. This doc is about operating the client 🤔

pdyraga: Yeah, I think we should give https://github.com/keep-network/keep-core/pull/1455 more attention though I think it makes sense to have it here temporarily before we can get staking doc out. What do you think?

Shadowfiend: Yeah that's probably fair. At least it'll give us something to point at.

pdyraga: Moved to the bottom.

PR #1583, opened 2020-04-12
Prefer `npx truffle` to a global truffle install

### LINKED ISSUE #1581: install.sh assumes a global truffle install
The docs should mention it, or we should use a local `npm exec truffle` type solution. This is probably an issue across the codebase

mhluongo:
No need to install `truffle` globally and deal with dueling version issues.

Closes #1581.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
           name: Generate solidity docs
           command: |
             cd solidity && npm install
-            node_modules/.bin/truffle compile
+            npx truffle compile
             mkdir -p output
             node scripts/generate-api-docs.js > output/index.adoc
             mkdir -p /tmp/docs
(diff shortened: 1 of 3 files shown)


Shadowfiend:
npx is its own global install, which is why none of our repos use it. If you want to go this path, add a script or two instead, which tbh isn’t a bad idea and future proofs us in case we move away from truffle.

Shadowfiend:
(Which is to say, I don’t consider npx anyone’s friend. It’s a dumb solution for an annoying problem. Not that I have an opinion or anything 😁)

mhluongo:
IIRC `npx` comes bundled with `npm` since version 5.2...

mhluongo:
If you still don't like it for some reason (again, no extra install), let's just reach into `node_modules` and avoid mismatching `truffle` issues?

Shadowfiend:
Hm, strange that I didn't have it. Still annoyed on a fundamental level that this is something that lives in a separate command, but willing to let that be a personal problem lol.

Shadowfiend:
(Though I'll say fwiw that there's still the framework agnosticness of doing these things as scripts.)

Shadowfiend:
Whelp, I accidentally stole this in https://github.com/keep-network/keep-core/pull/2021 . Closing, with apologies 🙈

Issue #1578, opened 2020-04-10
Confusing stake minimum error message when running without authorization


msimovic:
If a correct minimum stake has been delegated but KeepRandomBeaconOperator authorization has not happened the code will confusingly output as follows:

`20:06:56.614 CRITI  keep-main: stake is below the required minimum main.go:77`

pdyraga:
Thanks for reporting the issue @msimovic, I opened two PRs that should solve this problem.

PR #1587, opened 2020-04-13
Add more context to "no minimum stake" failure when starting client

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/1578

Previously, we were printing "stake is below the required minimum" which could be very confusing in case the minimum stake is there but the operator contract has not been authorized. Fixing it now, adding some more context about what's potentially happening.

```
Starting keep-core client...
13:53:38.486 CRITI  keep-main: no minimum KEEP stake or operator is not authorized to use it;
please make sure:
- operator address in the configuration is correct and it has KEEP tokens delegated
- operator contract has been authorized to operate on stake main.go:77
```


cmd/start.go
```diff
@@ -86,7 +86,10 @@ func Start(c *cli.Context) error {
 		return fmt.Errorf("could not check the stake [%v]", err)
 	}
 	if !hasMinimumStake {
-		return fmt.Errorf("stake is below the required minimum")
+		return fmt.Errorf(`no minimum KEEP stake or operator is not authorized to use it;
+please make sure:
+- operator address in the configuration is correct and it has KEEP tokens delegated
+- operator contract has been authorized to operate on stake`)
```
nkuba, line 92: Maybe it would be better to remove line breaks as logger outputs to plaintext and having this message printed across multiple lines may cause parsing problems.

pdyraga: Good point, updated and it's one-liner now:

> Starting keep-core client...
> 13:17:00.147 CRITI  keep-main: no minimum KEEP stake or operator is not authorized to use it; please make sure the operator address in the configuration is correct and it has KEEP tokens delegated and the operator contract has been authorized to operate on the stake main.go:77

PR #1591, opened 2020-04-13
Document and automate roles setup and upgrades 

ngrinkevich:
tbc...


pdyraga:
Contracts are on mainnet now so closing without merging.

Issue #1585, opened 2020-04-13
Do not publish DKG result with minimal number of supporting signatures


pdyraga:
For a group of 64 members, we need 33 signature shares to generate relay entry. DKG result is accepted if it has 33 supporting signatures. It means that in the worst scenario _[note]_, all 33 group members need to participate in relay entry signing. 

This is risky. When DKG result is not accepted, group members lose their ETH spent on tickets.  If DKG result is accepted and just some few members decide not to participate in threshold signing, the entire group will lose some of their KEEP.

One option is e.g. to require at least 55 signatures on DKG result so that it's accepted. It gives a safety margin of 22 inactive members.

_[note] In practice, the situation could be a little better because there is a chance that although some members do not support the result they may decide to stay in the group and participate in threshold signing because they support the group public key. It happens in a situation when for example there is a disagreement about a few inactive members._

eth-r:
We could use the DKG nofail threshold for this (75%, leaving 25% of the original or 33% of the reduced group as a safety margin). However, we should already be enforcing the 75% rule in the submitted DKG result; even if we only have 51% signatures the group should not be able to contain less than 75% members and thus the absence of a few signers wouldn't be catastrophic as the non-signer members would still be responsible for producing entries.

It would be possible for a colluding 51% adversary to draft unwilling members into the group to make it past the 75% threshold without their signatures, but in such a group the adversary bears 51~66% of the cost of failing to sign, so I don't think this is a massive problem in practice.

pdyraga:
> However, we should already be enforcing the 75% rule in the submitted DKG result; even if we only have 51% signatures the group should not be able to contain less than 75% members and thus the absence of a few signers wouldn't be catastrophic as the non-signer members would still be responsible for producing entries.

I am not following this one. Where do we enforce it, is it somewhere in the off-chain part?

On-chain, when DKG result is submitted, in `DKGResultVerification` library we check whether `signaturesCount >= self.signatureThreshold` and `self.signatureThreshold` is currently set to `33`.

eth-r:
We should be checking that the number of disqualified operators doesn't exceed 25%; I checked and it's true that we aren't performing this check currently. The lines:
```
        require(
            misbehaved.length <= self.groupSize - self.signatureThreshold,
            "Malformed misbehaved bytes"
        );
```
should be:
```
        require(
            misbehaved.length <= (self.groupSize - self.signatureThreshold) / 2,
            "Malformed misbehaved bytes"
        );
````

[...]

PR #1603, opened 2020-04-14
Adding signature safety margin on DKG result

dimpar:
Close https://github.com/keep-network/keep-core/issues/1585

In this PR we are increasing the DKG nofail signature threshold to 75%.  It gives us a 25% safety margin for relay entry signing.

### DIFF
--- a/pkg/beacon/relay/dkg/result/submission.go
+++ b/pkg/beacon/relay/dkg/result/submission.go
@@ -61,11 +61,15 @@ func (sm *SubmittingMember) SubmitDKGResult(
 		)
 	}
 
-	if len(signatures) < config.HonestThreshold {
+	// Chain rejects the result if it has less than 25% safety margin.
+	// If there are not enough signatures to preserve the margin, it does not
+	// make sense to submit the result.
+	signatureThreshold := config.HonestThreshold + (config.GroupSize-config.HonestThreshold)/2
+	if len(signatures) < signatureThreshold {
 		return fmt.Errorf(
-			"could not submit result with [%v] signatures for honest threshold [%v]",
+			"could not submit result with [%v] signatures for signature threshold [%v]",
 			len(signatures),
-			config.HonestThreshold,
+			signatureThreshold,
 		)
 	}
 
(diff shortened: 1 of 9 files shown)


pdyraga:
@eth-r @ngrinkevich would love to have your 👀 on this one.

solidity/test/random_beacon_operator/TestPublishDkgResult.js
```diff
@@ -5,28 +5,29 @@ const generateTickets = require('../helpers/generateTickets');
...
-const {createSnapshot, restoreSnapshot} = require("../helpers/snapshot.js")
-const {contract, accounts, web3} = require("@openzeppelin/test-environment")
-const {expectRevert, time} = require("@openzeppelin/test-helpers")
+const { createSnapshot, restoreSnapshot } = require("../helpers/snapshot.js")
+const { contract, accounts, web3 } = require("@openzeppelin/test-environment")
+const { expectRevert, time } = require("@openzeppelin/test-helpers")
 const stakeDelegate = require('../helpers/stakeDelegate')
 
-describe('KeepRandomBeaconOperator/PublishDkgResult', function() {
+describe.only('KeepRandomBeaconOperator/PublishDkgResult', function () {
```
dimpar, line 13: only

pkg/beacon/relay/dkg/result/submission.go
```diff
@@ -61,11 +61,15 @@ func (sm *SubmittingMember) SubmitDKGResult(
 		)
 	}
 
-	if len(signatures) < config.HonestThreshold {
+	// Chain rejects the result if it has less than 25% safety margin.
+	// If there are not enough signatures to preserve the margin, it does not
```
dimpar, line 65: Maybe we can also mention in the comment that we'd need total of 75% signatures?

pdyraga: Technically, from the client's perspective, it doesn't have to be 75% if the on-chain threshold parameter is different from what we currently have. That's also why do we need to re-calculate this parameter here.

eth-r:
The solidity side looks good to me

pdyraga:
And everything else looks good to me too. Let's merge.

PR #1604, opened 2020-04-14
Docker Publish: Contextual Tagging

sthompson22:
### Intro

We've been rolling on `latest` image publishes across our keep-dev and keep-test environments for awhile.  While this is fine for our dev environment, it's time we work against properly versioned images for our test/Ropsten environment.

### What we did

Circle builds that run against a git tag generate a Circle env var `CIRCLE_TAG`.  This is perfect for our Docker image tagging needs.  However, the `CIRCLE_TAG` enviroment variable isn't generated for builds that don't run on a tag.  e.g. all keep-dev / master builds.  We already use environment based Circle contexts to factor out environment specific configs from the Circle config space, so it's a natural fit to handle this in the same way.  The Circle Context environment variable is `DOCKER_IMAGE_TAG`.  For keep-dev it's set to `latest`, and for keep-test it's set to `$CIRCLE_TAG`.

### Test build

https://app.circleci.com/pipelines/github/keep-network/keep-core/6356/workflows/8a1eeca1-6605-4243-b842-8efa9969891a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,12 +144,12 @@ jobs:
           google-project-id: GOOGLE_PROJECT_ID
           registry-url: $GCR_REGISTRY_URL
           image: initcontainer-provision-keep-client
-          tag: latest
+          tag: DOCKER_IMAGE_TAG
       - gcp-gcr/push-image:
           google-project-id: GOOGLE_PROJECT_ID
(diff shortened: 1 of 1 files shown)


Shadowfiend:
High-level this doesn't seem crazy. Need to give it another once-over once we've restored the branch filters.

PR #1607, opened 2020-04-15
Add stake-timeout flag to keep-client start

ljk662:
Currently, the keep client immediately exits unsuccessfully if the operator address does not meet the following conditions:
- operator address in the configuration is correct and it has KEEP tokens delegated
- operator contract has been authorized to operate on stake

This is not ideal for certain use cases. To handle those use cases I would like the node to wait for a specified amount of time to allow the address to meet the conditions above before crashing.

This PR adds a `-w`/`--wait-for-stake` integer flag which will cause the node to crash after that number of minutes if the address is still in violation of any conditions listed above. Until the stake is found or the timeout is reached, every minute the node will log a WARN-level message about the missing stake.

If the flag is omitted, the start function will follow the same code path as before. 

Running the client using an operator address with no stake produces the following output:
```
keep-client --config /root/config/keep-client-config.toml start --stake-timeout 3
04:02:43.701 DEBUG keep-block: subscribing to new blocks blockcounter.go:163
0x87ea311606Cc04d86e956CcD7cc0926ea40dbf54 stake is below required minimum, wait duration: 0 minutes
0x87ea311606Cc04d86e956CcD7cc0926ea40dbf54 stake is below required minimum, wait duration: 1 minutes
0x87ea311606Cc04d86e956CcD7cc0926ea40dbf54 stake is below required minimum, wait duration: 2 minutes
04:08:44.913 CRITI  keep-main: timed out waiting for 0x87ea311606Cc04d86e956CcD7cc0926ea40dbf54 to have required minimum stake main.go:77
```

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -3,11 +3,13 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/keep-network/keep-common/pkg/chain/ethereum/ethutil"
 	"github.com/keep-network/keep-common/pkg/persistence"
 	"github.com/keep-network/keep-core/config"
 	"github.com/keep-network/keep-core/pkg/beacon"
+	"github.com/keep-network/keep-core/pkg/chain"
 	"github.com/keep-network/keep-core/pkg/chain/ethereum"
 	"github.com/keep-network/keep-core/pkg/net/key"
 	"github.com/keep-network/keep-core/pkg/net/libp2p"
@@ -20,9 +22,10 @@ import (
 var StartCommand cli.Command
 
 const (
-	bootstrapFlag = "bootstrap"
-	portFlag      = "port"
-	portShort     = "p"
+	bootstrapFlag    = "bootstrap"
+	portFlag         = "port"
+	portShort        = "p"
+	stakeTimeoutFlag = "stake-timeout"
 )
 
 const startDescription = `Starts the Keep client in the foreground. Currently this only consists of the
@@ -38,6 +41,9 @@ func init() {
 			Flags: []cli.Flag{
 				&cli.IntFlag{
 					Name: portFlag + "," + portShort,
+				},
+				&cli.IntFlag{
+					Name: stakeTimeoutFlag,
 				},
 			},
 		}
@@ -78,6 +84,12 @@ func Start(c *cli.Context) error {
 	stakeMonitor, err := chainProvider.StakeMonitor()
 	if err != nil {
 		return fmt.Errorf("error obtaining stake monitor handle [%v]", err)
+	}
+	if c.Int(stakeTimeoutFlag) != 0 {
+		err = waitForStake(stakeMonitor, config.Ethereum.Account.Address, c.Int(stakeTimeoutFlag))
+		if err != nil {
+			return err
+		}
 	}
 	hasMinimumStake, err := stakeMonitor.HasMinimumStake(
 		config.Ethereum.Account.Address,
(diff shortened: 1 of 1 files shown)


cmd/start.go
```diff
@@ -20,9 +22,10 @@ import (
 var StartCommand cli.Command
 
 const (
-	bootstrapFlag = "bootstrap"
-	portFlag      = "port"
-	portShort     = "p"
+	bootstrapFlag    = "bootstrap"
+	portFlag         = "port"
+	portShort        = "p"
+	stakeTimeoutFlag = "stake-timeout"
```
Shadowfiend, line 28: Would love to align the naming here to the excellent name you have for the function itself: `--wait-for-stake`. That also gives us a perfect short flag that won't conflict with any common expectations I think, `-w`.

Going to trigger the build here to see if there's anything lurking.

pdyraga:
Thanks for the contribution @ljk662!

> This is not ideal for certain use cases. To handle those use cases I would like the node to wait for a specified amount of time to allow the address to meet the conditions above before crashing.

Out of the interest, what are the use cases you have in mind? Failing the client to start was intentional as it can't go unnoticed that the operator is not properly staked and the client is not eligible for joining the network. Even if we implement a graceful wait period, bootstrap nodes will refuse to connect with this client.

Shadowfiend:
Might be interesting to fire up an issue template with a placeholder for Discord links actually… We discussed a bit [on Discord](https://discordapp.com/channels/590951101600235531/686646403057057968/699795159046553631). Quick version:

> B: having kube crashloop until stake appears is what we are trying to avoid.
> A: Curious though---why avoid the carshloop? [sic, I suck at typing]
> B: I typically alert on crashloops

cmd/start.go
ljk662: I agree. I have made this change

ljk662:
> Thanks for the contribution @ljk662!
> 
> > This is not ideal for certain use cases. To handle those use cases I would like the node to wait for a specified amount of time to allow the address to meet the conditions above before crashing.
> 
> Out of the interest, what are the use cases you have in mind? Failing the client to start was intentional as it can't go unnoticed that the operator is not properly staked and the client is not eligible for joining the network. Even if we implement a graceful wait period, bootstrap nodes will refuse to connect with this client.

I understand the intention of the client failing to start and because of this, I attempted to make my changes as minimal as possible. I wanted to make sure that if this flag is not set the code path is exactly the same as before.

As @Shadowfiend pointed out above (thanks @Shadowfiend), my use case is: if a potential user of a deployment system launches a node, which in turn generates an operator address, they need some time to delegate stake to that address. If I want to alert on crash loops in my cluster, that means I will be alerted during that delta time period each time a node is deployed.

cmd/start.go
```diff
@@ -154,3 +167,20 @@ func loadStaticKey(
...
+	waitMins := 0
+	for waitMins < timeout {
+		hasMinimumStake, err := stakeMonitor.HasMinimumStake(address)
+		if err != nil {
+			return fmt.Errorf("could not check the stake [%v]", err)
+		}
+		if hasMinimumStake {
+			return nil
+		}
+		fmt.Printf("%s stake is below required minimum, wait duration: %d minutes \n", address, waitMins)
```
Shadowfiend, line 181: My instinct says that this should be a log. Especially with @nkuba investigating the move to ipfs-log v2, which will allow logging to contain structured data, it would probably help surface things better.

Thoughts? Open to sticking to a `Printf` if that seems like overkill.

Shadowfiend: Oh one other note---we do typically wrap at 80 characters. Haven't figured out a great way to handle that on the auto-format front, so I have to apologize for that 😅

nkuba: I would vote to stick with log wherever it's possible. This allows operators to parse the logs in a consistent way.

nkuba: I'm wondering if this message maybe qualifies even to a warn level?

ljk662: I agree, using the logger is better. I've made this change. I've also shortened the log message so the line is under 80 chars.

ljk662: I went with info, thought it made sense. I can change to warn if you want @nkuba .

ljk662: On further thought, I think warn level is better. I've made that change.

Shadowfiend: A lot happened here since last I looked :laughing:

Shadowfiend:
Okayy, so usually we try to have more detailed commit messages; I'm a butt so I forgot to mention that when we first saw the PR, and we haven't yet landed a good version of the requirement in our contributing docs. I'm going to amend the commits to fill in some extra detail based on the PR description and then land this as-is!

Issue #1608, opened 2020-04-15
Handle duplicate events


pdyraga:
It may happen that Keep client receives duplicate events from Ethereum client. It often happens with Infura deployment. We need to make sure we don't trigger duplicate actions for those events. At a minimum, we should protect:
- `OnGroupSelectionStarted`
- `OnRelayEntryRequested`

When working on this issue we need to go through all other event handlers and validate if duplicate protection is needed.

See https://github.com/keep-network/keep-ecdsa/issues/355

PR #1610, opened 2020-04-15
Events tracking functionality

dimpar:
Refs https://github.com/keep-network/keep-core/issues/1608

In this PR we add tracking functionality for `GroupSelectionStarted` and `RelayEntryRequested` events. In prevents from receiving duplicate events that sometimes happen during the Infura deployment.

### DIFF
--- a/pkg/beacon/beacon.go
+++ b/pkg/beacon/beacon.go
@@ -63,12 +63,12 @@ func Initialize(
 		groupRegistry,
 	)
 
-	groupSelected := &event.GroupSelectionTrack{
+	pendingGroupSelections := &event.GroupSelectionTrack{
 		Data:  make(map[string]bool),
 		Mutex: &sync.Mutex{},
 	}
 
-	relayRequested := &event.RelayRequestTrack{
+	pendingRelayRequests := &event.RelayRequestTrack{
 		Data:  make(map[string]bool),
 		Mutex: &sync.Mutex{},
 	}
@@ -77,15 +77,15 @@ func Initialize(
 		previousEntry := hex.EncodeToString(request.PreviousEntry[:])
 		if node.IsInGroup(request.GroupPublicKey) {
 			go func() {
-				if ok := relayRequested.Add(previousEntry); !ok {
+				if ok := pendingRelayRequests.Add(previousEntry); !ok {
 					logger.Errorf(
-						"a new relay has been already requested using previous entry [0x%x]",
+						"relay entry requested event with previous entry [0x%x] has been registered already",
 						previousEntry,
 					)
 					return
 				}
 
-				defer relayRequested.Remove(previousEntry)
+				defer pendingRelayRequests.Remove(previousEntry)
 
 				logger.Infof(
 					"new relay entry requested at block [%v] from group [0x%x] using "+
@@ -130,18 +130,18 @@ func Initialize(
 
 		newEntry := event.NewEntry.Text(16)
 		go func() {
-			if ok := groupSelected.Add(newEntry); !ok {
+			if ok := pendingGroupSelections.Add(newEntry); !ok {
 				logger.Errorf(
-					"group selection with seed [0x%v] has already started",
+					"group selection event with seed [0x%x] has been registered already",
 					newEntry,
 				)
 				return
 			}
 
-			defer groupSelected.Remove(newEntry)
+			defer pendingGroupSelections.Remove(newEntry)
 
 			logger.Infof(
-				"group selection started with seed [0x%v] at block [%v]",
+				"group selection started with seed [0x%x] at block [%v]",
 				newEntry,
 				event.BlockNumber,
 			)
--- a/pkg/beacon/beacon.go
+++ b/pkg/beacon/beacon.go
@@ -3,6 +3,7 @@ package beacon
 import (
 	"context"
 	"encoding/hex"
+	"sync"
 
 	"github.com/ipfs/go-log"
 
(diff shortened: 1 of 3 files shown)


pkg/beacon/relay/event/events_tracking.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Remove clears a map from an entry that was used to start a group selection.
+func (gst *GroupSelectionTrack) Remove(entry string) {
+	gst.Mutex.Lock()
+	defer gst.Mutex.Unlock()
+
+	delete(gst.Data, entry)
+}
+
+// RelayRequestTrack is used to track requests of new entriees after RelayEntryRequested
```
pdyraga, line 41: `s/entriees/entries`

Also, shouldn't we say `for new entries`?

pkg/beacon/beacon.go
```diff
@@ -111,7 +128,24 @@ func Initialize(
 			)
 		}
 
+		newEntry := event.NewEntry.Text(16)
 		go func() {
+			if ok := groupSelected.Add(newEntry); !ok {
+				logger.Errorf(
+					"group selection with seed [0x%v] has already started",
```
pdyraga, line 135: `group selection event with seed [0x%x] already registered`? 

We should also take the opportunity and replace `%v` with `%x` in the info log below.

pkg/beacon/relay/event/events_tracking_test.go
```diff
@@ -0,0 +1,80 @@
+package event
+
+import (
+	"sync"
+	"testing"
+)
+
+func TestGroupSelectionTrack_Add(t *testing.T) {
```
pdyraga, line 8: Instead of doing it this way and setting up `GroupSelectionTrack` state manually on the test setup, I'd rather prefer to implement those tests the same way they are implemented in `keep-ecdsa`, that is using a public API (`Add` + `Remove`)

dimpar: Sure, I can do the same way as in `keep-ecdsa`, but what are the advantages / disadvantages? We still use `GroupSelectionTrack.Add/Remove` in both cases. I thought that a series of related checks might be a good candidate for table tests. Anyway, open for both.

pdyraga: Table tests are fine but I'd like to avoid setting internal state not using a public API and then making assumptions based on this state, e.g.:
```
groupSelected{Data: map[string]bool{"0x123456": true}, Mutex: &sync.Mutex{}},
```

What we want to test is that if `Add` is called twice with the same entry, it returns `true` for the first call and `false` for all subsequent calls until we `Remove`. Right now we are not testing this scenario. We are testing that if `{"0x123456": true}` is in `Data` we return `false` when called `Add` with `"0x123456"`. Technically, it's the same but practically, it's not what we wanted to test here. 

Also, if we structure the tests like in `keep-ecdsa` we treat `groupSelected` as a black box, we don't care what's happening inside as long as it's behaving as expected and we'll not have to modify unit tests if the internal implementation details change.

pkg/beacon/relay/event/events_tracking.go
dimpar: both versions sounds good to me actually. Changed

pkg/beacon/beacon.go
```diff
@@ -62,23 +63,45 @@ func Initialize(
...
 
-	relayChain.OnRelayEntryRequested(func(request *event.Request) {
-		logger.Infof(
-			"new relay entry requested at block [%v] from group [0x%x] using "+
-				"previous entry [0x%x]",
-			request.BlockNumber,
-			request.GroupPublicKey,
-			request.PreviousEntry,
-		)
+	groupSelected := &event.GroupSelectionTrack{
```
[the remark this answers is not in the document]
dimpar: ok

pkg/beacon/beacon.go
```diff
@@ -62,23 +63,45 @@ func Initialize(
...
+		Data:  make(map[string]bool),
+		Mutex: &sync.Mutex{},
+	}
+
+	relayRequested := &event.RelayRequestTrack{
```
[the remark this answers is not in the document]
dimpar: ok

pkg/beacon/beacon.go
```diff
@@ -62,23 +63,45 @@ func Initialize(
...
-				request.PreviousEntry,
-				relayChain,
-				signing,
-				request.GroupPublicKey,
-				request.BlockNumber,
-			)
+			go func() {
+				if ok := relayRequested.Add(previousEntry); !ok {
+					logger.Errorf(
+						"a new relay has been already requested using previous entry [0x%x]",
```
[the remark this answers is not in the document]
dimpar: ok

pkg/beacon/beacon.go
dimpar: done

pkg/beacon/relay/event/events_tracking_test.go
dimpar: done

dimpar:
I looked at the other events in `chain.go`. Looks like other events call "watchers" and I'm not sure how much value we'd get if we add a shield from duplicating events there.

dimpar:
Testing:
- emitted a duplicate event for group selection. Group selection has started and also I observed the error msg about duplicate group selection, which was expected. It didn't break the flow.
- emitted a duplicate event for a new relay entry. Entry was generated successfully and the error log was there informing about a duplicate relay entry event

pkg/beacon/beacon.go
```diff
@@ -111,7 +128,24 @@ func Initialize(
 			)
 		}
 
+		newEntry := event.NewEntry.Text(16)
```
pdyraga, line 131: Since we are using `[0x%x]` formatter I think we should be using `events.NewEntry` directly in `Errorf` and `Infof`. In the current setup we are hex-ing twice.

pkg/beacon/beacon.go
```diff
@@ -62,23 +63,45 @@ func Initialize(
...
+		Mutex: &sync.Mutex{},
+	}
+
+	pendingRelayRequests := &event.RelayRequestTrack{
+		Data:  make(map[string]bool),
+		Mutex: &sync.Mutex{},
+	}
 
+	relayChain.OnRelayEntryRequested(func(request *event.Request) {
+		previousEntry := hex.EncodeToString(request.PreviousEntry[:])
```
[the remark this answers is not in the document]
pdyraga: Specifically, here we are hexing twice:
```
logger.Errorf(
  "relay entry requested event with previous entry [0x%x] has been registered already",
  previousEntry,
)
```

pkg/beacon/beacon.go
dimpar: done

pkg/beacon/beacon.go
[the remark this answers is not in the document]
dimpar: If we get rid of `previousEntry` then in Add/Remove argument we'd need to do the conversion. It doesn't look clean and is redundant.
- left `previousEntry`
- fixed double hexing

Issue #1588, opened 2020-04-13
Minimum staking policy when establishing connection


pdyraga:
When two clients are about to connect with each other they execute a handshake. As a part of the handshake, they confirm the other peer has a minimum stake of KEEP. If not, handshake fails and the connection is not established. For every active connection, the minimum stake condition is monitored and if some client's minimum stake dropped below the required minimum, the connection is dropped.

It is all fine for beacon nodes. It may be problematic for ECDSA nodes. Even with a minimum stake, one client can participate in multiple keeps. If for one of those peers malicious signature is reported, client's stake is dropped to zero and all other clients will disconnect.

It means that for any other keep that client participates in as a signer, redemption is no longer possible.

There are two solutions:
- This is not a bug, it's a future. The malicious client is removed from the network. All keeps it participated in has to be liquidated
- We let that client remain in the network as long as it participates in some active keeps. Seems attractive, but that one malicious keep has nothing to lose but their bond. Not sure if this is enough collateral to prevent it from further misbehaving (e.g. trying to nuke some nodes).

I'd personally recommend the first option though there could be some implications I am not aware of.

eth-r:
In bonded ECDSA the node should still have its bond for each keep. With how the bonded ECDSA incentives are designed, the bond should be sufficient to incentivize good behavior from participants.

Especially with 3-of-3 keeps, if one member is removed it prevents the entire keep from functioning. ECDSA keeps shouldn't be designed to rely on TBTC behavior to cover edge cases, or if they are, this should be clearly documented as a way the user of the keeps is expected to act. Other applications might have different needs, so I don't think it's necessary to create a cascading failure where it can be avoided. Not to mention, KEEP stake may be lost because of e.g. other members in a beacon group failing to generate their shares and the entire group timing out. Thus bonded ECDSA keeps should retain members that have gone below the minimum KEEP stake.

With Beacon the high safety factor should be sufficient to avoid this cascading failure mode. It also suggests that unbonded ECDSA keeps should only be used in a honest (super)majority mode instead of all-of-all.

PR #1612, opened 2020-04-16
Firewall: define your own rules for connections with remote peers

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/1588

So far we had just one rule we had to enforce when establishing a connection: the remote peer needs to have a minimum stake and the minimum stake has to be maintained for the entire lifetime of the connection.

It is still the case but we are discovering that more rules are needed and that we need exceptions from some of the rules in certain circumstances.

For example, for ECDSA client, we should require a minimum stake if the client has no active keeps. If the given client does no longer have a minimum stake but it has a number of active keeps we should let the client join the network and execute signing for those remaining keeps.

To do that, we introduce `net.Firewall` interface which is a facade behind which we can hide even the most sophisticated remote peer validation rules. Instead of passing `chain.StakeMonitor` all the way down to libp2p internals, we now pass `net.Firewall` with a specific minimum stake policy defined in `firewall.go`.

From the functional side, no big changes here. However, this will allow defining more complex connection filtering rules for `keep-ecdsa`.

Bonus:
- Set testnet stake undelegation period to one minute

    Default undelegation period is 3 months and it is hard to execute various test scenarios with it. For testnet, stake initialization period is set to 1 second so adding a custom value for stake undelegation period does not seem weird.

- Extracted network key -> ECDSA key translation details to a separate function
- Fixed local implementation of `ConnectedPeers` function
- Increased firewall check tick from 1 minute to 10 minutes

    It does not make sense to check the minimum stake every minute. It only
generates noise and costs. 10 minutes check should be enough.

If you are not convinced yet, I'd like to note we are eliminating `libp2p->chain` dependency and we no longer have to create fake stake monitors for tests; we can use `net.NoFirewall` instead.

### DIFF
--- a/pkg/net/net.go
+++ b/pkg/net/net.go
@@ -166,13 +166,3 @@ type Firewall interface {
 	// describing what is wrong.
 	Validate(remotePeerPublicKey *ecdsa.PublicKey) error
 }
-
-// NoFirewall is an empty Firewall implementation enforcing no rules
-// on the connection.
-var NoFirewall = &noFirewall{}
-
-type noFirewall struct{}
-
-func (nf *noFirewall) Validate(remotePeerPublicKey *ecdsa.PublicKey) error {
-	return nil
-}
--- a/pkg/net/net.go
+++ b/pkg/net/net.go
@@ -153,15 +153,12 @@ type BroadcastChannel interface {
 type BroadcastChannelFilter func(*ecdsa.PublicKey) bool
 
 // Firewall represents a set of rules the remote peer has to conform to so that
-// a connection with that peer can be established. The rules are checked when
-// a new connection is established and periodically for all active connections.
-// If some of the expectations defined by firewall rules are no longer met, the
-// connection is dropped.
+// a connection with that peer can be approved.
 type Firewall interface {
 
 	// Validate takes the remote peer public key and executes all the checks
-	// needed to decide whether the connection with the remote peer should be
-	// established and maintained.
+	// needed to decide whether the connection with the remote peer can be
+	// approved.
 	// If expectations are not met, this function should return an error
 	// describing what is wrong.
 	Validate(remotePeerPublicKey *ecdsa.PublicKey) error
--- a/pkg/net/net.go
+++ b/pkg/net/net.go
@@ -151,3 +151,28 @@ type BroadcastChannel interface {
 // public key as its argument and returns true if the message should be
 // processed or false otherwise.
 type BroadcastChannelFilter func(*ecdsa.PublicKey) bool
+
+// Firewall represents a set of rules the remote peer has to conform to so that
+// a connection with that peer can be established. The rules are checked when
+// a new connection is established and periodically for all active connections.
+// If some of the expectations defined by firewall rules are no longer met, the
+// connection is dropped.
+type Firewall interface {
+
+	// Validate takes the remote peer public key and executes all the checks
+	// needed to decide whether the connection with the remote peer should be
+	// established and maintained.
+	// If expectations are not met, this function should return an error
+	// describing what is wrong.
+	Validate(remotePeerPublicKey *ecdsa.PublicKey) error
+}
+
+// NoFirewall is an empty Firewall implementation enforcing no rules
+// on the connection.
+var NoFirewall = &noFirewall{}
+
+type noFirewall struct{}
+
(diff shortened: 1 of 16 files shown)


pdyraga:
I tested it with four properly staked clients executing DKG and relay entry signing. All works as expected. Then, I used `manage-stake.js` script to undelegate one of the client's stake. Given that the delegation period for testnet and local environment is set to one minute (see https://github.com/keep-network/keep-core/pull/1612/commits/4c89496911fc873506c70572953b0319f6c474c4), after the next cycle of watchtower, I could see that the node with no stake is getting disconnected by everyone else. Accidentally, that node was one of the bootstrap nodes so I could easily observe no connection can be established again.

image

I'd appreciate some additional testing from reviewers here.

pkg/net/net.go
```diff
@@ -151,3 +151,28 @@ type BroadcastChannel interface {
...
+
+// NoFirewall is an empty Firewall implementation enforcing no rules
+// on the connection.
+var NoFirewall = &noFirewall{}
+
+type noFirewall struct{}
+
+func (nf *noFirewall) Validate(remotePeerPublicKey *ecdsa.PublicKey) error {
+	return nil
+}
```
lukasz-zimnoch, line 178: What do you think about moving `noFirewall` type to ` pkg/firewall/firewall.go` or making a separate file for each policy in `pkg/firewall` package? I think `net.go` should remain a place where we define only interfaces used by the network layer.

pkg/net/net.go
```diff
@@ -151,3 +151,28 @@ type BroadcastChannel interface {
...
+// connection is dropped.
+type Firewall interface {
+
+	// Validate takes the remote peer public key and executes all the checks
+	// needed to decide whether the connection with the remote peer should be
+	// established and maintained.
+	// If expectations are not met, this function should return an error
+	// describing what is wrong.
+	Validate(remotePeerPublicKey *ecdsa.PublicKey) error
+}
```
lukasz-zimnoch, line 168: I think we are going to far with the description here. All of the details about when firewall rules are checked and what are the final effects are implementation details and I think they should be in places where `net.Firewall` is used (`authenticated_connection.go`, `watchtower.go`). Here we can limit ourselves to:

```
// Firewall represents a set of rules the remote peer has to conform to so that
// a connection with that peer can be considered valid.
type Firewall interface {

	// Validate takes the remote peer public key and executes all the checks.
	// If expectations are not met, this function should return an error
	// describing what is wrong.
	Validate(remotePeerPublicKey *ecdsa.PublicKey) error
}
```

What do you think?

pkg/net/watchtower/watchtower.go
```diff
@@ -93,18 +93,15 @@ func (g *Guard) start(ctx context.Context) {
...
 				g.markAsChecking(connectedPeer)
-				go g.manageConnectionByStake(ctx, connectedPeer)
+				go g.manageConnectionByFirewall(connectedPeer)
 			}
 		}
 	}
 }
 
-func (g *Guard) manageConnectionByStake(ctx context.Context, peer string) {
+func (g *Guard) manageConnectionByFirewall(peer string) {
```
lukasz-zimnoch, line 102: `manageConnectionByFirewall` -> `checkFirewallRules`?

pkg/net/net.go
pdyraga: Done in https://github.com/keep-network/keep-core/pull/1612/commits/4838270d696b8dc87985f39fe9ea19ab5e91eae9, please let me know if you like it.

pdyraga:
Good points @lukasz-zimnoch, all comments addressed.

pkg/net/net.go
```diff
@@ -151,3 +151,15 @@ type BroadcastChannel interface {
 // public key as its argument and returns true if the message should be
 // processed or false otherwise.
 type BroadcastChannelFilter func(*ecdsa.PublicKey) bool
+
+// Firewall represents a set of rules the remote peer has to conform to so that
+// a connection with that peer can be approved.
+type Firewall interface {
```
nkuba, line 157: If we defined `Firewall` interface in `pkg/net` shouldn't the implementation be placed under `pkg/net/firewall` instead of `pkg/firewall`

pdyraga: It was my initial plan but then I realized we don't want `net` to depend on `chain`. Right now, `net` defines `Firewall` interface so it doesn't care what's behind it.

PR #1614, opened 2020-04-16
Fix revoked grant staking interactions

eth-r:
Currently revoking a grant when some of it is staked immediately returns the entire revoked amount to the grant manager, even though the staked tokens aren't present in the grant contract. If some of those staked tokens are later lost to penalties, the grant contract can enter a state of permanent token deficit. To correct this, use withdrawal pattern for revoked tokens and prevent withdrawal of tokens that are not available in the grant. To ensure that revoked tokens can't be held indefinitely by the grantee, add functions the grant manager can use to force undelegation or cancellation of revoked grants' stakes.


solidity/test/token_grant/TestTokenGrantRevoke.js
```diff
@@ -183,4 +200,108 @@ describe('TokenGrant/Revoke', function() {
...
+      operator,
+      magpie,
+      authorizer,
+      minimumStake,
+      grantId
+    );
+    await grantContract.revoke(grantId, { from: tokenOwner });
+
+    const grantDetails = await grantContract.getGrant(grantId);
+    const revokedAmount = grantDetails[3];
```
pdyraga, line 246: Not used.

solidity/test/token_grant/TestTokenGrantRevoke.js
```diff
@@ -183,4 +200,108 @@ describe('TokenGrant/Revoke', function() {
...
```
pdyraga, line 282: Not used.

solidity/test/token_grant/TestTokenGrantRevoke.js
```diff
@@ -183,4 +200,108 @@ describe('TokenGrant/Revoke', function() {
...
```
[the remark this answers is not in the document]
eth-r: I'm adding tests to check that the revoked amount is correct.

PR #1620, opened 2020-04-17
Fonts updates

r-czajkowski:
Closes: #1630 

This PR adds the Vitesse font to the KEEP token dashboard dapp and updates text styles.


dimpar:
found 2 moderate, 5 high vulnerabilities in scanned packages. But I think maybe we can fix them down the road, after all other waiting PRs are merged?

solidity/dashboard/src/components/DelegatedTokensTable.jsx
```diff
@@ -5,10 +5,11 @@ import UndelegateStakeButton from './UndelegateStakeButton'
 import StatusBadge, { BADGE_STATUS } from './StatusBadge'
 import { PENDING_STATUS, COMPLETE_STATUS } from '../constants/constants'
 import { DataTable, Column } from './DataTable'
+import Tile from './Tile'
 
 const DelegatedTokensTable = ({ delegatedTokens, cancelStakeSuccessCallback }) => {
   return (
-    <section className="tile">
+    <Tile className="tile">
```
dimpar, line 15: Just wanted to make sure that we don't need to add <LoadingOverlay> here?

solidity/dashboard/src/components/SideMenu.jsx
```diff
@@ -43,8 +43,10 @@ export const SideMenu = (props) => {
         )}
         <Web3Status />
         <div className="account-address">
-          <span className="text-label text-bold">ADDRESS&nbsp;</span>
-          <AddressShortcut classNames="text-small" address={yourAddress} />
+          <h5 className="text-grey-50">
```
dimpar, line 46: Is this needed for responsive design?

solidity/dashboard/src/components/DelegatedTokensTable.jsx
r-czajkowski: yep. I'll update

dimpar:
in TokenGrantOverview.jsx:
- do we want to change `<span className="text-label">` -> `<h5 className="text-grey-50">` ?
- do we want to use <Tile> instead of <section> `import Tile from './Tile' and use <Tile>` ?

r-czajkowski:
> in TokenGrantOverview.jsx:
> 
>     * do we want to change `<span className="text-label">` -> `<h5 className="text-grey-50">` ?
> 
>     * do we want to use `<Tile> : import Tile from './Tile'` instead of `section`  ?

I think we can leave it as is because this component will be changed in the next PR. I already have a new version of this component.

solidity/dashboard/src/components/SideMenu.jsx
r-czajkowski: yep. It's for smaller screens.

dimpar:
Tested while doing the review, looked good to me.

Issue #1600, opened 2020-04-14
Panic button improvements: soft disable or permanently approve


pdyraga:
We had a long discussion around the panic button and its undesirable ability to halt tBTC. There was a brief proposal to introduce the concept of a permanently approved contract that is not subject to panic buttoning, alongside a permanent authorization from an operator to grant authorization to a permanently-approved contract. Our understanding is that if we had such a permanently approved contract, the operator is effectively opting in to a scenario where they are assignable to and slashable by that operator contract unless and until the operator fully unstakes; that is, leaving that contract would require moving stake to a new staking contract, subject to all related wait periods.

We are currently talking about the two approaches (which could actually both land):
- Move the panic button to a soft-disable, in that any already-at-stake KEEP in that operator contract remains at stake, but no additional KEEP can be put at stake with that operator contract.
- Add permanently authorized contracts that can't be panic buttoned, meaning the only way to stop additional KEEP being at stake is to unstake the operator completely and, optionally, restake with a new operator who doesn't authorize the bad contract.

There are two draft propositions:
https://github.com/keep-network/keep-core/compare/soft-panic-button
https://github.com/keep-network/keep-core/compare/approval-permanence

PR #1623, opened 2020-04-17
Panic button per operator contract with a possibility of being disabled

pdyraga:
Closes: #1600

We now allow specifying a panic button per operator contract. Such individual panic buttons can be disabled what makes the operator contract permanently approved. There is a default panic button value that is assigned to each new operator contract.

The only role which can disable a panic button for operator contract is governance. We don't want the registry keeper to have a power of permanently approving operator contracts if this action can not be later undone.

### DIFF
--- a/solidity/contracts/Registry.sol
+++ b/solidity/contracts/Registry.sol
@@ -8,7 +8,9 @@ pragma solidity 0.5.17;
 contract Registry {
     enum ContractStatus {New, Approved, Disabled}
 
-    // Governance role is to enable recovery from key compromise by rekeying other roles.
+    // Governance role is to enable recovery from key compromise by rekeying
+    // other roles. Also, it can disable operator contract panic buttons
+    // permanently.
     address internal governance;
 
     // Registry Keeper maintains approved operator contracts. Each operator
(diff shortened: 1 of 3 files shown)


solidity/contracts/Registry.sol
```diff
@@ -69,45 +109,98 @@ contract Registry {
...
+            _operatorContract,
+            _panicButton
+        );
     }
 
-    function approveOperatorContract(address operatorContract) public onlyRegistryKeeper {
+    function disableOperatorContractPanicButton(address _operatorContract)
+        public
+        onlyForApprovedContract(_operatorContract)
+        onlyGovernance
```
pdyraga, line 141: I don't like the governance has to be involved here but at the same time we shouldn't give the registry keeper a power to permanently approve operator contracts if this action can not be later undone. The same applies to the default panic button which could be transferring it's panicking capabilities to another address.

eth-r: If we shift governance over to its own contract at some point, we can implement arbitrary ACL over there. Thus using governance as a catch-all "we'll sort this out later" grab bag isn't as awful as it may feel.

pdyraga:
Thanks @eth-r, https://github.com/keep-network/keep-core/pull/1623/commits/59a7f0e4e089344b1e73309cc2c64ef4cd917368 looks good to me :+1:

PR #1625, opened 2020-04-18
Slashing test improvements, make sure slashing/seizing cannot fail for empty stake

pdyraga:
Refs: https://github.com/keep-network/keep-ecdsa/issues/356

Two changes:
- Extracted `TokenStaking` slashing and seizing tests to a separate file. We were testing `TokenStaking` functionality inside of random beacon operator contract test file. Now it's all tested in `TokenStaking`-land.
- Added tests from seizing/slashing empty stake
    We need to make sure that if slashing/seizing is executed for operator not having anything on stake, we are not going to fail. It is possible that two slashings will be executed quickly one after another. For example, beacon relay entry timeout happens and immediatelly after that, someone reports signature fraud in ECDSA keep. If we slashed operator to zero for relay entry timeout, ECDSA fraud slashing cannot fail. Otherwise, it may fail the entire fraud report transaction.


solidity/test/token_stake/TestPunishment.js
```diff
@@ -0,0 +1,209 @@
...
+            // the first slash will slash to 0, the second one has nothing
+            // to slash; it should not fail
+            await stakingContract.slash(amountToSlash, [operator], { from: operatorContract })
+            await stakingContract.slash(amountToSlash, [operator], { from: operatorContract })
+
+            let balanceAfterSlashing = await stakingContract.balanceOf(operator)
+            expect(balanceAfterSlashing).to.eq.BN(0)
+        })
+
+        it("should fail when operator stake is not active yet", async () => {
```
eth-r, line 108: We should also test for a case where the stake is released. Slashing should fail, because it should not be trying to punish operators past the undelegation period without having placed a lock.

pdyraga: Added in https://github.com/keep-network/keep-core/pull/1625/commits/6311811ae4d73aaeed76f261c0123d3094a7ca7f, I hade additionally covered the case it's not possible to slash/seize unlocked stake after undelegation in https://github.com/keep-network/keep-core/pull/1625/commits/439a5792a8915b13b94f64bcd0fefb4982ae1e0d.

solidity/test/token_stake/TestPunishment.js
```diff
@@ -0,0 +1,235 @@
...
+                amountToSeize, rewardMultiplier, tattletale,
+                [operator], { from: operatorContract }
+            )
+
+            let operatorBalanceAfterSeizing = await stakingContract.balanceOf(operator)
+            let tattletaleBalanceAfterSeizing = await token.balanceOf(tattletale)
+
+            expect(operatorBalanceAfterSeizing).to.eq.BN(0)
+
+            // 1000000000000000000000 = (200000000000000000000100 * 5 / 100) * 10 / 100
```
eth-r, line 174: Should we test with a higher excess to ensure that the tattletale reward is calculated from the applied penalty, not the requested penalty?

pdyraga: Yep, very good point. Fixed in bbbe3dd2255.

PR #1628, opened 2020-04-19
Let the contract to delegate authority to another contract directly

pdyraga:
In some cases, when the operator contract creates another contract (e.g. factory) and wants to pass its authority to the created contract, going through claim mechanism may be a little too complex and cause problems with gas estimation. We solve it by another function `delegateAuthority` that has the same effect as `claimDelegatedAuthority` except that it is called by the contract granting the authorization and not by the contract claiming it.

### DIFF
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -76,7 +76,8 @@ contract TokenStaking is StakeDelegatable {
     mapping(address => LockUtils.LockSet) internal operatorLocks;
     uint256 public constant maximumLockDuration = 86400 * 200; // 200 days in seconds
 
-    // Granters of delegated authority to operator contracts.
+    // Grantees of delegated authority to operator contracts granting their
+    // authority.
     // E.g. keep factories granting delegated authority to keeps.
     // `delegatedAuthority[keep] = factory`
     mapping(address => address) internal delegatedAuthority;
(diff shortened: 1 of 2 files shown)


solidity/contracts/TokenStaking.sol
```diff
@@ -619,13 +671,27 @@ contract TokenStaking is StakeDelegatable {
...
+    /// @dev `msg.sender` has to be an approved operator contract.
+    /// Later, the claimant can slash, seize, place locks etc.
+    /// on operators that have authorized the `msg.sender`.
+    /// If the `msg.sender` is disabled with the panic button,
+    /// any recipients of delegated authority from it will also be disabled.
+    function delegateAuthority(address delegateContract)
+        public
+        onlyApprovedOperatorContract(msg.sender)
+    {
+        delegatedAuthority[delegateContract] = msg.sender;
```
eth-r, line 692: I considered an one-step process when making `claimDelegatedAuthority` but it violates some security principles. There is no check that the target of `delegateAuthority` is actually trying to get delegated authority from the caller, meaning that any approved contract A can delegate its authority to any other contract B, preventing B from functioning unless operators authorize A. This is particularly troublesome if a permanently approved contract turns out to have a bug which lets it delegate authority to arbitrary addresses, as it can shut down the entire network.

I settled on `claimDelegatedAuthority` because keep factories call `initialize` after creating the proxied keep, thus giving them time to record the keep's address before it tries to claim delegated authority. An alternative method would be to have the recipient of delegated authority define which contract it wishes to get delegated authority from, and check this in the call to `delegateAuthority`.

pdyraga: Good point. I went through the code once again, made some additional tests and I think we are good with what we have now. Even if the interaction here is probably more intuitive for the factory, the *claim* version is more robust. I'll close this one without merging.

PR #1629, opened 2020-04-20
Pass rpcurl/chainId during the build process- dapp

r-czajkowski:
Refs: #1538

This PR adds a .env file for the KEEP token dashboard dapp that stores environment variables such as:
* `ETH_RPC_URL`- to connect to the node via ws,
* `CHAIN_ID`- to verify the transaction signature.

Also, it updates a `Dockerfile` and a `build_token_dashboard_dapp` job to set environment variables during the build process.


solidity/dashboard/Dockerfile
```diff
@@ -12,6 +16,7 @@ RUN apk add --update --no-cache \
 
 COPY package.json /app/package.json
 COPY package-lock.json /app/package-lock.json
+COPY .env* /app/
```
liamzebedee, line 19: Should we be copying .env? We set the relevant variables above, is that enough?

liamzebedee, line 19: Should we be copying .env? We set the relevant variables above, is that enough?

solidity/dashboard/.env
```diff
@@ -0,0 +1,2 @@
+REACT_APP_ETH_RPC_URL=$ETH_RPC_URL
```
liamzebedee, line 1: Did you mean to include the `REACT_APP_` prefix? I think `ETH_RPC_URL` would suffice.

.circleci/config.yml
```diff
@@ -189,7 +189,10 @@ jobs:
           name: Run Docker build
           working_directory: ~/project/solidity/dashboard
           command: |
-            docker build --tag keep-dapp-token-dashboard .
+            docker build \
+            --build-arg ETH_RPC_URL=$ETH_RPC_URL \
```
liamzebedee, line 193: 👍

solidity/dashboard/.env
r-czajkowski: Yep. Pls check here https://create-react-app.dev/docs/adding-custom-environment-variables/

solidity/dashboard/Dockerfile
r-czajkowski: I think we should. We set environments for 'machine' and inject into the dapp using .env file. check here https://create-react-app.dev/docs/adding-custom-environment-variables/#expanding-environment-variables-in-env

r-czajkowski:
> I didn't realise we had these variables set in the CircleCI workflows. Nice find!

We don't have full rpc url. We would have to add but I think shouldn't be a problem. @sthompson22, @Shadowfiend

liamzebedee:
@r-czajkowski I believe we have `ETH_HOST`, which should be able to be combined with `ws://${ETH_HOST}:${ETH_WS_PORT}`?

r-czajkowski:
> @r-czajkowski I believe we have `ETH_HOST`, which should be able to be combined with `ws://${ETH_HOST}:${ETH_WS_PORT}`?

For keep-dev probably ok, but keep-test I'm not sure. Couse, we need infura url eg. `wss://ropsten.infura.io/ws/v3/YOUR-PROJECT-ID` 🤔

.circleci/config.yml
nkuba, line 193: Do we need RPC or WS? I thought we want to pass `wss` endpoint address, in such case this argument should be named `ETH_WS_URL`.

solidity/dashboard/Dockerfile
```diff
@@ -3,6 +3,10 @@ FROM node:11.15.0-alpine as build
 
 WORKDIR /app
 
+ARG ETH_RPC_URL
+ARG CHAIN_ID
+ENV ETH_RPC_URL=$ETH_RPC_URL
+ENV CHAIN_ID=$CHAIN_ID
```
nkuba, line 9: Maybe couple of updates for consistency with other scripts we've got?
```suggestion
ARG ETH_WS_URL
ARG ETH_NETWORK_ID
ENV ETH_WS_URL=$ETH_WS_URL
ENV ETH_NETWORK_ID=$ETH_NETWORK_ID
```

.circleci/config.yml
```diff
@@ -189,7 +189,10 @@ jobs:
...
+            --build-arg CHAIN_ID=$ETH_NETWORK_ID \
```
nkuba, line 194: Thoughts about renaming `CHAIN_ID` to `ETH_NETWORK_ID`?

```suggestion
            --build-arg ETH_NETWORK_ID=$ETH_NETWORK_ID \
```

r-czajkowski: It is related to the [eip-155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md). `networkId` is not used at all in the signing process (the `chainId` is used for signing). That's why I used `chainId` instead of `networkId`. But in most cases especially mainet(`chainId=1` `networkId = 1`) and public test networks eg. ropsten (`chainId = 3` `networkId = 3`) `networkId == chainId`. Maybe we should also add the `ETH_CHAIN_ID` var to circleCI context to avoid misunderstanding?

r-czajkowski:
Addressed with 78480be by adding a config.json file that stores websocket url for a given chainId.

PR #1643, opened 2020-04-20
Peer discovery tweaks

### LINKED ISSUE #1639: Peer discovery tweaks
The network does not heal itself as fast as we'd like to after total failure of all bootstrap peers. We need to tweak bootstrap rounds parameters to improve the situation.

lukasz-zimnoch:
Closes  #1639 

Here we introduce some tweaks related to peer discovery:
- Reduced the DHT refresh round period from `1 hour` to `30 minutes`. This change should make peer discovery a bit faster especially for edge peers. Also, this value should not harm performance with too frequent refreshes.
- Increased bootstrap round period from `10 seconds` to the default value of `30 seconds`. This change should help for problems with `libp2p` dial backoffs which can occur when an offline peer is dialled to often by frequent bootstrap rounds. Also, there are no benefits from running bootstrap round every `10 seconds` so `30 seconds` seems to be a reasonable value.
- Adjusted code to the new`go-libp2p-bootstrap` library version.

### DIFF
--- a/pkg/net/libp2p/libp2p.go
+++ b/pkg/net/libp2p/libp2p.go
@@ -53,10 +53,6 @@ const (
 	// FirewallCheckTick is the amount of time between periodic checks of all
 	// firewall rules against all peers connected to this one.
 	FirewallCheckTick = time.Minute * 10
-	// BootstrapCheckPeriod is the amount of time between periodic checks
-	// for ensuring we are connected to an appropriate number of bootstrap
-	// peers.
-	BootstrapCheckPeriod = 10 * time.Second
 )
 
 // Config defines the configuration for the libp2p network provider.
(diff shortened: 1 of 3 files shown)


pkg/net/libp2p/libp2p.go
```diff
@@ -209,13 +203,6 @@ func WithRoutingTableRefreshPeriod(period time.Duration) ConnectOption {
 	}
 }
 
-// WithBootstrapMinPeerThreshold set a minimal peer threshold for bootstrap process.
-func WithBootstrapMinPeerThreshold(threshold int) ConnectOption {
-	return func(options *ConnectOptions) {
-		options.BootstrapMinPeerThreshold = threshold
-	}
-}
```
pdyraga, line 217: Heads up, we still use this function in `keep-ecdsa` in `start.go`:
```
libp2p.WithBootstrapMinPeerThreshold(bootstrapMinPeerThreshold),
```

lukasz-zimnoch: Yep, I know. We can just remove it when we will update `keep-core` dependency.

PR #1646, opened 2020-04-20
Broadcast channel relay

### LINKED ISSUE #1641: Pubsub relay tweaks
TBD

lukasz-zimnoch:
Closes #1641

Here we introduce a new behavior of the client which now subscribes to all broadcast channels and relays received messages to connected peers. This way the client acts as an auxiliary node even if they don't take part in given protocol execution. This change should make the network
more robust.

### DIFF
--- a/pkg/altbn128/altbn128.go
+++ b/pkg/altbn128/altbn128.go
@@ -297,14 +297,3 @@ func (e *gfP2) pow(base *gfP2, exp *big.Int) *gfP2 {
 	}
 	return e
 }
-
-func UnmarshalG2Point(bytes []byte) (*G2Point, error) {
-	g2 := new(bn256.G2)
-
-	_, err := g2.Unmarshal(bytes)
-	if err != nil {
-		return nil, err
-	}
-
-	return &G2Point{g2}, nil
-}
--- a/pkg/altbn128/altbn128.go
+++ b/pkg/altbn128/altbn128.go
@@ -297,3 +297,14 @@ func (e *gfP2) pow(base *gfP2, exp *big.Int) *gfP2 {
 	}
 	return e
 }
+
+func UnmarshalG2Point(bytes []byte) (*G2Point, error) {
+	g2 := new(bn256.G2)
+
+	_, err := g2.Unmarshal(bytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return &G2Point{g2}, nil
+}
--- a/pkg/beacon/beacon.go
+++ b/pkg/beacon/beacon.go
@@ -104,6 +104,10 @@ func Initialize(
 			}()
 		}
 
+		go node.RelayMessagesForChannel(
+			hex.EncodeToString(request.GroupPublicKey),
+		)
+
 		go node.MonitorRelayEntry(
 			relayChain,
 			request.BlockNumber,
--- a/pkg/beacon/beacon.go
+++ b/pkg/beacon/beacon.go
@@ -5,8 +5,6 @@ import (
 	"encoding/hex"
 	"sync"
 
-	"github.com/keep-network/keep-core/pkg/altbn128"
-
 	"github.com/ipfs/go-log"
 
 	"github.com/keep-network/keep-common/pkg/persistence"
@@ -105,15 +103,7 @@ func Initialize(
 				)
 			}()
 		} else {
-			groupPublicKey, err := altbn128.UnmarshalG2Point(request.GroupPublicKey)
-			if err != nil {
-				logger.Warningf("could not relay messages: [%v]", err)
-			} else {
-				// Channel name used for relay entry request is the string
-				// representation of the compressed group public key G2 point.
-				name := hex.EncodeToString(groupPublicKey.Compress())
-				go node.RelayMessagesForChannel(name)
-			}
+			go node.ForwardSignatureShares(request.GroupPublicKey)
 		}
 
 		go node.MonitorRelayEntry(
--- a/pkg/beacon/beacon.go
+++ b/pkg/beacon/beacon.go
@@ -5,6 +5,8 @@ import (
 	"encoding/hex"
 	"sync"
 
+	"github.com/keep-network/keep-core/pkg/altbn128"
+
 	"github.com/ipfs/go-log"
 
 	"github.com/keep-network/keep-common/pkg/persistence"
@@ -102,11 +104,17 @@ func Initialize(
 					request.BlockNumber,
 				)
 			}()
+		} else {
+			groupPublicKey, err := altbn128.UnmarshalG2Point(request.GroupPublicKey)
+			if err != nil {
+				logger.Warningf("could not relay messages: [%v]", err)
+			} else {
+				// Channel name used for relay entry request is the string
+				// representation of the compressed group public key G2 point.
+				name := hex.EncodeToString(groupPublicKey.Compress())
+				go node.RelayMessagesForChannel(name)
+			}
 		}
-
-		go node.RelayMessagesForChannel(
-			hex.EncodeToString(request.GroupPublicKey),
-		)
 
 		go node.MonitorRelayEntry(
 			relayChain,
--- a/pkg/beacon/relay/node.go
+++ b/pkg/beacon/relay/node.go
@@ -71,10 +71,12 @@ func (n *Node) JoinGroupIfEligible(
 		}
 	}
 
+	// create temporary broadcast channel name for DKG using the
+	// group selection seed
+	channelName := newEntry.Text(16)
+
 	if len(indexes) > 0 {
-		// create temporary broadcast channel for DKG using the group selection
-		// seed
-		broadcastChannel, err := n.netProvider.BroadcastChannelFor(newEntry.Text(16))
+		broadcastChannel, err := n.netProvider.BroadcastChannelFor(channelName)
 		if err != nil {
 			logger.Errorf("failed to get broadcast channel: [%v]", err)
 			return
@@ -133,7 +135,16 @@ func (n *Node) JoinGroupIfEligible(
 				)
 			}()
 		}
+	} else {
+		go n.RelayMessagesForChannel(channelName)
 	}
 
 	return
+}
+
+// RelayMessagesForChannel enables an ability to subscribe to messages
+// from the given broadcast channel and relay them to other nodes thus
+// making the network more robust.
+func (n *Node) RelayMessagesForChannel(name string) {
+	n.netProvider.BroadcastChannelRelayFor(name)
 }
--- a/pkg/beacon/relay/node.go
+++ b/pkg/beacon/relay/node.go
@@ -3,8 +3,12 @@ package relay
 import (
 	"bytes"
 	"encoding/hex"
+	"fmt"
 	"math/big"
 	"sync"
+
+	bn256 "github.com/ethereum/go-ethereum/crypto/bn256/cloudflare"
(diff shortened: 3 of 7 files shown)


pkg/beacon/beacon.go
```diff
@@ -102,6 +104,16 @@ func Initialize(
...
 			}()
+		} else {
+			groupPublicKey, err := altbn128.UnmarshalG2Point(request.GroupPublicKey)
+			if err != nil {
+				logger.Warningf("could not relay messages: [%v]", err)
+			} else {
+				// Channel name used for relay entry request is the string
+				// representation of the compressed group public key G2 point.
+				name := hex.EncodeToString(groupPublicKey.Compress())
+				go node.RelayMessagesForChannel(name)
```
pdyraga, line 115: Thoughts on naming it `ForwardSignatureShares`?

It's just a name but we'll keep the "channel" abstraction lower. What matters here is that if you are not in the group, you can still help by forwarding signature shares if needed. I was also considering `RelaySignatureShares` but since we have `OnRelayEntryRequested` and `GenerateRelayEntry`, I think `Forward*` could be less confusing.

pkg/beacon/beacon.go
```diff
@@ -102,6 +104,16 @@ func Initialize(
...
 				)
 			}()
+		} else {
+			groupPublicKey, err := altbn128.UnmarshalG2Point(request.GroupPublicKey)
+			if err != nil {
+				logger.Warningf("could not relay messages: [%v]", err)
+			} else {
+				// Channel name used for relay entry request is the string
+				// representation of the compressed group public key G2 point.
+				name := hex.EncodeToString(groupPublicKey.Compress())
```
pdyraga, line 114: Instead of defining `UnmarshalG2Point` in `altbn128.sol` and having a part of the channel naming logic here, we can extract everything to `node.go`:
```
// channelNameForPublicKey takes group public key represented by G2 point
// and transforms it into a broadcast channel name.
// Broadcast channel name for group is the hexadecimal representation of
// compressed public key of the group.
func channelNameForPublicKey(groupPublicKey *bn256.G2) string {
	altbn128GroupPublicKey := altbn128.G2Point{G2: groupPublicKey}
	return hex.EncodeToString(altbn128GroupPublicKey.Compress())
}

// channelNameForPublicKey takes group public key represented by marshalled
// G2 point and transforms it into a broadcast channel name.
// Broadcast channel name for group is the hexadecimal representation of
// compressed public key of the group.
func channelNameForPublicKeyBytes(groupPublicKey []byte) (string, error) {
	g2 := new(bn256.G2)

	if _, err := g2.Unmarshal(groupPublicKey); err != nil {
		return "", fmt.Errorf("could not create channel name: [%v]", err)
	}

	return channelNameForPublicKey(g2), nil
}
```

and then, this part can be just:
```
} else {
  node.ForwardSignatureShares(request.GroupPublicKey)
}
```

pkg/altbn128/altbn128.go
```diff
@@ -297,3 +297,14 @@ func (e *gfP2) pow(base *gfP2, exp *big.Int) *gfP2 {
...
+func UnmarshalG2Point(bytes []byte) (*G2Point, error) {
+	g2 := new(bn256.G2)
+
+	_, err := g2.Unmarshal(bytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return &G2Point{g2}, nil
+}
```
pdyraga, line 310: I am not sure if we should have this implementation here. `altbn128.go` implements additional functions that are not available in `bn256` and `Unmarshal` is there. Anyway, please look at my next comment, I don't think we need this function here anyway.

pkg/beacon/relay/node.go
```diff
@@ -133,7 +135,16 @@ func (n *Node) JoinGroupIfEligible(
 				)
 			}()
 		}
+	} else {
+		go n.RelayMessagesForChannel(channelName)
```
pdyraga, line 139: Similarly to the entry signing counterpart, I'd name this function `ForwardDKGMessages`

pkg/beacon/relay/node.go
```diff
@@ -133,7 +135,16 @@ func (n *Node) JoinGroupIfEligible(
...
 	}
 
 	return
 }
+
+// RelayMessagesForChannel enables an ability to subscribe to messages
+// from the given broadcast channel and relay them to other nodes thus
+// making the network more robust.
```
pdyraga, line 147: Let's leave the details about subscription to `net` package. I'd suggest:
```
// ForwardDKGMessages enables the ability to forward DKG messages
// to other nodes even if this node has not been selected to the group.
```

pkg/net/net.go
```diff
@@ -62,6 +62,9 @@ type Provider interface {
 	// CreateTransportIdentifier creates a transport identifier based on the
 	// provided public key.
 	CreateTransportIdentifier(publicKey ecdsa.PublicKey) (TransportIdentifier, error)
+
+	// BroadcastChannelRelayFor creates a message relay for given channel name.
+	BroadcastChannelRelayFor(name string)
```
pdyraga, line 67: `ForwardMessagesForChannel`?

pkg/net/libp2p/channel_manager.go
```diff
@@ -30,6 +31,9 @@ type channelManager struct {
 	pubsub *pubsub.PubSub
 
 	retransmissionTicker *retransmission.Ticker
+
+	relaySubscriptionsMutex sync.Mutex
+	relaySubscriptions      map[string]*pubsub.Subscription
```
pdyraga, line 36: Maybe `forwarderSubscriptionsMutex` and `fowarderSubscriptions` to do not mix terms with threshold relay?

pkg/net/libp2p/libp2p.go
```diff
@@ -116,6 +116,19 @@ func (p *provider) CreateTransportIdentifier(publicKey ecdsa.PublicKey) (
 	return peer.IDFromPublicKey(&networkPublicKey)
 }
 
+func (p *provider) BroadcastChannelRelayFor(name string) {
```
pdyraga, line 119: `BroadcastChannelForwarderFor`?

pkg/net/libp2p/channel_manager.go
```diff
@@ -114,3 +119,53 @@ func (cm *channelManager) newChannel(name string) (*channel, error) {
 
 	return channel, nil
 }
+
+func (cm *channelManager) newRelay(name string, ttl time.Duration) error {
+	cm.relaySubscriptionsMutex.Lock()
+	defer cm.relaySubscriptionsMutex.Unlock()
+
+	if _, ok := cm.relaySubscriptions[name]; !ok {
```
pdyraga, line 127: What do you think about invalidating the previous relay and adding a new one? This way, we can increase a TTL. Right now, we may run into a situation when we should relay for the next hour but we stop one minute after the protocol started because the previous relayer was installed.

lukasz-zimnoch: I'm not sure. Subscribing to a topic is rather heavyweight and might take a similar amount of time as the signature shares exchanging process which is really quick now. Apart from that, we should look at this functionality as an auxiliary message relay. Even if it shutdowns during for example a relay entry (which is btw rather unlikely) nothing big happens.

lukasz-zimnoch: Maybe we can just set a higher value for relay entries? For example 24 hours or so?

lukasz-zimnoch:
@pdyraga all comment addressed!

pkg/beacon/relay/node.go
```diff
@@ -133,7 +139,51 @@ func (n *Node) JoinGroupIfEligible(
...
+	} else {
+		go n.ForwardDKGMessages(channelName)
 	}
 
 	return
 }
+
+// ForwardDKGMessages enables the ability to forward DKG messages
+// to other nodes even if this node has not been selected to the group.
+func (n *Node) ForwardDKGMessages(name string) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, agree.

pkg/net/libp2p/channel_manager.go
pdyraga: See it's not the problem here. The problem is that we install the forwarder and it's here for one hour. Even if we register another forwarder after 50 minutes, we are going to shutdown after 10 minutes.

pkg/beacon/relay/node.go
```diff
@@ -133,7 +139,51 @@ func (n *Node) JoinGroupIfEligible(
...
+	n.netProvider.BroadcastChannelForwarderFor(name)
+}
+
+// ForwardSignatureShares enables the ability to forward signature shares
+// messages to other nodes even if this node is not a part of the group which
+// signs the relay entry.
+func (n *Node) ForwardSignatureShares(groupPublicKeyBytes []byte) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Nope, this one is used from outside.

pdyraga:
@lukasz-zimnoch Do you mind opening an issue in libp2p pubsub asking about the reasoning of not forwarding messages in floodsub? For a network flash crash we observed, this behaviour does not help with message delivery. We could have a time-cache-based forwarder there. 

I think we are fine for now with the implementation proposed here but eventually, I think this should be solved in the network layer. The threshold relay should not be aware that this is happening, it simply shouldn't care.

pkg/net/libp2p/channel_manager.go
lukasz-zimnoch: Yeah, I know but I tried to avoid making this solution complex because as I said this is just an auxiliary forwarder which is not essential for our protocols to work. If you think we should address this edge case I can obviously make this logic more sophisticated.

lukasz-zimnoch: We can also make forwarders for relay entry channels eternal. This is the simplest way to mitigate the problem we are discussing here.

pdyraga: Let's leave it as-is and open an issue in `keep-core` referencing the issue from libp2p pubsub and also mention this problem there.

lukasz-zimnoch: Sure, let's do it.

lukasz-zimnoch:
> @lukasz-zimnoch Do you mind opening an issue in libp2p pubsub asking about the reasoning of not forwarding messages in floodsub? For a network flash crash we observed, this behaviour does not help with message delivery. We could have a time-cache-based forwarder there.
> 
> I think we are fine for now with the implementation proposed here but eventually, I think this should be solved in the network layer. The threshold relay should not be aware that this is happening, it simply shouldn't care.

I opened a question in pubsub repo: https://github.com/libp2p/go-libp2p-pubsub/issues/292

pkg/net/libp2p/channel_manager.go
lukasz-zimnoch: Created https://github.com/keep-network/keep-core/issues/1652

pdyraga:
Reviewed, tested, works as expected. We are aware it shouldn't be the final solution but given that:
- we are rate-limited by smart contract
- there is only one relay entry in progress at the time
- there is only one DKG in progress at the time

we should be fine.

This approach implements the strategy from https://github.com/libp2p/go-libp2p-pubsub/issues/28 where all nodes in the relay group must participate in the dissemination of broadcast channel (libp2p pubsub topic) and all nodes outside of the relay group volunteer in the dissemination. What's unfortunate in our solution is that it's not entirely solved in the network layer but leaks to the relay. We have to live with it for now.

PR #1649, opened 2020-04-21
Upgradeable Dashboard: Add contract resolution step to dashboard build

Shadowfiend:
This step ensures that the latest contract version will be resolved even
if docker layer caching is in place. It also makes it so that new
releases can bump the contract build and dashboard build both, knowing
the dashboard build will correctly resolve the contract build.

Also included: running keep-test builds on release branches, which start
with `releases\/`.

---------

After this, I think the release process for keep-core is this:
- Create a new release branch.
- Bump the changes.
- Commit + push.
- Approve the build.
- *PR optional.*

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -407,14 +407,14 @@ workflows:
               tags:
                 only: /^v.*/
               branches:
-                ignore: /.*/
+                only: /releases\/.*/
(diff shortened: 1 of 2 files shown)


Shadowfiend:
Rather extraordinarily here, I'd like all three approvals from the requested reviewers please :)

Issue #1545, opened 2020-04-02
Final on-chain parameters


pdyraga:
Finalize on-chain parameters.
- result publication block step
- group active time
- ticket submission timeout and group duration
- group profit fee
- DKG frequency

- curve shape (stake drop)
- when it starts getting effective / when it's off

eth-r:
Stakedrop begins at June 8 (interval 1), interval 0 covers everything before June 8 so should be dated May 8. 6 scheduled intervals (0 to 5 defined in reward schedule; from 6 onwards every interval allocates 100%) where 0 has a small reward allocation, 1 has the largest, and later intervals' allocations decay at some rate that gives a pretty curve.

If we want the cost of getting a random number to be somewhere in the region of $10, 500k gwei per member/32M (0.032 ETH) per group would yield roughly $5.5 of group profit at current prices. If we assume a minimum of 3 blocks per entry (`b = 3`) we have roughly 1M entries a year at full utilization, which gives an easy calculation of maximum beacon revenue.

Result publication block step is currently also used for entry timeout, which at 3 gives 192 blocks or around 30 to 50 minutes for an entry to be submitted before the group times out. This seems reasonable, and we could add an alarm functionality that tries to alert a human operator e.g. 5~10 minutes into the timeout to possibly see what's the problem and avoid getting slashed. I wouldn't be comfortable going much lower before we get experience and data in production.

Ticket submission timeout and result publication blocks are basically adjusting the rate at which new DKGs/group selections can be performed. If DKG frequency is expressed as `1/r`, whether group creation is constrained by time or pooled money at maximum utilization depends on whether `b*r > ticketSubmissionTimeout + timeDKG + maybe(resultPublicationBlockStep * groupSize)` (money-constrained) or not (time-constrained). Time-constrained DKG means that at max utilization the beacon accumulates a large pool of DKG fee money, which it will then use afterwards to keep the number of groups high. This seems moderately preferable, especially if `r` is low (otherwise we can get a lot of groups).

With 39 blocks for ticket submission, and 72 blocks for DKG, and a result publication step of 3 blocks, the total amount of time group selection+DKG can take varies from 112 to 303 blocks. This means that `r =< 37` is guaranteed to be time-constrained with an entry every 3 blocks, and `r =< 100` can become time-constrained.

[...]

PR #1650, opened 2020-04-21
Group expiration off-by-one bugfix

pdyraga:
We had a bug in `expireOldGroups` function of `Groups` library allowing the `expiredGroupOffset` to go too far and read from groups array at non-existing index. It resulted in blocking the beacon forever by `requestRelayEntry` reverting every single time it was called.

There are a couple of improvements here to fix this problem.

First of all, we fix the loop so that it never reads from groups array with an index equal or greater to groups array length. Secondly, in `selectGroup` function, we first `expireOldGroups` before checking if we have any active groups to proceed with. Last but not least, in `genesis` function we now run group expiration before checking the number of active groups.

There is still a little chicken and egg problem here addressed with appropriate documentation on `numberOfGroups` function. Specifically, `numberOfGroups` may return a number `> 0` and still, request relay entry will fail with no active groups message. The reason for that is because `numberOfGroups` takes into account the current state in the contract and `requestRelayEntry` modifies this state by running group expiration. If all group expired, `requestRelayEntry` can't continue and it reverts without updating chain state. There is no easy way to address this issue without a solid refactoring so I decided to explain it in `numberOfGroups` docs for now. If such a situation happens, it's now possible to run `genesis` to fix it as we'll first run group expiration there.

We found this problem because in one of the previous PRs we set the stake undelegation time to 60 sec for local and testnet environment. However, currently, the group expiration period is set to stake undelegation period. With 60-seconds undelegation, groups were expiring after 60 seconds. It complicates testing instead of making it easier so here we increase undelegation time back to the original value.

### DIFF
--- a/solidity/contracts/KeepRandomBeaconOperator.sol
+++ b/solidity/contracts/KeepRandomBeaconOperator.sol
@@ -131,10 +131,14 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
 
 
     /**
-     * @dev Triggers the first group selection. Genesis can be called only when
-     * there are no groups on the operator contract.
+     * @dev Triggers group selection if there are no active groups.
      */
     function genesis() public payable {
+        // If we run into a very unlikely situation there are no active
+        // groups on the contract because of slashing and groups terminated
+        // or because beacon has not been used for a very long time and all
+        // groups expired, we first want to make a cleanup.
+        groups.expireOldGroups();
         require(numberOfGroups() == 0, "Groups exist");
         // Set latest added service contract as a group selection starter to receive any DKG fee surplus.
         groupSelectionStarterContract = ServiceContract(serviceContracts[serviceContracts.length.sub(1)]);
@@ -671,8 +675,16 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
     }
 
     /**
-     * @dev Gets the number of active groups. Expired and terminated groups are
-     * not counted as active.
+     * @notice Gets the number of active groups as currently marked in the
+     * contract. Groups marked as expired or terminated does not count as
+     * active.
+     *
+     * @dev Even if numberOfGroups() > 0, it is still possible requesting for
+     * a new relay entry will revert with "no active groups" failure message.
+     * This function returns the number of active groups as they are currently
+     * marked on-chain. However, during relay request, before group selection,
+     * we run group expiration and it may happen some groups seen as active
+     * turns out to be expired.
      */
     function numberOfGroups() public view returns(uint256) {
         return groups.numberOfGroups();
--- a/solidity/contracts/KeepRandomBeaconOperator.sol
+++ b/solidity/contracts/KeepRandomBeaconOperator.sol
@@ -676,7 +676,7 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
 
     /**
      * @notice Gets the number of active groups as currently marked in the
-     * contract. This is the sate from when the expired groups were last updated
+     * contract. This is the state from when the expired groups were last updated
      * without accounting for recent expirations.
      *
      * @dev Even if numberOfGroups() > 0, it is still possible requesting for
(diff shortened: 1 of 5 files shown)


pdyraga:
### Testing
0. Revert bf443de to have stake undelegation period set to 60 seconds and group expiration period being the same.
1. Run DKG.
2. Request relay entry, see it works.
3. Wait for 60 seconds after DKG.
4. Request for a relay entry.

    Before: TX fails with no clear reason, it consumes an absurd amount of gas (`~6417134`), there is `"Missing opcode 0xfe",` in the execution trace.

    After: TX fails with `"No active groups"` reason and consumes around `~220000` gas.

5. Run `genesis()` see that it completes successfully
6. Request for a relay entry, see it works.

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -671,8 +675,16 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
     }
 
     /**
-     * @dev Gets the number of active groups. Expired and terminated groups are
-     * not counted as active.
+     * @notice Gets the number of active groups as currently marked in the
+     * contract. Groups marked as expired or terminated does not count as
```
nkuba, line 679: We're saying here that we return active groups and expired groups don't count as active. But in the next paragraph we state that it may include expired groups in specific cases. Maybe we could update `notice` description so we don't cause confusions?

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -131,10 +131,14 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
 
 
     /**
-     * @dev Triggers the first group selection. Genesis can be called only when
-     * there are no groups on the operator contract.
+     * @dev Triggers group selection if there are no active groups.
      */
     function genesis() public payable {
+        // If we run into a very unlikely situation there are no active
```
dimpar, line 137: "[..] situation there are [..]" ->  "situation when there are"?

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -671,8 +675,16 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
...
-     * not counted as active.
+     * @notice Gets the number of active groups as currently marked in the
+     * contract. Groups marked as expired or terminated does not count as
+     * active.
+     *
+     * @dev Even if numberOfGroups() > 0, it is still possible requesting for
+     * a new relay entry will revert with "no active groups" failure message.
+     * This function returns the number of active groups as they are currently
+     * marked on-chain. However, during relay request, before group selection,
+     * we run group expiration and it may happen some groups seen as active
```
dimpar, line 686: "happen some groups" -> "happen that some groups" ?

solidity/migrations/2_deploy_contracts.js
```diff
@@ -27,7 +27,6 @@ module.exports = async function(deployer, network) {
   // Set the stake initialization period to 1 block for local development and testnet.
   if (network === 'local' || network === 'ropsten' || network === 'keep_dev') {
     initializationPeriod = 1;
-    undelegationPeriod = 60;
```
nkuba, line 30: Don't we want to keep this value lower for the test environments than for production? Having that helped us to find the bug we're fixing in this PR.

pdyraga: We can make it lower but only if we merge changes from https://github.com/keep-network/keep-core/issues/1545 first. I'd propose to merge this PR as-is and after on-chain parameters are updated, consider changing this value for testnet.

solidity/contracts/KeepRandomBeaconOperator.sol
pdyraga: But we also say `... as currently marked in the contract`. 

I do agree it's confusing but I have no better idea how to express it. I am open to suggestions!

eth-r: "The number of active groups when the expired groups were last updated, without accounting for recent expirations"?

pdyraga: Please see https://github.com/keep-network/keep-core/pull/1650/commits/1a28cdad5e6d6ba449490a8e673f8604d84a6a66

eth-r:
Can be solved in its entirety: #1654

PR #1653, opened 2020-04-21
On-chain system parameters update

ngrinkevich:
Refs: #1545 

- Group member base reward: `1000000 * 1e11 wei = 1M Gwei`. 
    It gives group reward of `64*1M Gwei`.
- Group active time: `86400 * 7 / 15 blocks = ~7 days`.
- Stake initialization period: `43200 seconds = 12 hours`.
- Stake undelegation period: `7776000 seconds = 3 months`.
- DKG contribution margin: `5%`.
- Result publication block step stays at `3 blocks`.

### DIFF
--- a/solidity/test/helpers/initContracts.js
+++ b/solidity/test/helpers/initContracts.js
@@ -14,7 +14,7 @@ async function initContracts(KeepToken, TokenStaking, KeepRandomBeaconService,
     serviceContractImplV1, serviceContractProxy, serviceContract,
     operatorContract;
 
-  let dkgContributionMargin = 1, // 5% Represents DKG frequency of 1/20 (Every 20 entries trigger group selection)
+  let dkgContributionMargin = 5, // 5% Represents DKG frequency of 1/20 (Every 20 entries trigger group selection)
     withdrawalDelay = 1,
     stakeInitializationPeriod = 30, // In seconds
     stakeUndelegationPeriod = 300; // In seconds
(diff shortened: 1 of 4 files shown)


solidity/test/helpers/initContracts.js
```diff
@@ -14,7 +14,7 @@ async function initContracts(KeepToken, TokenStaking, KeepRandomBeaconService,
     serviceContractImplV1, serviceContractProxy, serviceContract,
     operatorContract;
 
-  let dkgContributionMargin = 1, // 1%
+  let dkgContributionMargin = 1, // 5% Represents DKG frequency of 1/20 (Every 20 entries trigger group selection)
```
pdyraga, line 17: In the comment, we say 5% but the value is not 1%.

pdyraga:
Looks good on my side except for DKG frequency. Issue #1545 specifies that the DKG frequency should be of `1/20 ~ 1/30`. We will have this situation with DKG contribution margin set to 5% only when the actual gas is the same as gas price ceiling used for DKG contribution fee estimation.

I tested locally with gas price set to `1Gwei` (same as Ropsten) and had DKG starting:
- 10 entries after genesis
- 5 entries after the previous DKG
- 1 entry after the previous DKG

I am fine with merging this as-is but it does not close #1545. We still need to work on DKG frequency.

ngrinkevich:
Looks good @pdyraga 👍

pdyraga:
I created separate issue for DKG frequency: https://github.com/keep-network/keep-core/issues/1660

PR #1654, opened 2020-04-21
Apply expiration correctly when calculating number of groups

eth-r:
We can return an up-to-date count of groups by calculating expired and terminated groups in view functions, and only applying those changes in `expireOldGroups()`.

### DIFF
--- a/solidity/contracts/libraries/operator/Groups.sol
+++ b/solidity/contracts/libraries/operator/Groups.sol
@@ -239,39 +239,52 @@ library Groups {
      */
     function numberOfGroups(
         Storage storage self
-    ) internal view returns(uint256) {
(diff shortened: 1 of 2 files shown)


pdyraga:
I was considering doing it but it leads to a problem with KEEP token dashboard. Specifically, KEEP token dashboard uses `getFirstActiveGroupIndex()` and `numberOfGroups()` to figure out from which groups rewards are available for withdrawal. And values returned from those functions need to be up to date with the current state as stored on-chain and not with the most recent situation, as we can calculate in view functions.

From KEEP token dahboard's perspective the current (`master`) behavior of `numberOfGroups()` is the correct one although I do agree it's not intuitive when we drain groups to zero.

pdyraga:
Closing without merging after releasing [v1.1.2](https://github.com/keep-network/keep-core/releases/tag/v1.1.2)

PR #1655, opened 2020-04-21
keep-prd: Bootstrap

sthompson22:
Here we provide the base for the environment that will run mainnet
resources.  This sets up a:

- Google project
- Terraform state backend bucket
- Basic VPC with public/private subnets

### DIFF
--- a/infrastructure/terraform/keep-prd/base.tf
+++ b/infrastructure/terraform/keep-prd/base.tf
@@ -1,0 +1,69 @@
+data "google_client_config" "default" {}
+
+# Configure the Google Cloud provider
+provider "google" {
+  version = "<= 1.19.0"
+  region  = "${var.region_data["region"]}"
+}
+
+provider "google-beta" {
+  version = "<= 1.19.0"
+  region  = "${var.region_data["region"]}"
+}
+
+provider "null" {
+  version = "<= 2.0.0"
+}
+
+provider "random" {
+  version = "<= 2.0.0"
+}
+
+provider "template" {
+  version = "<= 1.0.0"
+}
+
+/* Set your locals.
+ * Terraform doesn't allow for string interpolation in variable maps.
+ * We cheat it by defining a local. A local instance variable mapping
+ * allows for string interpolation in maps. Locals are also good for
+ * names who are a construct of multiple values, to keep module blocks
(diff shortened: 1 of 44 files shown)


infrastructure/terraform/keep-prd/base.tf
```diff
@@ -0,0 +1,69 @@
...
+
+  labels {
+    contact     = "${var.contacts}"
+    environment = "${var.environment}"
+    vertical    = "${var.vertical}"
+  }
+}
+
+module "project" {
+  source                = "git@github.com:thesis/terraform-google-bootstrap-project.git?ref=0.1.0"
```
Shadowfiend, line 46: Oh this is fun 🤓

infrastructure/terraform/keep-prd/base.tf
```diff
@@ -0,0 +1,69 @@
...
+  project_name          = "${var.project_name}"
+  org_id                = "${var.gcp_thesis_org_id}"
+  billing_account       = "${var.gcp_thesis_billing_account}"
+  project_owner_members = "${var.project_owner_members}"
+  project_service_list  = "${var.project_service_list}"
+  location              = "${var.region_data["region"]}"
```
Shadowfiend, line 52: Any reason this is called `location` when it's really `region`?

infrastructure/terraform/keep-prd/variables.tf
```diff
@@ -0,0 +1,87 @@
+# env vars
+variable "gcp_thesis_org_id" {
+  description = "The ID for the organization the project will be created under. Local ENV VAR."
```
Shadowfiend, line 3: When we say “local” here we literally mean “on the machine running `terraform apply`”?

sthompson22: Aye.

infrastructure/terraform/keep-prd/base.tf
sthompson22: I try to match the resource arguments to what we surface on the module.  In this case this is the `location` for the gcp bucket resource.  It doesn't have to be a region, I believe it can be zonal or multi-region, would have to double check.

We're doing everything in a regional context.

PR #1656, opened 2020-04-21
keep-prd: NATs

sthompson22:
Here we setup 3 NAT's across us-central1 to provide egress traffic routing for GKE and private subnet instances.

- 3 static IP's for each NAT
- us-central1-a/b/c NAT instance deployment
- Network routes for nat routing - included as network tags for devices that need to route through the NATs
- NAT firewall rules


Shadowfiend:
> included as network tags

Does this mean “labels”?

PR #1657, opened 2020-04-21
keep-prd: GKE + VPN

sthompson22:
### Intro

Includes a regional GKE cluster and VPN to connect to the cluster.

There are a couple differences between this setup and lesser
environments worth noting:

1. We moved away from the openvpn module.  It's a single resource, it doesn't need to be provisioned from a module.  The helm_release is configured directly in the vpn.tf file.

2. We're using updated Helm and Kubernetes providers.  I wanted to give Helm v3 a shot to see if we could keep from poisoning the new env with all of the Tiller overhead nonsense.  This works, so we're going to keep it.  We can use what's configured here as a template for migrating lesser envs.

### GKE Setup

The GKE cluster is running across 3 us-central1 availability zones with 2 nodes per zone.  We run 2 nodes per zone to protect against a node issue  that would require running pods to be migrated elsewhere.  We need to consider this because we'll be running StatefulSets and the volumes associated with them are bound zonally. 

Aside from this the GKE cluster is a mirror config of lesser environments.

Side Note: We're really starting to miss out on newer GKE configurations by still being on Terraform 11, which pegs us to older Google Terraform Provider versions.

### VPN Setup

We had to tweak one configuration value compared to lesser envs.  This is due to configuration values changing in the newer version of the OpenVPN chart.  The configuration effect is the same though.

`openvpn.conf` -> `openvpn.serverConf`

### Bonus Improvement

Here we migrate away from Helm v2 to Helm v3.  This allows us to ditch Tiller and the esoteric setup that goes along with it.  In a later step I will migrate lesser environments to this setup.


infrastructure/terraform/keep-prd/variables.tf
```diff
@@ -99,3 +99,63 @@ variable "nat_gateway_ip" {
...
+  description = "Subnet for deploying GKE cluster resources."
+
+  default {
+    primary_ip_cidr_range = "10.8.0.0/16"
+
+    services_secondary_range_name    = "keep-prd-gke-services-secondary-range"
+    services_secondary_ip_cidr_range = "10.108.100.0/24"
+
+    cluster_secondary_range_name    = "keep-prd-gke-cluster-secondary-range"
+    cluster_secondary_ip_cidr_range = "10.108.0.0/20"
```
Shadowfiend, line 145: What are the secondary ranges for vs primary?

infrastructure/terraform/keep-prd/variables.tf
```diff
@@ -99,3 +99,63 @@ variable "nat_gateway_ip" {
...
+variable "openvpn" {
+  description = "Configuration values for the keep-prd VPN server."
+
+  default {
+    name                          = "openvpn"
+    namespace                     = "default"
+    helm_chart                    = "stable/openvpn"
+    helm_chart_version            = "4.2.2"
+    route_all_traffic_through_vpn = "false"
+    gke_master_cidr               = "172.16.0.0"
```
Shadowfiend, line 159: How come this doesn't have a length suffix?

Shadowfiend: Nm I see this below. Feels like calling this `_cidr` is weird tbh. This isn't proper CIDR notation. But I also don't know what the right name is hah.

PR #1663, opened 2020-04-22
keep-prd: IAM Members

sthompson22:
Here's the initial list of folks who can participate in the env.  This will likely evolve over time.


infrastructure/terraform/keep-prd/variables.tf
```diff
@@ -50,6 +50,23 @@ variable "project_owner_members" {
   ]
 }
 
+# module IAM members: editor
+variable "editor_iam_role" {
+  description = "Editor gives create/modify/destroy privs, no IAM."
```
Shadowfiend, line 55: Should this be `on IAM`?

Shadowfiend:
Good to go, let's do it.

Issue #1644, opened 2020-04-20
Update token amounts after slashing


dimpar:
When someone gets slashed, we don't update the amount of tokens displayed in the overview section.

PR #1664, opened 2020-04-22
Adding available balance of tokens after slashing

dimpar:
Closes: #1644

This PR calculates the staked amount for a given grant from delegations instead of `TokenGrant.staked`. [The `staked` field in the TokenGrant struct](https://github.com/keep-network/keep-core/blob/master/solidity/contracts/TokenGrant.sol#L46) is not updated after slashing and must be calculated based on the delegations via the `TokenStaking` contract.


pdyraga:
Is this one something that is still occurring? Should we undraft it?

dimpar:
Tested token slashing/seizing by running `misbehaved-groups.js` two times. Expected deducted KEEP amount was 400k.

Screenshot 2020-05-19 at 16 28 16
Screenshot 2020-05-19 at 16 29 22

PR #1665, opened 2020-04-22
Added steps to cross compile go binary

nkuba:
We use xgo because of some c references in go.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,9 @@ executors:
   docker-thesis-buildpack:
     docker:
       - image: thesisco/docker-buildpack:bionic
+  docker-xgo:
+    docker:
+      - image: karalabe/xgo-latest
 
 orbs:
   gcp-gcr: circleci/gcp-gcr@0.0.4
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -67,6 +70,46 @@ jobs:
...
+      - gcp-cli/initialize:
+          google-project-id: GOOGLE_PROJECT_ID
+          google-compute-zone: GOOGLE_COMPUTE_ZONE_A
+          # This param doesn't actually set anything, leaving here as a reminder to check when they fix it.
+          gcloud-service-key: GCLOUD_SERVICE_KEY
+      - run:
+          name: Upload client binaries
+          command: |
+            cd /tmp/keep-client/bin
+            gsutil -m cp * gs://${CLIENT_BIN_BUCKET}/keep-core/
```
Shadowfiend, line 111: (1) I don't think we should upload these anywhere during regular builds.
(2) I think we should attach these to GitHub releases somehow during RC builds.
(3) For non-RC builds we need to consider a good way to provide well-packaged and signed artifacts, but that means we'll need to do local builds.

lionakhnazarov:
Closing as stale. These cross-compile docs are years old and superseded by the current build instructions on main.

PR #1675, opened 2020-04-23
Keep all Registry roles public

pdyraga:
After the contract is deployed, we need to transfer those roles. To
confirm roles were transfered correctly, we should have the fields
public.


pdyraga:
Added them to events as well and updated tests. Build is finally green so ready for another chance.

PR #1678, opened 2020-04-23
Add adaptive staking policy

eth-r:
Customize stakeahead (in seconds and whether it considers cliff), and set the policy minimum as a multiple of the staking contract minimum.

### DIFF
--- a/solidity/contracts/AdaptiveStakingPolicy.sol
+++ b/solidity/contracts/AdaptiveStakingPolicy.sol
@@ -29,6 +29,10 @@ contract AdaptiveStakingPolicy is GrantStakingPolicy {
         // Multiplier for the minimum stake;
         // with a minimumMultiplier = 5
         // the policy permits staking 5 times the minimum stake.
+        // If the multiplier is 0,
+        // only the unlocked amount,
+        // including stakeahead if applicable,
+        // can be staked.
         uint256 minimumMultiplier,
         // Stakeahead time in seconds;
         // the policy permits staking the amount that will be unlocked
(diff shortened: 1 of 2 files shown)


pdyraga:
After https://github.com/keep-network/keep-core/pull/1678/commits/100bbc3c91a9b5bfd16127a713c4dff5369d4046 should we still consider tests as WIP or are we good to go here?

solidity/contracts/AdaptiveStakingPolicy.sol
```diff
@@ -0,0 +1,73 @@
...
+    using SafeMath for uint256;
+    using UnlockingSchedule for uint256;
+    uint256 minimumStake;
+    uint256 stakeaheadTime;
+    bool useCliff;
+
+    constructor(
+        address _stakingContract,
+        uint256 minimumMultiplier,
+        uint256 _stakeaheadTime,
```
pdyraga, line 37: Can we document this parameter in constructor docs? Ideally, all the parameters but this one is the most confusing for me.

solidity/contracts/AdaptiveStakingPolicy.sol
```diff
@@ -0,0 +1,73 @@
...
+        bool _useCliff
```
pdyraga, line 38: Actually, documenting this one could also help. It's also not clear to me at first what is it supposed to do.

eth-r:
Tests are still not comprehensive but no longer WIP as all important functionality is checked, and there isn't much that could be going wrong.

Issue #1631, opened 2020-04-20
Left side menu in KEEP token dashboard


r-czajkowski:
Let's update the left side menu according to [Figma](https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=2280%3A71)

![obraz](https://user-images.githubusercontent.com/57687279/79739876-16dbdb00-82ff-11ea-988a-0c7e82d741e8.png)

PR #1679, opened 2020-04-24
Left side menu updates

r-czajkowski:
Closes: #1631
Depends on: #1620

This PR updates styles in the left side menu according to the Figma view.

obraz


dimpar:
do we have an icon for "Token Grants"?

r-czajkowski:
> @r-czajkowski do we have an icon for "Token Grants"?

No, we don't have. The "Token grants" page is visible only for the KEEP owner.

dimpar:
Tested, console also looks much cleaner now.

PR #1682, opened 2020-04-24
Fix wait-for-stake log

ljk662:
Fixed a mistake that got through in the last PR. There are two objects supplied to a string with 1 template. This will fix it.


pdyraga:
Thanks for the contribution, I'll merge once the tests pass.

PR #1685, opened 2020-04-25
Service Contract Keeper and non-ownable KeepRandomBeaconOperator contract

pdyraga:
Some operator contracts, such as `KeepRandomBeaconOperator`, may want to specify a list of service contracts they want to work with. Service contracts can be added and removed from that list.

Previously, `KeepRandomBeaconOperator` contract had an owner which could add and remove contracts from that list and owner was set to the address which deployed the contract.

We do not want to have our contracts ownable so here we introduce service contract keeper role, set per the operator contract. Service Contract Keeper is authorized to manage service contract list for the given operator contract. Specifically, `KeepRandomBeaconOperator` checks if the address calling `addServiceContract` and `removeServiceContract` is the service contract keeper.

### DIFF
--- a/solidity/contracts/KeepRandomBeaconOperator.sol
+++ b/solidity/contracts/KeepRandomBeaconOperator.sol
@@ -3,6 +3,7 @@ pragma solidity 0.5.17;
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "openzeppelin-solidity/contracts/utils/ReentrancyGuard.sol";
 import "./TokenStaking.sol";
+import "./Registry.sol";
 import "./cryptography/BLS.sol";
 import "./utils/AddressArrayUtils.sol";
 import "./utils/PercentUtils.sol";
@@ -56,12 +57,10 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
     Groups.Storage groups;
     DKGResultVerification.Storage dkgResultVerification;
 
-    // Contract owner.
-    address internal owner;
-
     address[] internal serviceContracts;
 
-    // TODO: replace with a secure authorization protocol (addressed in RFC 11).
+    Registry internal registry;
+
     TokenStaking internal stakingContract;
 
     // Each signing group member reward expressed in wei.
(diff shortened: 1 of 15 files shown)


Shadowfiend:
How far out the window would you fling me if I asked that we use this opportunity to rename `Registry.sol` and `contract Registry` to `KeepRegistry.sol` and `contract KeepRegistry`? :grimacing:

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -143,17 +142,14 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
         startGroupSelection(_genesisGroupSeed, msg.value);
     }
 
-    /**
-     * @dev Throws if called by any account other than the owner.
-     */
-    modifier onlyOwner() {
-        require(owner == msg.sender, "Caller is not the owner");
+    modifier onlyServiceContractKeeper() {
```
mhluongo, line 145: Are we sure we want to introduce the term "keeper" anywhere? It already has a meaning [in the cryptocurrency space](https://medium.com/@rzurrer/keepers-workers-that-maintain-blockchain-networks-a40182615b66) and it's easy to confuse with "keep"

mhluongo:
> How far out the window would you fling me if I asked that we use this opportunity to rename `Registry.sol` and `contract Registry` to `KeepRegistry.sol` and contract `KeepRegistry`

That means it's time to rename `KeepRandomBeacon` to `RandomBeacon`, right? :upside_down_face:

solidity/contracts/KeepRandomBeaconOperator.sol
ngrinkevich: is `serviceContractUpgrader` a good name?  similar to existing `operatorContractUpgrader` that can add operator contracts to service contract so `serviceContractUpgrader` in this case adds service contracts to operator contract

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -56,12 +57,10 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
     Groups.Storage groups;
     DKGResultVerification.Storage dkgResultVerification;
 
-    // Contract owner.
-    address internal owner;
-
     address[] internal serviceContracts;
 
-    // TODO: replace with a secure authorization protocol (addressed in RFC 11).
```
ngrinkevich, line 64: apologies I should have removed this a while ago 😅

ngrinkevich:
This looks good @pdyraga 💯👏

solidity/contracts/KeepRandomBeaconOperator.sol
pdyraga: We already have `registryKeeper` in the `Registry` but I do agree and my fingers paused for a minute before typing it. I have no better idea how to name it but I am of course open for suggestions, renaming it is two minutes of work :)

pdyraga: Good idea @ngrinkevich, let's go with `serviceContractUpgrader`. This guy is doing the same as `operatorContractUpgrader` for the service contract.

PR #1686, opened 2020-04-26
Escrow contract

pdyraga:
A token holder contract allowing contract owner to set beneficiary of all tokens held by the contract and allowing the beneficiary to withdraw the tokens.

### DIFF
--- a/solidity/contracts/Escrow.sol
+++ b/solidity/contracts/Escrow.sol
@@ -34,7 +34,7 @@ contract Escrow is Ownable {
     // are no tokens in escrow, function fails.
     function withdraw() public {
         require(beneficiary != address(0), "Beneficiary not assigned");
-        require(msg.sender == beneficiary, "Caller is not a beneficiary");
+        require(msg.sender == beneficiary, "Caller is not the beneficiary");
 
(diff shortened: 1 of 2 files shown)


solidity/contracts/Escrow.sol
```diff
@@ -0,0 +1,45 @@
...
+        beneficiary = _beneficiary;
+        emit BeneficiaryUpdated(beneficiary);
+    }
+
+    // @notice Withdraws all tokens from escrow to the beneficiary.
+    // If the beneficiary is not set, caller is not the beneficiary, or there
+    // are no tokens in escrow, function fails.
+    function withdraw() public {
+        require(beneficiary != address(0), "Beneficiary not assigned");
+        require(msg.sender == beneficiary, "Caller is not a beneficiary");
```
Shadowfiend, line 37: not the beneficiary* otherwise we're implying there can be more than one :grimacing:

pdyraga: I felt there is something wrong with `a` here but that entire `the`, `a` thing is just too hard for me 😆

pdyraga: Done in beaec6669698790

Issue #1690, opened 2020-04-26
Invalid signature share


pdyraga:
Rarely but still, we can observe [`invalid signature share`](https://console.cloud.google.com/logs/viewer?interval=P1D&project=keep-test-f3e0&minLogLevel=0&expandAll=false&timestamp=2020-04-25T17:20:37.722000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22keep-test-f3e0%22%0Aresource.labels.location%3D%22us-central1%22%0Aresource.labels.cluster_name%3D%22keep-test%22%0Aresource.labels.namespace_name%3D%22default%22%0A%22invalid%20signature%20share%22&scrollTimestamp=2020-04-24T05:22:08.599218225Z&dateRangeEnd=2020-04-25T17:19:27.194Z&dateRangeUnbound=backwardInTime) warnings in logs when signing relay entry. This was reported on public Ropsten testnet and can be seen in GCP logs as well as on Boar Network private testnet. It seems that the reason is not a manipulation of malicious client (happened on private testnet as well).

The first time it occurred on Ropsten testnet is `2020-04-16 03:11:12.210 CEST` which is around `v0.13.0` release but I don't see anything in `v0.13.0` that could break anything for shares validation so it could be a coincidence.

PR #1697, opened 2020-04-27
Check incoming relay requests events against the chain

lukasz-zimnoch:
Refs #1690

If there are some problems with ethereum node synchronization it may happen that multiple blocks arrive at the same time. It may happen that those blocks contain multiple relay entry request events. Currently, each event will be processed by the client. This is wrong because only the latest event is valid from the chain's perspective and only one request can be processed at the same time. 

Here we introduce a condition that checks event block number against the current entry start block from the chain. This way we handle only the latest event and ignore the ones which belong to the past from the chain's viewpoint.


pkg/chain/gen/abi/KeepRandomBeaconOperator.go
```diff
@@ -28,7 +28,7 @@ var (
...
...e\":\"uint256[]\"}],\"name\":\"submitDkgResult\",\"outputs\":[],\"payable\":false,\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"constant\":false,\"inputs\":[{\"internalType\":\"bytes32\",\"name\":\"ticket\",\"type\":\"bytes32\"}],\"name\":\"submitTicket\",\"outputs\":[],\"payable\":false,\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"constant\":true,\"inputs\":[],\"name\":\"submittedTickets\",\"outputs\":[{\"internalType\":\"uint64[]\",\"name\":\"\",\"type\":\"uint64[]\"}],\"payable\":false,\"stateMutability\":\"view\",\"type\":\"function\"},{\"constant\":true,\"inputs\":[],\"name\":\"ticketSubmissionTimeout\",\"outputs\":[{\"internalType\":\"uint256\",\"name\":\"\",\"type\":\"uint256\"}],\"payable\":false,\"stateMutability\":\"view\",\"type\":\"function\"},{\"constant\":false,\"inputs\":[{\"internalType\":\"address\",\"name\":\"operator\",\"type\":\"address\"},{\"internalType\":\"uint256\",\"name\":\"groupIndex\",\"type\":\"uint256\"},{\"internalType\":\"uint256[]\",\"name\":\"groupMemberIndices\",\"type\":\"uint256[]\"}],\"name\":\"withdrawGroupMemberRewards\",\"outputs\":[],\"payable\":false,\"stateMutability\":\"nonpayable\",\"type\":\"function\"}]"
```
pdyraga, line 31: We will have a conflict with https://github.com/keep-network/keep-core/pull/1708 here and let's use the version from #1708 when resolving. It has all the required functions as well as the fix for `from` caller address.

pkg/beacon/relay/chain/chain.go
```diff
@@ -38,6 +40,9 @@ type RelayEntryInterface interface {
 	// supposed to submit a relay entry, did not deliver it within a specified
 	// time frame (relayEntryTimeout) counted in blocks.
 	ReportRelayEntryTimeout() error
+	// CurrentEntryStartBlock returns the starting block of the
+	// current relay entry.
+	CurrentEntryStartBlock() (*big.Int, error)
```
pdyraga, line 45: I asked @dimpar to rename it to `CurrentRequestStartBlock` in #1698. I missed to rename it in the contract earlier and we are having a mess in the naming. Everything but this one is `currentRequest*`.

pkg/beacon/beacon.go
```diff
@@ -74,12 +74,33 @@ func Initialize(
 	}
 
 	relayChain.OnRelayEntryRequested(func(request *event.Request) {
-		previousEntry := hex.EncodeToString(request.PreviousEntry[:])
+		currentEntryStartBlock, err := relayChain.CurrentEntryStartBlock()
+		if err != nil {
+			logger.Warningf("could not check current entry start block")
```
pdyraga, line 79: `*current request start block` please. We need to rename this function. It's my fault, sorry.

pkg/beacon/beacon.go
```diff
@@ -74,12 +74,33 @@ func Initialize(
...
+		// the current entry start block from the chain. Most likely, there
+		// are some problems with chain synchronization and multiple events
+		// arrived at the same time but only the last should be handled.
+		if request.BlockNumber != currentEntryStartBlock.Uint64() {
+			logger.Warningf(
+				"ignoring relay entry request with block number [%v]; "+
+					"current entry start block stored on-chain is [%v]",
+				request.BlockNumber,
+				currentEntryStartBlock.Uint64(),
+			)
```
pdyraga, line 93: Shouldn't we `return` here?

pkg/chain/gen/abi/KeepRandomBeaconOperator.go
lukasz-zimnoch: Ok, I'll wait for #1708

pkg/beacon/beacon.go
lukasz-zimnoch: Yeah, for sure. Thanks for catching this!

pkg/chain/gen/abi/KeepRandomBeaconOperator.go
pdyraga: Even better, please take a look at #1708 and review this code, I think we can merge it pretty quickly.

Issue #1691, opened 2020-04-26
Resume relay entry signing after client restart


pdyraga:
In #1684 we exposed all the information needed so that the client can join in-progress relay entry signing when it starts and belongs to the signing group. The client can call `isEntryInProgress()`, and then looking at `currentEntryStartBlock`, `currentRequestGroupIndex`, and `currentRequestPreviousEntry` join the signing process, similarly how clients join the in-progress signing process for ECDSA keeps.

PR #1698, opened 2020-04-27
Resuming relay entry signing after client restart

dimpar:
Refs: https://github.com/keep-network/keep-core/issues/1691

In the scenario, when a client crashes and restarts, we want to add functionality for it to rejoin the 
signing group under the condition that this client belongs to the signing group.


dimpar:
Testing scenarios:
1. Run 2 Nodes and sent a `genesis` request. After the group was formed, I killed the second Node -> sent a `relay` request -> and run the previously killed Node. It rejoined the signing and entry was submitted.
2. As 1) but also run a 3rd Node which didn't take part in `genesis`. It stayed silent and didn't join the signing process.

pkg/beacon/relay/chain/chain.go
```diff
@@ -38,6 +40,14 @@ type RelayEntryInterface interface {
...
 	// time frame (relayEntryTimeout) counted in blocks.
 	ReportRelayEntryTimeout() error
+	// IsEntryInProgress checks if a new relay entry is currently in progress.
+	IsEntryInProgress() (bool, error)
+	// GurrentEntryStartBlock returns a start block of a current entry.
+	CurrentEntryStartBlock() (*big.Int, error)
+	// CurrentRequestPreviousEntry returns previous entry of a current request.
+	CurrentRequestPreviousEntry() ([]byte, error)
+	// GroupPublicKey returns group public key.
+	GroupPublicKey() ([]byte, error)
```
pdyraga, line 50: `CurrentRequestGroupPublicKey`?

pkg/chain/ethereum/ethereum.go
```diff
@@ -405,6 +405,27 @@ func (ec *ethereumChain) ReportRelayEntryTimeout() error {
 	return nil
 }
 
+func (ec *ethereumChain) IsEntryInProgress() (bool, error) {
+	return ec.keepRandomBeaconOperatorContract.IsEntryInProgress()
+}
+
+func (ec *ethereumChain) CurrentEntryStartBlock() (*big.Int, error) {
```
pdyraga, line 412: I just realized we have `currentRequest*` for all fields but this one and it's my fault because I did not notice it when refactoring `KeepRandomBeaconOperator` contract. Can we take the opportunity and rename `currentEntryStartBlock` to `currentRequestStartBlock` in the contract?

pkg/beacon/relay/node.go
```diff
@@ -157,6 +158,61 @@ func (n *Node) ForwardSignatureShares(groupPublicKeyBytes []byte) {
...
+			logger.Errorf(
+				"failed getting a current relay entry start block: [%v]",
+				err,
+			)
+			return
+		}
+		groupPublicKey, err := relayChain.GroupPublicKey()
+		if err != nil {
+			logger.Errorf(
+				"failed getting a group public key: [%v]",
```
pdyraga, line 196: `failed to get current relay request's group public key` maybe?

pkg/beacon/relay/node.go
```diff
@@ -157,6 +158,61 @@ func (n *Node) ForwardSignatureShares(groupPublicKeyBytes []byte) {
...
+			logger.Errorf(
+				"failed getting previous entry on current request: [%v]",
+				err,
+			)
+			return
+		}
+		entryStartBlock, err := relayChain.CurrentEntryStartBlock()
+		if err != nil {
+			logger.Errorf(
+				"failed getting a current relay entry start block: [%v]",
```
pdyraga, line 188: `current relay request start block` please and of course we need to rename it in the contract because I failed to do that in my PR, heh.

pkg/beacon/relay/node.go
```diff
@@ -157,6 +158,61 @@ func (n *Node) ForwardSignatureShares(groupPublicKeyBytes []byte) {
...
+		if err != nil {
+			logger.Errorf(
+				"failed getting a group public key: [%v]",
+				err,
+			)
+			return
+		}
+
+		logger.Infof(
+			"client is atempting to rejoin the current signing group [0x%x]",
```
pdyraga, line 203: I'd rather say `the current signing process` because the client is in the signing group, it just wasn't active.

pkg/chain/local/local.go
```diff
@@ -470,6 +473,22 @@ func (c *localChain) ReportRelayEntryTimeout() error {
...
+}
+
+func (c *localChain) CurrentRequestPreviousEntry() ([]byte, error) {
+	return c.currentRequestPreviousEntry, nil
+}
+
+func (c *localChain) GroupPublicKey() ([]byte, error) {
+	return c.groups[0].groupPublicKey, nil
+}
+
```
pdyraga, line 491: Honestly, I think it's better if we `panic("not implemented")` for them, we can save some time in the future while figuring out why unit test does not work as expected.

And if we do not decide for panic we should assign `currentRequestPreviousEntry` in `SubmitRelayEntry` and `currentEntryStartBlock` somewhere else - hard to say where.

pkg/chain/gen/abi/KeepRandomBeaconOperator.go
```diff
@@ -28,7 +28,7 @@ var (
 )
```
pdyraga, line 28: We will have a conflict with https://github.com/keep-network/keep-core/pull/1708 here and let's use the version from #1708 when resolving. It has all the required functions as well as the fix for `from` caller address.

pkg/beacon/relay/node.go
dimpar: > because the client is in the signing group, it just wasn't active.

I don't think we know this at this point yet. Looks like we validate if it belongs to the currently signing group down the stream [here](https://github.com/keep-network/keep-core/blob/master/pkg/beacon/relay/relay.go#L108)

But I can say `process`, kinda gives the same idea on what is happening.

pkg/chain/ethereum/ethereum.go
dimpar: just to make sure, we want `currentRequestStartBlock` not `currentRequestEntryStartBlock`?

pdyraga: Yes, `currentRequestStartBlock`.

pkg/chain/local/local.go
dimpar: sure, replaced with `panic("not implemented")`

pkg/beacon/relay/node.go
dimpar: Replaced all occurrences.

pkg/beacon/relay/node.go
```diff
@@ -157,6 +158,61 @@ func (n *Node) ForwardSignatureShares(groupPublicKeyBytes []byte) {
...
+			logger.Errorf(
+				"failed to get a previous entry for current request: [%v]",
+				err,
+			)
+			return
+		}
+		entryStartBlock, err := relayChain.CurrentRequestStartBlock()
+		if err != nil {
+			logger.Errorf(
+				"failed to get a current relay request entry start block: [%v]",
```
pdyraga, line 188: It should be `current relay request start block`; we either have relay request or relay entry.

We should also align those errors. For example:
- `failed to get a previous entry for the current request: [%v]"`
- `failed to get a start block for the current request: [%v]"`
- `failed to get group public key for the current request: [%v]`"

pkg/beacon/relay/node.go
```diff
@@ -157,6 +158,61 @@ func (n *Node) ForwardSignatureShares(groupPublicKeyBytes []byte) {
...
+		if err != nil {
+			logger.Errorf(
+				"failed to get current relay request's group public key: [%v]",
+				err,
+			)
+			return
+		}
+
+		logger.Infof(
+			"client is atempting to rejoin the current signing processs [0x%x]",
```
pdyraga, line 203: `attempting` and `process`

pdyraga: Also, let's just say `attempting to rejoin the current signing process...`. I don't think we use `client is...` anywhere else, we just say `leaving message loop...`, `accepting signature share...`, `rejecting signature share...` and so on.

pkg/beacon/relay/node.go
dimpar: done

pkg/beacon/relay/node.go
dimpar: sure, done

PR #1702, opened 2020-04-27
CODEOWNERS for smart contracts

pdyraga:
Setting @Shadowfiend and @pdyraga as code owners for `keep-core` smart contracts deployed to mainnet in 1.0.1 release.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,13 @@
-*.sol @Shadowfiend @pdyraga
+/solidity/contracts/KeepToken.sol @Shadowfiend @pdyraga
+/solidity/contracts/TokenGrant.sol @Shadowfiend @pdyraga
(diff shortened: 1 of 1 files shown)


Shadowfiend:
My thought was we would only be CODEOWNERS on the contracts that are deployed, rather than on all of them :grimacing:

pdyraga:
I am hoping to freeze them tomorrow 

> My thought was we would only be CODEOWNERS on the contracts that are deployed, rather than on all of them 😬

Updated in 6fb7538. Side note, I realized only 4 contracts are not locked/not deployed yet. 😆

Shadowfiend:
Two things:
- Let's make sure we've hit all libraries as well.
- Let's add package.json, since dependencies live in there.

pdyraga:
Ready for another chance!

PR #1703, opened 2020-04-28
KeepToken is ERC20Detailed with name and symbol

nkuba:
We updated KeepToken to be ERC20Detailed with name `KEEP Token`, symbol `KEEP` and decimals `18`. The values will help list the token.

### DIFF
--- a/solidity/contracts/KeepToken.sol
+++ b/solidity/contracts/KeepToken.sol
@@ -1,6 +1,7 @@
 pragma solidity 0.5.17;
 
 import "openzeppelin-solidity/contracts/token/ERC20/ERC20Burnable.sol";
+import "openzeppelin-solidity/contracts/token/ERC20/ERC20Detailed.sol";
 
 
 /// @dev Interface of recipient contract for approveAndCall pattern.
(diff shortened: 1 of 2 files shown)


solidity/contracts/KeepToken.sol
```diff
@@ -1,21 +1,22 @@
...
-contract KeepToken is ERC20Burnable {
+contract KeepToken is ERC20Burnable, ERC20Detailed {
     string public constant NAME = "KEEP Token";
     string public constant SYMBOL = "KEEP";
     uint8 public constant DECIMALS = 18; // The number of digits after the decimal place when displaying token values on-screen.
     uint256 public constant INITIAL_SUPPLY = 10**27; // 1 billion tokens, 18 decimal places.
 
     /// @dev Gives msg.sender all of existing tokens.
-    constructor() public {
+    constructor() public ERC20Detailed(NAME, SYMBOL, DECIMALS) {
```
mhluongo, line 19: `ERC20` already has a `(name, symbol)` constructor- what does ERC20Detailed get us?

mhluongo: (I think, so many openzeppelin versions!)

solidity/test/TestToken.js
```diff
@@ -9,7 +9,15 @@ describe('TestToken', function() {
     account_two = accounts[1];
 
   before(async () => {
-    token = await KeepToken.new({from: account_one});
+    token = await KeepToken.new({ from: account_one });
+  });
+
+  it("sets token details", async function () {
```
mhluongo, line 15: :heart:

solidity/contracts/KeepToken.sol
Shadowfiend: It doesn't, no: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v2.3.0/contracts/token/ERC20/ERC20.sol .

Shadowfiend: (The most recent OpenZeppelin versions collapse a lot of the variants for some of the ERC tokens into a single contract, rather than offer 8 different possibilities.)

mhluongo: Yep, [that was it](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol)

liamzebedee: It gives us the decimals detail. 

Apparently we're using the all-caps members (eg. `DECIMALS`) but they didn't do anything in the end.

liamzebedee: > (I think, so many openzeppelin versions!)

🤦this.

PR #1709, opened 2020-04-28
keep-dev/test: Kill Bitcoind

sthompson22:
### Intro

Bitcoind was the backing Bitcoin node for our Electrum server.  We use Electrum to power the tbtc-dapp and tbtc-maintainers.

Soon ™️ we'll be replacing tbtc-maintainers with another service, relay-maintainers.  Relay-maintainers **requires** a bcoin Bitcoin node due to the RPC calls it uses to do its work.

While the relay-maintainer hooks directly up to bcoin, I decided to see about swapping out the Bitcoin node backend for Electrum so we don't have to support two different Bitcoin node deployments.

#### keep-dev
- Migrate Electrum to point at bcoin
- Uninstall bitcoind via Terraform
- Manually delete bitcoind persistent volume on Kube cluster
- Remove bitcoind Terraform state and files

#### keep-test
- Migrate Electrum to point at bcoin
- Uninstall bitcoind via Terraform
- Manually delete bitcoind persistent volume on Kube cluster
- Remove bitcoind Terraform state and files

### Testing

- Electrum syncs to bcoin
- tbtc-maintainers syncs to Electrum while synced to bcoin
- tbtc-dapp works for deposit (keep-test)


Shadowfiend:
Is bcoin deployed in keep-core or in tbtc? Also,

> Manually delete bitcoind persistent volume on Kube cluster

Why did we have to do this manually, is it not something we can manage via terraform?

sthompson22:
> Is bcoin deployed in keep-core or in tbtc? Also,

tbtc

> Why did we have to do this manually, is it not something we can manage via terraform?

This I need to look into - bitcoind was removed via `terraform --destroy --target helm_release.bitcoind`.  I assumed this would have cleaned up everything...but in double checking the PV was still there.

PR #1710, opened 2020-04-29
Check current entry status after signature error

lukasz-zimnoch:
In case a threshold signer returns with an error, the current entry status should be checked. If the current entry is not in progress anymore, it means the result was already submitted by an other signed and the error should not be logged. If the current entry is still in progress, something went wrong and signature error should be logged as usual.


pkg/beacon/relay/relay.go
```diff
@@ -148,10 +148,35 @@ func (n *Node) GenerateRelayEntry(
...
+						err,
+						signatureErr,
+					)
+					return
+				}
+
+				// Check if we failed because someone else submitted in the
+				// meantime or because something wrong happened with
+				// our transaction.
+				if !isEntryInProgress {
```
pdyraga, line 168: I think that this check should be inside of the submitter. This is submitter-specific logic (`submission.go`). One idea could be not to write to `errorChannel <- err` in this case.

lukasz-zimnoch: Yeah right, I changed.

pkg/beacon/relay/entry/submission.go
```diff
@@ -77,7 +77,35 @@ func (res *relayEntrySubmitter) submitRelayEntry(
 					}
 					errorChannel <- err
 				})
-			return <-errorChannel
+
+			entryErr := <-errorChannel
```
pdyraga, line 81: I think this will block until we can read from this channel and we don't always write to this channel. We should have a non-blocking read here with default return.

lukasz-zimnoch: We block here as before (`return <-errorChan`) and we always write either an error or `nil`. Actually I didn't changed the flow here.

PR #1711, opened 2020-04-29
Hardcode callback gas limit

ngrinkevich:
Make sure not to spend more than 2 million gas on a callback. This is to protect members from relay entry failure and potential slashing in case of any changes in .call() gas limit.

### DIFF
--- a/solidity/contracts/KeepRandomBeaconOperator.sol
+++ b/solidity/contracts/KeepRandomBeaconOperator.sol
@@ -492,6 +492,11 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
         // contract for the callback
         uint256 gasLimit = callbackFee.div(gasPriceCeiling);
 
+        // Make sure not to spend more than 6 million gas on a callback.
+        // This is to protect members from relay entry failure and potential
+        // slashing in case of any changes in .call() gas limit.
+        gasLimit = gasLimit > 6000000 ? 6000000 : gasLimit;
+
         bytes memory callbackReturnData;
         uint256 gasBeforeCallback = gasleft();
         (, callbackReturnData) = currentRequestServiceContract.call.gas(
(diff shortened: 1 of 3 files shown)


solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -492,6 +492,11 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
         // contract for the callback
         uint256 gasLimit = callbackFee.div(gasPriceCeiling);
 
+        // Make sure not to spend more than 6 million gas on a callback.
+        // This is to protect members from relay entry failure and potential
+        // slashing in case of any changes in .call() gas limit.
+        gasLimit = gasLimit > 6000000 ? 6000000 : gasLimit;
```
pdyraga, line 498: Can we make it 1-2M? It is quite conservative but gives a greater safety margin in case the gas limit goes down.

solidity/contracts/KeepRandomBeaconServiceImplV1.sol
```diff
@@ -221,7 +221,7 @@ contract KeepRandomBeaconServiceImplV1 is ReentrancyGuard, IRandomBeacon {
      * once a new relay entry has been generated. Callback contract must
      * declare public `__beaconCallback(uint256)` function that is going to be
      * executed with the result, once ready.
-     * @param callbackGas Gas required for the callback.
+     * @param callbackGas Gas required for the callback (6 million gas max).
```
pdyraga, line 224: Don't we want to fail `entryFeeEstimate` if `callbackGas` is too high?

solidity/contracts/KeepRandomBeaconServiceImplV1.sol
```diff
@@ -232,6 +232,11 @@ contract KeepRandomBeaconServiceImplV1 is ReentrancyGuard, IRandomBeacon {
         address callbackContract,
         uint256 callbackGas
     ) public nonReentrant payable returns (uint256) {
+        require(
+            callbackGas <= 6000000,
+            "Callback gas exceeds 6000000 gas limit"
+        );
```
pdyraga, line 238: Can we add a test making sure it fails?

ngrinkevich:
@pdyraga ready!

PR #1725, opened 2020-04-30
Updating keep-core contracts dependency 1.1.2

pdyraga:
After releasing beacon contracts to mainnet we are setting the
dependency to keep-core 1.1.2 mainnet version.

Also, updated the version of dashboard to `1.1.0-pre` since `1.0.0` has been
released with `v1.1.2` tag of `keep-core`.


pdyraga:
Closing without merging as outdated.

PR #1736, opened 2020-05-02
Added operator contract balance checks to the pricing test script

pdyraga:
We can use operator contract balance check to make sure the value of
rewards stored in the contract matches the total ETH balance stored on
the contract.

### DIFF
--- a/solidity/scripts/pricing-test.js
+++ b/solidity/scripts/pricing-test.js
@@ -59,7 +59,8 @@ module.exports = async function() {
                 serviceContractBalance,
                 dkgFeePool.toString(),
                 requestSubsidyFeePool.toString(),
-                dkgContributionMargin.toString()
+                dkgContributionMargin.toString(),
+                dkgFeePool.add(requestSubsidyFeePool).toString() === serviceContractBalance
             )
 
             console.log("Service Contract Summary (before request)");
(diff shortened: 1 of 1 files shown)


ngrinkevich:
That looks much cleaner than mine 😄

What do you think of adding an error message just in case there's a mismatch? i.e.

```
if (!operatorContractBalance.eq(rewardsSum.add(dkgSubmitterReimbursementFee))) {
    console.log("ERROR: Operator Contract balance mismatch");
}
```

solidity/scripts/pricing-test.js
```diff
@@ -44,6 +44,42 @@ module.exports = async function() {
...
+
+            let rewardsSum = web3.utils.toBN(0);
+            for (let i = 0; i < accounts.length; i++) {
+                rewardsSum = rewardsSum.add(web3.utils.toBN(prevRewards[i]));
+            }
+
+            const serviceContractBalance = await web3.eth.getBalance(contractService.address);
+            const dkgFeePool = await contractService.dkgFeePool();
+            const requestSubsidyFeePool = await contractService.requestSubsidyFeePool();
+            const dkgContributionMargin = await contractService.dkgContributionMargin();
```
pdyraga, line 56: Those functions are not currently exposed on the service contract. I'll add them in a separate PR. Let's hold on with merging this one.

pdyraga: Done in https://github.com/keep-network/keep-core/pull/1738

pdyraga:
Working on this one. Should have it ready pretty soon.

pdyraga:
We now have both service and operator contract checks so I put them in the table:

image

pdyraga:
One sec, something is wrong here...

PR #1738, opened 2020-05-02
Service contract pool values and DKG contribution margin made public

pdyraga:
DKG contribution margin, DKG fee pool, and request subsidy fee pool can be public without hurting anything. Having them public makes the system more transparent and helps with debugging (see #1736).


ngrinkevich:
@pdyraga `_baseCallbackGas` has to be updated to reflect the contract size change, just tested locally it should be 10226. Plus an update for "should correctly evaluate entry fee estimate" test

Do you want me to push the change ? I have it ready

pdyraga:
Yes @ngrinkevich, please do it! I have another PR in the queue that depends on this one.

pdyraga:
Looks good to me, thank you @ngrinkevich :+1:

Leaving the final decision and merge honours to you.

PR #1739, opened 2020-05-02
Moved contracts stubs to `solidity/tests/contracts` directory

nkuba:
Moved contracts that were defined under `solidity/contracts/stubs` to `solidity/tests/contracts` as it seems to be an appropriate directory for these files. The contracts are used in tests and shouldn't be mixed with regular contract sources.


Shadowfiend:
I think this causes issues with the openzeppelin stuff, from what I remember from tbtc.

nkuba:
Unfortunately you're right :(

pdyraga:
Knowing this approach is not compatible with OpenZeppelin test engine and that we love OpenZeppelin test engine, do we want to close this PR without merging?

nkuba:
Sure, let's close it.

PR #1740, opened 2020-05-03
Display token amount with metric suffix

r-czajkowski:
Depends on: #1721 

obraz


solidity/dashboard/src/utils/token.utils.js
```diff
@@ -0,0 +1,86 @@
...
+ */
+export function fromTokenUnit(amount, decimals = 18) {
+  amount = new BigNumber(amount)
+  return amount.times(new BigNumber(10).pow(new BigNumber(decimals)))
+}
+
+/**
+ * Returns a number with a metric suffix eg.:
+ * * 10000000 => { value: 10, suffix: 'M', formattedValue: '10M' }
+ * * 1000.4 => { value: 1.4, suffix: 'K', formattedValue: '10K'}
```
dimpar, line 47: shouldn't it be 1.4K ?

solidity/dashboard/src/utils/token.utils.js
```diff
@@ -0,0 +1,86 @@
...
+ * * 10000.4 => { value: 10, suffix: 'K', formattedValue: '10K' }
```
dimpar, line 48: 10.4K ?

solidity/dashboard/src/components/OwnedTokensOverview.jsx
```diff
@@ -33,7 +38,7 @@ const OwnedTokensOverview = ({ keepBalance, stakedBalance }) => {
           <h5 className="text-grey-70">staked</h5>
           <h4 className="text-grey-70">{displayAmount(stakedBalance)}</h4>
```
dimpar, line 39: do we want to display formatted amount here as well?

solidity/dashboard/src/utils/token.utils.js
r-czajkowski: We want to display decimals only if a value is less than 10.

solidity/dashboard/src/components/OwnedTokensOverview.jsx
r-czajkowski: I did it according to Figma.

PR #1743, opened 2020-05-04
Return an etherscan url depending on the network

r-czajkowski:
This PR provides an etherscan url depending on the current network. It also creates a component `ViewContractInBlockExplorer` that renders a link to contract on etherscan.


solidity/dashboard/src/components/ViewContractInBlockExplorer.jsx
```diff
@@ -0,0 +1,22 @@
+import React from "react"
+import { useEtherscanUrl } from "../hooks/useEtherscanUrl"
+
+const ViewContractInBlockExplorer = ({ contractAddress, text }) => {
+  const etherscanDefaultUrl = useEtherscanUrl()
+
+  return (
+    <a
+      href={`${etherscanDefaultUrl}/address/${contractAddress}#code`}
```
dimpar, line 9: just want to confirm.. we want to navigate to the byte code of the contract? That's why we add #code?

dimpar:
Tested: links to [etherscan](https://ropsten.etherscan.io/) look fine

solidity/dashboard/src/components/ViewContractInBlockExplorer.jsx
r-czajkowski: exactly, but rather to contract source Code. If the contract is verified, we can check also the source code

PR #1745, opened 2020-05-04
Fix callback surplus return

ngrinkevich:
If callback fails make sure to return any surplus back to the customer.

### DIFF
--- a/solidity/contracts/KeepRandomBeaconOperator.sol
+++ b/solidity/contracts/KeepRandomBeaconOperator.sol
@@ -17,6 +17,7 @@ interface ServiceContract {
     function entryCreated(uint256 requestId, bytes calldata entry, address payable submitter) external;
     function fundRequestSubsidyFeePool() external payable;
     function fundDkgFeePool() external payable;
+    function callbackSurplusRecipient(uint256 requestId) external view returns(address payable);
 }
 
 /// @title KeepRandomBeaconOperator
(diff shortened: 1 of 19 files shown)


solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -476,15 +477,23 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
         // slashing in case of any changes in .call() gas limit.
```
pdyraga, line 477: We probably don't need `uint256 callbackFee = currentRequestCallbackFee;` in line 468 and we can use `currentRequestCallbackFee` directly. It should let us save some gas/space.

solidity/test/random_beacon_service/TestPricing.js
```diff
@@ -6,7 +6,7 @@ const {contract, accounts, web3} = require("@openzeppelin/test-environment")
 const {time} = require("@openzeppelin/test-helpers")
 const CallbackContract = contract.fromArtifact('CallbackContract')
 
-describe('TestKeepRandomBeaconService/Pricing', function() {
+describe.only('TestKeepRandomBeaconService/Pricing', function() {
```
pdyraga, line 9: `only` - we need to remove it

solidity/contracts/KeepRandomBeaconOperator.sol
pdyraga: Removed in https://github.com/keep-network/keep-core/pull/1745/commits/5de53849ac781bdef8aae22b7d5da0bc01352604

PR #1747, opened 2020-05-04
Circle: Mr Clean

sthompson22:
### Intro 

When Circle migrations were implemented I did not account for the frequency of failed migrations _(this is a separate issue we're going to address, albeit a more complicated one than what we have here)_.  In the current state, if the `migrate_contracts` job fails the utilitybox cleanup that removes migration artifacts for a given build doesn't run.  After a number of failed builds _(somewhere around 10)_, the utilitybox disk fills up and migrations are hosed.

With keep-core, keep-ecdsa, and tbtc migrations all running agains the same utilitybox, failed migrations can stack up in short order.

We need a way to cleanup the migration dir whether the `migrate_contracts` job passes or fails.

### What We Did

- Move the `clean` step from `circleci-migrate-contracts.sh` to its own script.
  - NOTE: I thought about using a command line arg with current script to trigger a "clean only" or "regular" migration run.  It ended up looking weird because the "regular" path didn't include the clean step, since it's executed via separate Circle `run` step.  I figure it will just cause more confusion when someone else looks at what's going on here.  Instead we took a bit of duplication to keep what's happening obvious.

- Include new `run` step in the `migrate_contracts` job that has `when: always` set.  This will run the new clean script right after the migration step runs.

### Testing

 - [ ] Clean executes on migration pass
-  [ ] Clean executes on migraation fail

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,6 +109,13 @@ jobs:
             export CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY=$CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY
             mkdir -p /tmp/keep-client/contracts
             solidity/scripts/circleci-migrate-contracts.sh
+      - run:
+          name: Clean Migration Directory
+          command:
+            export BUILD_TAG=$CIRCLE_SHA1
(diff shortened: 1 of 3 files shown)


sthompson22:
I'm going to close this draft because we're disabling migrations against dev, for now.  We'll revisit this when we revisit dev migrations.

PR #1770, opened 2020-05-07
run-random-beacon: Update Ropsten Contracts

sthompson22:
Updating the doc to reference v1.2.0-rc contracts.


sthompson22:
Ready here - doin it

PR #1773, opened 2020-05-08
Return the chanID as a number

r-czajkowski:
The hardware wallet providers require a `chainId` param as a number but the `getChainId` function returned this param as a string. This update addresses it and makes that a `getChainId` function returns a number.


dimpar:
Tested by mimic ropsten/master in `getChainId`. Transaction went through when delegating owned token.

PR #1774, opened 2020-05-08
Fix fetching a deleagtion from a TokenGrant

r-czajkowski:
This PR fixes fetching the delegated stake from a TokenGrant.


dimpar:
Validated delegated KEEP tokens from grant in Operations. Looks good

PR #1782, opened 2020-05-10
keep-prd: Relay request submitter.

sthompson22:
Here's a quick and dirty mainnet relay request submitter to help with seeding the network if needed.  Can't be applied or merged until things needed is satisfied.

#### Things needed before this can be used

- [ ] Mainnet infura account
- [ ] Funded ETH account


sthompson22:
This is looking stale so going to close,  can re-open if we need it.

PR #1789, opened 2020-05-14
TBTC-less bonding: Allow bonding before tBTC is out

Shadowfiend:
This PR comments out a few of the bits and pieces around bond funding to allow it to be deployable despite tBTC contracts not yet being deployed. This will allow bootstrapping of available bonds and keep-ecdsa nodes so that once tBTC goes live, operators will already be available to back deposits.

Additionally, this fixes a few small bugs:
- Non-operators don't see the option to set available bond on their own address.
- Owners with grants properly see their operators in the bonding list.

### DIFF
--- a/solidity/dashboard/src/services/tbtc-authorization.service.js
+++ b/solidity/dashboard/src/services/tbtc-authorization.service.js
@@ -3,6 +3,7 @@ import { TOKEN_STAKING_CONTRACT_NAME } from "../constants/constants"
 import {
   BONDED_ECDSA_KEEP_FACTORY_CONTRACT_NAME,
   KEEP_BONDING_CONTRACT_NAME,
+  MANAGED_GRANT_FACTORY_CONTRACT_NAME
 } from "../constants/constants"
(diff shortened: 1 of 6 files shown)


solidity/dashboard/src/services/tbtc-authorization.service.js
```diff
@@ -312,8 +331,38 @@ const fetchOperatorsOf = async (web3Context, yourAddress) => {
...
+      web3Context,
+      TOKEN_STAKING_CONTRACT_NAME,
+      "ownerOf",
+      yourAddress
+    )
+
+    if (ownerAddress !== "0x0000000000000000000000000000000000000000") {
+      ownerOperators[0] = yourAddress
+    }
   }
```
pdyraga, line 366: Case: _This is not possible_. Cooper: _No. It's necessary._

pdyraga:
I could see it working in all combinations. Let's do eeeet.

pdyraga:
cc @dimpar @r-czajkowski

Issue #1769, opened 2020-05-07
Link to the delegation tx


mhluongo:
![Screenshot from 2020-05-07 18-34-11](https://user-images.githubusercontent.com/427505/81351504-7ae9f780-9092-11ea-8176-b8de8a242fe5.png)

It should open in a new window for everyone's sanity :)

PR #1794, opened 2020-05-14
Link to the tx

r-czajkowski:
Closes: #1769 

This PR adds the link to the transaction on the etherscan and updates styles of the message component.

transaction-is-pending

obraz


dimpar:
Do you think we can keep an address showing for a bit longer?

too-fast2

r-czajkowski:
I think this situation ^ is only possible with local chains like Ganache. In the 'real-world' it will be much longer.

dimpar:
The other option would be adding the etherscan link in the success banner, and give a user control when he wants to close it. This way a user certainly won't miss the timeline for clicking the nav link.

r-czajkowski:
Agree 👍 , but maybe let's open an issue for that. I think it may take a while.

dimpar:
okay, I've created an issue for that: https://github.com/keep-network/keep-core/issues/1813

liz-shinn:
+1 to including an option to close the success notification so it stays on screen longer, as well as adding a link to the success notification.

PR #1795, opened 2020-05-14
Increase the number of address to fetch from hardware wallets

r-czajkowski:
This PR increases the number of addresses to fetch from hardware wallets.


dimpar:
Tested on Ledger Live and Trezor. 15 accounts were displayed in the modal.

Issue #1792, opened 2020-05-14
Support authorizers with multiple operators 


dimpar:
In the bonding section of tBTC, we need to add operators of an authorizer.

Issue #1793, opened 2020-05-14
Handling BondReassigned event


dimpar:
In case `BondReassigned` event occurrence, we need to fetch bonded Eth from `lockedBond[newReassignedBondId]`. The value of the "old" `lockedBond[bondID]` becomes 0

dimpar:
something like:
```
// TODO: add later when fetching bonding section
const fetchBondReassignedEvents = async (
  web3Context,
  operatorAddresses,
  referenceId
) => {
  return contractService.getPastEvents(
    web3Context,
    KEEP_BONDING_CONTRACT_NAME,
    "BondReassigned",
    {
      fromBlock: CONTRACT_DEPLOY_BLOCK_NUMBER[KEEP_BONDING_CONTRACT_NAME],
      filter: { operator: operatorAddresses, referenceID: referenceId },
    }
  )
}
```

lionakhnazarov:
Closing as stale. Targets the legacy keep-core v1 (random beacon / KEEP staking) architecture, superseded by the Threshold v2 stack. Not actionable against current main.

PR #1799, opened 2020-05-15
Adding operators of authorizer in bonding section

dimpar:
Closes: https://github.com/keep-network/keep-core/issues/1792

In this PR we add operators of an authorizer in the bonding section of tBTC application.

### DIFF
--- a/solidity/dashboard/src/services/tbtc-authorization.service.js
+++ b/solidity/dashboard/src/services/tbtc-authorization.service.js
@@ -1,9 +1,12 @@
 import { contractService } from "./contracts.service"
-import { TOKEN_STAKING_CONTRACT_NAME, TOKEN_GRANT_CONTRACT_NAME } from "../constants/constants"
+import {
+  TOKEN_STAKING_CONTRACT_NAME,
+  TOKEN_GRANT_CONTRACT_NAME,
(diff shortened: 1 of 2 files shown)


solidity/dashboard/src/services/tbtc-authorization.service.js
```diff
@@ -270,25 +256,34 @@ const fetchDelegationInfo = async (web3Context, operatorAddress) => {
...
+      operator: _.returnValues.operator,
+      holder: _.returnValues.holder,
+      referenceID: _.returnValues.referenceID,
     }
-  )
+  })
 }
 
+// TODO: add later when fetching bonding section
 const fetchBondReassignedEvents = async (
```
r-czajkowski, line 287: Let's comment out this function due to `fetchBondReassignedEvents is assigned a value but never used` warning.

dimpar: Actually, I'd rather just remove it for now. Copied & paste in the [issue](https://github.com/keep-network/keep-core/issues/1793) as an example what should be done later.

dimpar:
@r-czajkowski I don't think anyone would review it other than you, so since you're the only reviewer, you can go ahead and merge :)

PR #1811, opened 2020-05-19
Use alpine as the runtime container since Go binaries are self-contained

alphastorm:
This cuts the final Docker image size down by 90% from 391MB (v1.2.0) to 36.9MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,4 @@
-FROM golang:1.13.6-alpine3.10 AS runtime
-
-ENV APP_NAME=keep-client \
-	BIN_PATH=/usr/local/bin
-
-FROM runtime AS gobuild
+FROM golang:1.13.6-alpine3.10 AS gobuild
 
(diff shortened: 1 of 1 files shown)


alphastorm:
Before:
Screen Shot 2020-05-19 at 9 43 14 AM
After:
Screen Shot 2020-05-19 at 9 43 32 AM
If you want to go even slimmer can use [GoogleContainerTools/distroless](https://github.com/GoogleContainerTools/distroless), but I've never used it before and "Distroless images are minimal and lack shell access", so you need to use special debug images to troubleshoot.

mhluongo:
Very excited about this- we lost Alpine earlier in the dev process, and have been wanting it back for some time. Thanks @alphastorm, we'll get this reviewed!

liamzebedee:
Nice one, @alphastorm! Save those 🌲

lukasz-zimnoch:
Works as advertised!

Issue #1812, opened 2020-05-19
Adding withdraw button to the bonding section


dimpar:
For this first implementation, it can be a “withdraw all” that goes straight to a Metamask popup and sets the amount as the full unbonded amount. 

The “full” implementation here would be similar to “Add”, in that it lets you specify how much you want to withdraw. We should also notify users that they are withdrawing to the beneficiary address.

liz-shinn:
Working on this in the Figma file here: https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=2172%3A321

liz-shinn:
![add-eth](https://user-images.githubusercontent.com/57226633/82357550-4996f300-99d3-11ea-9638-2f7338805c4f.gif)

This GIF shows adding ETH. Note that the 'Withdraw` button state changes from inactive to active once there is a balance of ETH added.

liz-shinn:
![Withdraw ETH](https://user-images.githubusercontent.com/57226633/82360011-b9f34380-99d6-11ea-87d8-59aabccbeb74.png)


When you click `Withdraw` it pulls up a window like Add ETH does where you can enter an amount. Prominently notes that withdrawn ETH goes to the beneficiary address.

liz-shinn:
**SHORT TERM**
We'll skip the modal for withdrawal and just add a button that reads "Withdraw All" 

Also in the short-term we can bold and change the color of the instructions where it indicates ETH withdrawal goes to the beneficiary address.

![Image 2020-05-19 at 2 12 49 PM](https://user-images.githubusercontent.com/57226633/82362963-2bcd8c00-99db-11ea-9778-f8bb6687ebcf.png)

PR #1814, opened 2020-05-20
Withdraw all bonding

dimpar:
Refs: https://github.com/keep-network/keep-core/issues/1812

Adding functionality to withdraw all available bonds (aka unbondedValue) in the bonding section.

### DIFF
--- a/solidity/dashboard/src/services/tbtc-authorization.service.js
+++ b/solidity/dashboard/src/services/tbtc-authorization.service.js
@@ -212,9 +212,7 @@ const fetchBondingData = async (web3Context) => {
         operators[i][0]
       )
 
-      const bondedEth = operatorBondingDataMap.get(
-        operators[i][0]
-      )
+      const bondedEth = operatorBondingDataMap.get(operators[i][0])
         ? operatorBondingDataMap.get(operators[i][0])
         : 0
 
@@ -344,18 +342,11 @@ const fetchOperatorsOf = async (web3Context, yourAddress) => {
       "getGranteeOperators",
       managedGrantAddress
     )
-    for (let i = 0; i < managedGrantAddresses.length; ++i) {
-      const managedGrantAddress = managedGrantAddresses[i]
-      // operators of grantee (managedGrantAddress)
-      let operatorsOfManagedGrant = await contractService.makeCall(
-        web3Context,
-        TOKEN_GRANT_CONTRACT_NAME,
-        "getGranteeOperators",
-        managedGrantAddress
-        )
-        for (let i = 0; i < operatorsOfManagedGrant.length; i++) {
-          operators.set(web3Utils.toChecksumAddress(operatorsOfManagedGrant[i]), false)
-        }
+    for (let i = 0; i < operatorsOfManagedGrant.length; i++) {
+      operators.set(
+        web3Utils.toChecksumAddress(operatorsOfManagedGrant[i]),
+        false
+      )
     }
   }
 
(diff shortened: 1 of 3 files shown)


dimpar:
Tested by:
- withdrawing funds as an owner of operator. Eth were transferred to the beneficiary address.
- withdrawing funds as an operator. Eth were transferred to the beneficiary address.

r-czajkowski:
Tested:
* ^ paths above,
* withdrawing founds as an operator of the stake created by [ManagedGrant](https://github.com/keep-network/keep-core/blob/master/solidity/contracts/ManagedGrant.sol#L135). Eth were transferred to the beneficiary address.

solidity/dashboard/src/services/tbtc-authorization.service.js
```diff
@@ -328,35 +336,50 @@ const fetchOperatorsOf = async (web3Context, yourAddress) => {
...
+    yourAddress
+  )
+
+  if (ownerAddress !== "0x0000000000000000000000000000000000000000") {
+    // yourAddress is an operator
+    operators.set(web3Utils.toChecksumAddress(yourAddress), true)
   }
 
-  return [...new Set(operators)]
+  return [...operators]
```
r-czajkowski, line 382: Is there any reason why we return an array instead of Map?

solidity/dashboard/src/services/tbtc-authorization.service.js
```diff
@@ -168,7 +183,7 @@ const fetchBondingData = async (web3Context) => {
     const sortitionPoolAddress = await fetchSortitionPoolForTbtc(web3Context)
     const createdBonds = await fetchCreatedBonds(
       web3Context,
-      operators,
+      operators.map((_) => _[0]),
```
r-czajkowski, line 186: If we return a Map from `fetchOperatorsOf`, we could `Array.from(operators.keys())`.

solidity/dashboard/src/services/tbtc-authorization.service.js
```diff
@@ -190,18 +205,20 @@ const fetchBondingData = async (web3Context) => {
     for (let i = 0; i < operators.length; i++) {
```
r-czajkowski, line 205: Also, with operators as a Map, iteration through operators could be cleaner. We could: `for (const [operatorAddress, isWithdrawable] of operators.entries())`. By using Map, we can avoid unclear variables like `operators[i][0]`.

dimpar: yea, I like that!

solidity/dashboard/src/services/tbtc-authorization.service.js
dimpar: I guess it made sense while was drafting things. No reason why we should return an array. Will return a map to make further processing cleaner.

solidity/dashboard/src/services/tbtc-authorization.service.js
```diff
@@ -202,23 +202,26 @@ const fetchBondingData = async (web3Context) => {
...
-      const bondedEth = operatorBondingDataMap.get(operators[i][0])
-        ? operatorBondingDataMap.get(operators[i][0])
+      const bondedEth = operatorBondingDataMap.get(operatorAddress)
+        ? operatorBondingDataMap.get(operatorAddress)
         : 0
 
       const bonding = {
-        operatorAddress: operators[i][0],
-        isWithdrawable: operators[i][1],
+        operatorAddress: operatorAddress,
```
r-czajkowski, line 223: I think we can just `operatorAddress,` instead of `operatorAddress: operatorAddress`

dimpar:
@r-czajkowski your changes look good to me. I've addressed your comments above and if you accept them, let's merge this guy!

solidity/dashboard/src/services/tbtc-authorization.service.js
dimpar: nice, simplified!

Issue #1386, opened 2020-02-26
Bonded ECDSA keep rewards


pdyraga:
- Beacon rewards are paid in ETH. The beneficiary has to withdraw the reward. They are accumulating per group and can be withdrawn when the group expires. 
- Bonded ECDSA keep rewards are paid in ERC20 specific to the application. They are automatically paid to the beneficiary address and they don't require manual withdrawal.
- Bonded ECDSA keep refunds are paid in ETH. Operator nodes are automatically withdrawing them to beneficiary account but we may also skip this step and require beneficiaries to withdraw them manually. 

*Rewards* page in the dApp supports only beacon rewards right now. We need to discuss how to present and handle Bonded ECDSA keep rewards there.

PR #1816, opened 2020-05-21
TBTC rewards page in KEEP token dashboard

r-czajkowski:
Closes: #1386 

This PR adds the TBTC rewards page according to the Figma view. 

obraz


solidity/dashboard/src/services/tbtc-rewards.service.js
```diff
@@ -0,0 +1,72 @@
...
+  const bondedECDSAKeepContract = createBondedECDSAKeepContractInstance(
+    web3,
+    keepAddress
+  )
+
+  const bondedMembers = new Set(
+    await bondedECDSAKeepContract.methods.getMembers().call()
+  )
+
+  const benficiaryOperators = []
```
dimpar, line 54: benficiaryOperators -> beneficiaryOperators

dimpar:
Screenshot 2020-06-29 at 10 40 55

Looks like we display 0 rewards upon deposit. Deposit Token ID and operators are 0x0

dimpar:
Tested by running keep Node, 3 ecdsa nodes and tBTC dApp. Then going through the deposit -> redeem flows a couple of times. All seems to work fine.
Screenshot 2020-06-30 at 12 28 32

PR #1818, opened 2020-05-21
Improved modsqrt documentation

pdyraga:
Tonelli-shanks algorithm only works when p is a prime, and the Legendre symbol is only defined for odd primes. Since both functions are internal, we should be fine with just making the expectations about p clear in the documentation.

Also, removed incorrect `p == 2` condition in `modSqrt` implementation. This value of `p` is not a valid input.

### DIFF
--- a/pkg/altbn128/altbn128.go
+++ b/pkg/altbn128/altbn128.go
@@ -60,8 +60,11 @@ func mod(i, m *big.Int) *big.Int {
 	return new(big.Int).Mod(i, m)
 }
 
-func modSqrt(i, m *big.Int) *big.Int {
-	return new(big.Int).ModSqrt(i, m)
+// modSqrt returns square root of x mod p if such a square root exists. The
+// modulus p must be an odd prime. If x is not a square mod p, function returns
+// nil.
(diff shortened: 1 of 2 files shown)


pkg/altbn128/altbn128.go
```diff
@@ -60,8 +60,11 @@ func mod(i, m *big.Int) *big.Int {
 	return new(big.Int).Mod(i, m)
 }
 
-func modSqrt(i, m *big.Int) *big.Int {
-	return new(big.Int).ModSqrt(i, m)
+// modSqrt returns square root of x mod p if such a square root exists. The
+// modulus p must be an odd prime. If x is not a square mod p, function returns
+// nil.
```
dimpar, line 65: do we want to add that it panics if `p` is not odd?

pdyraga: I am not sure, we disallow this value. I think we don't have to define what happens if a disallowed value is provided in this case. What do you think?

dimpar: ah, that makes sense.

PR #1819, opened 2020-05-21
Align master to mainnet release

sthompson22:
After some discussion about how to handle versions in master while we're not continuously migrating @pdyraga made the suggestion that we just land on the most recent mainnet release versions for now.

Here we update keep-core to reflect those versions.  Notably absent here is an explicit version for the keep-client.  It's on the list.

I consider this a temporary setting, while we have deeper discussion about improvements to release management.

### DIFF
--- a/infrastructure/kube/keep-prd/keep-dapp-token-dashboard-deployment.yaml
+++ b/infrastructure/kube/keep-prd/keep-dapp-token-dashboard-deployment.yaml
@@ -21,6 +21,6 @@ spec:
     spec:
       containers:
       - name: keep-dapp-token-dashboard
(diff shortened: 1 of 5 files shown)


sthompson22:
Will close when the release is out.

sthompson22:
This is ready for another pass whenever someone has a breather.

sthompson22:
Do we want to revisit this?

pdyraga:
I think we want to close this one without merging after having the versions pointing to RC after the recent release. What do you think @sthompson22 ?

Shadowfiend:
I agree, let's do it.

Issue #1719, opened 2020-04-30
Disable "Undelegate" button when having 0 Keep


dimpar:
![Screenshot 2020-04-30 at 15 50 37](https://user-images.githubusercontent.com/3156420/80718692-c3397080-8afa-11ea-93a9-25e7608b4b84.png)

Issue #1632, opened 2020-04-20
Operator page UI updates


r-czajkowski:
Let's update the operator page according to [Figma](https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=2540%3A1637) views.

![obraz](https://user-images.githubusercontent.com/57687279/79743473-eb5bef00-8304-11ea-9497-ddf5b66c471f.png)
![obraz](https://user-images.githubusercontent.com/57687279/79743493-f6af1a80-8304-11ea-8c28-39b0c064a401.png)

dimpar:
Also, I think we should hide the text "Click undelegate below..." if all tokens have been already undelegated.
![Screenshot 2020-05-05 at 13 54 15](https://user-images.githubusercontent.com/3156420/81063426-0b300d00-8ed8-11ea-9222-351c592ec5c2.png)

PR #1820, opened 2020-05-21
Operator page ui updates

r-czajkowski:
Closes: #1632
Closes: #1719

This PR updates the operator page according to the Figma view.

### DIFF
--- a/solidity/dashboard/src/components/SlashedTokensList.jsx
+++ b/solidity/dashboard/src/components/SlashedTokensList.jsx
@@ -2,50 +2,52 @@ import React from "react"
 import { formatDate } from "../utils/general.utils"
 import { displayAmount } from "../utils/token.utils"
 import AddressShortcut from "./AddressShortcut"
+import { DataTable, Column } from "./DataTable"
 
 const SlashedTokensList = ({ slashedTokens }) => {
   return (
-    <div className="slashed-tokens-list">
-      <div className="flex flex-1">
-        <span className="text-label flex flex-1">amount</span>
-        <span className="text-label flex-2">details</span>
-      </div>
-      <ul className="flex column flex-1">
-        {slashedTokens.map(renderSlashedTokensItem)}
-      </ul>
-    </div>
+    <DataTable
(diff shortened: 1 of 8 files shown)


solidity/dashboard/src/services/token-staking.service.js
```diff
@@ -118,31 +121,31 @@ const fetchPendingUndelegation = async (web3Context) => {
...
-      ? COMPLETE_STATUS
-      : PENDING_STATUS
-  } else if (
-    delegation.undelegatedAt === "0" &&
-    delegation.createdAt !== "0" &&
-    delegation.amount === "0"
-  ) {
-    undelegationStatus = COMPLETE_STATUS
+
+  let delegationStatus
```
dimpar, line 132: Don't we need another status here "Pending"? What I noticed is when we undelegate Tokens, status is still "Processing" in the Operations, but in the Tokens it's complete for the same undelegation.
Screenshot 2020-05-25 at 15 14 35
Screenshot 2020-05-25 at 15 13 35

solidity/dashboard/src/components/SlashedTokensList.jsx
```diff
@@ -2,50 +2,52 @@ import React from "react"
...
+          </>
+        )}
+      />
+      <Column
+        header="details"
+        field="event"
+        renderContent={({ event, groupPublicKey, date }) => (
+          <>
+            <div className="text-big text-grey-70">
+              Group&nbsp;
```
dimpar, line 33: Looks like this whitespace is unnecessary after Group.

solidity/dashboard/src/services/token-staking.service.js
r-czajkowski: From the operator's perspective, it is still `processing` because it has not yet been recovered. But I am not sure if this is the correct approach.

dimpar: @Shadowfiend just to double check. Is this what we want (keeping processing status in Operations when it wasn't recovered yet)?

PR #1821, opened 2020-05-22
Fix the positioning of the arrow icon in the link

r-czajkowski:
This PR fixes the positioning of the arrow icon in the link. This problem occurs on the production build only.

Bug:
obraz

Fix:
obraz


dimpar:
Tested
- `npm build`
- `serve -s build`

Arrow was aligned properly

PR #1822, opened 2020-05-22
Fix the bonded eth amount

r-czajkowski:
It is possible that the operator is in multiple bonds (see [KeepBonding contract](https://github.com/keep-network/keep-ecdsa/blob/master/solidity/contracts/KeepBonding.sol#L32)). In that case the `bondedEth` value should be the sum of all locked bonds for a given operator.

### DIFF
--- a/solidity/dashboard/src/services/tbtc-authorization.service.js
+++ b/solidity/dashboard/src/services/tbtc-authorization.service.js
@@ -9,6 +9,7 @@ import {
   MANAGED_GRANT_FACTORY_CONTRACT_NAME,
 } from "../constants/constants"
 import { isSameEthAddress } from "../utils/general.utils"
+import { add } from "../utils/arithmetics.utils"
 import {
   CONTRACT_DEPLOY_BLOCK_NUMBER,
   getBondedECDSAKeepFactoryAddress,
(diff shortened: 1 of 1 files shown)


solidity/dashboard/src/services/tbtc-authorization.service.js
```diff
@@ -189,17 +190,25 @@ const fetchBondingData = async (web3Context) => {
...
+        operatorAddress,
         createdBonds[i].holder,
         createdBonds[i].referenceID
       )
 
-      operatorBondingDataMap.set(
-        web3Utils.toChecksumAddress(createdBonds[i].operator),
-        bondedEth
-      )
+      if (operatorBondingDataMap.has(operatorAddress)) {
```
dimpar, line 203: yea, looks good. I think we can simplified just a bit by checking for a _falsy_ bond value, but up to you:
```
      const currentBond = operatorBondingDataMap.get(operatorAddress)
      if (currentBond) {
        operatorBondingDataMap.set(operatorAddress, add(currentBond, bondedEth))
      } else {
        operatorBondingDataMap.set(operatorAddress, bondedEth)
      }
```

dimpar:
Tested by running 3 ecdsa clients. Opened keep a couple of times with the same operators resulting with new keep addresses. A new keep address becomes an input for `bondId` generation (holder, referenceId). Bonded Eth amount is expected to increase which what I've observed instead of being overwritten.

1. current amount:
Screenshot 2020-05-22 at 14 22 40
2. added bond of `100000` Wei
Screenshot 2020-05-22 at 14 35 05

PR #1839, opened 2020-06-02
Signature issue

dimpar:
Adding more logs around final signature reconstruction from member shares.


pdyraga:
@dimpar @lukasz-zimnoch Do you guys think we still need it after https://github.com/keep-network/keep-core/pull/1881?

lukasz-zimnoch:
> @dimpar @lukasz-zimnoch Do you guys think we still need it after #1881?

Nope, I think we can close.

Issue #1823, opened 2020-05-22
Rewards calculation in the dashboard is incorrect when there are active terminated groups


shaimo:
The 'numberOfGroups' method subtracts terminated groups - https://github.com/keep-network/keep-core/blob/master/solidity/contracts/libraries/operator/Groups.sol#L219. It means that when there are active terminated groups (and surely there will be), the calculation here is wrong: https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/services/rewards.service.js#L22

Issue #1804, opened 2020-05-16
Rewards not Visible on Token Dashboard


navneetlakra:
The rewards are no longer visible on the token dashboard. Used to see the rewards alright till last a few days back.
![image](https://user-images.githubusercontent.com/10936810/82108907-9a111500-9775-11ea-8b50-b2e2c6946dd2.png)

Please let me know if more information is required.

r-czajkowski:
Thanks for reporting! I think it could be related to #1823

Issue #1636, opened 2020-04-20
Rewards page UI updates in KEEP token dashboard


r-czajkowski:
The rewards page is probably ready. We need to confirm and check if everything is ok.

![obraz](https://user-images.githubusercontent.com/57687279/79745572-d719f100-8308-11ea-871a-a909e3160589.png)

r-czajkowski:
@pdyraga, what do we do with the ETH value in $?

liz-shinn:
Updates to this page addressing the concern for withdrawal by row (for the case where a beneficiary has multiple operators per one signing group), as well as including a potential for withdraw all at the top that has a gas cost estimate:

![beaconrewardspage](https://user-images.githubusercontent.com/57226633/84961029-445fbd80-b0d1-11ea-8651-f4848e36af6f.png)

liz-shinn:
Changing the labels of the rewards to more accurately reflect rewards status vs signing group status:
![rewards status](https://user-images.githubusercontent.com/57226633/85145266-539e5280-b21a-11ea-80b1-b728de9a3c4d.png)

Issue #1613, opened 2020-04-16
Format issue displaying reward ETH on dashboard


msimovic:
<img width="814" alt="formatissue" src="https://user-images.githubusercontent.com/1518898/79477786-a10f0100-7fd8-11ea-84e9-d2d3af4568ea.png">

PR #1843, opened 2020-06-05
Random Beacon rewards page

r-czajkowski:
Closes: #1804
Closes: #1823
Closes: #1636
Closes: #1613
Depends on: #1851

This PR updates the random beacon rewards page and fixes the bug with terminated groups described in #1823. These updates don't add an `operator` column in the `rewards status` table due to the beneficiary may have more than one operator in a given group and Figma views don't consider this. I've posted a [comment](https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx?node-id=2544:645#28944905) on Figma.

<details>
  <summary>Screenshoots</summary>

obraz

obraz
</details>


solidity/dashboard/src/services/rewards.service.js
```diff
@@ -13,26 +14,32 @@ const fetchAvailableRewards = async (web3Context) => {
...
+      fromBlock: CONTRACT_DEPLOY_BLOCK_NUMBER[OPERATOR_CONTRACT_NAME],
+    }
+
+    // get all created groups
+    const groups = await contractService.getPastEvents(
+      web3Context,
+      OPERATOR_CONTRACT_NAME,
+      "DkgResultSubmittedEvent",
+      operatorEventsSearchFilters
+    )
```
dimpar, line 27: thoughts about filtering return data a bit here to make a `for` loop below cleaner? Similar to what we do in `tbt-authorization.service.js`
```
.map((_) => _.returnValues. groupPubKey)
```

solidity/dashboard/src/services/rewards.service.js
```diff
@@ -164,18 +164,18 @@ const fetchWithdrawalHistory = async (web3Context) => {
...
             .call()
           return {
             blockNumber,
             groupPublicKey,
-            date: formatDate(withdrawnAt * 1000),
-            amount: utils.fromWei(amount, "ether"),
+            reward: utils.fromWei(amount, "ether"),
             transactionHash,
+            operator,
+            status: "WITHDRAWN",
```
dimpar, line 178: if we extract rewards statuses to `constants` then we'll be able to reuse it here

dimpar:
If we have a terminated group, what status should we display in the `Random Beacon Rewards` page? (see [screenshots](https://github.com/keep-network/keep-core/pull/1843#issue-428378230)). As for now, it will say `Active`, but seems confusing. @Shadowfiend @pdyraga can you please advice?

dimpar:
`see more` / `see less` is not visible after navigating to `/rewards` page. It will appear if you `withdraw` some eth.

solidity/dashboard/src/services/rewards.service.js
r-czajkowski: I think with `map` would be more complicated due to async calls and in some cases, we skip iteration with `continue` which is not possible with `map`.

solidity/dashboard/src/services/rewards.service.js
dimpar: seems that we still do not reference constant

solidity/dashboard/src/services/rewards.service.js
dimpar: It shouldn't affect skipping the iteration. If we create a new array by extracting public keys this way:
```
const groupPubKeys = (
    await contractService.getPastEvents(
      web3Context,
      OPERATOR_CONTRACT_NAME,
      "DkgResultSubmittedEvent",
      operatorEventsSearchFilters
    )
).map((_) => _.returnValues.groupPubKey)
```
then below, we can just write a simple for loop:
```
for (let i = 0; i < groupPubKeys.length; i++) {
 // groupPubKeys[i]
}
```

solidity/dashboard/src/services/rewards.service.js
```diff
@@ -45,29 +52,27 @@ const fetchAvailableRewards = async (web3Context) => {
           .call()
 
         if (gt(awaitingRewards, 0)) {
-          groupMemberIndices[groupPublicKey][memberAddress] = awaitingRewards
+          groupMemberIndices[groupPubKey][memberAddress] = awaitingRewards
         }
       }
-      if (Object.keys(groupMemberIndices[groupPublicKey]).length === 0) {
+      if (Object.keys(groupMemberIndices[groupPubKey]).length === 0) {
```
dimpar, line 58: maybe we can get rid of `continue` by
```
if (Object.keys(groupMemberIndices[groupPubKey]).length !== 0) { // code below continue }
```
?

pdyraga:
> If we have a terminated group, what status should we display in the `Random Beacon Rewards` page? (see [screenshots](https://github.com/keep-network/keep-core/pull/1843#issue-428378230)). As for now, it will say `Active`, but seems confusing. @Shadowfiend @pdyraga can you please advice?

I think we should display `Terminated`.

solidity/dashboard/src/services/rewards.service.js
```diff
@@ -19,21 +19,19 @@ const fetchAvailableRewards = async (web3Context) => {
     }
 
     // get all created groups
```
dimpar, line 21: // get all created public keys ?

liz-shinn:
Updates to this, also[ in the Figma file here](https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=2544%3A645)!

beaconrewardspage

Shadowfiend:
> If we have a terminated group, what status should we display in the Random Beacon Rewards page?

Looked at this more closely; what's not clear to me from these screenshots is what that status is meant to reflect. If the status is the status of the rewards, then the rewards should be “Accumulating”, “Available to withdraw”, or “Withdrawn”, I think? If the status is the status of the group, it would be different. Maybe I'm misunderstanding something though?

liz-shinn:
>Looked at this more closely; what's not clear to me from these screenshots is what that status is meant to reflect.

Initially was meant to tie the group status _to_ the status of the rewards, but yes that gets murky if you add in this element that a group is terminated, not sure how that impacts rewards that you already have accumulated...

It makes sense to limit these pill-shaped statuses for clarity to just refer to the status of your rewards. 

Can represent the group status differently (those statuses are completed, active, and terminated, right?).

liz-shinn:
Updated this @Shadowfiend and @r-czajkowski — thoughts on clarity here? More clear which status is related to rewards and which is related to signing group?

Image 2020-06-18 at 2 26 35 PM

liz-shinn:
For 'termination' – should that be highlighted to the user that that signing group behaved unexpectedly? And would it be considered an error state?

solidity/dashboard/src/components/Rewards.jsx
```diff
@@ -0,0 +1,349 @@
...
+  }
+
+  const updateRewardStatus = (status, groupIndex, operator) => {
+    const { indexInArray } = findIndexAndObject(
+      "groupIndex",
+      groupIndex,
+      rewards,
+      (object) => isSameRewardRecord(object, groupIndex, operator)
+    )
+    if (indexInArray === null) {
```
lukasz-zimnoch, line 165: This doesn't protect against `undefined`. Are we good here?

solidity/dashboard/src/services/rewards.service.js
```diff
@@ -1,153 +1,120 @@
...
+
+    const operatorsOfBeneficiary = await getOperatorsOfBeneficiary(
+      web3Context,
+      yourAddress
+    )
+    const rewards = []
+    // { groupIndex: { isStale, isTerminated, groupPubKey } }
+    const groups = {}
+
+    for (let groupIndex = 0; groupIndex < 2; groupIndex++) {
```
lukasz-zimnoch, line 42: `2` is a magic number. Can we extract and explain it a bit?

solidity/dashboard/src/services/rewards.service.js
```diff
@@ -1,153 +1,120 @@
...
-          continue
-        }
+    ).map((event) => event.returnValues.groupPubKey)
+
+    const operatorsOfBeneficiary = await getOperatorsOfBeneficiary(
+      web3Context,
+      yourAddress
+    )
+    const rewards = []
+    // { groupIndex: { isStale, isTerminated, groupPubKey } }
```
lukasz-zimnoch, line 39: Do we need to leave this comment?

solidity/scripts/rewards.js
```diff
@@ -75,33 +75,36 @@ module.exports = async function() {
...
   
+    // terminate groups
+    await keepRandomBeaconOperator.reportUnauthorizedSigning(25, { from: owner })
+    await keepRandomBeaconOperator.reportUnauthorizedSigning(26, { from: owner })
+
     const numberOfGroups = await keepRandomBeaconOperator.numberOfGroups()
     const firstActiveIndex = await keepRandomBeaconOperator.getFirstActiveGroupIndex()
   
-    for (let i = 0; i < firstActiveIndex; i++) {
+    for (let i = 0; i < 30; i++) {
```
lukasz-zimnoch, line 85: Please extract `30`.

solidity/scripts/rewards.js
```diff
@@ -75,33 +75,36 @@ module.exports = async function() {
...
-    await keepRandomBeaconOperator.emitRewardsWithdrawnEvent(accounts[1], 1)
-    await keepRandomBeaconOperator.emitRewardsWithdrawnEvent(accounts[1], 5)
-    await keepRandomBeaconOperator.emitRewardsWithdrawnEvent(accounts[1], 6)  
-  
+    
+    // await keepRandomBeaconOperator.emitRewardsWithdrawnEvent(accounts[1], 1)
+    // await keepRandomBeaconOperator.emitRewardsWithdrawnEvent(accounts[1], 3)
+    // await keepRandomBeaconOperator.emitRewardsWithdrawnEvent(accounts[1], 1)
+    // await keepRandomBeaconOperator.emitRewardsWithdrawnEvent(accounts[1], 5)
+    // await keepRandomBeaconOperator.emitRewardsWithdrawnEvent(accounts[1], 6)
```
lukasz-zimnoch, line 99: If we don't need those lines, maybe just remove them?

lukasz-zimnoch:
@r-czajkowski can you make a final pass across all comments and merge master?

solidity/dashboard/src/components/Rewards.jsx
```diff
@@ -0,0 +1,349 @@
...
+      returnValues: { groupIndex, operator, amount },
+    } = latestEvent
+    const { indexInArray } = findIndexAndObject(
+      "groupIndex",
+      groupIndex,
+      rewards,
+      (object) => isSameRewardRecord(object, groupIndex, operator)
+    )
+
+    if (indexInArray === null) {
```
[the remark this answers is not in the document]
r-czajkowski: Yep, because the `findIndexAndObject` fn returns null if object was not found.

solidity/dashboard/src/components/Rewards.jsx
r-czajkowski: ^ comment above

Issue #1778, opened 2020-05-08
Make sure button hover states are consistent to Design System figma file


liz-shinn:
There are a few instances of button states that deviate from Figma in the dApp. An example screenshot below:

<img width="442" alt="KEEP_Token_Dashboard" src="https://user-images.githubusercontent.com/57226633/81453449-90752500-9157-11ea-94cb-9a69fa1117b2.png">

Here's how the button states should be:
<img width="867" alt="Keep_Design_System" src="https://user-images.githubusercontent.com/57226633/81453675-36c12a80-9158-11ea-870b-5a73d2adc85e.png">

Button states can be referred to here in the design system in Figma (hoping to get this into a more dev friendly format at some point, but for now it's a component based Figma file)
https://www.figma.com/file/ARAMFLMOPal0JewjvlZGhX/Keep-Design-System?node-id=787%3A1141

liz-shinn:
<img width="454" alt="KEEP_Token_Dashboard" src="https://user-images.githubusercontent.com/57226633/81454259-b996b500-9159-11ea-8822-0b0ab6a3f5b0.png">

liz-shinn:
These two buttons have inconsistent padding – should be either stacked on top of each other or enough space in between them to be able to tell there are two distinct choices here. I would recommend stacking them.

<img width="476" alt="KEEP_Token_Dashboard" src="https://user-images.githubusercontent.com/57226633/81454313-dfbc5500-9159-11ea-9515-f9a7679bc805.png">

PR #1847, opened 2020-06-10
Token overview page

r-czajkowski:
Closes: #1778 

This PR adds the `Overview` page to KEEP token dashboard according to the Figma views. What is more, it fixes button hover states described in #1778.

obraz


solidity/dashboard/src/utils/arithmetics.utils.js
```diff
@@ -25,3 +30,12 @@ export const lt = (a, b) => {
...
   return web3Utils.toBN(a).lte(web3Utils.toBN(b))
 }
+
+export const isZero = (a) => web3Utils.toBN(a).isZero()
+
+export const percentageOf = (value, total) => {
+  return web3Utils
+    .toBN(value)
+    .mul(ONE_HUNDRED)
+    .div(isZero(total) ? ONE : total)
```
dimpar, line 40: if `total` is zero, then the percentageOf will return `value * 100` which is not the correct result, right?
Maybe we shouldn't calculate `percentOf` in case of wrong values being passed to the function and return an error instead?

dimpar:
1. clicking on "Owned Tokens" won't open "Owned" tab in Delegate Tokens.
2. in Figma we have Keep Balance, not Total Balance
3. In Figma, we also have Eth Balance, but it's not visible in Overview

dimpar:
Delegation fails, because of the changes in contracts

r-czajkowski:
>     1. clicking on "Owned Tokens" won't open "Owned" tab in Delegate Tokens.
> 
>     2. in Figma we have Keep Balance, not Total Balance
> 
>     3. In Figma, we also have Eth Balance, but it's not visible in Overview
> 
>     4. Are we adding a "progress bar" next to % Staked?

1. Right! I'll update.
2. 3. 4. The overview page has changed in Figma since the PR opened. Let's update it in a separate PR.

Issue #1783, opened 2020-05-11
Random beacon staker KEEP rewards


pdyraga:
Rewards to distribute KEEP tokens to group members based on a predetermined curve separated by intervals.

pdyraga:
The last nitpick we should make before deploying the contract (being addressed by https://github.com/keep-network/keep-core/pull/2117): 
- adjust allocations in tests so that May release beacon subsidy is properly included,
- make sure those tests do not start failing in the future (we can pass arbitrary group creation timestamps).

PR #1851, opened 2020-06-10
Prepare operator contract for rewards

eth-r:
Expose publicly the total number of groups created by the operator contract, and the registration blockheight and termination status of a group.

To enable this, add a function in the Groups library to query the registration blockheight. To prevent out of gas issues on deployment, change `setGroupMembers` to public instead of internal.

Refs #1783


pdyraga:
I'd normally merge it as-is but we are having other fixes in the queue and I am not sure how they will play together given the operator contract size problems we are having.

Let's have it marinate for a while once we'll finalize all the other fixes we need to implement in the next couple of days and then try to see if we are not running into contract size problems again having them all integrated together.

solidity/contracts/libraries/operator/Groups.sol
```diff
@@ -85,7 +85,7 @@ library Groups {
         bytes memory groupPubKey,
         address[] memory members,
         bytes memory misbehaved
-    ) internal {
+    ) public {
```
eth-r, line 88: From #1863:

> From the peanut gallery, this change looks odd without some sort of ACL. Anyway, flagging it for proper review

eth-r: The rationale is that this is one of the larger internal functions in `Groups`, and is only called once per group so the gas cost impact of `DELEGATECALL` instead of inlining should not be excessive. Because `Groups` is a library, it either gets all applicable ACL from the contract calling it or I'm thoroughly misunderstanding something.

Shadowfiend: > Because Groups is a library, it either gets all applicable ACL from the contract calling it or I'm thoroughly misunderstanding something.

Yep, we deal with deposits this way in tBTC as well.

pdyraga:
I am going to review/merge https://github.com/keep-network/keep-core/pull/1878 first, and then look at this branch if we get any out-of-gas failures. If not, we should be good to merge addressing the changes requested above.

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -696,6 +696,18 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
...
+    function getNumberOfCreatedGroups() public view returns (uint256) {
+        return groups.groups.length;
+    }
+
+    function getGroupRegistrationBlockHeight(uint256 groupIndex) public view returns (uint256) {
+        return groups.getGroupRegistrationBlockHeight(groupIndex);
+    }
+
+    function isGroupTerminated(uint256 groupIndex) public view returns (bool) {
+        return groups.isGroupTerminated(groupIndex);
```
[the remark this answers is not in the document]
pdyraga: Done in https://github.com/keep-network/keep-core/pull/1851/commits/dfe09c002883631f090a8139d51adf85c93cb020.

pdyraga:
Tests are passing, I deployed locally to make sure we are not getting out-of-gas problems, everything looks good.

pdyraga:
@lukasz-zimnoch @dimpar I was doing changes to this PR, can you please guys review?

PR #1852, opened 2020-06-11
Fix golangci-lint test(gofmt)

wakiyamap:
```
$golangci-lint --version
golangci-lint has version 1.27.0 built from fb74c2e on 2020-05-13T18:48:26Z

$golangci-lint run --deadline=10m --disable-all --enable=gofmt
```

I fixed so that warning does not appear.

I'm doing a separate commit now.
https://github.com/wakiyamap/keep-core/commits/Fix_lint


wakiyamap:
I didn't see it needed a signature.
I will fix it if I feel like it.

https://github.com/keep-network/keep-core/blob/master/CONTRIBUTING.adoc

Issue #1854, opened 2020-06-11
Compilation breaks if `abigen` is too new


ElderOrb:
This is similar issue as https://github.com/keep-network/keep-ecdsa/issues/439

go generate ./... produces the following using the latest abigen: 

```
pkg/chain/gen/abi/KeepRandomBeaconService.go:397:2: not enough arguments to return
pkg/chain/gen/abi/KeepRandomBeaconService.go:397:42: _KeepRandomBeaconService.contract.RawTransact undefined (type *bind.BoundContract has no field or method RawTransact)
```

ElderOrb:
finally this 'git commit --amend --no-edit -S' did the trick

PR #1855, opened 2020-06-11
Use package-specific version of abigen

ElderOrb:
Fixes: #1854

Running `go run github.com/ethereum/go-ethereum/cmd/abigen` should run the package-specific version of abigen and we can avoid problems with incompatible `abigen` versions this way.

My initial fix was based on extracting right version of abigen into keep-common, but after re-checking  on a clean environment seems like much simpler approach works too. Not sure what was wrong previously...

### DIFF
--- a/pkg/chain/gen/Makefile
+++ b/pkg/chain/gen/Makefile
@@ -39,7 +39,7 @@ abi/%.abi: ${solidity_dir}/contracts/%.sol
 		 -o abi $<
 
 abi/%.go: abi/%.abi
-	abigen --abi $< --pkg abi --type $* --out $@
+	go run github.com/ethereum/go-ethereum/cmd/abigen --abi $< --pkg abi --type $* --out $@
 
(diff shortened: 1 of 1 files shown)


pdyraga:
@ElderOrb Thanks for the contribution! I can't merge this PR until all your commits are signed, please see https://github.com/keep-network/keep-core/blob/master/CONTRIBUTING.adoc#development-tooling

I'd recommend you to set up commit signing and rebase your branch.

ElderOrb:
I've done 'git commit -S --amend' but still don't see 'verified' label, hmmm

mhluongo:
You also have to tell Github about the keypair for it to say `verified` -> I believe it's in your account settings

ElderOrb:
> You also have to tell Github about the keypair for it to say `verified` -> I believe it's in your account settings

I believe that was done. 

image

ElderOrb:
maybe amend & push is not enough and I need to recreate commit?

pdyraga:
Thank you @ElderOrb 🙌

Issue #1752, opened 2020-05-05
Support viewing grants by ID without a paired wallet


mhluongo:
Eg, dashboard.keep.network/grant/12345

Allowing anyone to view a grant will make management easier for custodians, as well as transparency with the greater community.

Note, it's still fine to lean on having a Web3 provider- we just don't want to require looking up whether someone has a particular grant to view a particular grant.

liz-shinn:
What happens if you click on other parts of the token dash in the menu (are we showing the lefthand menu as dapp context)? If we showed the menu it would then redirect you to the paired wallet requirement screen.

View from Figma:
![Image 2020-05-18 at 11 52 58 AM](https://user-images.githubusercontent.com/57226633/82233572-39611400-98fe-11ea-93cd-c8fad6a3b6e6.png)

PR #1861, opened 2020-06-16
Support viewing grants by ID without a paired wallet

r-czajkowski:
Closes: #1752

This PR adds support for viewing grants by id without a paired wallet. A new page is available at `grant/<grant id>`

obraz


solidity/dashboard/src/services/token-grants.service.js
```diff
@@ -218,10 +219,28 @@ const getOperatorsFromManagedGrants = async (web3Context) => {
   return operators
 }
 
+const fetchGrantById = async (web3Context, grantId) => {
+  const id = new BigNumber(grantId)
```
dimpar, line 223: Can we validate `grantId` in the if statement? Do we actually need to convert it to BigNumber since later we convert it to string?
`return await getGrantDetails(id.toString(), web3Context, true)`

solidity/dashboard/src/services/token-grants.service.js
```diff
@@ -218,10 +219,28 @@ const getOperatorsFromManagedGrants = async (web3Context) => {
...
+
+  if (!id.isInteger() || id.isNegative()) {
+    throw new Error({
+      type: "INVALID_GRANT_ID",
+      message: "Grant ID is not a number",
```
dimpar, line 228: Maybe just "Invalid grant ID"? Negative number is still a number

dimpar:
Do we display error messages anywhere if the `grantId` is invalid or not found?

lukasz-zimnoch:
@r-czajkowski the code looks good to me. Can you address pending comments and merge master?

solidity/dashboard/src/services/token-grants.service.js
r-czajkowski: With `BigNumber` because we can easily check if the provided grntId is an integer or negative value. In that case, a `grantId` could be of type `string/number/BigNumber.

r-czajkowski: If a `grantId` will be `asd`, `BigNumber.isInteger()` returns `false`.

PR #1863, opened 2020-06-16
Rewards

eth-r:
Contracts for distributing KEEP token rewards to keeps.

When a reward contract is created, the creator defines a reward schedule
consisting of one or more reward intervals and their interval weights,
the length of reward intervals, and the quota of how many keeps must be
created in an interval for the full reward for that interval to be paid out.

The amount of KEEP to be distributed is determined by funding the contract,
and additional KEEP can be added at any time.
The reward contract is funded with `approveAndCall` with no extra data,
but it also collects any KEEP mistakenly sent to it in any other way.

An interval is defined by the timestamps [startOf, endOf);
a keep created at the time `startOf(i)` belongs to interval `i`
and one created at `endOf(i)` belongs to `i+1`.

When an interval is over, it will be allocated a percentage of the remaining
unallocated rewards based on its weight, and adjusted by the number of keeps
created in the interval if the quota is not met.

The adjustment for not meeting the keep quota is a percentage that equals
the percentage of the quota that was met; if the number of keeps created is
80% of the quota then 80% of the base reward will be allocated for the
interval.

Any unallocated rewards will stay in the unallocated rewards pool,
to be allocated for future intervals. Intervals past the initially defined
schedule have a weight of 100%, meaning that all remaining unallocated
rewards will be allocated to the interval.

Keeps of the appropriate type can receive rewards once the interval they
were created in is over, and the keep has closed happily.
There is no time limit to receiving rewards, nor is there need to wait for
all keeps from the interval to close.
Calling `receiveReward` automatically allocates the rewards for the interval
the specified keep was created in and all previous intervals.

If a keep is terminated, that fact can be reported to the reward contract.
Reporting a terminated keep returns its allocated reward to the pool of
unallocated rewards.


solidity/contracts/libraries/operator/Groups.sol
```diff
@@ -85,7 +85,7 @@ library Groups {
         bytes memory groupPubKey,
         address[] memory members,
         bytes memory misbehaved
-    ) internal {
+    ) public {
```
mhluongo, line 88: From the peanut gallery, this change looks odd without some sort of ACL. Anyway, flagging it for proper review

eth-r: That change is inherited from #1851 and should be reviewed there; I wish there was a better way to chain PRs than to make the latter PRs merge into the earlier ones.

mhluongo: Making that PR the base branch for this one would help a bit

mhluongo: Probably a faux pas but I flipped it out. Much easier diff IMO

solidity/contracts/BeaconRewards.sol
```diff
@@ -0,0 +1,84 @@
...
+        _minimumKeepsPerInterval,
+        _firstIntervalStart,
+        _intervalWeights
+    ) {
+        operatorContract = KeepRandomBeaconOperator(_operatorContract);
+        tokenStaking = TokenStaking(_stakingContract);
+    }
+
+    function _getKeepCount() internal view returns (uint256) {
+        return operatorContract.getNumberOfCreatedGroups();
```
pdyraga, line 31: How are we going to deal with the existing operator contract deployed where this function is not available?

eth-r: I made a couple of mistakes in getting the operator contract rewards-compatible and didn't realize it until after #1851 was merged. It's possible to hunt for the group number by binary searching the highest registered group index, but the existing operator contract is also missing the registration timestamp storage, meaning that to properly support it the entire rewards contract would need to be adapted to use either timestamps or blockheights depending on configuration. It can be done, but it would be all around easier if we could deploy the operator contract with these changes.

solidity/test/TestBeaconRewards.js
```diff
@@ -0,0 +1,161 @@
...
+    before(async () => {
+        let contracts = await initContracts(
+            contract.fromArtifact('TokenStaking'),
+            contract.fromArtifact('KeepRandomBeaconService'),
+            contract.fromArtifact('KeepRandomBeaconServiceImplV1'),
+            contract.fromArtifact('KeepRandomBeaconOperatorBeaconRewardsStub')
+        )
+        const termLength = 10
+        const totalRewards = 6000
+        const minimumIntervalKeeps = 2
```
NicholasDotSol, line 37: would love to see a test playing with 0 <= minimumIntervalKeeps < 2

solidity/contracts/BeaconRewards.sol
```diff
@@ -0,0 +1,84 @@
...
+    function _distributeReward(bytes32 groupIndexBytes, uint256 _value) internal {
+        bytes memory groupPubkey = operatorContract.getGroupPublicKey(
+            uint256(groupIndexBytes)
+        );
+        address[] memory members = operatorContract.getGroupMembers(groupPubkey);
+
+        uint256 memberCount = members.length;
+        uint256 dividend = _value.div(memberCount);
+
+        require(dividend > 0, "Dividend value must be non-zero");
```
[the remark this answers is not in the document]
eth-r: `allocateRewards` only calculates the allocations of each preceding interval; it doesn't distribute anything to the operators, so `_distributeReward` isn't called on any other keep/group than the specific one targeted and this would only impact intervals where the per-keep reward is less than the group size. However, permitting the distribution of 0 dividend rewards with a nonzero remainder still makes sense.

solidity/contracts/Rewards.sol
```diff
@@ -0,0 +1,518 @@
...
+        uint256 beforeBalance = _unpaidRewards();
+        require(
+            currentBalance >= beforeBalance,
+            "Reward contract has lost tokens"
+        );
+
+        uint256 addedBalance = currentBalance.sub(beforeBalance);
+
+        totalRewards += addedBalance;
+        unallocatedRewards += addedBalance;
```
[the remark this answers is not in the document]
eth-r: Good call. I think there are justified cases for using unsafe math in tightly optimized loops where correctness is reliably maintained otherwise, but this is definitely not one of them.

solidity/contracts/BeaconRewards.sol
pdyraga: Yeah, I remember the previous proposition used the binary search and that's why I was confused. Long story short, are we good with the operator contract after the changes from #1851, or do we need some additional tweaks there?

If making the rewards contract support the current mainnet version of the operator contract and bonded ecdsa keep factory is painful, we can maybe agree on distributing KEEP rewards to existing stakers manually, with a separate contract. @mhluongo what do you think about it?

mhluongo: > we can maybe agree on distributing KEEP rewards to existing stakers semi-manually, with a separate contract. @mhluongo what do you think about it?

If semi-manually means we take custody, nope nope nope. But if you mean we are the ones okaying distributions... maybe? Definitely want to avoid it if possible.

pdyraga: No, I don't want us to take custody. I was rather thinking about having a clone of this contract adjusted to the existing operator contracts with some of the functionality bypassed. For example, we can hardcode the number of groups created and their public keys.

I don't know what's the cost of making such a separate contract versus the cost of adjusting this contract to work with the currently deployed versions. For sure, the rewards contract as-is does not support them.

solidity/contracts/Rewards.sol
```diff
@@ -0,0 +1,518 @@
...
+        internal
+    {
+        uint256 allocatedIntervals = intervalAllocations.length;
+        require(
+            !(interval < allocatedIntervals),
+            "Interval already allocated"
+        );
+        // Allocate previous intervals first
+        if (interval > allocatedIntervals) {
+            _allocateRewards(interval - 1);
```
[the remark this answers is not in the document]
eth-r: I updated this to permit manual allocation of intervals, so we can get unstuck if needed.

solidity/test/TestBeaconRewards.js
eth-r: With the new allocation logic (using `Math.max(minimumIntervalKeeps, keepCount)`) the only situation that's really different is when both are zero, which used to be a divide by zero revert from SafeMath but now allocates 0 correctly.

solidity/contracts/BeaconRewards.sol
pdyraga: We have `BeaconBackportRewards` now. Resolving this conversation.

solidity/contracts/Rewards.sol
```diff
@@ -0,0 +1,611 @@
...
+        uint256 addedBalance = currentBalance.sub(beforeBalance);
+
+        totalRewards = totalRewards.add(addedBalance);
+        unallocatedRewards = unallocatedRewards.add(addedBalance);
+    }
+
+    /// @notice Sends the reward for a keep to the keep members.
+    /// @param keepIdentifier A unique identifier for the keep,
+    /// e.g. address or number converted to a `bytes32`.
+    function receiveReward(bytes32 keepIdentifier)
```
pdyraga, line 156: Do we want to emit an event when revard is claimed? It should be cheap but can help us with some stats/debug later.

solidity/contracts/BeaconBackportRewards.sol
```diff
@@ -0,0 +1,141 @@
+pragma solidity ^0.5.17;
```
pdyraga, line 1: Do we want to add KEEP ASCII art to the new contracts?

solidity/contracts/Rewards.sol
```diff
@@ -0,0 +1,611 @@
...
+        if (__baseAllocation == 0) {
+            return 0;
+        }
+        uint256 keepCount = keepsInInterval(interval);
+        uint256 minimumKeeps = minimumKeepsPerInterval;
+        uint256 adjustmentCount = Math.max(keepCount, minimumKeeps);
+        if (adjustmentCount == 0) {
+            return 0;
+        }
+        return __baseAllocation.div(adjustmentCount).mul(keepCount);
```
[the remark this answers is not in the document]
eth-r: > Now, if it happens that we don't have enough keeps opened in the first interval, we'll not distribute ~17000000 between signers in that interval, but we'll use e.g. just 25% of this amount and distribute the remaining 75% between the rest of intervals proportionally by their weights. The same happens for all other intervals. So if that happens, for example, that we'll still not have enough keeps at the second interval, we'll not only distribute between the future intervals the remaining 75% of tokens from the first interval but also e.g. 75% of tokens from the second interval. The curve's shape is changing by "squeezing" to the right every time we don't have enough keeps in the interval.

Yes, this is how it's working

> Another interesting use case is when we don't use all the tokens available in the rewards contract for the first 12 months (let's say that it's how many intervals we defined when deploying the contract). In that case, we'll distribute all the remaining tokens in 13th month and the shape of the curve will look similar to:

The way each interval allocates a percentage of the remaining reward means that the last months of the schedule should be set to distribute the entire remainder, with month 12 defined as weight 100 (or left implicit with a weight of 100), month 11 with a weight of e.g. 60, and so on. With the correct schedule, month 13 would only have such a drastic jump if a massive number of keeps were to be reported as terminated after interval 12 was allocated.

solidity/contracts/BeaconRewards.sol
```diff
@@ -0,0 +1,85 @@
...
+
+    constructor (
+        uint256 _termLength,
+        address _token,
+        uint256 _minimumKeepsPerInterval,
+        uint256 _firstIntervalStart,
+        uint256[] memory _intervalWeights,
+        address _operatorContract,
+        address _stakingContract
+    ) public Rewards(
```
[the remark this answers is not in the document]
pdyraga: Revisiting it, `_firstIntervalStart` would be problematic but I still think `_termLength` and `_minimumKeepsPerInterval` should be constants.

pdyraga:
I'd love to give this code a test spin for at least a day but before I - or anyone else - do that, we need to define the numbers (see [this](https://github.com/keep-network/keep-core/pull/1863#discussion_r463523219)) comment, to test it on real values and approve those values for deployment. We don't want to come up with them when working on deployment scripts.

solidity/contracts/Rewards.sol
```diff
@@ -0,0 +1,611 @@
...
+    uint256 public unallocatedRewards;
+    // Rewards that have been paid out;
+    // `token.balanceOf(address(this))` should always equal
+    // `totalRewards.sub(paidOutRewards)`
+    uint256 public paidOutRewards;
+
+    // Length of one interval in seconds (timestamp diff).
+    uint256 public termLength;
+    // Timestamp of first interval beginning.
+    uint256 public firstIntervalStart;
```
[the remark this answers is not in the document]
eth-r: Yes, that's correct. I'll clarify it in the documentation. If we need to extend rewards further into the past without using the backporting, a further parameter can be added to prevent allocation of intervals until enough time has passed from the creation of the contract.

solidity/test/TestBeaconBackportRewards.js
```diff
@@ -0,0 +1,191 @@
...
+        let balance1 = await token.balanceOf(beneficiary1);
+        expect(balance1).to.eq.BN(500)
+
+        await rewards.receiveReward(2);
+        let balance2 = await token.balanceOf(beneficiary2);
+        expect(balance2).to.eq.BN(2000)
+
+        await rewards.receiveReward(3);
+        let balance1new = await token.balanceOf(beneficiary1)
+        expect(balance1new).to.eq.BN(2500)
```
pdyraga, line 180: I'd love to make this test a little more explanatory, it took me a while to figure out where those balances come from. My proposition:
```
        // 4500 allocated to the first interval:
        //   1500 allocated to group 0
        //   1500 not allocated to group 1 because it's terminated
        //   1500 allocated to group 2
        // 4500 allocated to the second interval:
        //   4500 allocated to group 3

        // 1500 allocated to group 0
        await rewards.receiveReward(0);
        expect(await token.balanceOf(beneficiary1)).to.eq.BN(500)
        expect(await token.balanceOf(beneficiary2)).to.eq.BN(1000)

        // 1500 allocated to group 2
        await rewards.receiveReward(2);
        expect(await token.balanceOf(beneficiary1)).to.eq.BN(1000) // 500+500
        expect(await token.balanceOf(beneficiary2)).to.eq.BN(2000) // 1000+1000

        // 4500 allocated to group 3
        await rewards.receiveReward(3);
        expect(await token.balanceOf(beneficiary1)).to.eq.BN(2500) // 1000+1500
        expect(await token.balanceOf(beneficiary2)).to.eq.BN(5000) // 2000+3000
```

solidity/contracts/Rewards.sol
```diff
@@ -0,0 +1,611 @@
...
+        require(
+            !(interval < allocatedIntervals),
+            "Interval already allocated"
+        );
+        // Allocate previous intervals first
+        if (interval > allocatedIntervals) {
+            allocateRewards(interval.sub(1));
+        }
+        uint256 totalAllocation = _adjustedAllocation(interval);
+        // Calculate like this so rewards divide equally among keeps
```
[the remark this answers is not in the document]
eth-r: The comment was misplaced and should've been moved to `_adjustedAllocation`.

solidity/contracts/BeaconBackportRewards.sol
```diff
@@ -0,0 +1,141 @@
...
+        return excludedGroups[uint256(groupIndexBytes)];
+    }
+
+    // A group is recognized if its index is at most `lastEligibleGroup`
+    // and it isn't listed as excluded.
+    function _recognizedByFactory(bytes32 groupIndexBytes) internal view returns (bool) {
+        return lastEligibleGroup() >= uint256(groupIndexBytes);
+    }
+
+    function _distributeReward(bytes32 groupIndexBytes, uint256 _value) internal {
```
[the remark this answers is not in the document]
eth-r: For each member, we do a token transfer which has a floor of 10k gas if the beneficiary has an existing balance, or 25k gas if the beneficiary has zero tokens. With 64 members this can be up to 1.6M gas. Reducing the beneficiaries to a map may not help much, as each new beneficiary still creates a 20k `SSTORE` operation. Optimization in the form of deduplicating member operators and then using the withdrawal pattern for token transfers, only querying the beneficiary address upon withdrawal, would slightly reduce the gas cost of calling `receiveReward` at the cost of possibly increasing the overall gas cost.

solidity/contracts/Rewards.sol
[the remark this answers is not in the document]
eth-r: Unless a convenient and gas-saving way to pay rewards to the members of keeps/groups is available, the gas limits may prevent executing `receiveReward` on a significant number of keeps in one call. `reportTermination` is a bit more amenable to batching.

solidity/contracts/BeaconBackportRewards.sol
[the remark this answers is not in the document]
eth-r: If the reward for each keep is recorded only once, and verifiable members of the keep can withdraw at their leisure, this can be significantly optimized, actually.

solidity/contracts/BeaconBackportRewards.sol
mhluongo: Yes we do :grin:

solidity/contracts/BeaconRewards.sol
```diff
@@ -0,0 +1,86 @@
...
+        _token,
+        _firstIntervalStart,
+        _intervalWeights
+    ) {
+        operatorContract = KeepRandomBeaconOperator(_operatorContract);
+        tokenStaking = TokenStaking(_stakingContract);
+    }
+
```
pdyraga, line 26: Thoughts on adding:
```
    function receiveReward(uint256 i) public {
        receiveReward(bytes32(i));
    }
``` 
here and to `BeaconBackportRewards`? This would make web3 calls easier.

solidity/contracts/Rewards.sol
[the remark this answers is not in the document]
pdyraga: We synced with @mhluongo on that. It works well except that we'd like to smooth it out in the last month - not all the remaining tokens at once but e.g. add more intervals.

solidity/contracts/BeaconRewards.sol
pdyraga, line 26: Thoughts on adding:
```
    function receiveReward(uint256 i) public {
        receiveReward(bytes32(i));
    }
``` 
here and to `BeaconBackportRewards`? This would make web3 calls easier.

pdyraga (commented):
Completed my second pass through this code. The remaining issues in the order of priority:
- We do not want to give 100% of remaining rewards in the last interval ([this](https://github.com/keep-network/keep-core/pull/1863#discussion_r463530166) conversation). One possible solution could be to add an interval with a weight of the previous interval until all remaining rewards are allocated.
- Add a protection so that `firstIntervalStart` does not have to be in the future ([this](https://github.com/keep-network/keep-core/pull/1863#discussion_r463575302) conversation) - this would give us additional flexibility for deployment time.
- Increase the precision of calculations: first multiply, then divide, use 1e16 multiplier trick from beacon rewards where possible.
- There is no contract for ECDSA rewards yet.

dimpar:
Looked at the 4 last commits made by @pdyraga Lgtm

pdyraga (approved):
This looks good, it's not yet the final solution and we'll have to do some adjustments but let's continue in separate PRs that will be easier to review.

pdyraga:
@benlongstaff Thanks for the review, we'll address all your comments in a separate PR.

solidity/contracts/Rewards.sol
```diff
@@ -0,0 +1,617 @@
...
+    event RewardReceived(bytes32 keep, uint256 amount);
+
+    constructor (
+        address _token,
+        uint256 _firstIntervalStart,
+        uint256[] memory _intervalWeights
+    ) public {
+        token = IERC20(_token);
+        firstIntervalStart = _firstIntervalStart;
+        intervalWeights = _intervalWeights;
```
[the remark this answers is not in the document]
pdyraga: It does not have to sum up to 100. The weight says "what percentage of not yet unallocated KEEP we want to allocate in this interval". The amount of KEEP will eventually go to 0.

PR #1867, opened 2020-06-18
Reduce staking contract size, extract Authorizations contract

pdyraga:
There were two motivations for changes in this PR: make ETH-only staking implementation easier and reduce contract size of staking contract. We were operating very close to the limit and adding any functionality to `TokenStaking` was problematic because of out-of-gas problems during deployment. In this PR we reduce `TokenStaking` contract size from `23904` to `22775` (`-1129`). Hopefully, this will be enough to implement top-ups in the future.

One unrelated change is a fix to `stakeDelegate` function. We were not returning anything from this function and `await` instructions were doing nothing. Now we return a promise and no tests should fail because of timing issues related to this function.

The summary of changes with `TokenStaking` size diff:
- Reduced length of strings in `TokenStaking` contract: `23904 -> 23839 (-65)`.
- Extracted minimum stake calculation to a library: `23839 -> 23663 bytes (-176)`.
- `StakeDelegateable`'s `ownerOperators` an `operators` mappings made internal: `23663 -> 23039 (-624)`.
- `TokenStaking` `token` and `registry` references made internal: `23039 -> 22802 (-237)`.
- `initializationPeriod` and `undelegationPeriod` moved to `TokenStaking`: `22802->22800 (-2)`.

Additionally, cleaned up unused libraries, added KEEP ASCII logo and extracted contract authorization code to a separate contract (`Authorizations.sol`).  We will need the functionality from `StakeDelegateable` and `Authorizations` to implement ETH-only staking and we may not need all the other functionality from `TokenStaking`. This change reduced `TokenStaking` contract size as well `22800 -> 22775 = (-25)`.


solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
```diff
@@ -0,0 +1,19 @@
+pragma solidity 0.5.17;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+library MinimumStakeSchedule {
```
nkuba, line 5: Would it be worth adding docs to this library?

solidity/contracts/StakeDelegatable.sol
```diff
@@ -33,14 +32,6 @@ contract StakeDelegatable {
...
     }
 
-    modifier onlyOperatorAuthorizer(address _operator) {
-        require(
-            operators[_operator].authorizer == msg.sender,
-            "Not operator authorizer"
-        );
-        _;
-    }
-
```
nkuba, line 43: I'm 50-50 on moving this part to `Authorizations`. The function suggests that it's related to authorizations (so +1 for placing it in `Authorizations`). But on the other hand, it requires data from `StakeDelegatable`'s `operators` mapping (a call to an abstract function `authorizerOf` that is implemented here in `StakeDelegatable`).

pdyraga: It's not only about modifier name - `authorizeOperatorContract` from `Authorizations.sol` uses this modifier.

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
pdyraga: Done 👍

PR #1872, opened 2020-06-22
Let the grant manager to withdraw from escrow deposit for revoked grant

pdyraga:
Added `withdrawRevoked` function allowing to withdraw the entire amount that is still deposited in `TokenStakingEscrow` in case the grant has been revoked. Anyone can call this function and the entire amount is transferred back to the grant manager.

`TokenStakingEscrow` does not implement an unlocking scheme for revoked grants and always returns the entire, not-yet-withdrawn amount to the grant manager. The reason for that is that it's very hard (if possible at all) to calculate the amount that should belong to the grantee of a revoked grant and that still remains in the escrow. If really needed, we leave this to the grant manager.

### DIFF
--- a/solidity/contracts/TokenStakingEscrow.sol
+++ b/solidity/contracts/TokenStakingEscrow.sol
@@ -281,7 +281,7 @@ contract TokenStakingEscrow is Ownable {
         address operator,
         address grantManager
     ) internal {
-        uint256 amount = deposit.amount - deposit.withdrawn;
+        uint256 amount = deposit.amount.sub(deposit.withdrawn);
 
         deposits[operator].withdrawn = deposit.amount;
         keepToken.safeTransfer(grantManager, amount);
--- a/solidity/contracts/TokenStakingEscrow.sol
+++ b/solidity/contracts/TokenStakingEscrow.sol
@@ -129,16 +129,23 @@ contract TokenStakingEscrow is Ownable {
     /// @notice Returns the currently withdrawable amount that was previously
     /// deposited in the escrow after undelegating it from the provided operator.
     /// Tokens are unlocked base on their grant unlocking schedule.
-    /// Function returns 0 for non-existing deposits and revoked grants.
+    /// Function returns 0 for non-existing deposits and revoked grants if they
+    /// have been revoked before they fully unlocked.
     /// @param operator Address of the operator for which undelegated tokens
     /// were deposited.
     function withdrawable(address operator) public view returns (uint256) {
         Deposit memory deposit = deposits[operator];
 
-        // Staked tokens can be only withdrawn by grantee for non-revoked grant.
+        // Staked tokens can be only withdrawn by grantee for non-revoked grant
+        // assuming that grant has not fully unlocked before it's been
+        // revoked.
+        //
         // It is not possible for the escrow to determine the number of tokens
         // it should return to the grantee of a revoked grant given different
         // possible staking contracts and staking policies.
+        //
+        // If the entire grant unlocked before it's been reverted, escrow
+        // lets to withdraw the entire deposited amount.
         if (getAmountRevoked(deposit.grantId) == 0) {
             (
                 uint256 duration,
@@ -218,7 +225,7 @@ contract TokenStakingEscrow is Ownable {
 
         require(
             getAmountRevoked(deposit.grantId) > 0,
-            "Grant not revoked"
+            "No revoked tokens to withdraw"
         );
 
         address grantManager = getGrantManager(deposit.grantId);
--- a/solidity/contracts/TokenStakingEscrow.sol
+++ b/solidity/contracts/TokenStakingEscrow.sol
@@ -48,6 +48,11 @@ contract TokenStakingEscrow is Ownable {
         address indexed grantee,
         uint256 amount
     );
+    event RevokedDepositWithdrawn(
+        address indexed operator,
+        address indexed grantManager,
+        uint256 amount
+    );
 
     IERC20 public keepToken;
     TokenGrant public tokenGrant;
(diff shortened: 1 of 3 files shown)


solidity/contracts/TokenStakingEscrow.sol
```diff
@@ -247,6 +269,19 @@ contract TokenStakingEscrow is Ownable {
         emit DepositWithdrawn(operator, grantee, amount);
     }
 
+    function withdrawRevoked(
```
nkuba, line 272: Why we decided to implement two functions rather than put it into one?

solidity/contracts/TokenStakingEscrow.sol
```diff
@@ -247,6 +269,19 @@ contract TokenStakingEscrow is Ownable {
...
+        Deposit memory deposit,
+        address operator,
+        address grantManager
+    ) internal {
+        uint256 amount = deposit.amount - deposit.withdrawn;
+
+        deposits[operator].withdrawn = deposit.amount;
```
nkuba, line 279: Can revocation be performed in a couple of turns with different amounts or is it happening just once for total amount of tokens?

solidity/contracts/TokenStakingEscrow.sol
```diff
@@ -247,6 +269,19 @@ contract TokenStakingEscrow is Ownable {
         emit DepositWithdrawn(operator, grantee, amount);
     }
 
+    function withdrawRevoked(
+        Deposit memory deposit,
+        address operator,
+        address grantManager
+    ) internal {
+        uint256 amount = deposit.amount - deposit.withdrawn;
```
nkuba, line 277: Are we certain that the amount calculated this way is always equal to the revoked amount?

solidity/test/token_stake/TestTokenStakingEscrow.js
```diff
@@ -16,13 +16,15 @@ const chai = require('chai')
 chai.use(require('bn-chai')(BN))
 const expect = chai.expect
 
-describe('TokenStakingEscrow', () => {
+describe.only('TokenStakingEscrow', () => {
```
nkuba, line 19: `.only` :bomb:

solidity/test/token_stake/TestTokenStakingEscrow.js
```diff
@@ -16,13 +16,15 @@ const chai = require('chai')
...
   
   const deployer = accounts[0],
+    grantManager = accounts[0],
```
nkuba, line 22: I usually prefer to distinguish accounts for roles tests from the default account. `0` is a default account used for deployment and in most cases we expect the ownership or roles to be transferred to other accounts. Using the default account can give false-positive test results at some point in the future.

solidity/test/token_stake/TestTokenStakingEscrow.js
```diff
@@ -273,7 +275,7 @@ describe('TokenStakingEscrow', () => {
 
     it('can not be called by third-party', async () => {
       await expectRevert(
-        escrow.withdraw(operator, {from: deployer}),
```
nkuba, line 276: Maybe we should also leave the cases to test if `deployer` is not able to call the function?

solidity/test/token_stake/TestTokenStakingEscrow.js
```diff
@@ -413,6 +415,85 @@ describe('TokenStakingEscrow', () => {
...
+
+      const balanceBefore = await token.balanceOf(grantManager)
+      await escrow.withdrawRevoked(operator, {from: grantManager})
+      const balanceAfter = await token.balanceOf(grantManager)
+
+      const diff = balanceAfter.sub(balanceBefore)
+      expect(diff).to.eq.BN(5000) // 10000 - 5000 = 5000
+    })
+
+    it('withdraws entire deposited amount if nothing has been withdrawn before', async () => {
```
nkuba, line 474: Do we want to add a case where the whole amount has already been withdrawn but `withdrawRevoked` function is called again?

solidity/test/token_stake/TestTokenStakingEscrow.js
pdyraga: 🤦

solidity/test/token_stake/TestTokenStakingEscrow.js
pdyraga: :+1: I'll add similar test to `withdrawToManagedGrantee` and `withdraw`.

solidity/contracts/TokenStakingEscrow.sol
pdyraga: Grant once revoked cannot be revoked again and grant one revoked cannot be un-revoked.

solidity/contracts/TokenStakingEscrow.sol
pdyraga: It's a different operation. `withdraw` and `withdrawToManagedGrantee` are called by grantee or grantee's operator to withdraw unlocked tokens to the grantee's account.

`withdrawRevoked` is a kind of a recovery function allowing grant manager to reclaim tokens from a revoked grant to do not lock them in the escrow. It's not a default flow, it's something called in exceptional cases. Those two operations are also separated in `TokenGrant` contract.

solidity/contracts/TokenStakingEscrow.sol
pdyraga: It's not equal to the revoked amount and your question motivated me to improve docs even further in https://github.com/keep-network/keep-core/pull/1872/commits/f7ba39562623f4de09f94e307d53d2bfe52b97c6. 

We can't calculate the exact amount for a revoked grant that should be withdrawable from escrow. There are different staking policies, we are going to introduce stake top-ups, part of the tokens could be withdrawn before... It's just too hard if possible at all. We do a simplification:
- grant not revoked - let grantee withdraw based on the unlocking schedule
- grant revoked but fully unlocked - let grantee withdraw everything
- grant revoked and not fully unlocked - do not let grantee withdraw more than already withdrawn; grant manager takes the rest and they may split the tokens out of band.

solidity/contracts/TokenStakingEscrow.sol
nkuba: I'm sorry I wasn't specific. What I mean are `withdrawRevoked` (public) and `withdrawRevoked` (internal).

pdyraga: Oh, I thought it's just easier to follow. We have the standard case (`internal withdraw`) above the revoked case (`internal withdrawRevoked`) and they look pretty similar except that they calculate the amount differently, transfer tokens to different accounts, and fire different events. It's probably not that clear on a diff, but if you open the entire file and look at lines `266 - 290`, you'll see it then.

If you don't think it's better this way, I am happy to make it one function.

solidity/contracts/TokenStakingEscrow.sol
```diff
@@ -247,6 +276,19 @@ contract TokenStakingEscrow is Ownable {
...
```
nkuba, line 284: Should we use SafeMath's `.sub`  here?

pdyraga: Yeah, I think it's a good practice. We should never have `depositAmount > deposit.withdraw` but better be safe than sorry.

solidity/contracts/TokenStakingEscrow.sol
nkuba: We can leave it as is.

Issue #1856, opened 2020-06-12
Keep a local list of peers to facilitate bootstrapping upon reboot


jeremyid:
@mhluongo said on Discord that "It would be really nice to have some way to save peers between restarts".

Indeed it would, as this would both limit the risk of eclipse attacks as underlined by Matt, as well as prevent issues such as outdated bootstrap peers list in the config file that would require manual update.

One way to do it would be for the node to keep a routing table in the persistence folder and use it as a bootstrap list.

jeremyid:
I have identified a possible way to implement this for this. I would like to do this first for the ECDSA client. I know which Go routines I want to change - however I need help setting up a local environment to build everything and test my changes as I am new to this and the documentation hasn't helped for what I want to do.

@mhluongo Would anyone from the team be willing to provide some help, perhaps through a call with screen sharing or something of the sort?

mhluongo:
I don't think we could do 1:1 for a bit, considering deadlines for existing work. We're happy to help here or we could maybe do a group dev call to make the time higher leverage

jeremyid:
I'm fine with a group dev call. Anything to make it clearer to community devs how to run things locally to test various kinds of changes. I think a few other people would be interested, maybe we could gauge interest on Discord.

mhluongo:
cc @Shadowfiend @sthompson22 ideas or pointers on a better local setup?

@jeremyid anything you can share about your issues here would help!

jeremyid:
I think I am almost there thanks to help from various community members. 

What I'm trying to do involves changes in both ``keep-core/.../libp2p.go`` and ``keep-ecdsa/.../start.go``. What I did:

- Forked both repos
- Added an instruction the ``replace``of ``keep-ecdsa``'s ``go.mod`` to point to my own forked repo: ``github.com/keep-network/keep-core => github.com/jeremyid/keep-core v.1.2.3-rc``
- Built the ECDSA docker image by running ``docker build .`` from ``keep-ecdsa`` 

I was able to run this image after creating the ``config.toml`` and other appropriate files / folders. 

The only thing left to do is figuring out how to work with ``v1.0.0`` instead of ``v1.1.1-rc`` which is the last release that was available when I forked my repo. I believe this is a noob Git problem though so I can probably figure it out on my own.

mhluongo:
On the last point, branching off a tag would be the way to go. I'm not certain you should be off `v1.0.0` though- @sthompson22 what are the bootstraps running?

[...]

PR #1877, opened 2020-06-24
Save last seen peers and use them as bootstrap upon restart

experiencedft:
Follow up to issue #1856. 

Sorry if the commits are not clean, it's my first time contributing to a project with Git and there was a lot of weird stuff going on.

These changes allow the user to specify a file where to store the last seen peers, and read from that file to use them as bootstrap peers upon restart.

The run command now requires a new flag ``--peers`` which specifies the path to the relevant file. For example, using the docker image for v1.1.3, one could use ``--peers /mnt/keep-client/config/peers.dat``.

The last seen peers are wrote to the specified file by the ``monitorConnectedPeers`` function in keep-core/libp2p.go.

Thanks to Discord users ssh, pantsme and postables for the help with either Go, Git, Docker or LibP2P stuff.


mhluongo:
Excited about this @jeremyid! We're heads down on this next release, but we'll get it triaged soon.

We'll likely need to rebase and squash most of this commit history. On the actual changes, I'm pretty sure the team will want to unify how state is treated and take a different approach on the re-connect, but hoping we can work together on those changes in this PR.

jeremyid:
> We'll likely need to rebase and squash most of this commit history. 

I wish I would know how to deal with all the version control things, sorry about that @mhluongo, I settled for a PR without conflicts and didn't try to go further. It was particularly hard figuring out how to be able to work on this live on testnet which is what led me to revert to 1.0.0 and do some revert commit things.

I'm happy to keep working on these PRs, eager to know what the devs think of this and if they have any suggestion to improve the approach I took.

mhluongo:
> I wish I would know how to deal with all the version control things, sorry about that

Don't be!

> eager to know what the devs think

:ok_hand:

pdyraga:
Thanks for the contribution @jeremyid 🙌  and I am sorry for being late here. 

I like very much your idea about storing the peer list locally but I would like to propose two changes.

First, I don't think we should treat those peers as bootstraps. Although there is no difference on the network level and bootstraps do not expose any additional services, bootstrap peers are treated differently than other discovered peers. The most important difference is that the client always tries to connect to all bootstrap peers and not being able to establish a connection with bootstrap is treated as a failure. Please see https://github.com/keep-network/go-libp2p-bootstrap for details on that behavior. Also, there are separate metrics for bootstrap peers (see https://github.com/keep-network/keep-core/pull/1850) and they allow evaluating the network health for the given node. 

Not being able to connect to a non-bootstrap peer is not great but also not something we should consider a failure. More importantly, for bigger networks, peers may decide to which other peers they want to connect to based on the topics they are subscribed to.

Instead of using last seen peers as bootstrap peers we should find a way to provide them to the routing table and let the node decide in which order to connect and whether to connect or not to connect. I hope that @lukasz-zimnoch can shed some light on whether this is easily achievable or if we need to contribute something to libp2p to make it happen.

My second comment is about moving the last seen peers file location to the config file. All options but the log level and config file location are configured there. Honestly, I think we could even reuse the existing storage option for that. Just add another directory next to `current` and `archive`, perhaps `network`?

jeremyid:
Hey @pdyraga -- I agree with most of your remarks, my PR is functional but was meant to be a proof of concept rather than a fully fleshed out addition. 

I am aware of the fact that failure to connect to some of the specified bootstrap peers triggers a warning, though I am not sure why the code is such that the node keeps calling the bootstrap function once it is in fact connected to the network. I looked at #1850 and saw the discussion about ``connected_bootstrap_count`` as a health metrics -- I don't think this is necessarily a great metric personally but I will comment there more specifically.

On the topic of routing tables, there is this related [LibP2P issue](https://github.com/libp2p/go-libp2p-kad-dht/issues/254) that might be of interest to you and the team.

PR #1878, opened 2020-06-24
Cleaning group seletion candidates before genesis

dimpar:
Cleared group selection candidates map in case previous genesis failed. Without cleaning this map, we might get a "duplicate ticket" error, because members that participated in the previous genesis cannot participate in the next one.


dimpar:
Testing:
- before the candidate map cleanup, a second genesis was failing because of the remaining tickets in the map from the previous genesis. Confirmed with the revert message.
- after the cleanup, recreated the same scenario, when the first genesis fails. However, the second genesis went through.

pdyraga:
`generate_docs_solidity` is probably failing because of https://github.com/keep-network/keep-core/pull/1880 but Solidity tests failing are because of the changes in this PR.

solidity/test/random_beacon_operator/TestPricingRewardsWithdraw.js
```diff
@@ -152,28 +152,31 @@ describe('KeepRandomBeaconOperator/PricingRewardsWithdraw', function() {
...
+    let firstActiveGroupIndex = await operatorContract.getFirstActiveGroupIndex()
+    assert.equal(firstActiveGroupIndex.toString(), 1, "Active group index should be equal to 1")
 
-    assert.isFalse(await operatorContract.isExpiredGroup('0x' + group2.toString('hex')), "Group should not be expired")
     let beneficiary1balance = web3.utils.toBN(await web3.eth.getBalance(beneficiary1))
 
     // Nothing can be withdrawn
     await expectRevert(
-      operatorContract.withdrawGroupMemberRewards(operator1, 1),
+      operatorContract.withdrawGroupMemberRewards(operator1, firstActiveGroupIndex),
```
pdyraga, line 162: Can we take this opportunity and refactor this test to use:
```
const BN = web3.utils.BN
const chai = require('chai')
chai.use(require('bn-chai')(BN))
const expect = chai.expect
```

and
```
expect(...).to.eq.BN(...)
```
construction? Please see `TestTokenStakingEscrow.js` for an example.

dimpar: sure, done

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -147,6 +147,8 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
         // groups expired, we first want to make a cleanup.
         groups.expireOldGroups();
         require(numberOfGroups() == 0, "Groups exist");
+        // Executing candidate cleanup in case of previous genesis failure.
+        groupSelection.cleanupCandidates();
```
pdyraga, line 151: Is there a reason why do we call only `clenupCandidates()` and not the full cleanup? Cleaning up just candidates is probably enough but executing the full cleanup could be less confusing. This could also be an opportunity to make compiler happier and eliminate:
```
    function stop(Storage storage self) public {
    ^ (Relevant source part starts here and spans across multiple lines).
,/Users/piotr/go/src/github.com/keep-network/keep-core/solidity/contracts/libraries/operator/GroupSelection.sol:100:5: Warning: Variable is shadowed in inline assembly by an instruction of the same name
    function stop(Storage storage self) public {
```

We could rename `groupSelection.stop()` to e.g. `groupSelection.finish()` and call the entire `groupSelection.finish()` (which is already `public`) here.

What do you think?

dimpar: > Is there a reason why do we call only clenupCandidates() and not the full cleanup?

In current implementation for genesis we do the full cleanup, but it is broken down to 2 pieces:
- candidates cleanup in `genesis()`
- tickets cleanup in `genesis -> startGroupSelection -> groupSelection.start()`

> We could rename groupSelection.stop() to e.g. groupSelection.finish() and call the entire groupSelection.finish() 

Calling `groupSelection.finish()` from `genesis()`? If yes, then we would need to refactor code so we won't call cleanupTickets 2 times. But at the same time, we execute `cleanupTickets()` every time (createGroup()) when DKG fee pool has enough eth.

pdyraga: Yeah, for sure we can'd change anything in `startGroupSelection` because we'll run into gas estimate problems again. We need to limit ourselves to changes in `genesis`.

I think it's not that bad if we `cleanupTickets` twice. It's not an expensive function and I think we can sacrifice just a little more gas during `genesis` for a cleaner code. I think we should have:

```
groupSelection.finish() // cleanup after potential failed DKG
startGroupSelection(...)
```

IMO it's a little cleaner compare to what we have now. What do you think?

dimpar: If it's not bad calling `cleanupTickets` twice, then sure, that was my only concern.

pdyraga: Yeah, I think in all other cases but genesis I'd say it's not the best solution. In this single case, I think we are fine making this tradeoff.

dimpar:
Run genesis request to double check that everything works. Looked fine.

PR #1880, opened 2020-06-26
Fixing eth-json-rpc-filter

dimpar:
Problem seems to be in `web3-provider-engine` and `web3`. Because of the `^` before the version number, it updated to latest versions which broke the dependencies.

- `web3` to fixed 1.2.7 (changed because of the `Error: Cannot find module 'scrypt-shim`)
- `web3-provider-engine` to fixed 15.0.6 (changed because of the `Error: Cannot find module eth-json-rpc-filter`)


pdyraga:
I'll leave the final decision to @r-czajkowski

PR #1881, opened 2020-06-29
Improvements for DKG reconstruction phase

lukasz-zimnoch:
Until now, the reconstruction of the public key share of an inactive/disqualified member was performed only when such member was part of the QUAL set (i.e. their shares broadcasted in phase 3 were valid). Unfortunately, this approach had an uncovered edge case when a member becomes inactive/disqualified exactly in phase 9. In that case, an inactive/disqualified member belongs to the QUAL set, it is selected for reconstruction and its reconstructed public key share is added to the group public key. But, becoming inactive/disqualified in phase 9 means that the member also provided valid public key share during phase 7 which are also added to the group public key. As a result, the group public key becomes invalid and causes errors while signing. The proposed fix improves the `membersForReconstruction` method by checking also the received valid public key shares map.

### DIFF
--- a/pkg/beacon/relay/gjkr/integration_test.go
+++ b/pkg/beacon/relay/gjkr/integration_test.go
@@ -904,16 +904,12 @@ func TestExecute_DQ_member2_notRevealsDisqualifiedQualMemberKey_phase11(t *testi
 	seed := dkgtest.RandomSeed(t)
 
 	interceptor := func(msg net.TaggedMarshaler) net.TaggedMarshaler {
-		// Member 1 misbehaves by sending an accusation with wrong private key.
+		// Member 1 misbehaves by sending an invalid public key share points message.
 		// As result, they should be disqualified by other members.
-		accusationsMessage, ok := msg.(*gjkr.PointsAccusationsMessage)
-		if ok && accusationsMessage.SenderID() == group.MemberIndex(1) {
-			randomKeyPair, _ := ephemeral.GenerateKeyPair()
-			accusationsMessage.SetAccusedMemberKey(
-				group.MemberIndex(2),
-				randomKeyPair.PrivateKey,
-			)
-			return accusationsMessage
+		publicKeyShareMessage, ok := msg.(*gjkr.MemberPublicKeySharePointsMessage)
+		if ok && publicKeyShareMessage.SenderID() == group.MemberIndex(1) {
+			publicKeyShareMessage.RemovePublicKeyShare(2)
+			return publicKeyShareMessage
 		}
 
 		// Member 2 does not reveal private key generated for the sake
--- a/pkg/beacon/relay/gjkr/protocol.go
+++ b/pkg/beacon/relay/gjkr/protocol.go
@@ -536,7 +536,7 @@ func (cm *CommittingMember) areSharesValidAgainstCommitments(
 // - accused inactive or already disqualified member and as a result, we do not
 //   have enough information to resolve that accusation
 // - shares of the accused member are valid against commitments
-// - accused member ID does not exist 
+// - accused member ID does not exist
 //
 // Accused member is disqualified if:
 // - shares of the accused member can not be decrypted
@@ -985,7 +985,7 @@ func (pjm *PointsJustifyingMember) ResolvePublicKeySharePointsAccusationsMessage
 	for _, message := range messages {
 		accuserID := message.senderID
 		for accusedID, revealedAccuserPrivateKey := range message.accusedMembersKeys {
-			if pjm.ID == accusedID || int(accusedID) > pjm.group.GroupSize()  {
+			if pjm.ID == accusedID || int(accusedID) > pjm.group.GroupSize() {
 				// The member does not resolve the dispute as an accused
 				// or the accussed member ID is not valid.
 				// Mark the accuser as disqualified immediately,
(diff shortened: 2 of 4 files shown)


pkg/beacon/relay/gjkr/integration_test.go
```diff
@@ -904,16 +904,12 @@ func TestExecute_DQ_member2_notRevealsDisqualifiedQualMemberKey_phase11(t *testi
...
-			randomKeyPair, _ := ephemeral.GenerateKeyPair()
-			accusationsMessage.SetAccusedMemberKey(
-				group.MemberIndex(2),
-				randomKeyPair.PrivateKey,
-			)
-			return accusationsMessage
+		publicKeyShareMessage, ok := msg.(*gjkr.MemberPublicKeySharePointsMessage)
+		if ok && publicKeyShareMessage.SenderID() == group.MemberIndex(1) {
+			publicKeyShareMessage.RemovePublicKeyShare(2)
+			return publicKeyShareMessage
```
pdyraga, line 912: I don't think I follow... why did we have to modify this test?

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -1186,18 +1186,26 @@ func (rm *RevealingMember) RevealMisbehavedMembersKeys() (
 func (rm *RevealingMember) membersForReconstruction() []group.MemberIndex {
 	members := make([]group.MemberIndex, 0)
 
-	// From disqualified members list filter those who provided valid shares in
-	// Phase 3 and are sharing the group private key.
+	// The given member needs reconstruction if two conditions are met:
```
pdyraga, line 1189: "member *shares* needs reconstruction", perhaps?

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -1186,18 +1186,26 @@ func (rm *RevealingMember) RevealMisbehavedMembersKeys() (
...
+	// - member provided qualified shares in Phase 3
```
pdyraga, line 1190: Maybe "member provided valid shares in Phase 3 and is in QUAL set"?

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -1186,18 +1186,26 @@ func (rm *RevealingMember) RevealMisbehavedMembersKeys() (
...
+	// - member DID NOT provide valid public key share in phase 7
+	needsReconstruction := func(member group.MemberIndex) bool {
+		_, providedQualifiedShares := rm.receivedQualifiedSharesS[member]
+		_, providedValidPublicKeyShare := rm.receivedValidPeerPublicKeySharePoints[member]
+
+		return providedQualifiedShares && !providedValidPublicKeyShare
+	}
+
+	// From disqualified members list filter those who need reconstruction.
```
pdyraga, line 1199: Maybe "whose *shares* need to be reconstructed"?

pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -1186,18 +1186,26 @@ func (rm *RevealingMember) RevealMisbehavedMembersKeys() (
...
 	for _, disqualifiedMemberID := range rm.group.DisqualifiedMemberIDs() {
-		if _, ok := rm.receivedQualifiedSharesS[disqualifiedMemberID]; ok {
+		if needsReconstruction(disqualifiedMemberID) {
 			members = append(members, disqualifiedMemberID)
 		}
 	}
 
-	// From inactive members list filter those who provided valid shares in
-	// Phase 3 and are sharing the group private key.
+	// From inactive members list filter those who need reconstruction.
```
pdyraga, line 1206: "whose *shares*" perhaps?

pkg/beacon/relay/gjkr/integration_test.go
lukasz-zimnoch: Here we are testing the following scenario:
1. Member 1 misbehaves, doesn't matter how
2. Member 2 doesn't reveal member 1 as misbehaving before the reconstruction phase. As a result, member 2 should be disqualified.

Until now, member 1 misbehaved by sending an invalid `PointsAccusationsMessage` thus became DQ in phase 9. Member 2 should reveal them as DQ before the reconstruction phase. But, because we changed the logic of `memberForReconstruction` and members marked IA/DQ in phase 9 are no longer selected for reconstruction if they provided valid public key shares in phase 8, this test started to fail because member 2 stopped to reveal member 1 as expected. Because of that, I changed the moment when member 1 becomes DQ in the way it remains selected for reconstruction.

pdyraga: But shouldn't we mark member 1 as disqualified if they send accusation with wrong private key?

pdyraga: OK, we do mark member 1 as disqualified. We had to change this logic so that we mark member 2 as disqualified and it makes perfect sense.

What we are testing here is whether member 2 should be disqualified. And given the current logic, member 2 is not required to reveal ephemeral public key of member 1, if that member sent and invalid `PointsAccusationMessage` in phase 9.

The change Lukasz made aims at disqualifying member 1 earlier, so that member 2 is required to reveal ephemeral key and it's not doing that - this is the behaviour we are stressing here.

pdyraga:
The code looks good, double-checked with DKG and two relay requests, no problems found.

PR #1882, opened 2020-06-29
Ropsten KEEP Faucet: Un retry

sthompson22:
### Intro

During heavy use usually right after a fresh release the faucet gets slammed by grant requests (this is a good thing).  Since we issue grants from a single account this can result in nonce issues when we allow for more than one faucet instance to run at a time.  

### What We Did

- Set the nonce accounting for `pending` transactions.  The  internet has mixed feelings on how useful this is since we're  beholden to the mempool of the ETH node we're connected to.  This can be especially problematic in solutions where you're connected to load balanced ETH nodes.

- Retry one time with an incremented nonce if the initial transaction errors with `Error: nonce too low`.   Hoping this will help tamp down the number of times someone has to hit the faucet again, keeping the overall number of faucet requests down.

### Notes

Ran the `issue-grant.js` script through `eslint`, per the existing config in `/keep-core`.


sthompson22:
Drafting to get some eyes.

infrastructure/gcp/keep-test/google-functions/keep-faucet/issue-grant.js
```diff
@@ -126,12 +155,43 @@ exports.issueGrant = async (request, response) => {
...
             `)
           })
-          .on('error', (error) => {
-            console.error(
-              `Error with account grant transaction: [${error}]; URL was [${request.url}].`
-            )
-            if (!response.headersSent) {
-              response.status(500).send(`
+          .on("error", (error) => {
+            if (error == "Error: nonce too low") {
```
Shadowfiend, line 159: This works, right? Sometimes it's `error.message` or `error.getMessage`.

sthompson22: With the testing I did locally, it worked 😬 ..not sure that's the answer you're looking for though.

sthompson22:
@pdyraga or @nkuba If one of y'all has a breather this week would you mind sanity checking this?

infrastructure/gcp/keep-test/google-functions/keep-faucet/issue-grant.js
```diff
@@ -47,77 +47,106 @@ const keepToken = new web3.eth.Contract(keepTokenAbi, keepTokenAddress)
...
           .approveAndCall(tokenGrantAddress, grantAmount, grantData)
-          .send({ from: keepContractOwnerAddress })
-          .on('transactionHash', (hash) => {
+          .send({ from: keepContractOwnerAddress, nonce: nonce })
+          .on("transactionHash", (hash) => {
             console.log(
               `Submitted grant for [${grantAmount}] to [${granteeAccount}] ` +
-              `with hash [${hash}].`,
+                `with hash [${hash}].`,
+              `with nonce [${nonce}]`
```
Shadowfiend, line 149: Not blocking, but this should probably be on the same line as the above.

infrastructure/gcp/keep-test/google-functions/keep-faucet/issue-grant.js
```diff
@@ -126,12 +155,43 @@ exports.issueGrant = async (request, response) => {
...
           })
-          .on('error', (error) => {
-            console.error(
-              `Error with account grant transaction: [${error}]; URL was [${request.url}].`
-            )
-            if (!response.headersSent) {
-              response.status(500).send(`
+          .on("error", (error) => {
+            if (error == "Error: nonce too low") {
+              const noncePlusOne = nonce + 1
```
Shadowfiend, line 160: Non blocking, but I would call this `nextNonce` or similar.

infrastructure/gcp/keep-test/google-functions/keep-faucet/issue-grant.js
```diff
@@ -126,12 +155,43 @@ exports.issueGrant = async (request, response) => {
...
+              console.error(
+                `Error with account grant transaction, Nonce too low: At [${nonce}], retry at [${noncePlusOne}]; URL was [${request.url}].`
+              )
+              console.log("Resubmitting transaction with higher nonce...")
```
Shadowfiend, line 164: A cleaner implementation here might be to give `issueGrant` an additional parameter, `abortOnFailure`. If it's set to `true`, `on("error"`, send the 500. If not, `on("error"` you can call `exports.issueGrant(request, response, true)`. That lets you save the repetition of this code, and can be adapted later to >1 retry if needed.

PR #1888, opened 2020-07-03
Stake locks extracted out of TokenStaking into a library

pdyraga:
This change allows us to save some bytecode space in TokenStaking for top-ups implementation. The contract size went down from `24926` to `22409` (-`2517` bytes).

I have also introduced initTokenStaking function in `initContracts.js` to keep the details about `TokenStaking` library linking and deployment there and not pollute every unit test with the same logic.

Last but not least, `LockUtils` library has been moved to sit next to `Locks` library.


pdyraga:
```
const TokenStaking = artifacts.require("TokenStaking.sol");

module.exports = async function() {
    let instance = await TokenStaking.deployed();

    let bytecode = instance.constructor._json.bytecode;
    let deployedBytecode = instance.constructor._json.deployedBytecode;
    
    let bytecodeSize = bytecode.length / 2; // size in bytes
    let deployedBytecodeSize = deployedBytecode.length / 2; // size in bytes

    console.log(
      "TokenStaking size of bytecode in bytes = ", 
      bytecodeSize
    );
    console.log(
      "TokenStaking size of deployed bytecode in bytes = ", 
      deployedBytecodeSize
    );
    console.log(
      "TokenStaking initialization and constructor code in bytes = ", 
      bytecodeSize - deployedBytecodeSize
    );
}
```

Before:
```
TokenStaking size of bytecode in bytes =  24926
TokenStaking size of deployed bytecode in bytes =  24434
TokenStaking initialization and constructor code in bytes =  492
```

After:
```
TokenStaking size of bytecode in bytes =  22409
TokenStaking size of deployed bytecode in bytes =  21919
TokenStaking initialization and constructor code in bytes =  490
```

solidity/contracts/libraries/staking/Locks.sol
```diff
@@ -0,0 +1,109 @@
...
+    uint256 public constant maximumLockDuration = 86400 * 200; // 200 days in seconds
+
+    struct Storage {
+        // Locks placed on the operator.
+        // `operatorLocks[operator]` returns all locks placed on the operator.
+        // Each authorized operator contract can place one lock on an operator.
+        mapping(address => LockUtils.LockSet) operatorLocks;
+    }
+
+    function lockStake(
```
nkuba, line 24: Don't we want to add docs for functions?

solidity/test/token_stake/TestTokenStakeLock.js
```diff
@@ -153,13 +137,12 @@ describe('TokenStaking/Lock', () => {
 
     it("should not permit locks that exceed the maximum lock duration", async () => {
       await time.increaseTo(initializationPeriod.add(createdAt).addn(1))
-      let maximumDuration = await stakingContract.maximumLockDuration();
-      let longPeriod = maximumDuration.addn(1);
+      const maximumDuration = time.duration.days(200);
```
nkuba, line 140: Why no longer read the value from the contract?

pdyraga: This value is no longer public on the operator contract. It exists in the library and only the library uses it. To make it public, I need to expose a public function on the operator contract that will cost some bytecode. Given that it's just one unit test, I don't think it's crazy to duplicate it. At least, we are also testing it's not some crazy 1-minute value. Every byte matters given the code we want to add to `TokenStaking`.

solidity/contracts/libraries/staking/Locks.sol
pdyraga: I am not sure... I had two approaches to this and docs will mostly duplicate what's already in `TokenStaking`. Given that it's a library used only by `TokenStaking`, I think we should be fine with having the documentation just there. WDYT?

Issue #1883, opened 2020-06-29
Allow bridging already-staked tokens on new staking contract


Shadowfiend:
This is a light proposal for how to allow existing stakers to port their stake over to the updated staking contract we'll be supporting.

## Background

We'll be leveraging the tBTC deposit pause to deploy a new staking contract with a few features that weren't possible in the v1 release of the Keep core contracts. In particular, this staking contract will support “top ups”, meaning the ability to delegate additional tokens to an existing delegation relationship (i.e., an existing operator/beneficiary/authorizer triplet). This introduces a key complication, however: tokens already delegated through the current staking contract must remain locked in that contract for 60 days, the length of the undelegation period. This means that existing token owners who have used the old staking contract must wait 60 days to port those balances over to the new contract.

This issue outlines a proposal for how to deal with this by providing some additional liquidity from the primary token supply.

## Proposal

The core goal is for the Keep team to make an amount of KEEP tokens available to temporarily stake on behalf of a token owner, provided that they have those same tokens currently locked up in the old staking contract. Undelegation of all tokens in the new staking contract would be blocked until the temporarily-staked tokens provided by the Keep team are repaid by the token owner.

### Mechanism

A number of KEEP tokens equivalent to the currently-staked token amount will be locked in a special contract referred to here as the `StakingPortBacker` contract.

#### “Copying” stake from the old contract to the new contract.

The new staking contract will gain an additional function, `copyStake(address operator)`. `copyStake(address operator)` will request the copy from the `StakingPortBacker` contract, which will look up the given operator's information on the old staking contract, particularly its balance and owner. Only the owner of the relationship can copy it to the new staking contract, the relationship can only be copied once, and undelegation must not be complete on that relationship. The copying operation will transfer an amount of tokens equal to the old staking contract's delegation amount for this relationship from `StakingPortBacker`'s balance to the new staking contract. Once this operation is complete, the new staking contract will configure that relationship on its end to mirror the old one and set a flag on the relationship, referred to here as `hasBackedBalance`. Copying will not be allowed if the token owner has one or more existing delegations on the new staking contract, unless there is exactly one existing delegation to the same operator/authorizer/beneficiary triplet.

#### Limitations on recovering stake and new delegations

[...]

PR #1893, opened 2020-07-06
Stake top-ups

pdyraga:
The idea behind a top-up is to let the staker add KEEP tokens to existing delegations. This is a two-step process: first, staker
initiates a top-up, locking new KEEPs in the staking contract. Once the initialization period is over, staker (or anyone else)
can commit the top-up, increasing operator's stake.

From the smart contract's perspective, top-ups are performed using the same interface as delegations: `receiveApproval` on `TokenStaking` contract. If the given operator is active, tokens are treated as a top-up. If the given operator is not known, tokens are used to create a new delegation.

Operators who have their stake undelegated cannot have their stake topped-up. It means that we still enforce operator's uniqueness - operator address cannot be reused after delegating and recovering the stake. This is important because token staking escrow assumes operator's uniqueness when accepting a token deposit. Without operator's uniqueness, bookkeeping in the escrow would become a nightmare. 

For the same reason, staking contract enforces the same source of token for a top-up as used for the initial delegation. If the initial delegation was done from a grant, the same grant has to be used for a top-up to the operator. If the initial delegation was done using owner's liquid tokens, liquid tokens from the same owner are required for a top-up.

Top-up is a two-step process for gas efficiency reasons. For the random beacon group selection I see no risk - staker is always free to send any number of tickets they want, so postponing commitment of a top-up gives no advantage. I see a little risk of manipulation for a sortition pool, when waiting with committing a top-up may allow for some degree of manipulation when the seed is known. However, given that anyone can commit the top-up, I don't think we add any new issues to the existing threat model given that operator undelegation could pose a similar thread in the pool.

The following use-cases are supported:
- stake topped-up from liquid tokens
- stake topped-up from a grant
- stake topped-up from the escrow

### DIFF
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -108,16 +108,16 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     }
 
     /// @notice Receives approval of token transfer and stakes the approved
-    /// amount or top-ups an existing delegation with the approved amount.
-    /// In case of a top-up, it is expected that the operator stake passed
-    /// initialization period, it is not undelegated and that the top-up is
+    /// amount or adds the approved amount to an existing delegation (a “top-up”).
+    /// In case of an existing delegation, it is required that the operator stake
+    /// passed initialization period, it is not undelegated and that the top-up is
     /// performed from the same source of tokens as the initial delegation.
     /// That is, if the tokens were delegated from a grant, top-up
     /// has to be performed from the same grant. If the delegation was done
     /// using liquid tokens, only liquid tokens from the same owner can be used
     /// to top-up the stake.
-    /// @dev Makes sure provided token contract is the same one linked to this
-    /// contract.
+    /// @dev Requires that the provided token contract be the same one linked to
+    /// this contract.
     /// @param _from The owner of the tokens who approved them to transfer.
     /// @param _value Approved amount for the transfer and stake.
     /// @param _token Token contract address.
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -21,6 +21,7 @@ import "./StakeDelegatable.sol";
 import "./libraries/staking/MinimumStakeSchedule.sol";
 import "./libraries/staking/GrantStaking.sol";
 import "./libraries/staking/Locks.sol";
+import "./libraries/staking/TopUps.sol";
 import "./utils/PercentUtils.sol";
 import "./utils/BytesLib.sol";
 import "./Authorizations.sol";
@@ -40,6 +41,7 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     using SafeERC20 for ERC20Burnable;
     using GrantStaking for GrantStaking.Storage;
     using Locks for Locks.Storage;
+    using TopUps for TopUps.Storage;
 
     event Staked(
         address owner,
@@ -48,6 +50,8 @@ contract TokenStaking is Authorizations, StakeDelegatable {
         address indexed authorizer,
         uint256 value
     );
+    event TopUpInitiated(address indexed operator, uint256 topUp);
+    event TopUpCommitted(address indexed operator, uint256 newAmount);
     event Undelegated(address indexed operator, uint256 undelegatedAt);
     event RecoveredStake(address operator, uint256 recoveredAt);
     event TokensSlashed(address indexed operator, uint256 amount);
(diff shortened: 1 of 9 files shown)


solidity/contracts/TokenStaking.sol
```diff
@@ -107,44 +107,166 @@ contract TokenStaking is Authorizations, StakeDelegatable {
...
+        address _operator,
+        bytes memory _extraData
+    ) internal {
+        uint256 operatorParams = operators[_operator].packedParams;
+        require(
+            _isInitialized(operatorParams),
+            "Initialization period is not over"
+        );
+        require(
+            !_isUndelegating(operatorParams),
```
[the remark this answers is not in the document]
pdyraga: It could also be saying that the stake is currently undelegating and it hasn't been recovered yet - the delegation still exists and the operator has to be active. Really depends on the context but I do agree it could be confusing, especially in this line.

solidity/contracts/TokenStaking.sol
```diff
@@ -107,44 +107,166 @@ contract TokenStaking is Authorizations, StakeDelegatable {
...
+    /// - Authorizer address (20 bytes), ignored for a top-up
+    /// - Grant ID (32 bytes) - required only when called by TokenStakingEscrow
+    function receiveApproval(
+        address _from,
+        uint256 _value,
+        address _token,
+        bytes memory _extraData
+    ) public {
         require(ERC20Burnable(_token) == token, "Unrecognized token contract");
         require(_value >= minimumStake(), "Value must be greater than the minimum stake");
```
nkuba, line 137: The minimum value for a top up has to be at least equal to the minimum stake, this seems not necessary as the minimum amount is already staked.

solidity/contracts/libraries/staking/TopUps.sol
```diff
@@ -0,0 +1,76 @@
...
+    /// @param initializationPeriod Stake initialization period.
+    function commit(
+        Storage storage self,
+        address operator,
+        uint256 initializationPeriod
+    ) public returns (uint256) {
+        TopUp memory topUp = self.topUps[operator];
+        require(topUp.amount > 0, "No top up to commit");
+        require(
+            now > topUp.createdAt.add(initializationPeriod),
```
nkuba, line 51: Why `now` instead `block.timestamp`?

nkuba: `now` will work fine as it's an alias for `block.timestamp`.  We're just inconsistent across our codebase.

pdyraga: The only reason for using `now` was probably that it was faster to type it, heh. I can change it to `block.timestamp` if that's what we use in the rest of the codebase, no problem.

solidity/test/token_stake/TestTopUps.js
```diff
@@ -0,0 +1,676 @@
...
+      await time.increase(initializationPeriod.addn(1))
+      delegationInfo = await tokenStaking.getDelegationInfo(operatorFour)
+      expect(delegationInfo.amount).to.eq.BN(delegatedAmount)
+  
+      await tokenStaking.commitTopUp(operatorFour, {from: managedGrantee})
+        
+      delegationInfo = await tokenStaking.getDelegationInfo(operatorFour)
+      expect(delegationInfo.amount).to.eq.BN(delegatedAmount.muln(2))
+    })
+  })
```
nkuba, line 675: Can we add a test case to cover `No top up to commit` check?

solidity/contracts/TokenStaking.sol
pdyraga: Agreed. Please let me deal with it once https://github.com/keep-network/keep-core/pull/1902 and https://github.com/keep-network/keep-core/issues/1883 are there. We'll have to fight for every single byte because of out-of-gas problems during deployment.

solidity/test/token_stake/TestTopUps.js
pdyraga: Done in https://github.com/keep-network/keep-core/pull/1893/commits/870a6caf5187c43e5fd2d48f9aa6540d31e4f19c

solidity/contracts/TokenStaking.sol
```diff
@@ -107,44 +107,166 @@ contract TokenStaking is Authorizations, StakeDelegatable {
...
         require(_extraData.length >= 60, "Corrupted delegation data");
```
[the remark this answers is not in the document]
pdyraga: Having the experience from #1902: adding a separate `requires` costs more but expanding an existing one shouldn't be a problem. Just to be safe, please let me deal with it in the PR with the existing stake bridging, along [with the comment from @nkuba](https://github.com/keep-network/keep-core/pull/1893#discussion_r450914363).

pdyraga:
> For bytecode after this PR, I think the question for me becomes “can we push more things into libraries if we give those libraries references to other contracts/libraries”. For example, pushing most of `initiateTopUp` into the `TopUp` library by giving it access to escrow, token grant, and grant staking or something of the sort.

That was my first choice. I can't recall exactly why I abandoned this path but I think it was a mix of the number of parameters I had to pass and the fact I couldn't easily make it work without `ABIEncoderV2`. Having all of it working together, I'll make another attempt.

pdyraga:
@Shadowfiend @nkuba Thanks for the review, this is ready for another chance.

solidity/contracts/TokenStaking.sol
nkuba: Covered in https://github.com/keep-network/keep-core/pull/1974

Issue #1900, opened 2020-07-06
Ramping up undelegation period


pdyraga:
Undelegation period has been introduced to protect the network and token. For early days, we could consider having it lower to allow for faster migration between staking contracts and avoid all the hard work in the future we now need to do in https://github.com/keep-network/keep-core/issues/1883.

(*) The shape of the undelegation period function probably depends on the contract size and out-of-gas issues we may have.

PR #1902, opened 2020-07-08
Reduce bytecode size of TokenStaking contract

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/1883
~~Depends on https://github.com/keep-network/keep-core/pull/1893~~

Extracting stake locks to a separate library (https://github.com/keep-network/keep-core/pull/1888) provided some space in `TokenStaking` bytecode but we quickly used all of that space for the implementation of top-ups (https://github.com/keep-network/keep-core/pull/1893). In this PR, we are doing small changes, one by one, to reduce the bytecode size to allow for stake bridge implementation (https://github.com/keep-network/keep-core/issues/1883). This PR reduces the bytecode from `24596` to `24270` (`-326` bytes) and will let us add just a few lines of code, an absolute minimum needed for the stake bridge implementation.

I recommend reviewing this PR commit-by-commit, it's easier to understand changes this way.

### DIFF
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -321,12 +321,11 @@ contract TokenStaking is Authorizations, StakeDelegatable {
             "Timestamp in the past"
         );
         uint256 oldParams = operators[_operator].packedParams;
-        uint256 existingCreationTimestamp = oldParams.getCreationTimestamp();
-        uint256 existingUndelegationTimestamp = oldParams.getUndelegationTimestamp();
         require(
-            _undelegationTimestamp > existingCreationTimestamp.add(initializationPeriod),
+            _undelegationTimestamp > oldParams.getCreationTimestamp().add(initializationPeriod),
             "Timestamp in initialization period"
         );
+        uint256 existingUndelegationTimestamp = oldParams.getUndelegationTimestamp();
         require(
             // Undelegation not in progress OR
             existingUndelegationTimestamp == 0 ||
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -208,7 +208,7 @@ contract TokenStaking is Authorizations, StakeDelegatable {
         );
         require(
             !_isUndelegating(operatorParams),
-            "Operator undelegated"
+            "Stake undelegated"
         );
 
         // Top-up comes from a grant if it's been initiated from TokenGrantStake
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -263,8 +263,7 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     ) internal returns (uint256) {
         uint256 oldParams = operators[_operator].packedParams;
         uint256 newAmount = oldParams.getAmount().add(_topUp);
-        uint256 newParams = oldParams.setAmount(newAmount);
-        operators[_operator].packedParams = newParams;
+        operators[_operator].packedParams = oldParams.setAmount(newAmount);
         return newAmount;
     }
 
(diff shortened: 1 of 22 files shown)


solidity/contracts/TokenStaking.sol
```diff
@@ -578,18 +579,16 @@ contract TokenStaking is Authorizations, StakeDelegatable {
...
-
-        if (isAuthorized && isActive && !isUndelegating) {
+        if (
+            isAuthorizedForOperator(_operator, _operatorContract) &&
+            _isInitialized(operatorParams) &&
+            !_isUndelegating(operatorParams)
+            // `undelegatedAt` may be set to a time in the future,
+            // to schedule undelegation in advance.
+            // In this case the operator is still eligible
+            // until the timestamp `undelegatedAt`.
```
nkuba, line 580: Thoughts on moving the clarification comment out of the if condition?

solidity/contracts/TokenStaking.sol
```diff
@@ -60,8 +60,8 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     event ExpiredLockReleased(address indexed operator, address lockCreator);
 
     uint256 public minimumStakeScheduleStart;
-    uint256 public initializationPeriod;
-    uint256 public undelegationPeriod;
+    uint256 public initializationPeriod; // varies between mainnet and testnet
+    uint256 public constant undelegationPeriod = 5184000; // ~60 days
```
nkuba, line 64: I'm not 100% happy with this change from testing perspective, but I can live with it.

pdyraga: Well... it does not change anything that much and there is also a bright side: we are testing against the value close to the real one.

solidity/contracts/TokenStaking.sol
pdyraga: I am happy you brought it up, I wasn't sure what's the best option here. This comment is only for the line `!_isUndelegating(operatorParams)` and originally I placed it above this line but then, the linter is unhappy about `&&` rules. What's the best solution here in your opinion?

solidity/contracts/TokenStaking.sol
nkuba: It may complicate future system tests a little bit when we won't be able to tweak the value for specific environment or test case.

pdyraga: Agreed but I have a nice surprise for you: https://github.com/keep-network/keep-core/issues/1900

Not sure how it's going to look like yet but I plan to work on it this week.

solidity/contracts/TokenStaking.sol
nkuba: Yep, linter does that 😞  I think that placing it above the whole if statement would be better than below the line it relates to, we could rephrase the doc to clarify to what part it relates to.

solidity/contracts/TokenStaking.sol
nkuba: Props!

solidity/contracts/TokenStaking.sol
pdyraga: Please see https://github.com/keep-network/keep-core/pull/1902/commits/00a7398bd0f4fd1a4955a120ef9d90d3b60aec6b

solidity/contracts/TokenStaking.sol
```diff
@@ -570,14 +570,17 @@ contract TokenStaking is Authorizations, StakeDelegatable {
         address _operatorContract
     ) public view returns (uint256 balance) {
         uint256 operatorParams = operators[_operator].packedParams;
+        // To be eligible for work selection, the operator must:
+        // - have the operator contract authorized
+        // - have the stake initialized
+        // - must not be undelegating; keep in mind the `undelegatedAt` may be
+        // set to a time in the future, to schedule undelegation in advance.
+        // In this case the operator is still eligible until the timestamp
+        // `undelegatedAt`.
```
nkuba, line 579: ❤️

PR #1904, opened 2020-07-09
Cache negative minimum stake check results

### LINKED ISSUE #1895: Cache all minimum stake check results
One part of the connection handshake protocol is to check the minimum stake of the other party. In case the other party has a minimum stake, we cache the positive verification result for the stake of this party for 12 hours. We do not cache negative verification results though and it means a party with no minimum stake trying to connect to a bootstrap node may flood Ethereum client with too many minimum stake check queries.

We should cache the positive verification result and negative verification result. Maybe just for a little shorter period of time for the latter.

lukasz-zimnoch:
Closes  #1895

Introduced cache for negative minimum stake check results in the same way as for positive results.

### DIFF
--- a/pkg/firewall/firewall.go
+++ b/pkg/firewall/firewall.go
@@ -22,10 +22,17 @@ func (nf *noFirewall) Validate(remotePeerPublicKey *ecdsa.PublicKey) error {
 	return nil
 }
 
-// MinimumStakeCachePeriod is the time period the cache maintains the result of
-// the last HasMinimumStake result. We use the cache to minimize calls to
-// Ethereum client.
-const MinimumStakeCachePeriod = 12 * time.Hour
+const (
+	// PositiveMinimumStakeCachePeriod is the time period the cache maintains
+	// the positive result of the last HasMinimumStake check.
+	// We use the cache to minimize calls to Ethereum client.
+	PositiveMinimumStakeCachePeriod = 12 * time.Hour
+
+	// NegativeMinimumStakeCachePeriod is the time period the cache maintains
+	// the negative result of the last HasMinimumStake check.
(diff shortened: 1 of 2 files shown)


pkg/firewall/firewall.go
```diff
@@ -22,25 +22,34 @@ func (nf *noFirewall) Validate(remotePeerPublicKey *ecdsa.PublicKey) error {
...
+const (
+	// PositiveMinimumStakeCachePeriod is the time period the cache maintains
+	// the positive result of the last HasMinimumStake check.
+	// We use the cache to minimize calls to Ethereum client.
+	PositiveMinimumStakeCachePeriod = 12 * time.Hour
+
+	// NegativeMinimumStakeCachePeriod is the time period the cache maintains
+	// the negative result of the last HasMinimumStake check.
+	// We use the cache to minimize calls to Ethereum client.
+	NegativeMinimumStakeCachePeriod = 6 * time.Hour
```
pdyraga, line 34: I'd lower it to one hour, at least initially. Six hours is a lot if someone is not yet very familiar with setting up a node and started it before setting up a delegation.

pkg/firewall/firewall_test.go
```diff
@@ -100,3 +103,50 @@ func TestCachesActiveKeepMembers(t *testing.T) {
 		)
 	}
 }
+
+func TestCachesInactiveKeepMembers(t *testing.T) {
```
pdyraga, line 107: I am really confused why did I name the test `TestCachesActiveKeepMembers`...

Can we rename it to `TestCachesHasMinimumStake` and this test to `TestCachesHasNoMinimumStake`?

PR #1906, opened 2020-07-09
Allow bridging already-staked tokens on new staking contract

pdyraga:
Closes: #1883 
~~Depends on #1902~~

# Background

We'll be leveraging the tBTC deposit pause to deploy a new staking contract with a few features that weren't possible in the v1 release of the Keep core contracts. In particular, this staking contract will support “top-ups” (see #1893), meaning the ability to delegate additional tokens to an existing delegation relationship (i.e., an existing operator/beneficiary/authorizer triplet). This introduces a key complication, however: tokens already delegated through the current staking contract must remain locked in that contract for 60 days, the length of the undelegation period. This means that existing token owners who have used the old staking contract must wait 60 days to port those balances over to the new contract.

# Proposal

The core goal is for the Keep team to make an amount of KEEP tokens available to temporarily stake on behalf of a token owner, provided that they have those same tokens currently locked up in the old staking contract. Undelegation of all tokens in the new staking contract would be blocked until the temporarily-staked tokens provided by the Keep team are repaid by the token owner.

The expected mode of usage for users who have already staked on the old contract is to:
1. Copy stake.
2. Undelegate from the old staking contract.
3. Operate as normal on the new staking contract.
4. Once the old staking contract's undelegated balance is available for recovery, recover it and use the recovered tokens to repay the `StakingPortBacker` contract.

# Implemenation

A number of KEEP tokens equivalent to the currently-staked token amount will be locked in a special `StakingPortBacker` contract.

The implementation is very similar to what's been proposed in #1883 except the four major differences:
- `copyStake` and repay mechanism (by `receiveApproval`) are defined on `StakingPortBacker`
- Until the delegation is paid back, the owner of the new delegation is the `StakingPortBacker`. Once the delegation is paid back, the ownership of the delegation is transferred to the owner of the original staking relationship. If the delegation has not been paid back on time, the owner of `StakingPortBacker` may decide to force-undelegate from the new staking contract.
- The owner of `StakingPortBacker` needs to explicitly mark which delegations are allowed to be copied. The reason for explicitly listing which relationships are allowed to use the supply is to avoid a situation when someone stakes on the old contract later, after deploying the new staking contract version, just to drain the supply from `StakingPortBacker` contract. Explicitly specifying which relationships are allowed (by calling `addAllowedOperator`) lets the contract owner to snapshot the staking state before transferring the tokens to provide additional liquidity for this snapshot.
- Top-ups are not allowed until the delegation is paid back. I assume it to be an additional incentivize to pay back sooner than later. I have also gone through the grants and it does not seem to be a very severe limitation given what's currently staked and how stakeaheads were used.


solidity/contracts/StakingPortBacker.sol
```diff
@@ -0,0 +1,269 @@
...
+
+    /// @notice The maximum allowed time for the token owner to repay the
+    /// delegation backed by tokens from this contract. After this time, the
+    /// owner of the contract can undelegate and recover not paid back
+    /// delegations.
+    uint256 public constant maxAllowedBackingDuration = 7776000; // ~3 months
+
+    IERC20 public keepToken;
+    TokenGrant internal tokenGrant;
+    StakeDelegatable public oldTokenStaking;
```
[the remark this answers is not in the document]
pdyraga: I am not sure about it. `oldStake` suggests to me that it's just a value - how many KEEP tokens you have staked. Maybe: `oldStakigContract` and `newStakingContract`?

solidity/contracts/StakingPortBacker.sol
```diff
@@ -0,0 +1,269 @@
...
+    /// staking contract that will be allowed to use the token supply provided
+    /// by this contract.
+    function addAllowedOperator(address operator) public onlyOwner {
+        allowedOperators[operator] = true;
+    }
+
+    /// @notice Copies staking relationship from the old token staking contract
+    /// to the new staking contract. Only the owner of the relationship can copy
+    /// it to the new staking contract, the relationship can only be copied once,
+    /// and undelegation must not be complete on that relationship.
```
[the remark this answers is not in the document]
pdyraga: The version of `TokenStaking` contract deployed to mainnet referred here as the old token staking contract does not recognize top-ups or escrow. `StakingPortBacker`, as currently implemented, is designed specifically for the current needs.

solidity/contracts/StakingPortBacker.sol
```diff
@@ -0,0 +1,269 @@
...
+    /// Operator, beneficiary, authorizer, and stake amount will be the same
+    /// as in the original delegation. Until the delegation is repaid, this
+    /// contract is the owner of the delegation on the new staking contract.
+    /// @param operator The operator from the staking relationship on the old
+    /// staking contract that should be copied to the new staking contract.
+    function copyStake(address operator) public {
+        uint256 oldStakeBalance = oldTokenStaking.balanceOf(operator);
+        require(oldStakeBalance > 0, "No stake on the old staking contract");
+        require(copiedStakes[operator].amount == 0, "Stake already copied");
+        require(allowedOperators[operator], "Not allowed operator");
```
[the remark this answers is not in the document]
pdyraga: Yep, that's a good idea. And during the rename, I realized this case is not covered with a unit test. I'll add one.

solidity/contracts/StakingPortBacker.sol
```diff
@@ -0,0 +1,269 @@
...
+
+        emit StakePaidBack(stake.owner, operator);
+    }
+
+    /// @notice Undelegates stake on the new staking contract from the provided
+    /// operator.
+    /// @param operator The operator address.
+    function undelegate(address operator) public {
+        CopiedStake memory stake = copiedStakes[operator];
+        require(stake.owner == msg.sender, "Not authorized");
```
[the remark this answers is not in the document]
pdyraga: Yes, good catch. The operator should be able to undelegate.

solidity/contracts/StakingPortBacker.sol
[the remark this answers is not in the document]
nkuba: > Maybe: `oldStakigContract` and `newStakingContract`?

👍

solidity/contracts/StakingPortBacker.sol
```diff
@@ -0,0 +1,273 @@
...
+///
+/// The expected mode of usage for users who have already staked on the old
+/// contract is to:
+/// 1. Copy stake.
+/// 2. Undelegate from the old staking contract.
+/// 3. Operate as normal on the new staking contract.
+/// 4. Once the old staking contract's undelegated balance is available for
+/// recovery, recover it and use the recovered tokens to repay the
+/// StakingPortBacker contract.
+contract StakingPortBacker is Ownable {
```
nkuba, line 50: Do we want to include this contract in `2_deploy_contracts.js`?

pdyraga: I... don't think so. We don't have two staking contracts there.

solidity/contracts/StakingPortBacker.sol
```diff
@@ -0,0 +1,273 @@
...
+    /// @notice Copies staking relationship from the old token staking contract
+    /// to the new staking contract. Only the owner of the relationship can copy
+    /// it to the new staking contract, the relationship can only be copied once,
+    /// and undelegation must not be complete on that relationship.
+    /// Operator, beneficiary, authorizer, and stake amount will be the same
+    /// as in the original delegation. Until the delegation is repaid, this
+    /// contract is the owner of the delegation on the new staking contract.
+    /// @param operator The operator from the staking relationship on the old
+    /// staking contract that should be copied to the new staking contract.
+    function copyStake(address operator) public {
```
nkuba, line 123: Do we expect the old stake to be undelegated before copying it to the new token staking contract?

solidity/contracts/StakingPortBacker.sol
```diff
@@ -0,0 +1,273 @@
...
+    /// relaitonships that will be allowed to use the token supply provided by
+    /// this contract.
+    /// @dev The reason for explicitly listing which relationships are allowed
+    /// to use the supply is to avoid a situation when someone stakes on the old
+    /// contract later, after deploying the new staking contract version, just
+    /// to drain the supply from this contract.
+    /// @param operator The operator from the staking relationship on the old
+    /// staking contract that will be allowed to use the token supply provided
+    /// by this contract.
+    function allowOperator(address operator) public onlyOwner {
```
Shadowfiend, line 110: Any chance we can make this take an array?

pdyraga: Yeah... that's a good point. I think I'll add another function `allowOperators` with an array calling this one in a loop.

solidity/contracts/StakingPortBacker.sol
pdyraga: I had to think about it a little more but it appears to me it doesn't really matter. We are going to deploy a new operator contract and that new operator contract will refer only to the new staking contract. In the worst scenario, we can have someone staking on the old staking contract and using the old operator contract and staking on the new staking contract using the additional liquidity we provide and using the new operator contract. After tBTC is relaunched, it's going to use the new operator contract and even if we assume the old operator contract has the potential of providing some yields (unlikely), we can still force-undelegate from the new contract if the delegation is not paid back on time.

tl;dr; I don't think it's worth introducing additional complexity.

solidity/contracts/StakingPortBacker.sol
Shadowfiend: Perhaps based on an environment variable or similar? I don't think it needs to block this PR though.

pdyraga: `StakingPortBacker` [is implemented with an assumption](https://github.com/keep-network/keep-core/pull/1906#discussion_r454402750) that the "old" staking contract version does not support top-ups and does not work with escrow. If we wanted to simulate mainnet situation, we should deploy the previous version of `TokenStaking` as well. I am not sure how important it is, we'll probably need it only one time when testing the migration.

pdyraga:
The last two commits contain a fix for properly identifying the source of staked tokens. We can have a situation when someone staked to operator `A` from a `TokenGrant` using one staking contract and later staked to another staking contract using liquid tokens and the same operator `A`. In this scenario, we need to make sure grant ID is captured only by the first staking contract, and the second staking contract identifies the stake as coming from staker's liquid tokens.

nkuba:
For each of the possible ways of staking (liquid tokens, grants and managed grants) I tested scenarios covering i.a.: 
- delegation to the old token staking,
- copying stake to the new token staking,
- undelegation+revoke+withdraw from the old token staking
- repayment of the `StakingPortBacker`
- undelegation+revoke from the new token staking contract in two ways: when the new stake was repaid and the new stake wasn't repaid
- tokens flow tracking for the scenarios listed above

I prepared test env in the following way:
1. Deployed contracts from v1.2.1.
2. Deployed new contracts from `bridge` branch: `TokenStaking`, `TokenStakingEscrow` and `StakingPortBacker` configuring them with previously deployed `KeepToken`, `TokenGrant` and `TokenStaking`.

The testing scripts were uploaded [here](https://github.com/keep-network/local-setup/tree/wip-test-staking-bridge/e2e/wip/bridge-test).

During system tests we should focus on mixing the scenarios even more with all the pieces combined: escrow, tops-ups, redelegations, bridging old/new token staking...

solidity/test/token_stake/TestStakingPortBacker.js
```diff
@@ -0,0 +1,635 @@
...
+        ),
+        "Not authorized to pay back"
+      )
+      await expectRevert(
+        token.approveAndCall(
+          stakingPortBacker.address, delegatedAmount, data, {from: thirdParty}
+        ),
+        "Not authorized to pay back"
+      )
+    })
```
nkuba, line 517: What should happen when `receiveApproval` is called directly (not via `token.approveAndCall`)? Do we want to add such test case?

solidity/test/token_stake/TestStakingPortBacker.js
```diff
@@ -0,0 +1,635 @@
...
+
+    it("allows owner to withdraw tokens", async () => {
+      const balanceBefore = await token.balanceOf(owner)
+      await stakingPortBacker.withdraw(9999, {from: owner})
+      const balanceAfter = await token.balanceOf(owner)
+
+      expect(balanceAfter.sub(balanceBefore)).to.eq.BN(9999)
+    })
```
nkuba, line 534: Do we want to add a case for withdraw where the balance is insufficient?

solidity/test/token_stake/TestStakingPortBacker.js
```diff
@@ -0,0 +1,635 @@
...
+      const data = web3.eth.abi.encodeParameters(["address"], [operatorOne])
+      await expectRevert(
+        token.approveAndCall(
+          stakingPortBacker.address, delegatedAmount.addn(1), data, {from: tokenOwner}
+        ),
+        "Unexpected amount"
+      )  
+    })
+
+    it("fails for operator with no tokens backed", async () => {
```
nkuba, line 439: Can we add a case where an operator's token balance is less then the stake?

solidity/test/token_stake/TestMultipleStakingContracts.js
```diff
@@ -0,0 +1,252 @@
...
+      await oldTokenStaking.undelegate(operator, {from: grantee})
+      await newTokenStaking.undelegate(operator, {from: tokenOwner})
+      
+      const undelegationPeriod = await newTokenStaking.undelegationPeriod()
+      await time.increase(undelegationPeriod.addn(1))
+
+      const tokenOwnerBalanceBefore = await token.balanceOf(tokenOwner)
+      const granteeBalanceBefore = await token.balanceOf(grantee)
+      oldTokenStaking.recoverStake(operator, {from: grantee})
+      newTokenStaking.recoverStake(operator, {from: tokenOwner})
```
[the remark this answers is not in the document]
pdyraga: Yes, we do need! Good catch.

solidity/test/token_stake/TestStakingPortBacker.js
pdyraga: I don't think it's a problem with this contract, I don't see any scenario for exploit here. 🤔 I would leave it as-is, for the sake of less complexity.

solidity/test/token_stake/TestStakingPortBacker.js
pdyraga: Hmm... I don't know, it seems to be covered by ERC20 implementation, right?

solidity/test/token_stake/TestStakingPortBacker.js
pdyraga: Sure, good idea! Added in 67e6349.

solidity/test/token_stake/TestMultipleStakingContracts.js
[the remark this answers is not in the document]
pdyraga: Done in 21fd77d.

PR #1907, opened 2020-07-10
Fix a ledger bug

r-czajkowski:
This commit fixes the `MULTIPLE_OPEN_CONNECTIONS_DISALLOWED` error. The problem was that the user could only send one transaction after connected to wallet.


dimpar:
Tested a couple of delegation/undelagations on physical Ledger. Consecutive transactions went through.

Issue #1896, opened 2020-07-06
Consider increasing gas price ceiling


pdyraga:
The gas price ceiling in keep random beacon operator contract is set to 30Gwei. This value proved to be not enough in cases when the gas price skyrockets on mainnet for a couple of hours (sometimes more than 24!). We should consider increasing the gas price ceiling keeping in mind its impact on the price of a relay entry, DKG frequency, and how does it impact staker yields.

pdyraga:
Relay entry fee estimate with no callback price difference:

- For `50 Gwei` ceiling, entry fee is `78970000000000000 wei = 0.07897 ETH`
- For `40 Gwei` ceiling, entry fee is `75976000000000000 wei = 0.075976 ETH`
- For `30 Gwei` ceiling, entry fee is `72982000000000000 wei = 0.072982 ETH`

Analyzing the [historical data](https://docs.google.com/spreadsheets/d/18gXkpBe1Qfpn6vaZ5jIfiMo8zfjvzXJb3-3e42V9FWU/edit#gid=911342864) from https://etherscan.io/chart/gasprice, the gas price values are as follows:
- `13/7/2019 - 13/7/2020` median is `14364979897 wei`
- `13/7/2019 - 13/7/2020` average is `22031400877 wei`
- `1/1/2020 - 13/7/2020` median is `13439926391 wei`
- `1/1/2020 - 13/7/2020` average is `27560064651 wei`
- `1/7/2020 - 13/7/2020` median is `40907231924 wei`
- `1/7/2020 - 13/7/2020` average is `41228001120 wei`

Increasing the ceiling to `40Gwei` is a must and increasing it to `50Gwei` is also reasonable. This decision affects the frequency of DKG and the impact on group creation should be also analyzed.

pdyraga:
Looking at the impact of increasing gas price ceiling on DKG frequency having in mind https://github.com/keep-network/keep-core/pull/1741:

- `30 Gwei` gas price ceiling, `20 Gwei` actual gas price
  Assuming the actual gas price of `20 Gwei` , operator spends
  `(30000000000 * 200000) + (20000000000 * 1740000) = 40800000000000000`
  on triggering group selection and DKG result submission while it received
  `(200000 + 1740000) * 30000000000 = 58200000000000000` for that.

[...]

PR #1914, opened 2020-07-14
Increased gas price ceiling to 60Gwei

pdyraga:
Closes: #1896 

Relay entry fee estimate with no callback price difference:

- For `60 Gwei` ceiling, entry fee is `81964000000000000 wei = 0.081964 ETH`
- For `50 Gwei` ceiling, entry fee is `78970000000000000 wei = 0.07897 ETH`
- For `40 Gwei` ceiling, entry fee is `75976000000000000 wei = 0.075976 ETH`
- For `30 Gwei` ceiling, entry fee is `72982000000000000 wei = 0.072982 ETH`

[Analyzing the historical data](https://docs.google.com/spreadsheets/d/18gXkpBe1Qfpn6vaZ5jIfiMo8zfjvzXJb3-3e42V9FWU/edit#gid=911342864) from https://etherscan.io/chart/gasprice, the gas price values are as follows:
- `13/7/2019 - 13/7/2020` median is `14364979897 wei`
- `13/7/2019 - 13/7/2020` average is `22031400877 wei`
- `1/1/2020 - 13/7/2020` median is `13439926391 wei`
- `1/1/2020 - 13/7/2020` average is `27560064651 wei`
- `1/7/2020 - 13/7/2020` median is `40907231924 wei`
- `1/7/2020 - 13/7/2020` average is `41228001120 wei`

Increasing the ceiling to `40 Gwei` is a must, increasing it to `50 Gwei` is reasonable, and increasing it even further to `60 Gwei` is even more reasonable given the current trends on Ethereum.

The impact on DKG with 1% contribution margin as it's now:
- `30 Gwei` ceiling, `20 Gwei` actual price: DKG triggered every `72` entries
- `30 Gwei` ceiling, `10 Gwei` actual price: DKG triggered every `42` entries
- `40 Gwei` ceiling, `20 Gwei` actual price: DKG triggered every `54` entries
- `40 Gwei` ceiling, `10 Gwei` actual price: DKG triggered every `34` entries
- `50 Gwei` ceiling, `20 Gwei` actual price: DKG triggered every `48` entries
- `50 Gwei` ceiling, `10 Gwei` actual price: DKG triggered every `30` entries
- `60 Gwei` ceiling, `20 Gwei` actual price: DKG triggered every `42` entries
- `60 Gwei` ceiling, `10 Gwei` actual price: DKG triggered every `27` entries
- `60 Gwei` ceiling, `1 Gwei` actual price: DKG triggered every `13` entries

Detailed breakdown for DKG available at https://github.com/keep-network/keep-core/issues/1896

### DIFF
--- a/solidity/test/random_beacon_service/TestRelayRequestCallback.js
+++ b/solidity/test/random_beacon_service/TestRelayRequestCallback.js
@@ -128,7 +128,7 @@ describe('KeepRandomBeacon/RelayRequestCallback', function() {
     let beneficiaryStartBalance = web3.utils.toBN(await web3.eth.getBalance(beneficiary));
 
     // use the same gas price as the gas price ceiling
-    let relayEntryTxGasPrice = web3.utils.toBN(web3.utils.toWei('50', 'Gwei'));
+    let relayEntryTxGasPrice = web3.utils.toBN(web3.utils.toWei('60', 'Gwei'));
     await operatorContract.relayEntry(blsData.groupSignature, {
       from: operator, 
       gasPrice: relayEntryTxGasPrice
@@ -200,7 +200,7 @@ describe('KeepRandomBeacon/RelayRequestCallback', function() {
     );
 
     // use the same gas price as the gas price ceiling
-    let relayEntryTxGasPrice = web3.utils.toBN(web3.utils.toWei('50', 'Gwei'));
+    let relayEntryTxGasPrice = web3.utils.toBN(web3.utils.toWei('60', 'Gwei'));
 
     let customerStartBalance = web3.utils.toBN(await web3.eth.getBalance(customer));
     let beneficiaryStartBalance = web3.utils.toBN(await web3.eth.getBalance(beneficiary));
@@ -232,10 +232,10 @@ describe('KeepRandomBeacon/RelayRequestCallback', function() {
     );
 
     // use higher price than the gas price ceiling
-    let relayEntryTxGasPrice = web3.utils.toBN(web3.utils.toWei('50', 'Gwei'));
+    let relayEntryTxGasPrice = web3.utils.toBN(web3.utils.toWei('70', 'Gwei'));
     // higher tx.gasprice should not be used for reimbursement - maximum gas
     // price is the one from the gas price ceiling
-    let gasPriceForReimbursement = web3.utils.toBN(web3.utils.toWei('50', 'Gwei'));
+    let gasPriceForReimbursement = web3.utils.toBN(web3.utils.toWei('60', 'Gwei'));
 
     let customerStartBalance = web3.utils.toBN(await web3.eth.getBalance(customer));
     let beneficiaryStartBalance = web3.utils.toBN(await web3.eth.getBalance(beneficiary));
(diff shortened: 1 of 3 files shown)


pdyraga:
> I would consider even 60 gwei as the price, to stay ahead of the gas price growth trend. Short peaks should be handled by clients submitting transactions at a greater price if necessary and eating the difference, but if the increase in prices continues we don't want our reimbursements to get completely buried under it.

It sounds reasonable to me. I put detailed calculations [in the issue](https://github.com/keep-network/keep-core/issues/1896#issuecomment-659393292), and summarizing them:

- For `60 Gwei` ceiling, entry fee (with no callback) is `81964000000000000 wei = 0.081964 ETH`
- `60 Gwei` gas price ceiling, `20 Gwei` actual gas price: DKG is triggered every 42 entries.
- `60 Gwei` gas price ceiling, `10 Gwei` actual gas price: DKG is triggered every 27 entries.
- `60 Gwei` gas price ceiling, `1 Gwei` actual gas price: DKG is triggered every 13 entries.

pdyraga:
If we go from `30Gwei` to `60Gwei` ceiling we'll increase base (no callback) entry price by `12.3%`: `0.072982 ETH -> 0.081964 ETH` and we'll increase [the fee for opening a keep](https://github.com/keep-network/keep-ecdsa/blob/08f6557bb69873709553c8c7a2f3ef798ce59777/solidity/contracts/BondedECDSAKeepFactory.sol#L430-L432) by `13.73%`: `0.07418878 ETH -> 0.08437756`.

Given the gas price trends, I tend to agree with @eth-r's proposition of going even further and increase to 60 Gwei.

mhluongo:
The `13.73%` is unfortunate for tBTC, but this just screams "replace the mechanism" to me. I'm for biting the bullet now and going back to the pricing drawing board after this release

solidity/test/random_beacon_service/TestRelayRequestCallback.js
```diff
@@ -128,7 +128,7 @@ describe('KeepRandomBeacon/RelayRequestCallback', function() {
     let beneficiaryStartBalance = web3.utils.toBN(await web3.eth.getBalance(beneficiary));
 
     // use the same gas price as the gas price ceiling
-    let relayEntryTxGasPrice = web3.utils.toBN(web3.utils.toWei('30', 'Gwei'));
+    let relayEntryTxGasPrice = web3.utils.toBN(web3.utils.toWei('60', 'Gwei'));
```
dimpar, line 131: What do you think about reading `gasPriceCeiling` from the contract, defining it in `before` block and then reuse it in the tests? If later on we decide to change `gasPriceCeiling`, we won't need to update it, only gas prices around it.
```
let gasPriceCeiling;
before(async () => {
...
gasPriceCeiling = await operatorContract.gasPriceCeiling()
...
}

exampleTest {
// use the same gas price as the gas price ceiling
let relayEntryTxGasPrice = web3.utils.toBN(gasPriceCeiling);
}
```

pdyraga: I was thinking about it but at the same time, I like the tests are checking a particular value, especially that we have tests using a higher and lower value. If someone makes a mistake and change value in the contract accidentally, tests will fail - I like it. This way we are asserting a specific expected behavior.

dimpar: right, makes sense.

PR #1915, opened 2020-07-15
Undelegation period set to two weeks for the first two months after TokenStaking deployment

pdyraga:
Closes: #1900 
~~Depends on #1906 (I had to make sure we are not running into out-of-gas problems again but it seems that #1902 saved us)~~

For the first two months after the deployment of `TokenStaking` contract, the undelegation period is two weeks and it goes back to two months once two months after the deployment passes.

The staking contract guarantees that an undelegated operator’s stakes will stay locked for a period of time after undelegation, and thus available as collateral for any work the operator is engaged in. Two weeks is the absolute minimum given that the beacon group's active time is also two weeks. Changing undelegation period does not affect `keep-ecdsa` deposits since the stake is locked by the sanctioned application until all deposits the given staker has are redeemed and keeps are closed.

Shorten undelegation period should allow less stressful onboarding of new stakers to the system as well as allow for faster migration in case a new staking contract is deployed.

### DIFF
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -66,7 +66,6 @@ contract TokenStaking is Authorizations, StakeDelegatable {
 
     uint256 public minimumStakeScheduleStart;
     uint256 public initializationPeriod; // varies between mainnet and testnet
-    uint256 public constant undelegationPeriod = 5184000; // ~60 days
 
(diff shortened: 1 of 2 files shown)


pdyraga:
@mhluongo I need an adult to review the numbers: 2 weeks for the first 2 months and 2 months after that?

PR #1919, opened 2020-07-16
Fix for not presenting slashes

ElderOrb:
The issue could result in not presenting information about slashing / seizing in a dashboard under some specific conditions (when transaction hash is neither in slashedTokensGroupedByTxtHash nor in seizedTokensGroupedByTxtHash). As the result user would observe correct total number of staked tokens with no explanation why it is lesser than used to be. 

Actual fix consists of two parts: 
- skipping uninitialized punishmentData-s 
- properly filling 'amount' (unrelated to my original issue but might result in similar behaviour if not fixed too)

@pdyraga @sthompson22 - it resolved my issue with 'missing 1k & no slashes'

### DIFF
--- a/solidity/dashboard/src/services/slashed-tokens.service.js
+++ b/solidity/dashboard/src/services/slashed-tokens.service.js
@@ -112,7 +112,10 @@ const groupByTransactionHash = (events) => {
         amount: add(returnValues.amount, prevData.amount),
       }
     } else {
-      groupedByTransactionHash[transactionHash] = { ...returnValues }
+      groupedByTransactionHash[transactionHash] = {
+        ...returnValues,
+        amount: returnValues.amount,
+      }
     }
   })
 
--- a/solidity/dashboard/src/services/slashed-tokens.service.js
+++ b/solidity/dashboard/src/services/slashed-tokens.service.js
@@ -78,6 +78,7 @@ const fetchSlashedTokens = async (web3Context) => {
       punishmentData = { amount, type: "SEIZED", event }
     }
 
+    if (!punishmentData.hasOwnProperty("amount")) continue
     if (lte(punishmentData.amount, 0)) continue
 
     punishmentData.date = moment.unix(
--- a/solidity/dashboard/src/services/slashed-tokens.service.js
+++ b/solidity/dashboard/src/services/slashed-tokens.service.js
@@ -69,7 +69,7 @@ const fetchSlashedTokens = async (web3Context) => {
       blockNumber,
       returnValues: { groupIndex },
     } = punishmentEvents[i]
-    let punishmentData = {}
+    let punishmentData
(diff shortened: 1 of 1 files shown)


solidity/dashboard/src/services/slashed-tokens.service.js
```diff
@@ -78,6 +78,7 @@ const fetchSlashedTokens = async (web3Context) => {
       punishmentData = { amount, type: "SEIZED", event }
     }
 
+    if (!punishmentData.hasOwnProperty("amount")) continue
```
r-czajkowski, line 81: I think the problem is in the `groupByTransactionHash` function. We missed the `amount` field in [`groupedByTransactionHash[transactionHash] = { ...returnValues }`](https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/services/slashed-tokens.service.js#L114).

Should be `groupedByTransactionHash[transactionHash] = { ...returnValues, amount: returnValues.amount }`

ElderOrb: Fixed that place too. But in my case issue seems to be different: transaction '0x7f71b6dda7239fb70e0b97cf214176185b23c29162fb5f33cf6e8f5b644071e0' is neither in slashedTokensGroupedByTxtHash nor in seizedTokensGroupedByTxtHash - so results in empty punishmentData, which causes an error without extra check I initially added.

r-czajkowski: Right! I see. But maybe let's just declare `let punishmentData` instead of `let punishmentData = {}` and in if statement
`if (!punishmentData || lte(punishmentData.amount, 0)) continue`

ElderOrb: sure, should be more clean / efficient

ElderOrb:
> Looks good, thanks! Waiting for a green light and I'm ready to merge.

Speaking about green light... Is there anything I should configure on my side to prevent circle ci from failing on my PRs?

pdyraga:
> Speaking about green light... Is there anything I should configure on my side to prevent circle ci from failing on my PRs?

Some Circle jobs require credentials for publishing resources and we can't make them public. For this reason, they need to be triggered by a project member. We usually do it manually at the end of the review process for simple PRs.

r-czajkowski:
Seems like your commits are not signed. Could you please fix it? We are close to merging 😄

ElderOrb:
> Seems like your commits are not signed. Could you please fix it? We are close to merging 😄

Omg, again? :))) Weird, I've configured signing (although maybe not on all the environments)

ElderOrb:
finally done

pdyraga:
Can you please summarize in the PR description what was the issue and what's the fix? Thank you!

ElderOrb:
> Can you please summarize in the PR description what was the issue and what's the fix? Thank you!

Done, hope its a bit more clear now.

PR #1922, opened 2020-07-16
Check if the accused member ID is non-zero

pdyraga:
In the resolve secret shares accusation phase and resolve public key share points accusations phase, we were checking whether the accused member ID does not exceed the group size. We were not, however, checking if it's non-zero which could result in returning an error from GJKR protocol.

### DIFF
--- a/pkg/beacon/relay/gjkr/protocol.go
+++ b/pkg/beacon/relay/gjkr/protocol.go
@@ -549,7 +549,8 @@ func (sjm *SharesJustifyingMember) ResolveSecretSharesAccusationsMessages(
 	for _, message := range messages {
 		accuserID := message.senderID
 		for accusedID, revealedAccuserPrivateKey := range message.accusedMembersKeys {
-			if sjm.ID == accusedID || int(accusedID) > sjm.group.GroupSize() {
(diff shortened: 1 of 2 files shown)


pkg/beacon/relay/gjkr/protocol.go
```diff
@@ -549,7 +549,8 @@ func (sjm *SharesJustifyingMember) ResolveSecretSharesAccusationsMessages(
 	for _, message := range messages {
 		accuserID := message.senderID
 		for accusedID, revealedAccuserPrivateKey := range message.accusedMembersKeys {
-			if sjm.ID == accusedID || int(accusedID) > sjm.group.GroupSize() {
+			isAccusedValid := accusedID > 0 && int(accusedID) <= sjm.group.GroupSize()
```
dimpar, line 552: can we add `id`? `isAccusedIdValid`

pdyraga: Yeah, good idea but it has to be `isAccusedIDValid` then because Go linter complains about `Id` vs `ID` (:untada:)

dimpar:
LGTM, if @lukasz-zimnoch has no comments, I think we can merge.

PR #1928, opened 2020-07-20
Update KEEP token dashboard for TokenStakingEscrow

r-czajkowski:
This PR updates a dapp for `TokenStakingEscrow` contract(see #1869). If the user recovers/cancels delegation, tokens will be deposited in the `TokenStakingEscrow` contract. To calculate the correct grant values such as: `withdraw`, `staked`, `withdrawable` we have to take into account deposits in a `TokenStakingEscrow` contract. The user can withdraw deposited tokens in TokenStakingEscrow via the `Release tokens` button in the dapp, but it may be necessary to confirm more than one transaction.


dimpar:
Granted 1M 
|-> delegated 200k
|-> delegated 300k

Total balance seems to be incorrect, but hopefully merging master and resolving the conflict should help, since we noticed it before.

Screenshot 2020-07-24 at 10 57 01

dimpar:
Shouldn't we update `Grants` back to 1.3M. ?

Screenshot 2020-07-24 at 11 47 52

liz-shinn:
>Shouldn't we update Grants back to 1.3M. ?
@dimpar @r-czajkowski @pdyraga 

There was a UX concern brought up about this in Flowdock. In the example we're showing in these screenshots, it's confusing why the `Grants` amount at the top of the page is a smaller amount (1.2 million) than the number shown in an individual grant (in this case, 2 million). I understand why it is (one shows stakeahead etc) but it's confusing.

I would recommend that the number under `Grants` display the total amount of tokens a user holds across all of their grants (regardless of how much has been staked).

Reason being, since `Grants` is at the top of the page and is the section header, you would expect the token number shown to be the largest number. Once you start breaking out into smaller sections of the larger whole (for instance, viewing 1 grant and the Grant Details panel) that's when you would expect to see granular amounts / information.

dimpar:
Makes sense to me. Probably not very difficult to fix, but not sure if Rafał would rather fix it here or we would handle it in a separate PR.

liz-shinn:
Ok! Also, meant to say that the same should hold true for the Owned token balance. The number shown by `Owned Tokens ` at the top should also be the total amount of tokens irrespective of how many are already staked.

dimpar:
> Ok! Also, meant to say that the same should hold true for the Owned token balance. The number shown by Owned Tokens at the top should also be the total amount of tokens irrespective of how many are already staked.

Screenshot 2020-07-24 at 14 49 37

Just to make sure, top right corner should display in this case 1.4M?

solidity/dashboard/src/reducers/tokens-page.reducer.js
```diff
@@ -116,6 +123,40 @@ const removeFromDelegationOrUndelegation = (array, id) => {
...
+    availableToWitdrawGrant,
+    operator,
+  }
+) => {
+  const { indexInArray, obj: grantToUpdate } = findIndexAndObject(
+    "id",
+    grantId,
+    grants
+  )
+  if (indexInArray === null) {
```
dimpar, line 141: `if (!indexInArray)`

solidity/dashboard/src/services/token-grants.service.js
```diff
@@ -53,6 +59,26 @@ const getGrantDetails = async (
...
+  for (const event of escrowDepositsEvents) {
+    const {
+      returnValues: { operator },
+    } = event
+    const withdrawable = await fetchWithdrawableAmountForDeposit(operator)
+    const withdraw = await fetchDepositWithdrawnAmount(operator)
+
+    escrowWithdrawTotalAmount = add(escrowWithdrawTotalAmount, withdraw)
+
+    if (gt(withdrawable, 0)) {
```
dimpar, line 76: wouldn't it be enough to `if(withdrawable)`? I think `falsy` result handles 0

solidity/dashboard/src/services/token-grants.service.js
```diff
@@ -77,18 +103,22 @@ const getGrantDetails = async (
     "unlockedAmount",
     grantId
   )
-  let readyToRelease = "0"
+  let withdrawableAmountGrantOnly = "0"
   try {
```
dimpar, line 107: wonder if we need try catch block here. In other places we do not use it. Did we add it here for a specific reason?

solidity/dashboard/src/contexts/TokensPageContext.jsx
```diff
@@ -158,14 +160,34 @@ const TokenPageContextProvider = (props) => {
...
   )
 
+  const grantDeposited = useCallback(async (grantId, operator, amount) => {
+    const availableToWithdrawEscrow = await contracts.tokenStakingEscrow.methods
+      .withdrawable(operator)
+      .call()
+    const availableToWitdrawGrant = await contracts.grantContract.methods
+      .withdrawable(grantId)
+      .call()
+    console.log("asdasdas", availableToWithdrawEscrow, availableToWitdrawGrant)
```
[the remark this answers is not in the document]
pdyraga: nike

liz-shinn:
>Just to make sure, top right corner should display in this case 1.4M?

@dimpar 
Shoot... I think we might need some more clarity in that Owned Tokens details section... I think a breakdown of that number would make sense. The mint green number is the Total Amount in both the top right and in the Details section of Owned. Then beneath it we would have additional breakout of Staked: Amount and Unstaked: Amount.

It's repeating what the Staked visual is showing beneath it, but I think the repetition will help reinforce what that mint green number represents.

solidity/dashboard/src/contexts/TokensPageContext.jsx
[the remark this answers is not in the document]
r-czajkowski: reebok. [Are those reebok or nike?](https://www.youtube.com/watch?v=BQ4c54rCJ_k) 🤔

solidity/dashboard/src/services/token-grants.service.js
r-czajkowski: We probably don't need this anymore, I'll double-check. We added `try catch` block to catch the `SafeMath underflow` bug on the smart contract side that was addressed with https://github.com/keep-network/keep-core/pull/1464.

solidity/dashboard/src/services/token-grants.service.js
r-czajkowski: `withdrawable` is `string` so `if(withdrawable)` where `withdrawable -> "0"` it's a non-empty string and evaluated to true.
I think we can just `if(withdrawable !== "0")`, what do you think?

solidity/dashboard/src/components/TokenGrantOverview.jsx
```diff
@@ -78,23 +82,114 @@ export const TokenGrantUnlockingdDetails = ({
...
-  const { yourAddress, grantContract } = useContext(Web3Context)
+  const { yourAddress, grantContract, tokenStakingEscrow } = useContext(
+    Web3Context
+  )
   const showMessage = useShowMessage()
+  const closeMessage = useCloseMessage()
 
-  const releaseTokens = async (onTransactionHashCallback) => {
+  const releaseTokens = async (
+    onTransactionHashCallback,
```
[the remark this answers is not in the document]
r-czajkowski: We use `openConfirmationModal`. I've refactored the modal approach in the dapp a bit in https://github.com/keep-network/keep-core/commit/e47b03253e541a375c43062a3db2d50ea220d3c7 , so it will probably fix this.

solidity/dashboard/src/services/token-grants.service.js
dimpar: oh, I didn't realize that `withdrawable` is a string, `0` in `gt(withdrawable, 0)` confused me. Anyway, I think it's fine then, let's leave it as is.

dimpar:
Screenshot 2020-07-27 at 14 52 32

Delegated 200k and then immediately canceled. Shouldn't we display 1M in the left top corner?

dimpar:
Screenshot 2020-07-27 at 15 26 35

There were 2 grants: 1M + 4M

Total Balance seems to be incorrect.

dimpar:
Screenshot 2020-07-27 at 15 37 09

Not sure if 4500000 is correct in the top left corner. @liz-shinn @r-czajkowski ?

liz-shinn:
> Not sure if 4500000 is correct in the top left corner. @liz-shinn @r-czajkowski ?

Yeah that should be 5,000,000 to match the total of the two token grants.

dimpar:
Testing: 
- created 2 Grants of 1M and 4M
- functionality tested for multiple: delegated, undelegated, canceled, recovered and released tokens. Everything seem to add up together.

Issue #1911, opened 2020-07-13
Balance reported on new overview page


sthompson22:
### Background

I believe this was introduced in  [this](https://github.com/keep-network/keep-core/pull/1847) PR.  Last week we deployed  an updated token-dashboard against Ropsten, from master.  Last fully functional version of the token dashboard deployed  was  https://github.com/keep-network/keep-core/releases/tag/v1.2.4-rc which did not include the the overview  page.

The token overview balance appears to be adding total grant value + delegated value from grant to produce total balance.

tl;dr think we're looking for some changes between and https://github.com/keep-network/keep-core/releases/tag/v1.2.4-rc and https://github.com/keep-network/keep-core/tree/sthompson22/token-dash/use-tbtc-v1.0.3-rc

### How the issue was surfaced

Using an account with a fresh 300k token grant from the  faucet  I was able to  observe the 300k  grant  balance in the overview page.  I did a 100k KEEP delegation which was  successful, however the overview page  was showing a 400k KEEP balance now.
- [Branch](https://github.com/keep-network/keep-core/tree/sthompson22/token-dash/use-tbtc-v1.0.3-rc) for deployment where issue was found.

I reproduced this by using a different account, getting  a 300k grant and delegating all 300k.  The result in the overview page was a token balance of 600k displayed.

### How to reproduce

1.  get a grant
2. check  balance in overview page
3. delegate some stake
4. check balance in overview page

PR #1933, opened 2020-07-20
Fix total balance of grants- KEEP token dashboard overview page

r-czajkowski:
Closes: #1911 

This PR fixes a bug with a total balance of grants on the overview page.

Before:
* total balance of grants- `1 000 000` KEEP,
* delegate `100 000` KEEP,
* total balance of grants on the overview page- `1 100 000` KEEP- which was not correct.

Currently: 
* total balance of grants- `1 000 000` KEEP,
* delegate `100 000` KEEP,
* we are able to observe `1 000 000` KEEP and `10%` Staked in `Granted Tokens` card.

### DIFF
--- a/solidity/dashboard/src/components/ChooseWallet.jsx
+++ b/solidity/dashboard/src/components/ChooseWallet.jsx
@@ -130,7 +130,7 @@ const Wallet = ({
         {renderModalContent()}
       </ModalComponent>
       <li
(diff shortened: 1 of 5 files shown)


dimpar:
Testing:
Created 2 grants, delegated some amount from each and then released the tokens which became "owned". Math adds up.
Screenshot 2020-07-21 at 14 09 19

PR #1939, opened 2020-07-23
Add ubuntu-setup.sh

lukasz-zimnoch:
Refs: https://github.com/keep-network/local-setup/pull/12

During the work on the aforementioned pull request, we developed some Ubuntu environment provisioning scripts. Here we introduce them as `ubuntu-setup.sh` as an alternative for `macos-setup.sh`.


pdyraga:
@mhluongo will be very happy about this PR 😆

Issue #1898, opened 2020-07-06
Update KEEP token dashboard for top-ups


pdyraga:
https://github.com/keep-network/keep-core/pull/1893 introduces the ability for staker to top-up the existing delegation. Adding tokens to existing delegation is a two-step process: first, staker initiates a top-up, locking new KEEPs in the staking contract. Once the initialization period is over, staker (or anyone else) can commit the top-up, increasing operator's stake.

There is one limitation. We expect the same source of tokens for a top-up as used for the initial delegation. If the initial delegation was done from a grant, the same grant has to be used for a top-up to the operator. If the initial delegation was done using owner's liquid tokens, liquid tokens from the same owner are required for a top-up.

An implication of this change is that tokens staked from a grant, undelegated and recovered does no longer come back to the grant contract. Instead, they are deposited in a special escrow contract from which they can be redelegated or withdrawn, according to grant unlocking schedule.

We need to update KEEP token dashboard to reflect this change.

pdyraga:
Another important requirement is that all calls should now go directly to `TokenStaking` contract.

liz-shinn:
Notes from call on this today in Coda 👍 
<img width="659" alt="Screen Shot 2020-07-08 at 11 12 23 AM" src="https://user-images.githubusercontent.com/57226633/86943922-09233e00-c115-11ea-8690-bd3074375e94.png">

pdyraga:
Posting some more thoughts about it. It's absolutely not a complete solution, just something we can iterate on.

For top-ups, we can have the following flow:

In *Delegate Tokens* section, we move the *Operator Address* input to be the first one. For a delegation, everything works just like now. When the user inputs operator address for existing delegation, *Authorizer Address* and *Beneficiary Address* are filled automatically and disabled for editing. *Delegate Stake* button turns into *Initiate Stake Top-Up*. 

Alternatively, in *Delegations* table below we can add another button, next to *Undelegate*, that will initiate a top-up. We need to display some form, so we could take the user to *Delegate Tokens* section with fields pre-filled as proposed earlier.

Initiated top-ups can be displayed in a table below, next to *Delegations* and *Undelegations*. Each initiated top-up needs to have one button, allowing to commit the top-up as soon as the initialization period passes.

If someone wants to add more tokens to an initializing top-up, we can solve it with a modal informing about the fact the timer will reset if they do that.

[...]

PR #1945, opened 2020-07-27
Top ups in dapp

r-czajkowski:
Ref: https://github.com/keep-network/keep-core/issues/1898

This PR adds support for top-ups in a keep token dashboard.

<details>
  <summary>Screenshoots</summary>

obraz
obraz
obraz

</details>

### DIFF
--- a/solidity/dashboard/src/hooks/useFetchData.js
+++ b/solidity/dashboard/src/hooks/useFetchData.js
@@ -1,4 +1,4 @@
-import { useEffect, useReducer, useContext, useCallback } from "react"
+import { useEffect, useReducer, useContext, useCallback, useState } from "react"
 import { Web3Context } from "../components/WithWeb3Context"
 import { wait } from "../utils/general.utils"
 import { usePrevious } from "./usePrevious"
@@ -15,8 +15,11 @@ const requestTimeDelay = 500 // 0.5s
 export const useFetchData = (
   serviceMethod,
   initialData,
-  ...serviceMethodArgs
+  ...initialServiceMethodArgs
 ) => {
+  const [serviceMethodArgs, setServiceMethodArgs] = useState(
+    initialServiceMethodArgs
+  )
   const web3Context = useContext(Web3Context)
   const [state, dispatch] = useReducer(dataFetchReducer, {
     isFetching: true,
@@ -40,6 +43,7 @@ export const useFetchData = (
           dispatch({ type: FETCH_REQUEST_SUCCESS, payload: data })
       })
       .catch((error) => {
+        console.log("error", error)
         shouldSetState &&
           dispatch({ type: FETCH_REQUEST_FAILURE, payload: error })
       })
@@ -49,7 +53,7 @@ export const useFetchData = (
     }
   }
 
-  useEffect(fetchData, [])
+  useEffect(fetchData, [serviceMethodArgs])
   useEffect(() => {
     if (
       prevSyncState === syncState.UP_TO_DATE &&
@@ -67,7 +71,7 @@ export const useFetchData = (
     dispatch({ type: REFRESH_DATA })
   }, [])
 
-  return [state, updateData, refreshData]
+  return [state, updateData, refreshData, setServiceMethodArgs]
 }
 
 const dataFetchReducer = (state, action) => {
--- a/solidity/dashboard/src/hooks/useFetchData.js
+++ b/solidity/dashboard/src/hooks/useFetchData.js
@@ -43,7 +43,6 @@ export const useFetchData = (
           dispatch({ type: FETCH_REQUEST_SUCCESS, payload: data })
       })
       .catch((error) => {
-        console.log("error", error)
         shouldSetState &&
           dispatch({ type: FETCH_REQUEST_FAILURE, payload: error })
       })
--- a/solidity/dashboard/src/services/top-ups.service.js
+++ b/solidity/dashboard/src/services/top-ups.service.js
@@ -1,8 +1,75 @@
-import { contracts } from "../contracts"
+import { contracts, CONTRACT_DEPLOY_BLOCK_NUMBER } from "../contracts"
+import { isEmptyArray } from "../utils/array.utils"
+import { add } from "../utils/arithmetics.utils"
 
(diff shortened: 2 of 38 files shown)


dimpar:
Switching between Metamask accounts doesn't work properly. Upon selection I can see the below error and it will switch only after page reloads.

Screenshot 2020-07-30 at 11 33 26

dimpar:
I think one of the things that can be improved a bit is disabling `commit top-up` or `add keep` button until the initialization period is over. As for now they are still clickable but MetaMask will reject that transaction.

dimpar:
I don't think we need "Clickable finger palm" over Current / New 

Screenshot 2020-07-30 at 11 57 14

r-czajkowski:
> I think one of the things that can be improved a bit is disabling `commit top-up` or `add keep` button until the initialization period is over. As for now they are still clickable but MetaMask will reject that transaction.

Yeah, nice catch!

r-czajkowski:
> I don't think we need "Clickable finger palm" over Current / New
> 
> Screenshot 2020-07-30 at 11 57 14

Implemented based on [KEEP Design system](https://www.figma.com/file/ARAMFLMOPal0JewjvlZGhX/Keep-Design-System?node-id=1168%3A80) 🤔

dimpar:
Yeah, to be clear, I'm not talking about changing colors but rather about "navigation hand / palm" whatever you call it :) Usually you expect to be navigated to somewhere upon click, but in this case it just change color.

https://upload.wikimedia.org/wikipedia/commons/thumb/3/39/Pointing_hand_cursor_vector.svg/1200px-Pointing_hand_cursor_vector.svg.png

r-czajkowski:
> Yeah, to be clear, I'm not talking about changing colors but rather about "navigation hand / palm" whatever you call it :) Usually you expect to be navigated to somewhere upon click, but in this case it just change color.
> 
> https://upload.wikimedia.org/wikipedia/commons/thumb/3/39/Pointing_hand_cursor_vector.svg/1200px-Pointing_hand_cursor_vector.svg.png

yeah, I see 👍

dimpar:
Scenario:
Delegated 100k from Grant 1 to 0xAAA. But right after I canceled that. And then again I've delegated 100k from Grant 1 to 0xAAA and it crashed:

Screenshot 2020-07-30 at 12 19 11

dimpar:
Problem: cannot undelegate tokens

Scenario:
Delegate tokens from Grant to 0xAAA -> Add Keep and Commit-topup. Then Undelegate tokens, but MetaMask won't let you.

Screenshot 2020-07-30 at 13 32 35

dimpar:
Problem: Can't recover undelegated tokens

Scenario: Delegate from Grant -> Undelegate. Then tried Recover but it failed.

Screenshot 2020-07-30 at 13 50 54 (3)

solidity/dashboard/src/hooks/useFetchData.js
```diff
@@ -40,6 +43,7 @@ export const useFetchData = (
           dispatch({ type: FETCH_REQUEST_SUCCESS, payload: data })
       })
       .catch((error) => {
+        console.log("error", error)
```
dimpar, line 46: console.log

solidity/dashboard/src/services/top-ups.service.js
```diff
@@ -0,0 +1,89 @@
...
+
+  return availableTopUps
+}
+
+const reduceByOperator = (result, event) => {
+  const {
+    returnValues: { operator },
+  } = event
+
+  ;(result[operator] = result[operator] || []).push(event)
```
dimpar, line 67: semicolon in a weird plac

solidity/dashboard/src/services/top-ups.service.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const reduceAmount = (result, { returnValues: { topUp } }) => {
+  return add(result, topUp)
+}
+
+const filterByAfterLatestCompletedTopUp = (latestTopUpCompletedEvent) => (
+  initiatedEvent
+) => {
+  const isAfterLatestCompleteedTopUpBlock =
```
dimpar, line 79: typo: isAfterLatestCompleteedTopUpBlock -> isAfterLatestCompletedTopUpBlock

solidity/dashboard/src/services/top-ups.service.js
r-czajkowski: eslint...

pdyraga:
The problems with canceled stake and top-ups are probably related to https://github.com/keep-network/keep-core/pull/1949.

dimpar:
Regarding https://github.com/keep-network/keep-core/pull/1945#issuecomment-666260428

I can switch accounts now just fine, but I still can see the same error in the console.

Also, what I noticed is that the flow changed a bit. Not sure if that was intentional, but upon switching MM accounts it won't navigate me to the wallets selection page (MM, ledger, trezor). Only after the page reload I am able to select a wallet or clicking at the tokenDashboard logo.

dimpar:
Problem: Can't commit top-up after adding tokens from Grants

Scenario 1: After init period is over -> add Keep -> you will see a row in top-ups table below. Click enabled "commit top-up" button. MM will reject you. 

Scenario 1: After init period is over -> add Keep -> you will see a row in top-ups table below with enabled button. Just refresh the page without committing. Button is disabled now.

dimpar:
Regarding: https://github.com/keep-network/keep-core/pull/1945#issuecomment-666274506

In happy path flow:

Delegate tokens (status is pending) -> add keep (available Top-up appear) -> commit top-up. It will reject you. Then wait until the init period is over. Commit top-up button is still disabled.

dimpar:
Problem: During init period it seems that I can skip 'Submit top-up' phase and add keep directly. MM will allow me to do so

dimpar:
Problem: Top-ups "disappear" upon delegation.

Scenario: In this example I've delegated 700k tokens -> added 500k tokens but did NOT commit -> undelegated. In the view below it is sort of uncertain where'd 500k top up went. Only if you click "Recover" then 500k will be added to the total available and everything sums up to 10M

Screenshot 2020-07-31 at 15 05 40

PR #1949, opened 2020-07-29
Do not allow to redelegate to an operator from which a delegation has been cancelled

pdyraga:
After we added a feature allowing to top-up in the initialization period (https://github.com/keep-network/keep-core/pull/1929) we unconsciously introduced a path allowing to redelegate to the operator from which the delegation has been canceled. It should not be allowed as it leads to rejecting the deposit by the escrow when undelegating/canceling that second delegation.

There are two scenarios when top-up can be performed on a canceled delegation:

Scenario 1: Someone tries to re-delegate from the escrow to operator that already has a deposit in the escrow:
```
delegate operator A -> escrow ->
redelegate operator B -> escrow ->
redelegate operator A -> escrow (TX reverted!)
 ```
 This scenario is now prohibited from happening by `TokenStakingEscrow.hasDeposit` check in redelegate function.

Scenario 2: Someone tries to top-up a stake to an operator that has been previously used and from which delegation was canceled.

This scenario is now prohibited from happening because of `TokenStakingEscrow.hasDeposit` check in the `TopUps.executeInOneStep`function.

The problem described here is not a vulnerability when a 3rd party can hurt a staker but rather an interesting edge case when the staker can hurt themselves.

When working on the fix, I had to move some code out of `TokenStaking` to `TopUps` library to avoid out-of-gas issues. The good part about it is that we now have the entire logic but grant rules validation in `TopUps` library.

### DIFF
--- a/solidity/contracts/libraries/staking/TopUps.sol
+++ b/solidity/contracts/libraries/staking/TopUps.sol
@@ -45,7 +45,7 @@ library TopUps {
         address operator,
         uint256 operatorParams,
         TokenStakingEscrow escrow
-    ) public returns (uint256) {
+    ) public returns (uint256 newParams) {
         // Stake is not yet initialized so we don't need to check if the
         // operator is not undelegating - initializing and undelegating at the
         // same time is not possible. We do however, need to check whether the
@@ -55,13 +55,12 @@ library TopUps {
         require(!escrow.hasDeposit(operator), "Stake canceled");
 
         uint256 newAmount = operatorParams.getAmount().add(value);
-        uint256 newParams = operatorParams.setAmountAndCreationTimestamp(
+        newParams = operatorParams.setAmountAndCreationTimestamp(
             newAmount,
             block.timestamp
         );
 
         emit TopUpCompleted(operator, newAmount);
-        return newParams;
     }
 
     /// @notice Initiates top-up of the given value for tokens delegated to
@@ -95,8 +94,9 @@ library TopUps {
     function commit(
         Storage storage self,
         address operator,
+        uint256 operatorParams,
         uint256 initializationPeriod
-    ) public returns (uint256) {
+    ) public returns (uint256 newParams) {
         TopUp memory topUp = self.topUps[operator];
         require(topUp.amount > 0, "No top up to commit");
         require(
@@ -104,16 +104,19 @@ library TopUps {
             "Stake is initializing"
         );
 
+        uint256 newAmount = operatorParams.getAmount().add(topUp.amount);
+        newParams = operatorParams.setAmount(newAmount);
+
         delete self.topUps[operator];
-        return topUp.amount;
+        emit TopUpCompleted(operator, newAmount);
     }
 
-    /// @notice Force-commits existing top-up without taking into account
-    /// stake initialization period. If there is no pending top-up for the
-    /// operator, function does nothing. This function should be used when
-    /// the stake is recovered to return tokens from a pending top-up.
+    /// @notice Cancels pending, initiating top-up. If there is no initiating
+    /// top-up for the operator, function does nothing. This function should be
+    /// used when the stake is recovered to return tokens from a pending,
+    /// initiating top-up.
     /// @param operator Operator The operator from which the stake is recovered.
-    function forceCommit(
+    function cancel(
         Storage storage self,
         address operator
     ) public returns (uint256) {
--- a/solidity/contracts/libraries/staking/TopUps.sol
+++ b/solidity/contracts/libraries/staking/TopUps.sol
@@ -37,6 +37,7 @@ library TopUps {
     /// the tokens should be added to.
     /// @param operatorParams Parameters of that operator, as stored in the
     /// staking contract.
+    /// @param escrow Reference to TokenStakingEscrow contract.
     /// @return New value of parameters. It should be updated for the operator
     /// in the staking contract.
     function executeInOneStep(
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by pdyraga** [comment]
Let's hold on with this PR, it seems there is another path that is problematic.

File: solidity/contracts/libraries/staking/TopUps.sol [hunk 1]
```diff
@@ -20,20 +28,71 @@ library TopUps {
...
+        uint256 operatorParams,
+        TokenStakingEscrow escrow
+    ) public returns (uint256 newParams) {
+        // Stake is not yet initialized so we don't need to check if the
+        // operator is not undelegating - initializing and undelegating at the
+        // same time is not possible. We do however, need to check whether the
+        // operator has not canceled its previous stake for that operator,
+        // depositing the stake it in the escrow. We do not want to allow
+        // resurrecting operators with cancelled stake by top-ups.
+        require(!escrow.hasDeposit(operator), "Stake canceled");
```
nkuba, line 56: We're not checking if the stake is canceled, but if the operator already deposited anything to the escrow. Maybe we should update the message?
```suggestion
        require(!escrow.hasDeposit(operator), "Operator already deposited to the escrow");
```

solidity/contracts/libraries/staking/TopUps.sol
```diff
@@ -20,20 +28,71 @@ library TopUps {
...
+        address operator,
+        uint256 operatorParams,
+        TokenStakingEscrow escrow
     ) public {
+        // Stake is initialized, the operator is still active so we need
+        // to check if it's not undelegating.
+        require(!isUndelegating(operatorParams), "Stake undelegated");
+        // We also need to check if the stake for the operator is not already
+        // in the escrow because it's been previously cancelled.
+        require(!escrow.hasDeposit(operator), "Stake canceled");
```
nkuba, line 91: We're not checking if the stake is canceled, but if the operator already deposited anything to the escrow. Maybe we should update the message?
```suggestion
        require(!escrow.hasDeposit(operator), "Operator already deposited to the escrow");
```

solidity/contracts/libraries/staking/TopUps.sol
```diff
@@ -73,4 +136,15 @@ library TopUps {
...
         return topUp.amount;
     }
+
+    /// @notice Returns true if the given operatorParams indicate that the
+    /// operator is undelegating its stake or that it completed stake
+    /// undelegation.
+    /// @param operatorParams Parameters of the operator, as stored in the
+    /// staking contract.
+    function isUndelegating(uint256 operatorParams)
+        internal view returns (bool) {
```
nkuba, line 146: Can this function be updated to `pure`?
```suggestion
        internal pure returns (bool) {
```
If so, can we do the same in the `TokenStaking`'s one?

solidity/test/token_grant/TestTokenGrantStake.js
```diff
@@ -287,20 +287,36 @@ describe('TokenGrant/Stake', function() {
...
     let createdAt = web3.utils.toBN((await web3.eth.getBlock(tx.receipt.blockNumber)).timestamp)
     await time.increaseTo(createdAt.add(initializationPeriod).addn(1))
     tx = await grantContract.undelegate(operatorOne, {from: grantee})
     await time.increaseTo(createdAt.add(grantUnlockingDuration))
     await grantContract.recoverStake(operatorOne, {from: grantee});
-    await stakingEscrowContract.withdraw(operatorOne, {from: grantee});
 
     await expectRevert(
-      delegateLiquid(grantee, operatorOne, minimumStake),
+      delegate(grantee, operatorOne, minimumStake),
```
nkuba, line 298: Don't we want to leave also the previous case where we tried to top up with liquid tokens?

solidity/contracts/libraries/staking/TopUps.sol
```diff
@@ -20,20 +28,71 @@ library TopUps {
...
+    /// @dev This function should be called only for not yet initialized stake.
+    /// @param value Top-up value, the number of tokens added to the stake.
+    /// @param operator Operator The operator with existing delegation to which
+    /// the tokens should be added to.
+    /// @param operatorParams Parameters of that operator, as stored in the
+    /// staking contract.
+    /// @param escrow Reference to TokenStakingEscrow contract.
+    /// @return New value of parameters. It should be updated for the operator
+    /// in the staking contract.
+    function executeInOneStep(
```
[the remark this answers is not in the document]
pdyraga: At first, I liked this suggestion but then I realized we call `TopUps` library functions as follows:
```
topUps.executeInOneStep(...)
topUps.initiate(...)
topUps.cancel(...)
topUps.commit(...)
```

`topUps.instantTopUp(...)` does not fit here very well in my opinion and just `topUps.instant(...)` sounds a little weird but... maybe?

solidity/contracts/libraries/staking/TopUps.sol
pdyraga: Maybe `Stake for the operator already deposited in the escrow`? Technically, it's not the operator who is depositing.

solidity/contracts/libraries/staking/TopUps.sol
pdyraga: Nope, it's not going to work:
```
TopUps.sol:148:41: TypeError: Function declared as pure, but this expression (potentially) reads from the environment or state and thus requires "view".
        return (undelegatedAt != 0) && (block.timestamp > undelegatedAt);
```

solidity/test/token_grant/TestTokenGrantStake.js
pdyraga: This scenario is covered in `TestTopUps.js` by two `""can not be done from liquid tokens""` test cases.

solidity/contracts/libraries/staking/TopUps.sol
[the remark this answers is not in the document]
nkuba: Yep, I had the same concern about `topUps.(...)TopUp`. Also, it feels like `topUps.instant(...)` is missing a verb and is imperfect either. 🤷 
Maybe `instantComplete` (refers to the emitted event) or other ideas: `instantBoost`, `instantStake`. If we won't find anything better we can leave it as is.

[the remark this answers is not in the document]
pdyraga: `instantComplete` sounds nice to me!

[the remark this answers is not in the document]
pdyraga: Done.

solidity/contracts/libraries/staking/TopUps.sol
pdyraga: Done.

PR #1950, opened 2020-07-30
Added test cases for escrow redelegations undelegated/cancelled

pdyraga:
Although this path works fine, we had no unit tests covering it. Adding them now to stay safe!

### DIFF
--- a/solidity/test/token_stake/TestStakingGrant.js
+++ b/solidity/test/token_stake/TestStakingGrant.js
@@ -656,5 +656,53 @@ describe('TokenStaking/StakingGrant', () => {
           "Not authorized"
         )
       })
+
+      it('lands back in the escrow when undelegated and recovered', async () => {
+        const redelegatedAmount = delegatedAmount
+        await tokenStakingEscrow.redelegate(
+          operatorOne, redelegatedAmount, data3, {from: grantee}
+        )
+
+        await time.increase(initializationPeriod.addn(1))
+        await tokenStaking.undelegate(operatorThree, {from: operatorThree})
+
+        await time.increase(undelegationPeriod.addn(1))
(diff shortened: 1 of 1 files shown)


solidity/test/token_stake/TestStakingGrant.js
```diff
@@ -656,5 +656,53 @@ describe('TokenStaking/StakingGrant', () => {
           "Not authorized"
         )
       })
+
+      it('lands back in the escrow when undelegated and recovered', async () => {
+        const redelegatedAmount = delegatedAmount
```
nkuba, line 661: Can we test it with a redelegation value different than the initial delegation value?

```suggestion
        const redelegatedAmount = delegatedAmount.subn(10)
```

solidity/test/token_stake/TestStakingGrant.js
```diff
@@ -656,5 +656,53 @@ describe('TokenStaking/StakingGrant', () => {
...
+        expect(
+          await tokenStakingEscrow.depositWithdrawnAmount(operatorThree)
+        ).to.eq.BN(0)
+        expect(
+          await tokenStakingEscrow.depositRedelegatedAmount(operatorThree)
+        ).to.eq.BN(0)
+      })
+
+      it('lands back in the escrow when cancelled', async () => {
+        const redelegatedAmount = delegatedAmount
```
nkuba, line 687: Can we test it with a redelegation value different than the initial delegation value?

```suggestion
        const redelegatedAmount = delegatedAmount.subn(10)
```

solidity/test/token_stake/TestStakingGrant.js
pdyraga: Sure, good idea!

Issue #1952, opened 2020-07-31
Upgrade libp2p dependencies


pdyraga:
Upgrade libp2p dependencies to the most recent versions:
```
libp2p/go-libp2p v0.10.2
libp2p/go-libp2p-autonat v0.3.1
libp2p/go-libp2p-autonat-svc v0.2.0 (deprecated)
libp2p/go-libp2p-circuit v0.3.1
libp2p/go-libp2p-core v0.6.1
libp2p/go-libp2p-kad-dht v0.8.3
libp2p/go-libp2p-mplex v0.2.4
libp2p/go-libp2p-pubsub v0.3.3
libp2p/go-libp2p-pubsub-router v0.3.1
libp2p/go-mplex v0.1.3
multiformats/go-multistream v0.1.2
```

PR #1954, opened 2020-07-31
Upgrade libp2p dependencies

lukasz-zimnoch:
Closes: #1952 

Here we upgrade libp2p dependencies to the most recent versions. 

Apart from bumping up the versions in `go.mod`, we have to set DHT `server` mode explicitly while creating the router instance to keep the current peer discovery behavior. This is because the new version of libp2p changed the default mode of the DHT router to `auto` which  effectively set the router to `client` mode so it doesn't respond to incoming DHT queries.

Also, the upgrade of libp2p forces the upgrade of the `go-log` library as well. [Switching from v0.0.1 to v1.0.4](https://github.com/ipfs/go-log/compare/v0.0.1...v1.0.4) brings some format changes.


go.mod
```diff
@@ -12,24 +12,22 @@ require (
...
-	github.com/libp2p/go-libp2p v0.4.1
-	github.com/libp2p/go-libp2p-connmgr v0.1.0
-	github.com/libp2p/go-libp2p-core v0.3.0
-	github.com/libp2p/go-libp2p-kad-dht v0.3.0
-	github.com/libp2p/go-libp2p-peerstore v0.1.4
-	github.com/libp2p/go-libp2p-pubsub v0.2.6-0.20200127182502-25c434f5f772
-	github.com/libp2p/go-libp2p-secio v0.2.1
-	github.com/libp2p/go-yamux v1.2.4 // indirect
+	github.com/libp2p/go-addr-util v0.0.2
+	github.com/libp2p/go-libp2p v0.10.2
```
mhluongo, line 20: I believe we want `v0.10.3` - it looks like `v0.10.2` has a regression

pdyraga:
Tested locally, works as advertised. @mhluongo's comment is addressed now so waiting for him to do the honors.

PR #1955, opened 2020-07-31
keep-dev: Refresh round 2

sthompson22:
### Intro

keep-dev was our first Keep env, and as such was behind a bit when compared to keep-test/prd structuring.  Here we bring keep-dev up to speed with the general structure of keep-prd.

### What We Did 

- Configs to more explicit files
- Migrate from helm v2 -> helm v3 (gets rid of Tiller)

### Notes

I ended up smashing the installed helm resource state, resulting in needing to recreate the VPN server.  Will need to create new users to go with that.


lionakhnazarov:
Closing as stale. This 2020 "keep-dev refresh" branch is long obsolete and conflicts with current main.

Issue #1930, opened 2020-07-20
Expose connected peers list in diagnostics endpoint


pdyraga:
We could have an endpoint listing IDs of connected peers (full multiaddr, if possible) both for `keep-core` and `keep-ecdsa` clients.

PR #1959, opened 2020-08-04
Expose peers list as JSON via Diagnostics package

ElderOrb:
~~Depends on https://github.com/keep-network/keep-common/pull/43~~
Refs https://github.com/keep-network/keep-core/issues/1930

Enables the diagnostics module which exposes information useful for debugging and diagnostic client's status. 

Diagnostics module exposes the following information:
- list of connected peers along with their network id and ethereum operator address
- information about the client's network id and ethereum operator address 

Diagnostics module can be enabled in the config TOML file:
```
[Diagnostics]
Port = 8002
```

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -255,4 +255,5 @@ func initializeDiagnostics(
 	)
 
 	diagnostics.RegisterConnectedPeersSource(registry, netProvider)
+	diagnostics.RegisterNodeInfoSource(registry, netProvider)
 }
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -2,9 +2,11 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
+	"github.com/keep-network/keep-core/pkg/diagnostics"
 	"github.com/keep-network/keep-core/pkg/metrics"
 	"github.com/keep-network/keep-core/pkg/net"
 
@@ -157,6 +159,7 @@ func Start(c *cli.Context) error {
 	}
 
 	initializeMetrics(ctx, config, netProvider, stakeMonitor, ethereumKey.Address.Hex())
+	initializeDiagnostics(ctx, config, netProvider, stakeMonitor)
 
 	select {
 	case <-ctx.Done():
@@ -231,5 +234,51 @@ func initializeMetrics(
 	metrics.ExposeLibP2PInfo(
 		registry,
 		netProvider,
+	)
+}
+
+func initializeDiagnostics(
+	ctx context.Context,
+	config *config.Config,
+	netProvider net.Provider,
+	stakeMonitor chain.StakeMonitor,
+) {
+	registry, isConfigured := diagnostics.Initialize(
+		config.Diagnostics.Port,
+	)
+	if !isConfigured {
+		logger.Infof("diagnostics are not configured")
+		return
+	}
+
+	registry.RegisterSource("peers", func() string {
+		connectedPeers := netProvider.ConnectionManager().ConnectedPeers()
+
+		peersList := make([]map[string]interface{}, len(connectedPeers))
+		for i := 0; i < len(connectedPeers); i++ {
+			peerAddress, err := netProvider.ConnectionManager().GetPeerPublicKey(connectedPeers[i])
+			if err == nil {
+				hasMinimumStake := false
+				hasMinimumStake, _ = stakeMonitor.HasMinimumStake(peerAddress.X.String())
+
+				peersList[i] = map[string]interface{}{
+					"PeerId":          connectedPeers[i],
+					"PeerAddress":     peerAddress.X.String(),
+					"HasMinimumStake": hasMinimumStake}
+			}
+		}
+
+		bytes, err := json.Marshal(peersList)
+		if err != nil {
+			return ""
+		}
+
+		logger.Debug("peers list: ", string(bytes))
+		return string(bytes)
+	})
+
+	logger.Infof(
+		"enabled diagnostics on port [%v]",
+		config.Diagnostics.Port,
 	)
 }
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -255,5 +255,5 @@ func initializeDiagnostics(
 	)
 
 	diagnostics.RegisterConnectedPeersSource(registry, netProvider)
-	diagnostics.RegisterNodeInfoSource(registry, netProvider)
+	diagnostics.RegisterClientInfoSource(registry, netProvider)
 }
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"time"
 
@@ -255,28 +254,5 @@ func initializeDiagnostics(
 		config.Diagnostics.Port,
 	)
 
-	registry.RegisterSource("connected_peers", func() string {
-		connectedPeers := netProvider.ConnectionManager().ConnectedPeers()
-
-		peersList := make([]map[string]interface{}, len(connectedPeers))
-		for i := 0; i < len(connectedPeers); i++ {
-			peerPublicKey, err := netProvider.ConnectionManager().GetPeerPublicKey(connectedPeers[i])
-			if err != nil {
-				logger.Error("Error on getting peer public key: [%v]", err)
-				continue
-			}
-
-			peersList[i] = map[string]interface{}{
-				"PeerId":        connectedPeers[i],
-				"PeerPublicKey": key.NetworkPubKeyToEthAddress(peerPublicKey),
-			}
-		}
-
-		bytes, err := json.Marshal(peersList)
-		if err != nil {
-			return ""
-		}
-
-		return string(bytes)
-	})
+	diagnostics.RegisterConnectedPeersSource(registry, netProvider)
 }
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -159,7 +159,7 @@ func Start(c *cli.Context) error {
 	}
 
 	initializeMetrics(ctx, config, netProvider, stakeMonitor, ethereumKey.Address.Hex())
-	initializeDiagnostics(ctx, config, netProvider, stakeMonitor)
+	initializeDiagnostics(ctx, config, netProvider)
 
 	select {
 	case <-ctx.Done():
@@ -241,7 +241,6 @@ func initializeDiagnostics(
 	ctx context.Context,
 	config *config.Config,
 	netProvider net.Provider,
-	stakeMonitor chain.StakeMonitor,
 ) {
 	registry, isConfigured := diagnostics.Initialize(
 		config.Diagnostics.Port,
@@ -251,20 +250,25 @@ func initializeDiagnostics(
 		return
 	}
 
-	registry.RegisterSource("peers", func() string {
+	logger.Infof(
+		"enabled diagnostics on port [%v]",
+		config.Diagnostics.Port,
+	)
+
+	registry.RegisterSource("connected_peers", func() string {
 		connectedPeers := netProvider.ConnectionManager().ConnectedPeers()
 
 		peersList := make([]map[string]interface{}, len(connectedPeers))
 		for i := 0; i < len(connectedPeers); i++ {
-			peerAddress, err := netProvider.ConnectionManager().GetPeerPublicKey(connectedPeers[i])
-			if err == nil {
-				hasMinimumStake := false
-				hasMinimumStake, _ = stakeMonitor.HasMinimumStake(peerAddress.X.String())
+			peerPublicKey, err := netProvider.ConnectionManager().GetPeerPublicKey(connectedPeers[i])
+			if err != nil {
+				logger.Error("Error on getting peer public key: [%v]", err)
+				continue
+			}
 
-				peersList[i] = map[string]interface{}{
-					"PeerId":          connectedPeers[i],
-					"PeerAddress":     peerAddress.X.String(),
-					"HasMinimumStake": hasMinimumStake}
+			peersList[i] = map[string]interface{}{
+				"PeerId":        connectedPeers[i],
+				"PeerPublicKey": key.NetworkPubKeyToEthAddress(peerPublicKey),
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


ElderOrb:
I've noticed merge conflict and rebased on the latest master

config/config.go
```diff
@@ -34,6 +35,11 @@ type Metrics struct {
 	EthereumMetricsTick int
 }
 
+// Diagnostics stores diagnostics-related configuration.
+type Diagnostics struct {
+	Port int
+}
```
lukasz-zimnoch, line 41: Please add those to the config sample: https://github.com/keep-network/keep-core/blob/master/configs/config.toml.SAMPLE

cmd/start.go
```diff
@@ -233,3 +236,49 @@ func initializeMetrics(
...
+		}
+
+		logger.Debug("peers list: ", string(bytes))
+		return string(bytes)
+	})
+
+	logger.Infof(
+		"enabled diagnostics on port [%v]",
+		config.Diagnostics.Port,
+	)
```
lukasz-zimnoch, line 283: Can we move this up and place it just below the `!isConfigured` check?

cmd/start.go
```diff
@@ -233,3 +236,49 @@ func initializeMetrics(
...
+) {
+	registry, isConfigured := diagnostics.Initialize(
+		config.Diagnostics.Port,
+	)
+	if !isConfigured {
+		logger.Infof("diagnostics are not configured")
+		return
+	}
+
+	registry.RegisterSource("peers", func() string {
```
lukasz-zimnoch, line 254: I'd name it `connected_peers` to be more specific

cmd/start.go
```diff
@@ -233,3 +236,49 @@ func initializeMetrics(
...
+		connectedPeers := netProvider.ConnectionManager().ConnectedPeers()
+
+		peersList := make([]map[string]interface{}, len(connectedPeers))
+		for i := 0; i < len(connectedPeers); i++ {
+			peerAddress, err := netProvider.ConnectionManager().GetPeerPublicKey(connectedPeers[i])
+			if err == nil {
+				hasMinimumStake := false
+				hasMinimumStake, _ = stakeMonitor.HasMinimumStake(peerAddress.X.String())
```
lukasz-zimnoch, line 262: What's the point of checking the minimum stake here? If we have a peer on the connected peers list we can assume they had the minimum stake because the handshake during the connection attempt was successful. Of course, it's possible the peer just loosed the stake and will be disconnected by the firewall soon. However, I think the only question answered by this diagnostic source should be "what specific peers my client is connected to at this point?"

cmd/start.go
```diff
@@ -233,3 +236,49 @@ func initializeMetrics(
...
+		logger.Infof("diagnostics are not configured")
+		return
+	}
+
+	registry.RegisterSource("peers", func() string {
+		connectedPeers := netProvider.ConnectionManager().ConnectedPeers()
+
+		peersList := make([]map[string]interface{}, len(connectedPeers))
+		for i := 0; i < len(connectedPeers); i++ {
+			peerAddress, err := netProvider.ConnectionManager().GetPeerPublicKey(connectedPeers[i])
```
lukasz-zimnoch, line 259: I think `peerAddress`should rather be `peerPublicKey` becasue it's the result of `GetPeerPublicKey` method.

cmd/start.go
```diff
@@ -233,3 +236,49 @@ func initializeMetrics(
...
+			if err == nil {
```
lukasz-zimnoch, line 260: I think we should check `err != nil` as usual and if its `true` we should log an error and continue the loop. Currently we won't know if something goes wrong.

cmd/start.go
```diff
@@ -233,3 +236,49 @@ func initializeMetrics(
...
+				hasMinimumStake := false
+				hasMinimumStake, _ = stakeMonitor.HasMinimumStake(peerAddress.X.String())
+
+				peersList[i] = map[string]interface{}{
+					"PeerId":          connectedPeers[i],
+					"PeerAddress":     peerAddress.X.String(),
```
lukasz-zimnoch, line 266: Regarding `peerAddress.X.String()`, we have a method which converts the network public key to the Ethereum address. See `key.NetworkPubKeyToEthAddress`.

cmd/start.go
```diff
@@ -233,3 +236,49 @@ func initializeMetrics(
...
+					"HasMinimumStake": hasMinimumStake}
+			}
+		}
+
+		bytes, err := json.Marshal(peersList)
+		if err != nil {
+			return ""
+		}
+
+		logger.Debug("peers list: ", string(bytes))
```
lukasz-zimnoch, line 276: I think we should remove this `debug`. It can introduce to much noise.

cmd/start.go
lukasz-zimnoch, line 254: I think we should also extract this part to a separate function (`RegisterConnectedPeersSource`) and place it in `pkg/diagnostics/diagnostics.go` (just like we do for metrics). This way we can reuse this code from `keep-ecdsa` as well.

lukasz-zimnoch (changes requested):
Thanks for that @ElderOrb! Left some comments. Apart from that, please update the PR description and keep in mind we should update the `keep-common` dependency in `go.mod` once https://github.com/keep-network/keep-common/pull/43 is merged.

cmd/start.go
ElderOrb: Well, yeah... Actually I hoped to raise discussion about other useful information which could be collected as the part of 'connected_peers' diagnostics source, but as @pdyraga suggested - I'll rather do it in separate issue. Deleting for now.

cmd/start.go
ElderOrb: done

cmd/start.go
ElderOrb: sure

cmd/start.go
ElderOrb: I've extracted code into RegisterConnectedPeersSource and placed it into pkg/diagnostics/diagnostics.go *of keep-core* but this way it doesn't really help to reuse it from keep-ecdsa (unless I'm missing something). Or you meant moving it into pkg/diagnostics/diagnostics.go of keep-common? If this is the case, should I also move 'Initialize' function from pkg/diagnostics/diagnostics.go of keep-core to pkg/diagnostics/diagnostics.go of keep-common (and drop keep-core's pkg/diagnostics/diagnostics.go completely)?

ElderOrb:
Note: unfortunately I can't get the latest keep-core connected so can't verify functionality end-to-end.

ElderOrb:
Also, should I add 'self' to connected peers? From one side, this will make peers list non-empty even in disconnected state and thus might lead to confusion. From other side - without 'self' - node list will be incomplete which will make life of monitoring tools more complicated. What if I add it as separate diagnostics source named 'node_info' or something?

pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,56 @@
...
+
+	registry := diagnostics.NewRegistry()
+
+	registry.EnableServer(port)
+
+	return registry, true
+}
+
+func RegisterConnectedPeersSource(registry *diagnostics.DiagnosticsRegistry, netProvider net.Provider) {
+
```
lukasz-zimnoch, line 30: Please remove this blank line and make sure all of the below lines doesn't exceed `80` columns of length [as per our go guidelines](https://github.com/keep-network/keep-core/blob/master/docs/development/go-guidelines.adoc#max-line-length)

pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,56 @@
...
+	registry.RegisterSource("connected_peers", func() string {
+		connectedPeers := netProvider.ConnectionManager().ConnectedPeers()
+
+		peersList := make([]map[string]interface{}, len(connectedPeers))
+		for i := 0; i < len(connectedPeers); i++ {
+			peerPublicKey, err := netProvider.ConnectionManager().GetPeerPublicKey(connectedPeers[i])
+			if err != nil {
+				logger.Error("Error on getting peer public key: [%v]", err)
```
lukasz-zimnoch, line 38: We usually start log messages with lowercase. Can we align here?

pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,56 @@
...
+
+			peersList[i] = map[string]interface{}{
+				"PeerId":        connectedPeers[i],
+				"PeerPublicKey": key.NetworkPubKeyToEthAddress(peerPublicKey),
+			}
+		}
+
+		bytes, err := json.Marshal(peersList)
+		if err != nil {
+			logger.Error("Error on serializing peers list to JSON: [%v]", err)
```
lukasz-zimnoch, line 50: And here.

cmd/start.go
lukasz-zimnoch: I mean `pkg/diagnostics/diagnostics.go` of `keep-core`. `keep-ecdsa` depends on `keep-core` directly so we can use that public function.

lukasz-zimnoch:
> Also, should I add 'self' to connected peers? From one side, this will make peers list non-empty even in disconnected state and thus might lead to confusion. From other side - without 'self' - node list will be incomplete which will make life of monitoring tools more complicated. What if I add it as separate diagnostics source named 'node_info' or something?

Personally, I'd vote for a separate source. `node_info` seems a good candidate. However, please note we already expose the peer ID through the metrics endpoint. It will probably make sense to move this from metrics to the diagnostics endpoint.

lukasz-zimnoch:
@ElderOrb thanks for the improvements! Just 3 minor comments and I think we can merge.

pkg/diagnostics/diagnostics.go
ElderOrb: done

pkg/diagnostics/diagnostics.go
ElderOrb: done

pkg/diagnostics/diagnostics.go
ElderOrb: done

ElderOrb:
> > Also, should I add 'self' to connected peers? From one side, this will make peers list non-empty even in disconnected state and thus might lead to confusion. From other side - without 'self' - node list will be incomplete which will make life of monitoring tools more complicated. What if I add it as separate diagnostics source named 'node_info' or something?
> 
> Personally, I'd vote for a separate source. `node_info` seems a good candidate. However, please note we already expose the peer ID through the metrics endpoint. It will probably make sense to move this from metrics to the diagnostics endpoint.

Added 'node_info'. For now I didn't remove anything from metrics as it might break dashboards which are already  using 'libp2p_info' but can do it in a separate PR if needed.

pdyraga (changes requested):
The functionality itself looks good, I commented mostly on naming/documentation/code styling.

We should also edit the PR description so that it reflects what functionality we are adding here. We should describe what diagnostics are exposed, how to configure them, and how to fetch them.

@lukasz-zimnoch I think we don't have to bother @ElderOrb with this, and if you are up for addressing my comments, then I think we can merge this PR quickly.

configs/config.toml.SAMPLE
```diff
@@ -54,7 +54,17 @@
 [Storage]
   DataDir = "/my/secure/location"
 
+# Uncomment to enable the metrics module which collects and exposes information
+# useful for external monitoring tools. The port on which the `/metrics`
+# endpoint will be available and the frequency with which the metrics will be
+# collected can be customized using the below parameters.
```
pdyraga, line 60: Let's list what kind of metrics are available as well as what types of values those are to explain the difference between `Metrics` and `Diagnostics` modules. My proposition:

```
# Uncomment to enable the metrics module which collects and exposes information
# useful for external monitoring tools usually operating on time series data. All values
# exposed by metrics module are quantifiable or countable. 
#
# The following metrics are available:
# - connected peers count
# - connected bootstraps count
# - eth client connectivity status
#
# The port on which the `/metrics` endpoint will be available and the frequency with
# which the metrics will be collected can be customized using the below parameters.
```

pdyraga, line 60: We are exposing `libp2p_info` in metrics. Given it's a static value available in diagnostics, shouldn't we remove it from metrics? Removing `ExposeLibP2PInfo` from `metrics.go` looks like a low-hanging fruit we can pick in this PR.

configs/config.toml.SAMPLE
```diff
@@ -54,7 +54,17 @@
...
 # [Metrics]
     # Port = 8080
     # NetworkMetricsTick = 60
     # EthereumMetricsTick = 600
+
+# Uncomment to enable the diagnostics module which exposes information
+# useful for debugging. The port on which the `/diagnostics` endpoint
+# will be available can be customized.
```
pdyraga, line 68: Let's list what diagnostic information is available and mention this endpoint is not only for debuging but also for diagnostic the current client's status. My proposition:
```
# Uncomment to enable the diagnostics module which exposes information useful
# for debugging and diagnostic client's status. 
#
# Diagnostics module exposes the following information:
# - list of connected peers along with their network id and ethereum operator address
# - information about the client's network id and ethereum operator address 
#
# The port on which the `/diagnostics` endpoint ewill be available can be customized
# below.
```

pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/ipfs/go-log"
+	"github.com/keep-network/keep-common/pkg/diagnostics"
+	"github.com/keep-network/keep-core/pkg/net"
+	"github.com/keep-network/keep-core/pkg/net/key"
+)
+
+var logger = log.Logger("keep-diagnostics")
+
+// Initialize set up the diagnostics registry and enables diagnostics server.
```
pdyraga, line 14: `sets up`

pdyraga:
Tested locally and works as expected. I edited PR description so that it explains the change.

Added 4 last comments, once they are addressed I am good to merge.

We should now work on enabling it in `keep-ecdsa` @lukasz-zimnoch

pdyraga (approved):

pdyraga:
Thanks for the contribution @ElderOrb :tada:

ElderOrb:
Welcome! And I welcome all interested to continue discussion here: https://github.com/keep-network/keep-core/issues/2009

Issue #1968, opened 2020-08-05
Remove redundant Address property in [ethereum.account] config section


pdyraga:
Beacon client expects `Address` property to be present in `[ethereum.account]` config section. This information is redundant because we can read address from the key file as we do in `keep-ecdsa`. We should remove `Address` property altogether and instead of having:
```
[ethereum.account]
  Address = "0x65ea55c1f10491038425725dc00dffeab2a1e28a"
  KeyFile = "/Users/piotr/ethereum/data/keystore/UTC--2018-11-01T06-24-04.418691008Z--65ea55c1f10491038425725dc00dffeab2a1e28a"
```

Have only:
```
[ethereum.account]
  KeyFile = "/Users/piotr/ethereum/data/keystore/UTC--2018-11-01T06-24-04.418691008Z--65ea55c1f10491038425725dc00dffeab2a1e28a"
```

PR #1971, opened 2020-08-05
Client provisioning without eth account address

nkuba:
As per https://github.com/keep-network/keep-core/issues/1968 keep client no longer requires ethereum account address to be provided in the config file. 

In this PR we update provisioning script not to set ethereum address in a config file. We also updated kube's configs for keep-dev and keep-test to remove ethereum addresses.

Depends on: https://github.com/keep-network/keep-core/pull/1970

### DIFF
--- a/infrastructure/kube/templates/keep-client/initcontainer/provision-keep-client/provision-keep-client.js
+++ b/infrastructure/kube/templates/keep-client/initcontainer/provision-keep-client/provision-keep-client.js
@@ -16,6 +16,8 @@ var authorizer = contractOwnerAddress
 
 var contractOwnerProvider = new HDWalletProvider(process.env.CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY, ethRPCUrl);
 
+const operatorKeyFile = process.env.KEEP_CLIENT_ETH_KEYFILE_PATH;
+
 /*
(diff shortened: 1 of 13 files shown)


infrastructure/kube/templates/keep-client/initcontainer/provision-keep-client/provision-keep-client.js
```diff
@@ -11,10 +11,18 @@ const ethWSUrl = process.env.ETH_WS_URL
 const ethNetworkId = process.env.ETH_NETWORK_ID;
 
 // Contract owner info
-var contractOwnerAddress = process.env.CONTRACT_OWNER_ETH_ACCOUNT_ADDRESS;
-var authorizer = contractOwnerAddress
+const contractOwnerAddress = process.env.CONTRACT_OWNER_ETH_ACCOUNT_ADDRESS;
+const authorizer = contractOwnerAddress
+const purse = contractOwnerAddress;
```
pdyraga, line 16: Given that it's now global constant, should we update `async function fundOperator(operatorAddress, purse, etherToTransfer)` by removing `purse` parameter from there?

pdyraga:
Given that #1970 is merged, should we un-draft this PR?

pdyraga:
Let's do it 💪

PR #1972, opened 2020-08-05
Release: 1.3.0-rc (Ropsten)

sthompson22:
1.3.0-rc release for Ropsten. 

Changes include here:
- We no longer use utilitybox for migrations
- We deploy `OldTokenStaking` stub for Ropsten so that we can test UI/UX of copy stake flow there
- Updated Tenderly account ID


.circleci/config.yml
```diff
@@ -91,24 +91,19 @@ jobs:
...
           name: Migrate Contracts
           command: |
-            export BUILD_TAG=$CIRCLE_SHA1
-            export TRUFFLE_NETWORK=$TRUFFLE_NETWORK
-            export CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY=$CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY
             mkdir -p /tmp/keep-client/contracts
-            solidity/scripts/circleci-migrate-contracts.sh
+            cd ./solidity
+            npm ci
+            ./node_modules/.bin/truffle migrate --reset --network $TRUFFLE_NETWORK
```
Shadowfiend, line 105: Notably, we lost tenderly here.

pdyraga: 😬 

What I think we need to do is to add:
```
tenderly login --authentication-method token --token $TENDERLY_TOKEN
tenderly push --networks $ETH_NETWORK_ID --tag keep-core \
    --tag $GOOGLE_PROJECT_NAME --tag $CIRCLE_SHA1 || echo "tendery push failed :("
```
here and remove `circleci-migrate-contract.sh` altogether since we no longer use it anywhere.

pdyraga: Please see https://github.com/keep-network/keep-core/pull/1972/commits/9166ac59605c274e5bfbc189be8eb45a8a0d3c8c and https://github.com/keep-network/keep-core/pull/1972/commits/5049f18309a2e2b005a1b72768fcd799cd8cf3c3.

pdyraga: ^ Those two commits do no longer reflect the whole solution. Tenderly is back and it's configured in contracts migration shell script.

nkuba (approved):

pdyraga (approved):

solidity/migrations/2_deploy_contracts.js
```diff
@@ -62,30 +62,24 @@ module.exports = async function(deployer, network) {
...
 
   let oldStakingContractAddress;
-  if (network === 'mainnet') {
-    // v1.0.1 TokenStaking contract address
-    oldStakingContractAddress = '0x6D1140a8c8e6Fac242652F0a5A8171b898c67600';
-  } else if (network === 'ropsten') {
-    // 1.3.0-rc.0 TokenStaking contract address
-    oldStakingContractAddress = '0x8117632eC1D514550b3880Bc68F9AC1A76c9C67B';
-  } else {
+  if (network === 'local' || network === 'ropsten') {
```
nkuba, line 65: For ropsten should we consider using already deployed contracts similiar as we will do for mainnet?

pdyraga: It's been explained in https://github.com/keep-network/keep-core/pull/1972/commits/b2aa91b9a4637d0c4f206d2028cad8aaf67cab4d:

```
We can't just point to the already deployed TokenStaking version on
Ropsten because it would require using the same KeepToken and TokenGrant
contracts and doing that is non-trivial given how our workflows are
configured: if we don't deploy a contract, it's not included in NPM
package and if it's not included in NPM package, provisioning scripts
and dashboard may not work.

Given that we are going to test copy stake flow using mainnet version of
contracts manually, we are deploying OldTokenStaking stub to allow for
UI/UX tests for copy flow on Ropsten but we don't consider those UI/UX
tests as ones giving us any guarantee the whole functionality works as
expected. To have this guarantee, we are going to test copy stake flow
using a clone of mainnet contracts.
```

PR #1974, opened 2020-08-06
Allow to top-up less than the minimum stake

pdyraga:
Having the minimum stake restriction on delegation makes perfect sense but users may want to top-up less than the minimum stake. It makes sense in two cases:
- As long as we slash 1% or 50% of the minimum stake (first 3 months and the next 3 months), users may delegate e.g. 100k KEEP and then to-up 1k KEEP to have 1% safety margin in case of slashing.
- Sortition pool weights by stake and if the `poolWeightDivisor` is lower than the minimum stake, having 101K KEEP delegated gives better chances of being selected than having 100k KEEP delegated.

Props to @r-czajkowski for finding this.

### DIFF
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -143,7 +143,6 @@ contract TokenStaking is Authorizations, StakeDelegatable {
         bytes memory _extraData
     ) public {
         require(ERC20Burnable(_token) == token, "Unrecognized token");
-        require(_value >= minimumStake(), "Less than the minimum stake");
         require(_extraData.length >= 60, "Corrupted delegation data");
 
         // Transfer tokens to this contract.
@@ -173,6 +172,8 @@ contract TokenStaking is Authorizations, StakeDelegatable {
         address _operator,
         bytes memory _extraData
     ) internal {
+        require(_value >= minimumStake(), "Less than the minimum stake");
+        
         address payable beneficiary = address(uint160(_extraData.toAddress(0)));
         address authorizer = _extraData.toAddress(40);
 
--- a/solidity/test/token_stake/TestTopUps.js
+++ b/solidity/test/token_stake/TestTopUps.js
@@ -400,7 +400,7 @@ describe('TokenStaking/TopUps', () => {
     })
 
     it("can be done with less than the minimum stake", async () => {
-      const topUpAmount = 9
+      const topUpAmount = (await tokenStaking.minimumStake()).subn(1)
 
       await time.increase(initializationPeriod.addn(1))
       await initiateTopUp(topUpAmount)
@@ -411,7 +411,7 @@ describe('TokenStaking/TopUps', () => {
         operatorOne,
         operatorContract
       )
-      expect(currentStake).to.eq.BN(delegatedAmount.addn(topUpAmount))
+      expect(currentStake).to.eq.BN(delegatedAmount.add(topUpAmount))
     })
   })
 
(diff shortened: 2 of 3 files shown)


solidity/test/token_stake/TestTopUps.js
```diff
@@ -581,18 +596,33 @@ describe('TokenStaking/TopUps', () => {
       )
       expect(currentStake).to.eq.BN(delegatedAmount.muln(4))
     })
+
+    it("can be done with less than the minimum stake", async () => {
+      const topUpAmount = 7
```
nkuba, line 601: It may be better to set this value relative to the minimum stake.
```suggestion
      const topUpAmount = (await tokenStaking.minimumStake()).subn(1)
```

solidity/test/token_stake/TestTopUps.js
```diff
@@ -763,6 +793,21 @@ describe('TokenStaking/TopUps', () => {
       )
       expect(currentStake).to.eq.BN(delegatedAmount.muln(4))
     })
+
+    it("can be done with less than the minimum stake", async () => {
+      const topUpAmount = 11
```
nkuba, line 798: It may be better to set this value relative to the minimum stake.
```suggestion
      const topUpAmount = (await tokenStaking.minimumStake()).subn(1)
```

solidity/test/token_stake/TestTopUps.js
```diff
@@ -843,5 +888,39 @@ describe('TokenStaking/TopUps', () => {
...
+        delegatedAmount,
+        grantId
+      )
+
+      await time.increase(initializationPeriod.addn(1))
+
+      let delegationInfo = await tokenStaking.getDelegationInfo(operatorFour)
+      expect(delegationInfo.amount).to.eq.BN(delegatedAmount)
+
+      const topUpAmount = 91
```
nkuba, line 909: It may be better to set this value relative to the minimum stake.
```suggestion
      const topUpAmount = (await tokenStaking.minimumStake()).subn(1)
```

solidity/contracts/TokenStaking.sol
```diff
@@ -143,7 +143,6 @@ contract TokenStaking is Authorizations, StakeDelegatable {
         bytes memory _extraData
     ) public {
         require(ERC20Burnable(_token) == token, "Unrecognized token");
-        require(_value >= minimumStake(), "Less than the minimum stake");
```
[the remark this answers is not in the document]
pdyraga: Good idea, please see https://github.com/keep-network/keep-core/pull/1974/commits/57d002f2a5f03bca6dba09f0b07c26e40406f959

solidity/test/token_stake/TestTopUps.js
pdyraga: Please see https://github.com/keep-network/keep-core/pull/1974/commits/81619c70a48b3f53828a3f0894ddfbb78e00343e

solidity/test/token_stake/TestTopUps.js
pdyraga: Please see https://github.com/keep-network/keep-core/pull/1974/commits/81619c70a48b3f53828a3f0894ddfbb78e00343e

solidity/test/token_stake/TestTopUps.js
pdyraga: Please see https://github.com/keep-network/keep-core/pull/1974/commits/81619c70a48b3f53828a3f0894ddfbb78e00343e

PR #1975, opened 2020-08-07
DefaultMaxGasPrice updated to 70Gwei

pdyraga:
Given the current trends of gas prices, it makes sense to increase the `DefaultMaxGasPrice` to `70Gwei`. As for today, Etherscan indicates `60Gwei` is the safe price and `75Gwei` is the proposed price, both with ~20 seconds expected mining time of a TX.

### DIFF
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -19,7 +19,7 @@
 	# allowed gas price is reached, no further resubmission attempts are
 	# performed.
 	#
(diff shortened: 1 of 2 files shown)


dimpar:
can we also update `config.toml.SAMPLE`? Now it says:
```
# MaxGasPrice = 50000000000 # 50 gwei (default value)
```

pdyraga:
Ops, very good catch. Updated!

PR #1977, opened 2020-08-10
Minimum stake schedule starts at the moment of deploying KEEP token

pdyraga:
We are deploying a new staking contract version but we want to maintain the same minimum KEEP stake schedule. To make it happen, instead of using staking contract deployment time as a schedule start, we now define a constant in `MinimumStakeSchedule` with a value equal to the time of deploying KEEP token. On the other hand, we want the slashing schedule (1% for the first 3 month and 50% for the next 3 months) start with the time of deploying the new `TokenStaking` contract. I renamed `TokenStaking.minimumStakeScheduleStart` to `TokenStaking.deployedAt` and updated `Groups.sol` to use the new name.

The most problematic part were unit tests. Since we now use a fixed point in the past as a minimum stake schedule start, I had to introduce `TokenStakingStub` for some tests with a fixed minimum stake of 100k KEEP.

### DIFF
--- a/solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
+++ b/solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
@@ -9,12 +9,20 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 library MinimumStakeSchedule {
     using SafeMath for uint256;
 
+    // Apr-28-2020 02:52:46 AM UTC when KEEP token has been deployed
+    // TX:  0xea22d72bc7de4c82798df7194734024a1f2fd57b173d0e065864ff4e9d3dc014
+    uint256 public constant scheduleStart = 1588042366;
+    
     // 2 years in seconds (seconds per day * days in a year * years)
     uint256 public constant schedule = 86400 * 365 * 2;
     uint256 public constant steps = 10;
     uint256 public constant base = 10000 * 1e18;
 
-    function current(uint256 scheduleStart) public view returns (uint256) {
+    function current() public view returns (uint256) {
+        return current(scheduleStart);
+    }
+
+    function current(uint256 scheduleStart) internal view returns (uint256) {
         if (now < scheduleStart.add(schedule)) {
             uint256 currentStep = steps.mul(now.sub(scheduleStart)).div(schedule);
             return base.mul(steps.sub(currentStep));
--- a/solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
+++ b/solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
@@ -4,8 +4,9 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 
 /// @notice MinimumStakeSchedule defines the minimum stake parametrization and
 /// schedule. It starts with a minimum stake of 100k KEEP. Over the following
-/// 2 years, the minimum stake is lowered periodically using a uniform stepwise
-/// function, eventually ending at 10k.
+/// 2 years, starting from the moment KEEP token has been deployed, the minimum
+/// stake is lowered periodically using a uniform stepwise function, eventually
+/// ending at 10k.
 library MinimumStakeSchedule {
     using SafeMath for uint256;
 
@@ -18,6 +19,9 @@ library MinimumStakeSchedule {
     uint256 public constant steps = 10;
     uint256 public constant base = 10000 * 1e18;
 
+    /// @notice Returns the current value of the minimum stake. The minimum
+    /// stake is lowered periodically over the course of 2 years since the time
+    /// KEEP token has been deployed and eventually ends at 10k KEEP.
     function current() public view returns (uint256) {
         return current(scheduleStart);
     }
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -63,7 +63,7 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     event LockReleased(address indexed operator, address lockCreator);
     event ExpiredLockReleased(address indexed operator, address lockCreator);
 
-    uint256 public minimumStakeScheduleStart;
+    uint256 public deployedAt;
     uint256 public initializationPeriod; // varies between mainnet and testnet
 
     ERC20Burnable internal token;
@@ -96,7 +96,7 @@ contract TokenStaking is Authorizations, StakeDelegatable {
         escrow = _escrow;
         registry = _registry;
         initializationPeriod = _initializationPeriod;
-        minimumStakeScheduleStart = block.timestamp;
+        deployedAt = block.timestamp;
     }
 
     /// @notice Returns minimum amount of KEEP that allows sMPC cluster client to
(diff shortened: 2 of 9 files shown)


solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
```diff
@@ -9,12 +9,20 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 library MinimumStakeSchedule {
     using SafeMath for uint256;
 
+    // Apr-28-2020 02:52:46 AM UTC when KEEP token has been deployed
+    // TX:  0xea22d72bc7de4c82798df7194734024a1f2fd57b173d0e065864ff4e9d3dc014
+    uint256 public constant scheduleStart = 1588042366;
```
pdyraga, line 14: Triple-check me please:

```
$ date -r 1588042366
Tue 28 Apr 2020 04:52:46 CEST
```

https://etherscan.io/tx/0xea22d72bc7de4c82798df7194734024a1f2fd57b173d0e065864ff4e9d3dc014

pdyraga: 04:52:46 CEST - I remember I observed sunshine back then so we should be good here 😆

dimpar: have the same:
```
date -r 1588042366
Tue Apr 28 04:52:46 CEST 2020
```
Tue Apr 28 04:52:46 CEST 2020 == Apr-28-2020 02:52:46 AM +UTC

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
```diff
@@ -9,12 +9,20 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
...
+    
     // 2 years in seconds (seconds per day * days in a year * years)
     uint256 public constant schedule = 86400 * 365 * 2;
     uint256 public constant steps = 10;
     uint256 public constant base = 10000 * 1e18;
 
-    function current(uint256 scheduleStart) public view returns (uint256) {
+    function current() public view returns (uint256) {
```
dimpar, line 21: since this is `public` and the one below is `internal` maybe we can replace it with 1 function, ie:
```
function current() public view returns (uint256) {
        if (now < scheduleStart.add(schedule)) {
            uint256 currentStep = steps.mul(now.sub(scheduleStart)).div(schedule);
            return base.mul(steps.sub(currentStep));
        }
        return base;
    }
```

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
```diff
@@ -9,12 +9,20 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 library MinimumStakeSchedule {
     using SafeMath for uint256;
 
+    // Apr-28-2020 02:52:46 AM UTC when KEEP token has been deployed
+    // TX:  0xea22d72bc7de4c82798df7194734024a1f2fd57b173d0e065864ff4e9d3dc014
+    uint256 public constant scheduleStart = 1588042366;
+    
     // 2 years in seconds (seconds per day * days in a year * years)
     uint256 public constant schedule = 86400 * 365 * 2;
```
dimpar, line 17: we didn't want to use SafeMath here? `86400.mul(365).mul(2)` and 10000.mul(1e18)

pdyraga: Not super worried about them given that those are constant set during deployment and we know how they are calculated, but...

it's probably a good practice to use SafeMath everywhere. Better be safe than sorry.

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
pdyraga: Negative, please see the minimum stake schedule tests. We can't go back in time in tests, that hurts.

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
pdyraga: Heh, no - it's not possible:
```
,/Users/piotr/go/src/github.com/keep-network/keep-core/solidity/contracts/libraries/staking/MinimumStakeSchedule.sol:19:36: TypeError: Initial value for constant variable has to be compile-time constant.
    uint256 public constant base = uint256(10000).mul(1e18);
```

dimpar:
I went through the code, no more comments

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
nkuba: Confirmed
```
$ date -u -r 1588042366
Tue Apr 28 02:52:46 UTC 2020
```

nkuba, line 14: Why not define this value in `TokenStaking` contract? The value could be provided in the `TokenStaking` constructor.

nkuba: Oh size limitation, huh?

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
nkuba, line 21: Can we add a doc?

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
pdyraga: Yes. Also, we can save on gas when this value is not passed as a parameter. It never changes and `MinimumStakeSchedule` is a part of `TokenStaking`.

nkuba: I don't like hardcoding this value, but we have to 🤷

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
pdyraga: Done!

nkuba: ❤️

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
```diff
@@ -4,17 +4,29 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
...
 /// schedule. It starts with a minimum stake of 100k KEEP. Over the following
-/// 2 years, the minimum stake is lowered periodically using a uniform stepwise
-/// function, eventually ending at 10k.
+/// 2 years, starting from the moment KEEP token has been deployed, the minimum
+/// stake is lowered periodically using a uniform stepwise function, eventually
+/// ending at 10k.
 library MinimumStakeSchedule {
     using SafeMath for uint256;
 
+    // Apr-28-2020 02:52:46 AM UTC when KEEP token has been deployed
```
Shadowfiend, line 13: Let's make this yyyy-mm-dd and drop the AM at some point. Which is to say, let's stick to ISO8601 for dates where possible.

solidity/contracts/libraries/staking/MinimumStakeSchedule.sol
Shadowfiend: I realize in another conversation a note about contract size came up, but it's worth fully articulating here would be that this library should "ideally" focus on the computation, and the staking contract should provide the start time, as was the case before. That also allows test flexibility, since tests can modify that start time to their heart's content, via stub. 

So it's strictly token staking contract size limitations (as I understand it) that have put us into this slightly odd setup. Correct me if I'm wrong, just trying to make sure the reasoning is fully captured.

PR #1981, opened 2020-08-11
Allow to copy stake only one time even if it's been recovered

pdyraga:
Modified `StakingPortBacker` to allow for copying stake only one time even if the copied stake has been recovered. This behavior was expected based on the documentation on `copyStake` which says:

> "(...)  the relationship can only be copied once (...)".

This scenario would be the most problematic in the situation when `StakingPortBacker` owner force-undelegates the stake after 3 months of not paying the loan back and recovers the stake after the undelegation time and the malicious staker copies the stake immediately before `StakingPortBacker` owner has a chance to withdraw tokens (gas auction).

### DIFF
--- a/solidity/test/token_stake/TestStakingPortBacker.js
+++ b/solidity/test/token_stake/TestStakingPortBacker.js
@@ -384,6 +384,38 @@ describe("TokenStaking/StakingPortBacker", () => {
       )
     })
 
+    it("allows to copy stake only one time even if it's been recovered", async () => {
+      await stakingPortBacker.copyStake(operatorOne, {from: tokenOwner})
+      await stakingPortBacker.copyStake(operatorTwo, {from: grantee})
+      await stakingPortBacker.copyStake(operatorThree, {from: managedGrantee})
+
+      await time.increase(time.duration.days(91))
+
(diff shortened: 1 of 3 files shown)


solidity/test/token_stake/TestStakingPortBacker.js
```diff
@@ -384,6 +384,38 @@ describe("TokenStaking/StakingPortBacker", () => {
       )
     })
 
+    it("allows to copy stake only one time even if it's been recovered", async () => {
```
nkuba, line 387: Can we add a case there we call `recoverStake` twice?

pdyraga: Not sure if that's what you asked for but it's in https://github.com/keep-network/keep-core/pull/1981/commits/0708dbca4e52105ba169c0a3af6d6bcc76b4e791 😆 

- The test case when `recoverStake` is called twice is already covered for a delegation created from a grant in `TestStakingGrant.js`,
- We had no test case for `recoverStake` called twice for delegation created from liquid tokens - added it in the mentioned commit,
- I don't think we need to add a test case for `recoverStake` in `TestStakingPortBacker` because this contract is not doing anything else than calling `recoverStake` on the staking contract. Did I miss something?

PR #1984, opened 2020-08-11
Pass minimum stake schedule start as a parameter to the library

pdyraga:
During the review of https://github.com/keep-network/keep-core/pull/1977 @nkuba and @Shadowfiend commented about not hardcoding the schedule start in the library. This is addressed in this PR. 

Instead of hardcoding schedule start in `MinimumStakeSchedule` library, we can pass it as a parameter. It's a little cleaner given that we don't have an alternative function for unit tests. It's a little more expensive gas-wise but we can live with it.

Having this value passed in the constructor is not possible given the contract size constraints. 

We also use ISO8601 format for schedule start date now.

### DIFF
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -77,6 +77,10 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     uint256 internal constant twoWeeks = 1209600; // [sec]
     uint256 internal constant twoMonths = 5184000; // [sec]
 
+    // 2020-04-28; the date of deploying KEEP token.
+    // TX:  0xea22d72bc7de4c82798df7194734024a1f2fd57b173d0e065864ff4e9d3dc014
+    uint256 internal constant minimumStakeScheduleStart = 1588042366;
+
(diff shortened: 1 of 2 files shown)


solidity/contracts/TokenStaking.sol
```diff
@@ -77,6 +77,10 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     uint256 internal constant twoWeeks = 1209600; // [sec]
     uint256 internal constant twoMonths = 5184000; // [sec]
 
+    // 2020-04-28; the date of deploying KEEP token.
+    // TX:  0xea22d72bc7de4c82798df7194734024a1f2fd57b173d0e065864ff4e9d3dc014
+    uint256 internal constant minimumStakeScheduleStart = 1588042366;
```
nkuba, line 82: I'm sorry to grumble again, is there any chance to pass the value to the constructor?

pdyraga: I am afraid that the answer is no, especially that we still need some bytecode: https://github.com/keep-network/keep-core/pull/1990

Issue #1813, opened 2020-05-19
Consider adding an etherscan link to success banner or make it visible longer


dimpar:
As for now, it takes a while on mainnet before a user can see a success banner which closes a banner with a transaction link. However, it would be nice to give a user control over when he/she wants to close it, giving more time to click on the etherscan link.

see https://github.com/keep-network/keep-core/pull/1794#issuecomment-630161238

liz-shinn:
Makes sense to me. Added to the design file:

<img width="708" alt="Token_Dashboard_dApp" src="https://user-images.githubusercontent.com/57226633/82941433-56cc5880-9f64-11ea-9ea8-70d056fafe13.png">

PR #1985, opened 2020-08-11
Migration to redux- dapp

r-czajkowski:
Closes: #1813 

This PR adds a `redux` and `redux-saga` to the keep token dashboard. As for now, the `redux` takes care of sending transactions and displaying messages such as (transaction is pending, success msg with link to the transaction on etherscan etc.).

In the future will be added:
* sagas tests,
* fetching data via redux and store in the global state,
* subscribing to the contracts events via [eventChannel](https://redux-saga.js.org/docs/advanced/Channels.html) and dispatching right actions to update the store data.

This PR also updates the flow of delegation. If a user delegates a stake from a grant, dapp first checks that the grant has the deposited tokens in `TokenStakingEscrow`. If so, dapp tries to redelegate tokens from `TokenStakingEscrow` deposits to the new operator and then staking directly through the `TokenGrant` contract.


dimpar:
Delegated 3 times: 100k, 200k, 300k. After refreshing the page there is no info on the delegated amounts

Screenshot 2020-08-13 at 16 35 20

dimpar:
Scenario: delegated 100k to 0xAAA and then canceled. Delegated to the same account 0xAAA. (didn't go through as expected). Delegated to a different account 0xBBB and then the error below:

Screenshot 2020-08-13 at 16 42 45

dimpar:
Suggestion to disable "Add Keep" when there's no Keep available

dimpar:
Experienced a scenario when after multiple delegation / undelegations / recover and releasing I added 90k Keep, but in the top-ups it actually added 100k. 
Just an assumption.. if there was some Keep in escrow, then it looked like it picked the wrong amount to top-up

dimpar:
Scenario:
Delegated to 0xAAA but then canceled right after.

Problem:
Available amount still displays 4.8M and become 4.9M only after the page refresh. 

Screenshot 2020-08-17 at 11 42 55

dimpar:
Scenario: Created a Grant_1 from 0xAAA to 0xBBB. Did some delegations/undelegations on 0xBBB. Switched account back to 0xAAA and created another Grant_2 for 0xBBB and switched to 0xBBB.

Problem: All of the 20M Delegations and Undelegation in the screen below weren't part of 0xBBB account. Looks like we display stuff from 0xAAA account.

Screenshot 2020-08-17 at 12 02 44

dimpar:
Summary of testing:

From an account that can create grants:
- delegations
- undelegations
- recover
- top-ups and adding Keep multiple times
- multiple grant creation
- add keep -> undelegate without committing a top-up

Checking delegations to the same account from different accounts. Only first goes through.

From granted account:
- delegations
- undelegations
- recover
- release
- multiple top-ups

Undelegation from Operators

dimpar:
@r-czajkowski I think the reason why in `Token Overview` page for a Grant Manager we display delegation from Grantee is because we don't have a filter like we have in `Delegated Tokens -> Owned` [here](https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/components/DelegationOverview.jsx#L13). Maybe we should apply it [here](https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/pages/TokenOverviewPage.jsx#L65) or somewhere in `tokens-page.service.js`?

Screenshot 2020-08-24 at 18 25 58

Screenshot 2020-08-24 at 18 25 28

dimpar:
Fix for the above error is done in [this](https://github.com/keep-network/keep-core/pull/2012/commits/4d97fcd15fcae15aad8fc11d1e737ce2e0861446) commit.

dimpar (approved):

PR #1986, opened 2020-08-11
Improved documentation around AuthorityVerifier

pdyraga:
Updated documentation around `AuthorityVerifier` to be explicit about the
fact it's not some 3rd party contract we cannot trust but that it's
something provided by us, with reasonable gas consumption and never
reverting on the check.

### DIFF
--- a/solidity/contracts/Authorizations.sol
+++ b/solidity/contracts/Authorizations.sol
@@ -42,6 +42,8 @@ interface AuthorityDelegator {
 /// listed there as authorized. This interface is a verifier that support verification
 /// of contract authorization in case of authority delegation from the source contract.
 interface AuthorityVerifier {
+    /// @notice Returns true if the given operator contract has been approved
+    /// for use. The function never reverts.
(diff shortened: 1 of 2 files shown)


solidity/contracts/Authorizations.sol
```diff
@@ -42,6 +42,8 @@ interface AuthorityDelegator {
 /// listed there as authorized. This interface is a verifier that support verification
 /// of contract authorization in case of authority delegation from the source contract.
 interface AuthorityVerifier {
+    /// @notice Returns true if the given operator contract has been approved
+    /// for use. The function never reverts.
```
nkuba, line 46: Do we need to mention that it doesn't revert? The name and the fact that it returns a boolean gives pretty binary result expression.
```suggestion
    /// for use.
```

pdyraga: This was brought by our auditor so I guess something that could be obvious for a person very familiar with the codebase is not that obvious for other contributors. I'd say it doesn't hurt to keep this information here if it can help others.

PR #1987, opened 2020-08-12
Allow to top-up less than the min stake- dapp

r-czajkowski:
Ref: #1974

This PR allows users to top-up less than the minimum stake.


dimpar:
We probably want to change text informing about minimum tokens that we can add

Screenshot 2020-08-12 at 17 26 29

PR #1988, opened 2020-08-13
Release Me: Add release management RFC, RFC 18

Shadowfiend:
This RFC has already gone through some review in our internal Coda, and
is ready for broader reading and comments.

Rendered version is at https://docs.keep.network/release-me/rfc/rfc-18-release-management.html .


docs/rfc/rfc-18-release-management.adoc
```diff
@@ -0,0 +1,489 @@
...
+
+Repositories sourced during release of the system.
+1. keep-common
+2. keep-core
+3. keep-ecdsa
+4. tbtc
+5. tbtc.js
+6. tbtc-dapp
+7. sortition-pools
+8. summa-tx/relays (fork)
```
pdyraga, line 59: When rendered, the formatting is weird - they are all in the same line.

docs/rfc/rfc-18-release-management.adoc
```diff
@@ -0,0 +1,489 @@
...
+latest contract artifacts from a Google storage bucket. This is a
+particularly hairy failure point as mistakes can go unaccounted for until
+deploy time, resulting in potentially having to redo large parts of the
+release. Handling these interdependencies programmatically, and with specific
+context of the upstream state as to minimize downstream work is critical to
+having a more robust, less error prone, and shorter release process.
+
+To illustrate: tbtc contract initialization depends on having the keep-ecdsa
+`BondedECDSAKeepFactory` contract address for the environment being released.
+The BondedECDSAKeepFactory address is produced outside of the release context
```
pdyraga, line 164: Missing backticks for `BondedECDSAKeepFactory`

pdyraga (commented):
I left two comments regarding the syntax, everything else here looks very good and we are already applying tagging strategy proposed here.

For posterity sake, I will mention two issues that were especially painful to me during the last RC hoping we'll address them when implementing this RFC:
- In some cases, it was not possible to resume from the last failed step and I had to execute the entire flow from the very beginning and some steps before the one that failed were very time consuming. Being able to resume the flow from any point would be very valuable.
- There is no way to test the deployment until we do the deployment. We disabled `keep-dev` deployments for `master` merges which is totally fine but being able to run `master` deployment against `keep-dev` Ethereum after pushing the button manually would be helpful for workflow development/adjustements. 

Also, one question regarding GitHub releases and RCs, especially for keep clients: do we want to publish RC releases on GitHub? I see pros and cons of both but I personally lean towards publishing there only final builds.

pdyraga (approved):

PR #1990, opened 2020-08-14
Replace StakeDelegatable.ownerOperators mapping with an event 

r-czajkowski:
`ownerOperators` mapping in `StakeDelegatable` was quite a problematic append-only data structure.

The first issue was that this mapping was not cleared up when a stake was recovered. The reason for that was to maintain a historical mapping between owner and operator so that the owner may withdraw rewards after recovering the stake even if it lost the operator address (operator address is a parameter to withdraw function).

Another issue was that this mapping maintained no longer up-to-date information in case the stake ownership was transferred with `transferStakeOwnership` function in `TokenStaking`. Clearing up this mapping would involve writing a loop and compering operator address bytes. This is problematic because of the contract size constraints as well as there is a tiny possibility such a loop may cause gas limit problems in case an owner has a lot of delegations.

The solution proposed here assumes we drop `ownerOperators` altogether, introduce `StakeDelegated` event with indexed owner and operator, as well as rename `Stake` event to `OperatorStaked` and remove owner field from there.

The relationship between owner and operator can be now looked up using `StakeDelegatedEvent` having both owner and operator indexed. This event is emitted when the delegation to the operator is registerred for the first time as well as when the delegation is transferred to another owner.

`OperatorStaked` event contains information about operator-beneficiary-authorizer triplet and this triplet is never changing. Since those values are set only one time and never updated, we emit this event only one time - when the delegation is created.

From KEEP token dashboard perspective:
- To find authorizer for operator, it's enough to filter out `OperatorStaked` event by indexed fields
- To find beneficiary for operator, it's enough to filter out `OperatorStaked` event by indexed fields
- To find operator for beneficiary, it's enough to filter out `OperatorStaked` event by indexed fields
- To find operator for authorizer, it's enough to filter out `OperatorStaked` event by indexed fields
- To find operator for owner, we should:  
  - Filter out `StakeDelegated` events by owner field
  - Filter out `StakeDelegated` events by operator field, checking more recent events to make sure the delegation ownership has not been transferred.

### DIFF
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -565,6 +565,7 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     function transferStakeOwnership(address operator, address newOwner) public {
         require(msg.sender == operators[operator].owner, "Not authorized");
         operators[operator].owner = newOwner;
+        ownerOperators[newOwner].push(operator);
         emit StakeOwnershipTransferred(operator, newOwner);
     }
 
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -51,7 +51,7 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     );
     event StakeOwnershipTransferred(
         address indexed operator,
-        address newOwner
+        address indexed newOwner
     );
     event TopUpInitiated(address indexed operator, uint256 topUp);
     event TopUpCompleted(address indexed operator, uint256 newAmount);
--- a/solidity/contracts/TokenStaking.sol
+++ b/solidity/contracts/TokenStaking.sol
@@ -42,16 +42,15 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     using Locks for Locks.Storage;
     using TopUps for TopUps.Storage;
 
-    event Staked(
-        address owner,
+    event StakeDelegated(
+        address indexed owner,
+        address indexed operator
+    );
+    event OperatorStaked(
         address indexed operator,
         address indexed beneficiary,
         address indexed authorizer,
         uint256 value
-    );
-    event StakeOwnershipTransferred(
-        address indexed operator,
-        address indexed newOwner
     );
     event TopUpInitiated(address indexed operator, uint256 topUp);
     event TopUpCompleted(address indexed operator, uint256 newAmount);
@@ -183,7 +182,6 @@ contract TokenStaking is Authorizations, StakeDelegatable {
             beneficiary,
             authorizer
         );
-        ownerOperators[_from].push(_operator);
 
         grantStaking.tryCapturingDelegationData(
             tokenGrant,
@@ -193,7 +191,8 @@ contract TokenStaking is Authorizations, StakeDelegatable {
             _extraData
         );
 
-        emit Staked(_from, _operator, beneficiary, authorizer, _value);
+        emit StakeDelegated(_from, _operator);
+        emit OperatorStaked(_operator, beneficiary, authorizer, _value);
     }
 
     /// @notice Performs top-up to an existing operator. Tokens added during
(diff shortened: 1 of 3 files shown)


pdyraga:
I think it makes sense but we should also consider removing the operator from the old owner to preserve information consistency.

solidity/contracts/TokenStaking.sol
```diff
@@ -42,17 +42,16 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     using Locks for Locks.Storage;
     using TopUps for TopUps.Storage;
 
-    event Staked(
-        address owner,
+    event StakeDelegated(
+        address indexed owner,
+        address indexed operator
+    );
+    event OperatorStaked(
```
pdyraga, line 49: Not sure about the verbiage but couldn't find anything better than `StakeDelegated` + `OperatorStaked`.

r-czajkowski: makes sense to me 👍

solidity/contracts/TokenStaking.sol
```diff
@@ -565,7 +564,7 @@ contract TokenStaking is Authorizations, StakeDelegatable {
     function transferStakeOwnership(address operator, address newOwner) public {
         require(msg.sender == operators[operator].owner, "Not authorized");
         operators[operator].owner = newOwner;
-        emit StakeOwnershipTransferred(operator, newOwner);
+        emit StakeDelegated(newOwner, operator);
```
nkuba, line 567: I'm not sure about using the same event in two flows. The previous `StakeOwnershipTransferred` event made here more sense to me. It may cause confusion because this function is not delegating the stake in a usual way. Why did we decide to change the previous approach?

solidity/contracts/TokenStaking.sol
```diff
@@ -193,7 +191,8 @@ contract TokenStaking is Authorizations, StakeDelegatable {
             _extraData
         );
 
-        emit Staked(_from, _operator, beneficiary, authorizer, _value);
+        emit StakeDelegated(_from, _operator);
+        emit OperatorStaked(_operator, beneficiary, authorizer, _value);
```
nkuba, line 195: Are we using two subsequent events because of the limitation of max three indexed parameters in an event?

nkuba, line 195: Did we update unit tests of `delegate` function to cover that both events are emitted?

pdyraga: Yes. No more than three indexed fields and we need to be able to map between `operator/authorizer/beneficiary`. Given that those three values never change, it probably makes more sense to keep them together in the event and have the `owner` indexed in a separate one.

pdyraga: We do not have any tests for emitted events there right now and writing them is quite complex I believe. We are having `approveAndCall` and `receiveApproval` and OpenZeppelin's `expectEvent` is not catching them.

solidity/contracts/TokenStaking.sol
pdyraga: I think I was just overprotective in regard to the contract size.

Instead of:
- Filter out `StakeDelegated` events by owner field
- Filter out `StakeDelegated` events by operator field, checking more recent events to make sure the delegation ownership has not been transferred.

We can:
- Filter out `StakeDelegated` events by owner field
- Filter out `StakeOwnershipTransferred` events by operator field, checking more recent events to make sure the delegation ownership has not been transferred.

pdyraga: See https://github.com/keep-network/keep-core/pull/1990/commits/681a8ba752ab5d5170d36337d43e10a0ea200f64 please.

solidity/contracts/TokenStaking.sol
nkuba: Can we use `getPastEvents` for verification?
```js
const startBlock = await web3.eth.getBlock("latest").number

// await delegate

const events = stakingContract.getPastEvents("OperatorStaked", {fromBlock: startBlock })

// assert length of events equals 1 and check events[0] details
```

pdyraga: Good idea! Please see https://github.com/keep-network/keep-core/pull/1990/commits/6111cc722fae61312e7c20b4cc880697994c99cc.

r-czajkowski:
looks good to me 👍

nkuba (approved):

PR #1991, opened 2020-08-14
Gas price ceiling updates

pdyraga:
Our 60Gwei gas price ceiling set in https://github.com/keep-network/keep-core/pull/1914 is not enough given the current gas price trends. 👨‍🌾 👩‍🌾 . 60Gwei ceiling we considered a bold move one month ago is not even 50% of today's current average gas price of 200Gwei, not even speaking about average 367Gwei price yesterday.

Until we come up with a pricing model taking gas price fluctuations into account we need to have a way to update gas price ceiling. That's what's implemented here.

`GasPriceOracle` is an ownable contract allowing the owner to modify the current gas price value with respect to 6 hours governance delay. The oracle notifies all subscribed contracts when the change is finalized and lets them pull the most recent gas price.

When working on this change I had to drop `removeServiceContract` function from `KeepRandomBeaconOperator` to save some bytecode and not exceed the maximum allowed contract size. There was little to no benefit of having this function. There are two places we guard by `onlyServiceContract` modifier: function triggering group selection and function requesting for a new entry signature. Both functions are designed and implemented in a way that is secure for staker and operator contract state even if the service contract gets compromised. It is quite unlikely we'll have to upgrade service contract at all but even if we do, we don't need to remove it from the allowed service contracts list.

### DIFF
--- a/solidity/contracts/GasPriceOracle.sol
+++ b/solidity/contracts/GasPriceOracle.sol
@@ -34,7 +34,7 @@ contract GasPriceOracle is Ownable {
 
     event GasPriceUpdated(uint256 newValue);
 
-    uint256 public constant governanceDelay = 6 hours; 
+    uint256 public constant governanceDelay = 1 hours; 
 
     uint256 public gasPrice;
 
--- a/solidity/contracts/GasPriceOracle.sol
+++ b/solidity/contracts/GasPriceOracle.sol
@@ -1,0 +1,99 @@
+/**
+▓▓▌ ▓▓ ▐▓▓ ▓▓▓▓▓▓▓▓▓▓▌▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▄
+▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▌▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
+  ▓▓▓▓▓▓    ▓▓▓▓▓▓▓▀    ▐▓▓▓▓▓▓    ▐▓▓▓▓▓   ▓▓▓▓▓▓     ▓▓▓▓▓   ▐▓▓▓▓▓▌   ▐▓▓▓▓▓▓
+  ▓▓▓▓▓▓▄▄▓▓▓▓▓▓▓▀      ▐▓▓▓▓▓▓▄▄▄▄         ▓▓▓▓▓▓▄▄▄▄         ▐▓▓▓▓▓▌   ▐▓▓▓▓▓▓
+  ▓▓▓▓▓▓▓▓▓▓▓▓▓▀        ▐▓▓▓▓▓▓▓▓▓▓         ▓▓▓▓▓▓▓▓▓▓▌        ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
+  ▓▓▓▓▓▓▀▀▓▓▓▓▓▓▄       ▐▓▓▓▓▓▓▀▀▀▀         ▓▓▓▓▓▓▀▀▀▀         ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▀
+  ▓▓▓▓▓▓   ▀▓▓▓▓▓▓▄     ▐▓▓▓▓▓▓     ▓▓▓▓▓   ▓▓▓▓▓▓     ▓▓▓▓▓   ▐▓▓▓▓▓▌
+▓▓▓▓▓▓▓▓▓▓ █▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+
+                           Trust math, not hardware.
+*/
+
+pragma solidity 0.5.17;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+/// @dev Interface expected to be implemented by contracts added as a gas price
+/// oracle consumers. Consumers are notified by GasPriceOracle every time gas
+/// price change is finalized by calling their refreshGasPrice function. Each
+/// consumer may decide to pull the new value from the oracle immediatelly or
+/// at the moment right for the consumer.
+interface GasPriceOracleConsumer {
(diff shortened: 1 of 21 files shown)


solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -218,14 +216,18 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
...
     }
 
-    /// @notice Removes service contract
-    /// @param serviceContract Address of the service contract.
-    function removeServiceContract(address serviceContract) public onlyServiceContractUpgrader {
-        serviceContracts.removeAddress(serviceContract);
+    /// @notice Pulls the most recent gas price from gas price oracle.
+    function refreshGasPrice() public {
+        gasPriceCeiling = gasPriceOracle.gasPrice();
     }
```
pdyraga, line 231: I am still not sure if we should use pull or push model here.

The benefit of the current solution is that we don't need to have an additional require with `msg.sender` and it's a little less developer-error-prone because of that. Also, the contract may decide whether it wants to pull the most recent price at this moment. I think it's similar to `SatWeiPriceFeed` in tBTC and MKR oracle.

On the other hand, `refreshGasprice(uint256 value) onlyGasPriceOracle` may look a little easier and consume less gas. I don't think we can have `onlyGasPriceOracle` modifier given the contract size problems but we probably have enough space to do:
```
    function refreshGasPrice(uint256 newPrice) public {
        require(msg.sender == address(gasPriceOracle), "Not authorized");
        gasPriceCeiling = newPrice;
    }
```

Really interested in reviewer's thoughts! 🙌

solidity/contracts/GasPriceOracle.sol
```diff
@@ -0,0 +1,99 @@
...
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+/// @dev Interface expected to be implemented by contracts added as a gas price
+/// oracle consumers. Consumers are notified by GasPriceOracle every time gas
+/// price change is finalized by calling their refreshGasPrice function. Each
+/// consumer may decide to pull the new value from the oracle immediatelly or
+/// at the moment right for the consumer.
```
Shadowfiend, line 25: I guess it's probably worth flagging that these consumers are trusted contracts whose `refreshGasPrice` must always succeed and never consume excessive gas?

pdyraga: Yes, very good point! I'll add it to docs.

solidity/contracts/KeepRandomBeaconOperator.sol
Shadowfiend: This seems fine as-is, honestly.

nkuba: >Also, the contract may decide whether it wants to pull the most recent price at this moment.

Wouldn't it be pulled to the operator contract on each update by `finalizeGasPriceUpdate`? What I mean is that the contract will not have a chance to be behind the current oracle's value. Which is fine I guess. Unless we want to have two versions of `finalizeGasPriceUpdate` where we call refresh or not.

I'm fine with the current solution! Just wanted to clarify with y'all that my understanding of how things work is correct.

solidity/test/TestGasPriceOracle.js
```diff
@@ -0,0 +1,175 @@
...
+
+const BN = web3.utils.BN
+const chai = require('chai');
+chai.use(require('bn-chai')(BN))
+const expect = chai.expect
+const assert = chai.assert
+
+describe("GasPriceOracle", () => {
+
+    const owner = accounts[0]
```
nkuba, line 17: Let's use non-default account.
```suggestion
    const owner = accounts[1]
```

solidity/contracts/GasPriceOracle.sol
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/// @notice Oracle presenting the current gas price. The oracle is manually
+/// updated by its owner.
+contract GasPriceOracle is Ownable {
+    using SafeMath for uint256;
+
+    event GasPriceUpdated(uint256 newValue);
+
+    uint256 public constant governanceDelay = 6 hours;
```
nkuba: Don't we want to configure this value on deployment? Probably not, but thought I should ask 😄

solidity/test/TestGasPriceOracle.js
```diff
@@ -0,0 +1,175 @@
...
+            const newValue = web3.utils.toBN(55555)
+            await oracle.beginGasPriceUpdate(newValue, {from: owner})
+            await time.increase(time.duration.minutes(361)) // 6 hours and 1 minute
+            const receipt = await oracle.finalizeGasPriceUpdate()
+            await expectEvent(receipt, "GasPriceUpdated", {
+                newValue: newValue
+            })
+        })
+    
+        it("notifies consumer contracts when finalizing the change", async () => {
```
nkuba, line 93: Can we add a case where we call `beginGasPriceUpdate` just before `governanceDelay` passes so we can check that the new value gets overwritten and the timer is set up again?

solidity/contracts/GasPriceOracle.sol
```diff
@@ -0,0 +1,99 @@
...
+        require(index < consumerContracts.length, "Invalid index");
+        consumerContracts[index] = consumerContracts[consumerContracts.length - 1];
+        consumerContracts.length--;
+    }
+
+    /// @notice Returns all consumer contracts currently registered in the
+    /// oracle.
+    function getConsumerContracts() public view returns (address[] memory) {
+        return consumerContracts;
+    }
```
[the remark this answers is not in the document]
pdyraga: With public `consumerContracts` we can do:
```
await oracle.consumers(0)
await oracle.consumers(1)
```

but we can't do `await oracle.consumers()`. 

I do not know a way of getting the number of contracts in this array without having `getConsumerContracts()`. Also having `consumerContracts` public and having this function let's us fetch all of them at once or fetch them individually if needed.

solidity/contracts/KeepRandomBeaconOperator.sol
```diff
@@ -172,15 +167,18 @@ contract KeepRandomBeaconOperator is ReentrancyGuard {
...
     ) public {
-        registry = KeepRegistry(_registryContract);
+        registry = KeepRegistry(_keepRegistry);
 
         serviceContracts.push(_serviceContract);
-        stakingContract = TokenStaking(_stakingContract);
+        stakingContract = TokenStaking(_tokenStaking);
 
-        groups.stakingContract = TokenStaking(_stakingContract);
+        gasPriceOracle = GasPriceOracle(_gasPriceOracle);
```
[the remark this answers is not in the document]
pdyraga: Sure!

solidity/test/TestGasPriceOracle.js
pdyraga: Done in https://github.com/keep-network/keep-core/pull/1991/commits/cc36463d1c83549fd4584481aec257f505af949b

solidity/contracts/KeepRandomBeaconOperator.sol
pdyraga: `KeepRandomBeacon` implements `refreshGasPrice` in a way that is pulling the change immediately in the same transaction where `finalizeGasPriceUpdate` is called, that's correct. What I wanted to say is that with the current interface we may have another contract added as the oracle's consumer which - in some circumstances - may decide to not update the price immediately.

pdyraga: Well, actually not updating the price immediately would be also possible with push model of `refreshGasPrice(uint256 newPrice)` as long as the new price is public on the oracle. I think the main advantage of the current solution is that there is harder to make a mistake in a consumer contract and let anyone update the price if the appropriate ACL is not in place. With this solution, the change is pulled from the oracle contract so no ACL is needed.

solidity/contracts/GasPriceOracle.sol
pdyraga: I think we can have it hardcoded. From testing perspective, it's nice to pass those values to constructor. On the other hand, having them hardcoed let's use write unit tests against real values and there is no chance we'll make a mistake during deployment.

PR #1993, opened 2020-08-15
dev-daramir/ dashboard update keep network deps

daramir:
# Description
This is an update to the references on `solidity/dashboard/package.json` dependencies, to the latest ones deployed on Ropsten.
Added a short getting started to running local against Ropsten in the Readme 

## Checklist
- [X] Fork keep-network/keep-core
- [X] Clone your fork
- [X] Describe the change you are intending to undertake in the PR description.

Before marking the PR as ready for review, make sure:
- [X] It passes the linter checks (npm run lint) (see Pre-commit to make this automatic).
Linter `npm run js:lint` ran successfully
- [X] It passes the continuous integration tests on CircleCI.
I don't have access to this CircleCI Org
- [X] Your changes have sufficient test coverage (e.g regression tests have been added for bug fixes, unit tests for new features)
- [X] Commits must be signed.


solidity/dashboard/package.json
```diff
@@ -5,9 +5,9 @@
   "license": "MIT",
   "dependencies": {
     "@0x/subproviders": "^6.0.8",
-    "@keep-network/keep-core": ">1.2.4-rc <1.2.4",
-    "@keep-network/keep-ecdsa": ">1.1.2-rc <1.1.2",
-    "@keep-network/tbtc": ">1.0.2-rc <1.0.2",
+    "@keep-network/keep-core": ">1.3.0-rc <1.3.0",
+    "@keep-network/keep-ecdsa": ">1.2.0-rc <1.2.0",
+    "@keep-network/tbtc": ">1.1.0-rc <1.1.0",
```
Shadowfiend, line 10: These bumps are in the RC branches, which have now landed on master.

solidity/dashboard/README.md
```diff
@@ -75,6 +75,13 @@ When you don’t want to use the local version of `@keep-network/keep-core` anym
 
 A new version of staking dApp is automatically deployed to `keep-dev` internal testnet after each `master` merge. dApp can be accessed at [https://dashboard.dev.keep.network/](https://dashboard.dev.keep.network/) and requires an initial setup in MetaMask before the first use. All the setup described below has to be done only one time. 
 
+### Ropsten testnet
+
+Change:
+In `solidity/dashboard/src/connectors/utils.js`
+- getChainId() return value to 3. As indicated, private chains ia the default, so you need to change if you use a different one (Ropsten = 3).
```
Shadowfiend, line 82: `getChainId` is deriving from the artifact dependencies, so it should not be hardcoded. It will be correct as long as the `package.json` points to the correct artifacts.

solidity/dashboard/README.md
```diff
@@ -75,6 +75,13 @@ When you don’t want to use the local version of `@keep-network/keep-core` anym
...
+- getWsUrl() return value to your Infura websocket URL (strongly recommend to get an Infura account if you don't have one), or find a websocket URL that connects to Ropsten and put it here.
```
Shadowfiend, line 83: `getwsUrl` is best updated @ https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/config/config.json, though you shouldn't need to update it.

solidity/dashboard/package.json
daramir: I've done an upstream sync. Resolving.

solidity/dashboard/README.md
daramir: Understood! I've modified the wording. this would now be part of point 2. Marked as optional and if desired should be done in the location suggested (i.e. `solidity/dashboard/src/config/config.json`)

solidity/dashboard/README.md
daramir: Here's a snippet of `getChainId()` [src](https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/connectors/utils.js#L25-L32)
```javascript
export const getChainId = () => {
  if (process.env.NODE_ENV === "development") {
    // private chains (default), change if you use a different one
    return 1337
  }
  // For KEEP internal testnet, ropsten and mainnet `chainId == networkId`
  return Number(getFirstNetworkIdFromArtifact())
}
```
When we run with `npm start` it runs in development mode, hence NODE_ENV would cause this to return the hard-coded value. I've explained this a bit more in https://github.com/daramir/keep-core/blob/47254ca5f2aca3fe615c6e2b2fcd00c5cc859959/solidity/dashboard/README.md#L80 but I see why this doesn't sound great. Should the:
- Readme specify that it's 1337=>3 and that none this changes shouldn't be committed?
- logic of `getChainId` change so that it doesn't depend on NODE_ENV or use a hard-coded value?

lionakhnazarov:
Closing — dashboard dependency update; the dashboard has been extracted to its own repository. No longer applicable here.

Issue #491, opened 2019-01-08
New interface for EventSubscriptions


rargulati:
PR #474 introduces a way for us to unsubscribe from callbacks. Typically our callbacks are attached to long running processes (ie. Ethereum Watchers: goroutines which parse log files and signal back on a channel). Unfortunately, in the implementation, there's quite a bit of code duplication (generics!) and the unsubscribe isn't deferred.

@Shadowfiend proposed the following:

```
type <T>EventSubscription interface {
    func OnEvent(func (T)) EventSubscription // unsubscribe just runs the unsubscribe
    func OnEventContext(func (T), context *context.Context) // unsubscribe unsubscribes + handles context
    func Pipe(chan T) EventSubscription // unsubscribe does the unsubscribe + closes chan
    func PipeContext(chan T, context *context.Context) // unsubscribe unsubscribes + closes chan + handles context
}

type EventSubscription interface {
    func Unsubscribe()
}

type Chain interface {
    ...
    DKGResultPublication() DKGResultPublicationEventSubscription
}

subscription := chainRelay.DKGResultPublication().OnEvent(...)
defer subscription.Unsubscribe()

publishChan := make(chan *event.DKGResultPublication)
subscription := chainRelay.DKGResultPublication().Pipe(publishChan)
defer subscription.Unsubscribe()
```

This presents a more ideal way forward (less duplication, deferrable unsubscribes, etc). I attempted this with a bit of code generation, but gave up half way through (time). Let's pick this up again.

lionakhnazarov:
Closing as stale. Targets the legacy keep-core v1 (random beacon / KEEP staking) architecture, superseded by the Threshold v2 stack. Not actionable against current main.

PR #1996, opened 2020-08-19
Add GoSec workflow

lukasz-zimnoch:
Closes: #2007

Here we introduce the [GoSec](https://github.com/securego/gosec) scan workflow to the Github Actions pipeline and deal with all discovered problems.

Summary of `#nosec` exceptions and their motivation:

1. https://github.com/keep-network/keep-core/blob/4bc88c6185b9df3b904733270f6ac091af45fc89/config/config.go#L18
Rule G101: Look for hardcoded credentials
Motivation: This line doesn't contain any credentials. It's just the name of the environment variable.

2. https://github.com/keep-network/keep-core/blob/4bc88c6185b9df3b904733270f6ac091af45fc89/pkg/chain/local/local.go#L508
Rule G404: Insecure random number source (rand)
Motivation: Local chain implementation doesn't require secure randomness. 

3. https://github.com/keep-network/keep-core/blob/4bc88c6185b9df3b904733270f6ac091af45fc89/pkg/net/local/identity.go#L26
Rule G404: Insecure random number source (rand)
Motivation: Local network identity doesn't require secure randomness.


pkg/beacon/relay/dkg/result/publish.go
```diff
@@ -16,9 +16,15 @@ import (
...
-	channel.RegisterUnmarshaler(func() net.TaggedUnmarshaler {
+	err := channel.RegisterUnmarshaler(func() net.TaggedUnmarshaler {
 		return &DKGResultHashSignatureMessage{}
 	})
+	if err != nil {
+		logger.Errorf(
+			"could not register DKGResultHashSignatureMessage unmarshaller: [%v]",
+			err,
+		)
+	}
```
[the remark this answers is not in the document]
pdyraga: Instead of capturing `err` from `RegisterUnmarshaler` we should go into `RegisterUnmarshaler` and change the function signature so that it does not return `error`. We changed the behavior a long time ago and this function never fails. I think we planned to do the cleanup in a followup PR but that followup PR never happened. We are now having a situation that `ReigsterUnmarshaler` always return `nil` error.

pkg/beacon/relay/gjkr/gjkr.go
```diff
@@ -19,27 +19,75 @@ var logger = log.Logger("keep-gjkr")
...
+
+	err = channel.RegisterUnmarshaler(func() net.TaggedUnmarshaler {
 		return &MisbehavedEphemeralKeysMessage{}
 	})
+	if err != nil {
+		logger.Errorf(
+			"could not register MisbehavedEphemeralKeysMessage unmarshaller: [%v]",
+			err,
+		)
+	}
```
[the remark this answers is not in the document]
pdyraga: > it's also not clear to me when RegisterUnmarshaler would fail, so that could be part of it

It's never failing, heh. We refactored it a long time ago as a part of a bigger channel-related change and aimed to clean up the code calling it in a followup PR and that followup PR never happened.

pkg/internal/entrytest/entrytest.go
```diff
@@ -107,7 +107,7 @@ func executeSigning(
 	}
 
 	entrySubmissionChan := make(chan *event.EntrySubmitted)
-	chain.ThresholdRelay().OnRelayEntrySubmitted(
+	_, _ = chain.ThresholdRelay().OnRelayEntrySubmitted(
```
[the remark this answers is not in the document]
pdyraga: The thing is that we can't have an error here, the interface [is not very precise](https://github.com/keep-network/keep-core/pull/1996#discussion_r475438385) 😬

pkg/internal/dkgtest/dkgtest.go
```diff
@@ -116,7 +116,7 @@ func executeDKG(
 	}
 
 	resultSubmissionChan := make(chan *event.DKGResultSubmission)
-	chain.ThresholdRelay().OnDKGResultSubmitted(
+	_, _ = chain.ThresholdRelay().OnDKGResultSubmitted(
```
pdyraga, line 119: Let's drop `error` from the returned parameters of `OnDKGResultSubmitter`, just like we did for all functions of `RelayEntryInterface`.

pkg/internal/entrytest/entrytest.go
```diff
@@ -107,11 +107,14 @@ func executeSigning(
 	}
 
 	entrySubmissionChan := make(chan *event.EntrySubmitted)
-	chain.ThresholdRelay().OnRelayEntrySubmitted(
+	_, err = chain.ThresholdRelay().OnRelayEntrySubmitted(
```
pdyraga: Let's remove the `error` from the interface of `OnRelayEntrySubmitted`. Let's keep it inside of `ethereum.go`.

pkg/net/libp2p/channel.go
```diff
@@ -326,7 +325,7 @@ func (c *channel) SetFilter(filter net.BroadcastChannelFilter) error {
 	c.pubsubMutex.Lock()
 	defer c.pubsubMutex.Unlock()
 
-	c.pubsub.UnregisterTopicValidator(c.name)
+	_ = c.pubsub.UnregisterTopicValidator(c.name)
```
pdyraga, line 328: Let's log this error. Agreed, we should never have this problem but better be safe than sorry.

lukasz-zimnoch:
@pdyraga all comments addressed.

pkg/chain/local/local.go
```diff
@@ -169,7 +169,7 @@ func (c *localChain) SubmitRelayEntry(newEntry []byte) *async.EventEntrySubmitte
 
 	currentBlock, err := c.blockCounter.CurrentBlock()
 	if err != nil {
-		relayEntryPromise.Fail(fmt.Errorf("cannot read current block"))
+		_ = relayEntryPromise.Fail(fmt.Errorf("cannot read current block"))
```
pdyraga, line 172: Let's log errors from failed promises in this file. It's a lot of boilerplate code but it's better than ignoring those errors.

pkg/chain/local/local.go
```diff
@@ -336,7 +336,7 @@ func (c *localChain) IsStaleGroup(groupPublicKey []byte) (bool, error) {
 	defer c.handlerMutex.Unlock()
 
 	bc, _ := BlockCounter()
-	bc.WaitForBlockHeight(c.simulatedHeight)
+	_ = bc.WaitForBlockHeight(c.simulatedHeight)
```
pdyraga, line 339: Let's log the error here.

pdyraga:
Some additional explanation about `On*` function changes in the chain interface: 

We decided to update signatures of all `On*` functions to do not return an error and log the possible error from `Watch*` inside `ethereum.go`. `On*` and `Watch*` never return an error. `Watch*`, in case of an error, tries to resubscribe until the subscription is canceled. Refactoring `Watch*` function signature is out of the scope of this PR and should be done when working on https://github.com/keep-network/keep-core/issues/491 but we could at least do some cleanup and not let the incorrect interface leak out of `ethereum.go`.

pdyraga (approved):

PR #2014, opened 2020-08-25
Optimize chain config fetching

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-ecdsa/issues/524

During the work on the rate limiter, we realized a significant slowdown of the relay entry generation process when the rate limiter was enabled with relatively small limits. We investigated the problem and discovered it is caused by the chain config fetch process which contains several contract calls and occurs for each member before signing the relay entry. Because all but one config parameters are invariants, there was no sense to fetch the configuration multiple times and send pointless calls against the Ethereum node. Hence, we provide some fixes to solve the problem:

1. Because `MinimumStake` parameter was the only field which may change over time (because of the minimum stake schedule), we extract it away of the `config.Chain` struct and fetch it independently where needed.
2. We fetch the entire chain config during the `ethereumChain` setup once and then just return it through the `GetConfig` method.


pkg/chain/ethereum/connect.go
```diff
@@ -6,6 +6,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/keep-network/keep-core/pkg/beacon/relay/config"
```
pdyraga, line 9: I looked at all the files with this import and all of them but `connect.go` reference `"github.com/keep-network/keep-core/pkg/beacon/relay/chain"` as well. I wonder if it makes sense to have a `config` package with `chain.go` inside. Perhaps we should rename `Chain` to `Config` and move it to `chain.go` in `pkg/beacon/relay/chain` and remove `pkg/beacon/relay/chain` altogether. What do you think?

pdyraga (approved):

PR #2018, opened 2020-08-27
Add Slither workflow

lukasz-zimnoch:
Here we introduce Slither scan as a new Github Actions workflow.


pdyraga:
No longer relevant. We have slither jobs configured for v2 code.

Issue #2011, opened 2020-08-22
Ledger Live HD Path Addresses not lining up


afmsavage:
Hi not sure if this is me being crazy or something is different between your implementation and Ledger's, but I am looking in both Ledger Live and Metamask's "Ledger Live" section for hardware wallets, and the addresses on the Dashboard (Mainnet) do not line up with the addresses in those other two apps.  Here are some screenshots to show you.  It appears that 0/0/1 lines up fine, but then the rest do not.

Metamask/Ledger Live
![image](https://user-images.githubusercontent.com/29739267/90947782-35152d00-e3fe-11ea-97db-c56264c7fff2.png)

KEEP Dashboard's Ledger Live
![image](https://user-images.githubusercontent.com/29739267/90947815-7d344f80-e3fe-11ea-8918-4f804039972a.png)

Additional Addresses on MyEtherWallet showing the same as in Metamask
![image](https://user-images.githubusercontent.com/29739267/90947842-ae148480-e3fe-11ea-8e69-5e49f9ed3949.png)

pdyraga:
Thanks for reporting it @afmsavage.

PR #2025, opened 2020-08-31
Ledger live derivation path fix

r-czajkowski:
Closes: #2011

This PR fixes the derivation path for Ledger Live accounts. The patter for Ledger Live should be `m/44'/60'/x'/0/0` instead of `m/44'/60'/0'/0/x`. The default implementation of `LedgerSubprovider` from `0x/subproviders` package doesn't support the derivation path for Ledger Live accounts.

Ledger Live
obraz
obraz
obraz
obraz


dimpar (approved):
Tested Ledger Live & Ledger Legacy. Compared with MetaMask accounts when connecting a hardware wallet. Results were the same.

PR #1989, opened 2020-08-14
Copy stake flow

r-czajkowski:
Ref: #1944 - just ref because we also need to allow users to `recover` old delegations and repay the `StakingPortBacker` contract.
Ref: #2022 

This PR adds a copy-stake flow to the KEEP token dashboard.

To fetch "old" delegations we need the old `TokenStaking` artifact- this will be resolved in a future PR. Currently, for local development, we need to copy the `OldTokenStaking` artifact to the `src/old-contracts-artifacts` in the dapp source code.

As a first step, a user needs to choose one of the old `TokenStaking` delegations. Only **non-copied** delegations and **delegations that were undelegated before the `StakingPortBacker` contract was deployed, are displayed**.

The user is able to choose one of the flow:
* copy-stake(recommended)- the 2 transactions may be required. First `copyStake` on the `StakingPortBacker` contract and then `undelegate` on the old `TokenStaking`,
* undelegate- the dapp calls `undelegate` on the old `TokenStaking` contract.

<details>
  <summary>Screenshoots</summary>

obraz

obraz

obraz

obraz

obraz

obraz

obraz

obraz

</details>

PR #2028, opened 2020-09-01
Fix fetching past events

r-czajkowski:
If we want to get past events eg. by all operators, and pass an empty array to indexed param eg. `{filter: { operator: [] }}` to the [`getPastEvent`](https://web3js.readthedocs.io/en/v1.2.0/web3-eth-contract.html#getpastevents), all emitted events will be returned. To avoid this, before calling `getPastEvent` with an empty array for the indexed param, we check that the filter param is not an empty array and then call ` getPastEvent` function.

### DIFF
--- a/solidity/dashboard/src/services/tbtc-rewards.service.js
+++ b/solidity/dashboard/src/services/tbtc-rewards.service.js
@@ -11,6 +11,7 @@ import {
 } from "../contracts"
 import web3Utils from "web3-utils"
 import { isSameEthAddress } from "../utils/general.utils"
+import { isEmptyArray } from "../utils/array.utils"
 
 const fetchTBTCRewards = async (web3Context, beneficiaryAddress) => {
   const transferEventSearchFilter = {
@@ -25,13 +26,16 @@ const fetchTBTCRewards = async (web3Context, beneficiaryAddress) => {
     transferEventSearchFilter
   )
 
+  const depositCreatedFilterParam = isEmptyArray(transferEventToBeneficiary)
+    ? {}
+    : {
+        _depositContractAddress: transferEventToBeneficiary.map(
+          (_) => _.returnValues.from
+        ),
+      }
   const depositCreatedSearchFilter = {
     fromBlock: CONTRACT_DEPLOY_BLOCK_NUMBER[TBTC_SYSTEM_CONTRACT_NAME],
-    filter: {
-      _depositContractAddress: transferEventToBeneficiary.map(
-        (_) => _.returnValues.from
-      ),
-    },
+    filter: depositCreatedFilterParam,
   }
 
   const depositCreatedEvents = await contractService.getPastEvents(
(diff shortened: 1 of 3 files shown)


solidity/dashboard/src/services/tbtc-rewards.service.js
```diff
@@ -25,13 +26,16 @@ const fetchTBTCRewards = async (web3Context, beneficiaryAddress) => {
     transferEventSearchFilter
   )
 
+  const depositCreatedFilterParam = isEmptyArray(transferEventToBeneficiary)
```
pdyraga, line 29: If my understanding is correct, I think the name of this `const` is a little misleading as we are not talking about created deposits but rather about deposits that have been successfully closed. This is out of scope of this PR though, so let's leave it as-is here.

solidity/dashboard/src/services/token-staking.service.js
```diff
@@ -218,12 +219,15 @@ export const getOperatorsOfOwner = async (owner) => {
...
-    })
-  ).reduce(reduceByOperator, {})
+  let transferEventsByOperators = {}
+  if (!isEmptyArray(operators)) {
+    transferEventsByOperators = (
+      await stakingContract.getPastEvents("StakeOwnershipTransferred", {
+        fromBlock: CONTRACT_DEPLOY_BLOCK_NUMBER.stakingContract,
+        filter: { operator: operators },
+      })
+    ).reduce(reduceByOperator, {})
```
pdyraga, line 229: I need help to understand it. Say we have `StakeOwnershipTransferred(operator = X, newOwner = A)` and then `StakeOwnershipTransferred(operator = X, newOwner = B)`. What we are doing here is that we are eliminating `newOwner = A` event, correct?

pdyraga:
Heads up @nkuba, we will have to backport this change to https://github.com/keep-network/keep-core/pull/1989 you are reviewing.

pdyraga:
Tested with multiple grants and delegations, works as advertised.

solidity/dashboard/src/services/tbtc-rewards.service.js
r-czajkowski: We scan `Created` event from `TBTCSystem` contract so `depositCreated...` it's ok I think.

solidity/dashboard/src/services/token-staking.service.js
r-czajkowski: yep, exactly. I'll try to add more color here in a moment

r-czajkowski: As a first step, we scan `StakeDelegated` by `owner` to get all operators. We can't 100% sure this relationship is still active since the owner can be changed.

Next, we need to scan `StakeOwnershipTransferred` by `newOwner` because another owner may have changed `owner` of a delegation to us. But again we can't 100% sure this relationship is still active since the owner can be changed.

So as the last step, we scan `StakeOwnershipTransferred` by all operators from these two results ^ and check the most recent `StakeOwnershipTransferred` event for a given operator to make sure we are the owner of the delegation.

pdyraga (approved):

PR #2031, opened 2020-09-01
Voluminous Containers: Fix initcontainers that require keyfile

Shadowfiend:
The provisioning script used in init containers recently moved to
reading Ethereum key and account information from a keyfile rather than
from the environment; however, the corresponding Kubernetes stateful set
and cron job configurations were not updated to mount the keyfile for
the initcontainer (it was already being mounted for the container
proper).

The initcontainer now has the keyfile mounted properly. In the process,
moved volumes to the top of the spec, since they are now used in both of
the subsequent sections.

This state is currently deployed on keep-test for client-2 and client-3,
as well as the relay requester.

### DIFF
--- a/infrastructure/kube/keep-dev/keep-client-0-statefulset.yaml
+++ b/infrastructure/kube/keep-dev/keep-client-0-statefulset.yaml
@@ -39,6 +39,19 @@ spec:
         type: beacon
         id: '0'
     spec:
+      volumes:
+      - name: keep-client-config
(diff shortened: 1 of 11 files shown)


Shadowfiend:
Tested this with relay requester and it worked. Rolling it out to the deployed stateful sets next, then will undraft.

pdyraga (approved):
Yes!

PR #2037, opened 2020-09-03
Linty faucet: Improved linting pre-commits, linting config for faucet, and a fix

Shadowfiend:
Three bits here:
- The faucet now has a proper `eslintrc`, based on the common Keep config.
- The faucet (and as a bonus the dashboard) now have linting pre-commit configurations (/cc @r-czajkowski ).
- The faucet now handles the newer error structures we get from web3 (this was an oversight from a web3 version bump).

See #1551.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,8 +22,20 @@ repos:
       additional_dependencies:
       - ethlint@1.2.3
     - id: lint-js
-      name: 'lint js'
+      name: 'lint solidity js'
       entry: /usr/bin/env bash -c "cd solidity && npm run lint:js"
-      files: '\.js$'
+      files: 'solidity\/.*\/\.js$'
+      language: script
+      description: "Checks JS code according to the package's linter configuration"
+    - id: lint-dashboard-js
+      name: 'lint dashboard js'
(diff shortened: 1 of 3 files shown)


nkuba (dismissed):

.pre-commit-config.yaml
```diff
@@ -22,8 +22,20 @@ repos:
...
+      name: 'lint solidity js'
       entry: /usr/bin/env bash -c "cd solidity && npm run lint:js"
-      files: '\.js$'
+      files: 'solidity\/.*\/\.js$'
+      language: script
+      description: "Checks JS code according to the package's linter configuration"
+    - id: lint-dashboard-js
+      name: 'lint dashboard js'
+      entry: /usr/bin/env bash -c "cd solidity/dashboard && npm run js:lint"
+      files: 'solidity\/.*\/\.js$'
```
nkuba, line 33: Don't we need to update this filter to respect the dashboard's file structure?

Shadowfiend: … I swear I'd set this one up right 😅

Shadowfiend: Should be fixed :grimacing:

nkuba (approved):

PR #2042, opened 2020-09-08
DefaultMaxGasPrice used by client set to 500 Gwei

pdyraga:
image

Source: https://etherscan.io/chart/gasprice

The last weeks brought an extreme gas price increase on mainnet, even to 500 Gwei average price. The default value of 70 Gwei could not be enough.

The `MaxGasPrice` can be still configured in the config file, here we just update the default value to keep it aligned with the reality.

### DIFF
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -19,7 +19,7 @@
 	# allowed gas price is reached, no further resubmission attempts are
 	# performed.
 	#
-	# MaxGasPrice = 70000000000 # 70 gwei (default value)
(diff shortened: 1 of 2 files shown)


nkuba (approved):
Should we consider automating obtaining this value at some point instead of hardcoding it, e.g. pulling from a contract or any other source?

pdyraga:
> Should we consider automating obtaining this value at some point instead of hardcoding it, e.g. pulling from a contract or any other source?

What we could be doing is to set the max price to be e.g. `N` (set in the config) times the average computed based on the previous blocks during the estimation. I am not yet sure what's better.

Issue #1992, opened 2020-08-15
Dashboard: List of slashing events all show amount slashed, but does not match balance


suprnrdy:
Multiple slashes are shown for a signing group and each one show's -1000 KEEP. This can be interpreted that my node is being slashed multiple times per signing group, but the total balance is showing me that I'm being slashed for each group, not each time there wasn't enough members participating. 

This can be seen for the Operator: 0x2e9c9d164e67E98204D13E5b6b7F8E61cD52e8Ef

<img width="1404" alt="Screen Shot 2020-08-14 at 11 14 26 PM" src="https://user-images.githubusercontent.com/1414501/90306672-66cd4780-de84-11ea-8e02-80b046e9afc8.png">

PR #2046, opened 2020-09-10
Fix fetching slashed tokens

r-czajkowski:
Closes: #1992

We should skip `UnauthorizedSigningReported\RelayEntryTimeoutReported` events if they were not emitted in the same transaction as the `TokenSlashed / TokenSeized` events because it means that these punishment events are not related to the provided operator.


dimpar:
Tested by creating a couple of new groups and reported unauthorized signing for them.
Slashed tokens were grouped by a group key for a specific punished group member.

dimpar (approved):

PR #2047, opened 2020-09-10
Skip data updates if its a copied delegation

r-czajkowski:
The dapp subscribes to the `StakeDelegated` event and updates data based on that event. If the `StakeCopied` event was emitted with a `StakedDelegated` event, we do not want to update grant data or KEEP token balance, since a copied delegations are from already staked tokens.


dimpar (approved):
lgtm

PR #2050, opened 2020-09-15
Format ETH amount on the Rewards page

r-czajkowski:
In this PR we display ETH amount rounded to 4 decimal places. In case where it's `0.0000`, we display `<0.0001 ETH` and then show the full amount in `Gwei` in a tooltip.

with formatting:
obraz

obraz

w/o formatting:
obraz


r-czajkowski:
@liz-shinn, Could you please take a look at this?

liz-shinn:
@r-czajkowski checking it out now!

Looks way better to have the numbers formatted. However, I wonder if it doesn't make more sense to show the full ETH amount with all the decimals in the tooltip. Is there a situation where someone would need to know the wei amount? Wei is a smaller unit of ETH, yes? We might want to denote that somewhere.

Also, I think what I may expect to happen here on hover is to hover and see something in ETH, since the hover is over the top of something in the unit of ETH. Wei is a smaller unit of ETH, yes? We might want to denote that somewhere.

And last thought, it would be good to make that a friendlier looking number by adding commas - 1,456,789,000 etc

pdyraga:
What happens if someone has less than 0.01 ETH accumulated in rewards for the group?

liz-shinn:
We could print the gwei/wei amount above and then show underneath the ETH amount, prevent a case where `0.00 ETH` is shown large without further context? Instead of in a tooltip.

Or, can we conditionally set it so it shows further decimal places if it's not within 0.001 ETH?

r-czajkowski:
> We could print the gwei/wei amount above and then show underneath the ETH amount, prevent a case where 0.00 ETH is shown large without further context? Instead of in a tooltip.

The main idea was to avoid displaying long values in a table row. That's why I decided to use a tooltip, since the long string is hidden. In a tooltip, we can also show `ETH` amount without rounding instead of `wei`.

If we display in `Gwei` or in `wei` we still have a long string Eg.
`0.001234560123345621 ETH` -> `1234560.123345621 Gwei` -> `1234560123345621 wei`

liz-shinn:
Further convo in FD here: https://www.flowdock.com/app/cardforcoin/tech/threads/KiSpaPANtnMCcWAGmXFp9fFWPA0

For situations where it's 0.00, we can show `<0.01 ETH` and then show the full amount either in the row underneath or in a tooltip. 

From @Shadowfiend  – 
>I wonder if the move is what you have right now with <0.01 ETH to deal with very small numbers, but with an explicit thing to click on that shows the hover with a full-precision gwei amount?

solidity/dashboard/src/utils/ethereum.utils.js
```diff
@@ -43,3 +45,19 @@ export const getEventsFromTransaction = async (contractToEventName, txHash) => {
 
   return events
 }
+
+export function displayEthAmount(
```
dimpar, line 49: I can't find an invocation of this method with parameters. Do you think we need these parameters here?

solidity/dashboard/src/components/Rewards.jsx
```diff
@@ -164,6 +167,10 @@ const RewardsComponent = ({ withdrawRewardAction }) => {
       : rewardsToReturn.slice(0, previewDataCount)
   }, [rewards, withdrawals, showAll, rewardFilter.status])
 
+  const amountTooltipText = (amount) => {
+    return `${displayEthAmount(amount, "Gwei", null)} Gwei`
```
Shadowfiend, line 171: I think `gwei` all lowercase is “most correct”.

Shadowfiend: Also, why pass `null` explicitly here?

solidity/dashboard/src/utils/ethereum.utils.js
```diff
@@ -46,14 +47,25 @@ export const getEventsFromTransaction = async (contractToEventName, txHash) => {
   return events
 }
 
+// We want to prevent long values being displayed in ETH unit(eg.
+// 0.000034560123345621 ETH). So in that case we want to display `<0.0001`.
+// More info here:
+// https://github.com/keep-network/keep-core/pull/2050#issuecomment-693434991
+export const MIN_ETH_AMOUNT_TO_DISPLAY_IN_WEI = "100000000000000" // 0.0001 ETH
```
Shadowfiend, line 54: This might be easier to understand if it was something like `toBN(10).pown(13)`.

solidity/dashboard/src/components/Rewards.jsx
r-czajkowski: > Also, why pass null explicitly here?

Because in the `displayEthAmount` function the latest param(`decimalsPlaces`) has default value(`4`). Displaying the amount in `gwei`, we want to get the full amount without rounding to decimal places.

r-czajkowski: > I think gwei all lowercase is “most correct”.

pls see 0bc3f45

solidity/dashboard/src/utils/ethereum.utils.js
r-czajkowski: pls see 6610f87

solidity/dashboard/src/components/Rewards.jsx
```diff
@@ -183,8 +190,11 @@ const RewardsComponent = ({ withdrawRewardAction }) => {
                 amount={totalRewardsBalance}
                 amountClassName="h1 text-primary"
                 displayAmountFunction={displayEthAmount}
-                withTooltip
-                tooltipText={`${totalRewardsBalance} wei`}
+                withTooltip={lt(
+                  totalRewardsBalance,
+                  MIN_ETH_AMOUNT_TO_DISPLAY_IN_WEI
+                )}
+                tooltipText={amountTooltipText(totalRewardsBalance)}
```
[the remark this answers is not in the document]
r-czajkowski: I think it might be usfule with others token (tBTC, KEEP) 🤔

solidity/dashboard/src/components/Rewards.jsx
Shadowfiend: I'm going to ponder this some more over the weekend, but `null` feels like a problematic value to have to pass explicitly. It doesn't explain itself at all, which makes it harder to understand what exactly is happening. It feels like if `null`/`undefined` is a valid value, it should be the default, even if that means we're passing `4` in many other places.

solidity/dashboard/src/components/Rewards.jsx
```diff
@@ -170,20 +167,39 @@ const RewardsComponent = ({ withdrawRewardAction }) => {
...
           ) : (
-            <>
-              <h1 className="balance">
-                {totalRewardsBalance}
-                <span className="h3 mr-1">&nbsp;ETH</span>
-              </h1>
-              <SpeechBubbleTooltip text="The total balance reflects the total Available and Acummulating rewards. Available rewards are ready to be withdrawn. Acummulating rewards become available after a signing group expires." />
-            </>
+            <div className="flex row mb-1 mt-1">
+              <TokenAmount
```
Shadowfiend, line 182: We appear to be displaying a pointer cursor for these elements, but I'm not sure if clicking is a valid action? If it is, I'm not sure what it does, which is also not ideal :grin: If it isn't, then a pointer cursor doesn't seem appropriate---though I'm open to reasoning there.

solidity/dashboard/src/components/TokenAmount.jsx
```diff
@@ -12,14 +12,21 @@ const TokenAmount = ({
   displayWithMetricSuffix,
   currencySymbol,
   displayAmountFunction,
+  withTooltip,
+  tooltipText,
 }) => {
   const { value, suffix } = displayWithMetricSuffix
     ? getNumberWithMetricSuffix(displayAmountFunction(amount, false))
```
Shadowfiend, line 19: The more I think about it the more it makes sense to simplify this and drop `displayAmountFunction`, `withTooltip`, and `tooltipText` entirely, and do all of the formatting internally.

To do this, we might rename `displayEthAmount` to `display18DecimalAmount` or similar. `TokenAmount` would instead allow passing in e.g. `smallestPrecisionDecimal` and `smallestPrecisionUnit`. For ETH, `currencySymbol` might be `ETH`, `14`, and `gwei`. TBTC might use this as `8` and `tSats`, for example, to display `<0.0001 TBTC` and a tooltip of `1000.01256 tSats`.

Thoughts? I think it can happen in a followup, of course.

solidity/dashboard/src/utils/ethereum.utils.js
```diff
@@ -43,3 +46,32 @@ export const getEventsFromTransaction = async (contractToEventName, txHash) => {
...
+  .pow(toBN(14))
+  .toString() // 0.0001 ETH
+
+export function displayEthAmount(
+  amountInWei,
+  unit = "ether",
+  decimalsPlaces = 4
+) {
+  if (!amountInWei) {
+    return 0
```
Shadowfiend, line 64: Let's return `"0"` as a string so this function consistently returns a string, and check for `amountInWei == "0"` or perhaps `parseInt(amountInWei) == 0`. Right now a 0 reward shows `<0.0001`.

Shadowfiend (approved):
Left some thoughts. I think we're currently substantially better than we were before, so I'm going to go ahead and approve + merge here.

solidity/dashboard/src/utils/ethereum.utils.js
r-czajkowski: pls see 51a8003

solidity/dashboard/src/components/Rewards.jsx
r-czajkowski: right, maybe `cursor: help`? But it's not a help heh 😆 so maybe we should set the `default`?

solidity/dashboard/src/components/TokenAmount.jsx
r-czajkowski: I think it makes sense. I'll open an issue for that, since it won't be a one-line fix.

r-czajkowski: #2095

solidity/dashboard/src/components/Rewards.jsx
```diff
@@ -170,20 +167,39 @@ const RewardsComponent = ({ withdrawRewardAction }) => {
...
+                  className: "eth-icon primary",
+                }}
+                displayWithMetricSuffix={false}
+                amount={totalRewardsBalance}
+                amountClassName="h1 text-primary"
+                displayAmountFunction={displayEthAmount}
+                withTooltip={lt(
+                  totalRewardsBalance,
+                  MIN_ETH_AMOUNT_TO_DISPLAY_IN_WEI
+                )}
```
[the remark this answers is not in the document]
r-czajkowski: Can we address it in #2095 ? I think its a good place to handle it.

solidity/dashboard/src/components/Rewards.jsx
Shadowfiend: I think `default` is just fine, yeah. I've seen `help` before, but it's always confusing tbh.

solidity/dashboard/src/components/Rewards.jsx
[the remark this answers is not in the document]
Shadowfiend: Yep, that sounds perfect.

Shadowfiend (approved):
Forgot to actually approve, whoopsie-daisy.

PR #2051, opened 2020-09-15
`useEtherscanUrl` hook fix

r-czajkowski:
This hook relied on the existence of the web3 provider, so it always returned a ropsten link during hardware wallet connection pop-up. Here we updated this custom hook and currently relies on the `@keep-network` package, since for a given build only support a single network ID.

### DIFF
--- a/solidity/dashboard/src/hooks/useEtherscanUrl.js
+++ b/solidity/dashboard/src/hooks/useEtherscanUrl.js
@@ -1,10 +1,8 @@
-import { useContext } from "react"
-import { Web3Context } from "../components/WithWeb3Context"
+import { getChainId } from "../connectors/utils"
 
+const chainID = getChainId()
 export const useEtherscanUrl = () => {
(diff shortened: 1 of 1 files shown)


solidity/dashboard/src/hooks/useEtherscanUrl.js
```diff
@@ -1,10 +1,8 @@
-import { useContext } from "react"
-import { Web3Context } from "../components/WithWeb3Context"
+import { getFirstNetworkIdFromArtifact } from "../contracts"
 
+const networkID = getFirstNetworkIdFromArtifact()
```
Shadowfiend, line 3: Would it be better ot use `getChainId` from `utils`? Feels like we're using it in similar scenarios elsewhere, and it lets us wrap the `getFirstNetworkIdFromArtifact` behavior centrally in case we decide to change it later.

Shadowfiend (approved):
Let's rock and roll :sunglasses:

PR #2054, opened 2020-09-16
Adding back mainnet bootstraps from Bison Trails and Figment Network

pdyraga:
Bison Trails and Figment Network bootstraps are active now and ready for work!


mattharrop:
Figment nodes are up, looks good to me.

Shadowfiend (approved):
Let's do it, we've got confirmation across the board.

PR #2058, opened 2020-09-17
Fix releasing tokens from a managed grant

r-czajkowski:
Set `args` param to an empty array when calling `withdraw` function from a `ManagedGrant` contract  via [`sendTransaction`](https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/sagas/web3.js#L111) helper.

### DIFF
--- a/solidity/dashboard/src/sagas/web3.js
+++ b/solidity/dashboard/src/sagas/web3.js
@@ -9,7 +9,7 @@ import {
 } from "../actions/messages"
 import { messageType } from "../components/Message"
 
-function createTransactionEventChannel(contract, method, args, options) {
+function createTransactionEventChannel(contract, method, args = [], options) {
   const infoMessage = Message.create({
     title: "Waiting for the transaction confirmation...",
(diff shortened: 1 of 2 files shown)


solidity/dashboard/src/sagas/web3.js
```diff
@@ -9,7 +9,7 @@ import {
 } from "../actions/messages"
 import { messageType } from "../components/Message"
 
-function createTransactionEventChannel(contract, method, args, options) {
+function createTransactionEventChannel(contract, method, args = [], options) {
```
Shadowfiend, line 12: In general I tend towards having a default value for all arguments after the first one with default value, just so it's clear to someone looking at it what's going on. Would it make sense to give `options` a default value of `{}`?

Shadowfiend: > Would it make sense to give options a default value of {}?

Let's make this change in a follow-up---I'd like to get the release tokens bug fix deployed.

PR #2060, opened 2020-09-17
Information on the gas cost of operating the client

pdyraga:
Added information about gas costs of operations the beacon client is performing. It is important that the operators understand the cost of running the client and can estimate the required ETH based on their stake, current gas price, and the expected traffic in the network.

### DIFF
--- a/docs/run-random-beacon.adoc
+++ b/docs/run-random-beacon.adoc
@@ -55,19 +55,23 @@ to submit more transactions as a result.
 
 Below is the average gas cost of the most important transactions the client is submitting:
 
-[%header,cols=2*]
+[%header,cols=3*]
 |===
 |TX
 |Gas Cost
+|Reimbursed
 
 |Submit group selection ticket
 |140 000
+|No
 
 |Submit DKG result
 |1 740 000
+|Yes
 
 |Submit relay entry
 |280 000
+|Yes
 |===
 
 For example, if the operator has 10 x minimum stake, it can submit 10 tickets. If the operator has
--- a/docs/run-random-beacon.adoc
+++ b/docs/run-random-beacon.adoc
@@ -43,6 +43,35 @@ Keep Random Beacon client.
 |2 vCPU / 4 GiB RAM / 1 GiB Persistent Storage
 |===
 
+== Gas Costs
+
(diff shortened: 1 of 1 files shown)


docs/run-random-beacon.adoc
```diff
@@ -43,6 +43,35 @@ Keep Random Beacon client.
 |2 vCPU / 4 GiB RAM / 1 GiB Persistent Storage
 |===
 
+== Gas Costs
+
+Random Beacon smart contracts reimburse the operator for successfully submitting relay entry and DKG result but they do not reimburse for submitting a group selection ticket. Reimbursements are sent to the beneficiary account and can be claimed along with rewards once the group expires. It is expected that the operators have enough ETH on the accounts used by clients to submit the required transactions and that the operator account balance is monitored and refilled as needed. Bear in mind that the higher stake is, the operator is selected more frequently and is expected to submit more transactions as a result.
```
Shadowfiend, line 48: Let's line wrap here and in the next paragraph.

docs/run-random-beacon.adoc
```diff
@@ -43,6 +43,35 @@ Keep Random Beacon client.
...
+|Gas Cost
+
+|Submit group selection ticket
+|140 000
+
+|Submit DKG result
+|1 740 000
+
+|Submit relay entry
+|280 000
```
Shadowfiend, line 64: Can we add a column on reimbursed yes/no?

docs/run-random-beacon.adoc
Shadowfiend: Maybe we can also clarify (could be its own paragraph after the table, not sure) that reimbursements happen after successful group formation?

pdyraga: > Maybe we can also clarify (could be its own paragraph after the table, not sure) that reimbursements happen after successful group formation?

We do have:

> Random Beacon smart contracts reimburse the operator for successfully submitting relay entry and DKG result

🤔 Isn't it clear from this sentence?

pdyraga:
It's ready for another chance @Shadowfiend

Shadowfiend (approved):
Yah let's do it.

PR #2062, opened 2020-09-18
Set the default value for a  `options` param in the `createTransactionEventChannel`

r-czajkowski:
Ref: https://github.com/keep-network/keep-core/pull/2058#discussion_r490433685 
Set the default value to `{}` for the `options` param in the `createTransactionEventChannel` function.


Shadowfiend (approved):
🚀

Issue #2071, opened 2020-09-21
Loading wrapper for Token Overview and Delegated Tokens page 


r-czajkowski:
Let's add loading skeletons for `Token Overview` and `Delegated Tokens` page that match the layout.

PR #2073, opened 2020-09-21
Loading wrappers

r-czajkowski:
Closes: #2071 

This PR adds the loading skeletons for Overview page and for top pane on the Delegated Tokens page to improve UX.

obraz

obraz


Shadowfiend:
Couple more notes here:
- @liz-shinn I think we should consider making the delegation form something that requires an explicit action to bring up. Right now it dominates the landing experience on the Delegate Tokens page. Its presence also 
- If there are no grants to display, we should really remove the “Grants” section altogether and only display the “Owned” section. Thoughts?
- Since that delegation form isn't useful until the page is loaded, if it's visible at page render (rather than hidden behind a click), it should also have some sort of loading indicator.

Shadowfiend (approved):
Landing this as-is, looks good.

PR #2074, opened 2020-09-21
Add peers multiaddresses to the diagnostics endpoint response

lukasz-zimnoch:
Here we add a new field with known multiaddresses of the connected peers to the diagnostics endpoint response. That enables the possibility of diagnosing issues related to the public availability of network members.

### DIFF
--- a/pkg/diagnostics/diagnostics.go
+++ b/pkg/diagnostics/diagnostics.go
@@ -43,9 +43,16 @@ func RegisterConnectedPeersSource(
 				continue
 			}
 
+			multiaddresses, err := connectionManager.GetPeerMultiaddresses(peer)
+			if err != nil {
+				logger.Error("error on getting peer multiaddresses: [%v]", err)
+				continue
+			}
+
 			peersList[i] = map[string]interface{}{
 				"network_id":       peer,
 				"ethereum_address": key.NetworkPubKeyToEthAddress(peerPublicKey),
+				"multiaddresses":   multiaddresses,
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/diagnostics/diagnostics.go
```diff
@@ -43,9 +43,16 @@ func RegisterConnectedPeersSource(
...
+			multiaddresses, err := connectionManager.GetPeerMultiaddresses(peer)
+			if err != nil {
+				logger.Error("error on getting peer multiaddresses: [%v]", err)
+				continue
+			}
+
 			peersList[i] = map[string]interface{}{
 				"network_id":       peer,
 				"ethereum_address": key.NetworkPubKeyToEthAddress(peerPublicKey),
+				"multiaddresses":   multiaddresses,
```
Shadowfiend, line 55: I wonder if this should be called `addresses` rather than `multiaddresses`. My reasoning here is that these interfaces are meant to be implementation-agnostic, and multiaddrs are a libp2p construct.

Another option is to make parts of the diagnostics endpoint implementation-specific via casting instead of pushing all diagnostics info into the implementation-agnostic `net.` interfaces.

Shadowfiend: (Also fine with keeping it as-is for now---the network layer is currently unlikely to be pulled away from libp2p in the near term.)

lukasz-zimnoch: Thanks for your point! According to [multiaddresses docs](https://multiformats.io/multiaddr/), they are not really a libp2p specific construct but seems they are a little bit above. I think this is a good justification to keeping it as is.

Shadowfiend: Yeah, sorry, they're not _meant_ to be a libp2p-specific construct, but de facto they are as no other p2p network I know of uses them at the moment.

Anyway, I'm comfortable leaving it as `multiaddresses`. A future move away from multiaddrs for any reason would likely start by transitioning `multiaddresses` to `addresses` and then leave `multiaddresses` empty before removing it.

lionakhnazarov:
Closing as stale. The diagnostics endpoint has changed since 2020; please re-cut against current main if peer multiaddress output is still wanted.

PR #2097, opened 2020-09-29
Remove the unused Ethereum address in the keep-client configuration

biluohc:
This Ethereum address is not used in the code. If the user fills in an address that does not match the keyfile, the error will be difficult to understand and mistakenly think that their address staking is incorrect.

In addition, there is no such field in keep-common and keep-ecdsa. Compared with adding this field and judgment logic in the code, it should be the simplest to delete directly.


Shadowfiend:
Thanks for this PR @biluohc ! We've been a little underwater pushing forward some other features, but this looks like a great addition 👌 We'll aim to merge it soon.

Issue #2082, opened 2020-09-21
Rename Glossary page to Resources


liz-shinn:
This section used to be only glossary / terms but now includes documentation and diagrams. Would be more aptly named "Resources"

PR #2098, opened 2020-09-29
Rename `glossary` to `resources`

r-czajkowski:
~~Depends on: #2096~~
Closes: #2082


r-czajkowski:
Not sure it's ok. Since we removed a footer, making the only last link as active is a bit tricky? 🤔  Currently out of idea how to handle it.

obraz

Shadowfiend (commented):
Left a couple of notes; I think the only blocker is figuring out that delegation diagram link, and again I'd suggest we just add some bottom margin there.

Another option would be to flip the terminology and diagram entries on the page, so that the tallest page is on the bottom (/cc @liz-shinn ).

liz-shinn:
> Another option would be to flip the terminology and diagram entries on the page, so that the tallest page is on the bottom (/cc @liz-shinn ).

This seems reasonable to me! Agreed that we could use some more margin there on the bottom of the page.

r-czajkowski:
> I'd suggest we just add some bottom margin there.

we are going to rebuild the app structure in #2101 so I'll add `margin-bottom` to the `main` tag in #2101.

solidity/dashboard/src/components/resources/Navigation.jsx
```diff
@@ -7,43 +7,33 @@ const Navigation = () => {
...
         activeClass="active"
-        to={to.hash.slice(1)}
+        to={to}
         spy={true}
         smooth={true}
         offset={offset}
         duration={500}
-        onSetActive={() => history.replace(to)}
+        onSetActive={() =>
+          history.replace({ pathname: "/resources", hash: to })
```
Shadowfiend, line 33: I guess the one downside is now this file has to know that resources are dropped at `/resources`... I think I'll have some more thoughts around this bundled into the suggestion regarding [sublinks in 2101](https://github.com/keep-network/keep-core/pull/2101#discussion_r500639319).

Shadowfiend (approved):
Outcome looks good here, any other tweaks we want feel like they can be taken care of in the stack of followup work :grin:

solidity/dashboard/src/components/resources/Navigation.jsx
Shadowfiend: Actually, it looks like we might should be able to use `let { path } = useRouteMatch();` to look up the parent route, maybe?

Issue #1999, opened 2020-08-19
Updates to the secondary navigation for token dashboard


liz-shinn:
Updates the architecture of the dApp to include a secondary navigation in the body of the dapp instead of nested within the main navigation, which will help to more clearly navigate the dapp.

Additionally:
+ Removes the token grant and owned tokens selector and makes them individual pages
+ Moves the "Overview" onto its own main navigation page instead of within Tokens
+ Page title moves closer to the top of the page and matches the main navigation selection

liz-shinn:
@pdyraga @r-czajkowski Added this to backlog for the dashboard.

liz-shinn:
Example of secondary nav [from Figma file](https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=5764%3A54122)

<img width="699" alt="Screen Shot 2020-08-19 at 3 50 57 PM" src="https://user-images.githubusercontent.com/57226633/90643385-d9974380-e233-11ea-9054-7a57e4583a8c.png">

PR #2101, opened 2020-09-30
Header and sec menu

r-czajkowski:
~~Depends on: #2096~~
Closes: #1999

This PR updates the header styles and adds the secondary nav into header.

obraz


solidity/dashboard/src/components/Header.jsx
```diff
@@ -1,30 +1,70 @@
...
-import { NetworkStatus } from "./NetworkStatus"
+import React, { useState, useCallback } from "react"
+import { NavLink } from "react-router-dom"
+import { isEmptyArray } from "../utils/array.utils"
+import { useContext } from "react"
 
-const Header = (props) => {
-  const { yourAddress } = useContext(Web3Context)
+const Header = () => {
+  const { title, subLinks } = useHeaderContext()
```
Shadowfiend, line 7: So, I _think_ if the page structure looks more like what I described above, we can have `PageWrapper` be in charge of rendering the header, at which point the title and sublinks can propagate hierarchically instead of out-of-band via context. I'm not 100% sure of this, but from what I see it should be true.

solidity/dashboard/src/components/Icons.jsx
```diff
@@ -518,4 +434,5 @@ export {
   Question,
   Home,
   FeesVector,
+  KeepDashboardLogo,
```
Shadowfiend, line 437: Random question for my own edification: what do we gain by including these SVGs directly vs using an `img` or `picture` element to reference them? Just reducing the number of requests we're making, or is there another gain?

solidity/dashboard/src/components/glossary/Navigation.jsx
```diff
@@ -31,6 +31,7 @@ const GlossaryLink = ({ label, to }) => {
   return (
     <li>
       <Link
+        containerId="main-content-wrapper"
```
Shadowfiend, line 34: Before I forget: we may want to check with @liz-shinn, but it feels like the side navigation used for the glossary (now resources) could just become the same top tab navigation that we're using for other subnavigation?

solidity/dashboard/src/css/header.less
```diff
@@ -0,0 +1,60 @@
+.header {
+    grid-area: header;
+    width: 100%;
+    z-index: 10;
+    display: flex;
```
Shadowfiend, line 5: Could use a comment here to indicate why we're using `display: flex`.

solidity/dashboard/src/css/header.less
```diff
@@ -0,0 +1,60 @@
...
+            bottom: 0;
+            background-color: @color-grey-60;
+            transition: all 0.3s ease;
+          }
+        }
+      }
+  
+       .sub-nav__item--active, .sub-nav__item:hover {
+        color: @color-black;
+        .cursor-pointer();
```
Shadowfiend, line 52: Could we just say `cursor: pointer` here?

solidity/dashboard/src/pages/ApplicationsPageContainer.jsx
```diff
@@ -1,25 +1,30 @@
...
+          exact
+          path="/applications/tbtc"
+          component={TBTCApplicationPage}
+        />
+        <Route
+          exact
+          path="/applications/random-beacon"
+          component={KeepRandomBeaconApplicationPage}
+        />
+        <Redirect to="/applications/random-beacon" />
```
Shadowfiend, line 25: This is the “if nothing else matches” fallthrough?

solidity/dashboard/src/pages/ApplicationsPageContainer.jsx
```diff
@@ -1,25 +1,30 @@
...
+        <Route
+          exact
+          path="/applications/tbtc"
+          component={TBTCApplicationPage}
+        />
+        <Route
+          exact
+          path="/applications/random-beacon"
+          component={KeepRandomBeaconApplicationPage}
+        />
```
Shadowfiend, line 24: Might be useful to specify these in the same order as the nav lists them.

I also immediately wonder if we could use the `subLinks` data to build this `Route` table, instead of having to specify the info in it twice 🤔 

Perhaps `PageWrapper` could actually take a `{title, path, component}` array and handle the `Route` piece directly? We could always use the first item as the default for the fallthrough below. Then we could probably eliminate a lot of these container files altogether in favor of a single declarative JS object map that describes the page hierarchy and component mapping..? Sounds a little crazy, but it also feels doable…

solidity/dashboard/src/pages/RewardsPage.jsx
```diff
@@ -1,13 +1,8 @@
...
 import { Rewards } from "../components/Rewards"
-import PageWrapper from "../components/PageWrapper"
 
 const RewardsPage = () => {
-  return (
-    <PageWrapper title="Random Beacon Rewards">
-      <Rewards />
-    </PageWrapper>
-  )
+  return <Rewards />
```
Shadowfiend, line 5: Why have a `RewardsPage` intermediary in the first place if all it does is pass through the component, out of curiosity?

Shadowfiend (commented):
I left… Thoughts. Possibly a lot of them :grimacing:

solidity/dashboard/src/components/glossary/Navigation.jsx
liz-shinn: Hadn't considered this 🤔 I'm not sure, tbh... 

I like the structure of the side navigation in this resources (glossary) as it stands for a couple reasons:
- it mirrors a common pattern for docs, so it looks at a glance like this is a doc section
- it allows for all of this content to be in one place (the diagram doesn't really warrant it's own page imo)

But I *do* like the idea of rearranging it to match the other pages for a few other reasons:
- it's consistent with other pages treatment of subnavigation, so teaches user to look in the same place
- it allows for more horizontal screen real estate, especially for diagrams (I can see there being a bunch of info viz we could include here)
- it's more scalable and better suited to adding more items, in particular links out to docs

So actually, TL;DR I've talked myself into doing what @Shadowfiend suggests here, and agreed that it makes sense to make the side nav for glossary the same top tab navigation that we're using for other subnavigation. 👍 💯

solidity/dashboard/src/App.js
```diff
@@ -16,19 +16,21 @@ const App = () => (
       <Web3ContextProvider>
         <ModalContextProvider>
           <ContractsDataContextProvider>
-            <SideMenuProvider>
-              <Router>
-                <main>
-                  <Header />
-                  <aside>
+            <HeaderContextProvider>
+              <SideMenuProvider>
```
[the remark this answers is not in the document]
r-czajkowski: yeah 100% right. let's drop a side menu context.

I decided to remove `MenuButton` component from a header because we don't have any mobile views on Figma yet and I think it's more reasonable to deal with mobile views in a separate PR.

Regarding the contexts, its on my plate too, to cleanup this mess.

solidity/dashboard/src/pages/RewardsPage.jsx
Shadowfiend: Writing this down while it's on my mind: keeping the [previous note](https://github.com/keep-network/keep-core/pull/2101#discussion_r500639319) around link data and the routing table in mind, I'm wondering if we reconsider some of our component vs page distinctions a little bit. In particular, my thinking here would be:
- Pages are defined as React components that bundle page structure + routing information.
- Anything that requires routing details (e.g., `ResourcesLink`) should either be used directly at the `Page` level, or it should be passed to the component that needs it (as a prop).
- A Page component is a function, but it has a special property, `route`, that indicates routing information (title, path, and `pages`, which specifies child `Page` components).
- The routing table is a list of top-level `Page` components, which in turn may define recursive children.

Quick sketch of this:
```
const pages = [OverviewPage, TokensPage, …, ApplicationsPage, …]

...

const ApplicationsPage = (...) {
  …
}
ApplicationsPage.route = {
  title: 'Applications',
  path: '/applications/',
  pages: [KeepRandomBeaconApplicationPage, TBTCApplicationPage]
}

…

const KeepRandomBeaconApplicationPage = (…) {
  …
}
KeepRandomBeaconApplicationPage.route = {
  title: 'Keep Random Beacon,
  path: '/random-beacon
}
```

Then we could use this to build up the routing table. There's a similar concept defined on the react-router site @ https://reactrouter.com/web/example/route-config . Thoughts?

solidity/dashboard/src/App.js
```diff
@@ -16,19 +16,21 @@ const App = () => (
...
+                    <Header />
                     <SideMenu />
-                  </aside>
-                  <div className="content">
-                    <Routing />
-                  </div>
-                </main>
-              </Router>
-            </SideMenuProvider>
+                    <div className="content-wrapper" id="main-content-wrapper">
```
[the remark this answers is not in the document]
r-czajkowski: In my opinion this
```
grid-template-areas:
    "header main" auto
    "footer main" max-content / 200px auto;
```
does not reflect the structure of the app in a right way.

I think we ca simplify and just `grid-template-columns: @side-menu-width 1fr`, and we get 2 columns layout.

What do you think about something like this?
```
<div class="wrapper"> // grid-template-columns: @side-menu-width 1fr
 <div class="side-menu">
  <header/>
  <nav/>
  <footer/>
 </div>
 <div class="content">
  <header/> <- as header component with page title and sublinks
  <main>
   page content here
  </main>
 </div>
</div>
```

solidity/dashboard/src/components/Header.jsx
r-czajkowski: I thought about it but opted for a context. The main idea with `HeaderContext` is that we only have the `Header` component in one place and for me, this solution fits better to the struct of app:
```
nav header
nav content
```
But now I think rendering a `Header` in a `PageWrapper` makes a lot of sense and it is less "complicated".

solidity/dashboard/src/components/Icons.jsx
r-czajkowski: We can style svg elements via css. In my opinion, is easier than changing svg styles inside img tag

solidity/dashboard/src/components/glossary/Navigation.jsx
r-czajkowski: so I'll remove the right side menu and use sub nav on the top 👍

solidity/dashboard/src/pages/ApplicationsPageContainer.jsx
r-czajkowski: yeah

> You can use the last `Route` in a `Switch` as a kind of "fallback" route, to catch 404 errors. A `Switch` renders the first child `Route` that matches

^ from the `react-router` docs

solidity/dashboard/src/pages/ApplicationsPageContainer.jsx
r-czajkowski: yeah sounds good.

> Sounds a little crazy, but it also feels doable…

 I'll do my best 😄

solidity/dashboard/src/css/header.less
r-czajkowski: actually we don't need flexbox here. I'll remove 👍

solidity/dashboard/src/components/glossary/Navigation.jsx
liz-shinn: Updating the Figma file to reflect this change to the Resources page 💯

solidity/dashboard/src/pages/ResourcesPage.jsx
```diff
@@ -3,21 +3,33 @@ import PageWrapper from "../components/PageWrapper"
...
+DelegationDiagram.route = {
+  title: "Diagrams",
+  path: "/resources/diagram",
+  exact: true,
+}
+
+TerminologyDataTable.route = {
+  title: "Quick Terminology",
+  path: "/resources/quick-terminology",
+  exact: true,
```
Shadowfiend, line 26: I wonder if `exact` can be inferred---by which I mean, if `pages` exists, then `exact: false`, if `pages` does not exist, then `exact: true`, without having to explicitly define it.

Obviously non-blocking.

solidity/dashboard/src/App.js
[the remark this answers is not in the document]
Shadowfiend: I think this achieves the visual aim but continues to leave the semantic hierarchy muddled. Can you explain how the structure of the app isn't reflected in my suggestion? It feels like your suggested revision tries to align the semantic hierarchy to the visual structure in a way that isn't necessary.

I find it helpful to think about two aspects of this when I'm trying to figure out the semantic hierarchy. The first is, “how would a typical page present this?” The second and most important one is “how should this flow for someone who cannot see the visual arrangement that we've chosen”. In that context, the `footer` is always the least important and last thing you want presented to someone browsing the structure (the most typical example being a user with a screen reader), because it's the one they're least likely to care about. The header and nav tend to be near the top of the ordering because they're the ones that tell the user the most information about what they're looking at and how the site hierarchy can be navigated.

This is where my proposed page structure comes from: it puts the site title and nav and page title and nav front and center, and finally puts the main content before the footer.

This also feeds my own driving principle for semantics, which is that a page that is well-structured *semantically* will also be easier to manage *programmatically*. In this case, I think a structure like this also allows the data flow to be hierarchical in more places, as opposed to having to use contexts to manage data flow non-hierarchically. Of course, this assumes hierarchical data flow is better—which I also happen to believe, but am open to debate on :)

Overall I still think my original suggestion has advantages over the current state, but I'm also not willing to block what we have here. It's solid work, and it looks and works well. Let's continue the conversation moving forward, but separately.

solidity/dashboard/src/pages/TokensPageContainer.jsx
```diff
@@ -479,10 +472,16 @@ const TokensPageContainerWithRedux = connect(
...
+    <TokensPageContainerWithRedux {...props} />
   </TokensPageContextProvider>
-)
+))
+
+TokensPageContainerWithContext.route = {
+  title: "Tokens",
+  path: "/tokens",
+  pages: [TokenOverviewPage, TokensPage, TokenGrantsPage],
+}
```
Shadowfiend, line 485: How does this approach feel? Is it better/worse/neutral compared to what we had before?

Shadowfiend (approved):
Solid, everything seems to work, let's roll it in.

I'm assuming the work to shift the glossary to the regular top nav is happening in another PR (via https://github.com/keep-network/keep-core/pull/2101#discussion_r500633059), and I also want to flag that the `main` element still includes the full page content instead of targeting the page's primary content (see my note @ https://github.com/keep-network/keep-core/pull/2101#discussion_r500548111).

Let's merge this 😎

solidity/dashboard/src/pages/TokensPageContainer.jsx
r-czajkowski: I like that abstraction and that we have a special type of component- `Page` component, that requires extra attributes and I think perfectly fits to our design.

Just wonder how could be problematic if we have a "special" type of page that will require something that we didn't predict in this solution. I mean I would like to avoid a situation where we need to pass a ton of props to the PageWrapper or other Routing components only to render this "special" page.

r-czajkowski:
> I'm assuming the work to shift the glossary to the regular top nav is happening in another PR

I've addressed this issue with https://github.com/keep-network/keep-core/pull/2101/commits/8e4c6c511d411e41f3596b4d4636045207624288

> and I also want to flag that the main element still includes the full page content instead of targeting the page's primary content

Hmm... the main tag contains only this part that specific page renders. `header`, `footer` and `nav` are outside the `main` tag. An then in `main` tag we have `section`'s.

solidity/dashboard/src/pages/ResourcesPage.jsx
r-czajkowski: Hmm, I see your point but on the other hand, explicitly defined attributes seem better(to me) than this kind of magic.

solidity/dashboard/src/App.js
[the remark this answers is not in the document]
r-czajkowski: For me the confusing part was:
```
grid-template-areas:
    "header main" auto
    "footer main" max-content / 200px auto;
```

since `header` it's not a `header` but `side bar`<- in visual structure. In my case, I thinking of the header as a container on the top, not on the left or right side. I think the grid templates should reflect the visual structure since its better to understand what's going on.

The more I wonder the more i understand your point. As I understand we want in React component something like this:

```
header that contains
 app-logo
 nav
routing- that renders the page content with sub heder and nav
footer
```
But we want to deal with element positioning in the css side.

It all depends on the approach that we use. I tried to align React components to what we see on the screen, but agree with your principle for semantics.

I'll open a new PR in a while to achive your semathic structure.

PR #2102, opened 2020-09-30
Make rewards upgradeable

eth-r:

A single rewards contract deployment is designed for only one keep factory. Because keeps may be created simultaneously in different factories, it would be quite complicated to accommodate different versions on one contract. Instead, the rewards contract can be given the address of an upgraded version. If the upgraded version is specified, rewards tokens that would normally be returned to the unallocated pool will instead be sent to the new rewards contract. This will gracefully bleed out the old rewards contract's reward tokens as the creation of new keeps (or beacon groups) dwindles over time.

### DIFF
--- a/solidity/contracts/Rewards.sol
+++ b/solidity/contracts/Rewards.sol
@@ -14,10 +14,13 @@
 
 pragma solidity ^0.5.17;
 
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
 import "openzeppelin-solidity/contracts/token/ERC20/SafeERC20.sol";
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "openzeppelin-solidity/contracts/math/Math.sol";
+
+import "./KeepToken.sol";
 
 /// @title KEEP Signer Subsidy Rewards
 /// @notice A contract for distributing KEEP token rewards to keeps.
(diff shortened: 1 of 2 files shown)


solidity/contracts/Rewards.sol
```diff
@@ -469,9 +478,36 @@ contract Rewards {
...
+        uint256 baseAllocation = _baseAllocation(interval);
+        uint256 adjustedAllocation = _adjustedAllocation(interval);
+        uint256 allocationDifference = baseAllocation.sub(adjustedAllocation);
+        intervalAllocations.push(adjustedAllocation);
+        // If a new rewards contract has been defined and the quota isn't met,
+        // send the excess to the new contract.
+        // Otherwise keep the excess unallocated.
+        if (newRewardsSpecified() && (allocationDifference > 0)) {
+            unallocatedRewards = unallocatedRewards.sub(baseAllocation);
+            sendToNewRewards(allocationDifference);
```
pdyraga, line 490: The assumption here is: Stakers will migrate to the new operator contract and we will not meet the minimum groups quota here. Over time, we can move unallocated tokens to the new rewards contract.

This is very elegant but I am not sure if we should handcuff ourselves. There are two issues I see:
- The minimum group quota for random beacon is set to 2 (group expiration time is two weeks, interval is 30 days). Anyone with 64 * min_stake, can block the migration.
- It will take a lot of time until we reallocate tokens, even if no groups will be created. And the rewards curve on the new contract will not have the expected shape.

Instead, I propose to set one interval governance delay and once the last interval is allocated, move all the unallocated tokens out from this contract to the new one.

WDYT?

pdyraga: An alternative proposition: https://github.com/keep-network/keep-core/pull/2108

eth-r:
Closing in favor of #2108

PR #2103, opened 2020-09-30
Flattening the rewards curve

pdyraga:

Instead of allocating yet unallocated rewards in the last interval with 100% interval weight, we are going to use the last defined interval's weight which should flatten the rewards curve.

Instead of:
image

We would have:
image


dimpar (approved):
Looks good to me!

PR #2104, opened 2020-09-30
add datetime and peers's network addr to diagnostics

biluohc:
Add peers' network addresses to the diagnostics API.

### DIFF
--- a/pkg/diagnostics/diagnostics.go
+++ b/pkg/diagnostics/diagnostics.go
@@ -2,11 +2,13 @@ package diagnostics
 
 import (
 	"encoding/json"
+	"time"
 
 	"github.com/ipfs/go-log"
 	"github.com/keep-network/keep-common/pkg/diagnostics"
 	"github.com/keep-network/keep-core/pkg/net"
 	"github.com/keep-network/keep-core/pkg/net/key"
+	manet "github.com/multiformats/go-multiaddr/net"
 )
 
 var logger = log.Logger("keep-diagnostics")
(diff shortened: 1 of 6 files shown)


pdyraga:
Thanks for your contribution! 🙌  We need to take a closer look, it's possible we have a duplicate with https://github.com/keep-network/keep-core/pull/2074

biluohc:
Ok, i will let ci pass tomorrow

pkg/diagnostics/diagnostics.go
```diff
@@ -78,6 +89,8 @@ func RegisterClientInfoSource(
 		clientInfo := map[string]interface{}{
 			"network_id":       clientID,
 			"ethereum_address": key.NetworkPubKeyToEthAddress(clientPublicKey),
+			"network_addrs":    connectionManager.NetAddrStrings(),
+			"datetime":         time.Now().Format("2006-01-02 15:04:05"),
```
Shadowfiend, line 93: Circling back here after a long time, I realize---what's the reasoning for including a timestamp in the JSON here? I believe Go's default HTTP response includes a `Date` header that should make this redundant.

biluohc: Thank you for your reminder, I mistakenly thought that Diagnostics also has a tick like Metrics.

But the timestamp in JSON may be more convenient than the HTTP header, and there is only a little overhead. I can also delete it if you want to delete it.

Is there any other problem with the code, let me update code together

PR #2105, opened 2020-09-30
Fund properly before allocating rewards

pdyraga:
Depends on #2103

We had a problem with `firstIntervalStart` property - it had to be set in a way that `firstIntervalStart + termLength` would be in the future. Otherwise, someone could call `allocateRewards(0)` before we fund the contract properly and the allocation for the first interval would not be the same as we'd want it to be.

Here we are making `Rewards` contract `Ownable` for only one purpose - the owner has to mark the contract as properly funded calling `markAsFunded` function. Only from then, rewards can be allocated.

Setting the contract as funded does not stop it from being funded even more in the future.

The expected order of operations is as follows:
- contract is deployed,
- contract is funded with the expected amount of KEEP,
- contract is marked as funded and ready to be used by stakers.

### DIFF
--- a/solidity/contracts/Rewards.sol
+++ b/solidity/contracts/Rewards.sol
@@ -14,6 +14,7 @@
 
 pragma solidity ^0.5.17;
 
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
 import "openzeppelin-solidity/contracts/token/ERC20/SafeERC20.sol";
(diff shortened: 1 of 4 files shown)


solidity/contracts/Rewards.sol
```diff
@@ -607,4 +616,9 @@ contract Rewards {
             "Keep not recognized by factory");
         _;
     }
+
+    modifier mustBeFunded() {
+        require(funded, "Contract not yet funded");
```
dimpar, line 621: "Contract has not been funded yet" ?

dimpar (approved):

Issue #2057, opened 2020-09-16
Change type style of address so it's not all caps


liz-shinn:
Addresses are case-sensitive; need to update the all caps type style to instead show both upper and lower case.

On the top right of the page it shows the first and last part of the address, but all letters are capital. So if an address is 0x559960cce06978064aa6AeBa5b2D8B3150d11aE6 it shows "1AE6" on the end.

Issue #2026, opened 2020-08-31
The wallet wall: address wallet integration


liz-shinn:
When to disable parts of the dapp, and when to show. Working with wallet integration of wallet into the dashboard.

- Remove the page "wall" that has four wallet choices and keeps you from seeing the rest of the dapp. 

- Show the Overview page instead as the first page you see when arriving at the dapp for the first time

- Show empty states, and where a wallet connection is needed have notation "Please connect wallet first"

- Wallet flow design: pictured below & in Figma

liz-shinn:
@r-czajkowski I've update how the wallet works to fix the issues with type size and also match what we're doing in the designs for tbtc dapp.

Looks like this, when disconnected and no wallet selected. Click to bring up the menu to select a wallet, then works like it does currently with pop ups. 
![wallet](https://user-images.githubusercontent.com/57226633/94871517-29af1b80-0418-11eb-8451-b6aab36098ab.png)

Then when wallet is connected and network is connected:
![Header](https://user-images.githubusercontent.com/57226633/94871535-32075680-0418-11eb-9da1-3a8e81ee9534.png)

liz-shinn:
In Figma here: https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=6474%3A33884

Issue #2002, opened 2020-08-19
Add Overview page separate from nested within Tokens page


liz-shinn:
Adds an overview page that's the first page a user comes to when they open the dashboard. It would have notifications (anything pertinent to the user/needs user action), overview of token balances. Potential to show activity here, or overview of rewards, maybe slashing notices.

Link to [Figma file here](https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=2172%3A322)

liz-shinn:
<img width="652" alt="Screen Shot 2020-08-19 at 6 44 18 PM" src="https://user-images.githubusercontent.com/57226633/90665186-fe97b080-e24b-11ea-90a4-25e2b6cc3777.png">

liz-shinn:
Noting that a future call needed here to discuss who the audience is for this page, what are their different needs, are we addressing all those different needs.

An ETH staker user journey, e.g.

cc/ @mhluongo @creckhow @Gluzman @r-czajkowski

PR #2107, opened 2020-10-01
Web3 status updates

r-czajkowski:
~Depends on: #2101~
Closes: #2057
Closes: #2026 

This PR updates the web3 status indicator in the KEEP token dashboard.

obraz

obraz

obraz

obraz

obraz

obraz


Shadowfiend:
🤔 is this based on any existing designs, or is it groundwork for implementing designs? This is what I'm seeing in Figma:

Screen Shot 2020-10-06 at 11 39 11

But I'm also seeing that the referenced issue only mentions capitalization, not any updates to styling or positioning of information.

liz-shinn:
Yes, the design for this wallet section had updated since these screenshots were taken, based on feedback from @r-czajkowski and @Gluzman. The styling you've screenshotted from Figma is the correct one.

liz-shinn:
Went into more detail on some different wallet states in this issue: https://github.com/keep-network/keep-core/issues/2026

liz-shinn:
>Not sure what we want to display when a user is not connected with a wallet.

@r-czajkowski Do you mean what we display in the body of the dapp?

The Overview page would be what is shown instead of the wallet wall page, even when there's no wallet connected. Instead of numerical data we can show some empty states. https://github.com/keep-network/keep-core/issues/2002

liz-shinn:
Noting that title of this page shouldn't be "Connect wallet" — should instead be "Overview" and show this:

Onboarding

liz-shinn:
I updated the Delegation page (for both Granted and Token subnavigation) to show an empty state for when a wallet isn't connected. 

Delegation Empty State


I don't think this is a page where it makes sense to show the full content with a bunch of N/A everywhere. I think it's better to explain what will be there (with a graphic) and give a clear action (Connect a Wallet).

@Shadowfiend @r-czajkowski

liz-shinn:
Sidenote that the case of the network status should be consistent. If we're using title case for "Network Disconnected," we should use title case for the other network statuses as well.

"Private"
"Please Connect a Wallet"
etc.

liz-shinn:
An empty state page for Token Grant page when a wallet isn't connected:

Grant Empty State

r-czajkowski:
> Updated the Delegation page to show an empty state for when a wallet isn't connected. Was thinking these could be the same for both Granted and Wallet subnavigation. Artboard in Figma file: https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=6710%3A10525
> 
> Delegation Empty State
> 
> I don't think this is a page where it makes sense to show the full content with a bunch of N/A (Not Available) for all the numbers and data. It also doesn't make sense to show 0s when a wallet isn't connected since that's a separate state (a zero balance =/= disconnected wallet).
> 
> Would recommend explaining what _will_ be there with a graphic and give a clear action (Connect a Wallet).
> 
> @Shadowfiend @r-czajkowski

Totally agree, this empty states are awesome! 🚀 I'll open an issue for that and add to the milestone!

solidity/dashboard/src/components/AddressShortcut.jsx
```diff
@@ -1,25 +1,26 @@
...
-      className={`address-shortcut tooltip address ${classNames}`}
-    >
-      <span className="tooltip-text address bottom">{copyStatus}</span>
-      {shortenAddress(address)}
-    </span>
+    <CopyToClipboard
+      toCopy={address}
+      render={({ reset, copyStatus, copyToClipboard }) => {
+        return (
+          <span
```
Shadowfiend, line 13: `button` if it's clickable, as a general rule.

solidity/dashboard/src/components/NetworkStatus.jsx
```diff
@@ -1,24 +1,27 @@
...
+import * as Icons from "./Icons"
 
 export const NetworkStatus = () => {
-  const { networkType, provider, error, yourAddress } = useContext(Web3Context)
+  const { networkType, provider, error, yourAddress } = useWeb3Context()
+
+  let status = "disconnected"
+  if (error) {
+    status = "error"
+  } else if (yourAddress && !error && provider !== null) {
```
Shadowfiend, line 11: This feels like something the web3 context should expose directly as an `isConnected` or something, though I'm not sure 🤔

solidity/dashboard/src/components/NetworkStatus.jsx
```diff
@@ -1,24 +1,27 @@
...
       />
-      <h5 className="text-grey-50">
-        {!yourAddress && "not connected"}
-        {yourAddress &&
-          !error &&
-          provider !== null &&
-          `connected: ${networkType}`}
-        {yourAddress && error && provider !== null && `wrong network`}
-      </h5>
+      <div className={`network-status__text--${status}`}>
```
Shadowfiend, line 20: Is this truly a `div`? Could we get away with no element or a `span`? It doesn't feel block-level at first blush.

Shadowfiend (approved):
I'm approving here, but the merge conflict isn't easily resolvable so I couldn't do a test run. I think once you handle the merge conflict you can merge the PR, as this functionality is central enough that I'll test it in-browser in any of the next PRs anyway.

PR #2108, opened 2020-10-01
Make rewards upgradeable

pdyraga:
This PR presents an alternative approach for Rewards upgrades. Compared to https://github.com/keep-network/keep-core/pull/2102, it's less graceful but allows reallocating tokens quicker.

# Background

`KeepRandomBeaconOperator` and `BondedECDSAKeepFactory` may get upgraded to the improved versions. When this happens, we want to incentivize stakers to move to the new contracts and do not lock the entire unallocated signer subsidy in rewards contract backing old operator contract versions.

# Competing approach

The approach presented in #2102 is more graceful. Rewards tokens that would normally be returned to the unallocated pool would instead be sent to the new rewards contract. This would gracefully bleed out the old rewards contract's reward tokens as the creation of new keeps (or beacon groups) dwindles over time as the stakers move to the new contract.

# Proposed approach

The approach presented here is less graceful than the one in #2102 but allows to relocate tokens quicker and lower the gas cost of allocating rewards.

Given that tokens are allocated at interval end, going with #2102 means that each interval of the new rewards contract needs to have 100% weight if we want to pay out the full amount of rewards from the interval defined on the old rewards contract. Additionally, it first requires allocating tokens on the old contract and has a risk that some malicious actor may race to allocate tokens on the new contract before we have a chance to move them there from the old contract for the given interval.

The approach presented here is a two-step process:
- upgrade is initiated,
- once the interval at which the upgrade was initiated ends, upgrade can be finalized.

When finalizing the upgrade, tokens are transferred to the new rewards contract but first, all past intervals are allocated. It means the upgrade mechanism may not change the amounts of tokens that would be distributed in any past or pending interval at the moment of initiating the upgrade.

### DIFF
--- a/solidity/contracts/Rewards.sol
+++ b/solidity/contracts/Rewards.sol
@@ -593,9 +593,8 @@ contract Rewards is Ownable {
             amountToTransfer,
             bytes("")
         );
-        if (!success) {
-            revert("Upgrade finalization failed");
-        }
+        require(success, "Upgrade finalization failed");
+        
 
         upgradeInitiatedTimestamp = 0;
     }
--- a/solidity/contracts/Rewards.sol
+++ b/solidity/contracts/Rewards.sol
@@ -20,6 +20,8 @@ import "openzeppelin-solidity/contracts/token/ERC20/SafeERC20.sol";
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "openzeppelin-solidity/contracts/math/Math.sol";
 
+import "./KeepToken.sol";
+
 /// @title KEEP Signer Subsidy Rewards
 /// @notice A contract for distributing KEEP token rewards to keeps.
 /// When a reward contract is created, the creator defines a reward schedule
@@ -67,9 +69,9 @@ import "openzeppelin-solidity/contracts/math/Math.sol";
 /// and the beacon operator contract acts as the "factory".
 contract Rewards is Ownable {
     using SafeMath for uint256;
-    using SafeERC20 for IERC20;
+    using SafeERC20 for KeepToken;
 
-    IERC20 public token;
+    KeepToken public token;
 
     // Array representing the percentage of unallocated rewards
     // available for each reward interval.
@@ -85,7 +87,8 @@ contract Rewards is Ownable {
     uint256 public totalRewards;
     // Rewards that haven't been allocated to finished intervals
     uint256 public unallocatedRewards;
-    // Rewards that have been paid out;
+    // Rewards that have been paid out from this contract - as a signer
+    // rewards or transferred to a new rewards contract.
     // `token.balanceOf(address(this))` should always equal
     // `totalRewards.sub(paidOutRewards)`
     uint256 public paidOutRewards;
@@ -112,7 +115,15 @@ contract Rewards is Ownable {
     // of the contract is possible with no owner's intervention.
     bool public funded = false;
 
+    // Owner of the contract may initiate an upgrade to a new rewards contract
+    // but the pending and past intervals must receive their rewards before any
+    // KEEP tokens are transferred out from this contract.
+    uint256 upgradeInitiatedTimestamp;
+    address newRewardsContract;
+
     event RewardReceived(bytes32 keep, uint256 amount);
+    event UpgradeInitiated(address newRewardsContract);
+    event UpgradeFinalized(uint256 amountTransferred);
 
     constructor (
         address _token,
@@ -121,7 +132,7 @@ contract Rewards is Ownable {
         uint256 _termLength,
         uint256 _minimumKeepsPerInterval
     ) public {
-        token = IERC20(_token);
+        token = KeepToken(_token);
         firstIntervalStart = _firstIntervalStart;
         intervalWeights = _intervalWeights;
         termLength = _termLength;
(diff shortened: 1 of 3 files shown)


eth-r:
This should work well if we don't mind the abrupt switch.

mhluongo:
This works on the biz / public incentives side and I think gives us a little more latitude in case of smart contract failure — I think it's the right call.

solidity/contracts/Rewards.sol
```diff
@@ -540,6 +551,55 @@ contract Rewards is Ownable {
...
+    /// initiated ended.
+    /// Before finalizing the upgrade, make sure all terminated groups are
+    /// reported.
+    function finalizeRewardsUpgrade() public onlyOwner {
+        require(upgradeInitiatedTimestamp != 0, "Upgrade not initiated");
+
+        // allocate all intervals that can be currently allocated
+        uint256 currentInterval = intervalOf(block.timestamp);
+        uint256 intervalToAllocate = 0;
+        if (currentInterval > 0) {
```
dimpar, line 575: My understanding of "0" intervals is that they all ended in September. Then from now on, we're dealing with intervals 1,2,3... so `currrentInterval` will always be > 0?

solidity/contracts/Rewards.sol
```diff
@@ -540,6 +551,55 @@ contract Rewards is Ownable {
...
+        paidOutRewards = paidOutRewards.add(amountToTransfer);
+
+        emit UpgradeFinalized(amountToTransfer);
+
+        bool success = token.approveAndCall(
+            newRewardsContract,
+            amountToTransfer,
+            bytes("")
+        );
+        if (!success) {
```
dimpar, line 596: maybe `require(success, "upgrade finalization failed")`?

solidity/test/rewards/TestRewardUpgrades.js
```diff
@@ -0,0 +1,222 @@
...
+
+    let token, rewards, newRewards
+
+    before(async () => {
+        token = await KeepToken.new({ from: owner })
+        newRewards = await NewRewardsStub.new()
+
+        firstIntervalStart = await time.latest()
+        timestamps = [
+            101, 150,    // 2 keep in interval 0
```
dimpar, line 34: Since here we are in beacon land, maybe we should use "groups" instead of "keeps"?

solidity/test/rewards/TestRewardUpgrades.js
```diff
@@ -0,0 +1,222 @@
...
+                newRewards.address,
+                {from: owner}
+            )
+            await expectRevert(
+                rewards.finalizeRewardsUpgrade({from: thirdParty}),
+                "Ownable: caller is not the owner."
+            )            
+        })
+
+        it("can not be finalized without initiating first", async () => {
```
dimpar, line 87: I think it should be written "[cannot](https://www.grammarly.com/blog/cannot-or-can-not/#:~:text=Can't%20is%20a%20contraction%20of%20cannot%2C%20and%20it's%20best,.%20.%20.%20but%20also.%E2%80%9D)" in these tests.

solidity/test/rewards/TestRewardUpgrades.js
```diff
@@ -0,0 +1,222 @@
...
+                newRewards.address,
+                {from: owner}
+            )
+            await expectRevert(
+                rewards.finalizeRewardsUpgrade({from: owner}),
+                "Interval hasn't ended yet"
+            )
+        })
+
+        it("can not be finalized another time without initiating again", async () => {
```
dimpar, line 105: another time, you mean in another interval?

solidity/contracts/Rewards.sol
```diff
@@ -85,7 +87,8 @@ contract Rewards is Ownable {
     uint256 public totalRewards;
     // Rewards that haven't been allocated to finished intervals
     uint256 public unallocatedRewards;
-    // Rewards that have been paid out;
+    // Rewards that have been paid out from this contract - as a signer
+    // rewards or transferred to a new rewards contract.
     // `token.balanceOf(address(this))` should always equal
     // `totalRewards.sub(paidOutRewards)`
     uint256 public paidOutRewards;
```
dimpar, line 94: I think that since KEEPs can be transferred to a new contract now, it is not technically "paid out" which associates with giving KEEPs to an operator in this context. Maybe we should rename it to ex. `dispensedRewards` or `distributedRewards` or something along this way?

solidity/test/rewards/TestRewardUpgrades.js
```diff
@@ -0,0 +1,222 @@
...
+const KeepToken = contract.fromArtifact('KeepToken')
+const RewardsStub = contract.fromArtifact("RewardsStub");
+const NewRewardsStub = contract.fromArtifact("NewRewardsStub");
+
+const BN = web3.utils.BN
+const chai = require('chai')
+chai.use(require('bn-chai')(BN))
+const expect = chai.expect
+
+describe.only("Rewards/Upgrades", () => {
```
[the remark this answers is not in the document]
pdyraga: lol it took me a while to figure out it's about that `only` 🤣

solidity/test/rewards/TestRewardUpgrades.js
pdyraga: `Rewards` contract thinks in terms of "keeps" and `BeaconRewards` contract translates "keeps" into "groups". Since we are testing the abstract `Rewards` contract here, this test talks about "keeps".

solidity/contracts/Rewards.sol
```diff
@@ -112,7 +115,15 @@ contract Rewards is Ownable {
     // of the contract is possible with no owner's intervention.
     bool public funded = false;
 
+    // Owner of the contract may initiate an upgrade to a new rewards contract
+    // but the pending and past intervals must receive their rewards before any
+    // KEEP tokens are transferred out from this contract.
+    uint256 upgradeInitiatedTimestamp;
```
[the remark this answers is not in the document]
pdyraga: I think we can even make them `public` - just checked, we are not getting problems with contract size.

solidity/test/rewards/TestRewardUpgrades.js
dimpar: Okay, it was just a bit confusing to me, because in `keep-core` we mostly use "groups" and this abstract rewards contract is in `keep-core` whereas in `keep-ecdsa` we use "keeps". Anyway, we can leave it.

solidity/contracts/Rewards.sol
```diff
@@ -540,6 +551,55 @@ contract Rewards is Ownable {
...
+    function initiateRewardsUpgrade(address _newRewardsContract) public onlyOwner {
+        upgradeInitiatedTimestamp = block.timestamp;
+        newRewardsContract = _newRewardsContract;
+        emit UpgradeInitiated(newRewardsContract);
+    }
+
+    /// @notice Finalizes the process of upgrading to another rewards contract
+    /// by allocating all past intervals and then, transferring the
+    /// not-yet-allocated tokens to a new rewards contract.
+    /// Can be called only when the interval during which the upgrade was
```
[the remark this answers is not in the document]
pdyraga: It fails when it's called before so I think "Can be called only" fits better here 🤔

solidity/contracts/Rewards.sol
pdyraga: There are two things to keep in mind:
- `Rewards` contract is abstract and it does not know when the first interval starts and ends,
- We are allocating previous intervals here - it does not matter if the interval ended but whether it was allocated.

solidity/contracts/Rewards.sol
pdyraga: Sure! More concise is better.

solidity/test/rewards/TestRewardUpgrades.js
pdyraga: No, I man calling `finalizeRewardsUpgrade` twice.

solidity/contracts/Rewards.sol
dimpar: Okay, makes sense.

solidity/contracts/Rewards.sol
pdyraga: I like it. I also had a feeling that something is wrong now with `paidOut*` part but couldn't find a better naming.

dimpar (approved):

PR #2111, opened 2020-10-06
Operator balance monitoring

### LINKED ISSUE #2171: Include mining waiter and block counter in rate limitter
Mining block waiter and block counter are not rate-limited in `keep-core` client. It has been fixed in `keep-ecdsa` https://github.com/keep-network/keep-ecdsa/pull/594 and now the same fix has to be applied to `keep-core`.

lukasz-zimnoch:
Closes https://github.com/keep-network/keep-core/issues/2163
Closes #2171
Refs: https://github.com/keep-network/keep-ecdsa/issues/551
Depends on: https://github.com/keep-network/keep-common/pull/51
Depends on: https://github.com/keep-network/keep-common/pull/59

Here we implement a simple balance monitor and enable it for the operator account balance. The monitoring routine checks the balance every 10 minutes and display a log error in case the balance falls below a threshold value passed from the config file.

### DIFF
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -39,9 +39,9 @@
 	# ConcurrencyLimit = 30
 	#
 	# BalanceAlertThreshold defines a minimum value of the operator's account
-        # balance below which an alert will be triggered.
-        #
-        # BalanceAlertThreshold = 500000000000000000 # 0.5 ETH (default value)
+	# balance below which an alert will be triggered.
+	#
+	# BalanceAlertThreshold = 500000000000000000 # 0.5 ETH (default value)
 
 [ethereum.account]
 	KeyFile            = "/Users/someuser/ethereum/data/keystore/UTC--2018-03-11T01-37-33.202765887Z--AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA8AAAAAAAAA"
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -37,6 +37,11 @@
 	# including view function calls.
 	#
 	# ConcurrencyLimit = 30
+	#
+	# BalanceAlertThreshold defines a minimum value of the operator's account
+    # balance below which an alert will be triggered.
+    #
+    # BalanceAlertThreshold = 500000000000000000 # 0.5 ETH (default value)
 
 [ethereum.account]
 	KeyFile            = "/Users/someuser/ethereum/data/keystore/UTC--2018-03-11T01-37-33.202765887Z--AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA8AAAAAAAAA"
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -39,7 +39,8 @@
 	# ConcurrencyLimit = 30
 	#
 	# BalanceAlertThreshold defines a minimum value of the operator's account
-	# balance below which an alert will be triggered.
+	# balance below which the client will start reporting errors in logs.
+	# The value should be passed in [wei].
 	#
 	# BalanceAlertThreshold = 500000000000000000 # 0.5 ETH (default value)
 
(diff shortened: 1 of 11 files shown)


pkg/chain/ethereum/balance_monitor.go
```diff
@@ -0,0 +1,70 @@
...
+	"math/big"
+	"time"
+
+	"github.com/keep-network/keep-core/pkg/chain"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// BalanceSource provides a balance info for the given address.
+type BalanceSource func(address common.Address) (*big.Int, error)
```
pdyraga, line 14: Do we need this function given that this code is located in `ethereum` package? Why not to call `WeiBalanceOf` directly (or `BalanceAt` per another comment)? I think we are adding another level of indirection here that is not really needed.

pdyraga (changes requested):

pkg/chain/ethereum/balance_monitor.go
lukasz-zimnoch: This code is also used by `keep-ecdsa` so this is why we need an indirection here.

pkg/chain/ethereum/ethereum.go
```diff
@@ -517,3 +519,15 @@ func (ec *ethereumChain) CalculateDKGResultHash(
...
+	return ec.accountKey.Address
+}
+
+func (ec *ethereumChain) WeiBalanceOf(address common.Address) (*big.Int, error) {
+	var result hexutil.Big
+
+	err := ec.clientRPC.Call(&result, "eth_getBalance", address, "latest")
+
+	return (*big.Int)(&result), err
+}
```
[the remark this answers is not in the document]
lukasz-zimnoch: Done. I've also switched the block counter and mining waiter to use the rate-limiting client by the way.

lukasz-zimnoch:
@pdyraga ready for the next round.

pkg/chain/ethereum/ethereum.go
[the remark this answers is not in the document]
pdyraga: Perfect! So this PR solves https://github.com/keep-network/keep-core/issues/2171 as well.

pkg/chain/ethereum/balance_monitor.go
pdyraga: Ah, I see. And `keep-ecdsa` has its own `ethereum` struct that is not referring to this one.

pdyraga:
We need to add an exception to `install.sh`. `BalanceAlertThreshold` property in `config.toml` is converted into a float:
```
2020-11-18T14:07:17.271+0100	FATAL	keep-main	error reading config file: unable to decode .toml file [/Users/piotr/go/src/github.com/keep-network/keep-core/configs/config.local.2.toml] error [toml: cannot load TOML value of type float64 into a Go integer]
```

lukasz-zimnoch:
> `BalanceAlertThreshold`

Done

configs/config.toml.SAMPLE
```diff
@@ -37,6 +37,12 @@
 	# including view function calls.
 	#
 	# ConcurrencyLimit = 30
+	#
+	# BalanceAlertThreshold defines a minimum value of the operator's account
+	# balance below which the client will start reporting errors in logs.
+	# The value should be passed in [wei].
+	#
+	# BalanceAlertThreshold = 500000000000000000 # 0.5 ETH (default value)
```
pdyraga, line 45: We should probably accept this value in Gwei. Right now, I can not set up a threshold alert for anything higher than 10 ETH. For small operators it is enough but for big ECDSA whales it's not enough if gas prices will skyrocket. 

Also, we accept `MaxGasPrice` in `Gwei` so those two values would be aligned.

lukasz-zimnoch: `MaxGasPrice` is passed in `wei` as well

pdyraga: I see. OK, let's leave it as-is for now. Do you mind opening a separate issue to change `MaxGasPrice` and `BalanceAlertThreshold` to Gwei?

lukasz-zimnoch: Here it is: https://github.com/keep-network/keep-core/issues/2176

pdyraga (approved):

PR #2114, opened 2020-10-07
Make sure upgrade initiated interval ends before finalizing the upgrade

pdyraga:

We were not properly checking whether the interval in which the upgrade was initiated ended before the upgrade was finalized. There was a unit test checking it but it was giving a false-positive given that the interval we were initiating on was 0 and it was not allocating any previous intervals.

We now have a separate `require` making sure the allocation condition holds as well as we have another unit test for non-zero interval.

I have also improved a comment in the code saying that the intervals must receive their reward before tokens are transferred out of the contract on upgrade. The right word to use here is "allocated", not "receive".


dimpar (approved):
lgtm

PR #2115, opened 2020-10-07
Dragons and invariants: Do not decrease dispensed rewards on finalizing upgrade

pdyraga:
Depends on https://github.com/keep-network/keep-core/pull/2114

We do not decrease `dispensedRewards` value when `finalizeUpgrade` is called to make the contract semantics easier and more transparent.

Additionally, we improve invariants defined in this contract by adding two that should always hold and clarifying the one already existing that is guaranteed to hold only right after `receiveApproval` is called.
    
Starting with invariants, we have two:
- `token.balanceOf(address(this)) >= totalRewards.sub(dispensedRewards)` should always hold,
- `token.balanceOf(address(this)) == totalRewards.sub(dispensedRewards)` should hold right after `receiveApproval` is called.

In regards to `finalizeUpgrade`, two options were considered:

**Option 1.** We do not decrease `totalRewards`, everything else stay as-is. In this case:
- `totalRewards` represents all KEEPs that came into this contract and were processed by `receiveApproval`,
- `dispensedRewards` represents all KEEPs that have already left this contract, either by `receiveReward` or `finalizeUpgrade`.

**Option 2.** We do decrease `totalRewards`, we do not increase `dispensedRewards`. In this case:
- `totalRewards` represents those KEEPs that came into this contract and were processed by `receiveApproval` and did not leave this contract by `finalizeUpgrade`.
- `dispensedRewards` represents all KEEPs that have already left this contract by `receiveReward`. It does not include those migrated to a new contract by `finalizeUpgrade`.

In addition to the invariants defined above, one more requirement has been added: after calling `finalizeUpgrade`, anyone can still call `receiveApproval` on the old contract and this call should not fail a produce a valid end state.

After considering two scenarios, it's been concluded that both Option 1 and 2 produce consistent outcomes and all invariants holds however Option 2 offers better semantics and that's what was done here.

Scenarios considered:

**Scenario 1.**  `totalRewards = 100`, `unallocatedRewards = 10`, `dispensedRewards = 90` and we initiate the upgrade.

  - With **Option 1**:
    - calling `finalizeUpgrade` produces end state: `totalRewards = 100`, `unallocatedRewards = 0`, `dispensedRewards = 100`, invariants holds.
    - calling `receiveApproval(5 KEEP)` after the upgrade has been finalized produces end state of: `totalRewards = 105`, `unallocatedRewards = 5`, `dispensedRewards = 100` and invariants still holds.

  - With **Option 2**: 
    - calling `finalizeUpgrade` produces end state: `totalRewards = 90`, `unallocatedRewards = 0`, `dispensedRewards = 90`, invariants holds.
    - calling `receiveApproval(5 KEEP)` after the upgrade has been finalized produces end state of: `totalRewards = 95`, `unallocatedRewards = 5`, `dispensedRewards = 90` and invariants still holds.

**Scenario 2.** `totalRewards = 100`, `unallocatedRewards = 90`, `dispensedRewards = 10` and we initiate the upgrade.

  - With **Option 1**:
    - calling `finalizeUpgrade` produces end state: `totalRewards = 100`, `unallocatedRewards = 0`, `dispensedRewards = 100`, invariants holds.
    - calling `receiveApproval(5 KEEP)` after the upgrade has been finalized produces end state of: `totalRewards = 105`, `unallocatedRewards = 5`, `dispensedRewards = 100` and invariants still holds

  - With **Option 2**:
    - calling `finalizeUpgrade` produces end state: `totalRewards = 10`, `unallocatedRewards = 0`, `dispensedRewards = 10`, invariants holds.
    - calling `receiveApproval(5 KEEP)` after the upgrade has been finalized produces end state of: `totalRewards = 15`, `unallocatedRewards = 5`, `dispensedRewards = 10` and invariants still holds.

### DIFF
--- a/solidity/test/rewards/TestRewardUpgrades.js
+++ b/solidity/test/rewards/TestRewardUpgrades.js
@@ -207,11 +207,10 @@ describe("Rewards/Upgrades", () => {
             // 50000 + 95000 = 145000
             expect(totalRewards).to.eq.BN(145000)
             expect(unallocatedRewards).to.eq.BN(0)
-            // nothing yet withdrawn but 855000 transferred to the new contract
-            expect(dispensedRewards).to.eq.BN(855000)
+            expect(dispensedRewards).to.eq.BN(0) // nothing yet withdrawn
         })
 
-        it("lets to withdraw outstanding rewards after migration", async () => {
+        it("lets to withdraw outstanding rewards after finalizing upgrade", async () => {
             await rewards.initiateRewardsUpgrade(
                 newRewards.address,
                 {from: owner}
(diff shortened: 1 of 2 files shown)


solidity/test/rewards/TestRewardUpgrades.js
```diff
@@ -231,5 +230,32 @@ describe("Rewards/Upgrades", () => {
...
+            await rewards.receiveReward(2, { from: beneficiary })
+
+            const totalRewards = await rewards.totalRewards()
+            const dispensedRewards = await rewards.dispensedRewards()
+            const unallocatedRewards = await rewards.unallocatedRewards()
+
+            // interval 0 allocates 50000
+            // interval 1 allocates 95000
+            // 50000 + 95000 = 145000
+            expect(totalRewards).to.eq.BN(145000)
```
dimpar, line 256: do you think we can add a comment here saying that `totalRewards = totalRewards - unallocatedRewards (before transfer)`?

pdyraga: Hmm... I think it's better explained by the previous tests:
- _moves all unallocated rewards to new contract_ explains where `855000` tokens go
-  _correctly updates reward balances_ explains - looking at this and previous test - why `145000` tokens stay in the contract

And the assertion pointed in this comment is inherited from that test. What we mostly care is `dispensedRewards` new value.

dimpar (approved):

Issue #2065, opened 2020-09-18
Remove instances of Vitesse in token dashboard and replace with Roboto Slab


liz-shinn:
Using Vitesse solely for branded instances (Keep logo, logo of the token dashboard dapp) and moving away from using a licensed font for within the dapp itself. 

Roboto Slab is an open sourced alternative to Vitesse that's a free Google Font. Replacing the instances of Vitesse within the body copy of the token dashboard in favor of this font. 

For example:

<img width="525" alt="Screen Shot 2020-09-18 at 5 38 17 PM" src="https://user-images.githubusercontent.com/57226633/93617082-cfed3100-f9d5-11ea-8f22-cd9bc8126e42.png">

Page title here (H1) is now Roboto Slab Light, 40pt, 48pt line height, -1.5p spacing
Secondary header (H2) is now Roboto Slab Bold, 32pt, 40pt line height, -1p spacing

PR #2125, opened 2020-10-15
Fonts updates in the KEEP token dashboard

r-czajkowski:
Closes: #2065

This PR updates the fonts styles in the KEEP token dashboard based on the KEEP Design System.


solidity/dashboard/src/css/typography.less
```diff
@@ -17,71 +17,81 @@ body {
...
 }
 
+h1, .h1,
+h2, .h2,
+h3, .h3,
+h4, .h4,
+h5, .h5,
+h6, .h6 {
+  font-family: 'Work Sans', sans-serif;
+  font-feature-settings: 'tnum' on, 'lnum' on, 'zero' on;
```
Shadowfiend, line 36: Are these two necessary, given inheritance from `body`? Are there cases where the headers need to have the font but intervening elements have changed it?

solidity/dashboard/src/css/typography.less
```diff
@@ -17,71 +17,81 @@ body {
...
+
 h1, .h1 {
-  font-family: 'Vitesse', sans-serif;
+  font-weight: 800;
+  font-size: 3rem;
+  line-height: 3.25rem;
+  letter-spacing: -2px;
+}
+
+.h1--alt {
```
Shadowfiend, line 47: Wondering if we have any way of expressing _why_ there's an `alt` here, i.e. name it in a more semantic way, or if this makes sense more as a class that can be mixed in separately. Would require a lot of reworking, so I don't think it's worth any time right now---but wanted to see if we could be thinking about that.

solidity/dashboard/src/css/typography.less
```diff
@@ -140,28 +159,9 @@ label {
...
-
-  &-big {
-    font-weight: normal;
-    font-size: 1.125rem !important;
-    line-height: 145%;
-    color: @color-grey-60;
-    font-feature-settings: 'tnum' on, 'lnum' on;
-  }
-
   &-samll {
```
Shadowfiend, line 162: Looks like a misspelling that's probably percolated through everything :grimacing:

solidity/dashboard/src/css/typography.less
```diff
@@ -222,20 +222,22 @@ label {
...
-    font-size: 0.7rem;
-    line-height: 1.375rem;
-    letter-spacing: 0.08rem;
+    font-weight: 500;
+    font-size: 0.5rem;
+    line-height: 0.75rem;
+    letter-spacing: 0.05rem;
   }
 
   &-intstruction {
```
Shadowfiend, line 231: Same re: misspelling/typo.

Shadowfiend (approved):
Approving here as everything looks good overall. Couple of things to flag, however:

- Font sizes in the right zone here look a bit small, both for buttons and text: Screen Shot 2020-10-26 at 09 14 42 IIRC this is undergoing some rework still, so that's probably fine.
- These bold 0s look very aggressive lol: Screen Shot 2020-10-26 at 09 15 31 The bold zeros in general feel a little in-your-face. /cc @liz-shinn 
- Applications / Keep Random Beacon navigation is misspelled (“Keep Rando Beacon”).
- Mouse movement between the web3 address and the dropdown can start the dropdown fading, which is a bit jarring. Couple of ways to manage this, including waiting to see if the mouse enters the dropdown before hiding it, but the easiest way would be to find some way to extend the element that does mouse tracking right up to the edge of the dropdown.
- The web3 dropdown has a pointer even when it's over things that aren't buttons—in particular, the amount of KEEP.

Let's roll this on in and keep trucking!

solidity/dashboard/src/css/footer.less
```diff
@@ -12,15 +12,16 @@ footer.footer {
         margin-bottom: 1rem;
         
         &__item {
+            line-height: 142%;
```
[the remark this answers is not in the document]
r-czajkowski: I got this value from Figma. In Figma `line-height` for this element is 17px or 142%.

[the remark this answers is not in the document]
r-czajkowski: w/o `line-height` there is too much space between items.

liz-shinn:
> * These bold 0s look very aggressive lol: Screen Shot 2020-10-26 at 09 15 31 The bold zeros in general feel a little in-your-face. /cc @liz-shinn

Think it will look more balanced when the unit is there as well, for example:
Screen Shot 2020-10-26 at 10 38 23 AM

Also, the addresses shouldn't be in all caps case since we ran into issue with case sensitivity with addresses

r-czajkowski:
I'll resolve this all issue in followup work, because almost all screens have changed so I'll go through page by page and update. In this PR, I focused on font updates on the css side.

solidity/dashboard/src/css/typography.less
r-czajkowski: heh 🤦  in UI components we use `text-small` 😄. It's probably the only place with this misspelling, I'll update.

solidity/dashboard/src/css/typography.less
r-czajkowski: same as above

solidity/dashboard/src/css/typography.less
r-czajkowski: Hmm based on the KEEP Design System `alt(alternative)` makes sense only with h1 and h2 tag. Out of idea which tag would fit better here 🤔

solidity/dashboard/src/css/buttons.less
```diff
@@ -32,27 +32,30 @@ button, .btn {
 
 .btn {
   &-lg {
-    padding: 1rem 2.5rem !important;
-    font-size: 1.125rem !important;
+    padding: 1rem 2.25rem !important;
+    font-size: 1.25rem !important;
+    line-height: 1.5rem !important
```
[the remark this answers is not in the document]
r-czajkowski: I'll take a look closer at it when I have a chance. Thanks for pointing this out 👍

solidity/dashboard/src/css/typography.less
r-czajkowski: pls see 092322f

solidity/dashboard/src/css/typography.less
r-czajkowski: pls see 092322f

r-czajkowski: pls see 092322f

solidity/dashboard/src/css/typography.less
r-czajkowski: pls see c3d1712

solidity/dashboard/src/css/footer.less
```diff
@@ -31,6 +32,6 @@ footer.footer {
     }
 
     .footer__app-version {
-        .footer-text-styles(0.5rem, @color-grey-40, 1.5rem);
+        .footer-text-styles(0.5rem, @color-grey-40, 1rem);
```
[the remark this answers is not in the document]
r-czajkowski: pls see c681076

Issue #2121, opened 2020-10-12
"relay entry requested event with previous entry [...] has been registered already" should not be logged as an error


pdyraga:
The situation with an event already registered is not an error and it might happen in certain Ethereum client deployments (nodes behind a load-balancer). This scenario does not deserve to be logged on error level and should be logged just as a warning.

This issue is similar to the one in `keep-ecdsa`: https://github.com/keep-network/keep-ecdsa/issues/560.

PR #2127, opened 2020-10-15
Logging "relay entry requested event with previous entry.." as warning

dimpar:
Closes https://github.com/keep-network/keep-core/issues/2121

In this PR we are changing logging level from Error to Warning. Event already registered is not an error and this scenario might happen in certain Ethereum client deployments where nodes are behind a load-balancer.


pdyraga (approved):

pdyraga:
Even a small chain reorg can make it happen. Good move.

Issue #2122, opened 2020-10-12
"could not confirm the expected relay request starting block..." should not be logged as an error


pdyraga:
Before we start the relay entry signing process we need to confirm the current relay request start block on the chain. This is to avoid having the client participating in an old relay request signing that has already completed with the rest of the signing group member clients and the result has been already published to the chain.

All this validation happens in `confirmCurrentRelayRequest` in `beacon.go` and works as expected. The only issue is that if the relay entry has been already submitted  to the chain and there is no other entry in progress (current request start block is 0), after 30 checks, client will log an error.

For this kind of a situation, we should log a warning or maybe even just info. But bear in mind - warning/info should be only the case for already submitted entry.

PR #2130, opened 2020-10-15
Replacing error with a warning when there is no entry in progress

dimpar:
Closes https://github.com/keep-network/keep-core/issues/2122

If the relay entry has been already submitted to the chain and there is no other entry in progress (current request start block is 0), then we should log a Warning instead of an Error.

### DIFF
--- a/pkg/beacon/beacon.go
+++ b/pkg/beacon/beacon.go
@@ -251,14 +251,24 @@ func confirmCurrentRelayRequest(
 			)
 			return
 		} else if i == maxRetries {
-			logger.Errorf(
-				"could not confirm the expected relay request starting block; "+
-					"the most recent one obtained from chain is [%v] and the "+
-					"expected one is [%v]; giving up after [%v] retries",
-				currentRequestStartBlock,
-				expectedRequestStartBlock,
-				maxRetries,
-			)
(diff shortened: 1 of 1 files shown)


pkg/beacon/beacon.go
```diff
@@ -251,14 +251,24 @@ func confirmCurrentRelayRequest(
...
 		} else if i == maxRetries {
-			logger.Errorf(
-				"could not confirm the expected relay request starting block; "+
-					"the most recent one obtained from chain is [%v] and the "+
-					"expected one is [%v]; giving up after [%v] retries",
-				currentRequestStartBlock,
-				expectedRequestStartBlock,
-				maxRetries,
-			)
+			if currentRequestStartBlock == 0 {
```
pdyraga, line 254: Looks very good, I'd just add a comment explaining that it's usually happening when entry was submitted very fast before this node received an event and/or was able to confirm request ID.

dimpar: Sure, done.

pdyraga (approved):

pdyraga:
This is quite tricky. We could even consider making it Info at some point but it's a double-edged sword - we don't know if the check is performed too late (Info) or too early (Warning).

Issue #2120, opened 2020-10-12
Failed to check if stale for group public key


pdyraga:
When DKG completes, client may report an error in logs saying `failed to check if stale for group with public key (...): Group does not exist`

This error is most probably related to the fact, the client is performing stale check for the group that completed DKG and the Ethereum client the request is directed to do not have the most recent state of the chain synced yet.

We should consider whather this situation really deserves an ERROR, or maybe it's just a WARNING.
Also, we should see if there a way to avoid performing stale check for the group just registered.

PR #2133, opened 2020-10-16
Failed to check if stale for group public key

dimpar:
Closes https://github.com/keep-network/keep-core/issues/2120

A node might run into situation when there is a slight delay with syncing to the recent state of the chain. It might result in error loggin "Group does not exist", even though it was just added to the chain. In order to avoid described scenario, we can skip checking the latest added group if it is a "stale group".

### DIFF
--- a/pkg/beacon/relay/registry/groups.go
+++ b/pkg/beacon/relay/registry/groups.go
@@ -1,6 +1,7 @@
 package registry
 
 import (
+	"bytes"
 	"encoding/hex"
 	"fmt"
 	"sync"
@@ -83,7 +84,7 @@ func (g *Groups) GetGroup(groupPublicKey []byte) []*Membership {
 // a new operation and it cannot have an ongoing operation for which it could be
 // selected before it expired. Such a group can be safely removed from the registry
 // and archived in the underlying storage.
-func (g *Groups) UnregisterStaleGroups() {
+func (g *Groups) UnregisterStaleGroups(latestGroupPublicKey []byte) {
 	g.mutex.Lock()
 	defer g.mutex.Unlock()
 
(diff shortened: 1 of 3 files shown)


pkg/beacon/relay/registry/groups.go
```diff
@@ -83,7 +84,7 @@ func (g *Groups) GetGroup(groupPublicKey []byte) []*Membership {
 // a new operation and it cannot have an ongoing operation for which it could be
 // selected before it expired. Such a group can be safely removed from the registry
 // and archived in the underlying storage.
-func (g *Groups) UnregisterStaleGroups() {
+func (g *Groups) UnregisterStaleGroups(latestGroupPublicKey []byte) {
```
pdyraga, line 87: This change makes me wonder whether we can optimize the number of calls to Ethereum client by:
- making `UnregistarStaleGroups` private (this parameter looks a little weird now)
- calling `unregisterStaleGroups` from `RegisterGroup` (now this parameter does not look weird at all).

We don't need to clean up the storage every time a new group is registered, especially if the client has relatlively low stake and do not participate in a lot of new groups. We can avoid an excessive number of `IsStale` calls this way.

pkg/beacon/relay/registry/groups_test.go
```diff
@@ -129,7 +129,7 @@ func TestUnregisterStaleGroups(t *testing.T) {
 
 	mockChain.markAsStale(signer2.GroupPublicKeyBytes())
 
-	gr.UnregisterStaleGroups()
+	gr.UnregisterStaleGroups(signer3.GroupPublicKeyBytes())
```
pdyraga, line 132: Can we add a simple assert `signer3.GroupPublicKeyBytes()` is not checked?

pkg/beacon/relay/registry/groups.go
dimpar: I'm confused a bit about this part "We don't need to clean up the storage every time a new group is registered..", but by calling `groups.unregisterStaleGroups()` from `groups.RegisterGroup()` this is exactly what we'll do, won't we?

pdyraga: Yes, but only for the current client, not in a global context. If you have just a min stake and you are qualifying to e.g. just 5% of new groups (random number, does not reflect the reality), you don't need to execute is stale check all the time, for 95% of DKGs, it's enough if you do it only when you qualified to a new group.

pkg/beacon/relay/registry/groups_test.go
dimpar: Yes, added.

pkg/beacon/relay/registry/groups.go
pdyraga: Sooo... this suggestion wasn't ideal as I just realized `RegisterGroup` is called for every member of the group so for highly staked nodes we increased the number of calls instead of reducing it. I think we should revert 7371160 and do it in the way originally proposed here 😬

pkg/beacon/relay/registry/groups_test.go
```diff
@@ -118,42 +117,72 @@ func TestLoadGroup(t *testing.T) {
 
 func TestUnregisterStaleGroups(t *testing.T) {
```
pdyraga, line 118: Instead of incorporating those assertions into this test, I'd implement a separate test checking whether `IsStale` was NOT executed for the last group passed - this is what we actually want to test. I'd also write "first group", "second group", and "third group" in assertions as it's hard to figure out by group public key what's happening when assertion fails.

```
func TestUnregisterStaleGroupsSkipLastGroupCheck(t *testing.T) {
	mockChain := &mockGroupRegistrationInterface{
		groupsToRemove:       [][]byte{},
		groupsCheckedIfStale: make(map[string]bool),
	}

	gr := NewGroupRegistry(mockChain, persistenceMock)

	gr.RegisterGroup(signer1, channelName1)
	gr.RegisterGroup(signer2, channelName1)
	gr.RegisterGroup(signer3, channelName1)

	gr.unregisterStaleGroups(signer3.GroupPublicKeyBytes())

	group1PublicKeyString := groupKeyToString(signer1.GroupPublicKeyBytes())
	if mockChain.groupsCheckedIfStale[group1PublicKeyString] != true {
		t.Fatalf("IsStaleGroup() was expected to be called for the first group")
	}

	group2PublicKeyString := groupKeyToString(signer2.GroupPublicKeyBytes())
	if mockChain.groupsCheckedIfStale[group2PublicKeyString] != true {
		t.Fatalf("IsStaleGroup() was expected to be called for the second group")
	}

	group3PublicKeyString := groupKeyToString(signer3.GroupPublicKeyBytes())
	if mockChain.groupsCheckedIfStale[group3PublicKeyString] != false {
		t.Fatalf("IsStaleGroup() was expected to not be called for the third group")
	}
}
```

dimpar: Agree, the original test looked cluttered with these many validations. Created a new one.

pkg/beacon/relay/registry/groups.go
dimpar: 😬 done

pdyraga (approved):

Issue #2079, opened 2020-09-21
Tooltip style


liz-shinn:
Update the previous old style of tooltip:
![Screen Shot 2020-09-21 at 4 34 24 PM](https://user-images.githubusercontent.com/57226633/93818420-a1cc5300-fc5a-11ea-824e-17f96e14238a.png)

To be this style of tooltip:
<img width="501" alt="Screen Shot 2020-09-21 at 4 33 32 PM" src="https://user-images.githubusercontent.com/57226633/93818485-b6105000-fc5a-11ea-9dea-ae362ddb99d3.png">

In Figma here: https://www.figma.com/file/ARAMFLMOPal0JewjvlZGhX/Keep-Design-System?node-id=1168%3A80

PR #2135, opened 2020-10-19
Tooltip updates

r-czajkowski:
Closes: #2079 

This PR updates the `Tooltip` component based on the Figma views.

obraz


solidity/dashboard/src/components/Tooltip.jsx
```diff
@@ -0,0 +1,79 @@
...
+          className="tooltip__content-wrapper"
+          onMouseEnter={() => setActive(true)}
+          onMouseLeave={hideTooltip}
+        >
+          {children ? (
+            children
+          ) : (
+            <>
+              <Tooltip.Header text={title} icon={IconComponent} />
+              <Tooltip.Divider />
```
Shadowfiend, line 56: The overall structure here seems a little more complicated that it might need to be for what we're trying to achieve… Nothing urgent, but would be nice to revisit at some point from the perspective of how we can use semantic markup and targeted styling to potentially collapse some of the layers a little bit.

Shadowfiend (approved):
Left some notes, but nothing blocking. Gave it a whirl, was thrown by the balance display being at 1000 KEEP then realized there's another PR actually hooking that number up to reality :grin:

solidity/dashboard/src/components/Tooltip.jsx
```diff
@@ -0,0 +1,79 @@
+import React, { useEffect, useState, useRef } from "react"
+import * as Icons from "./Icons"
+
+const Tooltip = ({
+  triggerComponent: TriggerComponent,
```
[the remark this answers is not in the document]
r-czajkowski: heh

It's about assigning a new name to variable.

From [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment):

> A property can be unpacked from an object and assigned to a variable with a different name than the object property.
> ```
> const o = {p: 42, q: true};
> const {p: foo, q: bar} = o;
>  
> console.log(foo); // 42 
> console.log(bar); // true
> ```
> Here, for example, const {p: foo} = o takes from the object o the property named p and assigns it to a local variable named foo.

solidity/dashboard/src/components/Tooltip.jsx
r-czajkowski: The main idea was to achieve maximum flexibility in case to create custom tooltip components via passing `children` and all necessary "logic" is inside tooltip component. But if skip `children` and pass `title`, `IconComponent` and `content` props we render the default tooltip component from [KEEP Design System](https://www.figma.com/file/ARAMFLMOPal0JewjvlZGhX/Keep-Design-System?node-id=1764%3A3)

solidity/dashboard/src/components/AddressShortcut.jsx
```diff
@@ -10,14 +11,22 @@ const AddressShortcut = ({ address, classNames }) => {
       toCopy={address}
       render={({ reset, copyStatus, copyToClipboard }) => {
         return (
-          <button
-            onClick={copyToClipboard}
-            onMouseOut={reset}
-            className={`address-shortcut tooltip address ${classNames}`}
+          <Tooltip
+            simple
+            triggerComponent={() => {
```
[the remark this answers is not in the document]
r-czajkowski: Hmm to be honest not follow. A Tooltip component handle all "logic" about showing and hiding tooltip. So we just pass a trigger component instead of handling feature outside tooltip component.

In that case you described we would have to pass a component to render to `showCopy` function and take care of hiding/showing every time when we want to use tooltip component.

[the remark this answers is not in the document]
Shadowfiend: My conceptualization here is that *where the tooltip is triggered* is not a concern of the tooltip, it's a concern of the caller using the tooltip—something I feel is hinted at by the hierarchy, where the caller is above the tooltip, and the trigger is never below the tooltip. In this formulation, the tooltip component provides all the tools to present tooltips, including functions to show and hide it, but actually hooking that up is something left to the caller.

solidity/dashboard/src/components/Tooltip.jsx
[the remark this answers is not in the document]
Shadowfiend: Aaaand I missed the curly braces on the opening line 😂

solidity/dashboard/src/components/Tooltip.jsx
Shadowfiend: Do we use that flexibility right now, or is it more hypothetical?

r-czajkowski: Probably more hypothetical, but we have some tooltips with redirect button and some without. Also, I think is better to have this flexibility instead of passing props like: `withBtn`, `topBtnProps`, `bottomBtnProps` etc. in case we would add a new type of tooltip.
Maybe we should just leave `{children}` and creates a something like `ResourceTooltip` that would render this:

obraz

Shadowfiend: That could be an interesting alternative to look into, definitely!

PR #2136, opened 2020-10-20
Improve rewards upgrade handling

eth-r:
Refs #1783

Finalizing a rewards upgrade can get blocked if the latest interval is allocated before the finalization. Fix this by only allocating intervals on finalization as required.

Send deallocated rewards or new funding to the new rewards contract if the upgrade has been finalized.

### DIFF
--- a/solidity/contracts/Rewards.sol
+++ b/solidity/contracts/Rewards.sol
@@ -120,6 +120,7 @@ contract Rewards is Ownable {
     // but the pending and past intervals must have their rewards allocated
     // before any KEEP tokens are transferred out from this contract.
     uint256 public upgradeInitiatedTimestamp;
+    uint256 public upgradeFinalizedTimestamp;
     address public newRewardsContract;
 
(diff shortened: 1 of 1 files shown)


solidity/contracts/Rewards.sol
```diff
@@ -601,6 +606,25 @@ contract Rewards is Ownable {
         
 
         upgradeInitiatedTimestamp = 0;
+        upgradeFinalizedTimestamp = block.timestamp;
+    }
+
+    /// @notice Return the given amount to the unallocated pool.
+    /// If the contract has been upgraded,
+    /// the deallocated amount will be sent to the new contract.
```
dimpar, line 614: `@param amount` is missing here.

dimpar (approved):
lgtm, leaving final approval to @pdyraga / @nkuba

nkuba (approved):
Code looks good to me. Tests are missing but they'll be added in https://github.com/keep-network/keep-core/pull/2137

PR #2137, opened 2020-10-21
Rewards deallocation tests

dimpar:
Depends on: https://github.com/keep-network/keep-core/pull/2136

Adding tests around:
- deallocation of rewards
- rewards allocation just before finalizing rewards contract upgrade.

### DIFF
--- a/solidity/test/rewards/TestRewardUpgrades.js
+++ b/solidity/test/rewards/TestRewardUpgrades.js
@@ -168,7 +168,7 @@ describe("Rewards/Upgrades", () => {
             )
         })
 
-        it("allocates interval before finalize rewards upgrade", async () => {
+        it("can be finalized with all previous intervals already allocated", async () => {
             await time.increase(termLength + 1) // interval 0 ends
 
             await rewards.initiateRewardsUpgrade(
--- a/solidity/test/rewards/TestRewardUpgrades.js
+++ b/solidity/test/rewards/TestRewardUpgrades.js
@@ -182,6 +182,27 @@ describe("Rewards/Upgrades", () => {
             await rewards.allocateRewards(1)
             await rewards.finalizeRewardsUpgrade({from: owner})
 
+            const allocation0 = await rewards.getAllocatedRewards(0)
+            const allocation1 = await rewards.getAllocatedRewards(1)
+            
+            expect(allocation0).to.eq.BN(50000) // 5% of 1000000
+            expect(allocation1).to.eq.BN(95000) // 10% of (1000000 - 50000)
+        })
+
+        it("should correctly update timestamps", async () => {
+            await time.increase(termLength + 1) // interval 0 ends
+
+            await rewards.initiateRewardsUpgrade(
+                newRewards.address,
+                {from: owner}
+            )
+
+            await time.increase(termLength + 1) // interval 1 ends
+            await rewards.setCloseTime(timestamps[2])
+
+            await rewards.allocateRewards(1)
+            await rewards.finalizeRewardsUpgrade({from: owner})
+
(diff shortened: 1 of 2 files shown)


eth-r (dismissed):
Looks good to me

solidity/test/rewards/TestRewardUpgrades.js
```diff
@@ -168,6 +168,60 @@ describe("Rewards/Upgrades", () => {
             )
         })
 
+        it("allocates interval before finalize rewards upgrade", async () => {
```
pdyraga, line 171: As far as I understand, what this test is checking, is whether the `finalizeRewardsUpgrade` works in case all previous intervals were allocated. It makes sense, though I think it's unclear from the test name. 

`finalizeRewardsUpgrade` do not allocate any interval in this case, because `await rewards.allocateRewards(1)` was called before `finalizeRewardsUpgrade`. 

The scenario with multiple intervals allocated is checked in the previous test (`"allocated all possible intervals"`).

What I think we can do here to improve the clarity is to, for example:
- rename `allocates all possible intervals` to `allocates all unallocated intervals when finalizing upgrade`,
- rename `allocates interval before finalize rewards upgrade` to `can be finalized with all previous intervals already allocated`. 

What do you think? It's possible I missed something.

solidity/test/rewards/TestRewardUpgrades.js
```diff
@@ -168,6 +168,60 @@ describe("Rewards/Upgrades", () => {
...
+            await rewards.setCloseTime(timestamps[2])
+
+            await rewards.allocateRewards(1)
+            await rewards.finalizeRewardsUpgrade({from: owner})
+
+            const upgradeInitiatedTimestamp = await rewards.upgradeInitiatedTimestamp()
+            const upgradeFinalizedTimestamp = await rewards.upgradeFinalizedTimestamp()
+
+            expect(upgradeInitiatedTimestamp).to.eq.BN(0)
+            expect(upgradeFinalizedTimestamp).not.to.eq.BN(0)
```
pdyraga, line 189: I'd extract those assertions into a separate test ensuring timestamps are correctly updated and instead, in this test, I would check allocations for intervals, just like in the previous one. We are testing interval allocations so let's assert interval allocations.

solidity/test/rewards/TestRewardUpgrades.js
```diff
@@ -168,6 +168,60 @@ describe("Rewards/Upgrades", () => {
...
+        })
+
+        it("transfers amount to a new contract upon rewards upgrade", async () => {
```
pdyraga, line 192: This test makes a lot of sense but the test name does not express what we are really testing here and at the first sight I was sure it's duplicating `""moves all unallocated rewards to new contract"` test case.

What do you think about renaming this test to e.g. `transfers any topped-up amount to a new contract after finalizing the upgrade`?

solidity/test/rewards/TestRewardUpgrades.js
dimpar: changed to `can be finalized with all previous intervals already allocated`

solidity/test/rewards/TestRewardUpgrades.js
dimpar: This test was rather about passing finalization process (hence checking the timestamp, cause they are at the end of the finalization function) and not allocations. I thought we had plenty of tests just for allocations values and that's why I didn't add allocations here. Anyway, changed now.

solidity/test/rewards/TestRewardUpgrades.js
dimpar: Yeah, sounds good, changed.

pdyraga (approved):

pdyraga:
Nice work!

PR #2141, opened 2020-10-27
Update app structure

r-czajkowski:
Adjust the app structure to the semantic hierarchy. The app layout starts from one of the most important parts- header and navigation. These containers tell the user the most information about what they're looking at and how the site hierarchy can be navigated. The next layout item is a content section that renders the page content. The last and least important layout item- footer is on the bottom of the app hierarchy. So to sum up, the current app structure is: 
```
header
 app-logo
 nav
content section
footer
```

but the visual structure looks like:
```
header content
footer content
```


solidity/dashboard/src/App.js
```diff
@@ -30,19 +30,24 @@ const App = () => (
...
+const AppHeader = () => {
+  return (
+    <header className="app__header">
+      <Link to="/" className="app-logo">
+        <Icons.KeepDashboardLogo />
+      </Link>
+      <SideMenu />
+    </header>
+  )
+}
```
Shadowfiend, line 49: One last suggestion here: I would encourage a rejiggering that makes `App` hold the layout and header directly (unless we're reusing them somewhere else), and wrap all of the context in `App` (Provider, Messages, etc) in a separate `AppHelpers` component. The thought here is that when you visit the App you want to first see what it's putting on the page—the components that wrap the content are largely mechanical, they don't tell us very much, so they can be moved along. Instead, if I look at the `App` component I should see the content: the header, the app content section, and the footer. Then I can move through those to look further down the hierarchy.

The overall goal that drives this kind of stuff for me is trying to make it as easy as possible to navigate the hierarchy and discover where content is coming from—working, in essence, to make the hierarchy a tool for understanding the app, and simplifying the navigation of that hierarchy. Extracting a mini-component like `AppHeader` can be useful in cases where it helps this goal, but if it isn't leading to reuse and doesn't have a super-deep hierarchy itself, it probably doesn't need the intermediate element. Simpler is better for navigation and understanding, and helps new folks get into the code more easily as well.

Not actually going to block on this though, the overall changes look good.

Shadowfiend (approved):
Let's do it :rocket:

PR #2142, opened 2020-10-27
Update external link styles

r-czajkowski:
Display the north east arrow as an HTML entity (`\2197`) after external link content instead of a custom svg icons.

obraz


solidity/dashboard/src/css/typography.less
```diff
@@ -101,24 +101,13 @@ a {
...
-    
-    &.grey {
-      &::after {
-        background-image: url("../static/svg/link-arrow-grey.svg");
-      }
+      // north east arrow code
+      content: '\2197';
+      margin-left: 0.25rem;
+      // Avoid `text-decoration` under `::after` content
+      display: inline-block;
```
Shadowfiend, line 110: Can't you just `text-decoration: none`? Also, do we _want_ the decoration not to cover the arrow? It's part of the hover hit area. This and the above change would also give us some consistency in how we handle things across our properties.

Last but not least: the approach I linked to above may be a good blanket approach here. No need to mark something with a class, you can match it based on the `href`.

Shadowfiend (changes requested):
Left a couple of notes about the overall approach; we're now doing this in several other sites/systems we're building, may make sense to aim for a consistent style across those.

solidity/dashboard/src/css/typography.less
r-czajkowski: > Can't you just text-decoration: none? Also, do we want the decoration not to cover the arrow? It's part of the hover hit area. This and the above change would also give us some consistency in how we handle things across our properties.

As I can see on Figma we skip `text-decoration` under the arrow. Pls [see](https://www.figma.com/file/ARAMFLMOPal0JewjvlZGhX/Keep-Design-System?node-id=1097%3A0)

r-czajkowski: >Last but not least: the approach I linked to above may be a good blanket approach here. No need to mark something with a class, you can match it based on the href.

Right, but we want an arrow only for external links.

r-czajkowski: and we can't just `text-decoration: none` because 
> Note that text decorations are not propagated to floating and absolutely positioned descendants, nor to the contents of atomic inline-level descendants such as inline blocks and inline tables.

[doc](https://www.w3.org/TR/CSS21/text.html#propdef-text-decoration)

solidity/dashboard/src/css/typography.less
```diff
@@ -101,24 +101,13 @@ a {
...
-
-    }
-    
-    &.grey {
-      &::after {
-        background-image: url("../static/svg/link-arrow-grey.svg");
-      }
+      // north east arrow code
+      content: '\2197';
+      margin-left: 0.25rem;
```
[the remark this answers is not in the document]
r-czajkowski: seems that spacing doesn't work. We also need `white-space: pre` pls [see].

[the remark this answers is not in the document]
Shadowfiend: 🤔 but this isn't at the end of a line, right? I would expect a space by itself not to work, but a space followed by the unicode character to render as expected.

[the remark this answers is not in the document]
Shadowfiend: Ah I see, this doesn't work because of the `inline-block` display, which makes the space be at the beginning of a block-level item, and thus unrendered. I think the right move here is to buck the design and underline the arrow. We can discuss with Liz + Michael on the back end, but right now that's the most consistent approach with how we've handled these elsewhere, so let's keep it consistent and then decide if we want to consistently switch to not underlining.

solidity/dashboard/src/css/typography.less
Shadowfiend: Ah yes, I remember running into this at some point. You can't override a parent's text-decoration if you're operating on an inline element.

r-czajkowski:
@Shadowfiend, fixed an issue with arrow. Going to merge this PR once the green light comes on.

PR #2149, opened 2020-11-03
Adding staker rewards escrow beneficiary contracts

dimpar:
In this PR we are adding 4 staker rewards beneficiary contracts. Each beneficiary contract can receive a withdrawal phase from a `PhasedEscrow` contract. Immediately transfers the received tokens to a designated `*RewardsEscrowBeneficiary` contract.
- `BeaconBackportRewardsEscrowBeneficiary`
- `BeaconRewardsEscrowBeneficiary`
-  `ECDSABackportRewardsEscrowBeneficiary`
- `ECDSARewardsEscrowBeneficiary`

### DIFF
--- a/solidity/contracts/PhasedEscrow.sol
+++ b/solidity/contracts/PhasedEscrow.sol
@@ -66,20 +66,27 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
         curveRewards = _curveRewards;
     }
 
-    function __escrowSentTokens(uint256 amount) external {
+    function __escrowSentTokens(uint256 amount) external onlyOwner {
         token.approve(address(curveRewards), amount);
         curveRewards.notifyRewardAmount(amount);
     }
 }
 
 /// @dev Interface of recipient contract for approveAndCall pattern.
-interface IStakerRewards { function receiveApproval(address _from, uint256 _value, address _token, bytes calldata _extraData) external; }
+interface IStakerRewards {
+    function receiveApproval(
+        address _from,
+        uint256 _value,
+        address _token,
+        bytes calldata _extraData
+    ) external;
+}
 
-/// @title BeaconBackportRewardsEscrowBeneficiary
-/// @notice A beneficiary contract that can receive a withdrawal phase from a
-///         PhasedEscrow contract. Immediately stakes the received tokens on a
-///         designated BeaconBackportRewardsEscrowBeneficiary contract.
-contract BeaconBackportRewardsEscrowBeneficiary is Ownable {
+/// @title StakerRewardsBeneficiary
+/// @notice An abstract beneficiary contract that can receive a withdrawal phase
+///         from a PhasedEscrow contract. Immediately stakes the received tokens
+///         on a designated rewards escrow beneficiary contract.
+contract StakerRewardsBeneficiary is Ownable {
     IERC20 public token;
     IStakerRewards public stakerRewards;
 
@@ -88,65 +95,67 @@ contract BeaconBackportRewardsEscrowBeneficiary is Ownable {
         stakerRewards = _stakerRewards;
     }
 
-    function __escrowSentTokens(uint256 amount) external {
-        token.approve(address(stakerRewards), amount);
-        stakerRewards.receiveApproval(address(this), amount, address(token), "");
+    function __escrowSentTokens(uint256 amount) external onlyOwner {
+        bool success = token.approve(address(stakerRewards), amount);
+        require(success, "Rewards token staking has failed");
+        
+        stakerRewards.receiveApproval(
+            address(this),
+            amount,
+            address(token),
+            ""
+        );
+    }
+}
+
+/// @title BeaconBackportRewardsEscrowBeneficiary
+/// @notice Stakes the received tokens on a designated
+///         BeaconBackportRewardsEscrowBeneficiary contract.
+contract BeaconBackportRewardsEscrowBeneficiary is StakerRewardsBeneficiary {
+    constructor(IERC20 _token, IStakerRewards _stakerRewards)
+        public
+        StakerRewardsBeneficiary(_token, _stakerRewards)
+    {
+        token = _token;
+        stakerRewards = _stakerRewards;
     }
 }
 
 /// @title BeaconRewardsEscrowBeneficiary
-/// @notice A beneficiary contract that can receive a withdrawal phase from a
-///         PhasedEscrow contract. Immediately stakes the received tokens on a
-///         designated BeaconRewardsEscrowBeneficiary contract.
-contract BeaconRewardsEscrowBeneficiary is Ownable {
-    IERC20 public token;
-    IStakerRewards public stakerRewards;
-
-    constructor(IERC20 _token, IStakerRewards _stakerRewards) public {
+/// @notice Stakes the received tokens on a designated
+///         BeaconRewardsEscrowBeneficiary contract.
+contract BeaconRewardsEscrowBeneficiary is StakerRewardsBeneficiary {
+    constructor(IERC20 _token, IStakerRewards _stakerRewards)
+        public
+        StakerRewardsBeneficiary(_token, _stakerRewards)
+    {
         token = _token;
         stakerRewards = _stakerRewards;
-    }
-
-    function __escrowSentTokens(uint256 amount) external {
-        token.approve(address(stakerRewards), amount);
-        stakerRewards.receiveApproval(address(this), amount, address(token), "");
     }
 }
 
 /// @title ECDSABackportRewardsEscrowBeneficiary
-/// @notice A beneficiary contract that can receive a withdrawal phase from a
-///         PhasedEscrow contract. Immediately stakes the received tokens on a
-///         designated ECDSABackportRewardsEscrowBeneficiary contract.
-contract ECDSABackportRewardsEscrowBeneficiary is Ownable {
-    IERC20 public token;
-    IStakerRewards public stakerRewards;
-
-    constructor(IERC20 _token, IStakerRewards _stakerRewards) public {
+/// @notice Stakes the received tokens on a designated
+///         ECDSABackportRewardsEscrowBeneficiary contract.
+contract ECDSABackportRewardsEscrowBeneficiary is StakerRewardsBeneficiary {
+    constructor(IERC20 _token, IStakerRewards _stakerRewards)
+        public
+        StakerRewardsBeneficiary(_token, _stakerRewards)
+    {
         token = _token;
         stakerRewards = _stakerRewards;
-    }
-
-    function __escrowSentTokens(uint256 amount) external {
-        token.approve(address(stakerRewards), amount);
-        stakerRewards.receiveApproval(address(this), amount, address(token), "");
     }
 }
 
 /// @title ECDSARewardsEscrowBeneficiary
-/// @notice A beneficiary contract that can receive a withdrawal phase from a
-///         PhasedEscrow contract. Immediately stakes the received tokens on a
-///         designated ECDSARewardsEscrowBeneficiary contract.
-contract ECDSARewardsEscrowBeneficiary is Ownable {
-    IERC20 public token;
-    IStakerRewards public stakerRewards;
-
-    constructor(IERC20 _token, IStakerRewards _stakerRewards) public {
+/// @notice Stakes the received tokens on a designated
+///         ECDSARewardsEscrowBeneficiary contract.
+contract ECDSARewardsEscrowBeneficiary is StakerRewardsBeneficiary {
+    constructor(IERC20 _token, IStakerRewards _stakerRewards)
+        public
+        StakerRewardsBeneficiary(_token, _stakerRewards)
+    {
         token = _token;
         stakerRewards = _stakerRewards;
-    }
-
-    function __escrowSentTokens(uint256 amount) external {
-        token.approve(address(stakerRewards), amount);
-        stakerRewards.receiveApproval(address(this), amount, address(token), "");
     }
 }
--- a/solidity/contracts/PhasedEscrow.sol
+++ b/solidity/contracts/PhasedEscrow.sol
@@ -75,6 +75,10 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
 /// @dev Interface of recipient contract for approveAndCall pattern.
 interface IStakerRewards { function receiveApproval(address _from, uint256 _value, address _token, bytes calldata _extraData) external; }
 
+/// @title BeaconBackportRewardsEscrowBeneficiary
+/// @notice A beneficiary contract that can receive a withdrawal phase from a
+///         PhasedEscrow contract. Immediately stakes the received tokens on a
+///         designated BeaconBackportRewardsEscrowBeneficiary contract.
 contract BeaconBackportRewardsEscrowBeneficiary is Ownable {
     IERC20 public token;
     IStakerRewards public stakerRewards;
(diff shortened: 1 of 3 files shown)


solidity/contracts/PhasedEscrow.sol
```diff
@@ -71,3 +71,82 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
         curveRewards.notifyRewardAmount(amount);
     }
 }
+
+/// @dev Interface of recipient contract for approveAndCall pattern.
+interface IStakerRewards { function receiveApproval(address _from, uint256 _value, address _token, bytes calldata _extraData) external; }
```
pdyraga, line 76: Let's apply some nicer formatting to this interface.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -71,3 +71,82 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
...
+
+/// @title BeaconBackportRewardsEscrowBeneficiary
+/// @notice A beneficiary contract that can receive a withdrawal phase from a
+///         PhasedEscrow contract. Immediately stakes the received tokens on a
+///         designated BeaconBackportRewardsEscrowBeneficiary contract.
+contract BeaconBackportRewardsEscrowBeneficiary is Ownable {
```
pdyraga, line 82: I don't think beneficiaries have to be `Ownable`. Same for `CurveRewardsEscrowBeneficiary` - I think we missed it. Is there any reason for having them `Ownable`?

solidity/contracts/PhasedEscrow.sol
```diff
@@ -71,3 +71,82 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
...
+    IStakerRewards public stakerRewards;
+
+    constructor(IERC20 _token, IStakerRewards _stakerRewards) public {
+        token = _token;
+        stakerRewards = _stakerRewards;
+    }
+
+    function __escrowSentTokens(uint256 amount) external {
+        token.approve(address(stakerRewards), amount);
+        stakerRewards.receiveApproval(address(this), amount, address(token), "");
```
pdyraga, line 112: Not that I expect any problems here, but I think it's worth calling `receiveApproval` only if `approve` [returned `true`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/85334995073fff746961155113b0906b689a044c/contracts/token/ERC20/ERC20.sol#L134).

solidity/test/TestPhasedEscrow.js
```diff
@@ -249,6 +268,157 @@ describe("PhasedEscrow", () => {
...
+        stakingContract = contracts.stakingContract
+        operatorContract = contracts.operatorContract
+
+        phasedEscrow = await PhasedEscrow.new(token.address, {from: owner})
+        const amount = web3.utils.toBN(baseBalance)
+        await token.transfer(phasedEscrow.address, amount, {from: owner})
+      })
+
+      beforeEach(createSnapshot)
+      afterEach(restoreSnapshot)
```
pdyraga, line 296: We don't need them, `beforeEach` and `afterEach` from lines 54-58 (under the main `describe` still work here).

solidity/contracts/PhasedEscrow.sol
```diff
@@ -71,3 +71,82 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
...
+    }
```
pdyraga, line 151: I do agree having 4 separate contracts is helpful as it's more transparent, easier to track where tokens go, and less chances of making a stupid mistake during the deployment. On the other hand, they have exactly the same code. Maybe we can extract the code to `StakerRewardsBeneficiary` contract and have 4 contracts inherit from it? This way, we can also simplify tests as we'll be testing only `StakerRewardsBeneficiary`.

solidity/test/TestPhasedEscrow.js
```diff
@@ -249,6 +268,157 @@ describe("PhasedEscrow", () => {
...
+
+      describe("BeaconBackportRewardsEscrowBeneficiary", () => {
```
pdyraga, line 298: Hmm... I see why you did that but I think we complicated the structure this way. Instead of:

```
PhasedEscrow
  when withdrawing to a CurveRewardsEscrowBeneficiary
  when withdrawing to a staker rewards escrow beneficiary
    BeaconBackportRewardsEscrowBeneficiary
    BeaconRewardsEscrowBeneficiary
    ECDSABackportRewardsEscrowBeneficiary
    ECDSARewardsEscrowBeneficiary
```

I think we should have:
```
PhasedEscrow
  when withdrawing to a CurveRewardsEscrowBeneficiary
  when withdrawing to a BeaconBackportRewardsEscrowBeneficiary
  when withdrawing to a BeaconRewardsEscrowBeneficiary
  when withdrawing to a ECDSABackportRewardsEscrowBeneficiary
  when withdrawing to a ECDSARewardsEscrowBeneficiary
```

Or - IMO even better - if we decide for a parant contract (previous comment), do:
```
PhasedEscrow
  when withdrawing to a CurveRewardsEscrowBeneficiary
  when withdrawing to a StakerRewardsBeneficiary
```

solidity/contracts/PhasedEscrow.sol
dimpar: Unless we are missing `onlyOwner` modifier in `__escrowSentTokens`?

solidity/test/TestPhasedEscrow.js
dimpar: We have a possibility to test against the actual `*BeaconRewards` contract that I set up in this test. But `*ECDSARewards` is a mock. If we decide to test only against the parent contract, then either we sacrifice the possibility of testing against real `*BeaconRewards` or not testing against a mock for `*ECDSARewards`. What do you think if we implement a parent contract as you suggested above, but having a second structure as you posted here?

solidity/contracts/PhasedEscrow.sol
pdyraga: Yeah, I took another look and we could go without `Ownable` for staker rewards but it's a good move to go with `Ownable` and `onlyOwner` on `__escrowSentTokens` for `CurveRewardsEscrowBeneficiary` as their `notifyRewardAmount` has `onlyRewardDistribution` modifier.

Let's do it consistend for all beneficiaries and add `onlyOwner` to `__escrowSentTokens` and make `PhasedEscrow` the owner.

solidity/test/TestPhasedEscrow.js
pdyraga: Yeah, I think it's a good move. Parent contract with the code but separate `describe` for each beneficiary:
```
PhasedEscrow
  when withdrawing to a CurveRewardsEscrowBeneficiary
  when withdrawing to a BeaconBackportRewardsEscrowBeneficiary
  when withdrawing to a BeaconRewardsEscrowBeneficiary
  when withdrawing to a ECDSABackportRewardsEscrowBeneficiary
  when withdrawing to a ECDSARewardsEscrowBeneficiary
```

solidity/contracts/PhasedEscrow.sol
dimpar: Done.

solidity/contracts/PhasedEscrow.sol
dimpar: Sound good, done.

solidity/contracts/PhasedEscrow.sol
dimpar: Done.

solidity/test/TestPhasedEscrow.js
dimpar: Removed.

solidity/test/TestPhasedEscrow.js
dimpar: Done.

solidity/contracts/PhasedEscrow.sol
dimpar: Sure, I think it's a good idea. Done.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -66,8 +66,96 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
...
+    IStakerRewards public stakerRewards;
+
+    constructor(IERC20 _token, IStakerRewards _stakerRewards) public {
+        token = _token;
+        stakerRewards = _stakerRewards;
+    }
+
+    function __escrowSentTokens(uint256 amount) external onlyOwner {
+        bool success = token.approve(address(stakerRewards), amount);
+        require(success, "Rewards token staking has failed");
```
pdyraga, line 100: Why *staking*? 🤔 We are not staking here. I'd say `Token transfer approval failed`.

I guess *staking* for Curve rewards makes sense, but here, I'd rather say about funding reward contracts.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -66,8 +66,96 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
...
+        uint256 _value,
+        address _token,
+        bytes calldata _extraData
+    ) external;
+}
+
+/// @title StakerRewardsBeneficiary
+/// @notice An abstract beneficiary contract that can receive a withdrawal phase
+///         from a PhasedEscrow contract. Immediately stakes the received tokens
+///         on a designated rewards escrow beneficiary contract.
```
pdyraga, line 88: I don't think we should talk about *staking* for staker rewards contracts. I'd rather say this contract is funding the designated staker rewards contract.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -66,8 +66,96 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
...
+            amount,
+            address(token),
+            ""
+        );
+    }
+}
+
+/// @title BeaconBackportRewardsEscrowBeneficiary
+/// @notice Stakes the received tokens on a designated
+///         BeaconBackportRewardsEscrowBeneficiary contract.
```
pdyraga, line 113: This contract is `BeaconBackportRewardsEscrowBeneficiary` so how can it stake to itself? All is it doing is funding the designated rewards contract.

solidity/test/TestPhasedEscrow.js
```diff
@@ -174,82 +194,220 @@ describe("PhasedEscrow", () => {
...
+    it("withdraws specified tokens from escrow", async () => {
+      await phasedEscrow.withdraw(transferAmount, {from: owner})
+
+      expect(await token.balanceOf(phasedEscrow.address)).to.eq.BN(
+        baseBalance - transferAmount,
+        "Unexpected amount withdrawn"
+      )
+    })
+
+    it("transfers specified tokens to beacon backport rewards contract", async () => {
```
pdyraga, line 382:
```
    when withdrawing to a CurveRewardsEscrowBeneficiary
      ✓ withdraws specified tokens from escrow (291ms)
      ✓ transfers specified tokens to beacon backport rewards contract (287ms).  <--- NOPE
```

pdyraga: Let's just drop `beacon backport` and we should be good.

solidity/contracts/PhasedEscrow.sol
dimpar: Correct, it is just transfer of tokens, not staking, we're not locking tokens here. Changed.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -66,8 +66,96 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
         curveRewards = _curveRewards;
     }
 
-    function __escrowSentTokens(uint256 amount) external {
+    function __escrowSentTokens(uint256 amount) external onlyOwner {
```
pdyraga, line 69: Given that this `onlyOwner` turns to be an important requirement, we should cover it with unit tests inside `TestPhasedEscrow.js`. Ideally, we should do the same for `StakerRewardsBeneficiary` as well:
```
describe("CurveRewardsEscrowBeneficiary", () => {
  describe("__escrowSentTokens", async () => {
    it("can be called by owner", async () => {

    })

    it("can not be called by non-owner", async () => {
    })    
  })
})
```

solidity/contracts/PhasedEscrow.sol
dimpar: Right, as above.

solidity/contracts/PhasedEscrow.sol
dimpar: Right, as above.

solidity/test/TestPhasedEscrow.js
dimpar: Done, removed.

solidity/contracts/PhasedEscrow.sol
dimpar: Added.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -66,8 +66,96 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
...
+    {
+        token = _token;
+        stakerRewards = _stakerRewards;
+    }
+}
+
+/// @title ECDSABackportRewardsEscrowBeneficiary
+/// @notice Trasfer the received tokens to a designated
+///         ECDSABackportRewardsEscrowBeneficiary contract.
+contract ECDSABackportRewardsEscrowBeneficiary is StakerRewardsBeneficiary {
```
nkuba, line 140: Wouldn't defining `ECDSABackportRewardsEscrowBeneficiary` and `ECDSARewardsEscrowBeneficiary` next to rewards contracts in `keep-ecdsa` be more accurate?

dimpar: One the one hand yes, but my take was this is a one time transfer of rewards and we wanted to have all beneficiaries in one place for simplicity.

solidity/test/TestPhasedEscrow.js
```diff
@@ -174,82 +198,294 @@ describe("PhasedEscrow", () => {
...
+          }),
+          "Ownable: caller is not the owner"
         )
       })
+    })
+  })
 
-      it("transfers specified tokens to curve rewards contract", async () => {
-        await phasedEscrow.withdraw(transferAmount, {from: owner})
+  describe("StakerRewardsBeneficiary", () => {
```
pdyraga, line 413: Let's define them on the same level as `PhasedEscrow` as they are covering a separate contract:
```
describe("StakerRewardsBeneficiary", () => {
  const owner = accounts[0]
  const thirdParty = accounts[1]

  const transferAmount = 1000

  let token
  let rewardsContract
  let rewardsBeneficiary

  before(async () => {
    token = await KeepToken.new({from: owner})
    rewardsContract = await TestECDSARewards.new(token.address)
    rewardsBeneficiary = await StakerRewardsBeneficiary.new(
      token.address,
      rewardsContract.address,
      {from: owner}
    )

    const amount = web3.utils.toBN(transferAmount)
    await token.transfer(rewardsBeneficiary.address, amount, {from: owner})
  })

  describe("__escrowSentTokens", async () => {
    it("can be called by the owner", async () => {
      await rewardsBeneficiary.__escrowSentTokens(transferAmount, {
        from: owner,
      })
      // ok, no revert
    })

    it("can not be called by the non-owner", async () => {
      await expectRevert(
        rewardsBeneficiary.__escrowSentTokens(transferAmount, {
          from: thirdParty,
        }),
        "Ownable: caller is not the owner"
      )
    })
  })
})
```

solidity/contracts/PhasedEscrow.sol
pdyraga: I think it's fine to have it here - `keep-core` defines all the foundation for token and how it is dispatched.

solidity/test/TestPhasedEscrow.js
dimpar: Right, done.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -66,8 +66,96 @@ contract CurveRewardsEscrowBeneficiary is Ownable {
...
+/// @title BeaconBackportRewardsEscrowBeneficiary
+/// @notice Transfer the received tokens to a designated
+///         BeaconBackportRewardsEscrowBeneficiary contract.
+contract BeaconBackportRewardsEscrowBeneficiary is StakerRewardsBeneficiary {
+    constructor(IERC20 _token, IStakerRewards _stakerRewards)
+        public
+        StakerRewardsBeneficiary(_token, _stakerRewards)
+    {
+        token = _token;
+        stakerRewards = _stakerRewards;
```
pdyraga, line 120: We don't need to set them given that we are calling the parent contract constructor, correct?

I think it should be just:
```
    constructor(IERC20 _token, IStakerRewards _stakerRewards)
        public
        StakerRewardsBeneficiary(_token, _stakerRewards) { }
```

dimpar: Correct.

pdyraga (approved):

solidity/contracts/PhasedEscrow.sol
nkuba: From migrations scripts perspective it would be better to define them in `keep-ecdsa` as ECDSA beneficaires need references to ECDSA Rewards contracts.

PR #2155, opened 2020-11-05
Added PhasedEscrow.withdrawFromEscrow function

pdyraga:
`withdrawFromEscrow` pulls all funds from a non-phased `Escrow` passed as a parameter. For `withdrawFromEscrow` to succeed, the `PhasedEscrow` has to be set as a beneficiary of the non-phased `Escrow`.

This PR is targeted to `solidity/v1.4.0` release branch and will have to be later backported to `master`.

### DIFF
--- a/solidity/contracts/PhasedEscrow.sol
+++ b/solidity/contracts/PhasedEscrow.sol
@@ -4,6 +4,8 @@ import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
 import "openzeppelin-solidity/contracts/token/ERC20/SafeERC20.sol";
 
+import "./Escrow.sol";
+
 interface IBeneficiaryContract {
     function __escrowSentTokens(uint256 amount) external;
 }
@@ -46,6 +48,13 @@ contract PhasedEscrow is Ownable {
         emit TokensWithdrawn(address(beneficiary), amount);
 
         beneficiary.__escrowSentTokens(amount);
+    }
+
+    /// @notice Withdraws all funds from a non-phased Escrow passed as
+    ///         a parameter. For this function to succeed, this PhasedEscrow
+    ///         has to be set as a beneficiary of the non-phased Escrow.
+    function withdrawFromEscrow(Escrow _escrow) public {
+        _escrow.withdraw();
     }
 }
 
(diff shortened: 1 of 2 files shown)


solidity/contracts/PhasedEscrow.sol
```diff
@@ -47,6 +49,13 @@ contract PhasedEscrow is Ownable {
 
         beneficiary.__escrowSentTokens(amount);
     }
+
+    /// @notice Withdraws all funds from a non-phased Escrow passed as
+    ///         a parameter. For this function to succeed, this PhasedEscrow
+    ///         has to be set as a beneficiary of the non-phased Escrow.
+    function withdrawFromEscrow(Escrow _escrow) public {
```
dimpar, line 56: I take that we don't care if a non-owner calls this functions since the address of beneficiary will be set by the owner?

solidity/test/TestPhasedEscrow.js
```diff
@@ -102,6 +103,40 @@ describe("PhasedEscrow", () => {
...
+      await escrow.setBeneficiary(phasedEscrow.address, {from: owner})
+      await phasedEscrow.withdrawFromEscrow(escrow.address)
+
+      const balanceAfter = await token.balanceOf(phasedEscrow.address)
+      expect(balanceAfter).to.eq.BN(amount)
+    })
+
+    it("pulls all funds from a non-phased Escrow when having some tokens", async () => {
+      const initialFunds = web3.utils.toBN(999)
+      await token.transfer(phasedEscrow.address, 999, {from: owner})
```
dimpar, line 128: I think you can pass `initialFunds` instead of `999`

solidity/contracts/PhasedEscrow.sol
pdyraga: Yes, [Escrow is Ownable](https://github.com/keep-network/keep-core/blob/solidity/v1.4.0/solidity/contracts/Escrow.sol) and the owner has to first set beneficiary via `setBeneficiary()` and then, the beneficiary can withdraw via `withdraw()`. I don't see a threat with this function being public as a non-phased `Escrow` can only drain funds to 0, we can not withdraw funds in portions.

nkuba (approved):

solidity/test/TestPhasedEscrow.js
pdyraga: Yeah, my bad. Fixed in 09f9529.

dimpar (approved):

PR #2158, opened 2020-11-09
Improve rewards contract flexibility

eth-r:
By separating the logic of intervals, allocations and payouts slightly more thoroughly from the logic of querying the number of keeps in an interval, it's possible to accommodate multiple factories in a single rewards contract without breaking compatibility with existing rewards implementations.


lionakhnazarov:
Closing — targets the legacy rewards contract (solidity-v1), which is being retired (#3947). No longer applicable to current main.

PR #2159, opened 2020-11-10
Allow PhasedEscrow to receive tokens via approveAndCall

pdyraga:
Added receiveApproval function to token escrow pulling all of the
approved tokens. We will use this functionality to update staker rewards
for the interval 3.

### DIFF
--- a/solidity/test/TestPhasedEscrow.js
+++ b/solidity/test/TestPhasedEscrow.js
@@ -2,29 +2,30 @@ const {initContracts} = require("./helpers/initContracts")
 const {accounts, contract, web3} = require("@openzeppelin/test-environment")
 const {createSnapshot, restoreSnapshot} = require("./helpers/snapshot.js")
 const {expectRevert, expectEvent} = require("@openzeppelin/test-helpers")
+const time = require("@openzeppelin/test-helpers/src/time")
+const crypto = require("crypto")
 
 const KeepToken = contract.fromArtifact("KeepToken")
 const Escrow = contract.fromArtifact("Escrow")
 const PhasedEscrow = contract.fromArtifact("PhasedEscrow")
+
+const TestSimpleBeneficiary = contract.fromArtifact("TestSimpleBeneficiary")
 const CurveRewardsEscrowBeneficiary = contract.fromArtifact(
   "CurveRewardsEscrowBeneficiary"
 )
 const StakerRewardsBeneficiary = contract.fromArtifact(
   "StakerRewardsBeneficiary"
 )
-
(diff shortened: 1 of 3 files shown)


solidity/test/TestPhasedEscrow.js
```diff
@@ -476,6 +509,69 @@ describe("StakerRewardsBeneficiary", () => {
     })
   })
 })
+
+describe("BeaconRewards to PhasedEscrow transfer", async () => {
+  const owner = accounts[0]
```
nkuba, line 514: We don't need to redefine `owner` as it's already done in L36.

solidity/test/TestPhasedEscrow.js
```diff
@@ -476,6 +509,69 @@ describe("StakerRewardsBeneficiary", () => {
...
+    let contracts = await initContracts(
+      contract.fromArtifact('TokenStaking'),
+      contract.fromArtifact('KeepRandomBeaconService'),
+      contract.fromArtifact('KeepRandomBeaconServiceImplV1'),
+      contract.fromArtifact('KeepRandomBeaconOperatorBeaconRewardsStub')
+    )
+
+    token = contracts.token
+    stakingContract = contracts.stakingContract
+    operatorContract = contracts.operatorContract
```
nkuba, line 530: Would this work?
```suggestion
    const {token, stakingContract, operatorContract} = await initContracts(
      contract.fromArtifact('TokenStaking'),
      contract.fromArtifact('KeepRandomBeaconService'),
      contract.fromArtifact('KeepRandomBeaconServiceImplV1'),
      contract.fromArtifact('KeepRandomBeaconOperatorBeaconRewardsStub')
    )
```

solidity/test/TestPhasedEscrow.js
pdyraga: It's another `describe`, owner from L36 is not available here. I wanted to keep it separately as we need to declare all contracts from scratch to perform this test (`initRewards`).

pdyraga: What also made me realize I missed declarations for some variables. Fixed in https://github.com/keep-network/keep-core/pull/2159/commits/058e0ca390eb4f4496395081d54516a8b54bf082.

solidity/test/TestPhasedEscrow.js
pdyraga: Unfortunately no - we need to have `token` and `operatorContract` declared on a global level so that they are accessible from `it`s.

nkuba (approved):

PR #2160, opened 2020-11-10
Update token grants page

r-czajkowski:
Depends on: #2150 

This PR updates the page where a user can preview all grants. Here we also updated the sec nav menu by grouping links by topic (some links are not available yet).

obraz

obraz

Grant preview
obraz


solidity/dashboard/src/components/CircularProgressBar.jsx
```diff
@@ -112,7 +116,11 @@ export const CircularProgressBars = React.memo(
           </g>
         </svg>
         <div className="mb-1">
-          {withLegend && items.map(renderProgressBarLegendItem)}
+          <ProgressBarLegendContext.Provider
+            value={{ displayLegendValuFn: displayAmountWithMetricSuffix }}
```
Shadowfiend, line 120: I believe this, and its derivatives in `ProgressBar.jsx`, should be `displayLegendValueFn`, unless that's taken for some reason?

solidity/dashboard/src/components/SideMenu.jsx
```diff
@@ -1,35 +1,72 @@
...
-          icon={<Icons.FeesVector />}
+        <li className="side-menu__route-wrapper">
+          <NavLink
+            to="/overview"
+            className="side-menu__route"
+            activeClassName="side-menu__route--active"
+          >
+            Overview
+          </NavLink>
+          <Divider style={styles.overviewDivider} />
```
Shadowfiend, line 23: I would encourage taking a look at this divider component with appropriate styling on the `li`. If not, I would consider dropping the intermediate component altogether, instead preferring to use `hr` directly. IMO the intermediate component doesn't seem to bring anything to the table here other than obscuring the final outcome.

solidity/dashboard/src/components/SideMenu.jsx
```diff
@@ -1,35 +1,72 @@
...
-        <NavLink
-          to="/applications"
-          label="applications"
-          icon={<Icons.Authorize />}
+        <NavLinkSection
+          label="earn"
+          icon={<Icons.FeesVector />}
+          subroutes={[
+            { label: "Earnings", path: "/earnings", exact: false },
+            { label: "Reweards", path: "rewards", exact: false },
```
Shadowfiend, line 58: Rewards*

solidity/dashboard/src/components/SideMenu.jsx
```diff
@@ -39,17 +76,36 @@ export const SideMenu = (props) => {
   )
 }
 
-const NavLink = ({ label, to, exact, icon }) => {
+const NavLinkSection = ({ label, icon, subroutes = [] }) => {
+  return (
+    <li className="side-menu__section">
+      <div className="side-menu__section__content">
```
Shadowfiend, line 82: Is this content or is it a header?

solidity/dashboard/src/pages/delegation/GrantedTokensPage.jsx
```diff
@@ -48,9 +45,11 @@ const GrantedTokensPageComponent = ({ onSubmitDelegateStakeForm }) => {
 
   useEffect(() => {
     if (!isEmptyArray(grants) && isEmptyObj(selectedGrant)) {
-      setSelectedGrant(grants[0])
+      const lookupGrantId = hash.substr(1)
+      const grantByHash = grants.find((grant) => grant.id === lookupGrantId)
+      setSelectedGrant(grantByHash || grants[0])
     }
-  }, [grants, selectedGrant, previousGrant.id])
+  }, [grants, selectedGrant, previousGrant.id, hash])
```
Shadowfiend, line 52: Why put this in the hash and not make it part of the URL? e.g. `/delegation/grant/<id>`.

solidity/dashboard/src/components/SideMenu.jsx
```diff
@@ -1,35 +1,72 @@
...
+          </NavLink>
+          <Divider style={styles.overviewDivider} />
+        </li>
+        <NavLinkSection
+          label="stake"
+          icon={<Icons.StakeDrop />}
+          subroutes={[
+            {
+              label: "Delegation",
+              path: "/delegation",
```
Shadowfiend, line 31: This name sits a little weird. It seems categorical, unlike the rest of the links (grants, applications, operations, etc). Perhaps “Delegations” would work? Since we also include information about other delegation activities.

Not really willing to die on this hill, just feels a little odd as is. cc @liz-shinn perhaps.

solidity/dashboard/src/components/Routing.jsx
```diff
@@ -28,7 +30,6 @@ class Routing extends React.Component {
         <Route exact path="/glossary">
           <Redirect to="/resources/quick-terminology" />
         </Route>
-        <Route exact path="/grant/:grantId" component={TokenGrantPreviewPage} />
```
Shadowfiend, line 31: Just to check---this isn't a URL that used to work before the redesign and now won't, right? If it is, let's introduce a redirect.

Shadowfiend (commented):
Left a few comments. This is largely looking good with the exception of the existing notes I left in #2150 around broken buttons/links.

liz-shinn:
Looks like "rewards" is misspelled on the left hand menu @r-czajkowski - not sure if we had already caught that in another issue.

liz-shinn:
Looks like "rewards" is misspelled on the left hand menu @r-czajkowski - not sure if we had already caught that in another issue.

solidity/dashboard/src/components/SideMenu.jsx
liz-shinn: Yup, I also think that "Delegations" is a good change. Makes sense to me @Shadowfiend

solidity/dashboard/src/components/CircularProgressBar.jsx
r-czajkowski: Here we wrap a legend section in `<ProgressBarLegendContext />` because component renders `ProgressBarLegendItem` which uses a context.

solidity/dashboard/src/components/Routing.jsx
r-czajkowski: I've set `TokenGrantPreviewPage` path to `/grant/:grantId` so redirect is no longer neede here.

solidity/dashboard/src/components/CircularProgressBar.jsx
Shadowfiend: I was obviously so tired I wasn't doing a good job of pointing out what I wanted to at this point haha. The note here was referring to a spelling issue, the missing `e` in `Value` in `displayLegendValueFn`.

Shadowfiend:
Max Stake button seems not to adjust for 18 decimals of precision:

Screen Shot 2020-11-30 at 14 50 07

That doesn't… Strictly belong in this PR though, sorry.

r-czajkowski:
> Max Stake button seems not to adjust for 18 decimals of precision:
> 
> Screen Shot 2020-11-30 at 14 50 07
> 
> That doesn't… Strictly belong in this PR though, sorry.

omg, indeed. sorryyyy. will fix!

solidity/dashboard/src/components/CircularProgressBar.jsx
r-czajkowski: sorry too 😆. I was out of brain power and thought we were wondering about why there is a `value={{ displayLegendValuFn: fn }}`instead of `<ProgressBarLegend displayAMuntFn={fn} />`

solidity/dashboard/src/components/DelegateStakeForm.jsx
```diff
@@ -113,7 +113,7 @@ const TokensAmountField = ({
   const { setFieldValue } = useFormikContext()
 
   const onAddonClick = () => {
-    setFieldValue("stakeTokens", availableToStake)
+    setFieldValue("stakeTokens", toTokenUnit(availableToStake).toString())
```
Shadowfiend, line 116: This looks like it's still blowing up (still doesn't show an error unfortunately):
```
Warning: An unhandled error was caught during low priority validation in <Formik validate /> Error: Error: [number-to-bn] while converting number "302619.18818968797564688" to BN.js instance, error: invalid number value. Value must be an integer, hex string, BN or BigNumber instance. Note, decimals are not supported. Given value: "302619.18818968797564688"
```
Looks like BigNumber returns a decimal value, but then the string gets routed to BN, and BN only takes integers. I think maybe we need to use `toFixed(0, BigNumber.ROUND_DOWN)` instead of `toString`.

solidity/dashboard/src/components/Routing.jsx
Shadowfiend: Is that a URL that's hooked up anywhere? Also, if we have `/token-grants` to display the list of grants, it seems reasonable to have `/token-grants/:grantId` be the place where you find a particular grant, right?

Not looking to bikeshed this further right now, we can handle this in follow-up work---but let's do a pass where we consider each URL and its place in the conceptual information hierarchy.

Shadowfiend (approved):
Ok, going to leave the rest of the Max Stake button fix for the redesign-dashboard branch; landing this PR for now, as the blocking concerns from the previous one have been landed :sunglasses:

Let's :shipit:

PR #2172, opened 2020-11-16
Add Random Beacon rewards page

r-czajkowski:
Depends on: #2160 
Closes: #2175  

Some changes against Figma views:
* "withdraw all" button is gone, as according to Figma comments,
* "Est. Allocation" and "Rewards History" are removed from "Rewards Details" based on Figma comments so in that case I've moved details below the amount of the rewards on the left side.
* the total amount refers to rewards withdrawn not those that have not yet been withdrawn.
 
obraz

- [x] testing with local contracts


Shadowfiend:
Believe package.json needs to be bumped to `@keep-network/keep-core` `1.4.1` here.

solidity/dashboard/src/pages/rewards/index.js
```diff
@@ -0,0 +1,16 @@
...
+import RewardsOverviewPage from "./RewardsOverviewPage"
+
+const RewardsPage = (props) => {
+  return <PageWrapper {...props} />
+}
+
+RewardsPage.route = {
+  title: "Rewards",
+  path: "/rewards",
+  pages: [RewardsOverviewPage /* RandomBeaconRewardsPage*/],
```
Shadowfiend, line 13: Wondering if we drop `RandomBeaconRewardsPage` altogether 🤔

solidity/dashboard/src/pages/rewards/RewardsOverviewPage.jsx
```diff
@@ -0,0 +1,78 @@
...
+  const dispatch = useDispatch()
+  const { yourAddress } = useWeb3Context()
+  const { beaconRewardsFetching, becaonRewardsBalance } = useSelector(
+    (state) => state.rewards
+  )
+
+  useEffect(() => {
+    dispatch({
+      type: "rewards/beacon_fetch_distributed_rewards_request",
+      payload: yourAddress,
```
Shadowfiend, line 20: Any chance we can at least make this work for the token holder as well, not just the beneficiary address? I actually think that might be a blocker, and it may make more sense to make this _only_ work for the token holder than to have it _only_ work for the beneficiary for this initial version.

solidity/dashboard/src/utils/beaconRewardsHelper.js
```diff
@@ -0,0 +1,35 @@
...
+
+  // There has to be at least 2 groups per interval to meet the group quota
+  // and distribute the full reward for the given interval.
+  static minimumBeaconGroupsPerInterval = 2
+
+  static intervalStartOf = (interval) => {
+    return moment
+      .unix(this.beaconFirstIntervalStart)
+      .add(interval * this.beaconTermLength, "seconds")
+  }
```
Shadowfiend, line 32: I'll say, a lot of this feels like it could be written to pull directly from the contracts rather than duplicating the numbers here. In fact it'll have to, since unallocated interval tokens are pushed to the next interval, and that's what's going to happen for the third interval of beacon rewards.

Not really a blocker right now though.

solidity/migrations/2_deploy_contracts.js
```diff
@@ -128,4 +129,13 @@ module.exports = async function(deployer, network) {
...
     KeepRandomBeaconOperator.address
   );
+
+  if(network === 'local') {
+    await deployer.deploy(
+        BeaconRewards,
+        KeepToken.address,
+        KeepRandomBeaconOperator.address,
+        TokenStaking.address,
+    )
```
Shadowfiend, line 139: Shouldn't this be deployed for development and Ropsten networks as well?

Shadowfiend (commented):
Left a few notes; in particular though, want to make sure token holders can see their rewards, even if it means the beneficiary account can't. If it's not a huge bundle of additional work, I would love for it to work for both beneficiary and token holder account.

Leaving unmerged for now while we land #2160.

solidity/dashboard/src/components/BeaconRewardsDetails.jsx
```diff
@@ -0,0 +1,59 @@
...
+    BeaconRewardsHelper.intervalStartOf(BeaconRewardsHelper.currentInterval + 1)
+  )
+
+  return (
+    <>
+      <h4 className="text-grey-70 mb-2">Beacon Rewards Details</h4>
+      <ul>
+        <li className="flex row center mb-1">
+          <Icons.Beacon width={12} height={12} />
+          <span className="text-small ml-1">Min. Keep Group</span>
```
Shadowfiend, line 23: I think this can be “Minimum required groups this interval”, or just “Min. Groups”.

solidity/dashboard/package.json
```diff
@@ -25,6 +25,7 @@
     "react-router-dom": "^5.1.2",
     "react-scripts": "^3.4.1",
     "react-transition-group": "^4.3.0",
+    "recharts": "^1.8.5",
```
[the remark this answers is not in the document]
r-czajkowski: For now, maybe we should use [React lazy loading](https://reactjs.org/docs/code-splitting.html#reactlazy) and rendering those components on demand(when a user goes to rewards page) ?

solidity/dashboard/src/components/RandomBeaconRewardsTable.jsx
```diff
@@ -0,0 +1,56 @@
...
+        header="group key"
+        field="groupPublicKey"
+        renderContent={renderGroupPublicKey}
+      />
+      {/* <Column header="operator" /> */}
+    </DataTable>
+  )
+}
+
+const renderAmount = ({ amount }) => <AmountCell amount={amount} />
```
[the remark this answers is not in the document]
r-czajkowski: Currently, we don't use `RandomBeaconRewardsTable` component. This component is based on the first version of rewards, not this simplified. But pushed this anyway for a future. Can we address it in the future?

solidity/migrations/2_deploy_contracts.js
r-czajkowski: If we could, it would be great! Since the app would probably crash on ropsten version(because we don't deploy rewards contract to ropsten network) and we would have to implement a workaround only for ropsten version of the dashboard.

solidity/dashboard/src/pages/rewards/RewardsOverviewPage.jsx
r-czajkowski: It's probably doable but probably lots of extra calls will be needed since rewards are transferred to a `beneficiary` and the beneficiary may have more than one operator. Also would have to call `getGroupMembers` and check if a given operator address is in a given group. The next step is computing a reward per operator. My point is that we don't have a direct match between transferred tokens from rewards and a given operator address. Probably it won't be very complicated, just a little bit tricky I guess.

Also, Rewards from `KeepRandomBeaconOperator` contract are displayed from the beneficiary's perspective.

solidity/dashboard/package.json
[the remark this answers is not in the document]
Shadowfiend: I'm not really worried enough about this to spend time working around it now---it's fine as-is for the time being. It's more of a general flag as something that may be worth looking at later in terms of general approach.

solidity/dashboard/src/pages/rewards/RewardsOverviewPage.jsx
Shadowfiend: A'ight, I think we can stick with what we've got for now.

solidity/migrations/2_deploy_contracts.js
Shadowfiend: Yeah, I think cc @pdyraga and let's not block the merge on it, but it makes sense to me to deploy these contracts on Ropsten, even if we leave them dormant somehow.

pdyraga: Agreed, it should be deployed to Ropsten. We were not doing an RC cut for reward contracts and that's the only reason why rewards contracts are not yet on Ropsten. I vote for removing the condition `if(network === 'local')` so that it all goes to Ropsten during the next RC cut.

Shadowfiend (approved):
And we are ready to rock :guitar:

Shadowfiend:
I'm actually _not_ going to wait for the builds here; instead, we'll deal with the builds in `redesign-dashboard`. The builds passed on #2177, which is effectively the same as this branch, since the last commit there merged this branch in.

liz-shinn:
There's not a `Withdraw` button on the screenshot of this page – just making sure we have the ability to withdraw these rewards? @Shadowfiend @r-czajkowski 

Screen Shot 2020-12-01 at 1 48 56 PM

PR #2177, opened 2020-11-21
tBTC rewards dapp

r-czajkowski:
Depends on: #2172 

This PR adds support for ECDSA rewards in a simplified version.

obraz

TODO:
- [x] testing with local contracts


solidity/dashboard/src/components/RewardsDetails.jsx
```diff
@@ -0,0 +1,111 @@
...
+  const nextIntervalStart = formatDate(
+    ECDSARewardsHelper.intervalStartOf(ECDSARewardsHelper.currentInterval + 1)
+  )
+
+  return (
+    <>
+      <h4 className="text-grey-70 mb-2">tBTC Rewards Details</h4>
+      <ul>
+        <MinKeepInInterval
+          label="Min. Keep"
```
Shadowfiend, line 88: I think this can be “Min. Deposits”. Keeps is strictly more correct, but I think deposits will be more familiar.

solidity/dashboard/src/pages/rewards/RewardsOverviewPage.jsx
```diff
@@ -54,19 +98,18 @@ const Balance = ({
...
-        <span className="text-validation">
-          The beneficiary account receives all withdrawn rewards.
-        </span>
-      </div> */}
+      {onWithdrawAll && (
+        <SubmitButton
+          onSubmitAction={onWithdrawAll}
+          className="btn btn-primary btn-lg mt-2"
+        >
+          withdraw all
```
Shadowfiend, line 106: I'm seeing this button even if there's nothing to withdraw, should we disable it in those cases?

Also, clicking on it changes it to say “Success” even if the process fails (e.g., if I change my address by overriding `yourAddress`, clicking the button won't work; similarly, if I click it and there are no tokens to withdraw, nothing happens). Not sure if that's normal, and not necessarily _hugely_ concerned, but worth a closer look.

solidity/dashboard/src/pages/rewards/RewardsOverviewPage.jsx
```diff
@@ -1,42 +1,86 @@
...
+        <section className="tile">
+          <Balance
+            title="tBTC Rewards"
+            rewardsBalance={ecdsaAvailableRewardsBalance}
+            isBalanceFetching={ecdsaAvailableRewardsFetching}
+            onWithdrawAll={onWithdrawECDSARewards}
+          />
+        </section>
+        <section className="tile">
+          <ECDSARewardsDetails pastRewards={ecdsaDistributedBalance} />
```
Shadowfiend, line 80: Were we intending on also displaying the rewards chart for tBTC? The distribution is different/higher here than for the beacon rewards... We should probably have neither or both (and it seems like neither is probably easier).

Net net we should have a consistent layout for the tBTC and beacon rewards, even if we don't have the same info available in both cases. I know for the beacon you dropped the middle pane because we were displaying less info, but I think the current result is strictly worse, as it's a bit visually unstable.

r-czajkowski: so what's a decision about chart for tBTC or we are going to drop beacon chart, thoughts?

Shadowfiend: If it's quick and easy I'd be fine with having the rewards chart for tBTC, but I also want to get this out last week, so I'm fine dropping the chart for the beacon. This is strictly a time-to-completion decision at this point. What we need: Keep and tBTC rewards to be visually consistent (2-3 panes, same width, ideally same height).

Let's do v2 from the designs, for both:

Screen Shot 2020-11-30 at 09 40 53

The beneficiary address can also be left off to get this shipped. For both rewards, let's also leave off the `?` next to the number of intervals left. For the beacon, we display both panes but skip the Withdraw All button.

r-czajkowski: Regarding beneficiary address, if we display rewards from beneficiary's perspective, should be current logged account address, right?

r-czajkowski: @Shadowfiend correct?
obraz

Shadowfiend: … Yeah, I think it makes less sense until we implement viewing from non-beneficiary addresses.

Shadowfiend (approved):
Hokayyy let's do this thing! Into beacon-rewards-dapp we go ↘️↙️

Issue #2167, opened 2020-11-13
Ocasionally failing TokenStakingEscrow test


pdyraga:
The following test is failing occasionally on some timing issue:
```
   1) TokenStakingEscrow
       withdrawRevoked
         withdraws part of deposited amount if something has been withdrawn before:
     AssertionError: expected 149999884259259259259260 to equal 150000000000000000000000
      at Context.<anonymous> (test/token_stake/TestTokenStakingEscrow.js:531:26)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

lionakhnazarov:
Closing as stale. Targets the legacy keep-core v1 (random beacon / KEEP staking) architecture, superseded by the Threshold v2 stack. Not actionable against current main.

PR #2179, opened 2020-11-23
Fixing test that occasionally fails

dimpar:
Closes #https://github.com/keep-network/keep-core/issues/2167

By increasing time programmatically using OpenZeppelin lib may add a split of a second to a current timestamp which results in difference of less than 1 KEEP during withdrawal of set amount. This PR takes this into account and accepts the diff if it is less than 1 KEEP.


lionakhnazarov:
Closing as stale. This 2020 flaky-test fix targets test code that has since been substantially rewritten on main.

Issue #2176, opened 2020-11-18
Pass `MaxGasPrice` and `BalanceAlertThreshold` properties in Gwei


lukasz-zimnoch:
Currently `MaxGasPrice` and `BalanceAlertThreshold` config properties are passed in `wei`. This can be insufficient for bigger values so it will be good to change the unit of these properties to `Gwei`.

nkuba:
We could also define them in the config as strings and handle as `big.Int` in the code.

@pdyraga any peference?

pdyraga:
Should we close this one @lukasz-zimnoch?

lukasz-zimnoch:
Yep I think so.

PR #2180, opened 2020-11-24
Configure Ether value with unit

nkuba:
Currently `MaxGasPrice` and `BalanceAlertThreshold` config properties are passed in `wei`. This can be insufficient for bigger values.

In this PR we are utilizing change introduced to `keep-common` (https://github.com/keep-network/keep-common/pull/60) that let us define a value in a config file with Ether unit, e.g. `500 Gwei` or `10.5 ether`.

Refs: https://github.com/keep-network/keep-core/issues/2176
Depends on: https://github.com/keep-network/keep-common/pull/60


cmd/start.go
```diff
@@ -42,16 +42,13 @@ const (
...
-	defaultBalanceAlertThreshold = 500000000000000000 // 0.5 ETH
-
-	// defaultBalanceMonitoringTick determines how often the monitoring
-	// check should be triggered.
-	defaultBalanceMonitoringTick = 10 * time.Minute
-)
+// Values related with balance monitoring.
+// defaultBalanceAlertThreshold determines the alert threshold below which
+// the alert should be triggered.
+var defaultBalanceAlertThreshold = big.NewInt(500000000000000000) // 0.5 ether
```
pdyraga, line 48: We could have a `WeiFromString` function declared in `github.com/keep-network/keep-common/pkg/chain/ethereum` that would let us handle this gracefully as: `ethereum.WeiFromString("0.5 ether")`

pdyraga (commented):
Were there any changes to the generated files so that we had to re-generate them again? Code generators do not seem to maintain the order of functions/fields so it's possible there were no functional changes but you have seen non-zero diff on the git level.

cmd/start.go
nkuba: This would be nice, but I'm not sure if the usage could be that elegant as the function would also have return an error returned by `UnmarshalText`. Maybe we could investigate that in a follow-up task as this requires changes to the keep-common repository.

What's more, we could update functions signatures (also in contracts code generators) to expect `ethereum.Wei` type instead of `big.Int`.

nkuba:
> Were there any changes to the generated files so that we had to re-generate them again? Code generators do not seem to maintain the order of functions/fields so it's possible there were no functional changes but you have seen non-zero diff on the git level.

We updated templates for code generators in https://github.com/keep-network/keep-common/pull/60.
Also I belive committed generated code might be behind the other changes to solidity contracts.

cmd/start.go
pdyraga: Yeah, agreed. It's a broader change. Right now, `ethereum.Wei` docs are clear about its usage:
> Wei is a custom type to handle Ether value parsing in configuration files

And we should expect nothing more. `var defaultBalanceAlertThreshold` is not a value from a configuration file and the way the conversion is handled later in this file is very graceful IMO.

go.sum
```diff
@@ -339,8 +339,10 @@ github.com/keep-network/cli v1.20.0 h1:mEufpPsovOVdduTTkk+a2CxS3crIrGFqUsvs58gSi
 github.com/keep-network/cli v1.20.0/go.mod h1:nzsst4JjU+rGE8Q5J839fYxectxWHpLhxKNohQWtQhA=
 github.com/keep-network/go-libp2p-bootstrap v0.0.0-20200423153828-ed815bc50aec h1:2pXAsi4OUUjZKr5ds5UOF2IxXN+jVW0WetVO+czkf+A=
 github.com/keep-network/go-libp2p-bootstrap v0.0.0-20200423153828-ed815bc50aec/go.mod h1:xR8jf3/VJAjh3nWu5tFe8Yxnt2HvWsqZHfGef1P5oDk=
-github.com/keep-network/keep-common v1.2.1-0.20201117163745-878626866ba7 h1:bmyyWcrg7dZ2TcgEIPV3d5uUrpoOncz9gEGBNg9eMHA=
-github.com/keep-network/keep-common v1.2.1-0.20201117163745-878626866ba7/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
+github.com/keep-network/keep-common v1.2.1-0.20201124134759-74da3a28a835 h1:jdEMdHaSWUW8/jfjISIeZdGHPBp4qjnFSS3Iq28WLaY=
+github.com/keep-network/keep-common v1.2.1-0.20201124134759-74da3a28a835/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
```
pdyraga, line 343: `go mod tidy` should remove those two lines.

pkg/chain/gen/tmp/cmd.go
```diff
@@ -0,0 +1,13 @@
+package cmd
```
pdyraga, line 1: I am confused, should we commit anything in `tmp` directory?

configs/config.toml.SAMPLE
```diff
@@ -42,7 +42,7 @@
 	# balance below which the client will start reporting errors in logs.
 	# The value should be passed in [wei].
 	#
-	# BalanceAlertThreshold = 500000000000000000 # 0.5 ETH (default value)
+	# BalanceAlertThreshold = "0.5 ether" # 0.5 ether (default value)
```
pdyraga, line 45: Let's tweak docs a little. `config.toml.SAMPLE` is the only documentation for the client configuration that exists now and we need to be very explicit here. 

We should explain this value can be passed in `wei`, `Gwei`, or `ether` with some examples, just like in the PR description for https://github.com/keep-network/keep-common/pull/60. We should also remove `# The value should be passed in [wei].`  comment as it is no longer up to date.

pdyraga, line 45: It's a little late but not too late since we haven't released this yet. Don't you think that setting value to `0.5 ETH` wouldn't be more natural than setting it to `0.5 ether`? Although the latter is also correct, the former is more widely used.

nkuba: `ether` is the correct name for the unit https://ethdocs.org/en/latest/ether.html#denominations. `ETH` is a currency name on the exchanges.

Our implementation is consistent with `web3js`: https://web3js.readthedocs.io/en/v1.2.11/web3-utils.html#towei

We shouldn't define `ETH` in the `unit` enum, as this is not a correct unit name.

pkg/chain/gen/tmp/cmd.go
nkuba: Yeah, we probably shouldn't. I'm not sure why we've got this file output by the code generator.

configs/config.toml.SAMPLE
pdyraga: OK, makes sense!

pdyraga:
Tested and works as advertised. Also, confirmed the change is backward-compatible so if someone has `BalanceAlertThreshold` or `MaxGasPrice` defined in the config file just as a number, this value is still interpreted as wei value.

pdyraga (approved):

PR #2181, opened 2020-11-27
fix typo

Leibniz137:
I just stumbled across this. Seems like this should be `RewardsPage` not `ReawrdsPage`, correct?

### DIFF
--- a/solidity/dashboard/src/pages/RewardsPageContainer.jsx
+++ b/solidity/dashboard/src/pages/RewardsPageContainer.jsx
@@ -6,7 +6,7 @@ import ReawrdsPage from "./RewardsPage"
 const RewardsPageContainer = (props) => {
   return (
     <Switch>
-      <Route exact path="/rewards/random-beacon" component={ReawrdsPage} />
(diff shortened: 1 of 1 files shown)


r-czajkowski:
Hi! Thanks for opening the PR!

We really appreciate your work but we've already fixed this at https://github.com/keep-network/keep-core/pull/2101 . These changes are not yet in a master branch (we are really close 🤞).

We encourage contributors to discuss a problem in issue before putting work in to make sure there isn't too much overlap of work.

Thanks again for your effort! 🤝

solidity/dashboard/src/pages/RewardsPageContainer.jsx
```diff
@@ -6,7 +6,7 @@ import ReawrdsPage from "./RewardsPage"
 const RewardsPageContainer = (props) => {
   return (
     <Switch>
-      <Route exact path="/rewards/random-beacon" component={ReawrdsPage} />
+      <Route exact path="/rewards/random-beacon" component={RewardsPage} />
```
r-czajkowski, line 9: We should also rename import above. This change would crash the dapp.

Issue #1824, opened 2020-05-22
Styling is off in the bonding section of tBTC


dimpar:
Also visible in [here](https://github.com/keep-network/keep-core/pull/1822#issuecomment-632669437)

![Screenshot 2020-05-22 at 14 50 56](https://user-images.githubusercontent.com/3156420/82676575-eca37300-9c46-11ea-93d2-96b4a9636257.png)

liz-shinn:
@dimpar I'd recommend setting a fixed width for all of these number input fields so that they're not changing in width & it doesn't expand and overlap with the buttons to the righthand side. 

Wouldn't change number of characters allowed in input – would scroll off to the right.

PR #2183, opened 2020-12-02
Fix bugs in KEEP token dashboard

r-czajkowski:
Closes: #1824 

This PR fixes a bunch of bugs from Antonio's notes. This PR:
* fixes fetching keep address from a `Deposit` contract. This bug was related to keep-network/tbtc@b0660f4,
* adds a temporary(because we don't have all on Figma yet) `Connect wallet` pages (Applications, Operations, Earnings page),
* fixes bug #1824,
* fixes overlapping tooltip content,
* adds loading feedback on the tBTC earning page,
* improves resource tooltip by setting a default link style instead of a link that looks like a button.

<details>
  <summary>Tooltip bug</summary>

### bug:
obraz

### fix:
obraz
</details>

<details>
  <summary>Bonding secion of tBTC</summary>

### bug:
obraz

### fix:
obraz

obraz

</details>

<details>
  <summary>Missing loading indicator in tBTC earnings page</summary>

obraz
</details>

<details>
  <summary>Improve resource tooltip:</summary>

obraz
</details>


michalsmiarowski:
**Bonding section of tBTC**

`Available Eth` has scrollbar on the bottom which i don't see on your screenshots (both on Chrome and Firefox):
image

Is this the desired look?

r-czajkowski:
> **Bonding section of tBTC**
> 
> `Available Eth` has scrollbar on the bottom which i don't see on your screenshots (both on Chrome and Firefox):
> image
> 
> Is this the desired look?

noope. On my Firefox the scrollbar is displayed only when scrolling and it looks much better than screenshot above 🤔 I'll dig into it

michalsmiarowski (changes requested):
To me the code itself looks good 👍 The only thing to check is the bottom scrollbar issue at the available field which i mentioned in the PR comment earlier.

Shadowfiend:
> On my Firefox the scrollbar is displayed only while scrolling

This is going to vary by system settings, amongst other things---and I would guess there will always be a scrollbar on non-macOS platforms. The way we've handled this elsewhere is to have a hover value that lets you see the full amount---we should be consistent about using that strategy everywhere IMO. Let's pick a precision (3-4 decimals seems reasonable), fix the width around that precision, and provide a hover that shows the full precision.

Shadowfiend:
It may also make sense to lay out the buttons in this table as their own `td`, and make the `Available ETH` header cell a `colspan="2"`. That should help the browser avoid overlap even in cases where things don't go quite as planned.

Shadowfiend:
Not a blocker, but other things I've noticed now that tBTC earnings is working:

- We round precision for the amount in the table, but no hover to show the full precision.
- The view in block explorer link links to Ropsten even if the dashboard is on mainnet.

Shadowfiend (approved):
Went through all of these, they look good overall. Flagged a couple of extra things I've noticed, none blocking or urgent. Let's roll it in.

r-czajkowski:
> The view in block explorer link links to Ropsten even if the dashboard is on mainnet.

If you run the dapp with `npm start`, it works in the development environment. In the development env, the dapp returns default chain ID otherwise gets chain id from `keep-network/*` artifacts. Pls [see](https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/connectors/utils.js#L25-L32). I believe it works at dashboard.keep.network.

PR #2184, opened 2020-12-02
Bump token dashboard release to v1.4.0

Shadowfiend:
Release notes to come in the release notes page.

This bundles the work in https://github.com/keep-network/keep-core/milestone/26?closed=1.

We should merge after things are deployed. Leaving as a draft in the meantime.

### DIFF
--- a/solidity/dashboard/package.json
+++ b/solidity/dashboard/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dashboard",
-  "version": "1.3.4",
+  "version": "1.4.0",
   "private": true,
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


solidity/dashboard/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dashboard",
-  "version": "1.3.4",
+  "version": "1.4.0",
```
pdyraga, line 3: Should we have the version in `package-lock.json` updated as well?

Shadowfiend: Oof. Let's… Leave it as-is, I don't think it has a material impact even on `npm ci`.

pdyraga (approved):

Issue #2188, opened 2020-12-03
Version number shows as "undefined" in sidebar menu


liz-shinn:
Version number for the token dashboard is showing as "Version undefined" in the menu sidebar instead of `version 1.4`

<img width="218" alt="Screen Shot 2020-12-03 at 1 58 14 PM" src="https://user-images.githubusercontent.com/57226633/101075967-b7ba9600-3570-11eb-96b0-ce1fc4d588af.png">

cc/ @r-czajkowski

r-czajkowski:
omg forgot to add `copy .env * / app /` to dashboard Dockerfile 🤦 will fix in a moment.

PR #2189, opened 2020-12-04
Fix dapp build process via docker

r-czajkowski:
Closes: #2188 

To get the current app version we added the `.env` file that contains `REACT_APP_VERSION` env variable. This env variable is provided by `create-react-app` under the hood and used in `Footer` component. For the `create-react-app` be able to inject env variable the .env file is needed during the build process and we missed the `COPY .env* /app/` in Dockerfile. This missed snippet of code should have been added in https://github.com/keep-network/keep-core/commit/e014af0bb2a1affff857a0ce037947bbdf6a85f4.

### Bug:
obraz

### Fix:
obraz

### DIFF
--- a/solidity/dashboard/Dockerfile
+++ b/solidity/dashboard/Dockerfile
@@ -12,6 +12,7 @@ RUN apk add --update --no-cache \
 
 COPY package.json /app/package.json
 COPY package-lock.json /app/package-lock.json
+COPY .env* /app/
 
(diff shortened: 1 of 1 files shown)


michalsmiarowski (dismissed):

michalsmiarowski (approved):

michalsmiarowski:
I tested locally and it looks ok 👍

PR #2190, opened 2020-12-04
Fix `withdraw tokens` btn

r-czajkowski:
The button should be disabled when there are no tokens to withdraw.

<details>
  <summary>Available to Withdraw label  bug</summary>

### Bug:
obraz

### Fix:
obraz

</details>

### DIFF
--- a/solidity/dashboard/src/components/ProgressBar.jsx
+++ b/solidity/dashboard/src/components/ProgressBar.jsx
@@ -1,7 +1,7 @@
 import React, { useMemo, useContext } from "react"
 import BigNumber from "bignumber.js"
 import CircularProgressBar from "./CircularProgressBar"
-import { percentageOf, sub } from "../utils/arithmetics.utils"
+import { percentageOf, sub, lt } from "../utils/arithmetics.utils"
 
 const defaultValue = 0
 const totalDefaultValue = 1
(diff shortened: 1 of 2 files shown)


michalsmiarowski:
I might be wrong but it looks like something is wrong with values of **Withdraw Unlocked Tokens** section when relogging into another account without refreshing. When I log in to my account1 with Metamask this is what i see:

image

which is correct. Then i log in to another Metamask account which I will call account2 (without refreshing), i go to another page just to make sure i relogged to another account (in this case i go to **Operations** page so i can see the values of `Beneficiary` and `Authorizer` changed) then go back to the **Delegations** -> **Granted Tokens** and i see the same values as on the screen above in **Withdraw Unlocked Tokens**, which are wrong. When i refresh the page and log in on account 2 again then the values are correct:

image

I don't think this is correct behaviour (and i'm not sure if it's related to this PR) but it will probably need more investigation. Thoughts?

michalsmiarowski (changes requested):
Code is good 👍 If the issue that i mentioned in the comment above is not related to this PR, then we can merge.

r-czajkowski:
> I might be wrong but it looks like something is wrong with values of **Withdraw Unlocked Tokens** section when relogging into another account without refreshing. When I log in to my account1 with Metamask this is what i see:
> 
> image
> 
> which is correct. Then i log in to another Metamask account which I will call account2 (without refreshing), i go to another page just to make sure i relogged to another account (in this case i go to **Operations** page so i can see the values of `Beneficiary` and `Authorizer` changed) then go back to the **Delegations** -> **Granted Tokens** and i see the same values as on the screen above in **Withdraw Unlocked Tokens**, which are wrong. When i refresh the page and log in on account 2 again then the values are correct:
> 
> image
> 
> I don't think this is correct behaviour (and i'm not sure if it's related to this PR) but it will probably need more investigation. Thoughts?

I'll open a PR that refreshes the app on `accountsChanged` event in a moment.

michalsmiarowski (approved):
Since the reload issue was addressed in https://github.com/keep-network/keep-core/pull/2196 i think we can merge this one.

PR #2193, opened 2020-12-04
Fix fetching ecdsa available rewards

r-czajkowski:
If the beneficiary has multiple operators, we should call `getWithdrawableRewards` one time in the given interval since `ECDSARewards` contract stores rewards per beneficiary not per operator.

https://github.com/keep-network/keep-ecdsa/blob/4fcb922ecaf319c0c4c14c2a9c5ee96b56756c31/solidity/contracts/ECDSARewards.sol#L155-L164


michalsmiarowski (approved):
I tested locally and it looks good 🚀 The `getWithdrawableRewards` methods is called once now for beneficiary who has multiple operators, which causes the reward value to be displayed properly

PR #2195, opened 2020-12-07
[Backport] Fix fetching ecdsa available rewards

r-czajkowski:
This PR is a backport of v1.4.1 fix done on releases/mainnet/token-dashboard/v1.4 release branch in PR #2193.

If the beneficiary has multiple operators, we should call `getWithdrawableRewards` one time in the given interval since `ECDSARewards` contract stores rewards per beneficiary not per operator.

https://github.com/keep-network/keep-ecdsa/blob/4fcb922ecaf319c0c4c14c2a9c5ee96b56756c31/solidity/contracts/ECDSARewards.sol#L155-L164.


michalsmiarowski (approved):
Looking good 👍

PR #2196, opened 2020-12-07
Refresh a dapp on `accountChanged` event

r-czajkowski:
This is a temporary solution to prevent a situation when a user changed an account but data has not been updated. After migrate to redux the dapp fetches data only once and updates data based on emitted events. This solution doesn't support a case where a user changed an account. We are going to address it in a follow-up work.


michalsmiarowski (approved):
The temporary solution works as expected - when we are changing the account the page refreshes and we need to log in again to see the data so there will be no case when we see wrong values 👍 As mentioned above we are going to address this further in a follow-up work

PR #2198, opened 2020-12-08
Fix subscription to TopUpCompleted event

r-czajkowski:
There was a typo in the event name. `TopUpCompleeted` -> `TopUpCompleted`.


michalsmiarowski (approved):
Tested locally and it's looking good - the TopUpCompleted event is called 👍

PR #2208, opened 2020-12-10
Support the new rewards mechanism in KEEP token dashboard

r-czajkowski:
Ref: https://github.com/keep-network/keep-ecdsa/pull/630

This PR adds support for the new rewards mechanism described in https://github.com/keep-network/keep-ecdsa/pull/644. In order to have all the required parameters to call the [`claim` function](https://github.com/keep-network/keep-ecdsa/blob/master/solidity/contracts/ECDSARewardsDistributor.sol#L72), the file generated by [merkle distributor](https://github.com/keep-network/keep-ecdsa/pull/636) is necessary. The output file should be in the `src/rewards-allocation/rewards.json` file. The dapp interprets `merkle_root` as a reward interval.

### This PR provides main changes:
* the new tbtc rewards page on which the beneficiary can view rewards history, view available rewards balance, and withdraw rewards (**Note: currently, the data on the tBTC rewards page is displayed from the beneficiary's perspective**),
* integration with `ECDSARewardsDistributor` contract and the output file generated by merkle distributor,
* `Timer` component #2212,
* separate pages for `Random Beacon` and `tBTC` instead of a single `Overview` page,
* update `start_dashboard.sh` script- generating necessary files to test the new rewards mechanism in KEEP token dashbord on the real merkle roots.

### This PR does not provide: 
* the `boost`, `eth score` and `APY` calculation because one beneficiary may have multiple operators and it is unclear how this information should be displayed in such a case,
* `witheld` status - at least for now - this information is not included in the input from reward distribution scripts,
* `accumulating` status - at least for now - computing this value is possible but also time-consuming when rendering the page.

### Release notes
For each interval, the dapp should be released with a file generated by merkle distributor which should contain previous merkle roots (intervals) and the new one. The file should be in the `src/rewards-allocation/rewards.json`. Note the `reward.json` file which is provided by this PR contains mock data and should not be used on the production build.

### Testing locally
To test the new rewards mechanism locally:
* please switch to `deploy-ecdsa-rewards` branch in `keep-ecdsa` repo or make sure the current branch includes files created in https://github.com/keep-network/keep-ecdsa/pull/630,
* run `start_dashbord.sh` script.

<details>
  
obraz

obraz

</details>


solidity/dashboard/src/pages/rewards/TBTCRewardsPage.jsx
```diff
@@ -0,0 +1,187 @@
...
+          </Tooltip>
+        </div>
+      </div>
+      <div className="rewards__overview__withdraw">
+        <SubmitButton
+          className="btn btn-primary btn-lg"
+          onSubmitAction={withdrawRewards}
+          disabled={!gt(balance, 0)}
+        >
+          withdraw all
```
michalsmiarowski, line 109: I think that we could add some width to this button to make it looks more to the one in the [figma](https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=7361%3A9712). Right now the button has less width than the text below which, in my opinion, doesn't look that good 🎨

solidity/dashboard/src/components/TBTCRewardsDataTable.jsx
```diff
@@ -1,84 +1,53 @@
...
+        </div>
       </div>
-    )
+    </div>
+  )
+}
+
+const renderStatus = ({ status }) => {
+  switch (status) {
+    case "AVAILABLE":
```
michalsmiarowski, line 44: Maybe we should use constants and put them in `constants.js` file instead of usign raw strings here?

solidity/dashboard/src/components/TBTCRewardsDataTable.jsx
```diff
@@ -1,84 +1,53 @@
...
+      return <Chip text={status} />
+    case "WITHDRAWN":
```
michalsmiarowski, line 46: Same as above

solidity/dashboard/src/pages/rewards/TBTCRewardsPage.jsx
```diff
@@ -0,0 +1,187 @@
...
+      <div className="rewards__overview__period">
+        <h5 className="text-grey-70">current rewards period</h5>
+        <span className="rewards-period__date">
+          {ECDSARewardsHelper.periodOf(ECDSARewardsHelper.currentInterval)}
+        </span>
+        <div className="rewards-period__remaining-periods">
+          <Icons.Time width="16" height="16" className="time-icon--grey-30" />
+          <span>84 rewards periods remaining&nbsp;</span>
+          <Tooltip simple delay={0} triggerComponent={Icons.MoreInfo}>
+            content
```
michalsmiarowski, line 99: Do we know what the content of the tooltip should be or we are just keeping it as 'content' for now?

solidity/dashboard/src/pages/rewards/TBTCRewardsPage.jsx
```diff
@@ -0,0 +1,187 @@
...
+        )}
+      </div>
+      <div className="rewards__overview__period">
+        <h5 className="text-grey-70">current rewards period</h5>
+        <span className="rewards-period__date">
+          {ECDSARewardsHelper.periodOf(ECDSARewardsHelper.currentInterval)}
+        </span>
+        <div className="rewards-period__remaining-periods">
+          <Icons.Time width="16" height="16" className="time-icon--grey-30" />
+          <span>84 rewards periods remaining&nbsp;</span>
```
michalsmiarowski, line 97: I assume this is the number of intervals remaining. In that case we should calculate it. We are keeping the intervals number in `ECDSARewards` class but i don't think we have a method to calculate how many intervals remains until the end, so I would create one in there.

michalsmiarowski (changes requested):
The code is looking good 🔥 I left a couple of notes below. Locally everything works as expected.

solidity/dashboard/src/pages/rewards/TBTCRewardsPage.jsx
r-czajkowski: pls see d0a3f55

solidity/dashboard/src/pages/rewards/TBTCRewardsPage.jsx
r-czajkowski: pls see d8d138a but I addressed it in a different way than you suggested. What do you think?

solidity/dashboard/src/components/TBTCRewardsDataTable.jsx
r-czajkowski: yeep 👍 we have already defined these statuses in `constants.js`. Will update!

solidity/dashboard/src/pages/rewards/TBTCRewardsPage.jsx
michalsmiarowski: I think it's ok, we can keep it that way 👌

r-czajkowski:
@michalsmiarowski ready for another 👀

michalsmiarowski (approved):
Looks ok. I think it is ready to be merged :boar:

michalsmiarowski:
One thing i noticed is that after clicking 'WITHDRAW ALL' button, and then rejecting the transaction in the Metamask, the button shows a success message (and also an error message at the top).

image

michalsmiarowski (approved):
Let's roll! 🚀

PR #2212, opened 2020-12-14
Implement timer for tBTC Rewards interval

michalsmiarowski:
This PR implements a timer on the tBTC Rewards page that shows when the next rewards will be released.

image

### DIFF
--- a/solidity/dashboard/src/components/Timer.jsx
+++ b/solidity/dashboard/src/components/Timer.jsx
@@ -1,12 +1,11 @@
 import React, { useEffect, useState } from "react"
+import moment from "moment";
 
-const Timer = ({ days, hours, minutes, seconds }) => {
-  const [remainingDays, setRemainingDays] = useState(days || 0)
-  const [remainingHours, setRemainingHours] = useState(hours || 0)
-  const [remainingMinutes, setRemainingMinutes] = useState(minutes || 0)
-  const [remainingSeconds, setRemainingSeconds] = useState(seconds || 0)
-
-  const zeroPad = (num, places) => String(num).padStart(places, "0")
+const Timer = ({ targetInUnix }) => {
+  const [remainingDays, setRemainingDays] = useState(0)
+  const [remainingHours, setRemainingHours] = useState(0)
+  const [remainingMinutes, setRemainingMinutes] = useState(0)
+  const [remainingSeconds, setRemainingSeconds] = useState(0)
 
   useEffect(() => {
     const myInterval = setInterval(() => {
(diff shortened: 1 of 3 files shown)


solidity/dashboard/src/pages/rewards/TBTCRewardsPage.jsx
```diff
@@ -8,15 +8,18 @@ import ProgressBar from "../../components/ProgressBar"
 import Timer from "../../components/Timer"
 import TBTCRewardsDataTable from "../../components/TBTCRewardsDataTable"
 import { colors } from "../../constants/colors"
+import { ECDSARewardsHelper } from "../../utils/rewardsHelper";
 
 const TBTCRewardsPage = () => {
+  const remainingTime = ECDSARewardsHelper.remainingTimeToFinishCurrentInterval()
```
r-czajkowski, line 14: I think we can easier. `const currentIntervalEndOf = ECDSARewardsHelper.intervalEndOf(ECDSARewardsHelper.currentInterval)` and then `<Timer target={currentIntervalEndOf} />`

solidity/dashboard/src/components/Timer.jsx
```diff
@@ -1,8 +1,54 @@
-import React from "react"
+import React, { useEffect, useState } from "react"
 
-const Timer = ({ target }) => {
-  // TODO implement timer
-  return <>{target}</>
+const Timer = ({ days, hours, minutes, seconds }) => {
+  const [remainingDays, setRemainingDays] = useState(days || 0)
```
r-czajkowski, line 4: We should sync props to state with `useEffect` eg:
```
useEffect(() => {
  setRemainingHours(props.hours)
 }, [props.hours]
)
```
Otherwise `useState` won’t initialize on props change.

solidity/dashboard/src/components/Timer.jsx
```diff
@@ -1,8 +1,54 @@
...
+          remainingMinutes === 0
+        ) {
+          clearInterval(myInterval)
+        } else {
+          setRemainingSeconds(59)
+          if (remainingMinutes > 0) {
+            setRemainingMinutes(minutes - 1)
+          } else if (remainingMinutes === 0) {
+            setRemainingMinutes(59)
+            if (remainingHours > 0) {
```
r-czajkowski, line 29: nested ifs... 🤯  out of idea now how we can do better, but let's keep 👀 on that.

r-czajkowski (changes requested):

solidity/dashboard/src/components/Timer.jsx
michalsmiarowski: One way to get cleaner code would be to keep the remainingSeconds only and then when displaying we could format it with moment js. However, I don't think it will be efficient to do such thing every re-render (which means every sec). Thoughts?

solidity/dashboard/src/components/Timer.jsx
```diff
@@ -1,8 +1,82 @@
...
+          }
+        }
+      }
+    }, 1000)
+    return () => {
+      clearInterval(myInterval)
+    }
+  })
+
+  useEffect(() => {
```
r-czajkowski, line 45: purely cosmetic thing. Could we set this hook as the first? It'll help to understand why we use `0` in `useState` hook.

r-czajkowski (commented):
Looks good 👍 just left one comment.

solidity/dashboard/src/components/Timer.jsx
r-czajkowski: > However, I don't think it will be efficient to do such thing every re-render (which means every sec)

Agree. Let's keep it in mind, maybe someday we'll do it better 😆

r-czajkowski (approved):
let's :shipit:

Issue #2095, opened 2020-09-28
Improve `TokenAmount` component


r-czajkowski:
The more I think about it the more it makes sense to simplify this and drop `displayAmountFunction`, `withTooltip`, and `tooltipText` entirely, and do all of the formatting internally.

To do this, we might rename `displayEthAmount` to `display18DecimalAmount` or similar. `TokenAmount` would instead allow passing in e.g. `smallestPrecisionDecimal` and `smallestPrecisionUnit`. For ETH, `currencySymbol` might be `ETH`, `14`, and `gwei`. TBTC might use this as `8` and `tSats`, for example, to display `<0.0001 TBTC` and a tooltip of `1000.01256 tSats`.

Thoughts?

_Originally posted by @Shadowfiend in https://github.com/keep-network/keep-core/pull/2050#discussion_r495274940_

PR #2220, opened 2020-12-17
Display tBTC rewards from the owner's/grantee's perspective

r-czajkowski:
This PR adds support for displaying the tBTC rewards from the `owner's/grantee's/operator's` perspective to use the KEEP token dashboard easier. In the previous implementation, the user was able to view tBTC rewards only from the `beneficiary's` perspective.

**What changed?**
* extracted the code snippets that fetch the grantee/managed grantee/copied delegations operators to the separate functions to be able to use this in other places,
* updated the saga that fetches the tBTC rewards to fetch rewards from the `owner's/grantee's/operator's` perspective.


solidity/dashboard/src/sagas/rewards.js
```diff
@@ -68,12 +75,40 @@ export function* watchWithdrawECDSARewards() {
...
+    // The same address can be used as beneficiary and owner. So addresses in
+    // array may be repeated. Let's convert to a Set to make sure the given
+    // address is in array only once.
+    const operators = new Set(
+      beneficiaryOperators
+        .concat(ownerOperators)
+        .concat(grenteeOperators)
+        .concat(mangedGranteeOperators)
+        .concat(copiedOperators)
+    )
```
michalsmiarowski, line 105: I think the UI is not user-friendly when we try to withdraw all tbtc rewards for more than one reward. For example when we have 5 rows in rewards history section

image

and we click **Withdraw All** button we have to do 5 MetaMask transactions on after another which might be a little problematic for higher number of rewards. I'm not sure how we can improve that tho 🤔 Any thoughts?

michalsmiarowski: There is also another issue with this. When, after clicking the **Withdraw All** button, I accept the first MetaMask transaction and then reject the second one I can't withdraw the rest of the rewards anymore and MetaMask shows a transaction error. After refresh it shows me that all the rewards are **Available**, which should not be the case because i withdrew the first one.

michalsmiarowski (changes requested):

michalsmiarowski:
Also i think we should remove the trailing zeroes from the main tBTC rewards section and just keep it as (for example) **1k** instead of **1.0k**

image

r-czajkowski:
> Also i think we should remove the trailing zeroes from the main tBTC rewards section and just keep it as (for example) **1k** instead of **1.0k**

We have an issue for improving the `TokenAmount` component #2095 . I think it's a good place to address an issue that you flagged.

solidity/dashboard/src/sagas/rewards.js
r-czajkowski: > I think the UI is not user-friendly when we try to withdraw all tbtc rewards for more than one reward. For example when we have 5 rows in rewards history section
> 
> image
> 
> and we click **Withdraw All** button we have to do 5 MetaMask transactions on after another which might be a little problematic for higher number of rewards. I'm not sure how we can improve that tho 🤔 Any thoughts?

We are limited by [`ECDSARewardsDistributor`](https://github.com/keep-network/keep-ecdsa/blob/master/solidity/contracts/ECDSARewardsDistributor.sol#L72) contract. We can't withdraw rewards for a given number of operators in a single transaction. Another solution is to add `withdraw` button per row.

PR #2216, opened 2020-12-16
Liquidity rewards

michalsmiarowski:
Depends on: #2224
Depends on: #2231

This PR adds new Liquidity Rewards page.

Main changes:
- add new **Liquidity Rewards** page
- add function that calculates the APY (although it is not used for now, since we are not sure about the results. All the APY results in UI are hardcoded and calculated for $1M pool )
- calculate % of the total pool for each liquidity pair
- calculate KEEP tokens reward for each liquidity pair
- add possibility to add more lp tokens for each liquidity pair
- add possibility to withdraw rewards for each liquidity pair
- santa hats for each card
- **NEW** label in the left menu for **Liquidity** page
- **NEW** label for the Liquidity page header

image

PR #2224, opened 2020-12-18
Liquidity rewards fetching data

r-czajkowski:
Ref: #2216 
This PR adds support for fetching and displaying the rewards data from the `LPRewards` contracts. All data are stored in the redux store.


solidity/dashboard/src/services/liquidity-rewards.js
```diff
@@ -0,0 +1,34 @@
+import { Web3Loaded, createERC20Contract } from "../contracts"
+
+// lp contract address -> wrapped ERC20 token as web3 contract instance
+const LPRewardsToWrappedTokenCache = {}
+
+// TODO implement functions
```
michalsmiarowski, line 6: The functions are implemented so I think we can remove this TODO 😉

solidity/dashboard/src/reducers/liquidity-rewards.js
```diff
@@ -0,0 +1,57 @@
...
+  TBTC_ETH: { ...wrappedTokenInitialData },
+  KEEP_TBTC: { ...wrappedTokenInitialData },
+}
+
+const liquidityRewardsReducer = (state = initialState, action) => {
+  if (!action.payload) {
+    return state
+  }
+
+  const { liquidityRewardPair, ...restPayload } = action.payload
```
michalsmiarowski, line 21: Hmmm... I would personally change the `liquidityRewardPair` variable name to `liquidityRewardPairName`. The `liquidityRewardPair` name may be misinterpreted as `LIQUIDITY_REWARD_PAIR` object from constants.js. 👍 if you agree or 👎 if you disagree.

solidity/dashboard/src/reducers/liquidity-rewards.js
```diff
@@ -0,0 +1,57 @@
...
+
+  switch (action.type) {
+    case `liquidity_rewards/${liquidityRewardPair}_fetch_data_start`:
```
michalsmiarowski, line 24: Maybe we should keep it all lowercased like other cases in the project? So it would be:

`liquidity_rewards/${liquidityRewardPair.toLowerCase()}_fetch_data_start`

What do you think?

solidity/dashboard/src/sagas/liquidity-rewards.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+}
+
+function* fetchLiquidityRewardsData(liquidityRewardPair, address) {
+  const contracts = yield getContractsContext()
+
+  const LPRewardsContract = contracts[liquidityRewardPair.contractName]
+  try {
+    yield put({
+      type: `liquidity_rewards/${liquidityRewardPair.name}_fetch_data_start`,
```
michalsmiarowski, line 27: Also change to lowercase if you agree with above comment.

michalsmiarowski (changes requested):
Left some comments here and there, but overall looks very good.

solidity/dashboard/src/sagas/liquidity-rewards.js
```diff
@@ -0,0 +1,79 @@
...
+      // Fetching total deposited liqidity tokens in the `LPRewards` contract.
+      const totalSupply = yield call(
+        fetchLPRewardsTotalSupply,
+        LPRewardsContract
+      )
+      // % of total pool in the `LPRewards` contract.
+      shareOfPoolInPercent = new BigNumber(lpBalance)
+        .div(totalSupply)
+        .multipliedBy(100)
+        .toFixed(2, BigNumber.ROUND_DOWN)
```
michalsmiarowski, line 53: I think this would look better if we remove the trailing 0's. So this values:

- 70.00%
- 70.20%
- 70.25%

should be displayed as:

- 70%
- 70.2%
- 70.25%

michalsmiarowski: Also it probably would be better to format it in the view instead of here.

solidity/dashboard/src/reducers/liquidity-rewards.js
r-czajkowski: I think it's fine and makes sense with uppercase since it's a kind of currency code, like USD PLN etc. Also in exchanges, the currency pairs are uppercased.

solidity/dashboard/src/services/liquidity-rewards.js
r-czajkowski: b09f05e

solidity/dashboard/src/reducers/liquidity-rewards.js
r-czajkowski: b09f05e

solidity/dashboard/src/sagas/liquidity-rewards.js
r-czajkowski: ce6e7ea

r-czajkowski:
@michalsmiarowski  I'm ready for the second round!

solidity/dashboard/src/sagas/liquidity-rewards.js
r-czajkowski: After thinking the whole thing I do agree with adding trailing format and rounding to two decimal places in the view. Since it's possible that the % of the total pool may be less than 0.01 so should we display `<0.01 %`, so we should store the total percentage value in store but convert it to a dedicated format in a component.

solidity/dashboard/src/sagas/liquidity-rewards.js
```diff
@@ -0,0 +1,80 @@
...
+      // Fetching total deposited liqidity tokens in the `LPRewards` contract.
+      const totalSupply = yield call(
+        fetchLPRewardsTotalSupply,
+        LPRewardsContract
+      )
+      // % of total pool in the `LPRewards` contract.
+      shareOfPoolInPercent = new BigNumber(lpBalance)
+        .div(totalSupply)
+        .multipliedBy(100)
+        .decimalPlaces(2)
```
michalsmiarowski, line 53: We should add BigNumber.ROUND_DOWN as second argument here (similar as it was before, when we used toFixed method).

michalsmiarowski (changes requested):
One more thing and we are ready to merge it to `liquidity-rewards` branch

solidity/dashboard/src/sagas/liquidity-rewards.js
r-czajkowski: 7547b7c

michalsmiarowski (approved):
Everything is ok! Tested it locally and it seems to work as expected! It's ready to be merged to `liquidity-rewards` branch.

PR #2225, opened 2020-12-18
Batched Phased Escrow

nkuba:
It's an implementation of `PhasedEscrow` that allows batched withdrawals to multiple beneficiaries in one function call.


nkuba:
Replaced by https://github.com/keep-network/keep-core/pull/2230

PR #2230, opened 2020-12-21
BatchedPhasedEscrow

pdyraga:
Possibly replaces #2225.

BatchedPhasedEsrow` is a token holder contract allowing contract owner to
approve a set of beneficiaries of tokens held by the contract, to
appoint a separate drawee role, and allowing that drawee to withdraw
tokens to approved beneficiaries in phases.

This contract is not the ultimate solution for a secure escrow with a
separate role withdrawing tokens. Notably, it does not limit the amount
or frequency of withdrawals so compromised drawee can drain the escrow
funds to one of the selected approved beneficiaries. The escrow owner
should proceed with caution and do not transfer all funds to
`BatchedPhasedEscrow`. Instead, funds should be transferred in tranches,
depending on how drawee spends them.

`CurveRewards` and `LPTokenRewards` contracts from `keep-ecdsa` that we are
are going to deploy for Uniswap pairs (see https://github.com/keep-network/keep-ecdsa/pull/645) 
have the same `notifyRewardAmount` interface. To do not duplicate the code,
 I abstracted `CurveRewardsEscrowBeneficiary` contract into a generic
`StakingPoolRewardsEscrowBeneficiary` contract.

### DIFF
--- a/solidity/contracts/PhasedEscrow.sol
+++ b/solidity/contracts/PhasedEscrow.sol
@@ -155,7 +155,7 @@ contract BatchedPhasedEscrow is Ownable {
             IBeneficiaryContract beneficiary = beneficiaries[i];
             require(
                 isBeneficiaryApproved(beneficiary),
-                "Beneficiary not approved"
+                "Beneficiary was not approved"
             );
             withdraw(beneficiary, amounts[i]);
         }
--- a/solidity/contracts/PhasedEscrow.sol
+++ b/solidity/contracts/PhasedEscrow.sol
@@ -96,18 +96,6 @@ contract BatchedPhasedEscrow is Ownable {
         drawee = msg.sender;
     }
 
-    /// @notice Funds the escrow by transferring all of the approved tokens
-    ///         to the escrow.
-    function receiveApproval(
-        address _from,
-        uint256 _value,
-        address _token,
-        bytes memory
-    ) public {
-        require(IERC20(_token) == token, "Unsupported token");
-        token.safeTransferFrom(_from, address(this), _value);
-    }
-
     /// @notice Approves the provided address as a beneficiary of tokens held by
     ///         the escrow. Can be called only by escrow owner.
     function approveBeneficiary(
@@ -136,6 +124,18 @@ contract BatchedPhasedEscrow is Ownable {
         require(newDrawee != address(0), "New drawee can not be zero address");
         emit DraweeRoleTransferred(drawee, newDrawee);
         drawee = newDrawee;
+    }
+
+    /// @notice Funds the escrow by transferring all of the approved tokens
+    ///         to the escrow.
+    function receiveApproval(
+        address _from,
+        uint256 _value,
+        address _token,
+        bytes memory
+    ) public {
+        require(IERC20(_token) == token, "Unsupported token");
+        token.safeTransferFrom(_from, address(this), _value);
     }
 
     /// @notice Withdraws tokens from escrow to selected beneficiaries,
--- a/solidity/contracts/PhasedEscrow.sol
+++ b/solidity/contracts/PhasedEscrow.sol
@@ -132,7 +132,7 @@ contract BatchedPhasedEscrow is Ownable {
 
     /// @notice Transfers the role of drawee to another address. Can be called
     ///         only by the contract owner.
-    function transferDraweeRole(address newDrawee) public onlyOwner {
+    function setDrawee(address newDrawee) public onlyOwner {
         require(newDrawee != address(0), "New drawee can not be zero address");
         emit DraweeRoleTransferred(drawee, newDrawee);
         drawee = newDrawee;
--- a/solidity/contracts/PhasedEscrow.sol
+++ b/solidity/contracts/PhasedEscrow.sol
@@ -126,7 +126,7 @@ contract BatchedPhasedEscrow is Ownable {
     ///         beneficiary of the escrow, `false` otherwise.
     function isBeneficiaryApproved(
         IBeneficiaryContract _beneficiary
-    ) external view returns (bool) {
+    ) public view returns (bool) {
         return approvedBeneficiaries[address(_beneficiary)];
     }
 
@@ -154,7 +154,7 @@ contract BatchedPhasedEscrow is Ownable {
         for (uint256 i = 0; i < beneficiaries.length; i++) {
             IBeneficiaryContract beneficiary = beneficiaries[i];
             require(
-                approvedBeneficiaries[address(beneficiary)],
+                isBeneficiaryApproved(beneficiary),
                 "Beneficiary not approved"
             );
             withdraw(beneficiary, amounts[i]);
--- a/solidity/contracts/PhasedEscrow.sol
+++ b/solidity/contracts/PhasedEscrow.sol
@@ -179,7 +179,7 @@ interface IStakingPoolRewards {
 /// @notice A beneficiary contract that can receive a withdrawal phase from a
 ///         PhasedEscrow contract. Immediately stakes the received tokens on a
 ///         designated IStakingPoolRewards contract.
-contract StakingPoolRewardsEscrowBeneficiary is Ownable {
+contract StakingPoolRewardsEscrowBeneficiary is Ownable, IBeneficiaryContract {
     IERC20 public token;
     IStakingPoolRewards public rewards;
 
--- a/solidity/contracts/PhasedEscrow.sol
+++ b/solidity/contracts/PhasedEscrow.sol
@@ -70,6 +70,103 @@ contract PhasedEscrow is Ownable {
     }
 }
 
+/// @title BatchedPhasedEscrow
+/// @notice A token holder contract allowing contract owner to approve a set of
+///         beneficiaries of tokens held by the contract, to appoint a separate
+///         drawee role, and allowing that drawee to withdraw tokens to approved
+///         beneficiaries in phases.
+contract BatchedPhasedEscrow is Ownable {
+    using SafeERC20 for IERC20;
+
+    event BeneficiaryApproved(address beneficiary);
+    event TokensWithdrawn(address beneficiary, uint256 amount);
+    event DraweeRoleTransferred(address oldDrawee, address newDrawee);
+
+    IERC20 public token;
+    address public drawee;
+    mapping(address => bool) private approvedBeneficiaries;
+
+    modifier onlyDrawee() {
+        require(drawee == msg.sender, "Caller is not the drawee");
+        _;
+    }
+
+    constructor(IERC20 _token) public {
+        token = _token;
+        drawee = msg.sender;
+    }
+
+    /// @notice Funds the escrow by transferring all of the approved tokens
+    ///         to the escrow.
+    function receiveApproval(
+        address _from,
+        uint256 _value,
+        address _token,
+        bytes memory
+    ) public {
+        require(IERC20(_token) == token, "Unsupported token");
+        token.safeTransferFrom(_from, address(this), _value);
+    }
+
+    /// @notice Approves the provided address as a beneficiary of tokens held by
+    ///         the escrow. Can be called only by escrow owner.
+    function approveBeneficiary(
+        IBeneficiaryContract _beneficiary
+    ) external onlyOwner {
+        address beneficiaryAddress = address(_beneficiary);
+        require(
+            beneficiaryAddress != address(0), 
+            "Beneficiary can not be zero address"
+        );
+        approvedBeneficiaries[beneficiaryAddress] = true;
+        emit BeneficiaryApproved(beneficiaryAddress);
+    }
+
+    /// @notice Returns `true` if the given address has been approved as a
+    ///         beneficiary of the escrow, `false` otherwise.
+    function isBeneficiaryApproved(
+        IBeneficiaryContract _beneficiary
+    ) external view returns (bool) {
+        return approvedBeneficiaries[address(_beneficiary)];
+    }
+
+    /// @notice Transfers the role of drawee to another address. Can be called
+    ///         only by the contract owner.
+    function transferDraweeRole(address newDrawee) public onlyOwner {
+        require(newDrawee != address(0), "New drawee can not be zero address");
+        emit DraweeRoleTransferred(drawee, newDrawee);
+        drawee = newDrawee;
+    }
+
+    /// @notice Withdraws tokens from escrow to selected beneficiaries,
+    ///         transferring to each beneficiary the amount of tokens specified
+    ///         as a parameter. Only beneficiaries previously approved by escrow
+    ///         owner can receive funds.
+    function batchedWithdraw(
+        IBeneficiaryContract[] memory beneficiaries,
+        uint256[] memory amounts
+    ) public onlyDrawee {
+        require(
+            beneficiaries.length == amounts.length,
+            "Mismatched arrays length"
+        );
+
+        for (uint256 i = 0; i < beneficiaries.length; i++) {
+            IBeneficiaryContract beneficiary = beneficiaries[i];
+            require(
+                approvedBeneficiaries[address(beneficiary)],
+                "Beneficiary not approved"
+            );
+            withdraw(beneficiary, amounts[i]);
+        }
+    }
+    function withdraw(IBeneficiaryContract beneficiary, uint256 amount) private  {
+        token.safeTransfer(address(beneficiary), amount);
+        emit TokensWithdrawn(address(beneficiary), amount);
+        beneficiary.__escrowSentTokens(amount);
+    }
+}
+
 interface ICurveRewards {
     function notifyRewardAmount(uint256 amount) external;
 }
--- a/solidity/contracts/PhasedEscrow.sol
+++ b/solidity/contracts/PhasedEscrow.sol
@@ -167,26 +167,30 @@ contract BatchedPhasedEscrow is Ownable {
     }
 }
 
-interface ICurveRewards {
-    function notifyRewardAmount(uint256 amount) external;
+// Interface representing staking pool rewards contract such as CurveRewards
+// contract deployed for Keep (0xAF379f0228ad0d46bB7B4f38f9dc9bCC1ad0360c) or
+// LPRewards contract from keep-ecdsa repository deployed for Uniswap.
+interface IStakingPoolRewards {
+   function notifyRewardAmount(uint256 amount) external; 
 }
 
-/// @title CurveRewardsEscrowBeneficiary
+
+/// @title StakingPoolRewardsEscrowBeneficiary
 /// @notice A beneficiary contract that can receive a withdrawal phase from a
 ///         PhasedEscrow contract. Immediately stakes the received tokens on a
-///         designated CurveRewards contract.
-contract CurveRewardsEscrowBeneficiary is Ownable {
+///         designated IStakingPoolRewards contract.
+contract StakingPoolRewardsEscrowBeneficiary is Ownable {
     IERC20 public token;
-    ICurveRewards public curveRewards;
+    IStakingPoolRewards public rewards;
 
-    constructor(IERC20 _token, ICurveRewards _curveRewards) public {
+    constructor(IERC20 _token, IStakingPoolRewards _rewards) public {
         token = _token;
-        curveRewards = _curveRewards;
+        rewards = _rewards;
     }
 
     function __escrowSentTokens(uint256 amount) external onlyOwner {
-        token.approve(address(curveRewards), amount);
-        curveRewards.notifyRewardAmount(amount);
+        token.approve(address(rewards), amount);
+        rewards.notifyRewardAmount(amount);
     }
 }
 
--- a/solidity/test/TestPhasedEscrow.js
+++ b/solidity/test/TestPhasedEscrow.js
@@ -776,6 +776,16 @@ describe("BatchedPhasedEscrow", () => {
       )
     })
 
+    it("reverts when there are not enough funds in the escrow", async () => {
+      await expectRevert.unspecified(
+        batchedPhasedEscrow.batchedWithdraw(
+          beneficiaries, 
+          [100, 200, 301], {
+          from: drawee,
+        })
+      )
+    })
+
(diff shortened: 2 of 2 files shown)


dimpar:
> I abstracted CurveRewardsEscrowBeneficiary contract into a generic StakingPoolRewardsEscrowBeneficiary contract.

Is this still relevant? I don't see `StakingPoolRewardsEscrowBeneficiary`

pdyraga:
> > I abstracted CurveRewardsEscrowBeneficiary contract into a generic StakingPoolRewardsEscrowBeneficiary contract.
> 
> Is this still relevant? I don't see `StakingPoolRewardsEscrowBeneficiary`

It's there, 25aa6bc4dbaaee940308b512d07b0e9159132d47, pull the lastest changes please :)

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
+    }
+
+    /// @notice Funds the escrow by transferring all of the approved tokens
+    ///         to the escrow.
+    function receiveApproval(
+        address _from,
+        uint256 _value,
+        address _token,
+        bytes memory
+    ) public {
```
dimpar, line 106: I think the order should be `constructor` -> `external` -> `public` -> `private`

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
+/// @notice A token holder contract allowing contract owner to approve a set of
+///         beneficiaries of tokens held by the contract, to appoint a separate
+///         drawee role, and allowing that drawee to withdraw tokens to approved
+///         beneficiaries in phases.
+contract BatchedPhasedEscrow is Ownable {
+    using SafeERC20 for IERC20;
+
+    event BeneficiaryApproved(address beneficiary);
+    event TokensWithdrawn(address beneficiary, uint256 amount);
+    event DraweeRoleTransferred(address oldDrawee, address newDrawee);
```
nkuba, line 83: Do we want to add some indexing?

```suggestion
    event BeneficiaryApproved(address indexed beneficiary);
    event TokensWithdrawn(address indexed beneficiary, uint256 amount);
    event DraweeRoleTransferred(address indexed oldDrawee, address indexed newDrawee);
```

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
+    ) public onlyDrawee {
+        require(
+            beneficiaries.length == amounts.length,
+            "Mismatched arrays length"
+        );
+
+        for (uint256 i = 0; i < beneficiaries.length; i++) {
+            IBeneficiaryContract beneficiary = beneficiaries[i];
+            require(
+                approvedBeneficiaries[address(beneficiary)],
```
nkuba, line 157: Why not use `isBeneficiaryApproved` function?

```suggestion
                isBeneficiaryApproved(beneficiary),
```

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
+    ) external view returns (bool) {
+        return approvedBeneficiaries[address(_beneficiary)];
+    }
+
+    /// @notice Transfers the role of drawee to another address. Can be called
+    ///         only by the contract owner.
+    function transferDraweeRole(address newDrawee) public onlyOwner {
+        require(newDrawee != address(0), "New drawee can not be zero address");
+        emit DraweeRoleTransferred(drawee, newDrawee);
+        drawee = newDrawee;
```
nkuba, line 138: We usually emit an event at the very end.

```suggestion
        drawee = newDrawee;
        emit DraweeRoleTransferred(drawee, newDrawee);
```

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
 }
 
-interface ICurveRewards {
-    function notifyRewardAmount(uint256 amount) external;
+/// @title BatchedPhasedEscrow
+/// @notice A token holder contract allowing contract owner to approve a set of
+///         beneficiaries of tokens held by the contract, to appoint a separate
+///         drawee role, and allowing that drawee to withdraw tokens to approved
+///         beneficiaries in phases.
+contract BatchedPhasedEscrow is Ownable {
```
nkuba, line 78: Should this contract implement `IBeneficiaryContract` in case we want to transfer funds between escrows?

nkuba, line 78: As we don't extend `PhasedEscrow` maybe we don't need to mention `Phased` escrow in the name? Would `BatchedEscrow` be fine?

Also maybe this contract should be placed in a separate file as we're not referring to `PhasedEscrow`?

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
+    ///         beneficiary of the escrow, `false` otherwise.
+    function isBeneficiaryApproved(
+        IBeneficiaryContract _beneficiary
+    ) external view returns (bool) {
+        return approvedBeneficiaries[address(_beneficiary)];
+    }
+
+    /// @notice Transfers the role of drawee to another address. Can be called
+    ///         only by the contract owner.
+    function transferDraweeRole(address newDrawee) public onlyOwner {
```
nkuba, line 135: Thoughts on renaming `transferDraweeRole` to `setDraweeRole`? Having in mind how `transferOwnership` from `Ownable` works, the current name suggest that this function transfers the role from the `msg.sender` to the `newDrawee`.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
+                approvedBeneficiaries[address(beneficiary)],
+                "Beneficiary not approved"
+            );
+            withdraw(beneficiary, amounts[i]);
+        }
+    }
+    function withdraw(IBeneficiaryContract beneficiary, uint256 amount) private  {
+        token.safeTransfer(address(beneficiary), amount);
+        emit TokensWithdrawn(address(beneficiary), amount);
+        beneficiary.__escrowSentTokens(amount);
```
nkuba, line 166: Let's emit an event at the end.
```suggestion
        beneficiary.__escrowSentTokens(amount);
        emit TokensWithdrawn(address(beneficiary), amount);
```

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
+}
+
+
+/// @title StakingPoolRewardsEscrowBeneficiary
 /// @notice A beneficiary contract that can receive a withdrawal phase from a
 ///         PhasedEscrow contract. Immediately stakes the received tokens on a
-///         designated CurveRewards contract.
-contract CurveRewardsEscrowBeneficiary is Ownable {
+///         designated IStakingPoolRewards contract.
+contract StakingPoolRewardsEscrowBeneficiary is Ownable {
```
nkuba, line 182: Can it implement `IBeneficiaryContract` interface?

solidity/test/TestPhasedEscrow.js
```diff
@@ -426,7 +425,378 @@ describe("PhasedEscrow", () => {
   }
 })
 
-describe("CurveRewardsEscrowBeneficiary", () => {
+describe("BatchedPhasedEscrow", () => {
```
nkuba, line 428: Could be worth extracting to a separate file.

nkuba (changes requested):

solidity/contracts/PhasedEscrow.sol
pdyraga: I am not sure, we'll approve 4 beneficiaries and transfer the drawee role one time. I'd be surprised if we had more than 10 those events emitted in total.

solidity/test/TestPhasedEscrow.js
```diff
@@ -426,7 +425,378 @@ describe("PhasedEscrow", () => {
...
+          amounts[i],
+          `Unexpected amount withdrawn for beneficiary ${i}`
+        )
+      }
+
+      expect(await token.balanceOf(batchedPhasedEscrow.address)).to.eq.BN(
+        0,
+        `Unexpected escrow balance`
+      )
+    })
```
nkuba, line 795: Can we add a test case where transaction reverts due to exceeding the available escrow balance?

nkuba (changes requested):

solidity/contracts/PhasedEscrow.sol
pdyraga: This will not work because `drawee` will already have a new value assigned. I am not sure if it makes sense to assign a temporary variable just to keep the event at the end, especially that after some updates based on Echidna test we no longer keep events at the end in some cases.

solidity/contracts/PhasedEscrow.sol
pdyraga: Yeah, good point. This line was written before `isBeneficiaryApproved` function. Will update.

solidity/contracts/PhasedEscrow.sol
pdyraga: I do not think so but let's validate what I say three times, please. When we'll be transferring from `PhasedEscrow` to `BatchedPhasedEscrow`, we'll set `StakerRewardsBeneficiary` on `PhasedEscrow` that will point to `BatchedPhasedEscrow` as a `_stakerRewards` beneficiary. This scenario is covered in `"can be funded from PhasedEscrow contract"` in unit tests. 

If we ever decide to transfer funds out of `BatchedPhasedEscrow`, we'll approve beneficiary using the standard mechanism... and that's it.

Did I miss something?

pdyraga: This contract is still a _phased_ escrow though. The reason why we introduced `PhasedEscrow` was that the standard `Escrow` allows transferring funds only in one shot. `PhasedEscrow` allows to transfer first X amount, then Y amount, then Z amount - that's the main distinction. And this contract is still a *phased* escrow.

solidity/contracts/PhasedEscrow.sol
pdyraga: Hmm... Yes, very good point. Will update.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
+contract BatchedPhasedEscrow is Ownable {
+    using SafeERC20 for IERC20;
+
+    event BeneficiaryApproved(address beneficiary);
+    event TokensWithdrawn(address beneficiary, uint256 amount);
+    event DraweeRoleTransferred(address oldDrawee, address newDrawee);
+
+    IERC20 public token;
+    address public drawee;
+    mapping(address => bool) private approvedBeneficiaries;
```
dimpar, line 87: do we need to make it private? If it is set to public, then I think we can remove external function `isBeneficiaryApproved()` and just call `approvedBeneficiaries(address)`

solidity/contracts/PhasedEscrow.sol
pdyraga: Actually, I think it should be just `setDrawee` - we are setting the property, that's all. We were transferring the role but now we just *set* who is a drawee. WDYT?

solidity/contracts/PhasedEscrow.sol
pdyraga: Oh, this is intentionally before `__escrowSentTokens`. I learned it from Echidna inspections and it would complain if we let it scan this code. Given that `__escrowSentTokens` is a call to another contract, the order of events might not be maintained in case we entered the recursion. Emitting the event before protects from that. Not a big deal, but that's also how it's done in `PhasedEscrow`, exactly for the same reason.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
+        require(
+            beneficiaries.length == amounts.length,
+            "Mismatched arrays length"
+        );
+
+        for (uint256 i = 0; i < beneficiaries.length; i++) {
+            IBeneficiaryContract beneficiary = beneficiaries[i];
+            require(
+                isBeneficiaryApproved(beneficiary),
+                "Beneficiary not approved"
```
dimpar, line 158: "Beneficiary not approved" -> "Beneficiary was not approved"?

solidity/contracts/PhasedEscrow.sol
pdyraga: Yes, good point. I will add it here.

solidity/test/TestPhasedEscrow.js
pdyraga: Yeah, I was struggling with it for a while. Though this is still a *phased* escrow, all beneficiaries, and all relationships (such as `PhasedEscrow` -> `BatchedPhasedEscrow`) are tested in this file. I'd prefer to keep it all together for now. WDYT?

solidity/contracts/PhasedEscrow.sol
pdyraga: Correct, good catch - I'll update it.

solidity/contracts/PhasedEscrow.sol
pdyraga: Yeah, I was thinking about it but `isBeneficiaryApproved(beneficiary)` reads better to me than `approvedBeneficiaries(address)` - that's why I decided to go with a separate function and making this mapping `private`.

solidity/contracts/PhasedEscrow.sol
nkuba: Yeah, you're right. So transfer between escrows requires a beneficiary contract deployed. I'm not sure why we don't do this transfer directly by implementing just one `__escrowSentTokens` function?

nkuba: Same for `LPRewards`, why don't we implement `__escrowSentTokens` there instead of implementing an intermediary beneficiary contract?

nkuba: It's confusing as this contract doesn't extend `PhasedEscrow`.

solidity/contracts/PhasedEscrow.sol
nkuba: YES! 💪🏻

solidity/test/TestPhasedEscrow.js
pdyraga: I was thinking about it but wasn't sure if that's the right thing to do. I do not want to add a custom `revert` because that would require adding a separate step to a loop summing up tokens withdrawn so far. The fact I can't spend more than I have should be enforced by ERC20 implementation and I do not want to duplicate this logic just to burn some gas. Given that, I can add a test but then I am testing ERC20 implementation: I can't spend more than I have.

solidity/contracts/PhasedEscrow.sol
pdyraga: Perfect. Done in https://github.com/keep-network/keep-core/pull/2230/commits/e63422335bbbac248d1ba6add95c473d540e6a36.

solidity/test/TestPhasedEscrow.js
```diff
@@ -426,7 +425,378 @@ describe("PhasedEscrow", () => {
...
+    let escrowBalance
+
+    beforeEach(async () => {
+      beneficiaries = [
+        beneficiary1.address,
+        beneficiary2.address,
+        beneficiary3.address,
+      ]
+
+      amounts = [100, 200, 300]
```
dimpar, line 698: Am I correct that we don't need a mechanism for cap a max amount of KEEP that a beneficiary can receive? Ie.  if a first beneficiary will withdraw 90 instead of 100, then a second beneficiary can withdraw 210.

solidity/contracts/PhasedEscrow.sol
pdyraga: I think the original intention was to do not leak any details about how do we handle transfers between escrow into the final recipient. In this design, a beneficiary with `__escrowSentTokens` performs the role similar to what adapters do in object-oriented design. I see a benefit in not leaking any of it into `LPRewards` though I absolutely do agree deploying a separate contract is quite unwieldy. Are you OK with leaving it as-is for now and reconsidering the design for the next escrow implementation (if we ever do that, hopefully, we'll not have to, heh)?

pdyraga: Depends on how you look at it - if you look from a perspective of inheritance then yes, I do agree if it's confusing. If you look from a perspective what those contracts offer, then I think it's not. `BatchedPhasedEscrow` explains what it's doing - allows withdrawing tokens in batches, and in phases. `PhasedEscrow` - allows withdrawing tokens in phases. If we named it just `BatchedEscrow`, the name would not reflect the full functionality.

solidity/contracts/PhasedEscrow.sol
dimpar: Ok

solidity/test/TestPhasedEscrow.js
nkuba: It's up to our contract how such situation is handled. We're using ERC20 impl for a single tansfer but our function performs multiple of them. We could handle revert of one transaction and let the others finish sucesfully. In the test we want to validate that if one transfer fails the whole call reverts.

solidity/test/TestPhasedEscrow.js
pdyraga: From the perspective of `BatchedPhasedEscrow`, we do not. In `withdraw` we do `token.safeTransfer(address(beneficiary), amount)` so we are not approving anything more than that. Individual beneficiary implementations do approve, but they approve a particular recipient address; for example from `__escrowSentTokens` of `StakingPoolRewardsEscrowBeneficiary`: `token.approve(address(rewards), amount)`

solidity/test/TestPhasedEscrow.js
```diff
@@ -426,7 +425,378 @@ describe("PhasedEscrow", () => {
...
+          from: owner,
+        }
+      )
+
+      expectEvent(receipt, "BeneficiaryApproved", {
+        beneficiary: beneficiary1.address,
+      })
+    })
+
+    it("maintains beneficiaries as non-approved by default", async () => {
```
dimpar, line 585: Idk, but it feels like we're testing the default Solidity values here, which for `bool` is always `false`

pdyraga: Yeah, could be. Though I was looking at `BatchedPhasedEscrow` as a black-box, without thinking how this mechanism is implemented. This could be, for example, a list of approved beneficiaries, with some implementation bug that would return `true` in case that list would be empty.

solidity/test/TestPhasedEscrow.js
dimpar: okay, makes sense.

solidity/contracts/PhasedEscrow.sol
nkuba: I don't fully agree, but if `BatchedPhasedEscrow` won't have a relation to `PhasedEscrow` we should move it to a separate file to reduce confusion. Perfectly to a dedicated directory where all escrow and beneficiary related stuff will be placed.

solidity/test/TestPhasedEscrow.js
pdyraga: Good point, I will add a test.

solidity/test/TestPhasedEscrow.js
dimpar: For sure it wouldn't hurt to have it. Okay, let's leave it

solidity/contracts/PhasedEscrow.sol
pdyraga: Can we do it in a separate PR? I started pulling it ou but that's a lot of work.

dimpar: > Yeah, you're right. So transfer between escrows requires a beneficiary contract deployed.

And also we would need to `setRewardDistribution(beneficiary)` if I understand the flow correctly.

solidity/test/TestPhasedEscrow.js
pdyraga: Added in the last commit, https://github.com/keep-network/keep-core/pull/2230/commits/686fbe387bc5e077839927265e2c32ad7c74214f.

pdyraga:
All comments addressed @dimpar @nkuba. The one about [extracting code to separate files](https://github.com/keep-network/keep-core/pull/2230#discussion_r546789480), I'd recommend implementing separately - it is a lot of work to make it right, and do not want to block here.

solidity/contracts/PhasedEscrow.sol
pdyraga: `setRewardDistribution` in this flow would be the escrow directly, not the beneficiary.

nkuba: Ok, let's create an issue or PR for that not to loose it.

pdyraga: I have already started the work locally on it, I should open PR tomorrow morning, no issue needed.

solidity/contracts/PhasedEscrow.sol
```diff
@@ -70,26 +70,127 @@ contract PhasedEscrow is Ownable {
...
+        for (uint256 i = 0; i < beneficiaries.length; i++) {
+            IBeneficiaryContract beneficiary = beneficiaries[i];
+            require(
+                isBeneficiaryApproved(beneficiary),
+                "Beneficiary was not approved"
+            );
+            withdraw(beneficiary, amounts[i]);
+        }
+    }
+    function withdraw(IBeneficiaryContract beneficiary, uint256 amount) private  {
```
nkuba, line 163:
```suggestion
    }
    
    function withdraw(IBeneficiaryContract beneficiary, uint256 amount) private  {
```

nkuba (approved):

Issue #61, opened 2018-03-27
Relationship between keys and bls.ID


Shadowfiend:
We've got a few public keys and keylike material swimming around, so let's clarify if they are related or derived or what:

 - `bls.ID` represents a member's identity in a BLS group. It is keylike.
 - A client is associated with a staking public key, which refers to the parent chain account that staked.
 - A client has a public key in libp2p.

`bls.ID` could conceivably be derived from either the staking public key or the libp2p public key. Do we want to do this? Or do we want BLS IDs to be per group rather than having one BLS ID for all groups per client? If group members can be derived in a stable order from chain data (see #58, but almost certainly this will be true), the BLS IDs can simply be indices from 1 to group size based on the staker's place in the member list for the group.

Can we derive or reuse the staking keys for libp2p communication? Do we want to (probably not, but want to record that)?

@mhluongo looking for your thoughts here to pull this conversation out from #56.

lionakhnazarov:
Closing as stale. Targets the legacy keep-core v1 (random beacon / KEEP staking) architecture, superseded by the Threshold v2 stack. Not actionable against current main.

PR #2242, opened 2020-12-28
Use ethereum and local signer from keep-common

pdyraga:
~~Depends on https://github.com/keep-network/keep-common/pull/61~~

`EthereumSigner` provides functions to sign a message and verify a
signature using Ethereum-specific signature format. It also provides
functions for the conversion of a public key to address.

So far, this code existed only in `keep-core` and could not be reused
from `keep-ecdsa`. In #61 the code has been extracted from `keep-core` 
with slight modifications that will allow it to be imported both from 
`keep-core` and `keep-ecdsa`.

The code is being extracted to allow us to decouple ethereum-specific
implementation functions we use in generic code of `keep-ecdsa` client,
especially in `node.go`.

In #61 also the `local` chain implementation of `keep-core`'s `localSigner`
has been extracted to `keep-common`. This way, we'll reuse the same
code from `keep-core` and `keep-ecdsa` instead of duplicating it.
Here, we remove the `keep-core` implementation and import the one from
`keep-ecdsa`.


pkg/chain/ethereum/ethereum.go
```diff
@@ -22,15 +22,19 @@ import (
...
+func (ec *ethereumChain) ThresholdRelay() relayChain.Interface {
 	return ec
 }
 
 func (ec *ethereumChain) GetKeys() (*operator.PrivateKey, *operator.PublicKey) {
 	return operator.EthereumKeyToOperatorKey(ec.accountKey)
 }
 
-func (ec *ethereumChain) GetConfig() *relaychain.Config {
+func (ec *ethereumChain) Signing() chain.Signing {
```
nkuba, line 33: Where are we using this function?

dimpar: Looks like we use it in `beacon.go` which implements an interface `Signing` from `chain.go`

dimpar:
Once 61 is merged, we'd need to update mod dependencies. I have no more comments here.

pkg/chain/ethereum/ethereum.go
pdyraga: It's fetched here:
https://github.com/keep-network/keep-core/blob/b6993e6dcf3bfb3c48a68e6e899954c4a220aeef/pkg/beacon/beacon.go#L52

and used later to generate relay entry.

There are also tests:
- `signing_test.go`
- `states_test.go`
- `membership_validator_test.go`
- `dkgtest.go`

pdyraga:
> Once 61 is merged, we'd need to update mod dependencies. I have no more comments here.

We can wait with updating dependencies for another PR. I was thinking about releasing new `keep-commons` version today or later this week.

pkg/chain/ethereum/ethereum.go
```diff
@@ -22,15 +22,19 @@ import (
...
+	return ethutil.NewSigner(ec.accountKey.PrivateKey)
+}
```
nkuba, line 35: At first I didn't like the naming that `NewSigner` returns `Signing` (but not `Signer`). But `ethutil.NewSigner` returns a struct that implements `chain.Signing`. I think it's fine.

nkuba: Nah, it should be `Signer` instead of `Signing`.

nkuba (approved):
I'm not sure about naming ([comment](https://github.com/keep-network/keep-core/pull/2242/files#r549675122)) but overall the PR looks fine.

PR #2244, opened 2020-12-28
Code linting

dimpar:
In this PR we are adding linting for Solidity contracts, their tests in JS and also JS script files to stay consistent across keep-network repositories.


.pre-commit-config.yaml
```diff
@@ -6,25 +6,19 @@ repos:
     - id: go-vet
     - id: go-lint
  - repo: https://github.com/keep-network/pre-commit-hooks.git
-   rev: 63e729f
+   rev: v1.3.0
```
nkuba, line 9: Don't we want to leave `63e729f` as this is newer than the `v1.3.0` tag?

solidity/.soliumrc.json
```diff
@@ -1,18 +1,16 @@
...
+      [
+        "timestamp"
+      ]
     ],
-    "indentation": [
-      "error",
-      4
-    ],
-    "security/no-block-members": ["off", ["timestamp"]],
+    "indentation": "off",
```
nkuba, line 13: Maybe we shouldn't turn off indentation checks?

```suggestion
```

solidity/.soliumrc.json
```diff
@@ -1,18 +1,16 @@
 {
-  "extends": "solium:recommended",
+  "extends": "keep",
```
nkuba, line 2: There are some rules that we use repetitievely across our repos, maybe we should review them and pull them into `solium-config-keep`. But this is not for the scope of this PR.

nkuba (commented):
First comments.

.pre-commit-config.yaml
dimpar: Yes, you're right. Reverted here 79153f6

solidity/.soliumrc.json
dimpar: This one actually behaves very strange. Initially it was not there, only when I copied the same config from `keep-ecdsa` I noticed why it might be turned off. When you turn it `on` and `npm run lint:sol` it will see that there are `.sol` files that need indentation changed. However, when you fix them manually (lint:fix:sol won't fix it) and run `npm run lint:fix:sol` again (ex. you have other changes that you want to fix) then it will revert you manual changes and running 'npm run lint:sol' will mark them as errors. So long story short.. `npm run lint:sol` indicates indentation errors made by `npm run lint:fix:sol`.

nkuba: Oh yeah, [I had the same problem in `keep-ecdsa`](https://github.com/keep-network/keep-ecdsa/commit/0befc5c9f84472c195a200cd13ced98f2d1d0ca6). We can turn off indentation check for solium as we have prettier running after solium and prettier will handle it.

solidity/package.json
```diff
@@ -18,9 +18,14 @@
     "test:quick": "mocha --exit --recursive --timeout 45000",
     "test:quick:watch": "mocha --watch --exit --recursive --timeout 45000",
     "demo": "truffle migrate --reset && truffle exec ./scripts/delegate-tokens.js",
-    "lint": "solium --dir ./contracts",
+    "lint": "npm run lint:js && npm run lint:sol",
     "lint:js": "eslint ${npm_package_config_eslintPaths}",
-    "lint:fix:js": "eslint --fix ${npm_package_config_eslintPaths}"
+    "lint:fix:js": "eslint --fix ${npm_package_config_eslintPaths}",
+    "lint:sol": "solium -d contracts/ && prettier --list-different contracts/ **/*.sol",
+    "lint:fix:sol": "solium -d contracts/ --fix && prettier --write contracts/ **/*.sol"
```
nkuba, line 25: Why we added `contracts/` to prettier? Isn't `prettier --list-different **/*.sol` and `prettier --write **/*.sol` enough?

dimpar: It didn't go to the nested `.sol` contracts under the `/contracts` directory. Ex. `/contracts/libraries`,  `/contracts/stubs` etc.

nkuba: Is this space here on purpose?

```suggestion
    "lint:sol": "solium -d contracts/ && prettier --list-different contracts/**/*.sol",
    "lint:fix:sol": "solium -d contracts/ --fix && prettier --write contracts/**/*.sol"
```

dimpar: yes, otherwise it didn't work for me

nkuba: Try using quotes around the path as in [an example in docs](https://prettier.io/docs/en/cli.html#--list-different).

```suggestion
    "lint:sol": "solium -d contracts/ && prettier --list-different '**/*.sol'",
    "lint:fix:sol": "solium -d contracts/ --fix && prettier --write '**/*.sol'"
```

solidity/package.json
```diff
@@ -18,9 +18,14 @@
     "test:quick": "mocha --exit --recursive --timeout 45000",
     "test:quick:watch": "mocha --watch --exit --recursive --timeout 45000",
     "demo": "truffle migrate --reset && truffle exec ./scripts/delegate-tokens.js",
-    "lint": "solium --dir ./contracts",
+    "lint": "npm run lint:js && npm run lint:sol",
     "lint:js": "eslint ${npm_package_config_eslintPaths}",
-    "lint:fix:js": "eslint --fix ${npm_package_config_eslintPaths}"
+    "lint:fix:js": "eslint --fix ${npm_package_config_eslintPaths}",
+    "lint:sol": "solium -d contracts/ && prettier --list-different contracts/ **/*.sol",
```
nkuba, line 24: Meybe we could use [`--check` instead `--list-different`](https://prettier.io/docs/en/cli.html#--check) so we have human readable output? We may need to bump up the version of prettier for that.
```suggestion
    "lint:sol": "solium -d contracts/ && prettier --check contracts/ **/*.sol",
```

solidity/package.json
```diff
@@ -18,9 +18,14 @@
     "test:quick": "mocha --exit --recursive --timeout 45000",
     "test:quick:watch": "mocha --watch --exit --recursive --timeout 45000",
     "demo": "truffle migrate --reset && truffle exec ./scripts/delegate-tokens.js",
-    "lint": "solium --dir ./contracts",
+    "lint": "npm run lint:js && npm run lint:sol",
     "lint:js": "eslint ${npm_package_config_eslintPaths}",
```
nkuba, line 22: Maybe we could simplify the configuration by removing `config.eslintPaths` files in all folders exept `dashboard`?
```suggestion
    "lint:js": "eslint --ignore-pattern 'dashboard/' .",
```

.pre-commit-config.yaml
```diff
@@ -11,20 +11,14 @@ repos:
...
-      - ethlint@1.2.3
+    - id: lint-sol
+      name: 'lint solidity sol'
+      entry: /usr/bin/env bash -c "cd solidity && npm run lint:sol"
+      language: script
+      description: "Checks solidity code according to the package's linter configuration"
     - id: lint-js
       name: 'lint solidity js'
       entry: /usr/bin/env bash -c "cd solidity && npm run lint:js"
-      files: 'solidity\/.*\/\.js$'
```
nkuba, line 27: We should probably leave `files` property configured for both `lint-sol` and `lint-js`. It would execute linting only if there are changes to files matching a pattern.

solidity/package.json
dimpar: Not sure if we can, it didn't work when I tried. According to the [docs](https://eslint.org/docs/user-guide/command-line-interface) we need to provide file or dir to run `eslint`. Did you try and it work for you?

.pre-commit-config.yaml
dimpar: According to the [docs](https://pre-commit.com/) it looks like it is an optional property. I've tried committing both `.js` & `.sol` with bad linting and in both cases it stopped me informing about it. It will just execute whatever command we have in `entry` property.

solidity/package.json
dimpar: Yup, quotes worked! 68db3a8

solidity/package.json
dimpar: Changed in 37696c0 but I don't think it helped much. Here's the diff:

--check
```
Checking formatting...
[warn] contracts/libraries/staking/TopUps.sol
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @keep-network/keep-core@1.7.0-pre.0 lint:sol: `solium -d contracts/ && prettier --check '**/*.sol'`
npm ERR! Exit status 1
```

--list-different
```
contracts/libraries/staking/TopUps.sol
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @keep-network/keep-core@1.7.0-pre.0 lint:sol: `solium -d contracts/ && prettier --list-different  '**/*.sol'`
npm ERR! Exit status 1
```

.pre-commit-config.yaml
nkuba: This property determines what files should be included in the check. One of the implications is that if the file doesn't match the pattern the check will be skipped. In our case, this is helpful as we don't want to run contracts linting verification when we're committing changes to the Go client, any other directory other than `solidity`, or to files that don't match extension that linting checks.

solidity/package.json
nkuba: We can do `eslint .` and it will check recursively all the files. If we want to skip some directories or files we can define them in a `.eslintignore` file or pass with `--ignore-pattern` as in the example above.

solidity/package.json
nkuba: Maybe it depends on a file type, but for some types `--check` will print a neat output.

solidity/package.json
dimpar: '2a87ff5' Adding a dot right after eslint `eslint .` did the trick!

.pre-commit-config.yaml
dimpar: `43873c4` ok, makes sense. But I think there was an error in that regex, cause it didn't check the JavaScript files when I tried to commit it. Only when I slightly changed it, it started to check `.js` files upon committing them.

solidity/.soliumrc.json
dimpar: https://github.com/keep-network/keep-core/issues/2303

nkuba (dismissed):

nkuba (approved):

nkuba (approved):

pdyraga (approved):

PR #2245, opened 2020-12-28
tBTC rewards allocation 2020-12-18 -> 2020-12-25

pdyraga:
Adding tBTC reward allocation computed in https://github.com/keep-network/keep-ecdsa/pull/652 to KEEP Token Dashboard.


r-czajkowski (approved):
LGTM 👍 tested data fetching locally against mainnet contracts, works as expected.

PR #2247, opened 2020-12-28
Liquidity rewards empty pool

michalsmiarowski:
Liquidity rewards empty pool ui changes

Issues changed/fixed by this PR:
- gray out liquidity info when the % of the pool is equal to 0
- change text of the button "add more lp tokens" to "deposit lp tokens" when the % of the pool is equal to 0
- enable 'withdraw all' button to allow the user to withdraw the lp tokens when there are no rewards yet

image

### DIFF
--- a/solidity/dashboard/src/components/LiquidityRewardCard.jsx
+++ b/solidity/dashboard/src/components/LiquidityRewardCard.jsx
@@ -67,7 +67,11 @@ const LiquidityRewardCard = ({
           View pool
         </a>
       </h4>
-      <div className={"liquidity__info text-grey-60 mt-2"}>
+      <div
+        className={`${
+          gt(lpBalance, 0) ? "liquidity__info" : "liquidity__info-locked"
+        } mt-2`}
+      >
         <div className={"liquidity__info-tile bg-mint-10"}>
           <h2 className={"liquidity__info-tile__title text-mint-100"}>
             {formattedApy}*
@@ -115,7 +119,7 @@ const LiquidityRewardCard = ({
           )
         }
       >
-        add more lp tokens
+        {gt(lpBalance, 0) ? "add more lp tokens" : "deposit lp tokens"}
       </SubmitButton>
 
       <SubmitButton
--- a/solidity/dashboard/src/components/LiquidityRewardCard.jsx
+++ b/solidity/dashboard/src/components/LiquidityRewardCard.jsx
@@ -68,9 +68,7 @@ const LiquidityRewardCard = ({
         </a>
       </h4>
       <div
-        className={`${
-          gt(lpBalance, 0) ? "liquidity__info" : "liquidity__info-locked"
-        } mt-2`}
+        className={`liquidity__info${gt(lpBalance, 0) ? "" : "--locked"} mt-2`}
       >
         <div className={"liquidity__info-tile bg-mint-10"}>
           <h2 className={"liquidity__info-tile__title text-mint-100"}>
--- a/solidity/dashboard/src/components/LiquidityRewardCard.jsx
+++ b/solidity/dashboard/src/components/LiquidityRewardCard.jsx
@@ -78,7 +78,7 @@ const LiquidityRewardCard = ({
         </div>
         <div className={"liquidity__info-tile bg-mint-10"}>
           {isFetching ? (
-            <Skeleton tag="h2" shining color="mint-20" />
+            <Skeleton tag="h2" shining color="color-grey-60" />
           ) : (
             <h2
               className={"liquidity__info-tile__title text-mint-100"}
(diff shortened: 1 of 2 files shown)


solidity/dashboard/src/components/LiquidityRewardCard.jsx
```diff
@@ -67,7 +67,11 @@ const LiquidityRewardCard = ({
           View pool
         </a>
       </h4>
-      <div className={"liquidity__info text-grey-60 mt-2"}>
+      <div
+        className={`${
+          gt(lpBalance, 0) ? "liquidity__info" : "liquidity__info-locked"
```
r-czajkowski, line 72: And let's simplify a bit. ```className={`liquidity__info${gt(lpBalance, 0) ? "" : "--locked"} mt-2`}```

solidity/dashboard/src/components/LiquidityRewardCard.jsx
```diff
@@ -115,19 +119,19 @@ const LiquidityRewardCard = ({
           )
         }
       >
-        add more lp tokens
+        {gt(lpBalance, 0) ? "add more lp tokens" : "deposit lp tokens"}
```
r-czajkowski, line 122: I believe it should be `gt(wrappedTokenBalance)` because we stake the wrapped token in the `LPRewards` contract.

r-czajkowski: Also, I think if the user doesn't have the wrapped token we should display a link `<Link to="<uniswapLink">deposit lp tokens</Link>` which redirects to the uniswap app.

r-czajkowski (changes requested):

solidity/dashboard/src/components/LiquidityRewardCard.jsx
r-czajkowski: ^ so probably should be:

```
gt(wrappedTokenBalance, 0) ? 

<SubmitButton
 className={`liquidity__add-more-tokens btn btn-primary btn-lg w-100`}
 onSubmitAction={(awaitingPromise) =>
  addLpTokens(
   wrappedTokenBalance,
   liquidityPairContractName,
   awaitingPromise
  )
 }
>
 add more lp tokens"
</SubmitButton>
:
<Link to="<uniswapLink" className="btn btn-primary btn-lg w-100>deposit lp tokens</Link>
```

r-czajkowski: ah sorry misunderstood Figma views. Everything is ok! pls ignore the comments above.

We want to display `deposit lp tokens` if the current logged user doesn't have any tokens in the LPRewards pool, so changes looking good. Sorry @michalsmiarowski.

michalsmiarowski:
@r-czajkowski I resolved all the issues you mentioned so this PR is ready for round 2! 🥊

solidity/dashboard/src/css/liquidity-page.less
```diff
@@ -29,7 +29,7 @@
     margin: 0.3rem 0;
   }
 
-  .liquidity__info {
+  .liquidity__info, .liquidity__info--locked {
```
r-czajkowski, line 32: What do you think about adding some transition effect to the `liquidity__info-tile` eg. `transition: background-color 0.5 ease`? On mainnet fetching data may take a little longer so the user will see the grey color definitely longer and a smooth transition from grey to mint could improve UX.

solidity/dashboard/src/css/liquidity-page.less
```diff
@@ -46,6 +46,15 @@
     }
   }
 
+  .liquidity__info--locked {
+    .liquidity__info-tile {
+      &:extend(.bg-grey-20);
+      .liquidity__info-tile__title {
+        color: @color-grey-60!important;
```
r-czajkowski, line 53: [Loading skeletons](https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/components/LiquidityRewardCard.jsx#L79) in `mint-20` color on the `grey-60` backround doens't look good. We should change the color of the loading skeleton.

r-czajkowski (changes requested):
Left 2 comments about UX. Overall looking very good 👍.

solidity/dashboard/src/css/liquidity-page.less
michalsmiarowski: I just noticed that too and i definitely agree. I will change that

solidity/dashboard/src/css/liquidity-page.less
michalsmiarowski: That's actually a great idea! I wouldn't have thought it up better myself! 🔥

michalsmiarowski:
The issues are fixed! Ready for another review!

r-czajkowski (approved):
looking good :shipit:

Issue #2243, opened 2020-12-28
Round percentage value to two decimal places in a circular progress bar


r-czajkowski:
Bug:
![obraz](https://user-images.githubusercontent.com/57687279/103204065-c8072c00-48f6-11eb-989e-c7cac6dfde32.png)

liz-shinn:
Yup was also going to post about this one; seeing it here:

<img width="298" alt="Screen Shot 2020-12-28 at 3 29 58 PM" src="https://user-images.githubusercontent.com/57226633/103241567-9e172f00-4921-11eb-85aa-2610fdbb5740.png">

PR #2251, opened 2020-12-29
Set two decimal places for the value of the progress bar

michalsmiarowski:
Closes: #2243

Set two decimal places for the percentage value of the circular progress bar

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>


r-czajkowski (approved):
:shipit:

PR #2270, opened 2021-01-05
Add SECURITY.md with vulnerability disclosure policy

Shadowfiend:
This outlines both the versions that are supported for vulnerabilities at this
time, as well as the requested process for disclosing vulnerabilities to the
team.

Goes hand-in-hand with keep-network/website#325.

### DIFF
--- a/SECURITY.adoc
+++ b/SECURITY.adoc
@@ -34,9 +34,10 @@ and include:
 - If desired, a suggestion on how to fix the issue.
 - If reporting a sensitive vulnerability in an encrypted message, please
   include a key for encrypting replies, or a secure medium to contact you.
-  Note that for simplicity, the Keep team prefers email, but can work with
-  Signal and Telegram on a case-by-case basis. Please always file an initial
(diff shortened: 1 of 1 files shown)


mhluongo (changes requested):
We're pushing Keybase in the security.txt, but don't mention it here and prefer encrypted email.. let's be consistent

Shadowfiend:
Great point, will flag Keybase as our preferred real-time mechanism for now.

Shadowfiend:
Great point, will flag Keybase as our preferred real-time mechanism for now. Note that point of first contact is still going to be encrypted email, and will remain our preferred approach for actual comms, since it allows a single address to lead to a potentially-changing set of responsible parties on the back end---something we can't do well and confidentially on any real-time platform. Note also that security.txt points to security@ as the point of contact, and uses Keybase solely as a PGP key lookup strategy.

Shadowfiend:
Updated and marked ready for review.

mhluongo (approved):
🎇

PR #2271, opened 2021-01-07
State machine performance improvements

lukasz-zimnoch:
Some POC work trying to improve state machine message handling.


lukasz-zimnoch:
There are no reliable proofs confirming those changes really give any benefits. Because they introduce some additional complexity and hence risks, we're closing this PR.

PR #2272, opened 2021-01-07
Add Liquidity Rewards notifications

michalsmiarowski:
This PR adds two new notifications:

1. "Your wallet has LP Tokens"

image

- the notification show each time current wallet gets the liquidity tokens
- it also show info about which pool those tokens are related to

2. "You've earned rewards!"

image

- we are checking every x minutes if the user's liquidity rewards balance is greater than the liquidity rewards balance x minutes ago
- x is 7 minutes (5 seemed too short to not get users annoyed and 10 seemed too long)


solidity/dashboard/src/sagas/liquidity-rewards.js
```diff
@@ -82,6 +94,62 @@ export function* watchFetchLiquidityRewardsData() {
...
+        showMessage({
+          type: messageType.REWARD,
+          title: `[${liquidityRewardPairName}] You've earned rewards!`,
+          sticky: true,
+          content: <ViewYourLiquidityBalance />,
+          classes: {
+            bannerDescription: "m-0",
+          },
+        })
+      )
```
michalsmiarowski, line 139: I was thinking that after showing the message we could call `fetchAllLiquidtyRewardsData()` so if the notification shows to the user while he is on Liquidity Page then he doesn't have to refresh to see the updated rewards. I'm not sure if it will be a good user experience though if he's on a Liquidity Page and suddenly the data starts fetching again. Thoughts?

solidity/dashboard/src/sagas/subscriptions.js
```diff
@@ -759,6 +767,27 @@ function* observeWrappedTokenMintAndBurnTx(liquidityRewardPair) {
       if (from === ZERO_ADDRESS || to === ZERO_ADDRESS) {
         yield* updateAPY(LPRewardsContract, liquidityRewardPair.name)
       }
+
+      // If the 'to' address is equal to the address of the connected wallet
+      // then it means that LP tokens are transferred to this address so we are
+      // displaying the proper notification
+      if (to === defaultAccount) {
```
r-czajkowski, line 774: Let's use [`isSameEthAddress`](https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/utils/general.utils.js#L50)

solidity/dashboard/src/sagas/subscriptions.js
```diff
@@ -759,6 +767,27 @@ function* observeWrappedTokenMintAndBurnTx(liquidityRewardPair) {
...
+        yield put(
```
r-czajkowski, line 775: I think should we also update the [`wrappedTokenBalance`](https://github.com/keep-network/keep-core/blob/master/solidity/dashboard/src/reducers/liquidity-rewards.js#L8) for a given liquidity rewards pair.

solidity/dashboard/src/App.js
```diff
@@ -26,6 +41,31 @@ const App = () => (
...
+      liquidityRewardNotificationFunc(yourAddress, false)
+      const liquidityRewardsNotificationInterval = setInterval(() => {
+        liquidityRewardNotificationFunc(yourAddress)
+      }, 420000) // every 7 minutes
+      return () => {
+        clearInterval(liquidityRewardsNotificationInterval)
+      }
+    }
+  }, [yourAddress, provider])
+
```
r-czajkowski, line 68: We should handle it on the redux side. We can something like:

```
while(true){
 delay(420000)
 // do the rest of logic
}
```

solidity/dashboard/src/sagas/liquidity-rewards.js
```diff
@@ -82,6 +94,62 @@ export function* watchFetchLiquidityRewardsData() {
...
+  // show the notification if the rewardBalance from LPRewardsContract is greater
+  // than the reward amount that was last time the notification was displayed
+  if (gt(currentReward, lastNotificationRewardAmount)) {
+    if (displayMessage) {
+      yield put(
+        showMessage({
+          type: messageType.REWARD,
+          title: `[${liquidityRewardPairName}] You've earned rewards!`,
+          sticky: true,
+          content: <ViewYourLiquidityBalance />,
```
r-czajkowski, line 134: I believe the redux shouldn't know about UI. We should handle displaying the specific message on the UI side and we can use `type` field for this.  An example of how we can do this -> https://stackoverflow.com/questions/35623656/how-can-i-display-a-modal-dialog-in-redux-that-performs-asynchronous-actions

r-czajkowski (changes requested):

solidity/dashboard/src/sagas/liquidity-rewards.js
r-czajkowski: We can dispatch acion like `liquidity_rewards/${liquidityRewardPair.name}_reward_updated` and pass the `currentReward` value. Then we can update current reward value via reducer.

r-czajkowski:
Also, let's sync with the current `master` branch.

michalsmiarowski:
@r-czajkowski Ready for another round of code review

solidity/dashboard/src/sagas/liquidity-rewards.js
michalsmiarowski: I'm not sure if you wanted me to do it exactly like in the first post from your StackOverlow link, but i just moved the `content` variable to the `messageType` constant in `Message.jsx` - that way we are displaying specific content component based on the type, like you proposed.

solidity/dashboard/src/components/ViewYourLiquidityBalance.jsx
```diff
@@ -0,0 +1,7 @@
+import React from "react"
+
+const ViewYourLiquidityBalance = () => {
```
r-czajkowski, line 3: Maybe let's create a dir `components/messages` where we can put all `Messages` component and then export them all from the `index.js` file? It will keep components by domain. What do you think?

solidity/dashboard/src/reducers/liquidity-rewards.js
```diff
@@ -7,6 +7,7 @@ const liquidityPairInitialData = {
   reward: 0,
   wrappedTokenBalance: 0,
   lpBalance: 0,
+  lastNotificationRewardAmount: "0",
```
r-czajkowski, line 10: Is that necessary? Maybe we can use the `reward` field?

solidity/dashboard/src/reducers/liquidity-rewards.js
```diff
@@ -143,6 +152,23 @@ const liquidityRewardsReducer = (state = initialState, action) => {
...
       }
+    case `liquidity_rewards/${liquidityRewardPairName}_reward_updated`:
+      return {
+        ...state,
+        [liquidityRewardPairName]: {
+          ...state[liquidityRewardPairName],
+          reward: restPayload.reward,
+        },
+      }
+    case `liquidity_rewards/${liquidityRewardPairName}_last_notification_reward_amount_updated`:
```
r-czajkowski, line 163: Same as above.

solidity/dashboard/src/sagas/subscriptions.js
```diff
@@ -809,3 +857,26 @@ export function* subscribeToLiquidityRewardsEvents() {
...
+        type: `liquidity_rewards/${pairName}_liquidity_rewards_earned_notification`,
+        payload: {
+          liquidityRewardPairName: pairName,
+          address: defaultAccount,
+          displayMessage,
+        },
+      })
+    }
+    displayMessage = true
+    yield delay(420000) // every 7 minutes
```
r-czajkowski, line 880: `moment.duration(7, "minutes").asMilliseconds()` ?

solidity/dashboard/src/sagas/subscriptions.js
```diff
@@ -809,3 +857,26 @@ export function* subscribeToLiquidityRewardsEvents() {
...
+export function* watchLiquidityRewardNotifications() {
+  const {
+    eth: { defaultAccount },
+  } = yield getWeb3Context()
+
+  // for the first iteration update the lastNotificationRewardAmount variable in redux without showing message
+  let displayMessage = false
+  while (true) {
+    for (const pairName of Object.keys(LIQUIDITY_REWARD_PAIRS)) {
+      yield put({
```
r-czajkowski, line 870: I think we can use `yield fork(processLiquidityRewardEarnedNotification, action)` so we can get rid of an additional `watchLiquidityRewardsEarnedNotification` watcher in `sagas/liquidity-rewards`.

solidity/dashboard/src/sagas/liquidity-rewards.js
```diff
@@ -89,6 +99,66 @@ export function* watchFetchLiquidityRewardsData() {
...
+    liquidityRewards[liquidityRewardPairName].lastNotificationRewardAmount
+  )
+  const currentReward = yield call(
+    [LiquidityRewards, LiquidityRewards.rewardBalance],
+    action.payload.address
+  )
+  // show the notification if the rewardBalance from LPRewardsContract is greater
+  // than the reward amount that was last time the notification was displayed
+  if (gt(currentReward, lastNotificationRewardAmount)) {
+    if (displayMessage) {
```
r-czajkowski, line 130: I think we can ` if (gt(currentReward, lastNotificationRewardAmount) && displayMessage)`

solidity/dashboard/src/sagas/subscriptions.js
```diff
@@ -809,3 +857,26 @@ export function* subscribeToLiquidityRewardsEvents() {
     )
   }
 }
+
+export function* watchLiquidityRewardNotifications() {
+  const {
+    eth: { defaultAccount },
+  } = yield getWeb3Context()
+
+  // for the first iteration update the lastNotificationRewardAmount variable in redux without showing message
```
r-czajkowski, line 866: Hmm, maybe we should display at the first iteration? I can imagine a situation when a user first goes to the other page than `liquidity rewards` after connecting to the wallet and we should let the user know that have available rewards.

r-czajkowski (changes requested):

solidity/dashboard/src/reducers/liquidity-rewards.js
r-czajkowski: ahh I see https://github.com/keep-network/keep-core/pull/2272/files#diff-84df92fe28afc557a6f217094ff02bda2900ddadeac35d401f09d2a94b9e0f07R129

michalsmiarowski: Yea, so just to clarify we need to save that value as `lastNotificationRewardAmount` so after seven minutes we can compare it with the `reward` field and display the notification only when `reward` is higher. 

It will also prevent the notification to display when for example user get some rewards 3 minutes after last notification check and then withdraw all rewards. In this case we won't show the notification since the reward balance is now 0 and is lower than `lastNotificationRewardAmount` which was checked 7 minutes ago. We save the new value (which is now 0) in the variable for future comparisons.

solidity/dashboard/src/sagas/subscriptions.js
michalsmiarowski: Hmm i'm not sure how to do that, because when a user connects to the wallet for the first time the `lastNotificationRewardAmount` is equal to 0. So right now we get the current `reward` value and save it to the `lastNotificationRewardAmount`, becuase other way it will just show the notification when the reward value is greater than 0 (for the first iteration only of course). To make it work properly with on the first iteration we would have to have `lastNotificationRewardAmount` value from user's last login, and i'm not really sure how we can do that 🤔 Any idea if we can resolve this somehow?

Unless you just meant to inform the user that he has any rewards (even if it's the same amount since he last logged in). In that case we can do that and when the user connects to the wallet the notifications will be displayed when the reward amount is greater than 0.

michalsmiarowski:
Ready for another round of review

solidity/dashboard/src/components/messages/LiquidityRewardsEarnedMessage.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+const LiquidityRewardsEarnedMessage = ({
+  liquidityRewardPairName,
+  sticky,
+  messageId,
+  onMessageClose,
+}) => {
+  const icon = Icons.Rewards
+  const title = `[${liquidityRewardPairName}] You've earned rewards!`
+  const content = <a href={"/liquidity"}>View your balance</a>
```
r-czajkowski, line 13: Let's use `Link` from `react-router-dom`. Currently, if click the link, the dapp refreshes and disconnects from the wallet.

r-czajkowski (changes requested):

solidity/dashboard/src/components/messages/LPTokensInWalletMessage.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+const LPTokensInWalletMessage = ({
+  liquidityRewardPairName,
+  sticky,
+  messageId,
+  onMessageClose,
+}) => {
+  const icon = Icons.Wallet
+  const title = `[${liquidityRewardPairName}] Your wallet has LP Tokens!`
+  const content = <a href={"/liquidity"}>Deposit them and earn rewards</a>
```
r-czajkowski, line 13: same as above

r-czajkowski (changes requested):

r-czajkowski (approved):
:shipit:

PR #2282, opened 2021-01-12
Separated dashboard build for keep-test

nkuba:
We can't build dashboard right after keep-core migration as we also have dependencies for keep-ecdsa and tbtc. So we first need to migrate all three projects and only after that can execute dashboard building.

The flow would look like:
- approve `build-test-migrate-publish-keep-test` execution
- kick off migrations and package publication for dependencies: `keep-ecdsa` and `tbtc` projects
- approve `build-publish-dashboard-keep-test`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -229,9 +229,10 @@ jobs:
           name: Resolve latest contracts
           working_directory: ~/project/solidity/dashboard
           command: |
-            npm update @keep-network/keep-core
-            npm update @keep-network/keep-ecdsa
-            npm update @keep-network/tbtc
+            npm update \
+              @keep-network/keep-core \
+              @keep-network/keep-ecdsa \
+              @keep-network/tbtc
       - run:
           name: Run Docker build
           working_directory: ~/project/solidity/dashboard
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -211,7 +211,7 @@ jobs:
     executor: docker-node
     steps:
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -229,7 +229,9 @@ jobs:
           name: Resolve latest contracts
           working_directory: ~/project/solidity/dashboard
           command: |
-            npm upgrade @keep-network/keep-core
+            npm update @keep-network/keep-core
+            npm update @keep-network/keep-ecdsa
+            npm update @keep-network/tbtc
```
Shadowfiend, line 234: You can run one `npm update` with multiple packages. Also, maybe use `--no-save` to avoid changing `package.json`---it's not really an issue, but I think it's conceptually cleaner (see [the docs](https://docs.npmjs.com/cli/v6/commands/npm-update#description) for more).

nkuba: Yeah, I tried with one `npm update` but I was confused with updates that happened to `package.json`. Maybe I just needed `--no-save`. Thanks for the hint!

nkuba: Actually there is one flaw of `--no-save`. It not only doesn't update `package.json` but also leaves `package-lock.json` unchanged. Taking into consideration that we're using `npm ci` that needs an updated `package-lock.json` I won't add `--no-save`.

Shadowfiend: Some software decisions really annoy me; this is one of them :laughing:

lukasz-zimnoch (approved):

PR #2285, opened 2021-01-13
Monitoring and Alerting Guide: Added initial monitoring and alerting guide

Shadowfiend:
The guide includes both metrics to monitor as well as recommended thresholds to
notify or alert on anomalies.


pdyraga (approved):
Should we add something in `keep-ecdsa` repository that would refer to this document?

pdyraga (approved):

PR #2286, opened 2021-01-13
Bump `@ledgerhq/hw-app-eth` dependency to 5.38

r-czajkowski:
One of the users reported a `0x6985 ledger error` which may occur in a random transaction. I was not able to reproduce this bug against ropsten and local network. Following the [btchip user recomendation](https://www.reddit.com/r/ledgerwallet/comments/j8bfng/when_trying_to_mint_a_nft_on_rarible_i_keep/g8aotp7\?utm_source\=share\&utm_medium\=web2x\&context\=3), we updated here `@ledgerhq/hw-app-eth` dependency to the `5.38`  to avoid this error in the future.


solidity/dashboard/src/connectors/utils.js
```diff
@@ -14,11 +14,18 @@ export const getEthereumTxObj = (txData, chainId) => {
...
 }
 
-// EIP-155 https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md
-// v = CHAIN_ID * 2 + 35 => CHAIN_ID = (v - 35) / 2
+/**
+ * This function returns the chain id from `v` param based on to the EIP-155[1]
+ * [1]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md
+ * Formula: v = CHAIN_ID * 2 + 35 => CHAIN_ID = (v - 35) / 2
+ *
+ * @param {Uint8Array} vInHex v param in hex.
```
michalsmiarowski, line 22: I would personally add hyphen between the name of the parameter and its description. As stated in [JSDoc](https://jsdoc.app/tags-param.html#names-types-and-descriptions) the hyphen is not necessary but it makes it more readable in my opinion. I'll leave it up to you to decide.

michalsmiarowski (dismissed):

solidity/dashboard/src/connectors/utils.js
Shadowfiend: I'm actually puzzled, is this a uint8array or a string (which is what "in hex" implies to me)?

On the subject of the dash, I think the dash becomes less clarifying if the parameter description is a full sentence (because it typically starts with a capital letter). Eg here, "The signature `v` value to extract the current chain from." or similar.

pdyraga:
Is it still relevant or are we good to close it?

lionakhnazarov:
Closing — this @ledgerhq/hw-app-eth bump targets the legacy dashboard dependencies, now extracted out of this repo. No longer applicable.

PR #2289, opened 2021-01-15
Calculate APY w/o connected wallet

r-czajkowski:
This PR adds support for calculating the APY value for each `LPRewards` contract without a connected wallet.


michalsmiarowski (approved):
Looking good 👍 Tested locally and it works fine - the APY is calculated before connecting to the Metamask wallet

PR #2290, opened 2021-01-15
KEEP Token Dashboard release v1.7.1

pdyraga:
bdee794 is the commit hash for `token-dashboard/v1.7.1`

This bundles a fix from https://github.com/keep-network/keep-core/pull/2289.


michalsmiarowski (approved):
👍

PR #2291, opened 2021-01-18
Bump web3 dependency to v1.3.1

r-czajkowski:
The previous version of web3 (1.2.7) doesn't support the breaking changes of
version 9.0.0 of the MetaMask extension described in
MetaMask/metamask-extension#8077.


michalsmiarowski (approved):
Tested locally on Firefox and Chrome with MetaMask 9.0.1 and it's looking good. No more info about "MetaMask no longer injecting web3" both in console and in MetaMask wallet.

PR #2292, opened 2021-01-18
Fetch liquidity reward per week from contract

r-czajkowski:
We can compute weekly rewards based on data in LPRewards contract
instead of hardcoded values -> `LPReward.rewardRate * 7 days in seconds`


michalsmiarowski (approved):
👍

PR #2295, opened 2021-01-19
Saddle apy calc

r-czajkowski:
This PR adds support for calculating the APY for the TBTC/Saddle liquidity rewards.


solidity/dashboard/src/constants/constants.js
```diff
@@ -47,26 +47,30 @@ export const LIQUIDITY_REWARD_PAIRS = {
...
     address: "0xe6f19dab7d43317344282f803f8e8d240708174a",
+    pool: "SADDLE",
   },
   KEEP_ETH: {
     contractName: LP_REWARDS_KEEP_ETH_CONTRACT_NAME,
     label: "KEEP + ETH",
     viewPoolLink:
       "https://info.uniswap.org/pair/0xe6f19dab7d43317344282f803f8e8d240708174a",
     address: "0xe6f19dab7d43317344282f803f8e8d240708174a",
+    pool: "UNISWAP",
```
michalsmiarowski, line 58: Not important but I think we could extract the "SADDLE" and "UNISWAP" to a new constant like LIQUIDITY_POOL, since the "UNISWAP" is used three times already.

michalsmiarowski (approved):
Exceptional code 👏

PR #2299, opened 2021-01-20
Fix creating the wrapped token of the LPRewards

r-czajkowski:
Store in the `LPRewardsToWrappedTokenCache` address of the wrapped token instead of the web3 contract instance. In case the web3 provider has changed. This PR also fixes the `No "from" address...` error on staking in `LPRewards`.


michalsmiarowski (approved):
Tested locally and it works well 👍 Staking and withdrawing work as expected.

PR #2302, opened 2021-01-20
LP Token Balance

michalsmiarowski:
DIsplay LP Token Balance for the pairs in Liquidity Rewards page.

image

Also adds coma thousands separator to the reward (and lp balance) numbers.
We are not adding LP Balance to the Saddle pool.


solidity/dashboard/src/sagas/liquidity-rewards.js
```diff
@@ -60,6 +64,11 @@ function* fetchLiquidityRewardsData(liquidityRewardPair, address) {
       )
       // % of total pool in the `LPRewards` contract.
       shareOfPoolInPercent = percentageOf(lpBalance, totalSupply).toString()
+
+      lpTokenBalance = yield call(
+        [LiquidityRewards, LiquidityRewards.calculateLPTokenBalance],
+        shareOfPoolInPercent
```
r-czajkowski, line 70: I think we can't calculate LP token balance correctly based on the % of total pool in the `LPRewards` contract. I think we should:
* get the deposited LP tokens in `LPRewards` contract for a given user,
* calculate % of **Uniswap** pool based on an amount above,
* and calculate token balance based on this formula https://uniswap.org/docs/v2/advanced-topics/understanding-returns/#why-is-my-liquidity-worth-less-than-i-put-in

Thoughts? 🤔

solidity/dashboard/src/components/LiquidityRewardCard.jsx
```diff
@@ -107,6 +122,44 @@ const LiquidityRewardCard = ({
     )
   }
 
+  const renderLPBalance = () => {
```
r-czajkowski, line 125: I believe it should be a separate component that is able to handle amount formatting.

solidity/dashboard/src/components/LiquidityRewardCard.jsx
```diff
@@ -55,6 +57,19 @@ const LiquidityRewardCard = ({
       : formatPercentage(bn)
   }, [percentageOfTotalPool])
 
+  const formattedLPTokenBalance = useMemo(() => {
```
r-czajkowski, line 60: The component that I mentioned above should handle amount formatting.

r-czajkowski (changes requested):

solidity/dashboard/src/services/liquidity-rewards.js
```diff
@@ -159,6 +167,46 @@ class UniswapLPRewards extends LiquidityRewards {
...
+
+    const constantProduct = new BigNumber(pairData.reserve0).multipliedBy(
+      pairData.reserve1
+    )
+
+    const token0LiquidityPool = constantProduct
+      .div(pairData.token1Price)
+      .squareRoot()
+    const token1LiquidityPool = constantProduct
+      .multipliedBy(pairData.token1Price)
```
r-czajkowski, line 196: Shouldn't be `pairData.token0Price`?

solidity/dashboard/src/utils/arithmetics.utils.js
```diff
@@ -33,6 +33,12 @@ export const lte = (a, b) => {
 
 export const isZero = (a) => web3Utils.toBN(a).isZero()
 
+export const calculatePercentage = (percentage, value) => {
```
r-czajkowski, line 36: Can we rename `value` to `total`? because we calculate a value from percentage so we need total value. Also wondering about a better name for this function.

r-czajkowski (changes requested):

solidity/dashboard/src/utils/arithmetics.utils.js
michalsmiarowski: Since we calculate a value **from** a **percentage** of another number then maybe we could just name it as `fromPercentage`? 🤔 Can't think of a better name right now 😄

solidity/dashboard/src/services/liquidity-rewards.js
michalsmiarowski: I think it should not. Based on [Uniswap docs](https://uniswap.org/docs/v2/advanced-topics/understanding-returns/#why-is-my-liquidity-worth-less-than-i-put-in) we are using the price of the second token (token1) and we use it to calculate `token0LiquidityPool` and `token1LiquidityPool`.

michalsmiarowski:
Ready for another round of review.

solidity/dashboard/src/utils/arithmetics.utils.js
r-czajkowski: > Can't think of a better name right now
me too 😆

I think `fromPercentage` it's better than `calculatePercentage`. So let's rename it to `fromPercentage`.

solidity/dashboard/src/services/liquidity-rewards.js
r-czajkowski: You're right! 🚀

solidity/dashboard/src/sagas/liquidity-rewards.js
r-czajkowski: @Battenfield We wonder if our understanding is correct. Could you please take a look at it?

solidity/dashboard/src/services/liquidity-rewards.js
```diff
@@ -159,6 +163,46 @@ class UniswapLPRewards extends LiquidityRewards {
...
+   * @param {string} lpBalance Balance of liquidity token for a given uniswap pair deposited in
+   * the LPRewards` contract.
+   * @return {Promise<{token0: string, token1: string}>}
+   */
+  calculateLPTokenBalance = async (lpBalance) => {
+    const pairData = await getPairData(this.wrappedTokenAddress.toLowerCase())
+
+    const shareOfUniswapPool = percentageOf(
+      lpBalance.toString(),
+      fromTokenUnit(pairData.totalSupply).toString()
```
dimpar, line 181: why do we convert `pairData.totalSupply` to Wei only total supply, but not other values?

solidity/dashboard/src/services/liquidity-rewards.js
```diff
@@ -159,6 +163,46 @@ class UniswapLPRewards extends LiquidityRewards {
...
+      pairData.reserve1
+    )
+
+    const token0LiquidityPool = constantProduct
+      .div(pairData.token1Price)
+      .squareRoot()
+    const token1LiquidityPool = constantProduct
+      .multipliedBy(pairData.token1Price)
+      .squareRoot()
+
```
dimpar, line 194: @michalsmiarowski @r-czajkowski guys, are you sure we need all these? What we have is:
`token_liquidity_pool` = `pairData.reserve0`, where token is ex. KEEP
`eth_liquidity_pool` = `pairData.reserve1`

```
const eth_price = token_liquidity_pool / eth_liquidity_pool // KEEP price for 1 ETH
const constantProduct = pairData.reserve0 * pairData.reserve1 => token_liquidity_pool * eth_liquidity_pool

const token1LiquidityPool = (token_liquidity_pool * eth_liquidity_pool / (token_liquidity_pool / eth_liquidity_pool)).squareRoot() => token1LiquidityPool = eth_liquidity_pool, which is the same.
```
It seems that it should be enough to write:
```
return {
      token0: fromPercentage(
        shareOfUniswapPool,
        pairData.reserve0
      ).toString(),
      token1: fromPercentage(
        shareOfUniswapPool,
        pairData.reserve1
      ).toString(),
    }
```

What do you think?

michalsmiarowski: @dimpar yes, that's a good point. I tested it and everything works wonderfully.

I think that we can go even further and remove the shareOfThePool calculation and just do:

```
return {
    token0: fromTokenUnit(
      new BigNumber(lpBalance)
        .multipliedBy(pairData.reserve0)
        .dividedBy(pairData.totalSupply)
    ).toString(),
    token1: fromTokenUnit(
      new BigNumber(lpBalance)
        .multipliedBy(pairData.reserve1)
        .dividedBy(pairData.totalSupply)
    ).toString(),
  }
```

becuase right now we multiply the result by 100 when calling `percentageOf` function, and then in return we divide the result by 100 by calling `fromPercentage` function. So it's just unnecessary and can be simplified as above.

Same calculation can be found in [uniswap-sdk repo](https://github.com/Uniswap/uniswap-sdk/blob/a88048e9c4198a5bdaea00883ca00c8c8e582605/src/entities/pair.ts#L225).

dimpar: yup, it can be simplified further as you described @michalsmiarowski

r-czajkowski (approved):
Looking very good :rocket:. Let's :shipit:

PR #2307, opened 2021-01-25
Top up updates dapp

r-czajkowski:
This PR adds updates regarding top-ups:
* moves top-up table to the top of delegation page- that lets people see what they already have before delegating more, and would also put the complete top-up button at the top of the page,
* adds support for displaying the `top up is ready to be committed` notification- we could notify the user, when they have a top up that's ready to be committed, to try and prevent not realizing there's another step. The notification will be displayed every 5 minutes.


solidity/dashboard/src/components/DelegationOverview.jsx
```diff
@@ -50,6 +50,17 @@ const DelegationOverview = ({
...
+        isFetching={areTopUpsFetching}
+        skeletonComponent={<DataTableSkeleton columns={3} />}
+      >
+        <Tile>
+          <TopUpsDataTable
+            topUps={filteredTopUps}
+            initializationPeriod={initializationPeriod}
+          />
+        </Tile>
+      </LoadingOverlay>
```
Shadowfiend, line 63: One thing I want to consider (cc @liz-shinn) is whether we should also move existing delegations to the top (though not a huge deal).

solidity/dashboard/src/sagas/staking.js
```diff
@@ -178,3 +191,72 @@ function* fetchTopUps() {
...
+  const initializationPeriod = yield select(
+    (state) => state.staking.initializationPeriod
+  )
+
+  const topUpsReadyToCommit = topUps.filter(({ createdAt }) =>
+    moment
+      .unix(createdAt)
+      .add(initializationPeriod, "seconds")
+      .isBefore(moment())
+  )
```
michalsmiarowski, line 218: I think there is a problem with the top-ups (unless this is how it should work, but probably not). When you stake some tokens for a new operator, that is not present in top-ups table yet, the new row in top-ups will be created  but the commit top-up button is disabled. The notification that it's ready to be committed does not appear also. After refresh and connecting with the wallet again, everything works as expected.

michalsmiarowski: So the first problem is with notifications - they do not appear for newly created top-ups and i managed to debug the code to find the issue. The top-ups that already exist in the table have type of the `createdAt` field as the number, while newly created top-ups have it as `Moment` object. It is changed to number after we relog and top ups update from fetching. This is why, for newly created top-ups, the notification does not appear - the `topUpsReadyToCommit` variable above does not filter the topUps that have `createdAt` fiels as `Moment` object properly. We can resolve this by changing this line of code:

https://github.com/keep-network/keep-core/blob/0c5748cd54e8f479a8ffbeedb2a17fe06264591b/solidity/dashboard/src/reducers/staking.js#L153

to `moment().unix()`. There is also similar issue with the code that is a few lines below:

https://github.com/keep-network/keep-core/blob/0c5748cd54e8f479a8ffbeedb2a17fe06264591b/solidity/dashboard/src/reducers/staking.js#L163

P.S.

There is also similar problem with `availableTopUpAmount` here:
https://github.com/keep-network/keep-core/blob/0c5748cd54e8f479a8ffbeedb2a17fe06264591b/solidity/dashboard/src/reducers/staking.js#L148-L157

For newly created top-ups the value of it is in **string** while for the fetched top-ups it is **BN**. I haven't noticed any issues with it, but it probably needs investigation.

solidity/dashboard/src/components/DelegationOverview.jsx
```diff
@@ -50,6 +50,17 @@ const DelegationOverview = ({
   return (
     <section>
       <DelegationHeader type={context} selectedGrant={selectedGrant} />
+      <LoadingOverlay
+        isFetching={areTopUpsFetching}
+        skeletonComponent={<DataTableSkeleton columns={3} />}
+      >
+        <Tile>
+          <TopUpsDataTable
```
michalsmiarowski, line 58: There is also the issue with disabled **commit top-up** button for newly created top-ups.

https://github.com/keep-network/keep-core/blob/0c5748cd54e8f479a8ffbeedb2a17fe06264591b/solidity/dashboard/src/components/TopUpsDataTable.jsx#L52-L58

I'm not sure what `initializationPeriod` with the value of 1 should do, but i assume that after 1 second it should enable the button. We should probably somehow inform the component that this delegation is ready to be committed, because right now we have to refresh the page to commit a top-up.

A quick idea that pops into my mind is that we can maybe add `readyToBeCommitted` property to the topUp objects that are in the redux state? 🤔 And then just disable/enable the button based on that property?

solidity/dashboard/src/components/Message.jsx
```diff
@@ -11,6 +11,7 @@ import PendingActionMessage from "./messages/PendingActionMessage"
 import SuccessMessage from "./messages/SuccessMessage"
 import ErrorMessage from "./messages/ErrorMessage"
 import DelegationAlreadyCopiedMessage from "./messages/DelegationAlreadyCopiedMessage"
+import TopUpReadyToBeCommittedMessage from "./messages/TopUpReadyToBeCommittedMessage"
```
michalsmiarowski, line 14: **TODO** something that i should have done in https://github.com/keep-network/keep-core/pull/2272 but forgot 🤭 :

Move all messages to `index.js` file to reduce the imports.

Not super important tho - we can do it in separate PR after merging this one.

michalsmiarowski (changes requested):
Left some comments below that should be looked at, before the merge 😉

michalsmiarowski:
Also one more thing - looks like the notifications are not displayed until we get into the **Delegations** page - similar problem that i had with the liquidity rewards notification, which I fixed [this](https://github.com/keep-network/keep-core/pull/2272/commits/eef111f2c87603b8196533176fdd111c64e517c9) and [that](https://github.com/keep-network/keep-core/pull/2272/commits/5787afa092566779e625e38930c665ebd34840c1) commits. Maybe it's a similar issue here.

r-czajkowski:
> Also one more thing - looks like the notifications are not displayed until we get into the **Delegations** page - similar problem that i had with the liquidity rewards notification, which I fixed in [this](https://github.com/keep-network/keep-core/pull/2272/commits/eef111f2c87603b8196533176fdd111c64e517c9) and [that](https://github.com/keep-network/keep-core/pull/2272/commits/5787afa092566779e625e38930c665ebd34840c1) commit. Maybe it's a similar issue here.

I am afraid that unfortunately, it must be like that, at least for now 😢. Fetching delegations it's time-consuming, especially for delegations from grants and we fetch this data only once and only on `Overview`, `Delegations` and `Token Grants` page. Not sure if it's a good option to fetch delegations data when the user connects to the wallet.

solidity/dashboard/src/components/Message.jsx
r-czajkowski: > we can do it in separate PR after merging this one.
yeah, let's address it in a separate PR.

solidity/dashboard/src/sagas/staking.js
```diff
@@ -166,15 +182,92 @@ function* fetchTopUps() {
...
+        showMessage({
+          messageType: messageType.TOP_UP_READY_TO_BE_COMMITTED,
+          messageProps: {
+            sticky: true,
+            grantId: stake.grantId,
+          },
+        })
+      )
+    }
+  }
```
michalsmiarowski, line 261: There is an issue with top up notifications - there are too many of them after a while, if user is not closing them 😛 

So i kept the site working in the background for a while, with my wallet connected, and this is what happened:
image


We had [similar problem with liquidity rewards notification](https://github.com/keep-network/keep-core/pull/2346).

I think we only need one message for the top-ups, and we don't need to display it if it's already displayed. So i propose similar solution that i applied:

```
const displayedMessages = yield select((state) => state.messages)
const topUpReadeToBeCommittedNotificationAlreadyDisplayed = displayedMessages.some(
  (message) => {
    return message.messageType === messageType. TOP_UP_READY_TO_BE_COMMITTED
  }
)
```

and then display the message `if (topUpReadeToBeCommittedNotificationAlreadyDisplayed === true)`

solidity/dashboard/src/sagas/staking.js
```diff
@@ -166,15 +182,92 @@ function* fetchTopUps() {
...
+  if (isFromLiquidTokens) {
+    yield put(
+      showMessage({
+        messageType: messageType.TOP_UP_READY_TO_BE_COMMITTED,
+        messageProps: {
+          sticky: true,
+        },
+      })
+    )
+  }
```
michalsmiarowski, line 272: DIsplay this also `if (topUpReadeToBeCommittedNotificationAlreadyDisplayed === true)`.

Similar as above.

solidity/dashboard/src/sagas/staking.js
```diff
@@ -166,15 +182,92 @@ function* fetchTopUps() {
...
 }
+
+export function* watchTopUpReadyToBeCommitted() {
+  // Waiting for top-ups data.
+  yield take("staking/fetch_top_ups_success")
+  yield call(notifyTopUpReadyToBeCommitted)
+
+  while (true) {
+    // Every 5 minutes.
+    yield delay(moment.duration(5, "minutes").asMilliseconds())
```
michalsmiarowski, line 211: We can remove this comment because we can read clearly that it's a 5 minutes delay 😉

solidity/dashboard/src/reducers/staking.js
```diff
@@ -150,7 +156,7 @@ const topUpInitiated = (topUps, { operator, topUp }) => {
       {
         operatorAddress: operator,
         availableTopUpAmount: topUp,
-        createdAt: moment.unix(),
+        createdAt: moment().unix(),
       },
       ...topUps,
```
michalsmiarowski, line 161: The commit-top up button is still disabled (until we refresh) because here we add a topUp (lines 156-159) and we did not define `readyToBeCommitted` variable so it's undefined.

solidity/dashboard/src/sagas/staking.js
```diff
@@ -166,15 +182,92 @@ function* fetchTopUps() {
...
+
+function* notifyTopUpReadyToBeCommitted() {
+  const topUps = yield select((state) => state.staking.topUps)
+  const initializationPeriod = yield select(
+    (state) => state.staking.initializationPeriod
+  )
+
+  const topUpsReadyToCommit = topUps.filter((topUp) =>
+    isTopUpReadyToBeCommitted(topUp, initializationPeriod)
+  )
```
michalsmiarowski, line 224: The notification is also not displaying without refresh, because we get the topUps from the state, and the newly added ones don't have `readyToBeCommitted` defined, so the `topUpsReadyToCommit` array does not take them into account when filtering. See the comment above also.

michalsmiarowski (changes requested):
Overall good code quality, but I think it's still need some changes before the merge - left some comments to look at 🐺

solidity/dashboard/src/reducers/staking.js
michalsmiarowski: So we have to think of a way to update this newly created top-up with `readyToBeCommited = true` variable, without refreshing the page but i don't think it's possible without fetching the topUps from BE again 🤔 Any ideas?

solidity/dashboard/src/sagas/staking.js
r-czajkowski: Function `isTopUpReadyToBeCommitted` doesn't check against `readyToBeCommitted` field but takes into account `createdAt` and `initializationPeriod`. So it should work 🤔

solidity/dashboard/src/reducers/staking.js
r-czajkowski: The top-up will be updated in `notifyTopUpReadyToBeCommitted` saga. Before displaying notification we dispatch `staking/top_ups_ready_to_be_committed` action which should update top-ups that are ready to be commited.

solidity/dashboard/src/sagas/staking.js
michalsmiarowski: Yes, you're right! My bad 🙏

michalsmiarowski (approved):
Looking good! Let's ship it :shipit:

PR #2308, opened 2021-01-25
tBTC rewards allocation: 2021-01-15 -> 2021-01-22

pdyraga:
Adding tBTC rewards allocation computed in keep-network/keep-ecdsa#678 to KEEP Token Dashboard.

There were no changes to KEEP Token Dashboard in `master` since `token-dashboard/v1.8.1` so pushing this change directly to `master` instead of doing the usual approach with release branch and backport.


r-czajkowski (approved):
:shipit:

PR #2310, opened 2021-01-27
KEEP-only pool dashboard support

r-czajkowski:
This PR adds support for KEEP-only pool in the KEEP token dashboard.

Main changes:
* the new liquidity pool strategy to handle `TokenGeyser` contract API,
* deploy `KeepTokenGeyser` contract to the local/Ropsten network,
* script that initializes the `KeepTokenGeyser` contract- for development purposes in KEEP token dashboard.

<details>

obraz

obraz

obraz

obraz

</details>


solidity/dashboard/src/utils/general.utils.js
```diff
@@ -74,3 +74,17 @@ export const formatPercentage = (value) => {
...
+  }
+
+  let prefix = ""
+  if (value === 0.01) {
+    prefix = `<`
+  } else if (value >= 999) {
+    prefix = `>`
+  }
+  return `${prefix}${value}%`
+}
```
michalsmiarowski, line 90: Maybe we could actually set min and max values as arguments of the function? And then we could just set the default values for them (0.01 and 999). This would make the function more reusable in the future, because we could set up the range, for which we display the prefixes, by passing the right arguments.

In this case we could probably also change `value === 0.01` to `value <= 0.01`

solidity/dashboard/src/reducers/liquidity-rewards.js
```diff
@@ -16,6 +16,7 @@ const initialState = {
   KEEP_ETH: { ...liquidityPairInitialData },
   TBTC_ETH: { ...liquidityPairInitialData },
   KEEP_TBTC: { ...liquidityPairInitialData },
+  KEEP: { ...liquidityPairInitialData },
```
michalsmiarowski, line 19: I don't really like the idea of the state name for keep only pool to be just "KEEP" - seems too generic. Maybe we could rename it to KEEP_ONLY? 🤔

solidity/dashboard/src/components/KeepOnlyPool.jsx
```diff
@@ -0,0 +1,127 @@
...
+                withdrawLiquidityRewards(
+                  liquidityContractName,
+                  lpBalance,
+                  pool,
+                  awaitingPromise
+                )
+              }
+            >
+              withdraw all
+            </SubmitButton>
```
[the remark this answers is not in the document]
r-czajkowski: Regarding `border-radius` I think we should address it in a separate PR because the button(and not only) styles have changed in the whole app.

solidity/dashboard/src/hooks/useSetMaxAmountToken.js
```diff
@@ -0,0 +1,18 @@
+import BigNumber from "bignumber.js"
+import { useFormikContext } from "formik"
+import { toTokenUnit } from "../utils/token.utils"
+
+const useSetMaxAmountToken = (filedName, availableAmount) => {
+  const { setFieldValue } = useFormikContext()
```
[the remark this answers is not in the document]
r-czajkowski: I totally agree. But can we address it in a cleanup PR? Currently, The PR has 29 changed files. I would like to avoid extra work for reviewers and make a mistake in moving files. 💣 😆

solidity/dashboard/src/services/liquidity-rewards.js
```diff
@@ -188,30 +225,128 @@ class SaddleLPRewards extends LiquidityRewards {
...
+      // error message in the wrong way when the `hanleRevert` option is enabled
+      // [1]. So here we clone the rewards contract instance and disable the
+      // `hanldeRevert` option.
+      // References: [1]:
+      // https://github.com/ChainSafe/web3.js/issues/3742
+      const clonedLPRewardsContract = this.LPRewardsContract.clone()
+      clonedLPRewardsContract.handleRevert = false
+      return await clonedLPRewardsContract.methods.unstakeQuery(amount).call()
+    } catch (error) {
+      return 0
```
Battenfield, line 276: If there's a problem with our call the user will see a 0 reward balance. I wonder if there's a better way to catch this error and render a message to the user. How are we error handling similar cases in the dapp?

solidity/dashboard/src/contracts.js
```diff
@@ -125,6 +127,10 @@ const contracts = {
     artifact: LPRewardsTBTCSaddle,
     withDeployBlock: true,
   },
+  [KEEP_TOKEN_GEYSER_CONTRACT_NAME]: {
```
Battenfield, line 130: We should make sure to deploy these as a part of the local-setup scripts so we don't see local errors.

Still unsure how what that PR would look like 😅. At the very least we should create a GH issue on the local-setup repo when we merge these in.

Battenfield (commented):
LGTM. I left just a couple of nit comments, one on error handling and one point on `local-setup` coordination.

I also have a minor design nit on the ` <KeepOnlyPool /> ` components overlap on small screens. It currently overlaps at ~1300px. Ideally, this would scale down with the banner above it. However, the pool component has too much inside to scale down. An alternative solution would be to have the banner above it also overlap, so the pool component looks less out of place.
Screen Shot 2021-02-05 at 3 49 23 PM

solidity/dashboard/src/contracts.js
r-czajkowski: It will work with `local-setup` repo. We deploy `KeepTokenGeyser` contract to the local network in `2_deploy_contracts.js` file, please see https://github.com/keep-network/keep-core/pull/2310/commits/1164b2d3325827d3c3efbd1371e5cfaa7f627f89. The `2_deploy_contracts` will be run via [`truffle migrate`](https://github.com/keep-network/keep-core/blob/master/scripts/install.sh#L35), command ([more here](https://www.trufflesuite.com/docs/truffle/getting-started/running-migrations#command)), which will be triggered by [this script](https://github.com/keep-network/local-setup/blob/master/install-keep-core.sh#L38) in `local-setup` repo.

solidity/dashboard/src/services/liquidity-rewards.js
r-czajkowski: Good point with error handling. But we don't have any mechanism to display that kind of errors. Also, I'm wondering how can we handle it because we fetch data step by step (I mean we can't get all data in a single request) and sometimes the next value depends on the previous.

So for example, the user is not able to fetch the `deposited KEEP` and `rewarded KEEP` but is able to fetch APY. Should we display a generic error message or specific messages for those values that failed?

r-czajkowski:
> I also have a minor design nit on the <KeepOnlyPool /> components overlap on small screens. It currently overlaps at ~1300px. Ideally, this would scale down with the banner above it. However, the pool component has too much inside to scale down. An alternative solution would be to have the banner above it also overlap, so the pool component looks less out of place.

@Battenfield fixed in https://github.com/keep-network/keep-core/pull/2310/commits/c19f44b538414dcffe51dc6e382c8cc7d7eb6416. Now should be better.

Battenfield (approved):

michalsmiarowski (approved):
👍

PR #2311, opened 2021-01-29
Fix reward pool per week for saddle pool

michalsmiarowski:
Get reward pool per week value based on rewardRate() method from the
LPRewardsContract instead of hardcoding the value.


r-czajkowski (approved):
:shipit:

PR #2314, opened 2021-02-01
tBTC rewards allocation: 2021-01-22 -> 2021-01-29

pdyraga:
Adding tBTC rewards allocation computed in keep-network/keep-ecdsa#682 to KEEP Token Dashboard.

There were no changes to KEEP Token Dashboard in `master` since `token-dashboard/v1.9.0` so pushing this change directly to `master`.


Shadowfiend (approved):
Confirmed merkle root and total against the [distribution transaction](https://etherscan.io/tx/0x6250fae94202de0c057f9c574b5a0ee23c56eb1a9ef327c0288003ce0528f85c#eventlog), let's do this.

PR #2315, opened 2021-02-01
Add jq to the macos Setup Script

beaushinkle:
The [manual setup doc](https://github.com/keep-network/local-setup/blob/master/docs/manual-setup.adoc#title) details that "If you are not on macOS or you don’t have Homebrew, you need to install:", and then lists `jq`, but the `macos-setup.sh` script doesn't install `jq` for you.

I also confirmed that `jq` is indeed necessary for the project via bricking by not having it XD

This PR adds a brew line to install it similar to how the rest of the dependencies are installed!


Shadowfiend (approved):
Not… Entirely sure how we missed this until now. Good catch!

Shadowfiend (approved):
Yes yes, it's… That-a-way ⤵️

PR #2318, opened 2021-02-08
Create GH Actions workflow for build&test of Go client

michalinacienciala:
and tBTC release processes in order to reduce human involvment and make
the processes faster and less error prone. A part of this task is
migration from CircleCI jobs to GitHub Actions. This commit creates a
GitHub Action workfow for building and testing Go client.

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -24,43 +24,19 @@ jobs:
             --tag keep-client . 
       - name: Create and enter test results directory
         run: |
-          mkdir -p /tmp/test-results/keep-core-go
-          cd /tmp/test-results/keep-core-go/
-      - name: DEBUG
-        run: |
-          pwd
-      - name: DEBUG
-        run: |
-          ls
-      - name: DEBUG
-        run: |
-          ls /tmp/test-results/keep-core-go/
-      - name: DEBUG
-        run: |
-          ls /home/runner/work/keep-core/keep-core
+          mkdir -p $GITHUB_WORKSPACE/test-results/
       - name: Run Go tests
         run: |
           docker run \
-            --volume /tmp/test-results/keep-core-go:/mnt/test-results \
+            --volume $GITHUB_WORKSPACE/test-results:/mnt/test-results \
             --workdir /go/src/github.com/keep-network/keep-core \
             go-build-env \
             gotestsum --junitfile /mnt/test-results/unit-tests.xml
-      - name: DEBUG
-        run: |
-          pwd
-      - name: DEBUG
-        run: |
-          ls
-      - name: DEBUG
-        run: |
-          ls /tmp/test-results/keep-core-go/
-      - name: DEBUG
-        run: |
-          ls /home/runner/work/keep-core/keep-core
       - name: Publish Unit Test results
         uses: EnricoMi/publish-unit-test-result-action@v1.7
         if: always() # guarantees that this action always runs, even if earlier steps fail
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          files:  unit-tests.xml
+          files:  ./test-results/unit-tests.xml
           check_name: Go Test Results # name under which test results will be presented in GitHub (optional)
+          comment_on_pr: false # turns off commenting on Pull Requests
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -40,3 +40,4 @@ jobs:
           files:  ./test-results/unit-tests.xml
           check_name: Go Test Results # name under which test results will be presented in GitHub (optional)
           comment_on_pr: false # turns off commenting on Pull Requests
+          # temporary comment
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -1,0 +1,30 @@
+name: Go
+
+#TODO: extend the conditions once workflow gets tested together with other workflows 
+on: [push, pull_request]
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: satackey/action-docker-layer-caching@v0.0.11
+        continue-on-error: true # ignore the failure of a step and avoid terminating the job
+      - name: Create test results directory
+        run: |
+          mkdir -p /tmp/test-results/keep-core-go
+      - name: Run Docker build
+        run: |
+          docker build --target gobuild --tag go-build-env .
+          docker build --tag keep-client . 
+      - name: Run Go tests
+        run: |
+          docker run --volume /tmp/test-results/keep-core-go:/mnt/test-results --workdir /go/src/github.com/keep-network/keep-core go-build-env gotestsum --junitfile /mnt/test-results/unit-tests.xml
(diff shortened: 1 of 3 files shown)


.github/workflows/client.yml
```diff
@@ -0,0 +1,30 @@
...
+        run: |
+          mkdir -p /tmp/test-results/keep-core-go
+      - name: Run Docker build
+        run: |
+          docker build --target gobuild --tag go-build-env .
+          docker build --tag keep-client . 
+      - name: Run Go tests
+        run: |
+          docker run --volume /tmp/test-results/keep-core-go:/mnt/test-results --workdir /go/src/github.com/keep-network/keep-core go-build-env gotestsum --junitfile /mnt/test-results/unit-tests.xml
+          docker run --volume /tmp/test-results/keep-core-go:/mnt/test-results --workdir /go/src/github.com/keep-network/keep-core go-build-env cat /mnt/test-results/unit-tests.xml > ./unit-tests.xml
```
nkuba, line 23: Thoughts on breaking the lines a little bit? This could improve readability.

```suggestion
          docker run \
            --volume /tmp/test-results/keep-core-go:/mnt/test-results \
            --workdir /go/src/github.com/keep-network/keep-core \
            go-build-env \
            gotestsum --junitfile /mnt/test-results/unit-tests.xml
          docker run \
            --volume /tmp/test-results/keep-core-go:/mnt/test-results \
            --workdir /go/src/github.com/keep-network/keep-core \
            go-build-env \
            cat /mnt/test-results/unit-tests.xml > ./unit-tests.xml
```

.github/workflows/client.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/checkout@v1
+      - uses: satackey/action-docker-layer-caching@v0.0.11
+        continue-on-error: true # ignore the failure of a step and avoid terminating the job
+      - name: Create test results directory
+        run: |
+          mkdir -p /tmp/test-results/keep-core-go
+      - name: Run Docker build
+        run: |
+          docker build --target gobuild --tag go-build-env .
+          docker build --tag keep-client .
```
nkuba, line 19: Maybe we could switch places for these two? First run docker build and then create test results directory? We don't need to create the dir for running the build but for the following testing steps.

pdyraga:
What do you think - and is it possible - to disable the bot to post a new comment with test results after each push? IMO it is cluttering communication and all the most recent test results are available from the checks list.

.github/workflows/client.yml
michalinacienciala: Agreed, done.

.github/workflows/client.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: satackey/action-docker-layer-caching@v0.0.11
+        continue-on-error: true # ignore the failure of a step and avoid terminating the job
+      - name: Run Docker build
+        run: |
+          docker build \
+            --target gobuild \
+            --tag go-build-env .
+          docker build \
+            --tag keep-client . 
+      - name: Create and enter test results directory
```
nkuba, line 25: We're not changing a directory, right?
```suggestion
      - name: Create test results directory
```

.github/workflows/client.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Run Docker build
+        run: |
+          docker build \
+            --target gobuild \
+            --tag go-build-env .
+          docker build \
+            --tag keep-client . 
+      - name: Create test results directory
+        run: |
+          mkdir -p $GITHUB_WORKSPACE/test-results/
```
nkuba, line 27: Maybe we could simplify it and do just `mkdir test-results`? In the next step, we've got to use `$GITHUB_WORKSPACE` to define the volume as a path, but I believe we don't need it here.

nkuba (commented):
Just one tiny suggestion. Overall it looks very good to me.

michalinacienciala:
I've noticed one problem with EnricoMi/publish-unit-test-result-action@v1 action used for publishing of tests result - when a commit causes two separate workflows to trigger in parallel, the test results in the form of a 'check' are sometimes getting attached to incorrect workflow (i.e. to the one that did not use the EnricoMi/publish-unit-test-result-action@v1 action).
This is a bug that creators of the action are aware of (https://github.com/EnricoMi/publish-unit-test-result-action/issues/12).  I subscribed to that ticket, so I should get notified once this is fixed.

nkuba (approved):

.github/workflows/client.yml
```diff
@@ -0,0 +1,42 @@
+name: Go
+
+#TODO: extend the conditions once workflow gets tested together with other workflows 
+on:  
+  push:
+    branches:
+      - master
```
Shadowfiend, line 7: One thing that may be worth considering here, if we're continuing to run the Circle build, is to change this filter to run on certain branches (e.g. `rfc-18/*`) and change Circle to _not_ run on that filter. This effectively gives you a branch namespace to use for the work.

.github/workflows/client.yml
```diff
@@ -0,0 +1,42 @@
...
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: satackey/action-docker-layer-caching@v0.0.11
+        continue-on-error: true # ignore the failure of a step and avoid terminating the job
```
Shadowfiend, line 17: I'll flag that the comment here mostly seems to be repeating what the flag already communicates (unlike `if: always()` which isn't as obvious).

.github/workflows/client.yml
```diff
@@ -0,0 +1,42 @@
...
+            go-build-env \
+            gotestsum --junitfile /mnt/test-results/unit-tests.xml
+      - name: Publish unit test results
+        uses: EnricoMi/publish-unit-test-result-action@v1.7
+        if: always() # guarantees that this action always runs, even if earlier steps fail
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          files:  ./test-results/unit-tests.xml
+          check_name: Go Test Results # name under which test results will be presented in GitHub (optional)
+          comment_on_pr: false # turns off commenting on Pull Requests
```
Shadowfiend, line 42: If things are working correctly, I believe the check should already be bubbling the important parts of this information up, including line annotations, e.g. like https://github.com/keep-network/keep-ecdsa/actions/runs/555260607 . I don't think we need the junit intermediary here, see https://github.com/actions/toolkit/blob/master/docs/commands.md#problem-matchers and the setup-go action's problem matcher at https://github.com/actions/setup-go/blob/main/matchers.json .

Shadowfiend: (Of note: we may want to add our own matcher that works with our go-to test output.)

Shadowfiend:
Re: temporary commits to test failures: when reasonable, this is a great place to use reset + force-push to clear the commit away completely while the PR is still in draft, rather than clutter history with a test + revert. Another option is to have a separate branch for testing that runs parallel to your primary.

It's also imperfect, but worth looking at https://github.com/nektos/act, which provides limited support for running GitHub Actions locally.

.github/workflows/client.yml
nkuba: Implemented in https://github.com/keep-network/keep-core/pull/2329

.github/workflows/client.yml
michalinacienciala: You're right. I'm not changing it though, as this fragment of code will be removed once #2331 gets merged to master.

PR #2319, opened 2021-02-08
tBTC rewards allocation: 2021-01-29 -> 2021-02-05

pdyraga:
Adding tBTC rewards allocation computed in https://github.com/keep-network/keep-ecdsa/pull/683/ to KEEP Token Dashboard.

There were no changes in `master` for KEEP Token Dashboard since v1.9.0 so we can either release the new allocation as-is or bundle it up with https://github.com/keep-network/keep-core/pull/2310 in v1.10.0 release.


r-czajkowski (approved):
:shipit:

PR #2321, opened 2021-02-09
Flatten the `npm link` commands into one line

beaushinkle:
Calling `npm link` multiple times in one directory, for whatever reason,
unlinks all links and then only links the executed command. So:

```
npm link foo
npm link bar
```

will link foo, unlink `foo`, and then link `bar`. This leaves you with only
`bar` linked. To link both, we flatten the command:

```
npm link foo bar
```

tagging @Shadowfiend for review since he also reviewed the nearly-identical https://github.com/keep-network/local-setup/pull/59


Shadowfiend (approved):
Yusss

PR #2322, opened 2021-02-10
Incorporate the `ethlike` package

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-common/pull/64

Updated the `keep-common` dependency version to use the new `ethlike` package and regenerated contract bindings. This change is one of the steps which prepares the codebase to be ready for supporting new chains.

### DIFF
--- a/pkg/chain/ethereum/connect.go
+++ b/pkg/chain/ethereum/connect.go
@@ -6,6 +6,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/keep-network/keep-common/pkg/rate"
+
 	"github.com/keep-network/keep-common/pkg/chain/ethlike"
 
 	relaychain "github.com/keep-network/keep-core/pkg/beacon/relay/chain"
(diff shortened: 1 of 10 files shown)


dimpar:
Is the goal be eventually replace all `ethutil` with `ethlike`?

lukasz-zimnoch:
> Is the goal be eventually replace all `ethutil` with `ethlike`?

Would be ideal but further PRs show that this is not possible.

dimpar:
Okay, wrapping my head around it. Did some testing and for now it all builds and the client generated a new public key.

pkg/chain/ethereum/connect.go
```diff
@@ -93,7 +94,9 @@ func connectWithClient(
 		transactionMutex: &sync.Mutex{},
 	}
 
-	blockCounter, err := blockcounter.CreateBlockCounter(pv.client)
+	blockCounter, err := ethlike.CreateBlockCounter(
+		ethutil.NewBlockSourceAdapter(pv.client),
+	)
```
pdyraga, line 99: Can we have this logic hidden inside of `keep-common`s `ethutil`? I don't have the full context grasped yet but I think it could make sense to have `ethereum` talking only to `ethutil` and `ethutil` using some components from `ethlike` internally.

lukasz-zimnoch: Done. Please see:
- https://github.com/keep-network/keep-common/pull/64/commits/a8c0a5ca0a5d094451e6d66d2d5d4c3130c77bce
- https://github.com/keep-network/keep-core/pull/2322/commits/07bf236ec33bb6f8f75bd3a17c483df182395343

pdyraga (approved):

PR #2324, opened 2021-02-10
Docker image for building celo blockchain client for development

pdyraga:
Setting up Celo client locally for development is similar to setting up Ethereum client locally, except that we need to use specially prepared genesis file which is more complex than one for Ethereum.

The process of setting up keystore and doing chain genesis is similar as for Ethereum client. Not very complex but time-consuming and cumbersome.

The Docker image implemented here should help with this process.

A developer needs to build the image:
```
$ docker build -t celo-dev .
```

And can later start Celo blockchain locally with:
```
$ docker run -it -v $(pwd)/data:/mnt/data -p 8545:8545 -p 8546:8546 celo-dev
```

`$(pwd)/data` directory is where the chain data will be stored. Since it is a Docker volume, chain data survive container restarts.

In case the developer wants to start from a fresh chain, it is enough to remove chain data: `$ rm -rf data/celo`,
or even remove the entire data directory: `$  rm -rf data` and start the container again. Keystore will be automatically populated and chain genesis executed.


Shadowfiend:
Let's stick to the directory structure we have here: https://github.com/keep-network/keep-core/blob/master/infrastructure/docker/

pdyraga:
> Let's stick to the directory structure we have here: https://github.com/keep-network/keep-core/blob/master/infrastructure/docker/

Lol, yeah. I was struggling whether I should open this PR as a draft or not, mainly because of `docker/celo` structure I hate. This thing implemented here is for local development purposes. Would it fit well with general-purpose images inside `infrastructure/docker`? I am afraid the answer is "it does not" but it is still better than `docker/celo`...

Shadowfiend:
The ethereum setup there is also meant for local/testnet dev (even though we haven't been using it), so I think it works.

infrastructure/docker/celo/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM golang:1.14-alpine AS gobuild
+
+RUN apk add --update --no-cache \
+    g++ \
+	linux-headers \
+    make \
+    git
+
+ENV BIN_PATH=/usr/local/bin \
+    DATA_DIR=/data
```
nkuba, line 10: We usually mount volumes under `/mnt`, so maybe we could change it to `/mnt/data`? This will also affect the `start.sh` script.

infrastructure/docker/celo/start.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+mkdir -p /data/keystore
```
nkuba, line 3: Would it be worth defining `/data` dir as an argument or env variable? The script would be a little more generic and maybe could be used outside of this specific Docker image.

infrastructure/docker/celo/start.sh
```diff
@@ -0,0 +1,16 @@
...
+cp init/keystore/* /data/keystore
+cp init/password.txt /data/password.txt
+[ ! -d "/data/celo" ] && geth --nousb --datadir=/data init init/genesis.json
+
+ geth --port 3000 --networkid 1101 --identity "somerandomidentity" \
```
nkuba, line 8: Some day we may want to make some properties (e.g. ports and network id) configurable (with defaults) in case we want to run it along with an Ethereum client (geth or ganache) or run it outside of the local dev machine (e.g. keep test envs).
Taking into account the first case, would it be worth doing now?

infrastructure/docker/celo/start.sh
```diff
@@ -0,0 +1,16 @@
...
+    --ws --wsaddr "0.0.0.0" --wsport "8546" --wsorigins "*" \
+    --rpc --rpcport "8545" --rpcaddr "0.0.0.0" --rpccorsdomain "" \
```
nkuba, line 10: Thoughts on using different default ports for RPC and WS, not to cause confusion with Ethereum ones?

infrastructure/docker/celo/start.sh
```diff
@@ -0,0 +1,16 @@
...
+    --rpcapi "db,ssh,miner,admin,eth,net,web3,personal" \
+    --wsapi "db,ssh,miner,admin,eth,net,web3,personal" \
+    --datadir /data --syncmode "fast" \
+    --mine --miner.threads 1 --nousb \
+    --unlock 0x2b2976824682233807a197081119da511af12f7a --password data/password.txt \
```
nkuba, line 15: I just wanted to confirm that we're unlocking just one account. Don't we want to unlock the rest of them?

nkuba:
Would it be worth adding a README file with the example commands from the PR description on how to build and run it?  Also, I believe we can drop `-it` flag from the run command.

infrastructure/docker/celo/start.sh
dimpar: This is a special "Validator" account in a proof-of-stake mechanism which mines new blocks. I believe a single validator chain should be enough to test Celo locally. We got a `genesis.json` from a Celo team with one validator which is also has it's BLS key written in `extraData` [property](https://github.com/keep-network/keep-core/blob/docker-my-celo/infrastructure/docker/celo/init/genesis.json#L23).

infrastructure/docker/celo/Dockerfile
pdyraga: Renamed. I have also renamed `/init` to `/celo-init`, the new name should be more explicit about the purpose of this directory. Please see https://github.com/keep-network/keep-core/pull/2324/commits/8f4b4322fc20038e67e1a1d88c4057d34c6ccb67.

dimpar:
I think password.txt path is bad. I have this error: `Fatal: Failed to read password file: open /mnt/data/password.txt: no such file or directory` when trying to run docker image.

pdyraga:
> I think password.txt path is bad. I have this error: `Fatal: Failed to read password file: open /mnt/data/password.txt: no such file or directory` when trying to run docker image.

I did a mistake in 8f4b432 and failed to update directory name when creating dir for keystore. It is possible this is the reason you observed this error. Please try again, 9f868b7  should fix it.

infrastructure/docker/celo/start.sh
pdyraga: Done. I defined `CELO_DATA_DIR` and `CELO_INIT_DIR`. Please see 9f868b7.

infrastructure/docker/celo/start.sh
pdyraga: This one unlock is a must-have for the reason @dimpar mentioned.

infrastructure/docker/celo/start.sh
pdyraga: In `truffle-config.js` we have just one local network, for Celo and Ethereum running on port 8545, so I'd rather prefer to keep it as-is and aligned with truffle config. I am afraid making it parametrizable does not help that much now but can cause some confusion during the environment setup. https://github.com/keep-network/keep-core/blob/61f9bb7012bf583b09afb63b0655b230cdadcf25/solidity/truffle-config.js#L8-L12

infrastructure/docker/celo/start.sh
pdyraga: We use the same local network in truffle config for Celo and Ethereum (see my previous comment). I also do not think right now we'll be using those two chains at the same time. Maybe in the future but then, we'll have to make it parametrizable and think about what to do with truffle config.

pdyraga:
Ready for another chance @nkuba @dimpar.

infrastructure/docker/celo/start.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+mkdir -p $CELO_DATA_DIR/keystore
+cp $CELO_INIT_DIR/keystore/* $CELO_DATA_DIR/keystore
+cp $CELO_INIT_DIR/password.txt $CELO_DATA_DIR/password.txt
+[ ! -d "$CELO_DATA_DIR/celo" ] && geth --nousb --datadir=$CELO_DATA_DIR init $CELO_INIT_DIR/genesis.json
```
nkuba, line 6: Would it be worth printing a message if we get into chain initialization?

infrastructure/docker/celo/README.md
```diff
@@ -0,0 +1,19 @@
+# Celo client Docker image for local development
```
nkuba, line 1: Can we add a `.gitignore` file containing `data` dir path?

nkuba (commented):
It works as advertised. I left just some minor comments.

infrastructure/docker/celo/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Building the image:
+```
+$ docker build -t celo-dev .
+```
+
+Starting Celo blockchain locally with the image built:
+```
+$ docker run -it -v $(pwd)/data:/data -p 8545:8545 -p 8546:8546 celo-dev
```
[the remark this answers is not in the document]
pdyraga: Yeah, that's why I left it but forgot to answer your comment...

infrastructure/docker/celo/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    g++ \
+    linux-headers \
+    make \
+    git
+
+ENV BIN_PATH=/usr/local/bin
+
+# Clone, compile, and install Celo blockchain v1.2.2
+RUN git clone https://github.com/celo-org/celo-blockchain.git
+WORKDIR celo-blockchain
```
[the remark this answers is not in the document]
pdyraga: Absolutely!

infrastructure/docker/celo/start.sh
pdyraga: Celo/Geth is printing information in logs:

> INFO [02-11|19:32:23.105] Initialised chain configuration          config="{ChainID: 1101 Homestead: 0 DAO: <nil> DAOSupport: false EIP150: 0 EIP155: 0 EIP158: 0 Byzantium: 0 Constantinople: 0 Petersburg: 0 Istanbul: 0 Churrito: <nil>, Donut: <nil>, Engine: istanbul}"

Do you think it is not enough?

nkuba: Yep I noticed this line finally. It gets lost very easy in this output. But after reconsideration I say leave it as is.

infrastructure/docker/celo/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN make geth-musl
+RUN cp build/bin/geth $BIN_PATH
+
+FROM alpine:3.10
+
+ENV BIN_PATH=/usr/local/bin
+ENV CELO_DATA_DIR=/mnt/data
+ENV CELO_INIT_DIR=/celo-init
+
+VOLUME [$CELO_INIT_DIR]
```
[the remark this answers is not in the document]
nkuba: On the other hand maybe it's a good practice to declare it here to be explicit that we expect a volume to be mounted. If we decide to leave it maybe we should also consider adding [`EXPOSE`](https://docs.docker.com/engine/reference/builder/#expose/) to document ports?

infrastructure/docker/celo/README.md
pdyraga: I wasn't sure about it given that the developer may point to any directory when starting the container. However, given that the README mentioned `$(pwd)/data`, ignoring it by default is good for developer's convenience when running everything as README suggests.

dimpar:
Tested a client by running: 
`$ docker build -t celo-dev .`
`$ docker run -it -v $(pwd)/data:/mnt/data -p 8545:8545 -p 8546:8546 celo-dev`
`$ ./scripts/install-celo.sh`
`$ ./scripts/start.sh`

Everything worked as expected! Just temporary had to add in `mod.go`:
`replace github.com/ethereum/go-ethereum => github.com/celo-org/celo-blockchain v0.0.0-20210118164417-a37e7f096fd1
`

I'll leave the final approval to @nkuba

dimpar (dismissed):

infrastructure/docker/celo/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN git clone https://github.com/celo-org/celo-blockchain.git
+WORKDIR /go/celo-blockchain
+RUN git checkout v1.2.2
+RUN make geth-musl
+RUN cp build/bin/geth $BIN_PATH
+
+FROM alpine:3.10
+
+# Volume where chain data are persisted
+VOLUME [$CELO_DATA_DIR]
```
nkuba, line 21: Shouldn't this be below `ENV CELO_DATA_DIR=/mnt/data`?

nkuba (commented):
Just one thing and we're good to take off.

infrastructure/docker/celo/Dockerfile
pdyraga: Yes, should be! Fixed.

nkuba (approved):

PR #2326, opened 2021-02-10
Official (but fake) release branch for token dashboard 1.10.1

Shadowfiend:
We've already released this version via #2325 , but this branch allows testing a workflow for publishing a preview build of new mainnet releases for sanity checking before pulling the final trigger.

To do this, a new GitHub workflow is added that does PR-only builds on `releases/mainnet/token-dashboard` branches and deploys them to preview.dashboard.keep.network. A tweak is added to the routing in the dApp to use the basename provided by the `PUBLIC_URL` environment variable, which react-scripts also uses and which we set for these preview builds.

Note that the preview domain will currently give an SSL warnings as it's being serviced directly by the GCS certification. Our current Cloudflare certs don't cover *.dashboard.keep.network, so unfortunately we're going to have to deal with the cert warning for the time being.


Battenfield:
The most recent preview looks fully functional with no errors in the console.  I wasn't able to check token grants but all else is looking good.

.github/workflows/dashboard.yml
```diff
@@ -0,0 +1,71 @@
...
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [12.11.x]
+
+    if: startsWith(github.head_ref, 'releases/mainnet/token-dashboard/')
```
nkuba, line 22: Do we want this workflow to eventually replace dashboard building in CircleCI for test environments? I understand that this branch prefix check will be moved to mainnet related steps.

.github/workflows/dashboard.yml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    #- run: npm run lint
+    #- if: github.event_name == 'push' TODO uncomment when mainnet builds happen this way
+    #  run: npm run build --if-present
+    - if: github.event_name == 'pull_request'
+      run: npm run build
```
nkuba, line 37: Is this specific to `mainnet` releases that we build it with `npm` and not use docker as in the current Circle CI config for keep-dev and keep-test?

.github/workflows/dashboard.yml
```diff
@@ -0,0 +1,71 @@
...
+    #  name: Deploy Master to GCP
+    #  uses: thesis/gcp-storage-bucket-action@v3.1.0
+    #  with:
+    #    service-key: ${{ secrets.KEEP_DASHBOARD_UPLOADER_SERVICE_KEY_JSON }}
+    #    project: keep-prd-210b
+    #    bucket-name: dashboard.keep.network
+    #    build-folder: solidity/dashboard/build
+    # A pull_request event is a PR; deploy to preview bucket.
+    - if: github.event_name == 'pull_request'
+      name: Deploy PR to GCP
```
nkuba, line 52: Can we mention in the step name that we're deploying it as a preview?

nkuba (commented):
Amazing feature! 😍  I left some minor comments mostly for me to understand where we want to head with this workflow in the context of RFC-19 and the planned transition from Circle CI to GH Actions.

.github/workflows/dashboard.yml
Shadowfiend: This will probably be the most straightforward way to deploy it across the board, yep---but I don't need us to get there immediately. I wanted to leave enough context for that work though.

For reference, this is almost identical to how we deploy the Keep and tBTC sites (see https://github.com/keep-network/website/blob/master/.github/workflows/build.yaml). I wanted the initial scope very limited until we can understand implications, though.

.github/workflows/dashboard.yml
Shadowfiend: Well, docker uses npm internally, but it's actually specific to the preview build right now. I want to slowly pull docker + kube out of the deployment process for anything that's just serving static assets---they're just needless complexity.

.github/workflows/dashboard.yml
Shadowfiend: Beyond the URL there's no real distinction, and the URL is handled in a separate step (the “post preview URL” one), so I'm tempted to keep it as-is. But if you feel strongly I can add it :)

nkuba: I'm afraid of confusion as the commented step above has a similar name `Deploy Master to GCP`, but only after looking at the `bucket-name` you can see the real difference between these two.

Shadowfiend:
```suggestion
      name: Deploy PR preview to GCP
```
How's that?

nkuba: Amazing!

Shadowfiend:
Fixed the merge issue. We should check the preview to make sure it's all good, and then look at merging.

pdyraga (dismissed):
Looks good to me, there is one outstanding comment from @nkuba so I'll leave him the final approval and merge honors once this comment is addressed.

nkuba (approved):

PR #2331, opened 2021-02-12
Switch to Docker's build-publish-action in GitHub Actions

nkuba:
In this PR we are switching the Go building workflow to Docker's official [Build and Publish Action](https://github.com/marketplace/actions/build-and-push-docker-images).

We were experiencing some issues with running out of disk space while using `satackey/action-docker-layer-caching`. It was happening because the cache size was incremented with each execution. During testing, it [grew up pretty quickly to 7 GB](https://github.com/keep-network/keep-core/runs/1880553919?check_suite_focus=true). There is an issue describing such a situation https://github.com/satackey/action-docker-layer-caching/issues/55. I tried the proposed workarounds but was not satisfied with them.

According to [documentation](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) caches will be cleaned up when their total size reaches 5 GB, but in this case, we had just one cache of 7 GB.

I noticed that Docker released [an action set](https://github.com/marketplace/actions/build-and-push-docker-images) that may suit our needs. It also supports layers caching. Additionally, we can use it later for image publication to registries. During testing the cache size was around 1 GB.

What is worth mentioning is a clever way how GH's cache action handles access to caches between branches. Long story short:
1. restore matches first caches on the current branch and later on parents - [read more](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key)
2. sibling branches cannot access each other's caches - [read more](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)


michalinacienciala (approved):
Approving. Analogical solution was tried out for keep-ecdsa (see [1st run](https://github.com/keep-network/keep-ecdsa/actions/runs/567855840), [2nd run](https://github.com/keep-network/keep-ecdsa/actions/runs/567914440), [3rd run](https://github.com/keep-network/keep-ecdsa/actions/runs/567962633)) and the results were satisfying (1st run lasted ~16 min, second lasted ~9 min). Used memory also doesn't seem to accumulate, but it would be good to monitor this for a while.

michalinacienciala (approved):

PR #2336, opened 2021-02-16
Unification of GH Action solidity workflows between keep-core and keep-ecdsa

michalinacienciala:
and tBTC release processes in order to reduce human involvment and make
the processes faster and less error prone. A part of this task is
migration from CircleCI jobs to GitHub Actions. For keep-core part of
the workflow responsible for testing and linting of Solidity build
was already developed as a GitHub Actions workflow. This commit
introduces four changes to the workflow, to make the implementation
similiar to analogous workflow in keep-core.
Those changes are:
1. Added triggering of workflow on push and pull request events for
'rfc-18/*' branches
2. Updated versions of actions
3. Added step that builds solidity contracts
4. Name of 'test' step changed to 'build-and-test'

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -3,6 +3,8 @@ name: Solidity
 on:
   push:
     branches:
+      # TODO: Run only on master after we're fully migrated from Circle CI
+      - "rfc-18/**"
       - master
     paths:
       - "solidity/contracts/**"
@@ -11,6 +13,9 @@ on:
       - ".github/workflows/contracts.yml"
   pull_request:
     branches:
+      # TODO: Run on all branches or only on master (to be decided) 
+      # after we're fully migrated from Circle CI
+      - "rfc-18/**"
       - master
     paths:
       - "solidity/contracts/**"
@@ -19,15 +24,15 @@ on:
       - ".github/workflows/contracts.yml"
 
 jobs:
-  test:
+  build-and-test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
         with:
           node-version: "12.x"
       - name: Cache node modules
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         env:
           cache-name: cache-solidity-node-modules
         with:
@@ -41,6 +46,10 @@ jobs:
         working-directory: ./solidity
         run: |
           npm ci
+      - name: Build solidity contracts
+        working-directory: ./solidity
+        run: |
+          npm run compile
       - name: Run tests
         working-directory: ./solidity
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/contracts.yml
```diff
@@ -3,6 +3,8 @@ name: Solidity
 on:
   push:
     branches:
+      # TODO: Run only on master after we're fully migrated from Circle CI
+      - "rfc-18/**"
       - master
```
nkuba, line 8: Can we remove `master` for now? In client workflow we're running it only on `rfc-18/**` branches. We should probably do the same here.

.github/workflows/contracts.yml
```diff
@@ -11,6 +13,9 @@ on:
       - ".github/workflows/contracts.yml"
   pull_request:
     branches:
+      # TODO: Run on all branches or only on master (to be decided) 
+      # after we're fully migrated from Circle CI
+      - "rfc-18/**"
       - master
```
nkuba, line 19: Can we remove `master` for now? In client workflow we're running it only on `rfc-18/**` branches. We should probably do the same here.

.github/workflows/contracts.yml
```diff
@@ -11,6 +13,9 @@ on:
...
     paths:
       - "solidity/contracts/**"
```
nkuba, line 21: We should also add `solidity/contracts/test/**`. But maybe it's not worth listing all the specific paths (as in the future we will add `solidity/migrations/**`) and we could generalize it to `solidity/**`. If so we would have to exclude `solidity/dashboard`. So we would have:
```yml
- "solidity/**"
- "!solidity/dashboard/**"
- ".github/workflows/contracts.yml"
```

.github/workflows/contracts.yml
```diff
@@ -41,6 +46,10 @@ jobs:
         working-directory: ./solidity
         run: |
           npm ci
+      - name: Build solidity contracts
+        working-directory: ./solidity
+        run: |
+          npm run compile
       - name: Run tests
         working-directory: ./solidity
         run: |
```
nkuba, line 55: I'm not sure we need `exit 0` command here. Can we remove it?

nkuba (changes requested):
Can we drop the pipeline from all the one-line run commands?

.github/workflows/contracts.yml
michalinacienciala: In case of client workflow we have equivalent of it present in CircleCI, executing for master, so there was no point in executing it for macter in GH Actions. 
In case of contracts workflow, there is no equivalent of this workflow in CircleCI, hence executing it for events on master is not duplicating anything.

.github/workflows/contracts.yml
michalinacienciala: In case of client workflow we have equivalent of it present in CircleCI, executing for master, so there was no point in executing it for macter in GH Actions.
In case of contracts workflow, there is no equivalent of this workflow in CircleCI, hence executing it for events on master is not duplicating anything.

nkuba: Oh, yeah :100: Thanks for the clarification!

nkuba (approved):

PR #2338, opened 2021-02-18
Refactor pkg/operator to support Celo chain

lukasz-zimnoch:
So far, the `operator` package strongly depended on the `go-ethereum/crypto` package. This fact excluded the possibility of supporting Celo as this require to rely on the `celo-blockchain/crypto` package. Normally we could try to hide implementation details behind interfaces but that's not a case. Under the hood `go-ethereum/crypto` and `celo-blockchain/crypto` packages use the same parts of native C code and this causes C linker errors during the `go build` due to duplicated symbols. To address that problem, Go build tags have been used. The `operator` package now contains two versions of the `key.go` file, one for Ethereum and one for Celo. During a normal `go build` the Ethereum version is selected by default. To use the Celo version, the `celo` build tag should be added. Apart from that, there was a need to remove the `go-ethereum/crypto` dependency from the `pkg/net/key` package. Now, this package directly depends on the `operator` package which plays nice with the build tag system. Last but not least, some function renaming has been done to replace ETH-focused naming and make it more generic.


pdyraga (approved):
Looks good, we'll have to revisit it once we finalize https://github.com/keep-network/keep-ecdsa/pull/693.

PR #2344, opened 2021-02-24
Chain specific balance monitor

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-common/pull/66

According to the changes made in `keep-common`, the balance monitor is no longer a chain-agnostic concept. Because of that, this change removes it from the `chain.Handle` interface. Instead of that, each chain implementation has a possibility to decide whether it provides a balance monitor or not. Apart from that, we also regenerate contract bindings in order to apply changes made in `keep-common` generators.


cmd/start.go
```diff
@@ -45,7 +48,10 @@ const startDescription = `Starts the Keep client in the foreground. Currently th
 // Values related with balance monitoring.
 // defaultBalanceAlertThreshold determines the alert threshold below which
 // the alert should be triggered.
-var defaultBalanceAlertThreshold = big.NewInt(500000000000000000) // 0.5 ether
+var defaultBalanceAlertThreshold = commoneth.WrapWei(
+	big.NewInt(500000000000000000),
+)
```
pdyraga, line 53: Non-blocking and not necessarily in this PR but I think it's worth considering to have `ethereum.Wei` function in`keep-core` that would call `keep-common/pkg/chain/ethereum` internally. 

This is an additional layer but it feels good when the only package talking with `keep-common`'s ethereum utils is `ethereum`.

cmd/start.go
```diff
@@ -267,30 +273,30 @@ func initializeBalanceMonitoring(
 	ctx context.Context,
 	chainProvider chain.Handle,
 	config *config.Config,
-	ethereumAddress string,
+	ethereumKey *keystore.Key,
 ) {
-	balanceMonitor, err := chainProvider.BalanceMonitor()
+	balanceMonitor, err := chainProvider.(*ethereum.EthereumChain).BalanceMonitor()
```
pdyraga, line 278: This change is contradicting with @Shadowfiend's approach presented in https://github.com/keep-network/keep-ecdsa/pull/693/. In the referenced PR we made `EthereumChain` not exported and `ethereum.Connect` function returns `chain.Handle` instead of `EthereumChain`.

If we consider `BalanceMonitor` Ethereum-specific concept, and it indeed makes sense and was noted in the referenced PR, we need to find better way to instantiate it than having to cast `chain.Handle`.

One option is to have `ethereum.Connect` return `chain.Handle` interface and `ethereum.BalanceMonitor`. What do you think about it?

pkg/chain/ethereum/ethereum.go
```diff
@@ -490,13 +492,24 @@ func (ec *ethereumChain) CalculateDKGResultHash(
 	return relayChain.DKGResultHashFromBytes(hash)
 }
 
-func (ec *ethereumChain) Address() common.Address {
+func (ec *EthereumChain) Address() common.Address {
 	return ec.accountKey.Address
 }
 
-func (ec *ethereumChain) WeiBalanceOf(address common.Address) (*big.Int, error) {
+func (ec *EthereumChain) WeiBalanceOf(
```
pdyraga, line 499: We can take the opportunity and make this function an anonymous one inside of `BalanceMonitor()`.

cmd/start.go
```diff
@@ -267,30 +273,30 @@ func initializeBalanceMonitoring(
 	ctx context.Context,
 	chainProvider chain.Handle,
 	config *config.Config,
-	ethereumAddress string,
+	ethereumKey *keystore.Key,
```
pdyraga, line 276: Why do we pass the entire `*keystore.Key` instead of passing just the address?

cmd/start.go
lukasz-zimnoch: This is just a temporary workaround. Once we'll do the multi-chain refactor this one will disappear from `start.go` entirely. However, I agree it can look better. Please see: https://github.com/keep-network/keep-core/commit/89f10b78f353c33ad427d4f7215725039126805b

pkg/chain/ethereum/ethereum.go
lukasz-zimnoch: Addressed in https://github.com/keep-network/keep-core/commit/89f10b78f353c33ad427d4f7215725039126805b

cmd/start.go
lukasz-zimnoch: Good point, see https://github.com/keep-network/keep-core/commit/b02c0e42d1a8db35f487348cb557145be2de8eab

cmd/start.go
lukasz-zimnoch: We can but I don't want to go too far in this PR. I Will post a TODO: https://github.com/keep-network/keep-core/commit/3f9f104b3d8be0b44ee6adb230d60314551487cc

lukasz-zimnoch:
@pdyraga all comments addressed!

pkg/chain/ethereum/connect.go
```diff
@@ -368,3 +369,29 @@ func fetchChainConfig(ec *ethereumChain) (*relaychain.Config, error) {
 		RelayEntryTimeout:          relayEntryTimeout.Uint64(),
 	}, nil
 }
+
+func BalanceMonitor(chain chain.Handle) (*ethutil.BalanceMonitor, error) {
```
pdyraga, line 373: I do not think this is the correct approach. We need to cast the interface to `*ethereumChain` so there is an assumption about the underlying type. Also, since we are creating Ethereum balance monitor, only the specific Ethereum implementation should be allowed as a parameter. I'd rather return the balance monitor pointer as a second parameter from `ethereum.Connect`.

cmd/start.go
pdyraga: We will still need to have a way to create the balance monitor and even though all of this code will be in Ethereum-specific connector, we will still have the same problem, nope?

pkg/chain/ethereum/connect.go
lukasz-zimnoch: But we're in the `ethereum` package. Also, it's not casting but type assertion. We receive a generic chain, we check wheter it's ethereum and if yes, return the balance monitor. I would like o to avoid returning I from `connect` as balance monitoring is optional.

cmd/start.go
lukasz-zimnoch: We can make the `ethereum.Connect` return `*ethereumChain` and use it only in a specific connector function which will return `chain.Handle`. We do this in `keep-ecdsa` https://github.com/keep-network/keep-ecdsa/blob/celo/cmd/connector_ethereum.go. Then we won't need type assertion in `ethereum.BalanceMonitor`.

pkg/chain/ethereum/connect.go
lukasz-zimnoch: Also, looks like it's a temporary inconvenience. Please see https://github.com/keep-network/keep-core/pull/2344#discussion_r582930532

pdyraga: I am not sure this is temporary, it depends on where the connector will live. It still feels squirmy to me and IMO is an indicator that something is wrong with our abstractions. I think we should either return a concrete type from `Connect()` function or if we want to return `chain.Handle` return chain-specific extensions by either expanding `chain.Handle` or returning them as second value from `Connect` function.

Balance monitor is not triggered until `Observe` function is called. Another option could be to make connector a chain-specific concept and pass the configuration to `Connect` function. This latter can be more complex given that there is no single chain-specific session for ECDSA client.

cmd/start.go
pdyraga: Unless we move the connector to `ethereum` package, we run into a conflict with the approach from https://github.com/keep-network/keep-ecdsa/pull/693 because we'll have to make `ethereumChain` exported and return it from the `Connect` function instead of `chain.Handle`.

pkg/chain/ethereum/connect.go
pdyraga: After an out-of-band discussion with @Shadowfiend, the last suggested option seems to us the best one, and given that https://github.com/keep-network/keep-ecdsa/pull/693 proposes to move tBTC system address from tBTC extension section into Ethereum configuration section, it should be enough to pass Ethereum configuration to `ethereum.Connect` function.

It plays well with build flags. Instead of having too much initialization logging in build-flag-dependent `cmd` package connector code, we just push it all behind chain's `Connect` facade that will do the right thing based on the configuration provided.

cmd/start.go
lukasz-zimnoch: Please see https://github.com/keep-network/keep-core/commit/5ac723255980b02ad039e199108b0de8202e2ece

pkg/chain/ethereum/connect.go
lukasz-zimnoch: Please see https://github.com/keep-network/keep-core/commit/5ac723255980b02ad039e199108b0de8202e2ece

pdyraga (approved):

PR #2346, opened 2021-02-25
Coalescence all liquidity rewards notifications

michalsmiarowski:
The liquidity rewards notifications are displayed too often right now and some user got too many of them.

<details>
  <summary>Example of the issue</summary>

image

</details>

This PR fixes that by:
 - extending the time when liquidity rewards are checked from 7 minutes to 30 minutes
 - merging notification from each pair into one (see how it looks like on the screenshot below)
 - adding new property in redux state for keeping data related to the notification type
 - not displaying another notification when it's already displayed - instead we just add a pair name to the content of it (unless it's also present in the currently displayed message)

Additional fixes:
- change pair name in the notification eg. KEEP_TBTC -> KEEP+TBTC

image


solidity/dashboard/src/sagas/liquidity-rewards.js
```diff
@@ -124,7 +124,7 @@ export function* watchLiquidityRewardNotifications() {
       )
     }
     displayMessage = true
-    yield delay(moment.duration(7, "minutes").asMilliseconds()) // every 7 minutes
+    yield delay(moment.duration(30, "minutes").asMilliseconds()) // every 7 minutes
```
r-czajkowski, line 127: I think we can remove comment `// every 7 minutes`. `moment.duration()` is enough.

solidity/dashboard/src/sagas/messages.js
```diff
@@ -5,13 +5,34 @@ import {
...
 } from "../actions/messages"
+import { messageType } from "../components/Message"
 
 function* addMessage(action) {
   yield put({ type: ADD_MESSAGE, payload: action.payload })
 }
 
 function* removeMessage(action) {
-  yield put({ type: REMOVE_MESSAGE, payload: action.payload })
+  console.log("action", action)
```
r-czajkowski, line 15: unnecessary `console.log`

solidity/dashboard/src/sagas/messages.js
```diff
@@ -5,13 +5,34 @@ import {
...
+          type:
+            "notifications_data/liquidityRewardNotification/pairs_displayed_updated",
+          payload: [],
+        })
+        break
+      default:
+        break
+    }
+  }
+  yield put({
```
r-czajkowski, line 31: Tbh I don't understand why we are dispatching `notifications_data/liquidityRewardNotification/pairs_displayed_updated` action . Could you please explain?

r-czajkowski (changes requested):

solidity/dashboard/src/sagas/messages.js
michalsmiarowski: Sure. So in this PR we only have one notification of liquidity rewards which displays all the pairs that the user get rewards for (see screenshot that i uploaded on this PR). These pairs are stored in redux store, so to update the pairs we are dispatching the action that you mentioned. This case we don't need to display several notifications for all the pairs, but only one notification, and update the pairs in redux store.

Here we dispatch the `notifications_data/liquidityRewardNotification/pairs_displayed_updated` action and put an empty array in the payload, because we want to clear the data when user closes the message. If he closes it, then the next time we will check if he got any rewards for all the pairs and update the state. If he doesn't close it though, then we keep the pairs already displayed in the notification (the store is not empty) and check if he got any rewards for the pairs that are not displayed.

I tried to write it as simple as possible. Hope it worked out 🙈

r-czajkowski: I understand what this action is responsible for but I don't understand why we dispatch this action even though we handle it in the `switch` statement.

solidity/dashboard/src/sagas/messages.js
```diff
@@ -5,13 +5,33 @@ import {
...
 function* addMessage(action) {
   yield put({ type: ADD_MESSAGE, payload: action.payload })
 }
 
 function* removeMessage(action) {
-  yield put({ type: REMOVE_MESSAGE, payload: action.payload })
+  yield put({ type: REMOVE_MESSAGE, payload: action.payload.id })
+  if (action.messageType) {
+    // switch for future notification data clearance for other message types
+    switch (action.payload.messageType) {
```
r-czajkowski, line 18: I think there is a mismatch with actions. With `switch` statement the `removeMessage` action creator is responsible for 2 things which in my opinion is not a good idea. We can handle it in the `LiquidityRewardsEarnedMessage` by dispatching  `notifications_data/liquidityRewardNotification/pairs_displayed_updated` in `onMessageClose` function.

In `LiquidityRewardsEarnedMessage.jsx`

```
<Message
 ...
 onMessageClose={(messageId) => {
  dispatch({
          type:
            "notifications_data/liquidityRewardNotification/pairs_displayed_updated",
          payload: [],
    })
  props.onMessageClose(messageId)
 }}
/>
```

r-czajkowski (changes requested):

solidity/dashboard/src/sagas/messages.js
michalsmiarowski: Oh, i see! 😮 You're right! We shouldn't dispatch the action again after the `switch` statement.

r-czajkowski (approved):
looking good 👍

PR #2350, opened 2021-02-26
KEEP Token Tracker

nkuba:
## KEEP Token Tracker

This is a tool for tracking KEEP Token ownership. It will walk over predefined sources of truth and inspect KEEP Token ownership at a specific block number.

The tool returns a map of addresses with their total KEEP Token holdings.

## Sources Of Truth

Holders of the tokens can be externally owned accounts or contracts. It is necessary to resolve the actual owner for tokens held by contracts. 

The following sources are currently defined:

- KEEP Token contract
- Token Staking contract
- Token Grant contract
- Liquidity Rewards contracts

The list of supported sources of truth can be enhanced. For guidelines on 
implementing additional sources please see Development section.

## Development

Additional implementation of supported sources should be placed in `truth-sources` directory. It should implement the `ITruthSource` interface, with the `getTokenHoldingsAtFinalBlock` function that should return `Map<Address, BN>` (map of owners to their token holding).

## tmp

Outputs from stages of the script execution will be stored in `./tmp` directory. This may be useful for results verification. 

The directory will also contain files with unknown contract addresses that hold tokens. These addresses should be closely analyzed as a field to further Source Of Truth implementations.

## Run

`ETH_HOSTNAME` environment variable is required to be configured with an address of Ethereum API.
The script requires `--final-block` to be configured with a block number at which holdings should be calculated.

```sh
ETH_HOSTNAME="wss://eth-mainnet.ws.alchemyapi.io/v2/....." \
  ETH_ACCOUNT_PRIVATE_KEY=..... \
  ./bin/inspec-token-ownership.js --final-block 10000000
```


token-tracker/bin/inspect-token-ownership.js
```diff
@@ -0,0 +1,85 @@
...
+}
+
+async function run() {
+  if (!existsSync(TMP_DIR)) {
+    mkdirSync(TMP_DIR)
+  }
+
+  const result = await getTokenOwnership(program.opts().finalBlock)
+
+  dumpDataToFile(result, RESULT_DUMP_PATH)
```
pdyraga, line 71: Nitpick, but I'd consider writing `result.json` to another directory than `tmp`. All the other files that eventually "add up" to `result.json` can be there, but `result.json` does not feel like something temporary.

token-tracker/src/lib/ethereum-helper.js
```diff
@@ -0,0 +1,126 @@
...
+      )
+
+      try {
+        const endBlock = toBlock || (await web3.eth.getBlockNumber())
+
+        while (fromBlock <= endBlock) {
+          let toBlock = fromBlock + GET_PAST_EVENTS_BLOCK_INTERVAL
+          if (toBlock > endBlock) {
+            toBlock = endBlock
+          }
```
pdyraga, line 107: We can log partial event pull progress - something similar to what we added for staker rewards in https://github.com/keep-network/keep-ecdsa/pull/665. Then, if it takes a lot of time to pull events, one running the script does not have a feeling it's frozen.

pdyraga (commented):
First thoughts, much more to come.

token-tracker/README.adoc
```diff
@@ -0,0 +1,86 @@
...
+Following arguments are expected by the program:
+
+- `final-block` (required) - block at which calculations should be made
+
+Command example:
+
+```sh
+ETH_HOSTNAME="wss://eth-mainnet.ws.alchemyapi.io/v2/....." \
+  ETH_ACCOUNT_PRIVATE_KEY=..... \
+  ./bin/inspec-token-ownership.js --final-block 10000000
```
dimpar, line 68: `inspec-token-ownership.js` -> `inspect-token-ownership.js`

token-tracker/bin/inspect-token-ownership.js
```diff
@@ -0,0 +1,85 @@
...
+    context.web3
+  )
+
+  logger.debug(`deployment block: ${context.deploymentBlock}`)
+
+  return context
+}
+
+export async function getTokenOwnership(finalBlockNumber) {
+  if (!finalBlockNumber) throw new Error("final block not defined")
```
dimpar, line 48: shouldn't it be: `final block _is_ not defined` or maybe `final block must be defined`? up to you

token-tracker/src/lib/context.js
```diff
@@ -0,0 +1,89 @@
...
+   * @type {Contracts}
+   */
+  #contracts
+
+  /**
+   * Deployment block is considered a starting
+   * block number that can be used as 'fromBlock' for ethereum events lookup.
+   * @type {Number}
+   */
+  #deploymentBlock
```
dimpar, line 19: What do you think about naming `deploymentBlock` to something more specific, ex. `deploymentKeepTokenBlock`? If we add `lp tokens` to `truth-sources` then we also need to take into account deployment blocks of `lp tokens` and probably add them in here as well.

token-tracker/src/inspector.js
```diff
@@ -0,0 +1,68 @@
...
+    return this.tokenHoldings
+  }
+
+  /**
+   * Adds balances of new tokens holdings to the existing map of holdings. 
+   * @param {Map<Address,BN>} newHoldings Map of token holders to add.
+   */
+  addTokenHoldings(newHoldings) {
+    newHoldings.forEach((value, holder) => {
+      holder = toChecksumAddress(holder)
```
dimpar, line 58: do you think we can use: `this.context.web3.utils.toChecksumAddress()` instead of importing `Web3`?

token-tracker/src/inspector.js
```diff
@@ -0,0 +1,68 @@
...
+      value = new BN(value)
+
+      if (!this.tokenHoldings.has(holder)) {
```
dimpar, line 61: alternative to this logic to avoid setting to zero and right after adding a value can be:
```
if (this.tokenHoldings.has(holder)) {
  this.tokenHoldings.get(holder).iadd(value)
} else {
  this.tokenHoldings.set(holder, new BN(value))
}
```
but we can leave as is, up to you.

token-tracker/src/truth-sources/truth-source.js
```diff
@@ -0,0 +1,136 @@
...
+   * @return {Promise<[Boolean,String]>} Boolean flag if the address should be ignored
+   * and resolved type of an address, that can be used for logging on other usage
+   * specific for the given type.
+   */
+  async isIgnoredAddress(address) {
+    for (const [ignoredContractName, ignoredContractAddress] of Object.entries(
+      AddressesResolver.IGNORED_CONTRACTS
+    )) {
+      if (
+        toChecksumAddress(ignoredContractAddress) === toChecksumAddress(address)
```
dimpar, line 104: we can probably use `toChecksumAddress` from `this.web3.utils`

token-tracker/src/truth-sources/token-grant.js
```diff
@@ -0,0 +1,142 @@
...
+    for (const id of tokenGrantIDs) {
+      const grant = await callWithRetry(
+        this.tokenGrant.methods.getGrant(id),
+        undefined,
+        undefined,
+        this.finalBlock
+      )
+
+      // Balance is the total grant amount minus revoked and withdrawn tokens. The
+      // value includes staked tokens.
```
pdyraga, line 86: But it does not take into account slashed tokens... 😬

token-tracker/src/truth-sources/token-grant.js
```diff
@@ -0,0 +1,142 @@
...
+    /** @type {Map<Address,BN>} */
+    const ownersBalances = new Map()
+
+    for (const tokenGrant of tokenGrants) {
+      const { grantee, balance } = tokenGrant
+
+      const owner = await resolveGrantee(this.context.web3, grantee)
+
+      if (owner !== grantee)
+        logger.debug(`resolved owner of grantee [${grantee}]: ${owner}`)
```
pdyraga, line 115: Maybe `resolved grantee for managed grant [${grantee}]`?

Managed grant's grantee is the owner, "owner of a grantee" is confusing and that's not how we express relationships between actors in contracts and docs. Although, yes. From the perspective of this script's function, this is correct.

token-tracker/src/truth-sources/token-grant.js
pdyraga: What I think we can do:
- Take all `TokenGrantStaked` events for the given grant ID from `TokenGrant` contract.
- Take all `DepositRedelegated` events for the given grant ID from `TokenStakingEscrow` contract.
- Extract operator addresses.
- Take all `TokensSlashed` events for extracted operators, note `amount`.
- Take all `TokensSeized` events for extracted operators, note `amount`.
- When calculating `const balance` in the next line, `sub` the two noted amounts.

token-tracker/src/truth-sources/token-staking.js
```diff
@@ -0,0 +1,187 @@
...
+
+      const amount = toBN(delegationInfo.amount)
+      const undelegatedAt = toBN(delegationInfo.undelegatedAt)
+
+      if (undelegatedAt.gtn(0)) {
+        logger.debug(
+          `skipping delegation to ${operator}, undelegated at: ${undelegatedAt.toString()}`
+        )
+        continue
+      }
```
pdyraga, line 151: Should we skip them?

`undelegatedAt` is set when the delegation owner starts undelegation:
https://github.com/keep-network/keep-core/blob/c9941e7923f03343eec1fc7f12e2d1d56bbbe262/solidity/contracts/TokenStaking.sol#L249-L251

Still, tokens are still staked at this point. When stake is recovered from `TokenStaking` contract, the amount is set to 0 but `undelegatedAt` is not updated:

https://github.com/keep-network/keep-core/blob/c9941e7923f03343eec1fc7f12e2d1d56bbbe262/solidity/contracts/TokenStaking.sol#L273

If the amount is non-zero, it means it is still in `TokenStaking` contract and we should include it.

token-tracker/src/truth-sources/token-staking.js
```diff
@@ -0,0 +1,187 @@
...
+
+import TokenStakingJSON from "@keep-network/keep-core/artifacts/TokenStaking.json"
+
+const TOKEN_STAKING_HISTORIC_STAKERS_DUMP_PATH =
+  "./tmp/token-staking-stakers.json"
+const TOKEN_STAKING_BALANCES_DUMP_PATH = "./tmp/token-staking-balances.json"
+const TOKEN_STAKING_UNKNOWN_OWNERS_CONTRACTS_DUMP_PATH =
+  "./tmp/token-staking-unknown_owners_contracts.json"
+
+// TODO: Add support for the old TokenStaking contract.
```
pdyraga, line 24: Correct. The logic should be identical, so we can probably just parameterize this code and call it twice.

token-tracker/bin/inspect-token-ownership.js
```diff
@@ -0,0 +1,85 @@
...
+  .requiredOption("--final-block <number>", "final block number")
+  .parse(process.argv)
+
+const TMP_DIR = "./tmp"
+const RESULT_DUMP_PATH = "./tmp/result.json"
+
+async function initializeContext() {
+  const context = await Context.initialize(
+    process.env.ETH_HOSTNAME,
+    process.env.ETH_ACCOUNT_PRIVATE_KEY || "01".repeat(32)
```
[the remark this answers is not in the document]
nkuba: f4ede5bafcb2b9b52c6bf5911fa444fb9ffa69d2

token-tracker/bin/inspect-token-ownership.js
```diff
@@ -0,0 +1,85 @@
...
+import { TokenStakingTruthSource } from "../src/truth-sources/token-staking.js"
+import { TokenGrantTruthSource } from "../src/truth-sources/token-grant.js"
+
+import KeepTokenJson from "@keep-network/keep-core/artifacts/KeepToken.json"
+
+import commander from "commander"
+const program = new commander.Command()
+
+program
+  .requiredOption("--final-block <number>", "final block number")
```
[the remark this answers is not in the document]
nkuba: 2ac1536b4

token-tracker/bin/inspect-token-ownership.js
nkuba: Good point, fa340c0f5

token-tracker/src/lib/ethereum-helper.js
nkuba: There won't be a frozen feeling as each call logs `making attempt number 0`. But I added what you proposed.
12200318b716d04bcfa35a4bffa3ba6031344c89

token-tracker/src/lib/context.js
dimpar: Actually, I take it back, sorry. KEEP is the "main" token here and LP tokens will be in the truth sources. I'll have deployments block of LP tokens defined in LP specific code.

token-tracker/src/truth-sources/token-grant.js
```diff
@@ -117,11 +117,11 @@ export class TokenGrantTruthSource extends ITruthSource {
       if (owner !== grantee)
         logger.debug(`resolved owner of grantee [${grantee}]: ${owner}`)
 
-      if (!ownersBalances.has(owner)) {
+      if (ownersBalances.has(owner)) {
+        ownersBalances.get(owner).iadd(balance)
+      } else {
         ownersBalances.set(owner, toBN(0))
```
dimpar, line 123: `ownersBalances.set(owner, toBN(0)) -> ownersBalances.set(owner, toBN(balance))`

token-tracker/src/truth-sources/keep-token.js
```diff
@@ -0,0 +1,215 @@
...
+import { EthereumHelpers } from "@keep-network/tbtc.js"
+const { callWithRetry } = EthereumHelpers
+
+import { getPastEvents } from "../lib/ethereum-helper.js"
+import { dumpDataToFile } from "../lib/file-helper.js"
+import { logger } from "../lib/winston.js"
+
+import KeepTokenJSON from "@keep-network/keep-core/artifacts/KeepToken.json"
+
+const KEEP_TOKEN_HISTORIC_HOLDERS_DUMP_PATH = "./tmp/keep-token-holders.json"
```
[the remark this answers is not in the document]
dimpar: there is also a function with `dump` -> `dumpDataToFile`

token-tracker/src/truth-sources/token-grant.js
nkuba: What if we subtract the staked amount in `token-grant.js` and resolve the staked tokens amount from grants in `token-staking.js`?

pdyraga: Resolving staked amounts from grants will be more complex unless I don't fully understand your proposition. The advantage of the current (and proposed) approach is that we don't have to care about the staking escrow used only for grants. It may happen that from the perspective of `TokenGrant`, the given amount is still staked while `TokenStaking` says the active stake is 0. This amount (and more!) can live in multiple slots in `TokenStakingEscrow`.

nkuba: > Take all DepositRedelegated events for the given grant ID from TokenStakingEscrow contract.

Do we need to bother about `TokenStakingEscrow`? We could check both `TokenStaking` and `OldTokenStaking` for the slash and seize events. Am I correct?

token-tracker/bin/inspect-token-ownership.js
dimpar: Do we want to add a new location of `result.json` or just this file to `.gitignore`? Everytime you run this `token tracker combine` it will appear with new git diff

token-tracker/src/truth-sources/token-grant.js
nkuba: Solution proposed in e8c7bb2d6c80fd7ddb693b9b8edd6bb2695bda33

token-tracker/src/truth-sources/token-staking.js
nkuba: a467f405b1109378314bf29da9071bd3147d3cb2

token-tracker/src/truth-sources/token-staking.js
nkuba: 069c0cf7eaf8c63dd779a3ee41c50d9fdd7311b1

token-tracker/src/truth-sources/keep-token.js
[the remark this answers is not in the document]
nkuba: Do you think I should also change the name for this function? Does _dump_ also have bad connotations with data?

token-tracker/bin/inspect-token-ownership.js
nkuba: 014557ed06475e8a91405973c4e6ecf152ecc6cb

token-tracker/src/truth-sources/token-grant.js
pdyraga: > > Take all DepositRedelegated events for the given grant ID from TokenStakingEscrow contract.
> 
> Do we need to bother about `TokenStakingEscrow`? We could check both `TokenStaking` and `OldTokenStaking` for the slash and seize events. Am I correct?

I haven't had a chance to review the proposed solution yet but the reason I mentioned `TokenStakingEscrow` is that the stake from a grant could be unstaked, tokens could be deposited into `TokenStakingEscrow`, and then redelegated to another operator. In this scenario, we need to inspect `TokensSlashed` and `TokensSeized` for two operator addresses.

token-tracker/src/truth-sources/keep-token.js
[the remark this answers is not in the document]
dimpar: Actually I don't know if it has a bad connotation, I figured I mention it just for consistency since you changed the `const` name of the path. I'm fine with both names.

token-tracker/src/truth-sources/token-grant.js
nkuba: Covered in 26779ac6e262614345fe3792aff74cf523276cf1. I also included any withdrwan or revoked tokens.

token-tracker/src/truth-sources/keep-token.js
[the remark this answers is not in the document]
pdyraga: I think it is fine in the context of the data dump, but I'd like to avoid it when it's about tokens.

token-tracker/README.adoc
```diff
@@ -0,0 +1,95 @@
...
+a holding can occur in few of them. E.g. when an owner staked tokens from a grant
+we could end up with duplications in TokenStaking and TokenGrant. To handle such
+situation we ignore any non externally owned accounts in TokenStaking.
+
+## Known Issues
+
+### error: artifact does not define network *
+
+Web3 provider may return invalid results for `getChainId()` function. The error
+looks like on this example:
```
pdyraga, line 89: Any idea what is the root cause?

pdyraga (approved):
Yes. 🧔

token-tracker/README.adoc
nkuba: Not sure. At first, I had problems when I was using `CacheSubprovider` from `web3-provider-engine`. After I removed it the problem disappeared.... for some time. I had no chance to dig any further.

PR #2352, opened 2021-03-01
Update truffle dependency version

nkuba:
A couple of changes here, that we're hoping to stabilize ropsten migrations:
- updated `truffle` and `@truffle/hdwallet-provider` to the latest ones. It also sorts the list of dependencies,
- increased some timeouts that were mentioned in https://github.com/trufflesuite/truffle/issues/3468
- updated object passed to HDWalletProvider to match the format mentioned in their doc.

We also switched `ETH_HOSTNAME` configuration from `https` to `wss`.

Migration was successful in execution https://github.com/keep-network/keep-core/pull/2352/checks?check_run_id=2005423441. The job failed though, but this was a post-migration step with Tenderly push that errored out.


nkuba:
Closing in favour of https://github.com/keep-network/keep-core/pull/2355

PR #2361, opened 2021-03-04
Token Distribution

nkuba:
This contract can be used to distribute ERC20 tokens with a merkle tree distribution mechanism in a cross-chain environment. 

An owner of the contract has to calculate a merkle tree for token assignments based on a mapping of recipient addresses with the amount of the assigned tokens. Next, the owner allocates the total amount of tokens in the contract by calling allocate function with the calculated merkle tree root. The owner has also a possibility to define a period after which unclaimed tokens will be allowed to recover by the owner. Once tokens allocation is made recipients can call the claim function to withdraw tokens to any destination address. The destination address should be signed by the recipient prior to calling the function and the signature should be provided on the claim.

The signing of destination addresses is implemented to allow cross-chain tokens distribution. It covers a situation when a recipient receives a tokens assignment based on their address on some chain (A) and operates under a different address
on the host chain (B) where tokens distribution takes place (e.g. a recipient is as a multi-sig wallet on chain A and receives tokens assignment on chain B, they have to sign an address on chain B they control with a signature calculated with a private key controlling the address on chain A).

The contract is based on the Uniswap's Merkle Distributor https://github.com/Uniswap/merkle-distributor with some modifications:
 - added 'allocate' function that will be called to allocate tokens for a merkle root,
- added a possibility for a recipient to redirect tokens withdrawal to another address by providing a signature over that address,
- added a possibility for the owner to recover unclaimed tokens after the unclaimed unlock duration that is configurable on tokens allocation.

## Destination address signing

When the allocation recipient gets tokens assigned they can redirect them to another address when claiming the tokens. To do that recipient has to provide an ethereum signature over the destination address and provide this signature to the claim function.

The signature has to be calculated over ethereum specific message format that should be obtained in the following way:
```
digest = keccak256(destinationAddress)
digestToSign = keccak256("\x19Ethereum Signed Message:\n" + len(digest) + digest)
```

When using web3.eth.sign function the ethereum prefix is added at the beginning automatically so one has to sign only the hash of the destination address, e.g.:
```js
web3.eth.accounts.sign(web3.utils.sha3(destinationAddress), privateKey)
```

## Unclaimed tokens recovery

Tokens that were not claimed within a recovery unlock period can be withdrawn by the owner of the contract. To allow the recovery the owner has to set unlock period duration on tokens allocation. If the parameter is provided with `0` value the unclaimed tokens won't be recoverable.


solidity/contracts/token-distribution/TokenDistributor.sol
```diff
@@ -0,0 +1,145 @@
...
+    event TokensAllocated(bytes32 merkleRoot, uint256 amount);
+
+    event TokensClaimed(
+        uint256 indexed index,
+        address indexed recipient,
+        address indexed destination,
+        uint256 amount
+    );
+
+    constructor(IERC20 _token) public {
```
dimpar, line 34: Not sure 100%, but if we pass an address of _token (cKEEP), wouldn't it simplify a deployment script a bit?

nkuba: I wanted this contract to be generic and usable with any ERC20.

dimpar: Sure, but I think if you pass an address `constructor(address _token)` it still stays generic. When we'll have a Token address (being it cKEEP or other token) we can just pass an address of the deployed token here, right? I guess it won't simplify things in this contract, but just asking about the deployment script whether it would be easier passing an address of already deployed token instead of initiating it first and then passing here. Not biggy, just wanted to ask.

solidity/test/token-distribution/TokenDistributorTest.js
```diff
@@ -0,0 +1,283 @@
...
+        ],
+      ])
+
+      for (const [testCaseName, testCaseData] of testCases) {
+        await createSnapshot()
+
+        console.log(`      ${testCaseName}`)
+        try {
+          const signature = web3.eth.accounts.sign(
+            web3.utils.sha3(destination),
```
dimpar, line 214: do you think we can use an alias of sha3 - `web3.utils.keccak256` to align these tests with the signature verification on smart contract?

nkuba: Good idea!

solidity/contracts/token-distribution/TokenDistributor.sol
```diff
@@ -0,0 +1,145 @@
...
+    ) internal pure returns (address) {
+        // Validate `s` and `v` values for a malleability concern described in EIP-2.
+        // Only signatures with `s` value in the lower half of the secp256k1
+        // curve's order and `v` value of 27 or 28 are considered valid.
+        require(
+            uint256(_s) <=
+                0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0,
+            "Invalid signature 's' value"
+        );
+        require(_v == 27 || _v == 28, "Invalid signature 'v' value");
```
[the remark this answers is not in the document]
nkuba: In Bonded ECDSA Keep we wanted to be compliant with ECDSA. The client produces signatures with recovery ID of 0,1 and then we convert it to Ethereum specific by adding 27. 

I would stay with 27, 28 here to be compliant with EIP-2, we expect the signature that is passed here to be an Ethereum signature.

solidity/package-lock.json
```diff
@@ -11427,6 +11427,27 @@
           "requires": {
             "has-flag": "^3.0.0"
           }
+        },
+        "ts-node": {
```
pdyraga, line 11431: Is this dependency needed now?

solidity/contracts/token-distribution/TokenDistributor.sol
```diff
@@ -0,0 +1,201 @@
...
+    event TokensAllocated(bytes32 merkleRoot, uint256 amount);
+
+    event TokensClaimed(
+        uint256 indexed index,
+        address indexed recipient,
+        address indexed destination,
+        uint256 amount
+    );
+
+    event TokensRecovered(address indexed destination, uint256 amount);
```
pdyraga, line 51: Do we need to index the destination? The recovery can happen only one time - the entire balance is transferred out and that's it.

solidity/contracts/token-distribution/TokenDistributor.sol
```diff
@@ -0,0 +1,201 @@
...
+                _unclaimedUnlockDuration;
+        }
+
+        emit TokensAllocated(_merkleRoot, _amount);
+    }
+
+    // Tokens not claimed within a given timeout should go to a treasury
+    // wallet address set on that contract.
+    // TODO: Is it fine to provide address on function call or do we want to declare
+    // it upfront on token deployment or allocation?
```
pdyraga, line 141: I think it is better to provide an address as a parameter to this function. More flexible and less gas-expensive.

solidity/helpers/merkle-distributor/dist/balance-tree.d.ts
```diff
@@ -0,0 +1,13 @@
+/// <reference types="node" />
```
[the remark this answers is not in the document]
nkuba: It's no longer used in tests. I removed the directory.

solidity/contracts/token-distribution/TokenDistributor.sol
nkuba: It doesn't hurt to index it, does it?

pdyraga: It is confusing, I was thinking `TokensRecovered` can be emitted multiple times and I was looking where is the catch and how is it possible.

solidity/package-lock.json
nkuba: It was there before, the order has just changed.

solidity/contracts/token-distribution/TokenDistributor.sol
```diff
@@ -0,0 +1,145 @@
...
+    function isClaimed(uint256 _index) public view returns (bool) {
+        uint256 claimedWordIndex = _index / 256;
+        uint256 claimedBitIndex = _index % 256;
+        uint256 claimedWord = claimedBitMap[claimedWordIndex];
+        uint256 mask = (1 << claimedBitIndex);
+        return claimedWord & mask == mask;
+    }
+
+    function recoverSignerAddress(
+        bytes memory _message,
```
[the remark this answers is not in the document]
pdyraga: Well... I don't like this name either but have no better idea and that's what is actually happening in the function so I think we are good with leaving it as-is 😆

solidity/test/token-distribution/TokenDistributorTest.js
```diff
@@ -0,0 +1,283 @@
...
+  describe("claim", async () => {
+    beforeEach(async () => {
+      await tokenDistributor.allocate(
+        testData.merkle.merkleRoot,
+        testData.merkle.tokenTotal,
+        { from: owner }
+      )
+    })
+
+    it("destination address signature verification", async function () {
```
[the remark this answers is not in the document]
nkuba: I'm not sure we're on the same page. This is a test that has a map of test data and it loops over these data. These are not single `it` tests.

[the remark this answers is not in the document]
nkuba: I'll remodel it to pull the common code to a function and execute test data in `it`s.

solidity/test/token-distribution/TokenDistributorTest.js
```diff
@@ -0,0 +1,651 @@
...
+        {
+          index: testData.merkle.claims[recipient].index.toString(),
+          recipient: recipient,
+          destination: claimDestination,
+          amount: toBN(testData.merkle.claims[recipient].amount),
+        }
+      )
+    })
+
+    describe("when verifying destination signature", async function () {
```
dimpar, line 297: Another way of testing an internal `recoverSignerAddress()` would be to create a contract stub with a public function that calls an internal one. It can make it more focused. But we can leave it as is, up to you.

solidity/test/token-distribution/TokenDistributorTest.js
```diff
@@ -0,0 +1,651 @@
...
+          signature.s,
+          testData.merkle.claims[recipient].index,
+          testData.merkle.claims[recipient].amount,
+          testData.merkle.claims[recipient].proof,
+          { from: submitter }
+        )
+
+        shouldRevert
+          ? await expectRevert(claimFuncCall, "invalid signature")
+          : await claimFuncCall
```
dimpar, line 322: I think it is better to expect something, maybe an emitted event at the end to make sure it was a successful call?

dimpar:
Contract looks good to me, comments above are not blockers.

solidity/contracts/token-distribution/TokenDistributor.sol
```diff
@@ -0,0 +1,251 @@
...
+/// a possibility to define a period after which unclaimed tokens will be allowed
+/// to recover by the owner. Once tokens allocation is made recipients can call
+/// claim function to withdraw tokens to any destination address. The destination
+/// address should be signed by the recipient prior to calling the function and
+/// the signature should be provided on claim. For details of the signature see
+/// claim function documentation.
+/// Signing of destination addresses is implemented to allow cross-chain tokens
+/// distribution. It covers a situation when recipient receives a tokens assignment
+/// based on their address on some chain (A), and operates under a different address
+/// on the host chain (B) where tokens distribution takes place (e.g. a recipient
```
pdyraga, line 38: I'd drop `host` from this sentence. If anything is host, it is the first chain based on which tokens are allocated on the other chain.

solidity/contracts/token-distribution/TokenDistributor.sol
```diff
@@ -0,0 +1,251 @@
...
+contract TokenDistributor is Ownable {
+    using SafeERC20 for IERC20;
+
+    // Distributed token.
+    IERC20 public token;
+
+    // Merkle tree root.
+    bytes32 public merkleRoot;
+
+    // Timestamp after which allocated tokens can be recovered from the contract
```
pdyraga, line 59: `...allocated and unclaimed...`?

solidity/contracts/token-distribution/TokenDistributor.sol
```diff
@@ -0,0 +1,251 @@
...
+    /// for a transfer to be provided. The destination address should be signed
+    /// by the recipient. To construct the message to signing recipient has to
+    /// hash destination address with keccak256 and prefix the obtained digest with
+    /// Ethereum specific `\x19Ethereum Signed Message:\n32`. When using web3's
+    /// signing function the prefixing is done automatically, so the signing operation
+    /// can look like: `sign(keccak256(destination), privateKey)`.
+    /// The signature should be provided with the call to this function. This is
+    /// to confirm that the recipient approved the destination address for a transfer.
+    /// The solution allows cross-chain allocations, where a recipient from a
+    /// different chain does not operate under the same address on the host chain
```
pdyraga, line 93: Here also I'd drop `host` and say just `... on the chain where tokens were allocated ...`

solidity/test/token-distribution/TokenDistributorTest.js
```diff
@@ -0,0 +1,651 @@
...
+    before(async () => {
+      await tokenDistributor.allocate(
+        testData.merkle.merkleRoot,
+        testData.merkle.tokenTotal,
+        unclaimedUnlockDuration,
+        { from: owner }
+      )
+    })
+
+    after(freshDeployment)
```
pdyraga, line 241: This is confusing. I'd rather drop it and changed `before` into `beforeEach`. WDYT?

solidity/contracts/token-distribution/TokenDistributor.sol
```diff
@@ -0,0 +1,251 @@
...
+
+        token.safeTransfer(_destination, _amount);
+
+        emit TokensClaimed(_index, _recipient, _destination, _amount);
+    }
+
+    /// Allocates amount of tokens for the merkle root.
+    /// @param _merkleRoot The merkle root.
+    /// @param _amount The amount of tokens allocated for the merkle root.
+    /// @param _unclaimedUnlockDuration Duration of a period after which unclaimed
```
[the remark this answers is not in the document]
nkuba: I was thinking about that. I'm glad you mentioned it as it makes two of us. Will do!

solidity/test/token-distribution/TokenDistributorTest.js
nkuba: Yep, this could be an alternative way to test it, we use it quite often for internal functions tests isolation. Here I decided to not add another stub contract as the current way of testing doesn't hurt that much in my opinion and we can easily test `recoverSignerAddress` results with the exposed function.

solidity/test/token-distribution/TokenDistributorTest.js
nkuba: Yeah, `beforeEach` will do the work!

dimpar (approved):

pdyraga (approved):

PR #2363, opened 2021-03-05
Migrate contracts job for Celo

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-core/pull/2376

Added a `migrate-and-publish-celo` job which performs contract migration on Celo chain. For now it publishes artifacts to GCS bucket only. Publishing to Tenderly and NPM should be added later.


.github/workflows/contracts.yml
```diff
@@ -150,3 +150,40 @@ jobs:
...
+        with:
+          project_id: ${{ secrets.GOOGLE_PROJECT_ID }}
+          service_account_key: ${{ secrets.KEEP_TEST_GCR_JSON_KEY }}
+
+      - name: Upload contract data
+        env:
+          CONTRACT_DATA_BUCKET: ${{ secrets.KEEP_TEST_CONTRACT_DATA_BUCKET }}
+        run: |
+          cd build/contracts
+          gsutil -m cp * gs://"$CONTRACT_DATA_BUCKET"/keep-core-celo
```
nkuba, line 187: As we name the directory here `keep-core-celo`, should we rename it for ethereum to `keep-core-ethereum`?

lukasz-zimnoch: I thought about it but decided to leave it in order to maintain backward compatibility. There are several CI jobs that read from that directory and renaming this will require changing all places and moving the files on GCS bucket. I prefer limiting the number of changes in existing jobs and focus on Celo so I think we should leave it as is. After all, it's a quite common assumption that resources without the chain prefix/suffix refer to the default ethereum host chain.

.github/workflows/contracts.yml
```diff
@@ -189,3 +189,62 @@ jobs:
...
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Migrate contracts
+        env:
+          CONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY: |
+            ${{ secrets.KEEP_TEST_CELO_CONTRACT_OWNER_PRIVATE_KEY }}
```
michalinacienciala, line 233: @nkuba: In `migrate-and-publish-ethereum` job we use this variable:
`CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY: ${{ secrets.CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY }}`
Maybe it would be good to change its name to `KEEP_TEST_ETH_CONTRACT_OWNER_PRIVATE_KEY` to make it  similar to what we use here for Celo?
We could change that as part of https://github.com/keep-network/keep-core/pull/2376 or leave it for later and change it with other variables that also may need some name changes (like adding `KEEP_TEST_` prefix).

nkuba: Yep @michalinacienciala, that was something I was thinking about when adding the celo secret.

.github/workflows/contracts.yml
```diff
@@ -189,3 +189,62 @@ jobs:
...
+        run: npx truffle migrate --reset --network alfajores
```
nkuba, line 234: We should probably pull `alfajores` to an env property, but I have no good idea how to handle it best together with ethereum, so the current code is fine with me.

.github/workflows/contracts.yml
```diff
@@ -189,3 +189,62 @@ jobs:
...
+
+      # TODO: if Celo works with Tenderly, add appropriate push step
```
nkuba, line 236: I don't think it's supported by Tenderly.

.github/workflows/contracts.yml
nkuba: Sure thing, I wasn't even thinking about changing it in this PR. Let's just keep it in mind for the future.

nkuba (commented):
What we've got currently looks fine to me. I'd just like to see a passing job after we're done with the configuration.

.github/workflows/contracts.yml
michalinacienciala: `KEEP_TEST_` prefix has been added now to several variables:
```KEEP_TEST_ETH_CONTRACT_OWNER_PRIVATE_KEY
KEEP_TEST_ETH_HOSTNAME
KEEP_TEST_ETH_NETWORK_ID
KEEP_TEST_ETH_TRUFFLE_NETWORK```
(changes done both in GitHub settings and in `contracts.yml`).

.github/workflows/contracts.yml
lukasz-zimnoch: Yeah definitely, I can leave a TODO about that.

nkuba: I pulled `alfajores` to a secret `KEEP_TEST_CELO_TRUFFLE_NETWORK`. It's consistent with how we handle it for ethereum.

lukasz-zimnoch: See https://github.com/keep-network/keep-core/commit/946d20a56bb31b93d968b6cfb2f9103d35dad665

.github/workflows/contracts.yml
lukasz-zimnoch: I removed the TODO then: https://github.com/keep-network/keep-core/commit/20a3ff1df46d84699c33ecaabe0a4d21e52ba892

lukasz-zimnoch:
@nkuba everything addressed!

michalinacienciala (commented):
Looks good to me. Leaving for @nkuba to review as well.

nkuba (approved):

Issue #2339, opened 2021-02-18
Visual inaccuracy after delegating liquid tokens


daramir:
### In: 
https://dashboard.keep.network/delegations/wallet

### When + What: 
Delegating a liquid stake, after the transaction was confirmed, the wallet balance shows negative tokens and the Tokens staked donut chart goes over 100%.
### Screenshot
![Screenshot from 2021-02-19 00-46-14](https://user-images.githubusercontent.com/4935829/108432819-130b4300-7299-11eb-9789-177143f8d444.png)

### Additional details
I had a balance that wasn't rounded to the thousands and used the "Max Stake" button, aiming to use all my balance for this stake. My balance was close to the min stake, but I still don't know how the dashboard derived this number since "my balance" modulo "10000" was not the number in the picture. 

After logging back in, the issue was gone and page reflected the actual (positive) balance of some decimals left behind.

I hope that helps to reproduce the issue.

PR #2375, opened 2021-03-10
Fix the problem with displaying KEEP tokens in the Overview page

michalsmiarowski:
Closes: #2339 

In some cases Wallet Balance in the **Overview** page was not displayed properly.

<details>
<summary>Recreating the issue</summary>

I managed to recreate the issue on Ropsten. To do that you must have 2 accounts.

1. Log in on account 1 on Google Chrome
2. Log in on account 2 on Firefox
3. Delegate some tokens to another account with account 1
4. Create token grant for account 1 using account 2
5. After a while the wallet balance of the account #1 will change even though we didn't withdraw the tokens from the grant yet. The issue disappear after refreshing the page.
</details>

There was a problem with `arithmeticOperation` variable in `subscriptions.js` file. We define it in a while loop without setting a value and assign some value later in the if statement. The problem is that this value persists in further iterations of the while loop. I fixed it by assigning a null value to it from the start.

Still, it might be a good idea to do some investigation into the code, because unfortunately i couldn't find the reason why the previous code worked in such an unexpected way.

### DIFF
--- a/solidity/dashboard/src/sagas/subscriptions.js
+++ b/solidity/dashboard/src/sagas/subscriptions.js
@@ -934,4 +934,3 @@ export function* subscribeToLiquidityRewardsEvents() {
     )
   }
 }
-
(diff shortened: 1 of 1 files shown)


r-czajkowski (approved):
This fixes described bug👍 but tbh I don't understand why the variable refers to value from the previous iteration 🤷.

PR #2376, opened 2021-03-10
Move migration and publishing of contracts (Ethereum) to GH Actions

michalinacienciala:
and tBTC release processes in order to reduce human involvement and
migration from CircleCI jobs to GitHub Actions. This commit enhances
the workflow for building and testing Solidity with a job migrating
contracts and publishing them to the npm registry. Contracts also get
pushed to Tenderly and uploaded to Google Cloud Storage. For now, only
keep-test configuration is implemented. Implementation for other
environments will be done at a later stage of work on RFC-18.

The `migrate-and-publish-ethereum` job is configured to execute only if
workflow is started by a push or workflow_dispatch event on `master` or
`rfc-18/*` branch. The `rfc-18/*` condition will need to be removed
once whole `build-test-migrate-publish-keep-test` CircleCI process
gets migrated to GH Actions.
Additionally, as we may not always want to migrate and publish
contracts on every push (or workflow_dispatch) on `master`, the job was
configured to require manual approval in GitHub GUI. This is done by
configuration of environment protection rules on `keep-test` and using
`environment: keep-test` config in the workflow.

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -12,23 +12,32 @@ on:
       - ".github/workflows/contracts.yml"
   pull_request:
     branches:
-      # TODO: Run on all branches or only on master (to be decided) 
-      # after we're fully migrated from Circle CI
+      # TODO: Run on all branches after we're fully migrated from Circle CI
       - "rfc-18/**"
       - master
     paths:
       - "solidity/**"
       - "!solidity/dashboard/**"
       - ".github/workflows/contracts.yml"
+  workflow_dispatch:
 
 jobs:
   build-and-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
+    defaults:
+      run:
+        working-directory: ./solidity
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
         with:
-          node-version: "12.x"
+          node-version: ${{ matrix.node-version }}
+
       - name: Cache node modules
         uses: actions/cache@v2
         env:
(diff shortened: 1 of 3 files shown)


michalinacienciala:
[Here](https://github.com/keep-network/keep-core/runs/2069777538?check_suite_focus=true) is an example of passing `migrate-and-publish-ethereum` job (on different branch, but containing the changes introduced in this PR). In that example `Push contracts to Tenderly` optional step is failing. Since then there were changes to the `keep-network/tenderly-push-action` action - we need to check if the step is passing after those changes.

michalinacienciala:
@lukasz-zimnoch, @nkuba - migration and push to Tenderly are finally working (see [run #650](https://github.com/keep-network/keep-core/actions/runs/642912058)) :tada: You can review and merge if no changes needed.

lukasz-zimnoch (approved):
Looks good to me! Leaving final approval for @nkuba

.github/workflows/contracts.yml
```diff
@@ -67,9 +86,106 @@ jobs:
...
+
+      # TODO: consider swithing to below action, after tweaking 
+      # (requires some changes, as below config throws an error)
+      # - name: Publish to npm
+      #   uses: JS-DevTools/npm-publish@v1
+      #   with:
+      #     package: ./solidity/package.json
+      #     token: ${{ secrets.NPM_TOKEN }}
+      #     access: public
+      #     dry-run: true # TODO: to be removed once testing of workflow is done
```
nkuba, line 191: We can remove this whole part, we will stick with the current one for now.

nkuba (approved):
Good to merge. The comments I left can be handled in another PR.

michalinacienciala:
> Good to merge. The comments I left can be handled in another PR.

Thanks for the input :) I applied the proposed changes as part of this cleanup PR: https://github.com/keep-network/keep-core/pull/2378.

PR #2378, opened 2021-03-11
Remove unnecessary configs after moving to GH Actions

michalinacienciala:
and tBTC release processes in order to reduce human involvement and
migration from CircleCI jobs to GitHub Actions.

The GH Actions contracts.yml workflow has been verified to successfully
build, test, deploy and upload contracts. This means that we can
finalize moving from CircleCI to GH Actions for those operations. Due
to that we are removing no longer necessary
`build-test-migrate-publish-keep-test` workflow from `config.yml` file,
together with jobs it was consisting of. We also no longer need to
perform `dry-run` when publishing to npm registry, or use `-wip` suffix
for names of uploaded docker images. Conditions for Solidity workflow
triggering also can be now changed to final ones.

This PR also removes the keep-dev related workflows and jobs from
CircleCI config, as it was decided that for now keep-test will be the main
supported env (we may add support for other envs at later stage).


michalinacienciala:
Before merging the changes GitHub config needs to be changed to no longer require passing of `ci/circleci: build_client_and_test_go` job. @nkuba could you change that?

nkuba (approved):

Issue #2185, opened 2020-12-02
Adjusting contrast of minimum stake and staking amount available


liz-shinn:
Contrast issues currently with the minimum stake amount and the staking amount available, two key pieces of information for staking. 

- [x] Adjust design to remedy font sizes and color contrast issues.

- [x] Minimum staking schedule indication in tooltip

![unnamed-1](https://user-images.githubusercontent.com/57226633/100896778-fe36c480-348c-11eb-81ed-8228649e74d6.png)

liz-shinn:
Adjusted the design to be the following:

<img width="746" alt="Screen Shot 2020-12-07 at 2 31 22 PM" src="https://user-images.githubusercontent.com/57226633/101396133-e4382000-3898-11eb-8286-e43ffdbbcb83.png">

Battenfield:
One option would be to show the user that the min stake is about to decrease to X amount in Y amount of days. This could be in the tooltip or on the delegation page.

Battenfield:
https://github.com/orgs/keep-network/projects/30#card-56281528

PR #2379, opened 2021-03-11
Staking keep token form redesign

michalsmiarowski:
Closes: #2185 

This PR makes some UI changes to the staking tokens input in Overview page.

Main changes:
- remove progress bar with available KEEP tokens below the input
- fix the ui problem with higher number of KEEP tokens written inside input that would cover the Max Stake button
- add a tooltip that will inform user that the min stake is about to decrease to 50,000 soon
- change the order of the Wallet Tokens page and Granted Tokens page on the top menu of Delegations page
- some UI changes based on [Figma](https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=8642%3A13364)

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

### DIFF
--- a/solidity/dashboard/src/components/DelegateStakeForm.jsx
+++ b/solidity/dashboard/src/components/DelegateStakeForm.jsx
@@ -131,8 +131,8 @@ const TokensAmountField = ({
             <MaxAmountAddon onClick={onAddonClick} text="Max Stake" />
           }
         />
-        <div className="text-caption text-grey-60 text-right ml-a">
-          {displayAmount(availableToStake)} KEEP available
+        <div className="text-caption--green-theme text-right ml-a">
+          {displayAmount(availableToStake)} available to stake
         </div>
       </div>
     </div>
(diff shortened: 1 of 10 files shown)


solidity/dashboard/src/components/DelegateStakeForm.jsx
```diff
@@ -123,27 +131,21 @@ const TokensAmountField = ({
...
-            minStake
-          )} KEEP`}
+          additionalInfoText={`MIN STAKE ${displayAmount(minStake)} KEEP`}
           leftIcon={<Icons.KeepOutline className="keep-outline--mint-100" />}
           inputAddon={
             <MaxAmountAddon onClick={onAddonClick} text="Max Stake" />
           }
+          tooltipText={
+            <>
+              The minimum stake will decrease to 50,000 KEEP on 04/28/2021. You
```
r-czajkowski, line 141: Can we create a kind of helper that returns the next minimum stake value and date according to the https://staking.keep.network/about-staking/staking-minimums ? Otherwise, we will have to update manually each time.

r-czajkowski (changes requested):
Looking good just left a tiny comment. Also could you please run `js lint` script to fix errors? I have no idea why that [pre-commit hook](https://github.com/keep-network/keep-core/blob/master/.pre-commit-config.yaml#L26-L31) doesn't work.

r-czajkowski (approved):

PR #2382, opened 2021-03-15
Add "incentives removed" warning to the KEEP-TBTC pool

michalsmiarowski:
Adds a warning to the KEEP-TBTC pool that informs the users that incentives for
this pool has been removed. Also blocks the possibility to add lp tokens to
KEEP-TBTC pool.

Also removes APY and % of the share pool from KEEP-TBTC pool.

image


Battenfield (approved):
LGTM

Issue #2328, opened 2021-02-11
Display your boost in the KEEP-only pool


mhluongo:
The KEEP pool shows your deposited tokens and rewards, but it doesn't show your boost. Withdrawing KEEP before you hit the max 3x boost means leaving money on the table, so it would be good for users to be able to see that and maybe get a warning before they withdraw... otherwise people will try to "churn" tokens in the pool to make more money, and not realize that they're actually losing out.

The KEEP-only pool contract is based on Ampleforth's geyser. [In their interface](https://www.ampleforth.org/geyser/beehiveV3-uniswap-weth-ampl), they include the boost / multiplier as well as an info icon to explain how long it takes to hit the max boost.

![Screenshot from 2021-02-11 09-31-48](https://user-images.githubusercontent.com/427505/107650012-0a3cc080-6c4c-11eb-8db1-96e9edfaa380.png)

Thoughts @liz-shinn @nkuba @r-czajkowski?

mhluongo:
This would address people's questions about "auto-compounding" @willthesis @nkuba — where the question behind the question is "should I withdraw rewards and re-deposit?" and the answer is likely no

r-czajkowski:
Yeah sounds good. Maybe we should display `rewards multiplier` [here](https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=8175%3A14003) instead of `your KEEP rewards` because this metric is duplicated(see screenshot below)

![obraz](https://user-images.githubusercontent.com/57687279/107651779-3114ea00-6c80-11eb-8b44-1760f64bb21f.png)

liz-shinn:
@michalsmiarowski Does this look like what you were describing on the call? Wanted to update the Figma design to match.

<img width="646" alt="Screen Shot 2021-03-29 at 11 14 48 AM" src="https://user-images.githubusercontent.com/57226633/112858647-11931e80-9080-11eb-9aeb-d5ff6aea2dfd.png">

liz-shinn:
Ah, I see "rewards multiplier" is what we're calling it. Will update.

liz-shinn:
<img width="753" alt="Screen Shot 2021-03-29 at 11 19 11 AM" src="https://user-images.githubusercontent.com/57226633/112859208-9a11bf00-9080-11eb-92a7-c56faafc44d8.png">

PR #2386, opened 2021-03-18
Display boost for KEEP-only pool

michalsmiarowski:
Closes: #2328 

Displays the reward boost for he KEEP ONLY pool:

image

### DIFF
--- a/solidity/dashboard/src/components/liquidity/RewardMultiplier.jsx
+++ b/solidity/dashboard/src/components/liquidity/RewardMultiplier.jsx
@@ -25,7 +25,13 @@ export const RewardMultiplier = ({
 
 RewardMultiplier.TooltipContent = () => {
   // TODO: content of the tooltip
-  return <>Reward multiplier tooltip</>
+  return (
+    <>
+      Deposit liquidity tokens for 30 days to achieve a 3x reward multiplier.
+      The multiplier is individual for each stake, so this value is an average
+      of the boosts of all your stakes.{" "}
+    </>
+  )
 }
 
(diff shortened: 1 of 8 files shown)


solidity/dashboard/src/sagas/liquidity-rewards.js
```diff
@@ -89,6 +89,11 @@ function* fetchLiquidityRewardsData(liquidityRewardPair, address) {
       )
     }
 
+    const rewardMultiplier = yield call(
```
r-czajkowski, line 92: We probably should update `rewardMultiplier` when the user has withdrawn or deposited more tokens.

solidity/dashboard/src/components/liquidity/RewardMultiplier.jsx
```diff
@@ -0,0 +1,31 @@
...
+        duration={1}
+        suffix={"x"}
+      />
+    </h2>
+  )
+}
+
+RewardMultiplier.TooltipContent = () => {
+  // TODO: content of the tooltip
+  return <>Reward multiplier tooltip</>
```
r-czajkowski, line 28: I realize that we don't have wireframes for this but maybe: `Deposit liquidity tokens for 30 days to achieve a 3x reward multiplier.` + something about how we calculate it? I mean we take into account all stakes. What do you think?

r-czajkowski (changes requested):

michalsmiarowski:
Ready for another round of review @r-czajkowski

r-czajkowski (approved):
LGTM!

PR #2387, opened 2021-03-19
go-ethereum dependency update to v1.10.1

pdyraga:
`ethereum/go-ethereum` dependency has been updated to the most recent version, `v1.10.1`. The generated Go contract code has been regenerated after some adjustments done in keep-network/keep-common#73 to support the new go-ethereum version. Some small adjustments were needed, mostly about EIP155-comatible signer, earlier merged into `keep-common` in https://github.com/keep-network/keep-common/pull/72


pdyraga:
Tested by doing genesis and requesting a couple of relay requests.

lukasz-zimnoch (approved):

PR #2391, opened 2021-03-22
Fix tests that fail occasionally with values close to expected

nkuba:
Some tests are failing occasionally due to a value not exactly as expected. There is a time factor so the value may slightly diverge from expectations. With these updates, we're checking if values are close to the expected ones within a specific range.

Sample CI execution that had failures: 
https://github.com/keep-network/keep-core/runs/2113275081
https://github.com/keep-network/keep-core/runs/2352840657

Refs https://github.com/keep-network/keep-core/pull/2354.

### DIFF
--- a/solidity/test/geyser/TestKeepTokenGeyser.js
+++ b/solidity/test/geyser/TestKeepTokenGeyser.js
@@ -371,24 +371,29 @@ describe("KeepTokenGeyser", async () => {
       })
 
       // Validate stakers' stake token balances.
-      expect(
+      expectCloseTo(
         await stakeToken.balanceOf.call(staker1),
+        stakerInitialBalance1,
         "invalid staker's 1 token balance"
-      ).to.eq.BN(stakerInitialBalance1)
-      expect(
+      )
+      expectCloseTo(
         await stakeToken.balanceOf.call(staker2),
+        stakerInitialBalance2,
         "invalid staker's 2 token balance"
-      ).to.eq.BN(stakerInitialBalance2)
+      )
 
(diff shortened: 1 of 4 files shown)


solidity/test/geyser/TestKeepTokenGeyser.js
```diff
@@ -371,24 +369,29 @@ describe("KeepTokenGeyser", async () => {
       })
 
       // Validate stakers' stake token balances.
-      expect(
+      expectCloseTo(
```
pdyraga, line 372: Would it be possible to use `expect(...).to.be.closeTo(...)`, [here](https://github.com/keep-network/coverage-pools/blob/140403c6cc9bf3f93e39f4894075f09f28536cae/test/AssetPoolTest.js#L150-L155) is an example.

https://www.chaijs.com/api/bdd/#method_closeto

pdyraga (approved):
I'd prefer to eventually use `.to.be.closeTo` but given that we extracted already existing code, I do not consider this a blocker.

solidity/test/geyser/TestKeepTokenGeyser.js
nkuba: `expect(...).to.be.closeTo(...)` expects params of type `Number`, here we've got `BN`.

When I pass the values as `BN` I get: `AssertionError: expected <BN: 1fc3842bd1f071c00000> to be a number`.
When I pass the values as `number` I get: `Error: Number can only safely store up to 53 bits`.

I'm not sure it's possible with truffle.

PR #2398, opened 2021-03-26
Wallet improvements dapp

### LINKED ISSUE #2081: MetaMask disconnects when you refresh the page
Discord user review: 
>When refreshing or clicking the logo on the top left it automatically signs out and requires to re-select signin method. I would prefer to be kept signed in and have a sign-out option.

### LINKED ISSUE #2091: Ability to disconnect / sign out of wallet
How to disconnect from a wallet (that's not refreshing your browser.)

r-czajkowski:
~Depends on: #2403~
Close #2091 
Close #2081 

Thie PR improves integration with the wallets in the KEEP token dashboard.

Main changes:
* ability to disconnect wallet;
* use an auto-connect pattern for MetaMask like in other popular dapps. The auto-connect pattern is about when a user has MetaMask extension installed and is still connected to the dapp site, then the dapp uses the MetaMask extension as a default sign-in option. The user is able to disconnect a site for a given account directly in the MetaMask extension (Metamask doesn't provide a function to disconnect a MetaMask via their API). If the user disconnects account or log-out in the MetaMask extension the patter will not work and a user will have to select sign-in option;
* ability to switch account and connector;
* also this PR is an introduction to writing test in the dapp. In follow-up work, we are going to define GitHub actions to auto-run tests on every PR.


solidity/dashboard/src/store.js
```diff
@@ -2,10 +2,13 @@ import { createStore, applyMiddleware } from "redux"
 import createSagaMiddleware from "redux-saga"
 import rootReducer from "./reducers"
 import rootSaga from "./sagas"
-
+import { composeWithDevTools } from "redux-devtools-extension"
```
michalsmiarowski, line 5: I think we don't need the redux devtools on prod so maybe we could do
```
import { composeWithDevTools } from "redux-devtools-extension/developmentOnly"
```

as stated [here](https://www.npmjs.com/package/redux-devtools-extension)?

solidity/dashboard/jsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+    "typeAcquisition": {
+        "include": ["jest"]
+    }
+    
+}
```
michalsmiarowski, line 6: New line is missing there. You can also check if there are any other js lint errors just in case you didn't 😉

michalsmiarowski (changes requested):
Add some comments but overall very good 👍

solidity/dashboard/src/App.js
```diff
@@ -26,6 +27,32 @@ const App = () => (
...
+      connector.once("disconnect", disconnectHandler)
+    }
+
+    return () => {
+      if (connector) {
+        connector.removeListener("accountsChanged", accountChangedHandler)
+        connector.removeListener("disconnect", disconnectHandler)
+      }
+    }
+  }, [isConnected, connector, dispatch, yourAddress])
```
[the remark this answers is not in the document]
r-czajkowski: The main idea was to avoid connecting the `Web3Context` with redux. The `Web3Context` can exist without third-party components and can handle events in specific ways that only make sense inside `Web3Context`. One of the disadvantages of the proposed solution is that if, eg., we put `Web3Context` over the Redux provider in the app tree, `Web3Context` will not work.

michalsmiarowski (approved):
Looking good! Good job with connector switching fix 🚀

PR #2403, opened 2021-03-30
Support accounts switching

r-czajkowski:
Ref: #2398

This PR adds support for switching accounts on all pages in the dapp. Here we moved fetching data for `Operations`, `Applications/RandomBeacon` pages to Redux. For Other pages (`Applications/tBTC`, `Earnings`) to avoid migrate the whole app to a Redux in a single PR, we decided to use `useFetchData` custom hook with few modifications to handle re-fetching data if the address has changed.


solidity/dashboard/src/actions/index.js
```diff
@@ -11,3 +11,35 @@ export const RESET_COPY_STAKE_FLOW = "copy-stake/reset_flow"
...
+export const FETCH_OPERATOR_SLASHED_TOKENS_RERQUEST =
+  "operator/fetch_slashed_tokens_request"
+export const FETCH_OPERATOR_SLASHED_TOKENS_START =
+  "operator/fetch_slashed_tokens_start"
+export const FETCH_OPERATOR_SLASHED_TOKENS_SUCCESS =
+  "operator/fetch_slashed_tokens_success"
+export const FETCH_OPERATOR_SLASHED_TOKENS_FAILURE =
+  "operator/fetch_slashed_tokens_failure"
+export const OPERATOR_DELEGATION_UNDELEGATED = "operator/delegation_undelegated"
+export const OPERATR_DELEGATION_CANCELED = "operator/delegation_canceled"
```
michalsmiarowski, line 33: Just a note: I think it would be more readable if we keep all action names in the same tense 🤔 As far as i see we usually use past simple so maybe it would be better to change it to _REQUESTED, _STARTED, _SUCCESS?

Although if I can be honest I would prefer to keep it all in present simple 😅 

EDIT: Actually I see that we used present simple in other places also for START, SUCCESS and FAILURE, but still maybe we could refactor and use only one tense in the near future?

michalsmiarowski (approved):
Looking good. Haven't found any issues 🚀

michalsmiarowski:
I will merge it to `wallet-improvements-dapp` so any issues related to this will be resolved on the `wallet-improvements-dapp` branch 😉

Issue #2024, opened 2020-08-31
Explorer mode / Read-only address


liz-shinn:
Ability to view the Keep Token Dapp without needing to connect via metamask/ledger/etc.

Additionally, you can personify a beneficiary and see your rewards, allows for modality.

Benefits of an "explorer" mode:
1. Having that capability evens the playing field. anyone can view anyone's grant info w/ an address – transparency
2. For demo purposes and for trouble shooting / help
3. Cold storage audience who maintains large token grants they may want to view it without connecting a cold wallet
4. Multi sig actually possible (can enter a multi sig address). Can't currently do anything in the UI and has to be all on command line

liz-shinn:
Check out Zapper.fi for example & Zerion

mhluongo:
Plus :100:

This + [WalletConnect](https://github.com/keep-network/keep-core/pull/2340) and we'll be golden on the access side. Note that I believe WalletConnect supports the Gnosis Safe, so adding instructions around that would also improve the lives of us multisig users

PR #2405, opened 2021-03-31
Explorer mode

michalsmiarowski:
Closes: #2024

Adds an ability to explore the site using any wallet address.

**Main changes:**

- adds Explorer Mode
- adds icon of the wallet on the top-right of the page
- adds the ability to log in to the wallet and send a transaction that you executed while being in explorer mode (you have to connect to the same account that you have used in explorer mode)
- adds the ability to go into explorer mode by pasting the url with the correct address in it (example pattern: `https://dashboard.keep.network/<address>/delegations/granted`) 

**Side notes:**

- creates a custom subprovider for Explorer Mode that catches the 'eth_sendTransaction' method and emit the payload of it to the ui to save it in redux store
- creates a transaction queue in the store where we are saving the payloads of the 'eth_sendTransaction' methods are added while being in explorer mode. Then these methods will be executed when user connect to the wallet.
- background of address in the top-right corner is greyed out to emphasise clearly that user is in Explorer Mode
- **[DESIGN NEEDED?]** We probably need to think about some different icon for the explorer mode. Right now i used the wallet icon, but i think it does not match it. Maybe it will be better to add like svg icon of the eye since it's the read-only mode? 🤔 Please see screenshots below.

**Applied UI solutions:**

<details>
  <summary>Wallet icons of the connected wallet on the top-right corner of the page</summary>

**MetaMask**:
image

image

**Explorer Mode:**

image

image

</details>

<details>
  <summary>Connect wallet dropdown</summary>

image

</details>

<details>
  <summary>Explorer mode modal window</summary>

This windows pops up after choosing `ExplorerMode` in the connect wallet dropdown.

image

</details>

<details>
  <summary>Connect to wallet modal window</summary>

This windows pops up after trying to do a transaction while being in Explorer Mode

image

Hovering on `Connect Wallet` button:

image

</details>


solidity/dashboard/src/connectors/explorerModeSubprovider.js
```diff
@@ -0,0 +1,22 @@
...
+class ExplorerModeSubprovider extends Subprovider {
+  constructor(_eventEmitter) {
+    super()
+    this.eventEmitter = _eventEmitter
+  }
+  async handleRequest(payload, next, end) {
+    switch (payload.method) {
+      case "eth_sendTransaction":
+        this.eventEmitter.emit("chooseWallet", payload)
+        next()
```
r-czajkowski, line 13: It should be `end()`, shouldn't it? This type of eth method has been already handled by emitting an event, so we don't want to pass it to the next subprovider.

solidity/dashboard/src/connectors/explorerModeSubprovider.js
```diff
@@ -0,0 +1,22 @@
+import { Subprovider } from "@0x/subproviders";
+
+
+class ExplorerModeSubprovider extends Subprovider {
```
r-czajkowski, line 4: I would consider inheriting from `EventEmitter` then we can get rid of `eventEmitter` as a field in the class. We would have to do `this.explorerSubproivder.eventEmitter.on()` to handle subprovider events which is a little weird and could be misleading. Also not sure if we really need `Subprovider` from `@0x/subproviders`.

solidity/dashboard/src/connectors/read-only-connector.js
```diff
@@ -0,0 +1,40 @@
...
+import ExplorerModeSubprovider from "./explorerModeSubprovider";
+import { EventEmitter } from "events"
+
+export class ReadOnlyConnector extends AbstractConnector {
+  /** @type {string} To store selected account by user. */
+  selectedAccount = ""
+
+  constructor() {
+    super(WALLETS.READ_ONLY_ADDRESS.name)
+    this.websocketSubprovider = new WebsocketSubprovider({ rpcUrl: getWsUrl() })
```
r-czajkowski, line 15: I think we should create `WebsocketSubprovider` in the `enable` method. Otherwise, it creates an unnecessary connection to the node after creating the `ReadOnlyConnector` object.

r-czajkowski (changes requested):
Left my first thoughts. Custom subprovider looks like a really interesting solution 👍.

michalsmiarowski:
@liz-shinn You can check the top post for the UI solutions that i applied in this task. Feel free to propose some improvements 😃

solidity/dashboard/src/connectors/explorerModeSubprovider.js
michalsmiarowski: Yeah i used it like that mainly because I wanted the `ExplorerModeSubprovider` to inherit from `@0x/subprovider` which seems like a good idea, because it  make sure that it forces us to have the `handleRequest` method defined 🤔

liz-shinn:
Awesome! Have a few thoughts:

>Wallet icons of the connected wallet on the top-right corner of the page

I like the solution of adding the wallet icon into the header when you have a particular wallet connected!
I think that the Explorer Mode icon could look more like "explore" though – perhaps it's a 🚀 or a 🌐 , potentially 🔍 but that could look like search so maybe not. I lean towards 🌐 

> Connect wallet dropdown

Would add a space so it's "Explorer Mode" instead of explorermode, and swap out the icon for whatever the more "explore" icon is from above.

> Explorer mode modal window

The modal title would make more sense as "Connect Ethereum Address" instead of Connect Wallet.

> Connect to wallet modal window

Modal title would be "Select Wallet" with a capital "W" to be consistent since we use title case in other modal instances
Would alter the modal text to read: "You're viewing the Dashboard in Explorer Mode. Connect a wallet to proceed."

liz-shinn:
Screen Shot 2021-04-06 at 5 13 31 PM
 
This line on the modals should be above the button as a footer indication:
Screen Shot 2021-04-06 at 5 14 46 PM

Should also have a cancel option next to the action button

solidity/dashboard/src/components/WalletSelectionModal.jsx
```diff
@@ -0,0 +1,33 @@
...
+            triggerComponent={() => (
+              <span
+                className={`btn btn-primary btn-lg empty-state__connect-wallet-btn`}
+              >
+                Connect wallet
+              </span>
+            )}
+          >
+            <WalletOptions displayExplorerMode={false} />
+          </Tooltip>
```
r-czajkowski, line 28: I would consider extracting this to a separate component eg. `ConnectWalletBtn`. The same component is under `EmptyState.ConnectWalletBtn`.

solidity/dashboard/src/css/web3-status.less
```diff
@@ -5,6 +5,13 @@
         align-items: center;
     }
 
+    &__wallet-connected-icon-wrapper {
+        display: flex;
+        margin-left: 1rem;
+    }
+
+    &__wallet-connected-tooltip {}
```
r-czajkowski, line 13: Hmm do we really need this? If yes could you please add more details?

r-czajkowski (changes requested):

solidity/dashboard/src/components/Web3ContextProvider.jsx
```diff
@@ -48,10 +49,24 @@ class Web3ContextProvider extends React.Component {
...
       yourAddress = accounts[0]
 
+      if (
+        this.state.connector?.name === WALLETS.EXPLORER_MODE.name &&
+        connector.name !== WALLETS.EXPLORER_MODE.name
+      ) {
+        if (
+          this.state.yourAddress &&
+          !isSameEthAddress(this.state.yourAddress, yourAddress)
+        ) {
```
r-czajkowski, line 59: Nested ifs. We can simplify it to one. If a condition is too long let's extract it to a separate function.

solidity/dashboard/src/components/Web3Status.jsx
```diff
@@ -7,17 +7,26 @@ import { shortenAddress } from "../utils/general.utils"
...
+  const { yourAddress, isConnected, connector } = useWeb3Context()
 
   return (
     <div className="web3-status">
       <div className="web3-status__content-wrapper">
         <div className="web3-status__network-status">
           <NetworkStatus />
         </div>
-        <div className="web3-status__wallet">
+        {connector && <CurrentWalletIconTooltip />}
```
r-czajkowski, line 22: Maybe we can use `isConnected`?

solidity/dashboard/src/connectors/explorer-mode-connector.js
```diff
@@ -0,0 +1,39 @@
...
+    this.provider.addProvider(new ExplorerModeSubprovider(this.eventEmitter))
+    this.provider.addProvider(new WebsocketSubprovider({ rpcUrl: getWsUrl() }))
+
+    this.provider.start()
+
+    return [this.selectedAccount]
+  }
+
+  disconnect = async () => {
+    this.provider.stop()
```
r-czajkowski, line 37: We should emit a `disconnect` event here to clear a `Web3Context` state after disconnecting from the provider.

r-czajkowski (changes requested):

solidity/dashboard/src/components/Routing.jsx
```diff
@@ -84,4 +84,31 @@ const CustomRoute = ({
   )
 }
 
+export const renderExplorerModePage = (PageComponent, index) => {
+  const finalPath = "/:address" + PageComponent.route.path
```
r-czajkowski, line 88: I am not convinced about this solution. I think we should not modify routes on render. Maybe we can use query params in the url? Eg. `https://dashboard.keep.network/delegations/granted?address=0x0`. Then we can detect query param in the `App` component and use `Explorer` connector. Thoughts? 🤔

r-czajkowski (changes requested):

solidity/dashboard/src/components/Routing.jsx
michalsmiarowski: Hmmm I think we are not modifying the routes on render here. I kept the `renderPage` method untouched and added another one (`renderExploreModePage`) to create routes for all pages with the address at the beginning. So for example for Liquidity Page we will have two routes:

/liquidity
/:address/liquidity

and this applies for all the page components.

But i agree with the second part of your comment - we can detect query param in the `App` component and use `Explorer` connector there 👍

r-czajkowski: ahh right... I misunderstood 👍

solidity/dashboard/src/hooks/useWalletAddressFromUrl.js
```diff
@@ -0,0 +1,19 @@
+import { useLocation } from "react-router-dom"
+import web3Utils from "web3-utils"
+
+const useWalletAddressFromUrl = () => {
+  const location = useLocation()
```
r-czajkowski, line 5: I think we can use [`useParams`](https://reactrouter.com/web/example/url-params) from the `react-router-dom` pacakge. eg:

```
const { address } = useParams() 
```

michalsmiarowski: Hmm I think we cannot because when we use the `useWalletAddressFromUrl` hook inside `useSubscribeToConnectorEvents`, which in turn is used in `AppLayout`, we are outside of a `<Routing/>` component which resulting in returning an empty object from `useParams()` 🤔

solidity/dashboard/src/components/ExplorerModeModal.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const ExplorerModeModal = ({ connector, connectAppWithWallet, closeModal }) => {
+  const [selectedAddress, setSelectedAddress] = useState("")
+
+  useEffect(() => {
+    connector.setSelectedAccount(selectedAddress)
+  }, [selectedAddress])
+
+  const submitAction = (values) => {
+    setSelectedAddress(values.address)
```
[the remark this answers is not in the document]
michalsmiarowski: Hmmm I think in this case the `ExplorerModeModal` component won't re-render, because we won't be changing the state. So even though we will set the `selectedAddress` for connector in `submitAction` (and set proper condition in return statement) the return statement won't display `<SelectedWalletModal>` component

liz-shinn:
@michalsmiarowski Added the icon to the Figma file! It's here: https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=8739%3A28

Screen Shot 2021-04-13 at 9 00 21 AM

solidity/dashboard/src/components/CustomNavLink.jsx
```diff
@@ -0,0 +1,12 @@
+import React from "react"
+import { NavLink } from "react-router-dom"
+import useWalletAddressFromUrl from "../hooks/useWalletAddressFromUrl"
+
+const CustomNavLink = ({ to, ...props }) => {
```
r-czajkowski, line 5: I think we can skip `Custom` and just `NavLink`, `Redirect` etc. IMO w/o `Custom` it will be clear that this is the right component to use in the entire app.

solidity/dashboard/src/hooks/useWalletAddressFromUrl.js
r-czajkowski: I see. How about `useRouteMatch` ? It can be used outside the `Route` component and parses the url to params based on the provided path. Please see https://reactrouter.com/web/api/Hooks/useroutematch

solidity/dashboard/src/connectors/explorer-mode-connector.js
```diff
@@ -0,0 +1,40 @@
...
+  /** @type {string} To store selected account by user. */
+  selectedAccount = ""
+
+  constructor() {
+    super(WALLETS.EXPLORER_MODE.name)
+    this.eventEmitter = new EventEmitter()
+  }
+
+  setSelectedAccount = (address) => {
+    this.selectedAccount = address
```
r-czajkowski, line 19: I think we should emit `accountsChanged` event here. Probably will be not used in our dapp via `ExplorerModeConnector`  but seems reasonable to emmit this type of event according to the rest connectors.

solidity/dashboard/src/connectors/explorerModeSubprovider.js
r-czajkowski: Maybe we can use `mixins` concept? Please see https://javascript.info/mixins.

solidity/dashboard/src/hooks/useSubscribeToConnectorEvents.js
```diff
@@ -15,19 +26,71 @@ const useSubscribeToConnectorEvents = () => {
...
     if (isConnected && connector) {
       dispatch({ type: "app/login", payload: { address: yourAddress } })
       connector.on("accountsChanged", accountChangedHandler)
       connector.once("disconnect", disconnectHandler)
+
+      if (connector.name === WALLETS.EXPLORER_MODE.name) {
+        connector.eventEmitter.on(
+          "chooseWalletAndSendTransaction",
+          showChooseWalletModal
+        )
```
r-czajkowski, line 68: If we deal with https://github.com/keep-network/keep-core/pull/2405/files#r606091686 the connector API will be consistent(highly recomended 😆 ) and `if (connector.name === WALLETS.EXPLORER_MODE.name)` will be unnecessary.

solidity/dashboard/src/hooks/useSubscribeToConnectorEvents.js
```diff
@@ -15,19 +26,71 @@ const useSubscribeToConnectorEvents = () => {
...
     return () => {
       if (connector) {
         connector.removeListener("accountsChanged", accountChangedHandler)
         connector.removeListener("disconnect", disconnectHandler)
+        if (connector.name === WALLETS.EXPLORER_MODE.name) {
+          connector.eventEmitter.removeListener(
+            "chooseWalletAndSendTransaction",
+            showChooseWalletModal
+          )
+        }
```
r-czajkowski, line 90: same as above.

r-czajkowski (changes requested):

solidity/dashboard/src/hooks/useExplorerModeConnect.js
```diff
@@ -0,0 +1,74 @@
...
+const useExplorerModeConnect = () => {
+  const location = useLocation()
+  const history = useHistory()
+  const { openModal, closeModal } = useModal()
+  const { connector, connectAppWithWallet, yourAddress } = useWeb3Context()
+
+  useEffect(() => {
+    const pathnameSplitted = location.pathname.split("/")
+    if (pathnameSplitted.length > 1 && pathnameSplitted[1]) {
+      const address = pathnameSplitted[1]
```
r-czajkowski, line 30: Can we use `useWalletAddressFromUrl` ?

r-czajkowski (changes requested):

solidity/dashboard/src/components/SelectedWalletModal.jsx
```diff
@@ -62,7 +62,9 @@ const SelectedWalletModal = ({
     if (connector && connectWithWalletOnMount) {
       setIsConnecting(true)
       wait(1000) // Delay request and show loading indicator.
-        .then(() => connectAppWithWallet(connector))
+        .then(async () => {
+          await connectAppWithWallet(connector)
```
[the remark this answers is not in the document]
michalsmiarowski: I think that without `async/await` this modal window dissapears before we are connected to Explorer Mode, which is probably not a good thing

image

michalsmiarowski:
@r-czajkowski All comments addressed - ready for another round of review

solidity/dashboard/src/connectors/explorer-mode-connector.js
```diff
@@ -0,0 +1,57 @@
...
+
+    this.provider.start()
+
+    return [this.selectedAccount]
+  }
+
+  disconnect = async () => {
+    this.explorerModeSubprovider.removeListener(
+      "chooseWalletAndSendTransaction",
+      () => {}
```
r-czajkowski, line 52: It won't remove listener added in:
```
    this.explorerModeSubprovider.on(
      "chooseWalletAndSendTransaction",
      (payload) => {
        this.emit("chooseWalletAndSendTransaction", payload)
      }
    )
```

It must be the same function, so we need to define this funcion in a class or ` this.explorerModeSubprovider.removeAllListeners("chooseWalletAndSendTransaction")`.

solidity/dashboard/src/connectors/explorer-mode-connector.js
```diff
@@ -0,0 +1,57 @@
...
+  /** @type {string} To store selected account by user. */
+  selectedAccount = ""
+
+  constructor() {
+    super(WALLETS.EXPLORER_MODE.name)
+    this.explorerModeSubprovider = Object.assign(
+      ExplorerModeSubprovider.prototype,
+      Subprovider.prototype,
+      EventEmitter.prototype
+    )
```
r-czajkowski, line 20: Hmmm I think it would be better to handle it in `./explorerModeSubprovider`.
```
// explorerModeSubprovider.js
import { EventEmitter } from "events"
import { Subprovider } from "@0x/subproviders"

class ExplorerModeSubprovider extends Subproivder {...}

Object.assign(ExplorerModeSubprovider.prototype, EventEmitter.prototype)

export default ExplorerModeSubprovider
```

solidity/dashboard/src/components/NavLink.jsx
```diff
@@ -0,0 +1,13 @@
+import React from "react"
+import { NavLink as NavLinkReactRouterDOM } from "react-router-dom"
+import useWalletAddressFromUrl from "../hooks/useWalletAddressFromUrl"
+import useFinalPath from "../hooks/useFinalPath"
+
+const NavLink = ({ to, ...props }) => {
+  const walletAddressFromUrl = useWalletAddressFromUrl()
+  const finalPath = useFinalPath(to, walletAddressFromUrl)
```
r-czajkowski, line 8: I think the `useFinalPath`  hook should use `useWalletAddressFromUrl()`:

```
const useFinalPath = (to) => {
 const address = useWalletAddressFromUrl()
  return address ? `/${address}${to}` : to
}
```

r-czajkowski (changes requested):

solidity/dashboard/src/connectors/explorer-mode-connector.js
michalsmiarowski: Oh, i see! My bad! 😅

solidity/dashboard/src/components/SelectedWalletModal.jsx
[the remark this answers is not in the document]
r-czajkowski: If we do `() => connectAppWithWallet(connector)` or
```
async () => {
          return connectAppWithWallet(connector)
}
```

it should work.

r-czajkowski (approved):
:shipit:

PR #2406, opened 2021-04-01
Move generation of docs to GH Actions

michalinacienciala:
and tBTC release processes in order to reduce human involvement and to
migration from CircleCI jobs to GitHub Actions. This commit creates
the GH Actions workflow for generation of documentation and removes the
CircleCI jobs doing the same.
The workflow is set to trigger automatically when there are changes to
files in `docs` or 'solidity' directories.
When changes are part of PR, the docs will be deployed to preview bucket.
When changes are part of merge to master (PR approve), the docs will be
deployed to main bucket.

The PR also fixes issue with non-parsable character in one of the docs,
which was causing incorrect generation of PDF file (not caught before,
because CircleCI job didn't use `--failure-level=ERROR` flag for PDF
generation and now we do use that flag in the GH Actions workflow).


michalinacienciala:
Generated documentation is now being deployed to Google Cloud Storage.
The bucket being used is not final yet, it needs to be updated before
merging the change to master. @Shadowfiend, could you advise where
to publish the artifacts? Would following be ok?

```
# in case of PRs:
          project: cfc-production
          bucket-name: docs.keep.network
          bucket-path: core/${{ github.head_ref }
# in case of pushes to master:
          project: cfc-production
          bucket-name: docs.keep.network
          bucket-path: core
```

michalinacienciala:
I don't have access to `docs.keep.network` bucket and due to that I cannot verify if docs generated by 'my' GH Actions workflow are the same as the docs generated by 'old' CircleCI jobs. Before uploading the docs to bucket I tested publishing them as run artifcact (they can be previewed here: https://github.com/keep-network/keep-core/actions/runs/708305606). It looks to me that everything that needs to be published is published, but I would appreciate close inspection of that by reviewers.
The one thing I'm not sure about is `./solitity/index.adoc` file - I think it was being published by CircleCI job, but I suspect it actually does not need to be (I don't see any other `*adoc` files being published) and due to that I decided to not publish it by the GH Actions workflow. Please let me know if that needs to be changed.

michalinacienciala:
Also, I noticed that `thesis/gcp-storage-bucket-action@v3.1.0` action that is used for upload of artifacts to Google Cloud Storage performs following command:
`gsutil -m rsync -R "$INPUT_BUILD_FOLDER" gs://"$INPUT_BUCKET_NAME"/"$INPUT_BUCKET_PATH"`
I wonder whether we shouldn't use `rsync` with additional `-d` option (see more info [here](https://cloud.google.com/storage/docs/gsutil/commands/rsync)) - it may be needed in case we want to no longer generate some document. Without it the no longer generated document will still be kept in the bucket, because it was generated and uploaded in previous deployments.

mhluongo:
> Without it the no longer generated document will still be kept in the bucket, because it was generated and uploaded in previous deployments.

Hopping in to say that I'd rather keep the docs "accidentally immutable" than "accidentally deleted", since we serve these publicly. I also expect we need to consider a bigger doc reworking, so lean against changing the mechanism for now

michalinacienciala:
> > Without it the no longer generated document will still be kept in the bucket, because it was generated and uploaded in previous deployments.

> Hopping in to say that I'd rather keep the docs "accidentally immutable" than "accidentally deleted", since we serve these publicly. I also expect we need to consider a bigger doc reworking, so lean against changing the mechanism for now

Makes sense. I'm leaving it as it is for now.

michalinacienciala:
Before approving this PR please add `docs-generate` job as a required for PR merge and remove the requirements for CircleCI jobs related to docs generation/upload.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,191 @@
...
+      #     registry: ghcr.io
+      #     image: keepnetwork/texlive:3
+      #     options: -v ${{ github.workspace }}:/volume-dir
+      #     run: |
+      #       mkdir /volume-dir/tmp/docs/img
+      #       cd /volume-dir/docs
+      #       make clean
+      #       make docs
+
+      - name: Compile TeX files to PDFs
```
nkuba, line 55: Thoughts on running this job in the following way:
```
    runs-on: ubuntu-latest
    container: keepnetwork/texlive:3
    defaults:
      run:
        working-directory: ./docs
    steps:
      - uses: actions/checkout@v2
      - run: make clean 
      - run: make docs
      # now publish PDF to bucket
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,111 @@
+name: Docs
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - "docs/**"
+      - "solidity/scripts/generate-api-docs.js"
```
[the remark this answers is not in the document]
michalinacienciala: Applied in a24d498c178a4fbf093e7c74657661bbd8ecc328

.github/workflows/docs.yml
michalinacienciala: I did end up applying slightly modified solution. I had to add moving generated file to separate folder, as later action publishing the files to bucket expects the files to be in `build-folder` and providing the path to `docs` there would upload too much.

michalinacienciala:
I attach the archive containing files uploaded to bucket as a result of running the workflow:
[move-generate-and-upload-docs-wip2.zip](https://github.com/keep-network/keep-core/files/6365208/move-generate-and-upload-docs-wip2.zip)
For comparison purposes I also attach files generated by old CircleCI job:
[explorer-mode.zip](https://github.com/keep-network/keep-core/files/6365244/explorer-mode.zip)
The difference is that current workflow does not upload `./solitity/index.adoc` (discussed in one of the previous comments) and also does not upload `/img` folder containing files needed for generation of `relay-states.pdf`.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,194 @@
+name: Docs
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - "docs/**"
+      - "solidity/**"
```
nkuba, line 9: Let's exclude the dashboard directory.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,194 @@
...
+    if: |
+      github.event_name != 'pull_request'
+        || needs.docs-detect-changes.outputs.path-filter == 'true'
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Generate solidity docs
+        run: |
+          cd solidity && npm install
+          node_modules/.bin/truffle compile
```
nkuba, line 89:
```suggestion
          cd solidity && npm ci
          npx truffle compile
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,194 @@
...
+    continue-on-error: true
+    if: |
+      github.event_name != 'pull_request'
+        || needs.docs-detect-changes.outputs.path-filter == 'true'
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Generate solidity docs
+        run: |
+          cd solidity && npm install
```
nkuba, line 88: Can we define a working directory for the step instead of using `cd solidity`?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,194 @@
...
+          node_modules/.bin/truffle compile
+          mkdir -p ${{ github.workspace }}/docs/solidity
```
nkuba, line 90: Could we define `${{ github.workspace }}/docs/solidity` as a job level variable and use it in all steps referring to this path? Similar to what we did with `DOCS_DIR` in `docs-tex` job.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,194 @@
...
+          name: Solidity Docs
+          path: ${{ github.workspace }}/docs/solidity/index.adoc
+
+  docs-html:
+    runs-on: ubuntu-latest
+    needs: [docs-detect-changes, docs-solidity]
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Download a Build Artifact
```
nkuba, line 105:
```suggestion
      - name: Download Solidity Docs Artifact
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,194 @@
...
+
+      - name: Upload a Build Artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: Solidity Docs
+          path: ${{ github.workspace }}/docs/solidity/index.adoc
+
+  docs-html:
+    runs-on: ubuntu-latest
+    needs: [docs-detect-changes, docs-solidity]
```
nkuba, line 101: We want to run `docs-html` and `docs-pdf` jobs even if `docs-solidity` failed ([`if: always()`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-not-requiring-dependent-jobs-to-be-successful)).

.github/workflows/docs.yml
```diff
@@ -0,0 +1,194 @@
...
+          bucket-path: core-wip2 # TODO: remove `-wip2` after tests
+          build-folder: tmp/docs
+
+      # A pull_request event is a PR; deploy to preview bucket.
+      - if: github.event_name == 'pull_request'
+        name: Upload preview of generated files
+        uses: thesis/gcp-storage-bucket-action@v3.1.0
+        with:
+          service-key: ${{ secrets.DOCS_KEEP_NETWORK_UPLOADER_SERVICE_KEY_JSON }}
+          project: keep-test
```
nkuba, line 71: We need to remember to replace it in all jobs with the production project/bucket before we merge this PR.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,194 @@
...
+          bucket-name: docs.keep.network
+          bucket-path: core/${{ github.head_ref }}-wip2 # TODO: remove `-wip2` after tests
+          build-folder: ${{ steps.html.outputs.asciidoctor-artifacts }}/docs
+
+      - name: Upload a Build Artifact # TODO: Remove this after tests finished
+        uses: actions/upload-artifact@v2
+        with:
+          name: HTML Docs
+          path: ${{ steps.html.outputs.asciidoctor-artifacts }}/docs
```
nkuba, line 145: We don't need this anymore, right?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,194 @@
...
+
+  docs-pdf:
+    runs-on: ubuntu-latest
+    needs: [docs-detect-changes, docs-solidity]
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Download a Build Artifact
```
nkuba, line 153:
```suggestion
      - name: Download Solidity Docs Artifact
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,194 @@
...
+          project: keep-test
+          bucket-name: docs.keep.network
+          bucket-path: core/${{ github.head_ref }}-wip2 # TODO: remove `-wip2` after tests
+          build-folder: ${{ steps.pdf.outputs.asciidoctor-artifacts }}/docs
+
+      - name: Upload a Build Artifact # TODO: Remove this after tests finished
+        uses: actions/upload-artifact@v2
+        with:
+          name: PDF Docs
+          path: ${{ steps.pdf.outputs.asciidoctor-artifacts }}/docs
```
nkuba, line 194: 💣 ❓

nkuba (changes requested):

.github/workflows/docs.yml
michalinacienciala: See c2acfd995d97eb540f6d2c84824f77c4682b0b19.

.github/workflows/docs.yml
michalinacienciala: See c2acfd995d97eb540f6d2c84824f77c4682b0b19.

.github/workflows/docs.yml
michalinacienciala: See 4d7d113b4f3111d55bbf054dd5ea9dc6aa9ea131.

.github/workflows/docs.yml
michalinacienciala: See 4d7d113b4f3111d55bbf054dd5ea9dc6aa9ea131.

.github/workflows/docs.yml
michalinacienciala: See def6d8a4167d963cc3682c3a18d35d9fca0e925a.

.github/workflows/docs.yml
michalinacienciala: See 6f8839b5b10e9d059ba583a702e503c131e879b2. I also removed the `continue-on-error: true` from `docs-solidity`.

.github/workflows/docs.yml
michalinacienciala: Right, I checked the content of the artifact for latest build (everything ok) and finally removed the step in 8c33123f4fc6bb176a66fd8bfc69e9a9437d3458.

.github/workflows/docs.yml
michalinacienciala: Right, I checked the content of the artifact for latest build (everything ok) and finally removed the step in 8c33123f4fc6bb176a66fd8bfc69e9a9437d3458.

.github/workflows/docs.yml
michalinacienciala: Yes, and also `-wip2`suffixes will need to be removed from couple of places and small change in path-filter criteria will need to change before merging. I will commit those changes if there will be no other issues after re-review.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,187 @@
...
+    outputs:
+      path-filter: ${{ steps.filter.outputs.path-filter }}
+    steps:
+      - uses: actions/checkout@v2
+        if: github.event_name == 'pull_request' 
+      - uses: dorny/paths-filter@v2
+        if: github.event_name == 'pull_request' 
+        id: filter
+        with:
+          filters: | # TODO: Remove '|(.github/workflows/docs.yml)' once workflow gets tested
```
nkuba, line 26: Let's remember about this TODO.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,187 @@
...
+
+      # A push event is a master merge; deploy to primary bucket.
+      - if: github.event_name == 'push'
+        name: Upload asciidocs
+        uses: thesis/gcp-storage-bucket-action@v3.1.0
+        with:
+          service-key: ${{ secrets.DOCS_KEEP_NETWORK_UPLOADER_SERVICE_KEY_JSON }}
+          project: keep-test
+          bucket-name: docs.keep.network
+          bucket-path: core-wip2 # TODO: remove `-wip2` after tests
```
nkuba, line 131: I was under an impression that I already commented on it but looks like the comment was not submitted, so... 
For keep-core docs, we're probably going to push it to the root directory of the bucket. I believe we do that now. But would love to hear @Shadowfiend thoughts around that.

nkuba (commented):
It looks fine to me. I think we can handle all the TODOs and `wip` suffixes now and get it ready for master merge. Before merging I'd love @Shadowfiend to give it a second approval.

.github/workflows/docs.yml
michalinacienciala: Yes, in one of the previous comments I also was considering dropping `core` in `bucket-path`, as in CircleCI job we are currently uploading files directly to `cfc-production` project's `docs.keep.network` bucket.

.github/workflows/docs.yml
michalinacienciala: Removed, see 0d01d5f5d9b81843e5c1cdc9b0fe1daa58a502bd.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,187 @@
...
+
+      # A push event is a master merge; deploy to primary bucket.
+      - if: github.event_name == 'push'
+        name: Upload generated files
+        uses: thesis/gcp-storage-bucket-action@v3.1.0
+        with:
+          service-key: ${{ secrets.DOCS_KEEP_NETWORK_UPLOADER_SERVICE_KEY_JSON }}
+          project: cfc-production
+          bucket-name: docs.keep.network
+          bucket-path: core
```
Shadowfiend, line 63: Agree that this should be the root for master push. We _should_ implement a similar build for ECDSA (it doesn't exist yet), and we _should_ make that push to `/ecdsa` for master push.

Ideally, we would do a docs build that would include `ecdsa`, `tbtc`, and `core` in one build---but the split repos make that very hard to deal with. We've done the `/tbtc` approach for tBTC, and I think we can do the same for ECDSA. We should continue to consider keep-core our “root” project for all intents and purposes, however.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,187 @@
...
+
+      # A pull_request event is a PR; deploy to preview bucket.
+      - if: github.event_name == 'pull_request'
+        name: Upload preview of generated files
+        uses: thesis/gcp-storage-bucket-action@v3.1.0
+        with:
+          service-key: ${{ secrets.DOCS_KEEP_NETWORK_UPLOADER_SERVICE_KEY_JSON }}
+          project: cfc-production
+          bucket-name: docs.keep.network
+          bucket-path: core/${{ github.head_ref }}
```
Shadowfiend, line 74: Similar to above, using `${{ github.head_ref }}` is perfect here.

The “correct” way to do this per other builds is to have a `preview.docs.keep.network/<branch>/<everything else>`. That's actually not straightforward, because securing that with HTTPS requires some certificate wrangling that probably isn't worth doing right now.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,187 @@
...
+    env:
+      SOLIDITY_DOCS_DIR: ${{ github.workspace }}/docs/solidity
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Generate solidity docs
+        working-directory: ./solidity
+        run: |
+          npm ci
+          npx truffle compile
```
Shadowfiend, line 92: Would definitely be interesting if we could reuse the build artifacts from the Solidity build here, fwiw. Maybe by attaching the Solidity docs build to the Solidity workflow rather than the docs workflow.

None of that is straightforward, so it's more of a “food for thought” thing---for now this is just fine.

.github/workflows/docs.yml
Shadowfiend: Does this also make it run if `docs-detect-changes` reports no changes though?

Shadowfiend (dismissed):
Other than the upload path for master builds, I did a quick pass and everything looks good to me. Left a couple of other longer-term thoughts as well. :rocket:

.github/workflows/docs.yml
michalinacienciala: Changed to root directory, see 858826317d4776d487008b896b378d6b995c5446.

.github/workflows/docs.yml
michalinacienciala: You're right, it does. I've decided to provide additional conditions - jobs will be executed even if earlier jobs fail, but only if `github.event_name != 'pull_request' || needs.docs-detect-changes.outputs.path-filter == 'true'` criterium is met.
I also considered setting `continue-on-error` for `docs-solidity` job and removing `if` conditions from `docs-html` and `docs-pdf` jobs, but that would not mark the workflow as failed in case of `docs-solidity` job failure (making it possible for us to not notice the failure in some situations).

.github/workflows/docs.yml
michalinacienciala: Will keep that in mind as something for consideration for the future, but merging the workflows might be hard to achieve, as the conditions on which we want to trigger Solidity and Docs are different (maybe this can be somehow accommodated, but likely with some code acrobatics, making the code harder to read and sustain).
But as I said, I will keep that in mind, maybe I'll think of some reasonable solution to this at later stage...

michalinacienciala:
> Before approving this PR please add `docs-generate` job as a required for PR merge and remove the requirements for CircleCI jobs related to docs generation/upload.

As I've split the `docs-generate` job to four separate jobs, what would need to be added as required is this:
* `docs-tex`
* `docs-solidity`
* `docs-html`
* `docs-pdf`

nkuba (approved):
LGTM!

nkuba:
I'm merging this PR as all @Shadowfiend comments were adressed.

PR #2413, opened 2021-04-08
Improve fetching contract deployment block

r-czajkowski:
This PR improves fetching contract deployment block. In the previous implementation, the contract deployment block was fetched when dapp was connecting to the wallet. In this case, it took too long to connect the dapp with a wallet. Here we added a `getContractDeploymentBlockNumber` function which returns the block number from a cache or fetches this data from a node. So the deployment block will be fetched when needed.


michalsmiarowski (approved):
Good job 👍 Let's :shipit:

PR #2416, opened 2021-04-12
Fix fetching APY w/o connected wallet

r-czajkowski:
The APY on the Liquidity rewards page should be fetched w/o connected wallet.
The saga which fetches APY should start in the root saga before the
login/logout flow.


michalsmiarowski (approved):
Tested locally on mainnet and the code does what it suppose to do. Let's ship it!

PR #2418, opened 2021-04-13
Fix fetching operators of owner

r-czajkowski:
This PR fixes fetching operators of owner. In the `getOperatorsOfOwner` function we used `getContractDeploymentBlockNumber` in wrong way.


michalsmiarowski (approved):
🚢

PR #2421, opened 2021-04-14
Fix fetching Random Beacon rewards

r-czajkowski:
The saga that fetches the Beacon rewards expects object with `address`
property in a payload.


michalsmiarowski (approved):
Looking good 👍 Tested locally and it catches the address properly

PR #2424, opened 2021-04-16
Fix fetching paid back delegations

michalsmiarowski:
When fetching paidBackDelegations we used the old way of getting the `fromBlock`
number which was changed in 221fea5fd9da01e933e1b1f854243478cdb857c2 commit.


r-czajkowski (approved):
My fault 🤦. Thanks for fixing this :shipit:

PR #2425, opened 2021-04-16
Fix Reward Multiplier calculation

michalsmiarowski:
The current formula for reward multiplier was wrong. This PR fixes it.

The new formula:
We are calling `updateAccounting()` method to get the reward that the user has
accumulated with the max multiplier. Then we divide the current reward by the
reward with max boost and scale the boost to be in range 1.0x - 3.0x instead of
0.3 - 1.

Also, the `formatPercentage()` function was renamed to `formatValue()` for 
better readability.


solidity/dashboard/src/utils/general.utils.js
```diff
@@ -93,3 +97,30 @@ export const displayPercentageValue = (
...
+ * @param {BigNumber|String|Number} inputX Number on number line
+ * @param {BigNumber|String|Number} xMin - Minimal value of the current number
+ * line
+ * @param {BigNumber|String|Number} xMax - Maximal value of the current number
+ * line
+ * @param {BigNumber|String|Number} yMin - Minimal value of the new number line
+ * @param {BigNumber|String|Number} yMax - Maximal value of the new number line
+ * @return {BigNumber} Number from the new number line
+ */
+export const scaleInputForNumberRange = (inputX, xMin, xMax, yMin, yMax) => {
```
r-czajkowski, line 117: Could we add some tests for this function?

r-czajkowski (changes requested):
Looking good and it works as expected 🚀 . Let's just add some unit tests for `scaleInputForNumberRange` function.

solidity/dashboard/src/utils/general.utils.js
```diff
@@ -93,3 +97,30 @@ export const displayPercentageValue = (
   }
   return `${prefix}${value}%`
 }
+
+/**
+ * Returns a number that is in the same spot on <yMin, yMax> number line that
+ * inputX is on <xMin, xMax> number line
+ *
+ * e.g 2 on the number line of <1, 3> is exactly in the middle so for the
+ * number line of <1, 10> it will return 5.
```
r-czajkowski, line 106: It will return `5.5` instead of `5`.

r-czajkowski (changes requested):
Just a tiny comment.

solidity/dashboard/src/utils/general.utils.js
michalsmiarowski: That's right! Thanks 🔥

r-czajkowski (approved):
:shipit:

PR #2427, opened 2021-04-19
Add Go linting to the client.yml workflow

michalinacienciala:
As part of work on RFC-18 a job linting the Go code has been added to
`client.yml` workflow. Action for linting of Go code has been also added
(as separate repository) as part of this work.

Note: running the workflow with linting revealed a number of issues found
by the linter ([see here](https://github.com/keep-network/keep-core/actions/runs/763198160)).
Those issues will also be fixed as part of this PR.
Changes include:
- adding missing comments
- improving existing comments
- renaming `DiagnosticsRegistry` type/func to `Registry` (same change
  applied also in keep-common)
- renaming `SetSenderId` func to `SetSenderID`
- exporting all `initiatorAct*` and `responderAct*`

Also, @nkuba, after merging this PR into master please configure the
`client-lint` job in GitHub's settings as required for PR merging.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/ipfs/go-datastore v0.4.4
 	github.com/ipfs/go-log v1.0.4
 	github.com/keep-network/go-libp2p-bootstrap v0.0.0-20200423153828-ed815bc50aec
-	github.com/keep-network/keep-common v1.4.1-0.20210426115544-f1c11be35fce
+	github.com/keep-network/keep-common v1.4.1-0.20210429125057-3f1aa45d984a
 	github.com/libp2p/go-addr-util v0.0.2
 	github.com/libp2p/go-libp2p v0.10.3
 	github.com/libp2p/go-libp2p-connmgr v0.2.4
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/ipfs/go-datastore v0.4.4
 	github.com/ipfs/go-log v1.0.4
 	github.com/keep-network/go-libp2p-bootstrap v0.0.0-20200423153828-ed815bc50aec
-	github.com/keep-network/keep-common v1.4.1-0.20210319095805-ebf46d0b62db
+	github.com/keep-network/keep-common v1.4.1-0.20210426115544-f1c11be35fce
 	github.com/libp2p/go-addr-util v0.0.2
 	github.com/libp2p/go-libp2p v0.10.3
 	github.com/libp2p/go-libp2p-connmgr v0.2.4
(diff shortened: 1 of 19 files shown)


pkg/net/libp2p/authenticated_connection.go
```diff
@@ -146,12 +146,12 @@ func (ac *authenticatedConnection) runHandshakeAsInitiator() error {
 	// Act 1
 	//
 
-	initiatorAct1, err := handshake.InitiateHandshake(ac.protocol)
+	InitiatorAct1, err := handshake.InitiateHandshake(ac.protocol)
```
lukasz-zimnoch, line 149: Why do we capitalize the variable name as well (here and all changes below)? Even if we need to export `InitiatorAct*` etc., we should leave all variables names uncapitalized. This is the convention in Go.

pkg/altbn128/altbn128.go
```diff
@@ -9,10 +9,12 @@ import (
 	"github.com/keep-network/keep-core/pkg/internal/byteutils"
 )
 
+// G1Point represents G1 point
```
nkuba, line 12: We need to add dots at the ends of all these comments we're adding in this PR.

nkuba (changes requested):

pkg/net/libp2p/authenticated_connection.go
michalinacienciala: Sorry, my bad, fixed in 553773d9cb7555c8ef5ed6fde47829718774f59d.

pkg/altbn128/altbn128.go
michalinacienciala: Fixed in abb1a2151c4f605c7447e4817f710d80c01fe4b8.

lukasz-zimnoch (approved):
LGTM! Leaving final approval to @nkuba

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/ipfs/go-datastore v0.4.4
 	github.com/ipfs/go-log v1.0.4
 	github.com/keep-network/go-libp2p-bootstrap v0.0.0-20200423153828-ed815bc50aec
-	github.com/keep-network/keep-common v1.4.1-0.20210319095805-ebf46d0b62db
+	github.com/keep-network/keep-common v1.4.1-0.20210426115544-f1c11be35fce
```
nkuba, line 19: Can we update this version to the latest  merge commit (`3f1aa45d984ac25a1c5b4b1f511a790f69817072`)?

nkuba (changes requested):

go.mod
michalinacienciala: Updated, see 1e06979505d9024b05fa6ebe1f77bce74ec0a01c.

nkuba (approved):

Issue #2357, opened 2021-03-02
Show a hover tooltip for abbreviated numbers


liz-shinn:
Introduce a pattern of showing a tooltip that displays the full number for all abbreviated numbers in the token dashboard. 

Throughout the dashboard, we show abbreviated numbers in instances where there is limited UI space or the number is larger. However, we don't consistently show the user what the specific number is. This would allow the user to hover over any abbreviate number (e.g. 150K) in the dashboard and have it show the full number (e.g. 150,000) in a tooltip.

liz-shinn:
![tooltip](https://user-images.githubusercontent.com/57226633/111187863-d1fe0a00-858a-11eb-81b5-485e14f4ec9d.png)

PR #2430, opened 2021-04-19
Tooltips for abbreviated numbers

r-czajkowski:
Closes: #2095 
Closes: #2357 

This PR adds support for showing a tooltip that displays the full number for all abbreviated numbers. It also improves a `TokenAmount` component based on the #2095.


solidity/dashboard/src/forms/common-validators.js
```diff
@@ -22,7 +22,12 @@ export const validateEthAddress = (address, required = true) => {
   }
 }
 
-export const validateAmountInRange = (value, maxValue, minValue = 0) => {
+export const validateAmountInRange = (
+  value,
+  maxValue,
+  minValue = 0,
+  token = KEEP
+) => {
```
michalsmiarowski, line 30: There is a problem with validation for depositing KEEP in KEEP-only pool.

image

When you write "0" it shows the message "The value should be greater or equal than <1" (it should just be "1" instead of "<1")

I think the problem lies there:

https://github.com/keep-network/keep-core/blob/d2e9cbc3ceed0610a857d1daae594277b738563a/solidity/dashboard/src/components/KeepOnlyPool.jsx#L297

We pass 1 to the validateAmountInRange but it's 1 wei instead of 1 token. That's why it fails on this if:
https://github.com/keep-network/keep-core/blob/d2e9cbc3ceed0610a857d1daae594277b738563a/solidity/dashboard/src/utils/token.utils.js#L164

It's because `MIN_AMOUNT_TO_DISPLAY` is 1 token (10^18) not 1 wei.

michalsmiarowski: Also let's check another places where we use 'validateAmountInRange' because i think same thing is happening in `CreateTokenGrantForm.jsx`

solidity/dashboard/src/services/liquidity-rewards.js
```diff
@@ -140,7 +140,7 @@ class LiquidityRewards {
 
 class UniswapLPRewards extends LiquidityRewards {
   calculateAPY = async (totalSupplyOfLPRewards) => {
-    totalSupplyOfLPRewards = toTokenUnit(totalSupplyOfLPRewards)
+    totalSupplyOfLPRewards = Token.toTokenUnit(totalSupplyOfLPRewards)
```
michalsmiarowski, line 143: Something is not right here. For the `totalSupplyOfLPRewards` that equals `105552000000000000000000` the `Token.toTokenUnit()` function returns BigNumber that is NaN. Because of that APY for KEEP-only pool is not displayed.

image

solidity/dashboard/src/css/tooltip.less
```diff
@@ -115,20 +115,9 @@
         position: relative;
         
         .tooltip {
-            &__trigger {
```
michalsmiarowski, line 118: Look into the styles of the tooltip. For big numbers it doesn't show full number and it does not look appealing for the eye because it's missing padding on the right 😛 

Example:
image

Real value of totalKEEP:
image

solidity/dashboard/src/utils/token.utils.js
```diff
@@ -1,96 +1,192 @@
...
+export const ETH = new Token("Ether", 18, "ETH", "gwei", 14, Icons.ETH)
+export const TBTC = new Token("tBTC", 18, "TBTC", "tSats", 8, Icons.TBTC)
+export const LPToken = new Token(
+  "Liqudity Provider Token",
+  18,
+  "LP",
+  "LP",
+  18,
+  null,
+  4
```
michalsmiarowski, line 191: Last argument is not needed here because default value for `_decimalsToDisplay` is 4 so i think we can remove it.

michalsmiarowski (changes requested):
Looking good! Great job with that Token component. Left some comments there and there to look into.

r-czajkowski:
@michalsmiarowski, ready for another 👀.

michalsmiarowski (approved):
Looking good. Let's 🚢

Issue #2187, opened 2020-12-03
Make clearer in UI that there's 12 hour delay before you commit top-up


liz-shinn:
Show in the UI flow for adding KEEP as a top up that there's a 12 hour delay.

liz-shinn:
Can add some more information in the top up modal that tells about 12 hour delay before you commit top up.

liz-shinn:
Proposal to adjust this addressed as well in https://github.com/keep-network/keep-core/issues/1983

Would add a final modal to the end that states there's a 12 hour delay:
![Add KEEP 4](https://user-images.githubusercontent.com/57226633/113731201-94daf280-96c6-11eb-87a3-a91c3ce829ec.png)

@Battenfield

Issue #2077, opened 2020-09-21
Make the 12 hour delay for delegation clearer


liz-shinn:
Discord user feedback
>making the 12 hour delay for delegation clearer, maybe a modal or something that explains why there is a delay

liz-shinn:
Could show this in the staking modal, a bit of copy that indicates, maybe a note/warning icon

cc/ @Battenfield

Issue #1983, opened 2020-08-11
Top-up cannot be cancelled note in KEEP token dashboard


pdyraga:
A top-up cannot be cancelled so it's important that the user is very careful with the amount of KEEP tokens added to the stake. The only way to get those tokens back is to undelegate and recover the stake.

Should we employ some UX pattern or at least add a note to the top-up modal being explicit about the fact a top-up cannot be cancelled and the user has to be careful?

liz-shinn:
Yes — I think adding a note to the top up modal is a good call. 

We could add a confirmation interaction here as well, something along the lines of:
"I understand that the amount of KEEP tokens I am adding cannot be edited once submitted as a top up. To retrieve my tokens I will need to undelegate and recover the entire stake delegation."

liz-shinn:
Could we instead offer an option to partially withdraw a delegation, discussing with @Battenfield

liz-shinn:
I propose some updates to this modal series that would adjust how we display current balance, added amount, and final balance. Would also add a note about top up cancellation:

<img width="967" alt="Screen Shot 2021-04-06 at 10 52 50 AM" src="https://user-images.githubusercontent.com/57226633/113730844-3ca3f080-96c6-11eb-8c63-97e26ffb3fd1.png">

liz-shinn:
![Add KEEP 1](https://user-images.githubusercontent.com/57226633/113730941-56453800-96c6-11eb-9fc0-912305ea1f17.png)
![Add KEEP 2](https://user-images.githubusercontent.com/57226633/113730943-56ddce80-96c6-11eb-8a30-0af0a7a439de.png)
![Add KEEP 3](https://user-images.githubusercontent.com/57226633/113730944-56ddce80-96c6-11eb-9aae-d4deab205dda.png)
![Add KEEP 4](https://user-images.githubusercontent.com/57226633/113730945-57766500-96c6-11eb-86ba-3c4a86dbf2d3.png)

@Battenfield

PR #2434, opened 2021-04-23
Top up flow- UI improvements

r-czajkowski:
~Depends on: #2430~
Closes: #2077
Closes: #1983
Closes: #2187

This PR improves the top up UI flow. Here we added a confirmation modal with a note that a top-up cannot be cancelled and a final modal to the end that states there's a 12 hour delay. Please see screenshots below.

<details>
  
  obraz

obraz

obraz

</details>


michalsmiarowski (commented):
@r-czajkowski Shouldn't we display the same modal when clicking **DELEGATE STAKE** button when staking for the same address again (so basically delegating a stake to an address that we have already delegeted a stake before)? Because in that situation it also creates a top-up, similar when you click **ADD KEEP** in the Delegations table. Thoughts?

solidity/dashboard/src/components/AmountForm.jsx
```diff
@@ -57,20 +34,29 @@ const AmountForm = ({
           normalize={normalizeAmount}
           format={formatFormAmount}
         />
-        <div className="text-caption text-grey-50">
-          {KEEP.displayAmountWithSymbol(availableAmount)} available.
+        <div className="flex row center" style={styles.availableAmountWrapper}>
```
michalsmiarowski, line 37: I think we should add `align-items: baseline` here becase the word "available" looks off kompared to the KEEP amount

solidity/dashboard/src/components/AddTopUpModal.jsx
```diff
@@ -1,39 +1,209 @@
...
+    <DelegationAddress address={operatorAddress} label="operator" />
+    <DelegationAddress address={beneficiary} label="beneficiary" />
+  </>
+)
+
+const TopUpInitiatedConfirmationModal = ({
+  // In token unit.
+  addedAmount,
+  // In the smallest token unit.
+  currentAmount,
```
michalsmiarowski, line 163: Just wondering - wouldn't it be better to keep the amount props in the same unit? We could change it when passing the values to the props instead of doing it inside component. I think it would be better for readability.

r-czajkowski:
> @r-czajkowski Shouldn't we display the same modal when clicking **DELEGATE STAKE** button when staking for the same address again (so basically delegating a stake to an address that we have already delegeted a stake before)? Because in that situation it also creates a top-up, similar when you click **ADD KEEP** in the Delegations table. Thoughts?

Good point! But maybe we should display just a modal indicating that delegation for this address exists and should use top-up? My point is that if a user fills a staking form with the same operator address but with a different `authorizer` and `beneficiary`, the top-up modal will display the actual `authorizer` and `beneficiary` address of this delegation, which may be misleading. Also staking form is for new delegations and the user should use top-up flow so I think this approach should be ok. What do you think?

solidity/dashboard/src/components/AddTopUpModal.jsx
```diff
@@ -1,39 +1,209 @@
...
+        authorizerAddress={authorizerAddress}
+        beneficiary={beneficiary}
+        operatorAddress={operatorAddress}
+      />
+      <div className="flex row center mt-2 mb-2">
+        <div className="flex-1">
+          <Tag text="Current" IconComponent={Icons.KeepToken} />
+        </div>
+        <TokenAmount amount={currentAmount} />
+      </div>
```
michalsmiarowski, line 97: image

I'm thinking that we could also show the amount of KEEP placed in the top-ups for this address just below the amount of KEEP in the delegation. E.g right now this address has a delegation of a 60k KEEP but there is also a top-up for this address with the amount of 178k KEEP that is not commited. Might be a good idea to display this in this modal (and probably also in Step2 modal). Thoughts?

michalsmiarowski (changes requested):

michalsmiarowski:
>Good point! But maybe we should display just a modal indicating that delegation for this address exists and the user should top-up delegation via add KEEP button? My point is that if a user fills a staking form with the same operator address but with a different authorizer and beneficiary, the top-up modal will display the actual authorizer and beneficiary address of this delegation, which may be misleading. Also staking form is for new delegations and the user should use top-up flow so I think this approach should be ok. What do you think?

Sounds good to me!! Let's do that! :muscle:

solidity/dashboard/src/components/AddTopUpModal.jsx
r-czajkowski: Makes sense for me but not sure if we should add not committed top-ups to the current amount because not committed tokens do not participate in the work of the network. So I would suggest rather list pending top-ups but not sure how we can handle this in the UI, so I'll ping @liz-shinn here.

Also sounds like we should address it in a separate PR to not block the v.1.13.0 release. I assume that multiple top-ups for the same delegation are rather rare.

solidity/dashboard/src/components/AddTopUpModal.jsx
```diff
@@ -1,39 +1,209 @@
...
+  const onSubmit = (values) => {
+    if (step === 1) {
+      setAmount(values.amount)
+      setStep((preveStep) => preveStep + 1)
+    } else if (step === 2) {
+      submitConfirmationModal({ amount })
+    }
+  }
+
+  const amountInTokenUnit = KEEP.fromTokenUnit(amount)
```
michalsmiarowski, line 35: Bad variable name. We change the `amount` from Token Unit, so it's not in token unit after

michalsmiarowski (approved):
LGTM! 🚀

solidity/dashboard/src/components/AddTopUpModal.jsx
liz-shinn: >E.g right now this address has a delegation of a 60k KEEP but there is also a top-up for this address with the amount of 178k KEEP that is not commited. Might be a good idea to display this in this modal (and probably also in Step2 modal). Thoughts?

So this is a scenario in which the user has a previously submitted top up but has not committed it and wants to submit an additional top up, correct?

Also, I think we should display the full addresses here since there's room in the modal to do so versus abbreviating them.

liz-shinn: What about something like this @michalsmiarowski @r-czajkowski 
In figma here: https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=8916%3A0

Screen Shot 2021-04-29 at 11 33 08 AM

liz-shinn: Idk if "top-up" or "top up" – whichever is consistent with how we've referred to it in the UI previously

Issue #2297, opened 2021-01-19
Disable the Beacon contract on the Applications page


liz-shinn:
We should show a disabled state for the Beacon contract – currently the Authorize button is active and allows you to get to a transaction window with gas fees of $1000+ USD, no bueno.

<img width="347" alt="Screen Shot 2021-01-19 at 5 20 44 PM" src="https://user-images.githubusercontent.com/57226633/105100895-01f1cb00-5a7c-11eb-8bdb-6da4bced35c0.png">

liz-shinn:
<img width="347" alt="Screen Shot 2021-01-19 at 5 20 44 PM" src="https://user-images.githubusercontent.com/57226633/105100989-2ea5e280-5a7c-11eb-94db-40620489f485.png">

Disable the "Authorize" button here ^^

liz-shinn:
cc/ convo with @Battenfield 

1. Error in the contract, doesn't have enough gas to cover what it does cost
2. Whole thing blows up when the user doesn't have enough gas in their wallet to begin with for what it would cost
3. Is this is a MetaMask issue

![Screen Shot 2021-03-02 at 1 03 06 PM](https://user-images.githubusercontent.com/57226633/109693504-bb51bf00-7b57-11eb-849d-2fafaf437a90.png)

Course of action: 
1. If Metamask issue, then: 
-- we can render a model that says "This is an expensive transaction to authorize. please make sure you have enough ETH to cover the gas cost. If you don't have enough gas, MetaMask will show an incorrect balance usually of $1,000 or more."
-- when they reject the transaction then say "please make sure you have enough gas to cover which is xyz amount" but that shows an error when we don't know that's why they closed a transaction

2. If contract error, then the transaction will fail - MetaMask does the same thing
-- Then we would disable the button

Battenfield:
- [ ] Spike out reason for the contract exception
    - [ ] Disable button if it's contract-related or design modal if it's just a lack of ETH issue

Shadowfiend:
Sorry for not noticing this folks. The reason this transaction fails is that the random beacon operator contract has been panic buttoned (see https://docs.keep.network/random-beacon/#_the_registry_and_panic_button) due to issues from the November geth hard fork (see https://docs.keep.network/status-reports/2020-11-11-retro-geth-hardfork.html).

[...]

PR #2441, opened 2021-04-27
Disable the Beacon contract on the Applications page

r-czajkowski:
Closes: #2297

This PR disables the `authorize` button for the Beacon contract on the Application page and adds a tooltip indicating the operator contract is disabled.


solidity/dashboard/src/components/AuthorizeContracts.jsx
```diff
@@ -65,41 +66,27 @@ const AuthorizeContracts = ({
...
+                <AuthorizeContractItem
+                  key={contract.contractName}
+                  {...contract}
+                  operatorAddress={operatorAddress}
+                  onAuthorizeBtn={onAuthorizeBtn}
+                  onDeauthorizeBtn={onDeauthorizeBtn}
+                  onSuccessCallback={onSuccessCallback}
+                />
+              ))}
+            </ul>
```
michalsmiarowski, line 80: Not sure why we are getting rid of `Contracts` component here. Both solutions are valid but i think the best practise is to separate everything as much as possible into separate components. Who knows, maybe we would want to reuse it in the future?

solidity/dashboard/src/components/AuthorizeContracts.jsx
```diff
@@ -153,14 +140,39 @@ const AuthorizeContractItem = ({
...
           </div>
         ) : (
-          <SubmitButton
-            onSubmitAction={onAuthorize}
-            className="btn btn-secondary btn-sm"
-            style={{ marginLeft: "auto" }}
-            successCallback={onSuccess}
+          <Tooltip
+            shouldShowTooltip={
+              contractName === "Keep Random Beacon Operator Contract"
```
michalsmiarowski, line 145: Would be best to save this `contractName` as constants somewhere. Not only `Keep Random Beacon Operator Contract` but i see there are also `TBTCSystem` and `BondedECDSAKeepFactory` that are used in multiple places.

Those name are different than the names in the `constants.js` file (e.g in `constants.js` there is `bondedEcdsaKeepFactoryContract` instead of `BondedECDSAKeepFactory`) which makes it confusing 🤔 Maybe we should create separate constants and call it `CONTRACT_LABEL` or something? Thoughts?

michalsmiarowski (changes requested):
Overall looks good and works fine! Left some comments regarding best code practices.

solidity/dashboard/src/components/AuthorizeContracts.jsx
r-czajkowski: I decided to remove `Contracts` due to the prop-drilling. But yeah it makes sense to have it in a separate component, I'll update.

michalsmiarowski (approved):
Looking good!

PR #2442, opened 2021-04-27
KEEP Token Dashboard v1.13.0 release

r-czajkowski:
~Depends on: #2447, #2446~

KEEP Token Dashboard v1.13.0 release branch. This release will bundle the work from https://github.com/keep-network/keep-core/milestone/48?closed=1.


liz-shinn:
Not sure that this helper text is giving enough context as to what an address is:
Screen Shot 2021-04-29 at 11 50 35 AM

Maybe this should read "Enter an Ethereum address to preview its read-only version of the dashboard."

So then it hits the points:
- what do I need to put in here (eth address)
- what will it do (read-only version of dashboard)
- why (preview that eth address's version of the dashboard)

I also wonder if we shouldn't just write this in the modal vs. hiding it in helper text.

@sashatanase thoughts here?

Battenfield (dismissed):
Approving for dashboard.test 🚀

Battenfield:
Screen Shot 2021-04-30 at 12 25 09 AM
+1 on changing the text to: `Enter an Ethereum address to preview its read-only version of the dashboard.`

Another small nit, it would be great to have the tooltip over above the input instead of over it.

Battenfield (dismissed):

Battenfield:
These features are already shipped right? Should be merged in after conflicts are resolved.

pdyraga:
> These features are already shipped right? Should be merged in after conflicts are resolved.

We do not merge release branches, all fixes should be backported to `master` separately and `master` should stay as `-pre`-version bumped. 

The approach is:
- We create a release branch, bump up versions there, create PR, the action uploads the code for preview.
- We create another PR that increases the`-pre` version in `master`.
- Once we release, we merge the `-pre` version bump to `master`.
- We leave the release branch PR until the next version is released. Once we are there, we close the PR without merging and without deleting the branch.

PR #2443, opened 2021-04-29
Wallet address in url for every wallet

michalsmiarowski:
Adds wallet addres in url for every wallet


r-czajkowski:
If a user disconnects from the wallet the dapp dispatch `app/login` action with the previously selected address and redux sagas start fetching data and subscribing to events.

obraz

obraz

Not sure if it's related to those changes but anyway we need to fix this and seems reasonable to address it in this PR.

michalsmiarowski:
> If a user disconnects from the wallet the dapp dispatch `app/login` action with the previously selected address and redux sagas start fetching data and subscribing to events.
> (...)
> Not sure if it's related to those changes but anyway we need to fix this and seems reasonable to address it in this PR.

Good catch! It was indeed related to this PR! Thanks!

Ready for another review.

r-czajkowski (approved):
:shipit:

Issue #2392, opened 2021-03-23
Responsive Dashboard: Adding a modal for user


liz-shinn:
Add a modal for devices that are mobile to say that the user should use desktop for the dashboard. 

CTA - 
1. Link to the documentation
2. Join the Discord
3. Back to the website

liz-shinn:
Still workshopping this, but a potential option:
cc/ @Battenfield 

![modal-draft](https://user-images.githubusercontent.com/57226633/116468031-9d3dcd80-a82d-11eb-839c-be425296e3bf.png)

liz-shinn:
These version are in Figma here: https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=8913%3A133

PR #2444, opened 2021-04-30
Create modal window for mobile users

michalsmiarowski:
Closes: #2392

When the screen resolution width will be **1000px** or lower then the modal window will pop up and inform the user that the dashboard is not ready to be used on mobile.

image

**Note:** We are not addressing the border radius of the modal window in this PR, since we will probably round it for all modals (not only this one). This will probably be addressed in a separate PR.


solidity/dashboard/src/components/Modal.jsx
```diff
@@ -2,60 +2,81 @@ import React, { useEffect, useRef, useCallback, useState } from "react"
...
+    const onOverlayClick = useCallback(
+      (event) => {
+        if (modalOverlay.current === event.target) {
+          closeModal()
+        }
+      },
+      [closeModal]
+    )
+
+    const modalWrapperBorderStyle = isRounded
```
r-czajkowski, line 41: I think we should get rid of `isRounded` props because as I can see on Figma we are going to round corners in most of the components, so we should address it in a separate PR. For now, I suggest using a custom class name or leaving w/o rounded corners.

solidity/dashboard/src/components/Modal.jsx
```diff
@@ -70,6 +91,11 @@ export const ModalContextProvider = ({ children }) => {
   const [isOpen, setIsOpen] = useState(false)
   const [modalOptions, setModalOptions] = useState(null)
   const awaitingPromiseRef = useRef()
+  const showModalWindowForMobileUsers = useCurrentWidthLowerThanThreshold(1000)
+  const [
+    modalWindowForMobileUsersDisplayed,
+    setModalWindowForMobileUsersDisplayed,
+  ] = useState(false)
```
r-czajkowski, line 98: I think displaying the modal for mobile users should be handled in the `AppLayout` component. The `ModalContextProvider` is a tool for showing modal and it should not decide which modal will be displayed. It just provides an API to open/close a modal. I believe the `AppLayout` component is the place to handle this logic. Thougts?

solidity/dashboard/src/hooks/useCurrentWidthLowerThanThreshold.js
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from "react"
+
+const useCurrentWidthLowerThanThreshold = (widthThreshold) => {
+  const [width, setWidth] = useState(window.innerWidth)
+
+  const handleWindowSizeChange = () => {
+    setWidth(window.innerWidth)
+  }
```
r-czajkowski, line 8: The `handleWindowSizeChange` function should be rather inside the` useEffect` hook because a new instance of the `handleWindowSizeChange` is created on every render and `handleWindowSizeChange` function won't be removed in `window.removeEventListener("resize", handleWindowSizeChange)`. Thougts? 🤔

solidity/dashboard/src/components/Modal.jsx
```diff
@@ -114,6 +140,26 @@ export const ModalContextProvider = ({ children }) => {
...
+    openModal(
+      <MobileUsersModal closeModal={customModalWindowForMobileUsersClose} />,
+      {
+        closeModal: customModalWindowForMobileUsersClose,
+        hideTitleBar: true,
+        isRounded: true,
+      }
+    )
+  }
+
```
r-czajkowski, line 162: It should be wrapped by `useEffect` shouldn't it? Otherwise, if the width of the window will change and be greater than a threshold (in our case`1000`), the modal is still displayed.

r-czajkowski (changes requested):

solidity/dashboard/src/components/Modal.jsx
michalsmiarowski: I fixed it but I'm also wondering if we should display the modal more than once. Right now we display it exactly once, so for example when we resize the window below 1000 (modal will appear), then resize again to >1000px (modal will disappear) and then again resize it back to <1000px, then the modal will not appear again. I think we don't need to display it many times and one is enough (until refresh) but I'm open to suggestions.

r-czajkowski: I would prefer to display a modal each time the window resizes. But I will leave the final decision to @liz-shinn.

r-czajkowski (approved):
Looking good on my side 👍. Regarding https://github.com/keep-network/keep-core/pull/2444#discussion_r628229820 - leaving a final decision to @liz-shinn.

liz-shinn:
@michalsmiarowski A note on the text treatment – we have one word in the tope sentence that is isolated on its own line. The bottom one also looks a bit off-balance. The lines will look more balanced visually if we line-break the sentences like this:

line break

solidity/dashboard/src/components/Modal.jsx
liz-shinn: Hmm 🤔 Let's discuss on the standup call. In the scenario Michal describes it would make sense to show the modal again when the window re-sizes. But, I'm wondering if trying to show it more than once will result in it showing over and over on refreshes, or navigating to a different page?

r-czajkowski (approved):

solidity/dashboard/src/components/Modal.jsx
michalsmiarowski: Note: we discussed it on the standup call and decided to keep it like it is now, which means displaying modal only once (unless user refreshes)

r-czajkowski (approved):
LGTM! :shipit:

PR #2445, opened 2021-05-03
Load non-sensitive data from file instead of from secrets

michalinacienciala:
Non-sensitive data previously accessed by using `secrets` context is
now accessed by invoking `keep-network/load-env-variables` action and
using `env` context.
BACKGROUND:
We've been storing id of the test environment as
`KEEP_TEST_ETH_NETWORK_ID` organization secret in GitHub's settings to
allow for easy access to that value from multiple projects.
But it turned out to be problematic, as GHA in some situations redacts
workflow logs and actions outputs/inputs based on the values stored in
secrets. So if the secret holds small numeric value, it is quite
probable that at some point in time this value will be recognized as a
substring of `github.sha` or another variable and the `github.sha` will
be redacted with `***` in place of the recognized pattern. If such
redacted value is configured as input of some action, the action gets
invoked by the GHA without that input, which can cause run failure or
can mess up the workflow results.
As a solution for this problem we decided to no longer store
frequently-used, non-sensitive data such as `KEEP_TEST_ETH_NETWORK_ID`
in GitHub's secrets, but instead store them in a file kept in
`keep-network/ci` repository and read that data to `env` context in
order to be able to use it when running workflows. A
`keep-network/load-env-variables` action has been implemented to allow
for loading of the variables to `env` context in one step. The scope of
the loaded variables is limited to the job in which they were loaded.
To use the variables in other job, action `load-env-variables` needs to
be executed there as well.


michalinacienciala:
Test configuration (`-wip`, references to development branch) is now removed.
Here are the links to the successfully executed workflows which used that test configuration:
Go: https://github.com/keep-network/keep-core/actions/runs/806823489
Solidity: https://github.com/keep-network/keep-core/actions/runs/806678601
Dashboard Testnet: https://github.com/keep-network/keep-core/actions/runs/806669356

.github/workflows/client.yml
```diff
@@ -38,6 +38,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Load environment variables
+        uses: keep-network/load-env-variables@main # TODO: use @v1 once availible
+        env: 
+          CI_GITHUB_TOKEN: ${{ secrets.CI_GITHUB_TOKEN }}
+        with:
+          environment: 'ropsten'
+          ref: 'master'
```
nkuba, line 47: Could `master` be a default value for `ref` defined in the action? We wouldn't have to provide anything here. Non-default `ref` would be super rare and used mostly to test some new configurations.
```suggestion
```

.github/workflows/client.yml
```diff
@@ -38,6 +38,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Load environment variables
+        uses: keep-network/load-env-variables@main # TODO: use @v1 once availible
```
nkuba, line 42:
```suggestion
        uses: keep-network/load-env-variables@v1
```

.github/workflows/client.yml
```diff
@@ -38,6 +38,14 @@ jobs:
...
+        env: 
+          CI_GITHUB_TOKEN: ${{ secrets.CI_GITHUB_TOKEN }}
+        with:
+          environment: 'ropsten'
```
nkuba, line 46: Can we leave a `TODO` comment here that once we integrate RFC-18 changes we should pass the environment from workflow_dispatch?

.github/workflows/contracts.yml
```diff
@@ -113,6 +113,14 @@ jobs:
     steps:
```
nkuba, line 113: Same comment as for `client.yml`.

.github/workflows/dashboard-testnet.yml
```diff
@@ -37,6 +37,14 @@ jobs:
     steps:
```
nkuba, line 37: Same comment as for `client.yml`.

nkuba (changes requested):

.github/workflows/client.yml
michalinacienciala: I'm not sure if passing environment here is the way to go. There are 2 factors that will need to be taken in mind:
1. When triggering upstream build in `ci` project, we don't provide environment in form of the full name (like `ropsten`), but in form of the id that will be used in versioning of artifacts (like `rc`). And this value is then passed to downstream workflows. So the action and/or config files would need to be modified to support such value.
2. We have workflows which use the `load-env-variables` action for both `ropsten` and `alfajores`. They would need to be split...
I'll add a comment about considering substitution of the `ropsten`/`alfajores` values with `environment` input and we will decide about the approach at later stage.

nkuba (approved):

PR #2447, opened 2021-05-03
Improve helper text in explorer mode address form

r-czajkowski:
The helper text is not giving enough context as to what an address
is/what this will do. Also here we moved a tooltip over above the form
input instead of over it for better readability. Based on the comments
posted in #2442.


michalsmiarowski (approved):
Looking good!

PR #2448, opened 2021-05-03
Update keep-common to Latest

beaushinkle:
We'd like to update keep-ecdsa for [this](https://github.com/keep-network/keep-ecdsa/pull/756#discussion_r619166506), but that puts ecdsa and core on different versions of common. To remedy that, we'll update core, and then update both core and common in ecdsa.

Tagging @Shadowfiend for review


beaushinkle:
closing. behavior covered by: https://github.com/keep-network/keep-core/pull/2449

Issue #2334, opened 2021-02-15
Broken links in contract docs


RhettCreighton:
The links to contract documentation are broken here

https://github.com/keep-network/keep-core/tree/master/solidity/docs

PR #2451, opened 2021-05-05
Fix broken links in solidity/docs

r-czajkowski:
Closes: #2334 

This README file is a part of the abandoned Solidity API docs CI build so is no longer valid. More context here: https://www.flowdock.com/app/cardforcoin/tech/threads/-9kWxCim5nVgKG4TttURbggBKgO


Shadowfiend (approved):
:fire:

PR #2453, opened 2021-05-05
Execute chain of workflows by manual triggering of upstream build

michalinacienciala:
This commit is part of a bigger work, spanning multiple repositories.
The main purpose of the work is to automate the process of workflows`
executions during product deployment, so that the workflows would be
triggered in a particular logical order and at the correct moment. By
that we mean:
* workflows requiring the presence of artifacts created by other
  workflows would not start before those upstream workflows are
  finished
* successfully executed workflows would automatically kick-off
  downstream workflows specified in the config file
* if any workflow in that chain of executions fails, the downstream
  workflows will not be triggered

To achieve the above requirements, following changes were introduced:
* A repository (called `ci`) managing the order of workflow executions
  was created
* The `config/config.json` file in `ci` repo describes the
  upstream/downstream relations between product modules and workflows
* The `Main` GH Actions workflow in `ci` repo allows for manual
  triggering of workflow deploying specified module, by providing the
  details of its upstream builds (to trigger deploy of the most
  upstream module, `Main` must be triggered with empty
  `upstream_builds` input)
* The triggering of module workflows by the `ci`'s `Main` workflow
  happens thanks to a `keep-network/run-workflow` action working as an
  intermediary (`Main` calls the `run-workflow`, which triggers
  workflow deploying the specified module)
* All NPM modules built as a part of the release process get versioned
  with `<base-version>-<environment>.<build-number>+<branch>.<commit>`
  version (thanks to `keep-network/upstream-builds-query` action that
  reads versions of older deployments and thanks to
  `keep-network/npm-version-bump@v2` action that bumps-up the version
  of the new artifact)
  (note: tagging of artifacts and versioning of docker images will be
  done at the later stage, outside this PR)
* Deployment workflows that are about to successfully finish will be
  executing the `keep-network/notify-workflow-completed` action as the
  last step - the action will then trigger the `ci`'s `Main` workflow
  with appropriate inputs
* Module workflows configuration has been adjusted to support the above
  features
* Conditions of execution of particular jobs/steps have been modified
  - after the changes contracts migration and deployments of artifacts
  happens only for workflows triggered by the `workflow-dispatch`
  event; actions like linting and sec scans are no longer executed for
  workflows triggered by `workflow-dispatch`
* Additionally, a couple of other changes were introduced to improve
  the process of build management:
  - listing of workflow metadata
  - unification of workflows' code
  - decreased amount of the gas sent for Ropsten env (to fix tests
    failing with `... exceeded the block limit`)
  - no longer pushing migrated contracts artifacts to GCP Bucket (see
    #2450 for more details)

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -12,6 +12,18 @@ on:
       - "solidity/dashboard/**"
   pull_request:
   workflow_dispatch:
+    inputs:
+      environment:
+        description: 'Environment for workflow execution'
+        required: false
+        default: 'dev'
+      upstream_builds:
+        description: 'Upstream builds'
+        required: false
+      ref:
+        description: 'Git reference to checkout (e.g. branch name)'
+        required: false
+        default: 'master'
 
 jobs:
   client-detect-changes:
@@ -19,15 +31,31 @@ jobs:
     outputs:
       path-filter: ${{ steps.filter.outputs.path-filter }}
     steps:
-    - uses: actions/checkout@v2
-      if: github.event_name == 'pull_request' 
-    - uses: dorny/paths-filter@v2
-      if: github.event_name == 'pull_request' 
-      id: filter
-      with:
-        filters: |
-          path-filter:
-            - './!((docs|infrastructure|scripts|token-stakedrop|(solidity/dashboard))/**)'
+      - name: Show workflow metadata
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+            echo "Workflow dispatched on branch: ${{ github.ref }}"
+            echo "environment = ${{ github.event.inputs.environment }}"
+            echo "upstream_builds = ${{ github.event.inputs.upstream_builds }}"
+            echo "ref = ${{ github.event.inputs.ref }}"
+            echo "sha = ${{ github.sha }}"
+
+      - name: Show workflow metadata
+        if: github.event_name != 'workflow_dispatch'
+        run: |
+            echo "ref = ${{ github.ref }}"
+            echo "sha = ${{ github.sha }}"
+
+      - uses: actions/checkout@v2
+        if: github.event_name == 'pull_request'
+
+      - uses: dorny/paths-filter@v2
+        if: github.event_name == 'pull_request'
+        id: filter
+        with:
+          filters: |
+            path-filter:
+              - './!((docs|infrastructure|scripts|token-stakedrop|(solidity/dashboard))/**)'
 
   client-build-test-publish:
     needs: client-detect-changes
@@ -79,9 +107,7 @@ jobs:
             gotestsum
 
       - name: Login to Google Container Registry
-        if: |
-          github.ref == 'refs/heads/master'
-            && github.event_name != 'pull_request'
+        if: github.event_name == 'workflow_dispatch'
         uses: docker/login-action@v1
         with:
           registry: ${{ env.GCR_REGISTRY_URL }}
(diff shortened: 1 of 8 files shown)


michalinacienciala:
Couple of things still TODO or open for consideration:
* What to do with Token Dashboard workflow? Currently it fails when it is triggered by events different than `workflow_dispatch`, because it requires information about upstream builds. Can we build the dashboard (without publishing) without referencing the data from upstream builds?
* Do we still need the environment protection switched on for contracts deployments? Maybe we could get rid of it, as triggering of deployments will be done as a result of manual action (triggering of `ci`'s `Main` workflow)
* I added steps listing workflow metadata to the jobs that are executed at the beginning of workflows. But now I consider moving those steps to a separate job, executed in parallel with others
* Maybe we could move the `Notify CI...` step in `Solidity` workflow to the end of `contracts-migrate-and-publish-ethereum` job instead of having it at the end of the workflow (after building of initcontainer)? We are not using initcontainer artifacts in other jobs in the chain, so maybe we don't need to wait for the initcontainer publishing?

.github/workflows/client.yml
```diff
@@ -12,22 +12,50 @@ on:
...
   workflow_dispatch:
+    inputs:
+      environment:
+        description: 'Environment for workflow execution'
+        required: false
+        default: 'dev'
+      upstream_builds:
+        description: 'Upstream builds'
+        required: false
+      ref:
```
nkuba, line 23: According to RFC-18 this property should be named `upstream-ref`.

.github/workflows/client.yml
```diff
@@ -12,22 +12,50 @@ on:
...
+            echo "environment = ${{ github.event.inputs.environment }}"
+            echo "upstream_builds = ${{ github.event.inputs.upstream_builds }}"
+            echo "ref = ${{ github.event.inputs.ref }}"
+            echo "sha = ${{ github.sha }}"
+
+      - name: Show workflow metadata
+        if: github.event_name != 'workflow_dispatch'
+        run: |
+            echo "ref = ${{ github.ref }}"
+            echo "sha = ${{ github.sha }}"
```
nkuba, line 47: Weren't we using these just for debugging? Do we want to keep them?

michalinacienciala: Initially I introduced them for debugging purposes during feature development, but now I'm thinking of keeping them, as they provide better visibility of why the workflow was started and what parameters were used (GitHub Actions don't list the workflow's dispatch parameters anywhere afaik).

.github/workflows/client.yml
michalinacienciala: Right, this will be delivered in following PRs:
https://github.com/keep-network/ci#3
https://github.com/keep-network/run-workflow/pull/2
https://github.com/keep-network/notify-workflow-completed#1
https://github.com/keep-network/keep-core/pull/2466
https://github.com/keep-network/keep-ecdsa/pull/786
https://github.com/keep-network/tbtc/pull/793
https://github.com/keep-network/tbtc-dapp/pull/395
https://github.com/keep-network/tbtc.js/pull/123

michalinacienciala:
After discussing some of the open issues with Kuba on call, we agreed on following:

> * What to do with Token Dashboard workflow? Currently it fails when it is triggered by events different than `workflow_dispatch`, because it requires information about upstream builds. Can we build the dashboard (without publishing) without referencing the data from upstream builds? (similar problem applies to `tbtc-dapp` and `tbtc.js` builds)

If Token Dashboard workflow gets invoked manually (workflow_dispatch event), it will update the dependencies by checking upstream modules versions and then running `npm install --save-exact...`. However, when Token Dashboard is executed as a result of `push` or `pull_request` event, we will update the dependencies based on values configured in 'package.json', by running `npm update...` command (see 8d82c3b8c813808e48480fa13d0ecf787350ff83)


> * Do we still need the environment protection switched on for contracts deployments? Maybe we could get rid of it, as triggering of deployments will be done as a result of manual action (triggering of `ci`'s `Main` workflow)

No we don't. For now we will leave the `environment: keep-test` in the workflow .yml files, but we plan to turn off the protection of the environments after merging `rfc-18/integration-2` branch to master.

> * I added steps listing workflow metadata to the jobs that are executed at the beginning of workflows. But now I consider moving those steps to a separate job, executed in parallel with others

The steps listing workflow metadata will be removed in one of the latests commits in `rfc-18/integration-2` branch.

.github/workflows/client.yml
michalinacienciala: Removed in ce1219c404fecaeecc49904dfe79cfc552352f3d

solidity/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@keep-network/keep-core",
-  "version": "1.8.0-pre.0",
+  "version": "1.8.0-rc",
```
nkuba, line 3: This doesn't match `package.json`.

nkuba (changes requested):

solidity/package-lock.json
michalinacienciala: Fixed in e6254d503bf7b6ab559501bffe6f4e7cd369788c.

nkuba (approved):

pdyraga (approved):

PR #2471, opened 2021-05-18
KEEP Token Dashboard v1.14.0 release

r-czajkowski:
KEEP Token Dashboard v1.14.0 release branch. This release will bundle the work from https://github.com/keep-network/keep-core/milestone/50?closed=1


michalinacienciala:
@r-czajkowski , there is a global outage of GH Actions today, that's why no checks were executed. You can check current state of the incident [here](https://www.githubstatus.com/incidents/m16jzl31gnqt). You may need to re-run following workflows once the incident gets resolved:
https://github.com/keep-network/keep-core/actions/runs/853684933
https://github.com/keep-network/keep-core/actions/runs/853684934
https://github.com/keep-network/keep-core/actions/runs/853684935
https://github.com/keep-network/keep-core/actions/runs/853684937

michalsmiarowski (changes requested):
Looks like there is a bug with modal window titles. They are not displaying probably because of changes from #2444

michalsmiarowski (approved):
The modal window's bug was fixed in #2486. Tested the preview locally and it's looking good 👍

infrastructure/kube/keep-prd/keep-dapp-token-dashboard-deployment.yaml
```diff
@@ -21,6 +21,6 @@ spec:
     spec:
       containers:
       - name: keep-dapp-token-dashboard
-        image: keepnetwork/token-dashboard:v1.12.0
+        image: keepnetwork/token-dashboard:v1.14.0
```
pdyraga, line 24: Let's prepare a separate PR to increase the `master` `version` in `package.json` to `1.15.0-pre` and that changes this line to `v1.14.0` just like it is done here.

PR #2474, opened 2021-05-20
Compile contracts before publishing

michalinacienciala:
`NPM` workflow will be used to publish the contracts needed for the
setup of development environment to NPM registry. Before the change we
were not compiling the contracts before publishing, which was a problem,
because this way not all the files needed by the downstream workflows
were being published. As a result of contracts compilation system will
generate `*.json` files needed by the later workflows. We will then
place them in the `artifacts` directory. By the `package.json`
configuration this directory is being picked by the publish step along
with the `contracts/**/*.sol` files.
Apart from this change, couple of small updates were introduced to `NPM`
workflow, mainly update to higher versions of used external actions.

Corresponding changes in `keep-ecdsa` project: https://github.com/keep-network/keep-ecdsa/pull/794

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -4,21 +4,27 @@ on:
   push:
     branches:
       - master
+      - "rfc-18/**" # TODO: Remove after testing
     paths:
       - 'solidity/contracts/**'
       - 'solidity/package.json'
       - 'solidity/package-lock.json'
+      - '.github/workflows/npm.yml' # TODO: Remove after testing
 
 jobs:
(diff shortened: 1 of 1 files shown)


michalinacienciala:
> TODO: once I get the OK for rest of the proposed changes, I will remove the pieces of code marked with # TODO

TODOs removed. Here is the last run before that change, with `--dry-run` in the publish step: https://github.com/keep-network/keep-core/runs/2654389484?check_suite_focus=true

nkuba (approved):

PR #2492, opened 2021-06-10
Refactor workflows to add deployment of contracts migrated on Celo

michalinacienciala:
Workflows have been refactored so that Celo-related artifacts could be
deployed in the chain of automatic execusions triggered by the manual
dispatch of the `Main` workflow in the `ci` project (with `environment`
set to `alfajores`).

Similar changes have been introduced to keep-core and tbtc
workflows. See:
https://github.com/keep-network/keep-ecdsa/pull/823
https://github.com/keep-network/tbtc/pull/804

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -53,13 +53,11 @@ jobs:
 
       - name: Load environment variables
         uses: keep-network/load-env-variables@v1
+        if: github.event_name == 'workflow_dispatch'
         env: 
           CI_GITHUB_TOKEN: ${{ secrets.CI_GITHUB_TOKEN }}
         with:
-          # TODO: Consider passing of `environment` input instead of using 
-          # hardcoded value. Would require some rework in action's code or
-          # in config files.
-          environment: 'ropsten' 
+          environment: ${{ github.event.inputs.environment }}
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
@@ -99,7 +97,17 @@ jobs:
           username: _json_key
           password: ${{ secrets.KEEP_TEST_GCR_JSON_KEY }}
 
+      - name: Build Docker Runtime Image
+        if: github.event_name != 'workflow_dispatch'
+        uses: docker/build-push-action@v2
+        env:
+          IMAGE_NAME: 'keep-client'
+        with:
+          labels: revision=${{ github.sha }}
+          push: false
+
       - name: Build and publish Docker Runtime Image
+        if: github.event_name == 'workflow_dispatch'
         uses: docker/build-push-action@v2
         env:
           IMAGE_NAME: 'keep-client'
@@ -109,7 +117,7 @@ jobs:
           # We don't use TAG yet, will be added at later stages of work on RFC-18.
           tags: ${{ env.GCR_REGISTRY_URL }}/${{ env.GOOGLE_PROJECT_ID }}/${{ env.IMAGE_NAME }}
           labels: revision=${{ github.sha }}
-          push: ${{ github.event_name == 'workflow_dispatch' }}
+          push: true
       
       - name: Notify CI about completion of the workflow
         if: github.event_name == 'workflow_dispatch'
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -87,14 +87,6 @@ jobs:
             go-build-env \
             gotestsum
 
-      - name: Login to Google Container Registry
-        if: github.event_name == 'workflow_dispatch'
-        uses: docker/login-action@v1
-        with:
-          registry: ${{ env.GCR_REGISTRY_URL }}
-          username: _json_key
-          password: ${{ secrets.KEEP_TEST_GCR_JSON_KEY }}
-
       - name: Build Docker Runtime Image
         if: github.event_name != 'workflow_dispatch'
         uses: docker/build-push-action@v2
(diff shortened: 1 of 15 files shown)


.github/workflows/client.yml
```diff
@@ -99,7 +97,17 @@ jobs:
           username: _json_key
           password: ${{ secrets.KEEP_TEST_GCR_JSON_KEY }}
 
+      - name: Build Docker Runtime Image
```
nkuba, line 100: Why are we splitting the docker building step? The main difference is pushing, can't we do this with `push: ${{ github.event_name == 'workflow_dispatch' }}` as we did before?

solidity/truffle-config.js
```diff
@@ -48,7 +48,7 @@ module.exports = {
 
     alfajores: {
       provider: function () {
-        const kit = Kit.newKit("https://alfajores-forno.celo-testnet.org")
+        const kit = Kit.newKit(process.env.CELO_HOSTNAME)
```
nkuba, line 51: Maybe we should expect the same-named env variables for all networks.

So instead of having `ETH_HOSTNAME` and `CELO_HOSTNAME` we could expect `CHAIN_API_URL`,
and instead of `CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY` and `CONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY` we could expect `CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY`.

This will make the migrations more generic, hence from the CI workflows perspective, we could handle both Ethereum and Celo migrations in one step.

.github/workflows/contracts.yml
```diff
@@ -247,7 +243,7 @@ jobs:
       - name: Build and publish initcontainer
         uses: docker/build-push-action@v2
         env:
-          IMAGE_NAME: 'initcontainer-provision-keep-client'
+          IMAGE_NAME: initcontainer-provision-keep-client-${{ github.event.inputs.environment }}
```
nkuba, line 246: Renaming the image will require aligning the kubernetes configs.

.github/workflows/contracts.yml
```diff
@@ -156,12 +148,21 @@ jobs:
...
             ${{ secrets.KEEP_TEST_ETH_CONTRACT_OWNER_PRIVATE_KEY }}
         run: npx truffle migrate --reset --network $TRUFFLE_NETWORK
+
+      - name: Migrate contracts on Celo
+        if: github.event.inputs.environment == 'alfajores'
+        env:
+          CELO_HOSTNAME: ${{ secrets.KEEP_TEST_CELO_HOSTNAME }}
+          CONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY: |
+            ${{ secrets.KEEP_TEST_CELO_CONTRACT_OWNER_PRIVATE_KEY }}
+        run: npx truffle migrate --reset --network $TRUFFLE_NETWORK
```
nkuba, line 165: We could handle these two in a one step if we update env variables names expected by the truffle configuration (see the other comment).

nkuba (changes requested):

.github/workflows/contracts.yml
michalinacienciala: Are those configs stored on the repo? Could you guide me where to find them?

.github/workflows/contracts.yml
michalinacienciala: My understanding is that if we'd switch to setting `CHAIN_API_URL` and `CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY` environment variables, we would still need to populate those variables with different values, depending on the network used. Is there a way to do that in one step?
Maybe there is a way to do something like this?:
```
CHAIN_API_URL: if github.event.inputs.environment != 'alfajores' then set ${{ secrets.KEEP_TEST_ETH_HOSTNAME }}; else set ${{ secrets.KEEP_TEST_CELO_HOSTNAME }}; 
```
(I don't know how to do that with [operators supported by GitHub](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#operators))

solidity/truffle-config.js
michalinacienciala: I did unified the variables (see 76adcf0cbc3f47ab59cdc1bb3d774799de51705c), but I'm not sure if it will allow us to handle both migrations in one step (see my comment below).

.github/workflows/contracts.yml
nkuba: Please grep ocurrences under `./infrastructure/kube/keep-dev` and `./infrastructure/kube/keep-test`.

.github/workflows/contracts.yml
nkuba: I'm afraid it's not yet supported by GHA 😞  
See: https://github.com/actions/runner/issues/409

.github/workflows/client.yml
nkuba: > Execute the step reading the env variables even when the workflow was triggered by PR or push. As environment specify ropsten (reading of the values would not have much consequences)

Maybe we could use `local` as the default environment, we would add a `local.env` file that will hold some dummy variables?

nkuba: On the second thought, let's not overengineer it. The current solution is fine. Thanks for clarifying everything in your first comment!

I would just move `Build Docker Runtime Image` above `Login to Google Container Registry`, so it keeps the stages more sperated.

solidity/truffle-config.js
nkuba: I was also thinking about loading secrets for the environemnt based on the environment property.  So we would name the GH environments the same as we expect workflow dispatch `environment` property to be set.

.github/workflows/client.yml
michalinacienciala: Order of the steps changed in c0b590d5982c38392a7822d0eaa43783ab4c2682.

solidity/truffle-config.js
michalinacienciala: The downside of this solution would be the need to repeat the same settings across different repositories. Sticking with current solution seems to be the least complicated.

.github/workflows/contracts.yml
michalinacienciala: I've changed the Kubernetes config. and also modified the image name in the workflow to use chain name in the suffix instead of the environment/network. In order for this to work, https://github.com/keep-network/ci/pull/6 needs to be merged to master, so please merge that PR before merging the current one.

nkuba (approved):

PR #2501, opened 2021-06-16
CI workflows - Move cache instead of copying

michalinacienciala:
In `keep-ecdsa` we've seen a number of workflows failing with `No space
lefton device` error.
The problem was a combination of a big cache size and significant memory
allocation after building Docker image.
According to moby/buildkit#1850:
"At the moment caches are copied over the existing cache so it keeps
growing".
As a temporary fix (until issue gets fixed by GH), we introduced a step
that moves the cache (similarily to how it's described in
moby/buildkit#1896) instead of copying.
Although we havent seen the problems with cache size yet in other
projects than `keep-ecdsa`, we're applying the solution across the
repositories to decrease the likelyhood of encountering problem in the
future.

See also:
https://github.com/keep-network/keep-ecdsa/pull/822
https://github.com/keep-network/tbtc/pull/807
https://github.com/keep-network/tbtc-dapp/pull/398


.github/workflows/client.yml
```diff
@@ -108,6 +108,16 @@ jobs:
           tags: ${{ env.GCR_REGISTRY_URL }}/${{ env.GOOGLE_PROJECT_ID }}/${{ env.IMAGE_NAME }}
           labels: revision=${{ github.sha }}
           push: ${{ github.event_name == 'workflow_dispatch' }}
+
+      - # Temp fix - move cache instead of copying (added below step and
+        # modified value of `cache-to`).
+        # https://github.com/docker/build-push-action/issues/252
+        # https://github.com/moby/buildkit/issues/1896
+        # Without the change some jobs were failing with `no space left on device`
+        name: Move cache
```
nkuba, line 117: Let's move it to be the very last step in workflows.

nkuba (changes requested):

.github/workflows/client.yml
michalinacienciala: Done, see ec61e35f2ff5f1ea6f51c631464e94be2ee0c1e8.

nkuba (approved):

PR #2522, opened 2021-06-28
Coverage pool- fetching data

r-czajkowski:
This PR adds support for fetching coverage pool data from the contracts.

**The main changes**
* `web3` library that provides wrappers for external web3 libraries. These wrappers will help to keep consistent API for external web3 libraries in the entire dapp. This is an introduction to the new approach to how we want to fetch the data from the on-chain contracts. It will make writing unit tests and for example migrations from `web3` to the `ethers.js ` easier.
* `keep` library that provides all contracts that are used in the KEEP Token Dashboard and services(currently provides `CoveragePoolV1` and `ExchangeService`). It will help to maintain and add new services like eg. `CoveragePoolV1` which are responsible for fetching data. It will also make writing unit tests easier. In the future, we will migrate all functions from the `services` dir to this library. It also requires some reorg because some utils from the dapp are used in this library- it should be independent.
* redux sagas that fetching cov pool data via the new `keep` library.
* update the `start_dashboard` script- build and link the `coverage-pool` contracts. Please see https://github.com/keep-network/coverage-pools/pull/103.

### DIFF
--- a/solidity/dashboard/src/sagas/coverage-pool.js
+++ b/solidity/dashboard/src/sagas/coverage-pool.js
@@ -60,7 +60,7 @@ function* fetchCovPoolData(action) {
       balanceOf
     )
     const estimatedKeepBalance = yield call(
-      Keep.coveragePoolV1.estimatedCorateralTokenBalance,
+      Keep.coveragePoolV1.estimatedCollateralTokenBalance,
       shareOfPool
     )
 
@@ -133,7 +133,7 @@ export function* subscribeToCovTokenTransferEvent() {
     )
 
     const estimatedKeepBalance = yield call(
-      Keep.coveragePoolV1.estimatedCorateralTokenBalance,
+      Keep.coveragePoolV1.estimatedCollateralTokenBalance,
       shareOfPool
     )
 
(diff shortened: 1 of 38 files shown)


solidity/dashboard/src/sagas/coverage-pool.js
```diff
@@ -1,27 +1,182 @@
...
+    } else if (isSameEthAddress(to, ZERO_ADDRESS)) {
+      updatedCovTotalSupply = add(covTotalSupply, value)
+    }
+
+    let arithmeticOpration = null
+    if (isSameEthAddress(address, from)) {
+      arithmeticOpration = sub
+    } else if (isSameEthAddress(address, to)) {
+      arithmeticOpration = add
+    }
```
michalsmiarowski, line 124: I think the same piece of code is placed in the `sagas/subscriptions.js` (`observeKeepTokenTransfer` method). Maybe we could actually rewrite it as a function inside `utils`? Not sure how to name it tho 🤔 Thoughts?

michalsmiarowski (changes requested):
Overall very good. `Web3jsContractWrapper` wrapper is a fantastic idea! Finally no more `methods.<methodName>(args..).call()` which was not that readable 😅

Also +1 for the `<OnlyIf>` component 👍 Simple but definitely very usable!

Left some comments, which may improve the readability of the code.

solidity/dashboard/src/sagas/coverage-pool.js
r-czajkowski: Can we address it in a follow-up work? I've thought about it, but I don't have a good idea of how to deal with it in the most elegant way.

michalsmiarowski (approved):
LGTM

PR #2535, opened 2021-07-14
Coverage pools banner

r-czajkowski:
This PR adds a banner about coverage pools on the Overview page.

obraz


solidity/dashboard/src/components/coverage-pools/LearnMoreBanner.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+const LearnMoreBanner = ({ onClose }) => {
+  return (
+    <Banner className="banner banner--coverage-pools">
+      <Banner.CloseIcon onClick={onClose} />
+      <div className="flex row space-between full-center">
+        <Banner.Icon icon={Icons.CoveragePool} />
+        <Banner.Title className="h3 text-white">
+          Deposit KEEP in the coverage pool to
+          <br />
```
michalsmiarowski, line 14: Hmm i don't think using `<br />` tag here is a good practise based on [`<br>` documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br#accessibility_concerns) (see **Accessibility concerns**)

Maybe the better idea would be to keep it as default in the div and and margin/padding on the right.

solidity/dashboard/src/components/coverage-pools/LearnMoreBanner.jsx
```diff
@@ -0,0 +1,28 @@
...
+          secure the network and earn rewards.
+        </Banner.Title>
+        <NavLink
+          to="coverage-pools/how-it-works"
+          className="btn btn-tertiary btn-lg"
+        >
+          learn more
+        </NavLink>
```
michalsmiarowski, line 22: Comparing it to the design in figma i think the banner looks slighty different - mainly because the Banner title is centered while on figma it looks like the title is closer to the icon (so both - icon and title - are kind of closer to the left side) with the `Learn More` button more on the right side.

So I would add margin-left auto to the NavLink to really make it stick to the right side and then add some proper margins to the `Banner.Title` to make it not exactly on center, but almost on center. You see what i'm sayin'? 😅 

When doing that we can also remove the `space-between` class from the parent.

solidity/dashboard/src/components/coverage-pools/LearnMoreBanner.jsx
```diff
@@ -0,0 +1,28 @@
...
+      <Banner.CloseIcon onClick={onClose} />
+      <div className="flex row space-between full-center">
+        <Banner.Icon icon={Icons.CoveragePool} />
+        <Banner.Title className="h3 text-white">
+          Deposit KEEP in the coverage pool to
+          <br />
+          secure the network and earn rewards.
+        </Banner.Title>
+        <NavLink
+          to="coverage-pools/how-it-works"
```
michalsmiarowski, line 18: I think you missed `/` at the beginning here. But I think that we will actually put it in a variable when we create `How it works` page so this comment might be pointless.

solidity/dashboard/src/css/banner.less
```diff
@@ -88,4 +88,33 @@
...
+    .banner__icon {
+        width: auto;
+    height: auto;
+    }
+
+    .banner__close-icon {
+        width: 16px;
+        height: 16px;
+        top: 16px;
+        right: 16px;
```
michalsmiarowski, line 114: I think we should use `rem` whenever possible, so i would change those `pixels` to `rem` - what do you think?

solidity/dashboard/src/css/banner.less
```diff
@@ -88,4 +88,33 @@
 .banner--warning {
     border-radius: 8px;
     background-color: @color-yellow-30;
+}
+
+.banner--coverage-pools {
+    width: 100%;
+    max-width: none;
+    background-color: @color-brand-violet-80;
+    border-radius: 8px;
```
michalsmiarowski, line 97: Same here if you agree with above

michalsmiarowski (changes requested):
Looking really good!!

Left some comments to look into though 😄

solidity/dashboard/src/components/coverage-pools/LearnMoreBanner.jsx
r-czajkowski: I used a `<p>` tag. I think this is the best that we can do according to the [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br#accessibility_concerns).

michalsmiarowski (approved):
Looking good!

PR #2542, opened 2021-07-19
Timeline component

r-czajkowski:
This PR adds a `Timeline` component based on the Figma views. There are a few things that we can do to make a `Timeline` component more flexible. Eg. this PR will add support for displaying the `Timlnie` component on the left side but in the future, we may want to display events on alternating sides. Based on the Figma views, this is enough and we can add more features as needed.

obraz

obraz


solidity/dashboard/src/css/timeline.less
```diff
@@ -0,0 +1,101 @@
...
+
+.timeline {
+  position: relative;
+
+  &__element {
+    position: relative;
+    min-width: 200px;
+    padding: 1.5rem 2rem;
+    border: 1px solid #AE96E1; // @color-brand-violet-40
+    border-left: @left-border-width solid #633EB1; // @color-brand-violet-80
```
michalsmiarowski, line 12: Add these brand-violet colors to the `vars.scss` (i added some of them in my PR also):

https://github.com/keep-network/keep-core/blob/e05a336380ac46563ab490d736757b056899fc4d/solidity/dashboard/src/css/vars.less#L10

solidity/dashboard/src/pages/coverage-pools/CoveragePoolPage.jsx
```diff
@@ -71,6 +72,43 @@ const CoveragePoolPage = ({ title, withNewLabel }) => {
...
+            </span>
+          </Timeline.Element>
+
+          <Timeline.Element>
+            <Timeline.ElementBreakpoint active>3</Timeline.ElementBreakpoint>
+            <h4 className="text-violet-80">Claim tokens</h4>
+            <span className="text-grey-60">
+              <strong>
+                You have a 2 day grace period to claim your tokens and rewards.
+              </strong>
```
michalsmiarowski, line 103: I think we can extract the `<h4>` element to a separate component like `<Timeline.ElementTitle>`.

I was thinking that we could also to the same with the `<span>` and extract it to something like ` <Timeline.ElementDescription>` since both Timelines (the one on `How it works` subpage and the one in modal window) are pretty similar so I don't think we need to put it all as `children` element for now (but we can also keep the possibility to specify `children` here and keep `title` and `description` optional as it might be more reusable in the future.

solidity/dashboard/src/components/Timeline.jsx
```diff
@@ -0,0 +1,51 @@
...
+    </div>
+  )
+}
+Timeline.Element = TimelineElement
+
+Timeline.ElementBreakpoint = ({
+  children,
+  className = "",
+  // start, center
+  // TODO add support for a start position.
```
michalsmiarowski, line 39: Just pointing out that there is a `TODO` here. Not sure if it's done, or you planning to do it later, but if it's done then we should remove it.

solidity/dashboard/src/components/Timeline.jsx
```diff
@@ -0,0 +1,51 @@
...
+    <TimelineContext.Provider value={{ position: position }}>
+      <div className="timeline">{children}</div>
+    </TimelineContext.Provider>
+  )
+}
+const TimelineElement = ({ position, children }) => {
+  const { position: timelinePosition } = useTimelineContext()
+  const _position = position || timelinePosition
+  return (
+    <div className={`timeline__element timeline__element--${_position}`}>
```
michalsmiarowski, line 28: I think we should get used to writing a classNames as property when creating new components and adding it immediately to (preferably) every div possible 😅  This is some light idea that would be great to discuss, but it could make the components even more reusable.

We could just keep it in one property as object (e.g classes) and assign them to the divs inside components.

For example:
```
const TimelineElement = ({ position, **classes**, children }) => {
  const { position: timelinePosition } = useTimelineContext()
  const _position = position || timelinePosition
  return (
    <div className={`timeline__element timeline__element--${_position} ${classes.root ? classes.root : ''}`}>
      {children}
    </div>
  )
}
```
If later we add the component title, then we do `${classes.title ? classes.title : ''}`.

And when using the components we will just write:
```
<TimelineElement classes={{
  root: 'root-class-name',
  title: 'title-class-name',
}} />
```

This idea is actually something i got from `Angular Material` when they use the classes similarly like here.


So, not saying that you have to do it like that, but I'm interesting in your opinion about this solution.

michalsmiarowski: Also the background of it on Figma is violet, but we could actually specific it with the added class like i mentioned above and keep it with white background as default.

michalsmiarowski (changes requested):
It's a pleasure to read a high quality code like this 💎 

Left some comments anyway to consider before the merge 👀

solidity/dashboard/src/css/timeline.less
r-czajkowski: I've added the brand-violet colors in my previous PRs please see 27e9b716b138146ac4b81615b2f5d970d9cad48c. I need to merge `cov-pools-dapp` branch and I'll update here.

michalsmiarowski: Kk, i forgot about it. Resolving this convo then.

solidity/dashboard/src/components/Timeline.jsx
r-czajkowski: > Also the background of it on Figma is violet, but we could actually specific it with the added class like i mentioned above and keep it with white background as default.

I believe the parent component should define the background. So in this case the section tag should set the background. eg

```
<section className="tile **bg-vilolet**">
 <Timeline>
  ...
 </Timeline>
</section>
```

solidity/dashboard/src/components/Timeline.jsx
r-czajkowski: I'm going to address it in follow-up work. I'll add this when I work on a [timeline in modal](https://www.figma.com/file/Su61MwHLrQn3ld24q85pMx/Token-Dashboard-dApp?node-id=9934%3A33738).

solidity/dashboard/src/pages/coverage-pools/CoveragePoolPage.jsx
r-czajkowski: TBH I think it's unnecessary. Passing children give us the most flexibility. Based on few other libs they define only a kind of `Content` component where we can pass children. In our cases, the timelines on the `How it works page` and in `modal` are slightly different so I think it's hard to define common styles. Probably on another page, the timeline would be different.

Also, I would consider getting rid of common styles(like violet borders) for `Timeline.Element` component- I think it should be defined by parent component inside `Timeline.Element` component.

solidity/dashboard/src/components/Timeline.jsx
r-czajkowski: IMO (ofc I'm open to discussion 😆) passing `classes` props only make sense in non-compound style components eg.:
```
const Component = ({ classes, title, subtitle, description, }) => {
  <section className={`default-style ${classes.root}`}>
    <h1 className={`default-title-style ${classes.title}`}>{title}</h1>
    <h3 className={`default-subtitle-style ${classes.subtitle}`}>{subtitle}</h3>
    <span className={`default-description-style ${classes.description}`}>
      {description}
    </span>
  </section>
}
```

In compound components, we are able to add a custom class name that can override default styles. Thoughts?

I missed the `className` prop for the `Timeline.Element` component- I'll update.

r-czajkowski: Regarding background: I temporarily added the `Timeline` just to show how it looks like. I'll create the right background and component grid in #2543.

solidity/dashboard/src/pages/coverage-pools/CoveragePoolPage.jsx
michalsmiarowski: I see your point and I think you are right. Let's keep it that way then.

And i also agree with the seconds paragraph too.

solidity/dashboard/src/components/Timeline.jsx
```diff
@@ -0,0 +1,51 @@
+import React, { useContext } from "react"
+
+const TimelineContext = React.createContext({
+  // May be useful in the future, eg. if we would like to be able to display the
+  // timeline events on alternating sides.
+  position: "left",
```
michalsmiarowski, line 6: Can we specify in the comment what are the values available here? eg. `left` and `right`? Maybe we could event create constants for this the for the `elementBreakpoint` position?

solidity/dashboard/src/css/timeline.less
```diff
@@ -0,0 +1,101 @@
...
+      content: '';
+      position: absolute;
+      width: 2px;
+      left: calc(-3rem - 8px);
+      height: calc(100% + 2rem);
+      top: -1rem;
+      bottom: -1rem;
+      overflow: visible;
+    }
+  }
```
michalsmiarowski, line 65: The dashed line is not looking good in certain cases (tested only on Chrome).

Examples:
<details>
  <summary>For 1 element:</summary>

image
</details>

<details>
  <summary>For 5 element:</summary>

image
</details>

What i propose is to write the line for each element only below it (instead of 50% on top and 50% on bottom). I refactor the code here and there and i ended up with something like this:

```
    // Drawing the dashed line for the first element. NOTE: only valid when timeline
    // breeakpoints are centered relative to the event content.
    &:nth-of-type(1) {
        &::before {
          top: 50%;
          bottom: 0;
          height: calc(100% + 2rem);
          .draw-dashed-vertical-line(@color-grey-30);
        }
    }

    // Drawing the dashed line for the last element. NOTE: only valid when timeline
    // breeakpoints are centered relative to the event content.
    &:nth-last-of-type(1) {
      &::before {
        top: 0;
        bottom: 50%;
        height: unset;
        background: transparent;
      }
    }

    // Drawing the dashed line for the rest elements between the first and last
    // element. NOTE: only valid when timeline breeakpoints are centered relative to
    // the event content.
    &::before {
      .draw-dashed-vertical-line(@color-grey-30);
      content: '';
      position: absolute;
      width: 2px;
      left: calc(-3rem - 8px);
      height: calc(100% + 2rem);
      top: 50%;
      bottom: 0;
      overflow: visible;
    }
  }
```

It makes the lines look ok in both cases that i mentioned.

michalsmiarowski (changes requested):

solidity/dashboard/src/components/Timeline.jsx
```diff
@@ -0,0 +1,51 @@
...
+      {children}
+    </div>
+  )
+}
+Timeline.Element = TimelineElement
+
+Timeline.ElementBreakpoint = ({
+  children,
+  className = "",
+  // start, center
```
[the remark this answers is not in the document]
r-czajkowski: According to the Figma design system currently, we need `start` and `center`, so I think we can skip this for now. Thoughts?

solidity/dashboard/src/components/Timeline.jsx
r-czajkowski: Currently, we only support timeline on the `left` side. The `position` prop may be helpful in the future development of this component.

solidity/dashboard/src/css/timeline.less
r-czajkowski: > to write the line for each element only below it (instead of 50% on top and 50% on bottom)

This is only true when all elements have the same height. We need to draw a line with the height of the element so I think the only way is to draw 50% on bottom and 50% on top. I'll try to make the dashed line looks better.

liz-shinn: For the dashed line element, it looks like it's missing the rounded radius here, like this: 

Screen Shot 2021-07-22 at 2 27 47 PM

I'm not sure the exact radius is though since Figma does dashed lines like this:
Screen Shot 2021-07-22 at 2 29 02 PM
And doesn't give a radius setting for individual dash marks.

r-czajkowski: Fixed in 3ee96a6.

For one timeline element.
obraz

r-czajkowski: > For the dashed line element, it looks like it's missing the rounded radius here, like this:

Regarding rounded dots. Firefox correctly draws a dotted line - with rounded corners. Chrome has a problem with this and it draws a dashed line. What the dotted line looks like depends on the browser and I'm afraid there is nothing we can do with this.

Chrome:
obraz

Firefox:
obraz

cc @liz-shinn

liz-shinn: Ahhh ok, woof! Thx for explaining that quirk, alas cursed chrome

solidity/dashboard/src/components/Timeline.jsx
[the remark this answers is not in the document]
r-czajkowski: I think this comment is outdated now because the dots are always on the start position (according to the Figma views). We can add support for this position when needed.

solidity/dashboard/src/css/chip.less
```diff
@@ -4,6 +4,7 @@
     pending: @pending-color;
     error: @error-color;
     disabled: @color-grey-10;
+    violet-80: #633EB1; // @color-brand-violet-80
```
michalsmiarowski, line 7: The `violet-80` name doesn't really match the previous ones (emphasis, success, pending). Maybe we should change it for `info` for example? Or something like that?

solidity/dashboard/src/css/timeline.less
```diff
@@ -1,101 +1,90 @@
...
-    padding: 1.5rem 2rem;
-    border: 1px solid #AE96E1; // @color-brand-violet-40
-    border-left: @left-border-width solid #633EB1; // @color-brand-violet-80
-    border-radius: 0.5rem;
-    &:extend(.box-shadow);
-    background-color: @white;
-    margin-bottom: @margin-bottom-between-elements;
-    
+    display: flex;
+    // min-height: 80px;
```
michalsmiarowski, line 11: Let's remove this commented `min-height` if you are not planning to use it

michalsmiarowski (changes requested):
I love how you handled the dashed line (the `BreakpointLine` component) ❤️ Left some last small comments and we are good to go here! 🚀

solidity/dashboard/src/css/chip.less
r-czajkowski: I updated the name according to the [Keep Design System (Sigils component)](https://www.figma.com/file/ARAMFLMOPal0JewjvlZGhX/Token-dApp-Design-System?node-id=2612%3A259).

michalsmiarowski (approved):
LGTM :shipit:

PR #2544, opened 2021-07-20
Code formatting with eslint and prettier 

r-czajkowski:
This PR updates the `eslint` configuration and adds support for `prettier` in KEEP Token Dashboard. It also fixes a pre-commit hook and adds a GH workflow for code formatting.

To run formatting check execute: `npm run format`
To automatically fix errors execute: `npm run format:fix`

TODO:
- [x] add `.git-blame-ignore-revs` file ([ref](https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame))
- [x] run workflow only if changes to `solidity/dashboard` are detected.


.github/workflows/dasboard-format.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
+          dashboard:
+            - './solidity/dashboard/**'
+
+  dashboard-code-format:
+    needs: dashboard-detect-changes
+    if: ${{ needs.dashboard-detect-changes.outputs.dashboard == 'true' }}
```
michalinacienciala, line 30: Three comments regarding this section:
1. I would like to keep the implementation of different workflows quite similar, when possible (it makes it easier to do bulk edits when some changes are needed across different workflows), hence I would suggest changing name of the filter from `dashboard` to `path-filter`, as we call it in other workflows (although I see why you picked the current name and if you feel it's better, I'm ok with keeping it).
2. I don't remember the exact details now, but I believe that there was some issue with using `dorny/paths-filter` action for workflows started by `push` event (if I recall correctly, the problem with correct filtering could occur if feature branch was ahead of main certain amount of commits). That's why in other workflows we use the action only for workflows started by `pull-request` event. I think we need to do the same here.
3. I believe it's almost always a good idea to add `workflow_dispatch` trigger to the workflows.

```suggestion
on:
  push:
    branches:
      - main
    paths:
      - "solidity/dashboard/**"
  pull_request:
  workflow_dispatch:

defaults:
  run:
    working-directory: ./solidity/dashboard

jobs:
  dashboard-detect-changes:
    runs-on: ubuntu-latest
    outputs:
      dashboard: ${{ steps.filter.outputs.dashboard }}
    steps:
    - uses: actions/checkout@v2
      if: github.event_name == 'pull_request'

    - uses: dorny/paths-filter@v2
      if: github.event_name == 'pull_request'
      id: filter
      with:
        filters: |
          path-filter:
            - './solidity/dashboard/**'

  dashboard-code-format:
    needs: dashboard-detect-changes
    if: |
      github.event_name != 'pull_request'
        || needs.dashboard-detect-changes.outputs.path-filter == 'true'
```

.github/workflows/dasboard-format.yaml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: "12.x"
```
michalinacienciala, line 37: The `setup-node` action now [supports caching](https://github.com/actions/setup-node#caching-packages-dependencies), we can add following:

```suggestion
          node-version: "12.x"
          cache: 'npm'
```

solidity/dashboard/.git-blame-ignore-revs
```diff
@@ -0,0 +1,6 @@
+# This file contains commit hashes of bulk changes that can be omitted by the
```
nkuba, line 1: Can we put `.git-blame-ignore-revs` file in the repository root directory?

solidity/dashboard/.prettierrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  ...require("@keep-network/prettier-config-keep"),
+}
```
nkuba, line 3: We've got a `Dockerfile` that I believe can get formatted by `"prettier-plugin-sh"` plugin.

```suggestion
  plugins: ["prettier-plugin-sh"],
}
```

nkuba (changes requested):

solidity/dashboard/package.json
```diff
@@ -45,18 +45,18 @@
...
   },
   "devDependencies": {
-    "@keep-network/prettier-config-keep": "github:keep-network/prettier-config-keep",
+    "@keep-network/prettier-config-keep": "github:keep-network/prettier-config-keep#a1a333e",
     "@redux-saga/testing-utils": "^1.1.3",
     "@testing-library/react-hooks": "^5.1.2",
     "@types/jest": "^26.0.21",
     "eslint": "^6.8.0",
-    "eslint-config-keep": "git+https://github.com/keep-network/eslint-config-keep.git",
+    "eslint-config-keep": "github:keep-network/eslint-config-keep#0c27ade",
```
michalinacienciala, line 59: In some places (like [here](https://github.com/keep-network/tbtc-v2/blob/6f1e52375d1cdcf259ff84af035e5d538730c752/solidity/package.json#L12) we use `#0.3.0` ref. I wonder if there are clear advantages of using one approach over the other, or maybe some conventions of which approach should be used when?
(I'm not requesting a change, I'ts more about educating myself :) )

nkuba: [`0.3.0` tag](https://github.com/keep-network/eslint-config-keep/releases/tag/0.3.0) is older than [0c27ade commit](https://github.com/keep-network/eslint-config-keep/commit/0c27ade54e725f980e971c3d91ea88bab76b2330).

We need to make publishing and tagging of our tools more automatic.

solidity/dashboard/package.json
```diff
@@ -45,16 +45,19 @@
...
   "devDependencies": {
+    "@keep-network/prettier-config-keep": "github:keep-network/prettier-config-keep#a1a333e",
     "@redux-saga/testing-utils": "^1.1.3",
     "@testing-library/react-hooks": "^5.1.2",
     "@types/jest": "^26.0.21",
     "eslint": "^6.8.0",
-    "eslint-config-keep": "git+https://github.com/keep-network/eslint-config-keep.git",
-    "prettier": "^2.0.4",
+    "eslint-config-keep": "github:keep-network/eslint-config-keep#0c27ade",
+    "prettier": "^2.3.2",
```
michalinacienciala, line 60: We'll also need the `"prettier-plugin-sh": "^0.6.1",` dependency if `prettier-plugin-sh` will be added to `solidity/dashboard/.prettierrc.js` according to Kuba's suggestion.

michalinacienciala:
I think we should also add `.eslintignore` file on the `keep-core/solidity` level and exclude the `./dashboard/` there

```
# Following directories have their own formatting configuration:
./dashboard/
```

nkuba:
> I think we should also add `.eslintignore` file on the `keep-core/solidity` level and exclude the `./dashboard/` there
> 
> ```
> # Following directories have their own formatting configuration:
> ./dashboard/
> ```

We already exclude it here https://github.com/keep-network/keep-core/blob/main/solidity/package.json#L22

.github/workflows/dasboard-format.yaml
r-czajkowski: In this case, I think it makes sense to put `if` statement on the job level instead of for each `step`. 

```
jobs:
  dashboard-detect-changes:
    if: github.event_name == 'pull_request'
```

Thoughts?

r-czajkowski:
@nkuba, @michalinacienciala PR is ready for the next round!

.github/workflows/dasboard-format.yaml
michalinacienciala: Normally I would agree, but in this particular job we need to keep the `if`s on the step level, as we want the `dashboard-detect-changes` job (but not it's step) to execute for workflow_dispatch and pull events. If we had the `if` on the job level, the job wouldn't be skipped for this type of events, resulting in skipping of jobs that have it configured as a dependency in the `needs` field.

But now when I think about it, maybe we can completely get rid of the `dashboard-detect-changes` job... This job is needed in the workflow if we have a situation where we want to make a passing of some job (like `dashboard-code-format`) mandatory for the merge of PR and when we want that job to be executed only if the PR touches the files in some specific location. I'm not sure if the `dashboard-code-format` satisfies that first condition.

@nkuba, will we be adding the `dashboard-code-format` as mandatory for PR merge? I can see that in `keep-ecdsa` we don't have the formatting job set as required...

nkuba: We should set the formatting job as required for PR merge.

michalinacienciala: Ok, so then we need to keep the solution with `dashboard-detect-changes` job. @r-czajkowski please move the `if`s back to the step level.

michalinacienciala (changes requested):
Please move the `if`s in the `dashboard-detect-changes` job to the step level. Other than that looks ok to me.

solidity/dashboard/.git-blame-ignore-revs
r-czajkowski: please see e4c5ae3

michalinacienciala (approved):
Looks good to me!

nkuba (approved):

PR #2549, opened 2021-07-29
Coverage pool- display metrics

r-czajkowski:
This PR adds a common component- `MetricsTile` to the `How it Works` and `Deposit` page. This component displays following metrics:
* tvl,
* tvl in USD,
* reward rate- the rate of rewards that you will receive annually,
* lifetime rewards- rewards distributed from the rewards pool contract
since the start of the pool,
* lifetime covered- amount of KEEP used from the coverage pool to cover
a loss since the start of the pool.

<details>
  
  
obraz

obraz

obraz

</details>


solidity/dashboard/src/css/coverage-pools-page.less
```diff
@@ -1,59 +1,63 @@
...
+      grid-area: tvl;
 
-  .checklist-wrapper {
-    display: flex;
-    flex-wrap: wrap;
+      .tvl.tvl--fiat-currency {
+        color: @color-grey-40;
+        font-weight: 600;
+        margin-top: 0.75rem;
+        margin-left: 2.25rem;
```
michalsmiarowski, line 18: Can we add a little bit more left-margin here?

image

The dollar sine should start be placed exactly under the first digit of the amount.

solidity/dashboard/src/css/tooltip.less
```diff
@@ -127,3 +128,6 @@
     }
   }
 }
+
+.tooltip--resources {
+}
```
michalsmiarowski, line 133: Do we need this?

solidity/dashboard/src/components/coverage-pools/MetricsSection.jsx
```diff
@@ -0,0 +1,103 @@
...
+          {isTotalAllocatedRewardsFetching ? (
+            <Skeleton tag="h2" shining color="grey-10" />
+          ) : (
+            <TokenAmount
+              amount={totalAllocatedRewards}
+              withIcon
+              withSymbol={false}
+              withMetricSuffix
+            />
+          )}
```
michalsmiarowski, line 73: Just a thought to discuss - we could also write it as two `<OnlyIf>` components. Maybe it will be more readable? What do you think?

solidity/dashboard/src/components/coverage-pools/MetricsSection.jsx
```diff
@@ -0,0 +1,103 @@
...
+          {isLifetimeCoveredFetching ? (
+            <Skeleton tag="h2" shining color="grey-10" />
+          ) : (
+            <TokenAmount
+              amount={lifetimeCovered}
+              withIcon
+              withSymbol={false}
+              withMetricSuffix
+            />
+          )}
```
michalsmiarowski, line 95: Same as above if you agree the `<OnlyIf>`s components would be more readable

michalsmiarowski (changes requested):
Looking awesome! 

And also left some small notes to look at 😉

solidity/dashboard/src/components/coverage-pools/MetricsSection.jsx
r-czajkowski: I used `OnlyIf` as you suggested but not sure if it's ok due to the repeating condition eg:
```
<OnlyIf={condition}>
...
</OnlyIf>
<OnlyIf={!condition}>
...
</OnlyIf>
```
I think the `Skeleton` component can handle this. I mean something like this:
```
<Skeleton condition={condition}>
 <Component />
</Skeleton>
```

I think we can use `OnlyIf` for now, but we should improve this component in the future or use another approach for cases as above.

michalsmiarowski (approved):
Tested locally and it looks good! 🚢

PR #2574, opened 2021-08-18
Fix group selection event deduplication logic

lukasz-zimnoch:
We deduplicate `GroupSelectionStarted` events by keeping them into the `pendingGroupSelections` cache. Duplicated events may occur due to chain reorgs or because of the subscription monitoring mechanism which re-fetches past events in case something bad happens with the WebSocket connection pointing to the Ethereum node.

So far, we removed the current group selection from cache just after ticket submission terminated. We can't do that because a duplicated event can come right after the removal and trigger a new DKG breaking the ongoing one.

To fix the problem, we need to hold the cache entry for the maximum DKG duration time. We should also do it regardless of the protocol outcome on the host node as other nodes may still execute the protocol successfully.

Worth flagging that removing the cache entry after hitting the DKG timeout is enough and we don't need to monitor for a successful group registration. A successful group registration means we never have to trigger group selection
again for the group seed we hold in the cache so we don't have to bother about removing the cache entry just after the group is registered. In the contrary, when the current group selection fails, we need to remove the cache entry once the timeout is hit to allow the node to retry the group selection for that seed as soon as possible.

### DIFF
--- a/pkg/beacon/beacon.go
+++ b/pkg/beacon/beacon.go
@@ -11,7 +11,9 @@ import (
 	"github.com/keep-network/keep-common/pkg/persistence"
 	"github.com/keep-network/keep-core/pkg/beacon/relay"
 	relaychain "github.com/keep-network/keep-core/pkg/beacon/relay/chain"
+	dkgresult "github.com/keep-network/keep-core/pkg/beacon/relay/dkg/result"
 	"github.com/keep-network/keep-core/pkg/beacon/relay/event"
+	"github.com/keep-network/keep-core/pkg/beacon/relay/gjkr"
 	"github.com/keep-network/keep-core/pkg/beacon/relay/groupselection"
 	"github.com/keep-network/keep-core/pkg/beacon/relay/registry"
 	"github.com/keep-network/keep-core/pkg/chain"
@@ -157,8 +159,6 @@ func Initialize(
 				return
 			}
 
-			defer pendingGroupSelections.Remove(newEntry)
-
 			logger.Infof(
 				"group selection started with seed [0x%x] at block [%v]",
 				event.NewEntry,
@@ -175,8 +175,58 @@ func Initialize(
 				onGroupSelected,
 			)
 			if err != nil {
-				logger.Errorf("Tickets submission failed: [%v]", err)
+				logger.Errorf("tickets submission failed: [%v]", err)
 			}
+
+			// We deduplicate `GroupSelectionStarted` events by keeping them
+			// into the `pendingGroupSelections` cache. Duplicated events
+			// may occur due to chain reorgs or because of the subscription
+			// monitoring mechanism which re-fetches past events in case
+			// something bad happens with the websocket connection pointing
+			// to the Ethereum node. However, we can't remove current group
+			// selection from cache just after ticket submission terminates
+			// because a duplicated event coming right after the removal
+			// will trigger a new DKG which will break the current one.
+			// We need to hold the cache entry for the maximum DKG duration
+			// time. We should also do it regardless of the outcome on this
+			// node as other nodes may still perform DKG successfully.
+			// Removing the cache entry after hitting the DKG timeout is
+			// enough and we don't need to monitor for a successful group
+			// registration. A successful group registration means we never
+			// have to trigger group selection again for the group seed we hold
+			// in the cache so we don't have to bother about removing the cache
+			// entry just after the group is registered. In contrary, when
+			// current group selection fails, we need to remove the cache
+			// entry once the timeout is hit to allow the node to retry
+			// the group selection for that seed as soon as possible.
+
+			dkgDurationBlocks :=
+				chainConfig.TicketSubmissionTimeout +
+					gjkr.ProtocolBlocks() +
+					dkgresult.PrePublicationBlocks() +
+					(uint64(chainConfig.GroupSize) * chainConfig.ResultPublicationBlockStep)
+
+			dkgTimeout := event.BlockNumber + dkgDurationBlocks
+
+			if err := blockCounter.WaitForBlockHeight(dkgTimeout); err != nil {
+				// We assume the average block time is 15s.
+				delayDuration := time.Duration(dkgDurationBlocks*15) * time.Second
+
+				logger.Errorf(
+					"could not set the block counter to wait "+
+						"until the DKG timeout block [%v] to remove the "+
+						"current group selection from cache: [%v]; "+
+						"current group selection will be removed from cache "+
+						"after a fixed delay of [%v]",
+					dkgTimeout,
+					err,
+					delayDuration,
+				)
+
+				<-time.After(delayDuration)
+			}
+
+			pendingGroupSelections.Remove(newEntry)
 		}()
 	})
 
--- a/pkg/beacon/relay/gjkr/states.go
+++ b/pkg/beacon/relay/gjkr/states.go
@@ -36,6 +36,25 @@ const (
 	combinationStateActiveBlocks = 20
 )
 
+// ProtocolBlocks returns the total number of blocks it takes to execute
+// all the required work defined by the GJKR protocol.
+func ProtocolBlocks() uint64 {
+	return ephemeralKeyPairStateDelayBlocks +
+		ephemeralKeyPairStateActiveBlocks +
+		commitmentStateDelayBlocks +
+		commitmentStateActiveBlocks +
+		commitmentVerificationStateDelayBlocks +
+		commitmentVerificationStateActiveBlocks +
+		pointsShareStateDelayBlocks +
+		pointsShareStateActiveBlocks +
+		pointsValidationStateDelayBlocks +
+		pointsValidationStateActiveBlocks +
+		keyRevealStateDelayBlocks +
+		keyRevealStateActiveBlocks +
+		combinationStateDelayBlocks +
+		combinationStateActiveBlocks
+}
+
 // ephemeralKeyPairGenerationState is the state during which members broadcast
 // public ephemeral keys generated for other members of the group.
 // `EphemeralPublicKeyMessage`s are valid in this state.


pkg/beacon/beacon.go
```diff
@@ -175,8 +175,58 @@ func Initialize(
...
+						"until the DKG timeout block [%v] to remove the "+
+						"current group selection from cache: [%v]; "+
+						"current group selection will be removed from cache "+
+						"after a fixed delay of [%v]",
+					dkgTimeout,
+					err,
+					delayDuration,
+				)
+
+				<-time.After(delayDuration)
```
pdyraga, line 226: I'd rather emit an error from the deduplicator and not allow the client to start ticket submission than to assume the block time. I think it's less error-prone and if one does not submit their ticket nothing bad happens.

pkg/beacon/beacon.go
```diff
@@ -175,8 +175,58 @@ func Initialize(
...
+
+			dkgDurationBlocks :=
+				chainConfig.TicketSubmissionTimeout +
+					gjkr.ProtocolBlocks() +
+					dkgresult.PrePublicationBlocks() +
+					(uint64(chainConfig.GroupSize) * chainConfig.ResultPublicationBlockStep)
+
+			dkgTimeout := event.BlockNumber + dkgDurationBlocks
+
+			if err := blockCounter.WaitForBlockHeight(dkgTimeout); err != nil {
```
pdyraga, line 211: Can we extract this logic and one for the relay entry to a dedicated deduplicator, just like we have in [`keep-ecdsa`](https://github.com/keep-network/keep-ecdsa/blob/f57fe591a1f14ff95bd22241e5934f48f83c8d0a/pkg/client/event/deduplicator.go)? Having it in a separate struct allows to unit test all cases properly. Also, the time spent on the implementation is not time lost - there is pretty complex logic here calculating dkg duration for the sake of deduplication and I think it deserves to be extracted out of `beacon.go`.

lukasz-zimnoch: But there are already dedicated deduplicators (see `GroupSelectionTrack` and `RelayRequestTrack`). The logic here just determines when to call `Remove` on `GroupSelectionTrack` and belongs to the client code using the deduplicator.

pdyraga: I do not think it's the client code, I think it's deduplicator logic to decide when to consider the given event is no longer a duplicate. I'd also love to have this code covered in unit tests, `beacon.go` in future as well. Keeping the code in separate structures would help us with it.

In the deduplicator's notify we'd registered an event in the track struct and scheduled an update to remove it from track when it's no longer needed. Notify completed function in the deduplicator for group selection event is not needed IMO.

pkg/beacon/beacon.go
lukasz-zimnoch: This is not so simple. If we emit an error here we must remove the current group selection seed from `GroupSelectionTrack` deduplicator immediately. That means we can still obtain a duplicated `GroupSelectionStarted` event and the client will begin a new ticket submission process. That round will fail as the contract state is "we are after ticket submission, DKG is in progress" but at the end, `SelectedParticipants` will return a valid participant list (from previous legit ticket submission) and the node will start a new DKG on behalf of its virtual stakers. This new DKG will break the ongoing one executed by other nodes as we are observing on `keep-test`. This is why we must wait with cache removal for a certain period of time. This period should ideally be the moment when DKG timeout occurs as we are sure the recent DKG is completed. However, in case the block counter returns an error (rather unlikely) we must wait a period of time close enough to the DKG duration. It doesn't have to be an exact value and I think the presented way is fair enough. Unfortunately, there is no ideal solution here. But, even if real block time is bigger than 15s, and the cache is cleared a little bit too early it will be not so easy to break the ongoing DKG at this stage. We still need a duplicated `GroupSelectionStarted` event which is hard to occur after such a long time from the original event. Also, even if it occurs, the original DKG should be close to their end and a new duplicated DKG will have a hard work to break things as members from the original DKG will not accept messages from members of duplicated DKG.

pkg/beacon/beacon.go
lukasz-zimnoch: Ok, I misunderstood. I mixed up things and was sure that when you say "deduplicator" you have `*Track` structures in mind. All clear now, I will extract dedicated deduplicators as we have in `keep-ecdsa`.

pkg/beacon/beacon.go
pdyraga: I see. In this case, maybe it makes more sense to store last group selection start event block number instead of storing the group selection seed in the event track map? We would allow another group selection to start if the last group selections start block + DKG time is less than the current block height.

In this case, we do not have to delete anything, we are just storing one integer.

lukasz-zimnoch: Good idea! Done in: https://github.com/keep-network/keep-core/pull/2574/commits/ea353ddbeff0915b5e9a11c030515c461707ff35

pkg/beacon/beacon.go
lukasz-zimnoch: Ok partially done in: https://github.com/keep-network/keep-core/pull/2574/commits/ea353ddbeff0915b5e9a11c030515c461707ff35. I'll add it for relay entries and add some tests for the deduplicator in further commits.

pkg/beacon/beacon.go
lukasz-zimnoch: But, we should keep in mind that if one DKG terminates with success, the next one will not be processed until DKG execution time elapses from the start of the latest DKG.

pkg/beacon/beacon.go
lukasz-zimnoch: Added for relay entry as well: https://github.com/keep-network/keep-core/pull/2574/commits/fb94fed2acef5178dff02250a02e610347540e02

pkg/beacon/beacon.go
pdyraga: I like where it is heading but I think we could simplify and optimize the logic. Also, I think it makes sense to solve the problem of the relay request event deduplication in this PR so that we do not have to go back to it after a couple of days. After reading your TODO I realized it's enough to have two relay requests executed one after another to break the deduplication logic.

There are two problems I see:
- Relay request deduplication does not work as intended; It deduplicates events only for a very short period of time.
- DKG deduplication is not efficient; We execute on-chain checks for most of the events that are getting deduplicated instead of deduplicating most of them using in-memory track first, and reaching to the chain only when it is absolutely necessary.

What I'd suggest doing for relay request:
Let's get rid of `NotifyRelayEntryCompleted`. As you noticed in the TODO, all the logic is done asynchronously and we currently can't provide a clear "completed" signal. We might want to change it in the future but without significant refactoring, it is not possible. Let's pass to the deduplicator how long it takes to execute relay entry signing protocol *without* the part for the result publication. When notifying about signing started, let's pass both block number, group public key, and previous entry (or the whole event). In case the block number is < than the last event block number from the track plus entry signing protocol execution time, the previous entry is the same, and group public key is the same, we consider the event as a duplicate given that no new signing could happen in that time for the same entry and group. If the block number is not under the threshold, we need to check the chain if signing is in progress and what is the previous relay entry. So two on-chain calls, that could be in future reduced to one (first we need to clean up operator contract).

What I'd suggest doing for DKG:
Let's pass the time it takes to select a group and execute DKG *without* a time to submit the result to the deduplicator. If we see an event under the threshold of the last DKG start block plus that time, we consider it as a duplicate. If the start block is above the threshold, we let it go.

lukasz-zimnoch: Ok, you're right and we can simplify as you proposed. However, for relay entry, we can't determine the duration in blocks as we have in DKG because the relay entry protocol is time-based. We also don't need to check the relay entry target group as requests occur synchronously. Here is my proposed solution: https://github.com/keep-network/keep-core/pull/2574/commits/4f0ac71660f7153909d79da5c874cbde38bfcf58

pkg/beacon/relay/event/deduplicator.go
```diff
@@ -0,0 +1,132 @@
...
+// time. It is deduplicator's responsibility to decide whether the given
+// event is a duplicate and should be ignored or if it is not a duplicate and
+// should be handled.
+//
+// Those events are supported:
+// - group selection started
+// - relay entry requested
+type Deduplicator struct {
+	chain                           chain
+	minGroupSelectionDurationBlocks uint64
```
pdyraga, line 29: `minGroupCreationDurationBlock`? For the same reason as explained earlier.

pkg/beacon/relay/event/deduplicator.go
```diff
@@ -0,0 +1,132 @@
...
+				if err != nil {
+					return false, fmt.Errorf(
+						"could not get current request previous entry: [%v]",
+						err,
+					)
+				}
+
+				if newRequestPreviousEntry ==
+					hex.EncodeToString(currentRequestPreviousEntryOnChain[:]) {
+					return true, nil
```
pdyraga, line 110: I think we cannot escape taking into account the time it takes to calculate relay entry off-chain.

Scenarios we observed on mainnet with no deduplication:
- The same previous entry but block number higher by 1 or 2 (minor chain reorg)
- The same previous entry and block number significantly higher (another group retrying)
- Another previous entry and block number significantly higher (happy path)

Scenarios we will observe on mainnet with deduplication:
- The same previous entry but block number higher by 1 or 2 (minor chain reorg)
- The same previous entry and block number significantly higher (another group retrying)
- Another previous entry and block number significantly higher (happy path)
- The same previous entry, block number from the past (event pull)
- Another previous entry, block number from the past (event pull, multiple requests)

Still possible, but the probability is quite low:
- Two requests with the same block number but different previous entries (chain fork)
- Two requests with block number close to each other but different previous entries (chain fork)

With the current implementation, the first scenario described (the same previous entry but block number higher by 1 or 2 (minor chain reorg)) will return true although the event should be filtered out.

I assume that in all cases but the old event pull, the new event received will have the block number higher or equal to the last one. This includes reorgs - if Ethereum client switches to another fork, that fork is longer. Taking it into account, I'd propose the following algorithm:
- If the previous entry is different and the block is higher than the previous one - let it pass.
- If the previous entry is the same and the block is higher than the previous one at least for the number of blocks it takes to sign entry off-chain - let it pass.
- Otherwise, filter out as a duplicate.

pdyraga: FWIW, I wouldn't care about reorgs for DKG. The penalty is not that severe and it takes a long time to create a group off-chain.

lukasz-zimnoch: You're right about the missing pieces about minor reorgs. However, we can't determine the time needed to calculate relay entry off-chain because the relay entry protocol is time-based and doesn't use the state machine (I mentioned this in one of the above comments). Anyway, I think we should just additionally check the current request starting block on-chain. This should protect us against duplicated events coming with a slightly higher block number than the original one. I added this here: https://github.com/keep-network/keep-core/pull/2574/commits/951a2bd578b0d6b7b3fd84ae4deb11a66a841e0d. What do you think?

pkg/beacon/beacon.go
```diff
@@ -147,25 +166,26 @@ func Initialize(
 			)
 		}
 
-		newEntry := event.NewEntry.Text(16)
 		go func() {
-			if ok := pendingGroupSelections.Add(newEntry); !ok {
+			if ok := eventDeduplicator.NotifyGroupSelectionStarted(
+				event.BlockNumber,
+			); !ok {
 				logger.Errorf(
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2574/commits/46ff2a7fc759ef8d22e1e99a6d360ecd56ae9544

pkg/beacon/beacon.go
```diff
@@ -62,35 +63,53 @@ func Initialize(
...
+	// time needed for publication as we are interested about the minimum
+	// possible group duration.
+	minGroupSelectionDurationBlocks :=
+		chainConfig.TicketSubmissionTimeout +
+			gjkr.ProtocolBlocks() +
+			dkgresult.PrePublicationBlocks()
+
+	eventDeduplicator := event.NewDeduplicator(
+		relayChain,
+		minGroupSelectionDurationBlocks,
```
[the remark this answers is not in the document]
lukasz-zimnoch: Right, renamed: https://github.com/keep-network/keep-core/pull/2574/commits/7423b139421edc721f4cf33b75d62509fa98dc1a

pkg/beacon/relay/event/deduplicator.go
lukasz-zimnoch: https://github.com/keep-network/keep-core/pull/2574/commits/7423b139421edc721f4cf33b75d62509fa98dc1a

pkg/beacon/beacon.go
```diff
@@ -147,25 +166,26 @@ func Initialize(
...
 
-		newEntry := event.NewEntry.Text(16)
 		go func() {
-			if ok := pendingGroupSelections.Add(newEntry); !ok {
+			if ok := eventDeduplicator.NotifyGroupSelectionStarted(
+				event.BlockNumber,
+			); !ok {
 				logger.Errorf(
-					"group selection event with seed [0x%x] has been registered already",
+					"group selection event with seed [0x%x] and "+
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, but I'd leave it just in case. It can help during debugging.

PR #2500, opened 2021-06-15
Coverage pool in KEEP Token Dashboard

r-czajkowski:
Depends on: #2522 

This is an integration PR for the coverage pool in KEEP Token Dashboard.

<details>
  <summary>Prototype</summary>

obraz

obraz

</details>

PR #2581, opened 2021-08-30
Update cov-pools dependency

michalinacienciala:
Coverage-pools contracts intended to be used for development of Token
Dashboard will be published to the NPM registry using under name with
`-dev` suffix. We need to update the dependencies configuration.
Note: for deployment of Dashboard on Testnet we're not using the
dependencies configured in the `package.lock`, instead we specify the
exact version we want to use when running the deployment workflow.

Ref:
https://github.com/keep-network/keep-core/pull/2500
https://github.com/keep-network/coverage-pools/pull/186
https://github.com/keep-network/keep-core/pull/2582


michalinacienciala:
I've run the NPM workflow for coverage-pools, we now have the [0.0.1-dev.1](https://www.npmjs.com/package/@keep-network/coverage-pools/v/0.0.1-dev.1) package in the NPM registry.

r-czajkowski (approved):

PR #2582, opened 2021-08-30
Coverage pools- fix UI bugs and subscriptions to AssetPool events

r-czajkowski:
~Depends on: https://github.com/keep-network/coverage-pools/pull/187~

This PR:
* updates `Deposit` and `Withdraw` form according to the Figma views(https://github.com/keep-network/keep-core/commit/adaffeec8a74f0d828a918c8ec311232d730ac48, https://github.com/keep-network/keep-core/commit/a471d52c499ff6360a442044a8f7fc12a82f1f03, 
ba8358f),
* fixes subscribing to the `AssetPool` contract events  and updating data based on them (https://github.com/keep-network/keep-core/commit/73fa742df2cac572c606a788fb18c5f0bc342ade),
* creates a common component, that displays info about cooldown period, used in the cov pools confirmation modals (https://github.com/keep-network/keep-core/commit/dadd81bdddbc9098c071af7f48a3ee0929fdf817),
* fixes other minor UI bugs (https://github.com/keep-network/keep-core/commit/6d690449937febec7853175af2a7ed53123181f4, https://github.com/keep-network/keep-core/commit/fc9487c142b0284524f9e69977a408cf520a1245, dd2de1a, 5354a22, e5bd767).


solidity/dashboard/src/sagas/coverage-pool.js
```diff
@@ -153,7 +152,7 @@ export function* watchFetchCovPoolData() {
   )
 }
 
-export function* subscribeToCovTokenTransferEvent() {
+export function* subscribeToAssetPoolDepositedEvent() {
   const requestChan = yield actionChannel(
     COVERAGE_POOL_COV_TOKEN_TRANSFER_EVENT_EMITTED
```
michalsmiarowski, line 157: Can we also change this const name to `COVERAGE_POOL_ASSET_POOL_DEPOSITED_EVENT_EMITTED`?

solidity/dashboard/src/sagas/coverage-pool.js
```diff
@@ -352,28 +325,71 @@ export function* subscribeToWithdrawalCompletedEvent() {
...
+      covTokenUpdated({
+        pendingWithdrawal: "0",
+        withdrawalInitiatedTimestamp: "0",
+        shareOfPool,
+        covBalance: updatedCovBalance,
+        covTotalSupply: updatedCovTotalSupply,
+        estimatedRewards,
+        estimatedKeepBalance,
+      })
+    )
```
michalsmiarowski, line 392: We should also update `tvl` here because without that it will take the old value for calculating `covKeep -> KEEP` `and thus calculating it wrong.

michalsmiarowski (changes requested):
Left some comments to look at but overall looks good 👍

solidity/dashboard/src/sagas/coverage-pool.js
```diff
@@ -352,28 +325,71 @@ export function* subscribeToWithdrawalCompletedEvent() {
...
-      returnValues: { underwriter, amount },
+      returnValues: { underwriter, amount, covAmount },
     } = event
     const address = yield select(selectors.getUserAddress)
 
-    if (isSameEthAddress(address, underwriter)) {
+    const { covTotalSupply, covBalance } = yield select(
+      selectors.getCoveragePool
+    )
+    const updatedCovTotalSupply = sub(covTotalSupply, amount)
```
michalsmiarowski, line 335: Should we subtract `covAmount` here from the `covTotalSupply`? I think `amount` is in KEEPs

solidity/dashboard/src/sagas/coverage-pool.js
michalsmiarowski: I think the solution for that would be to subtract `amount` from `tvl`

solidity/dashboard/src/components/FormCheckbox.jsx
```diff
@@ -7,10 +7,10 @@ const FormCheckbox = ({ label, withError = false, ...props }) => {
...
   return (
     <>
-      <label>
-        <input checked={field.value} {...field} {...props} />
-        &nbsp;{label}
-      </label>
+      <div className="form-checkbox">
+        <input id={props.name} checked={field.value} {...field} {...props} />
+        <label htmlFor={props.name}>{label}</label>
+      </div>
```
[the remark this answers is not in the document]
r-czajkowski: I agree the `text` cursor seems off. I believe it should be `cursor: default` instead of `pointer` according to [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox).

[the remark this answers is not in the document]
r-czajkowski: Please see 80e9a10

solidity/dashboard/src/sagas/coverage-pool.js
r-czajkowski: Please see 9129337

solidity/dashboard/src/sagas/coverage-pool.js
r-czajkowski: Please see da1d84e

solidity/dashboard/src/sagas/coverage-pool.js
r-czajkowski: Please see 16271a4

michalsmiarowski (approved):
LGTM

PR #2585, opened 2021-08-31
RFC 19: Random Beacon v2

pdyraga:
Proposal to simplify the pricing, rewards, and make timeouts more forgiving. Lessons learned on v1 of the random beacon.

### DIFF
--- a/docs/rfc/rfc-19-random-beacon-v2.adoc
+++ b/docs/rfc/rfc-19-random-beacon-v2.adoc
@@ -31,6 +31,7 @@ ceiling value impossible to maintain up-to-date. Given that the gas price
 ceiling had to be higher than the current gas price and that the fee for future
 group creation was calculated based on the gas price ceiling, new groups were 
 getting created too often. 
+
 With too many groups being created, operators were spending too much ETH on
 ticket submission and too low rewards were getting accumulated per group making
 the reward withdrawal operation not profitable enough or not profitable at all
@@ -55,6 +56,34 @@ operators noticed the problem.
 
 == Proposal
 
+=== Group creation
+
+New groups should be created with a fixed, governable frequency of relay
+requests. Instead of a ticket-based approach for group selection, we should
+use sortition pools. Group creation start transaction should be embedded into
+relay request transaction and should lock a sortition pool. From this moment,
+no operator can enter or leave the pool. All off-chain clients should perform
+group selection optimistically, using a view sortition function call and perform
+DKG. Once the group submits the result to the chain, a challenge period starts.
+
+During the challenge period, anyone can notify that the submitted DKG result
+contains group members not selected by the pool. This leads to slashing all
+members who signed the result and the notifier receives 5% from the slashed amount.
+The length of the challenge period and slashing amount are governable parameters.
+
+Once the challenge period passes, anyone can unlock the sortition pool and mark
+the DKG result as accepted. In this transaction, the DKG result submitter receives
+a reward, as described in <<fees-and-rewards, Fees and Rewards>> section.
+
+In case the DKG result was not submitted before a timeout, anyone can notify DKG
+timed out and receive a reward, as described in <<fees-and-rewards, Fees and Rewards>> 
+section.
+
+The sortition pool should weigh operators by stake and allow to select the same 
+operator multiple times. DKG protocol should execute in the same way as for v1
+and inactive/disqualified members should be marked and kicked out of the
+sortition pool for at least 2 weeks losing their rewards for that time.
+
 [[fees-and-rewards]]
 === Fees and rewards
 
@@ -63,43 +92,66 @@ deposit. Even though they are not reimbursed for on-chain actions and they
 receive KEEP rewards and TBTC fees, this scheme proved to be successful and no
 operators complained about it.
 
-For v2 of the random beacon, all rewards and reimbursements should be done in T.
+For v2 of the random beacon, all rewards should be done in T.
 This scheme is compatible with tBTC v2 design where all rewards and
-reimbursements are also done only in T.
+are also done only in T.
 
 Relay requester should provide a fee in T. The value of the fee is a governable
-parameter. The fee is split into two parts. One part goes to a group creation
-pool and another part is used as a reimbursement for the relay entry submitter.
-What percent of the fee goes to the group creation pool is a governable
-parameter.
+parameter. The entire fee is deposited in the maintenance pool that is used to
+reimburse for different actions related to DKG.
 
-Given that we want to incentivize the submission of a relay entry as soon as possible
-and that the price relation between gas price and T reimbursement for relay
-entry submission is unknown at the moment the fee is set by the governance, we
-should calculate a delay factor and delay penalty as in v1 of the random beacon.
-The delay penalty should be subtracted from reimbursement and sent to the
-sortition-pool-wide reward pool. The remaining part is sent to the relay entry
-submitter. Over the course of 64x6 blocks, reimbursement goes to zero. On-chain we
-are not going to enforce any order of submitters but off-chain clients should
-agree to the order where the member with index 1 has 6 blocks to submit, after 6
-blocks member with index 2 tries to submit, after 12 blocks member with index 3
-tries to submit and so forth.
+The transaction submitting relay entry is not reimbursable and implementation
+should ensure the gas cost of this transaction is as low as possible, below
+200k gas. The first group member eligible to submit the result is 
+`previous_entry MOD group_size`, then also `(previous_entry MOD group_size) + 1`.
+At the end, all group members should be eligible to submit the result. If the
+given member did not submit the result in their round, they should be removed
+from the sortition pool for at least 2 weeks and do not earn rewards for
+that time. The operation removing members from the sortition pool should be
+as cheap as possible so that the member submitting relay entry does not have to
+pay any additional costs for removing inactive members from the pool. Smart
+contract patterns such as gas station could be helpful in achieving this goal.
 
-Anyone should be able to submit a transaction creating a new group and this
-action should be detached from relay entry submission. The submitter of this
-transaction receives a part of the group creation pool and the remaining part
-should be reserved for the DKG result submitter as reimbursement for a future
-transaction. What percentage of the pool goes where is a governable parameter
-and should be determined based on the gas cost of the two transactions.
+For example, if `previous_entry MOD group_size = 60` and the entry was submitted 
+by the member with index `3`, members with indexes `{60, 61, 62, 63, 64, 1, 2}`
+should be removed from the sortition pool in the same transaction in which
+member with index `3` submitted relay entry.
+
+The default time for a single member to submit a relay entry should be increased 
+from 6 to 10 blocks and should become a governable parameter.
+
+There is a fixed, governable reward for submitting a DKG result, paid from the
+maintenance pool. The reward is paid to the DKG result submitter in the transaction
+approving the DKG result. The logic triggering new group selection will be embedded 
+in relay entry request transaction and should be as cheap as possible - it only
+locks the sortition pool and emits a group creation start event.
+
+In case the DKG result has not been submitted on time, anyone can unlock the pool
+and receive a fixed, governable reward for reporting DKG timeout. The reward is
+paid from the maintenance pool.
+
+Just like in the case of relay entry submission, the order in which operators
+are supposed to submit DKG result is enforced on-chain and operators who missed
+their round are removed from the sortition pool for at least 2 weeks and do
+not earn rewards for that time. The first member eligible to submit the DKG result
+is a member with index `1`, then also `2`. For example, if member `9` submitted
+DKG result, members with indexes `{1, 2, 3, 4, 5, 6, 7, 8}` should be removed
+from the sortition pool in the same transaction in which member `9` submitted
+DKG result.
+
+The default time for each member to submit DKG result should be set to 10 blocks
+and should become a governable parameter.
 
 To make this scheme work, both the relay entry and the DKG result submission
 transactions need to have predictable gas costs. Although it is guaranteed for
 the latter, the former, in v1 of the beacon, depends on the gas cost of
-executing a callback and needs to be optimized.
+executing a callback and needs to be optimized - more on that in the
+ <<callbacks,Callbacks>> section.
 
 T rewards will be distributed continuously to all operators in the beacon
 sortition pool, just like in the case of tBTC v2.
 
+[[callbacks]]
 === Callbacks
 
 In v1 of the random beacon, a callback is executed in the same transaction in
@@ -110,46 +162,37 @@ balance. This approach is not compatible with the idea that all rewards will be
 solely in T given that it is impossible to establish the cost of executing a
 callback in T.
 
-Instead of executing any callback in the same transaction in which relay entry
-is submitted, we can assume that only the simplest callbacks are allowed, up to
-some gas limit set by the governance. Applications wanting to use a relay entry
-should submit another transaction using the relay entry value previously set by
-the random beacon. They should also employ an additional security check making
-sure the entry submitted is only valid for a certain number of blocks to avoid
-the situation when relay entry beacon transaction and application-specific
-transaction are executed far from each other.
-
-For example, in tBTC v2, the random beacon transaction which submits the relay
-entry would execute a simple callback that sets the value of the entry on the
-wallet factory. Another transaction executes on the tBTC v2 side to the wallet
-factory that would use the entry for sortition.
-
 The fact the full - even the most complex - callback is executed in the same
 transaction in which the relay entry is submitted gives an impression of better
 security. This impression is false though, given that the entry to be submitted
 is visible in the mempool and smart attackers can have their transactions mined
-faster to - for example - put the sortition pool in the desired state for the
-new entry. This issue needs to be solved on the sortition pool side with
-initiation time for new operators in the pool.
+faster to, for example, put the sortition pool using the entry for group
+selection in the desired state. This issue needs to be solved on the sortition
+pool side with initiation time for new operators in the pool and/or state lock.
 
-=== Group creation
-
-As mentioned in the section about <<fees-and-rewards,fees and rewards>>, anyone 
-should be able to submit a transaction to create a new group and we should switch 
-from a ticket-based approach to sortition pools. The sortition pool should weigh 
-operators by stake and allow to select the same operator multiple times. DKG 
-protocol should execute in the same way as for v1 and inactive/disqualified 
-members should be marked and kicked out of the sortition pool losing their
-rewards.
+To ensure predictable cost of relay entry submission transaction, we should
+eliminate callbacks altogether. It is up to the application to act when a new
+relay entry has been submitted and execute the necessary logic based on the relay
+entry value. Random beacon will store the block in which the entry was submitted
+next to the entry so that the application may decide if the entry timed out in
+case the application transaction reading the entry was delayed for whatever reason.
+In this model, the application requests for a relay entry in one transaction,
+and reads the entry performing the necessary logic in another transaction.
+We do not care about free-riders. They can read the last entry value but can not
+have the entry produced in the frequency they need without paying a fee in T.
 
 === Timeouts
 
 We should extend the timeout for submitting relay entry to give operators more
-time to react. After 64x6 blocks, relay entry submission reward goes to 0. From
-this point, all operators in the group should start bleeding and losing their
-stake. The bleeding should increase linearly from 0 to the minimum stake per
-operator over the course of 48 hours. Both the initial and hard timeout should
-be governable parameters.
+time to react. After 64x10 blocks from the relay request, if no entry was
+provided, all operators in the group should start bleeding and losing their
+stake. The bleeding should increase linearly from 0 to the governable slashing 
+amount per operator over the course of 48 hours. The time for a single group
+member to become eligible to submit result and the hard relay entry timeout are
+governable parameters. This gives a chance to start with more forgiving penalties
+and increase them over time. In general, the slashing penalty should
+be proportional to rewards and the frequency of relay requests and associated
+risk.
 
 === Reduced complexity of contract relationships
 
@@ -162,6 +205,19 @@ create a new group on the most recent operator contract. All these
 functionalities, needed in v1, are simplified in this RFC so the service
 contract may no longer be needed, and removing it from the execution path could
 save some gas.
+
+== List of governable parameters
+
+* Relay request fee in T
+* Reward for submitting DKG result
+* Reward for unlocking the sortition pool if DKG timed out
+* Slashing amount for not submitting relay entry
+* Slashing amount for submitting malicious DKG result
+* The number of blocks for a member to become eligible to submit relay entry
+* The number of blocks for a member to become eligible to submit DKG result
+* Hard timeout for a relay entry
+* The frequency of a new group creation
+* Group lifetime
 
 [bibliography]
 == Related Links
--- a/docs/rfc/rfc-19-random-beacon-v2.adoc
+++ b/docs/rfc/rfc-19-random-beacon-v2.adoc
@@ -26,8 +26,11 @@ tBTC v2.
 === ETH-based pricing
 
 Having reimbursements for various on-chain actions in ETH was a nice feature but
-also complex in practice. Highly volatile gas price led to creating new groups
-too often and gas price ceiling value was impossible to maintain up-to-date.
+also complex in practice. Highly volatile gas price on mainnet made the gas price
+ceiling value impossible to maintain up-to-date. Given that the gas price
+ceiling had to be higher than the current gas price and that the fee for future
+group creation was calculated based on the gas price ceiling, new groups were 
+getting created too often. 
 With too many groups being created, operators were spending too much ETH on
 ticket submission and too low rewards were getting accumulated per group making
 the reward withdrawal operation not profitable enough or not profitable at all
--- a/docs/rfc/rfc-19-random-beacon-v2.adoc
+++ b/docs/rfc/rfc-19-random-beacon-v2.adoc
@@ -55,6 +55,7 @@ operators noticed the problem.
 
 == Proposal
 
+[[fees-and-rewards]]
 === Fees and rewards
 
 In tBTC v1, operators have to execute a lot of on-chain transactions per
@@ -133,12 +134,13 @@ initiation time for new operators in the pool.
 
 === Group creation
 
-As mentioned in the section about fees and rewards, anyone should be able to
-submit a transaction to create a new group and we should switch from a
-ticket-based approach to sortition pools. The sortition pool should weigh operators
-by stake and allow to select the same operator multiple times. DKG protocol
-should execute in the same way as for v1 and inactive/disqualified members
-should be marked and kicked out of the sortition pool losing their rewards.
+As mentioned in the section about <<fees-and-rewards,fees and rewards>>, anyone 
+should be able to submit a transaction to create a new group and we should switch 
+from a ticket-based approach to sortition pools. The sortition pool should weigh 
+operators by stake and allow to select the same operator multiple times. DKG 
+protocol should execute in the same way as for v1 and inactive/disqualified 
+members should be marked and kicked out of the sortition pool losing their
+rewards.
 
 === Timeouts
 
--- a/docs/rfc/rfc-19-random-beacon-v2.adoc
+++ b/docs/rfc/rfc-19-random-beacon-v2.adoc
@@ -17,9 +17,11 @@ For the time it was working, the beacon produced more than 1700 relay entries
 and 70 groups.
 
 Both on-chain and off-chain components proved to be successful but - just like
-with any other product - some pain points were identified by the users. This RFC
-aims at describing the pain points and proposing changes that should make the
-beacon easier to operate and more consistent with tBTC v2.
+with any other product - some pain points were identified by the users.
+The most problematic aspects were the construction of pricing scheme and too
+aggressive timeouts. This RFC aims at describing these problems and proposing
+changes that should make the beacon easier to operate and more consistent with
+tBTC v2.
 
 === ETH-based pricing
 
--- a/docs/rfc/rfc-19-random-beacon-v2.adoc
+++ b/docs/rfc/rfc-19-random-beacon-v2.adoc
@@ -78,6 +78,11 @@ Once the challenge period passes, anyone can unlock the sortition pool and mark
 the DKG result as accepted. In this transaction, the DKG result submitter receives
 a reward, as described in <<fees-and-rewards, Fees and Rewards>> section.
 
+There is a timeout before which DKG result should be submitted. The timeout equals
+the group size multiplied by the number of blocks for a member to become eligible
+to submit DKG result. The timer starts at the moment when the first member becomes
+eligible.
+
 In case the DKG result was not submitted before a timeout, anyone can notify DKG
 timed out and receive a reward, as described in <<fees-and-rewards, Fees and Rewards>> 
 section. DKG timeout includes the situation when no new relay entry was produced
--- a/docs/rfc/rfc-19-random-beacon-v2.adoc
+++ b/docs/rfc/rfc-19-random-beacon-v2.adoc
@@ -160,3 +160,8 @@ create a new group on the most recent operator contract. All these
 functionalities, needed in v1, are simplified in this RFC so the service
 contract may no longer be needed, and removing it from the execution path could
 save some gas.
+
+[bibliography]
+== Related Links
+
+- link:rfc-16-pricing.adoc[RFC 16: Pricing]
--- a/docs/rfc/rfc-19-random-beacon-v2.adoc
+++ b/docs/rfc/rfc-19-random-beacon-v2.adoc
@@ -62,9 +62,11 @@ New groups should be created with a fixed, governable frequency of relay
 requests. Instead of a ticket-based approach for group selection, we should
 use sortition pools. Group creation start transaction should be embedded into
 relay request transaction and should lock a sortition pool. From this moment,
-no operator can enter or leave the pool. All off-chain clients should perform
-group selection optimistically, using a view sortition function call and perform
-DKG. Once the group submits the result to the chain, a challenge period starts.
+no operator can enter or leave the pool. Once a new relay entry appears on the
+chain, all off-chain clients should perform group selection optimistically, 
+using the new entry value and a view sortition pool function call.
+After determining group members, clients should perform DKG as usual. 
+Once the group submits the result to the chain, a challenge period starts.
 
 During the challenge period, anyone can notify that the submitted DKG result
 contains group members not selected by the pool. This leads to slashing all
@@ -77,7 +79,8 @@ a reward, as described in <<fees-and-rewards, Fees and Rewards>> section.
 
 In case the DKG result was not submitted before a timeout, anyone can notify DKG
 timed out and receive a reward, as described in <<fees-and-rewards, Fees and Rewards>> 
-section.
+section. DKG timeout includes the situation when no new relay entry was produced
+and sortition could not be performed.
 
 The sortition pool should weigh operators by stake and allow to select the same 
 operator multiple times. DKG protocol should execute in the same way as for v1
--- a/docs/rfc/rfc-19-random-beacon-v2.adoc
+++ b/docs/rfc/rfc-19-random-beacon-v2.adoc
@@ -87,6 +87,11 @@ operator multiple times. DKG protocol should execute in the same way as for v1
 and inactive/disqualified members should be marked and kicked out of the
 sortition pool for at least 2 weeks losing their rewards for that time.
 
+Each group created in the system remains active for the same period of time
+which should be a governable parameter. Group that expired is no longer selected
+for any new work and group expiration will be handled in the relay request
+transaction.
+
 [[fees-and-rewards]]
 === Fees and rewards
 
--- a/docs/rfc/rfc-19-random-beacon-v2.adoc
+++ b/docs/rfc/rfc-19-random-beacon-v2.adoc
@@ -1,0 +1,157 @@
+:toc: macro
+
+= RFC 19: Random Beacon v2
+
+:icons: font
+:numbered:
+toc::[]
+
+== Background
+
+The last release of the random beacon got its genesis Sep 16th 2020 and worked
+until November 11th when it was stopped with a panic button as a result of an
+unannounced https://github.com/keep-network/keep-core/blob/main/docs/status-reports/2020-11-11-retro-geth-hardfork.adoc[Ethereum
+hard fork].
+
+For the time it was working, the beacon produced more than 1700 relay entries
+and 70 groups.
+
+Both on-chain and off-chain components proved to be successful but - just like
+with any other product - some pain points were identified by the users. This RFC
+aims at describing the pain points and proposing changes that should make the
+beacon easier to operate and more consistent with tBTC v2.
+
+=== ETH-based pricing
+
+Having reimbursements for various on-chain actions in ETH was a nice feature but
+also complex in practice. Highly volatile gas price led to creating new groups
+too often and gas price ceiling value was impossible to maintain up-to-date.
+With too many groups being created, operators were spending too much ETH on
+ticket submission and too low rewards were getting accumulated per group making
+the reward withdrawal operation not profitable enough or not profitable at all
+in the case of small stakers.
+
+Extra submitter rewards added on top of reimbursements were hard to understand
+for staking providers and calculating the provider’s fee was complicated given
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+unannounced https://github.com/keep-network/keep-core/blob/main/docs/status-reports/2020-11-11-retro-geth-hardfork.adoc[Ethereum
+hard fork].
+
+For the time it was working, the beacon produced more than 1700 relay entries
+and 70 groups.
+
+Both on-chain and off-chain components proved to be successful but - just like
+with any other product - some pain points were identified by the users. This RFC
+aims at describing the pain points and proposing changes that should make the
+beacon easier to operate and more consistent with tBTC v2.
```
beaushinkle, line 22: Can you include a high level summary of the pain points in the background? like maybe bullet points or something

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+
+=== ETH-based pricing
+
+Having reimbursements for various on-chain actions in ETH was a nice feature but
+also complex in practice. Highly volatile gas price led to creating new groups
+too often and gas price ceiling value was impossible to maintain up-to-date.
```
beaushinkle, line 28: why did the gas price volatility lead to creating new groups?

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+With too many groups being created, operators were spending too much ETH on
+ticket submission and too low rewards were getting accumulated per group making
```
beaushinkle, line 30: can this hyperlnk to what ticket submission is?

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+the reward withdrawal operation not profitable enough or not profitable at all
+in the case of small stakers.
+
+Extra submitter rewards added on top of reimbursements were hard to understand
+for staking providers and calculating the provider’s fee was complicated given
+the non-reimbursable cost of ticket submission. Surprisingly, it is much easier
+for staking providers to calculate their fee and costs in tBTC v1, having a
+clean rule that no operations are reimbursed and the only reward their receive
```
beaushinkle, line 38:
```suggestion
clean rule that no operations are reimbursed and the only reward they receive
```

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+This scheme is compatible with tBTC v2 design where all rewards and
+reimbursements are also done only in T.
+
+Relay requester should provide a fee in T. The value of the fee is a governable
+parameter. The fee is split into two parts. One part goes to a group creation
+pool and another part is used as a reimbursement for the relay entry submitter.
+What percent of the fee goes to the group creation pool is a governable
+parameter.
+
+Given that we want to incentivize to submit a relay entry as soon as possible
```
beaushinkle, line 70:
```suggestion
Given that we want to incentivize the submission of a relay entry as soon as possible
```

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+and that the price relation between gas price and T reimbursement for relay
+entry submission is unknown at the moment the fee is set by the governance, we
+should calculate a delay factor and delay penalty as in v1 of the random beacon.
+The delay penalty should be subtracted from reimbursement and sent to the
+sortition-pool-wide reward pool. The remaining part is sent to the relay entry
+submitter. Over the course 64x6 blocks, reimbursement goes to zero. On-chain we
```
beaushinkle, line 76:
```suggestion
submitter. Over the course of 64x6 blocks, reimbursement goes to zero. On-chain we
```

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+tries to submit and so forth.
+
+Anyone should be able to submit a transaction creating a new group and this
+action should be detached from relay entry submission. The submitter of this
+transaction receives a part of the group creation pool and the remaining part
+should be reserved for DKG result submitter as reimbursement for future
+transaction. What percentage of the pool goes where is a governable parameter
+and should be determined based on the gas cost of the two transactions.
+
+To make this scheme works both relay entry and DKG result submission
```
beaushinkle, line 89:
```suggestion
To make this scheme work, both the relay entry and the DKG result submission
```

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+for staking providers and calculating the provider’s fee was complicated given
+the non-reimbursable cost of ticket submission. Surprisingly, it is much easier
+for staking providers to calculate their fee and costs in tBTC v1, having a
+clean rule that no operations are reimbursed and the only reward their receive
+is KEEP plus TBTC fees.
+
+=== Aggressive timeouts
+
+Relay entry timeouts were pretty aggressive. Set to 64x6 = 384 blocks, they were
+giving about 1.5 hour to submit a relay entry before the minimum stake from each
```
beaushinkle, line 44:
```suggestion
giving about 1.5 hours to submit a relay entry before the minimum stake from each
```

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+submitter. Over the course 64x6 blocks, reimbursement goes to zero. On-chain we
+are not going to enforce any order of submitters but off-chain clients should
+agree to the order where the member with index 1 has 6 blocks to submit, after 6
+blocks member with index 2 tries to submit, after 12 blocks member with index 3
+tries to submit and so forth.
+
+Anyone should be able to submit a transaction creating a new group and this
+action should be detached from relay entry submission. The submitter of this
+transaction receives a part of the group creation pool and the remaining part
+should be reserved for DKG result submitter as reimbursement for future
```
beaushinkle, line 85:
```suggestion
should be reserved for the DKG result submitter as reimbursement for a future
```

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+transaction. What percentage of the pool goes where is a governable parameter
+and should be determined based on the gas cost of the two transactions.
+
+To make this scheme works both relay entry and DKG result submission
+transactions need to have predictable gas cost. Although it is guaranteed for
```
beaushinkle, line 90:
```suggestion
transactions need to have predictable gas costs. Although it is guaranteed for
```

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+solely in T given that it is impossible to establish the cost of executing a
+callback in T.
+
+Instead of executing any callback in the same transaction in which relay entry
+is submitted, we can assume that only the simplest callbacks are allowed, up to
+some gas limit set by the governance. Applications wanting to use a relay entry
+should submit another transaction using the relay entry value previously set by
+the random beacon. They should also employ an additional security check making
+sure the entry submitted is only valid for a certain number of blocks to avoid
+situation when relay entry beacon transaction and application-specific
```
beaushinkle, line 113:
```suggestion
the situation when relay entry beacon transactions and application-specific
```

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+transaction are executed far from each other.
```
beaushinkle, line 114:
```suggestion
transactions are executed far from each other.
```

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+
+On the example of tBTC v2, random beacon transaction submitting relay entry
+would only execute a simple callback setting the value of the entry on wallet
+factory, and another transaction executed on tBTC v2 side to the wallet factory,
+would use the entry for sortition.
```
beaushinkle, line 119: I think this can be reworded for clarity - maybe something like:

> For example, in tBTC v2, the random beacon transaction which submits the relay entry would execute a simple callback that sets the value of the entry on the wallet factory. Another transaction executes on the tBTC v2 side to the wallet factory that would use the entry for sortition.

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+transaction in which the relay entry is submitted gives an impression of better
+security. This impression is false though, given that the entry to be submitted
+is visible in the mempool and smart attackers can have their transactions mined
+faster to - for example - put the sortition pool in the desired state for the
+new entry. This issue needs to be solved on the sortition pool side with
+initiation time for new operators in the pool.
+
+=== Group creation
+
+As mentioned in the section about fees and rewards, anyone should be able to
```
beaushinkle, line 131: Can you hyperlink the fees and rewards section?

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+submit a transaction to create a new group and we should switch from a
+ticket-based approach to sortition pools. Sortition pool should weigh operators
```
beaushinkle, line 133:
```suggestion
ticket-based approach to sortition pools. The sortition pool should weigh operators
```

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+time to react. After 64x6 blocks, relay entry submission reward goes to 0. From
+this point, all operators in the group should start bleeding and losing their
+stake. The bleeding should increase linearly from 0 to the minimum stake per
+operator over the course of 48 hours. Both the initial and hard timeout should
+be governable parameters.
+
+=== Reduced complexity of contract relationships
+
+With all changes implemented as above, we need to answer the question if the
+service contract is still needed and whether we should not implement a simple
```
beaushinkle, line 150:
```suggestion
service contract is still needed and whether we should implement a simple
```

beaushinkle:
Did a grammar pass which was a really useful was to try to carefully understand exactly what was being discussed.

Also! I know that I made a borderline obnoxious amount of grammar suggestions, but I think upfront clarity work pays long-term dividends. https://www.grammarly.com/grammar-check is also super useful for catching subject/verb agreement and whatnot

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Please see https://github.com/keep-network/keep-core/pull/2585/commits/827175f76620b810c5520e3eef2a7af5f9941c81. The two sections below are the major pain points, I should be more explicit about it.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Please see https://github.com/keep-network/keep-core/pull/2585/commits/ce809c28d2dea1e5edb21493021560b14b718084. This is related to the pricing scheme. Because we did not have a gas price oracle that could not be manipulated, we decided to use a governable gas price ceiling. However, the gas price over the course of a week could jump from 20 Gwei to 200 Gwei, and some fees calculated based on the ceiling were just too high. As a result, some pools, like a pool for group creation were getting filled too soon and groups were getting created too often.

Looking how gas prices are volatile today I think we should decouple a frequency of system events with the gas price, wherever possible.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Unfortunately, we do not have an RFC describing it. I could link to https://docs.keep.network/random-beacon/index.html#_group_formation but this content will change for v2.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,157 @@
...
+
+Both on-chain and off-chain components proved to be successful but - just like
+with any other product - some pain points were identified by the users. This RFC
+aims at describing the pain points and proposing changes that should make the
+beacon easier to operate and more consistent with tBTC v2.
+
+=== ETH-based pricing
+
+Having reimbursements for various on-chain actions in ETH was a nice feature but
+also complex in practice. Highly volatile gas price led to creating new groups
```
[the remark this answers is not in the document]
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Partially done in https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25. Regarding "transaction" vs "transactions": there should be one callback transaction submitting the result and one (separate) transaction to do something with that result. That's why I went with singular here. Does it make sense?

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Same explanation as for the previous comment.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/2a0d1e1c217328743c47a0b1363a3e19c10e170e

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: https://github.com/keep-network/keep-core/pull/2585/commits/8211cc1bfdab624eb0ffaf593ad9eaed7e1e6b25

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,169 @@
...
+and that the price relation between gas price and T reimbursement for relay
+entry submission is unknown at the moment the fee is set by the governance, we
+should calculate a delay factor and delay penalty as in v1 of the random beacon.
+The delay penalty should be subtracted from reimbursement and sent to the
+sortition-pool-wide reward pool. The remaining part is sent to the relay entry
+submitter. Over the course of 64x6 blocks, reimbursement goes to zero. On-chain we
+are not going to enforce any order of submitters but off-chain clients should
+agree to the order where the member with index 1 has 6 blocks to submit, after 6
+blocks member with index 2 tries to submit, after 12 blocks member with index 3
+tries to submit and so forth.
```
eth-r, line 86: We can have this but someone is probably going to modify their client to front-run it whenever profitable; I definitely would.

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,169 @@
...
+
+Anyone should be able to submit a transaction creating a new group and this
+action should be detached from relay entry submission. The submitter of this
+transaction receives a part of the group creation pool and the remaining part
+should be reserved for the DKG result submitter as reimbursement for a future
+transaction. What percentage of the pool goes where is a governable parameter
+and should be determined based on the gas cost of the two transactions.
```
eth-r, line 93: This is going to require some work to figure out with the new static sortition pool design. Ideally we want to lock the sortition pool before the entry for the new group creation can be known, so the exact sequence has some interactions that need sorting out.

docs/rfc/rfc-19-random-beacon-v2.adoc
beaushinkle: I love this link-to-the-fixing-commit strategy

docs/rfc/rfc-19-random-beacon-v2.adoc
beaushinkle: Yeah absolutely!

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: I think this is generally problematic if we do not enforce the order on-chain.

Options are:
- Implement a delay factor, do not enforce order on-chain - we can end up with the situation you described.
- Eliminate delay factor, do not enforce the order on-chain, the submitter is always getting the whole reward. This option is problematic because if the reward is hefty enough - and it probably has to be so that we do not end up adjusting it every day - we can end up with the same situation you described.
- Implement a delay factor, enforce order on-chain - last members in the group are in a much worse position. 
- Eliminate delay factor, enforce order on-chain - everyone in the group is responsible to submit the entry to avoid slashing, rules are the same for everyone.

It feels that eliminating delay factor and enforcing the order is the simplest option and most fair in the model when the entire sortition pool is rewarded. WDYT?

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: One option would be to have a governable number determining the frequency of creating new groups per the number of entries generated. With the new static sortition pool design, the transaction requesting to create a new group is no longer expensive so we can not use it to guess the price of T and gas and assume DKG result submission will be net positive.

Let's say that a new group would be created every 20 entries. When relay request n*20+1 is submitted to the chain, it marks that the new group creation is in progress and locks the sortition pool. Now, stakers execute sortition view function to determine who is selected and start the DKG. Once the result is published, the call to sortition pool is executed saying it to unblock the state once 100(?) additional blocks passes.

What do you think?

docs/rfc/rfc-19-random-beacon-v2.adoc
eth-r: Eliminating delay factor and enforcing the order has the problem that while any operator has an equal chance of being the first in the long term the windfall from having position 1 in a group would still be relatively significant. If we pay the full submitter reward but some operator has a persistent advantage in being the first one allowed to submit it would increase the noisiness of operator returns a lot.

If we use delay factor while enforcing order on-chain it would mean that we could start at zero reward and grow it until some operator determines it to be worth it to submit.

That would make entries slower to be submitted while eliminating the windfall from being the first; an actor controlling the first few operators could deliberately stall the submission to get a bit of an extra reward without risk of being frontrun by someone else, but that is unlikely to be massively impactful (someone controlling 25% of operators has a 1:1,000 chance of getting the first 5 members in a group).

docs/rfc/rfc-19-random-beacon-v2.adoc
eth-r: > When relay request n*20+1 is submitted to the chain, it marks that the new group creation is in progress and locks the sortition pool.

If we lock the pool when someone _requests_ the group-creating entry, it would work fine. It would slightly increase the cost of requesting entries sometimes, but if we do the sortition pool correctly it should only be 6-7k gas + change.

However, we need to make sure that relay requests aren't submitted too often; if the pool is locked for an average of 120 blocks every DKG, if entries are requested every 6th block on average the pool would practically always be locked. We could enforce a minimum number of blocks or seconds between group creations to alleviate this risk faster than governance can respond.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: In the model proposed, we'd have:

delay factor = [ (T_deadline - T_received) / (T_deadline - T_begin)]^2
submitter reward = base reward * (1 - delay factor)

and there is no extra reward for other group members, correct?

The client would have to read KEEP/ETH price from some oracle and decide if it makes sense to submit. I like that the reward is adjusted to the current market conditions.

On the other hand, there are two pieces I dislike. First, it will take a longer time to submit a relay entry as you noted. Secondly, governance needs to adjust the base reward to make sure 100% of it covers the gas costs. It feels similar to the gas price ceiling. 

Another option would be not giving any submitter reward and assume rewards given to sortition pool members are enough. We would not enforce any order on-chain but off-chain, we'd implement a handshake agreement that the first member to submit is `entry % group_size`, then `entry % group_size + 1` and so on, and the only incentivize for submitting the entry is not to get slashed. It might not be that scary as it might feel given how ECDSA clients worked for tBTC v1. They were not reimbursed for any transaction and we had no complaints.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Good point. So let me try to summarize it:
- New group creation is triggered with a fixed (yet governable) frequency of entries, `dkg_freq`.
- When someone requests for a new entry and the counter says we should trigger DKG with the new entry  (we are at `dkg_freq - 1`), we lock the sortition pool.
- Sortition happens optimistically: the pool is locked, members are determined via a view function call, there is some time for complaints before the group is considered active, sortition pool is unlocked after the complaints period times out.
- Relay request transaction locking the pool, triggers new DKG.
- No one is getting reimbursed for triggering DKG.

Does it make sense?

In this model, we will have to figure out how to reward DKG result submitter - this model does not require a third party to submit expensive transaction so we can not estimate current market conditions based on it. 

Given that this action is not happening as often as relay entry submission but is much more expensive, maybe it is enough if we set a fixed, governable reward for submitting an entry and enforce order on-chain? This reward does not always need to cover the costs but might be used just to alleviate the pain. What do you think?

eth-r: Given that triggering DKG is just a simple check + state change (locking the pool) I agree that it can be wrapped into the entry request transaction.

For DKG result submission we could allocate a reward (possibly a fraction of the maintenance reward pool), but have the submitter reward grow over time from zero to the maximum so that higher gas costs would lead to DKG results being submitted slower and vice versa. Since the submission has to verify `threshold + 1` ECDSA signatures, plus create a new wallet and record the members, it's still relatively expensive. (We might be able to reduce the cost by encoding the operator IDs as `immutable` and only having a mutable bitstring denoting their status, but it would be more complex than densely packed storage slots.)

After the challenge period is over, one more transaction is needed for confirming the group and unlocking the sortition pool and releasing the submitter reward (we don't want to pay out rewards for fraudulent submissions). This again would be fairly cheap, and because it releases the reward the submitter would have an incentive to call it anyway. The group could be mildly griefed by the submitter not confirming it, but any other member (or indeed any address whatsoever) could do it just as well. We could also wrap the confirmation into the next entry request or submission after the challenge period is over, to make it happen automatically, but I think requiring a separate transaction is simpler and doesn't rely on entry frequency.

docs/rfc/rfc-19-random-beacon-v2.adoc
eth-r: > In the model proposed, we'd have:
> 
> delay factor = [ (T_deadline - T_received) / (T_deadline - T_begin)]^2
> submitter reward = base reward * (1 - delay factor)
> 
> and there is no extra reward for other group members, correct?

Yes. I had to read that math a couple of times but I think it's correct (reward starts at 0 and grows).

> On the other hand, there are two pieces I dislike. First, it will take a longer time to submit a relay entry as you noted. Secondly, governance needs to adjust the base reward to make sure 100% of it covers the gas costs. It feels similar to the gas price ceiling.

The base reward would not entirely need to cover the gas costs, as the group would get slashed if it misses the deadline; thus we only need to ensure that the amount slashed from a single operator covers the gas costs in all conditions.

If we want the price to increase faster, we could remove the `^2` and have the reward grow linearly. This increases the chance of the first submitter receiving a windfall but I think it's fine if the deadline is reasonably long.

> Another option would be not giving any submitter reward and assume rewards given to sortition pool members are enough. We would not enforce any order on-chain but off-chain, we'd implement a handshake agreement that the first member to submit is `entry % group_size`, then `entry % group_size + 1` and so on, and the only incentivize for submitting the entry is not to get slashed. It might not be that scary as it might feel given how ECDSA clients worked for tBTC v1. They were not reimbursed for any transaction and we had no complaints.

I like this way of ordering the submission order to spread the pole positions out over time, and this could even work with a windfall submitter reward that regularly exceeds gas costs. An off-chain handshake works reasonably well but it still has the problem that an operator/staking provider who makes their clients never submit entries (or only submit entries if they're rewarded above the gas cost) would still gain a benefit over time.

One possibility is that we could use the entry submission as a heartbeat; if you are in line for the submission (by the `entry % group_size` rule) but don't submit, you're deemed to be inactive and removed from the sortition pool, forcing the operator to rejoin the pool (possibly after a short ban period) and forfeit the rewards.

pdyraga: I really like the idea of correlating the submission with a heartbeat. I think that if we eliminate most of the existing pricing complexity from the relay entry submission transaction, we could go below 200k gas (eyeballing) with just BLS verification and submission order check/inactivity. Assuming that the responsibility is spread between group members, and the gas cost is relatively low, would you be comfortable with eliminating the submitter reward for relay entry? I make a quiet assumption that sortition pool rewards are good enough so that operators are not complaining about 200k gas transaction submitted from time to time.

eth-r: > Assuming that the responsibility is spread between group members, and the gas cost is relatively low, would you be comfortable with eliminating the submitter reward for relay entry? I make a quiet assumption that sortition pool rewards are good enough so that operators are not complaining about 200k gas transaction submitted from time to time.

Yes, I think round robin duty to submit would be perfectly suitable without separate rewards.

If we want to penalise inactivity more than just being kicked out of the sortition pool, we could add slashing for each operator who failed to submit within their window. If the amount slashed per operator grows over time, and the number of slashed operators also grows, the total slashed amount would grow to the square of the time that has elapsed and it could also achieve the effect of bleeding the group if it is too late to submit.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Increasing the reward seems interesting but I am worried that the additional complexity may not be worth it. It is similar to relay entry submission - except that the time it takes to submit DKG result might not be that important - it may happen that even 100% of the reward does not cover the cost, and we do not want operators to race for submission and have 63 transactions reverted when the price reaches profitability point plus some bonus. 

What I think we could do: 
- have a fixed, governable reward for DKG result submission
- enforce the order on-chain, member 1 is eligible first; then 1 and 2; then 1, 2, 3; then 1, 2, 3, 4;...
- member who was supposed to submit the result but did not is removed from the sortition pool

The rest stays as you say: two transactions, first locking the pool, and second approving the result, sending the reward to submitter and unlocking the pool. There will be some timeout for DKG result submission and in case it was not submitted on time, anyone can unlock the pool and be rewarded for it. Locking the pool will be wrapped in the entry request transaction.

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,225 @@
...
+are also done only in T.
+
+Relay requester should provide a fee in T. The value of the fee is a governable
+parameter. The entire fee is deposited in the maintenance pool that is used to
+reimburse for different actions related to DKG.
+
+The transaction submitting relay entry is not reimbursable and implementation
+should ensure the gas cost of this transaction is as low as possible, below
+200k gas. The first group member eligible to submit the result is 
+`previous_entry MOD group_size`, then also `(previous_entry MOD group_size) + 1`.
```
eth-r, line 106: why not `new_entry % group_size`? this way it would be unpredictable so there would be no safe way to be inactive

pdyraga: Hmm... yes, even better!

pdyraga: I really like it!

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+In case the DKG result has not been submitted on time, anyone can unlock the pool
+and receive a fixed, governable reward for reporting DKG timeout. The reward is
+paid from the maintenance pool.
+
+Just like in the case of relay entry submission, the order in which operators
+are supposed to submit DKG result is enforced on-chain and operators who missed
+their round are removed from the sortition pool for at least 2 weeks and do
+not earn rewards for that time. The first member eligible to submit the DKG result
+is a member with index `1`, then also `2`. For example, if member `9` submitted
```
eth-r, line 137: we could also use `hash(public_key) % group_size` for unpredictability

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,225 @@
...
+The fact the full - even the most complex - callback is executed in the same
+transaction in which the relay entry is submitted gives an impression of better
+security. This impression is false though, given that the entry to be submitted
+is visible in the mempool and smart attackers can have their transactions mined
+faster to, for example, put the sortition pool using the entry for group
+selection in the desired state. This issue needs to be solved on the sortition
+pool side with initiation time for new operators in the pool and/or state lock.
+
+To ensure predictable cost of relay entry submission transaction, we should
+eliminate callbacks altogether. It is up to the application to act when a new
```
eth-r, line 174: I'm not 100% sure of this, compared to enforcing a low gas budget (e.g. 100k) for the callback, but may warm up to the idea

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+With all changes implemented as above, we need to answer the question if the
+service contract is still needed and whether we should implement a simple
+beacon upgradeability scheme in the application using the beacon. In v1, the
+service contract takes care of executing the callback and reimbursing the
+operator. It also estimates entry fee, and determines if it is possible to
+create a new group on the most recent operator contract. All these
+functionalities, needed in v1, are simplified in this RFC so the service
+contract may no longer be needed, and removing it from the execution path could
+save some gas.
```
eth-r, line 207: agreed on removing service contract

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+=== Group creation
+
+New groups should be created with a fixed, governable frequency of relay
+requests. Instead of a ticket-based approach for group selection, we should
+use sortition pools. Group creation start transaction should be embedded into
+relay request transaction and should lock a sortition pool. From this moment,
+no operator can enter or leave the pool. All off-chain clients should perform
+group selection optimistically, using a view sortition function call and perform
+DKG. Once the group submits the result to the chain, a challenge period starts.
```
eth-r, line 67: we should note that the sortition is based on the entry whose request locked the sortition pool

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Good point. I removed low gas budget callbacks in the last commit making sure relay entry submit transaction is as cheap as possible, moving the responsibility/complexity entirely to the application. But now I think it could be too much for the application to handle...

Let's look at it from the perspective of tBTC v2.

With a callback: someone sees that the maintenance pool allows creating a new wallet and they submit a transaction. In this transaction, sortition pool is locked and a new relay entry is requested. Beacon produces an entry and executes a callback on the application. The callback stores a new seed and emits an event that is handled by clients performing sortition and generating group public key.

With no callback: someone sees that the maintenance pool allows creating a new wallet and they submit a transaction. In this transaction, sortition pool is locked and a new relay entry is requested. Beacon produces an entry and stores it in the beacon contract. Off-chain clients of tBTC v2 need to monitor the beacon contract to know when a new entry was submitted so that they can perform sortition and start generating group public key. This version is more complex but still possible to handle. The real complexity though begins when someone wants to challenge the submitted DKG result: we'd need to make sure the same relay entry is used for resolving the challenge as the one that was used by clients to perform sortition. 

It seems to me that implementing a low gas budget callback is much simpler than handling synchronization of entry on the application side.

What I'd propose to do:
- non-reimbursable, low gas budget callback,
- initial limit set to 50k gas or less: should be enough to SSTORE entry and emit an event,
- gas limit for the callback will be a governable value,
- for simplicity, we'd require that the contract being called has a function with a specific signature so that the only callback parameter is the contract address.

What do you think?

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Agreed, it's better.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Good point, the current version is not clear about it at all. I'll improve it.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Please see https://github.com/keep-network/keep-core/pull/2585/commits/5ca9187cf42019592e814451eb2d8935c9d41dde.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Please see https://github.com/keep-network/keep-core/pull/2585/commits/5ca9187cf42019592e814451eb2d8935c9d41dde.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Please see https://github.com/keep-network/keep-core/pull/2585/commits/0c97106900a6469bcb4e2be2d7927fed24398f54.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Please see https://github.com/keep-network/keep-core/pull/2585/commits/c719de341690bfeb6c20cf0e47b566157a37b8ca.

eth-r (dismissed):
Looking good

lukasz-zimnoch (dismissed):

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,252 @@
...
+    /// @param blockNumber Block number at which the relay entry was submitted
+    ///                    to the chain.
+    function __beaconCallback(uint256 relayEntry, uint256 blockNumber) external;
+}
+```
+
+=== Timeouts
+
+We should extend the timeout for submitting relay entry to give operators more
+time to react. After 64x10 blocks from the relay request, if no entry was
```
michalinacienciala, line 213: It is not clear to me whether the `64x10 blocks` number comes from `group size` x `the number of blocks for a member to become eligible to submit relay entry` or is it a hardcoded or governable value.

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,252 @@
...
+relay request transaction and should lock a sortition pool. From this moment,
+no operator can enter or leave the pool. Once a new relay entry appears on the
+chain, all off-chain clients should perform group selection optimistically, 
+using the new entry value and a view sortition pool function call.
+After determining group members, clients should perform DKG as usual. 
+Once the group submits the result to the chain, a challenge period starts.
+
+During the challenge period, anyone can notify that the submitted DKG result
+contains group members not selected by the pool. This leads to slashing all
+members who signed the result and the notifier receives 5% from the slashed amount.
```
michalinacienciala, line 73: What should happen (who will get paid the 5%) in case there will be more than one notification (is that even possible)?

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,252 @@
...
+faster to, for example, put the sortition pool using the entry for group
+selection in the desired state. This issue needs to be solved on the sortition
+pool side with initiation time for new operators in the pool and/or state lock.
+
+Instead of executing any callback in the same transaction in which relay entry
+is submitted, we should allow only simple, low gas budget callbacks with a gas
+limit controlled by the governance. Applications wanting to use a relay entry
+should submit another transaction using the relay entry value previously set by
+the random beacon. They should also employ an additional security check making
+sure the entry submitted is only valid for a certain number of blocks to avoid
```
michalinacienciala, line 183: Is that `certain number of blocks` a governable parameter?

michalinacienciala (commented):
I'm adding a couple of comments regarding some areas that were not clear to me.

meercode-keep (dismissed):

dimpar (approved):

docs/rfc/rfc-19-random-beacon-v2.adoc
```diff
@@ -0,0 +1,252 @@
...
+contract may no longer be needed, and removing it from the execution path could
+save some gas.
+
+== List of governable parameters
+
+* Relay request fee in T
+* Reward for submitting DKG result
+* Reward for unlocking the sortition pool if DKG timed out
+* Slashing amount for not submitting relay entry
+* Slashing amount for submitting malicious DKG result
```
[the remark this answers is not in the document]
pdyraga: Good catch. Fixed in https://github.com/keep-network/keep-core/pull/2585/commits/b0b53b62c4d5e6a493dbaca96e1f314888d1faa4.

docs/rfc/rfc-19-random-beacon-v2.adoc
pdyraga: Not for the random beacon. The random beacon does not care if and how quickly the application will use the relay entry the beacon produced. It is up to the application to have the checks needed.

pdyraga:
I improved some parts after the feedback from Michalina. @eth-r @michalinacienciala can you please take another look? I think we should be good to merge if you approve the changes.

michalinacienciala (approved):

eth-r (approved):

PR #2594, opened 2021-09-06
Adding cron for daily tests and builds on Meercode

dimpar:
The purpose of this PR is to set a cron job that would run tests and builds on a daily basis (at midnight) so the results can be displayed on Meercode dashboard.

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -1,4 +1,4 @@
-name: Go / keep-core
+name: Go
 
 on:
   schedule:
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -1,6 +1,8 @@
-name: Go
+name: Go / keep-core
 
 on:
+  schedule:
+    - cron: '0 0 * * *'
   push:
     branches:
       - main
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -1,6 +1,8 @@
-name: Solidity
+name: Solidity / keep-core
 
 on:
+  schedule:
+    - cron: '0 0 * * *'
   push:
     branches:
       - main
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -1,4 +1,4 @@
-name: Solidity / keep-core
+name: Solidity
 
 on:
   schedule:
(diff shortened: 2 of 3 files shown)


.github/workflows/client.yml
```diff
@@ -1,6 +1,8 @@
-name: Go
+name: Go / keep-core
```
michalinacienciala, line 1: I wonder if we really need to change the name. I know it would make it easier to differentiate the workflows on the meercode dropdown list, but in the filters we apply there we'll want to display all or no `Go` workflows, so differentiating is not that important, I think.
The downside of having the name change is that having some of the workflows with the ` / keep-core` name and some without them in the GH Actions view would look a bit inconsistent. Apart from that we would see the `/ keep-core` duplicated on the PR checks lists.
Both downsides are not that serious, so I'm ok with the change if it is really necessary.

.github/workflows/client.yml
```diff
@@ -1,6 +1,8 @@
...
 
 on:
+  schedule:
+    - cron: '0 0 * * *'
```
michalinacienciala, line 5: As we have E2E tests in `local-setup` running at 0:00, I would suggest changing the cron here for `0 2 * * *`, to decrease the load on the GH Actions.

michalinacienciala (commented):
The comments I left apply to all the modified files.

.github/workflows/client.yml
dimpar: Yes, it was added to make it easier to navigate in Meercode. Okay, we can remove it and we'll see if that would be a problem navigating in Meercode. We can always modify the name if needed. 15d0c90

.github/workflows/client.yml
dimpar: Right, but doesn't GitHub Actions manage this sort of issues itself through some balancers? Or adding some delay to the existing job? We will need to change around 10 repos in total and potentially more in the future. Keeping track of what is scheduled when might be a bit problematic when each repo is scheduled at different times. I would leave it at 0:00 and let GH sort things out, but if you know that it would be a problem, then sure, let's change it.

michalinacienciala: Yeah, maybe I'm just being over-cautious here - I don't know of any problems with parallel runs. You're right about keeping track of the times of cron executions, let't keep it at 0:00, we can always change it in the future if any problems occur.

michalinacienciala (approved):

PR #2597, opened 2021-09-08
Decimal places in Cov Pools

michalsmiarowski:
Allows to write decimal places in Cov Pools forms and also prints KEEPs and CovKEEPs with decimals

### DIFF
--- a/solidity/dashboard/src/forms/common-validators.js
+++ b/solidity/dashboard/src/forms/common-validators.js
@@ -3,7 +3,6 @@ import { KEEP } from "../utils/token.utils"
 
 const REGEXP_NOT_BLANK_STRING = /^\s*$/
 const REGEXP_ONLY_NUMBERS = /^\d+$/
-const REGEX_DOT = /[.]/g
 
 const isBlankString = (value) => {
   return !value || REGEXP_NOT_BLANK_STRING.test(value)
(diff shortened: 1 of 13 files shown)


solidity/dashboard/src/forms/common-validators.js
```diff
@@ -22,25 +23,43 @@ export const validateEthAddress = (address, required = true) => {
...
-  const formatedValue = value
-    ? web3Utils
-        .toBN(value)
-        .mul(web3Utils.toBN(10).pow(web3Utils.toBN(18)))
-        .toString()
-    : 0
-  const validateValueInBN = web3Utils.toBN(formatedValue)
+  let formattedValue = value
+  try {
+    formattedValue = value ? web3Utils.toWei(value) : 0
```
r-czajkowski, line 52: I would suggest using `const formatedValue = token.fromTokenUnit(value)` because not every token has a precision of 18 decimal places. Thoughts? Please see https://github.com/keep-network/keep-core/blob/2a0d1859ec814d3a77394d529abf0214c59027d0/solidity/dashboard/src/forms/common-validators.js#L33-L56 and let me know if this makes sense to you.

r-czajkowski (changes requested):

solidity/dashboard/src/forms/common-validators.js
michalsmiarowski: 8ce176f3cc8cd0e842499b0c2d5f7a601d1a3c08

r-czajkowski (approved):
:shipit:

PR #2598, opened 2021-09-10
Verify contracts on Etherscan [WIP]

michalinacienciala:
We're adding a step verifying the migrated contracts on Etherscan. We're
using `truffle-plugin-verify` plugin. The `etherscan-verify.sh` script
contains a list of contracts which are being migrated and which should
be verified. If we'll add a new contract to the project, we'll need to
add the name of the new contract to the list in the script.


michalinacienciala:
Here is the run in which contracts were successfully verified (it didn't have the annotations set yet): https://github.com/keep-network/keep-core/runs/3595973768?check_suite_focus=true.
Here is how annotations are displayed (tested on tbtc): https://github.com/keep-network/tbtc/actions/runs/1240880015.

michalinacienciala:
I decided to store the `annotation.json` in `.github` folder, but maybe there's some better place to keep it? Any suggestions are welcome :)

michalinacienciala:
Functionality will be developed in a separate PR: https://github.com/keep-network/keep-core/pull/2644. It was more convenient to create a new feature branch from `main` than resolve the conflicts after implementing the changes.

Issue #2591, opened 2021-09-02
Visual QA Coverage Pool -- Part: the First


liz-shinn:
# Visual QA review of the testnet version of the coverage pool. Overall looking really solid! Awesome job to all 🎊 

### Overview Page (https://dashboard.test.keep.network/overview)

**Pending withdrawal should only show when there is a pending withdrawal.**
Recommended fix: Remove this when there is no pending withdrawal.
<details>
<summary>Screenshot</summary>
<img width="1130" alt="Screen Shot 2021-09-02 at 8 18 39 PM" src="https://user-images.githubusercontent.com/57226633/131897189-2347243a-8fe9-43eb-b53d-67dea9dae070.png">
</details>

**Type style in purple banner should be Work Sans SemiBold. Sentence should end with a period.**
Recommended fix: Update font weight to SemiBold. Add period to the end of the sentence `...and earn rewards.`
<details>
<summary>Screenshot</summary>
<img width="1110" alt="Screen Shot 2021-09-02 at 8 33 18 PM" src="https://user-images.githubusercontent.com/57226633/131898100-14139b72-0030-476c-873d-4656d34f9ab8.png">
</details>

**Button height for secondary button doesn't match button height for primary button.**
Recommended fix: Update button heights for primary & secondary button styles to both be 56pt.
<details>
<summary>Screenshots</summary>
<img width="1067" alt="Screen Shot 2021-09-02 at 8 41 31 PM" src="https://user-images.githubusercontent.com/57226633/131899037-aafb6ada-6736-4b15-b7fe-1758ca3f554a.png">
<img width="288" alt="Screen Shot 2021-09-02 at 8 41 26 PM" src="https://user-images.githubusercontent.com/57226633/131899055-de42b9af-2bd3-45cd-9425-cbcbe176f400.png">
</details>

------

### How it Works Page (https://dashboard.test.keep.network/coverage-pools/how-it-works)

**Timeline cards: body copy line height is too loose. Need more space between timeline card title and body copy.**
Recommended fix: Body copy line height looks to be 1.5rem -- updating it to 1.4rem should do the trick. Add 8px more space between card title and body copy in timeline. 
<details>
<summary>Screenshots</summary>
<img width="494" alt="Screen Shot 2021-09-02 at 8 48 02 PM" src="https://user-images.githubusercontent.com/57226633/131899777-37d48b53-0402-4861-8e48-797bfca98bd9.png">
[...]

PR #2600, opened 2021-09-13
Cov pools improvements

michalsmiarowski:
Fixes: #2591 

Styling improvements for cov pools:
- [x] hide **Pending Withdrawal** section in **Overview** page when there are no pending withdrawals
- [x] add weight and dot at the end to the tile of the coverage pools banner
- [x] set same height for primary, secondary and tertiary buttons
- [x] add more space between title and description in the timeline card
- [ ] ~~add more side margins to the cov pools page~~ (the main objective behind this was to prevent tooltips from being cutted out of the viewport. Since it was resolved i decided to not adding more margins here)
- [x] make sure tooltips aren't cut out in some resolutions (when they are close to borders)<details><summary>Screenshot - tooltip fix</summary>
image
</details>

### DIFF
--- a/solidity/dashboard/src/css/banner.less
+++ b/solidity/dashboard/src/css/banner.less
@@ -123,7 +123,10 @@
     .banner__title {
       &:extend(.h3);
       margin-bottom: 2rem;
-      font-weight: 600;
+
+      &--font-weight-600 {
+        font-weight: 600;
+      }
     }
 
     .banner__icon {
--- a/solidity/dashboard/src/css/banner.less
+++ b/solidity/dashboard/src/css/banner.less
@@ -123,6 +123,7 @@
     .banner__title {
       &:extend(.h3);
       margin-bottom: 2rem;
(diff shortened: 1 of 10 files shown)


solidity/dashboard/src/css/banner.less
```diff
@@ -123,6 +123,7 @@
     .banner__title {
       &:extend(.h3);
       margin-bottom: 2rem;
+      font-weight: 600;
```
r-czajkowski, line 126: Shouldn't we update only for coverage pool banner on the overview page? It will change styles of all banners.

r-czajkowski (changes requested):

solidity/dashboard/src/css/banner.less
michalsmiarowski: 91482d3373715dbe087f6d1974a53f69891171e2

r-czajkowski (approved):
:shipit:

PR #2603, opened 2021-09-16
Change type of amount sent when depositing

michalsmiarowski:
We are changing the amount type from BigNumber to string because it looks like
it transactions doesn't work with the amount as BigNumber.


r-czajkowski (approved):
👍

PR #2604, opened 2021-09-17
Coverage Pools tests

michalsmiarowski:
Adds test cases to make sure the coverage pools events subscription works perfectly.

Also adds test that check if the correct success modal was displayed after a successful transaction.

The events that are tested here:
- Deposit
- WithdrawalInitiated
- WithdrawalCompleted


solidity/dashboard/src/__tests__/saga/coverage-pool.test.js
```diff
@@ -257,104 +268,899 @@ describe("Coverage pool saga test", () => {
...
-  //         totalValueLockedInUSD: tvlInUSD,
-  //         apy: updatedAPY,
-  //       })
-  //       .run()
-  //   })
-  // })
+
+  describe("Subscribe to Asset Pool's Deposit event", () => {
+    it("should update data correctly if the `Deposit` event has been emitted by current logged user", () => {
+      const address = "0x086813525A7dC7dafFf015Cdf03896Fd276eab60"
```
r-czajkowski, line 274: Some values are duplicated in test cases. Maybe we can define them one level higher?

r-czajkowski (changes requested):
Good job! Left a tiny comment otherwise looking very good! 🚀

solidity/dashboard/src/__tests__/saga/coverage-pool.test.js
michalsmiarowski: 8e0cbf567a0e304548f8bd8a0b9d8ec3efd59e6b

r-czajkowski (approved):

PR #2605, opened 2021-09-17
Fix bug when updating data after withdrawal is completed

michalsmiarowski:
There is a bug that causes the Pending Withdrawal to disappear when another user
complete his own withdrawal. The pending withdrawal for current user was not
visible in such case unless he refresh the page. This PR fixes that.


r-czajkowski (approved):
LGTM! 🦅 👁️

PR #2571, opened 2021-08-16
tBTC token upgrade portal

r-czajkowski:
Depends on: #2500 - we need some changes from this branch.

WIP...

TODO:
- [ ] integration with tbtc v2 Vending Machine contract

<details>
  <summary>Screenshots</summary>

  
obraz

obraz

obraz

</details>

PR #2609, opened 2021-09-21
Update dashboard dependendencies to tbtc-v2 project

michalinacienciala:
The `tbtc-v2` contracts intended to be used for building of Token
Dashboard will be published to the NPM registry under name with the
`-dev` suffix. We need to update the dependencies configuration in
`package.json` and `package-lock.json` dashboard files to use those
published packages.
For deployment of Dashboard on Testnet we're not using the dependencies
configured in the `package-lock.json`, instead we specify the exact version
we want to use when running the deployment workflow.

Ref: https://github.com/keep-network/keep-core/pull/2571

### DIFF
--- a/.github/workflows/dashboard-testnet.yml
+++ b/.github/workflows/dashboard-testnet.yml
@@ -81,6 +81,7 @@ jobs:
               @keep-network/keep-core \
               @keep-network/keep-ecdsa \
               @keep-network/tbtc \
+              @keep-network/tbtc-v2 \
               @keep-network/coverage-pools
 
       - name: Get upstream packages' versions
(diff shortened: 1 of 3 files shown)


.github/workflows/dashboard-testnet.yml
```diff
@@ -102,6 +104,7 @@ jobs:
               @keep-network/keep-core@${{ steps.upstream-builds-query.outputs.keep-core-contracts-version }} \
               @keep-network/keep-ecdsa@${{ steps.upstream-builds-query.outputs.keep-ecdsa-contracts-version }} \
               @keep-network/tbtc@${{ steps.upstream-builds-query.outputs.tbtc-contracts-version }} \
+              @keep-network/tbtc-v2@${{ steps.upstream-builds-query.outputs.tbtc-contracts-version }} \
```
r-czajkowski, line 107: Shouldn't be`steps.upstream-builds-query.outputs.tbtc-v2-contracts-version` instead of `...tbtc-contracts-version`?

r-czajkowski (changes requested):

.github/workflows/dashboard-testnet.yml
michalinacienciala: Good catch, fixed in 606353bfb00998df6073fee9c6a7719e5aa335cc.

r-czajkowski (approved):
LGTM! :shipit:

PR #2614, opened 2021-09-23
Liquidity page updates

r-czajkowski:
Depends on: #2571 

This PR updates the Liquidity page by adding info banners to the KEEP only and TBTC/Saddle cards indicating the
incentives for these pools have been removed. Also it adds a new card for the tbtcv2/Saddle pool.

<details>
  
  
obraz

obraz

obraz
</details>

### DIFF
--- a/solidity/dashboard/src/components/KeepOnlyPool.jsx
+++ b/solidity/dashboard/src/components/KeepOnlyPool.jsx
@@ -19,6 +19,8 @@ import { useModal } from "../hooks/useModal"
 import TokenAmount from "./TokenAmount"
 import MetricsTile from "./MetricsTile"
 import RewardMultiplier from "./liquidity/RewardMultiplier"
+import Banner from "./Banner"
+import { LINK } from "../constants/constants"
 
 const KeepOnlyPool = ({
   apy,
@@ -139,11 +141,37 @@ const KeepOnlyPool = ({
               &nbsp;KEEP
             </h4>
           </div>
-
+          <Banner
+            className="liquidity-info-banner liquidity-info-banner--warning mt-2"
+            style={{ maxWidth: "100%" }}
+          >
+            <Banner.Icon
+              icon={Icons.Warning}
+              className="liquidity-info-banner__icon liquidity-info-banner__icon--warning"
+            />
+            <div className="liquidity-info-banner__content">
+              <Banner.Title className="text-grey-60">
+                Incentives removed
+              </Banner.Title>
+              <Banner.Description className="liquidity-info-banner__content__description text-grey-60">
+                The incentives for this pool has been removed and you can no
+                longer deposit the KEEP tokens. You can still withdraw deposited
+                KEEP tokens and rewards that you already earned.&nbsp;
+                <a
+                  target="_blank"
+                  rel="noopener noreferrer"
+                  href={LINK.proposals.shiftingIncentivesToCoveragePools}
+                  className="text-grey-60"
+                >
+                  More info
+                </a>
+              </Banner.Description>
+            </div>
+          </Banner>
           <div className="flex row space-between mt-2">
             <SubmitButton
               className="btn btn-primary btn-lg"
-              disabled={!gt(wrappedTokenBalance || 0, 0)}
+              disabled={true}
               onSubmitAction={addKEEP}
             >
               {gt(lpBalance, 0) ? "add more keep" : "deposit keep"}
--- a/solidity/dashboard/src/components/KeepOnlyPool.jsx
+++ b/solidity/dashboard/src/components/KeepOnlyPool.jsx
@@ -161,7 +161,7 @@ const KeepOnlyPool = ({
                   target="_blank"
                   rel="noopener noreferrer"
                   href={LINK.proposals.shiftingIncentivesToCoveragePools}
-                  className="text-grey-60"
+                  className="text-link text-grey-60"
                 >
                   More info
                 </a>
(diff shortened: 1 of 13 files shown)


solidity/dashboard/src/components/KeepOnlyPool.jsx
```diff
@@ -139,11 +141,37 @@ const KeepOnlyPool = ({
...
+              </Banner.Title>
+              <Banner.Description className="liquidity-info-banner__content__description text-grey-60">
+                The incentives for this pool has been removed and you can no
+                longer deposit the KEEP tokens. You can still withdraw deposited
+                KEEP tokens and rewards that you already earned.&nbsp;
+                <a
+                  target="_blank"
+                  rel="noopener noreferrer"
+                  href={LINK.proposals.shiftingIncentivesToCoveragePools}
+                  className="text-grey-60"
```
michalsmiarowski, line 164: Can we also add `text-link` class here so the style of the "More info" link is similar to the ones in TBTC + Saddle and KEEP + TBTC banners?

solidity/dashboard/src/contracts.js
```diff
@@ -346,10 +353,19 @@ export const createLPRewardsContract = async (web3, contractName) => {
   return await getContract(web3, artifact, {})
 }
 
-export const createSaddleSwapContract = (web3) => {
+export function createSaddleSwapContract(web3) {
```
michalsmiarowski, line 356: Just wondering - is there any reason to change `const` to `function` to define function here? 🤔

solidity/dashboard/src/sagas/utils.js
```diff
@@ -108,6 +110,7 @@ export function* subscribeToEventAndEmitData(
         : action(event)
       yield put(_action)
     } catch (error) {
+      console.log("error", error)
       console.error(`Failed subscribing to ${_debugName} event`, error)
```
michalsmiarowski, line 114: Do we need `console.log` here since we have `console.error`?

solidity/dashboard/src/services/liquidity-rewards.js
```diff
@@ -422,20 +419,24 @@ class TokenGeyserLPRewards extends LiquidityRewards {
...
+  [POOL_TYPE.SADDLE]: SaddleLPRewards,
+  [POOL_TYPE.TOKEN_GEYSER]: TokenGeyserLPRewards,
 }
 
 export class LiquidityRewardsFactory {
   /**
    *
-   * @param {('UNISWAP' | 'SADDLE' | 'TOKEN_GEYSER')} pool - The supported type of pools.
-   * @param {Object} LPRewardsContract - The LPRewardsContract as web3 contract instance.
+   * @param {('UNISWAP' | 'SADDLE' | 'TOKEN_GEYSER')} pool - The supported type
```
michalsmiarowski, line 430: We have `POOL_TYPE` constant defined, so maybe we could use it here as a `pool` type instead of `{('UNISWAP' | 'SADDLE' | 'TOKEN_GEYSER')}`

solidity/dashboard/src/pages/liquidity/LiquidityPage.jsx
```diff
@@ -16,16 +16,82 @@ import {
...
+    id: "TBTCV2_SADDLE",
+    title: LIQUIDITY_REWARD_PAIRS.TBTCV2_SADDLE.label,
+    liquidityPairContractName:
+      LIQUIDITY_REWARD_PAIRS.TBTCV2_SADDLE.contractName,
+    MainIcon: Icons.TBTC,
+    SecondaryIcon: Icons.Saddle,
+    viewPoolLink: LIQUIDITY_REWARD_PAIRS.TBTCV2_SADDLE.viewPoolLink,
+    pool: LIQUIDITY_REWARD_PAIRS.TBTCV2_SADDLE.pool,
+    lpTokens: LIQUIDITY_REWARD_PAIRS.TBTCV2_SADDLE.lpTokens,
+    wrapperClassName: "tbtc-v2-saddle",
```
michalsmiarowski, line 32: I think we should style the logos in this card and fill the tbtc logo with white color (we did the same in TBTC-SADDLE pool):

```
.tbtc-saddle& {
    .main-icon {
      circle {
        fill: @white;
      }
    }
  }
```

in `liquigidy-page.less`.

michalsmiarowski (changes requested):
Left some small comments, but overall looks really good 👍

solidity/dashboard/src/services/liquidity-rewards.js
r-czajkowski: We can add jsdoc for the `POOL_TYPE` variable like this:
```
/**
 * Enum defines a type of the liquidity pools.
 * @enum {string}
 */
const POOL_TYPE = { ... }
```

and then use it instead of `{('UNISWAP' | 'SADDLE' | 'TOKEN_GEYSER')}`. What do you think?

solidity/dashboard/src/contracts.js
r-czajkowski: Without going into details(because I don't know how to describe it correctly 😆) is about javascript hoisting. If we use the arrow function the app crashes with `ReferenceError: can't access lexical declaration 'createSaddleTBTCMetaPool' before initialization` because we import functions in `constants.js` from `contracts.js` and `contract.js` file imports from `constants.js`. Not sure if it's an actual solution so we should probably take a look at it in the future.

solidity/dashboard/src/components/KeepOnlyPool.jsx
r-czajkowski: 06f0146

solidity/dashboard/src/sagas/utils.js
r-czajkowski: c64211f

solidity/dashboard/src/pages/liquidity/LiquidityPage.jsx
r-czajkowski: c64211f

solidity/dashboard/src/contracts.js
michalsmiarowski: I see, let's keep it that way for now then.

solidity/dashboard/src/services/liquidity-rewards.js
michalsmiarowski: Let's keep it as it is

michalsmiarowski (approved):
🚢

PR #2627, opened 2021-09-29
Fix broken path filtering in dashboard-format.yaml

michalinacienciala:
Filtering applied in `dashboard-detect-changes` job was creating output
called `dashboard`, whereas the `dashboard-code-format` job was executng
conditionally if (non-existing) `path-filter` output was true or if the
triggering event was different than PR. This way the
`dashboard-code-format` job was never executed for the Pull Requests.
The naming of the output in `dashboard-detect-changes` has been changed
to `path-filter`, which makes it consistent with other workflows.

### DIFF
--- a/.github/workflows/dashboard-format.yaml
+++ b/.github/workflows/dashboard-format.yaml
@@ -17,7 +17,7 @@ jobs:
   dashboard-detect-changes:
     runs-on: ubuntu-latest
     outputs:
-      dashboard: ${{ steps.filter.outputs.path-filter }}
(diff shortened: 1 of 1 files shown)


michalinacienciala:
Workflow run on the modified code: https://github.com/keep-network/local-setup/runs/3729731465?check_suite_focus=true. It is still failing, but no longer due to incompatible Node versions (the new reason of failure will be fixed in https://github.com/keep-network/coverage-pools/pull/198).

r-czajkowski (approved):
:shipit:

PR #2628, opened 2021-09-29
Update KEEP balance after deposit and withdraw

michalsmiarowski:
Update KEEP Balance after depositing or withdrawing tokens in AssetPool.

Separate KEEP token transfer events to two different ones (from and to)

Also fix typo `arithmeticOpration` -> `arithmeticOperation`


r-czajkowski (approved):
LGTM! :shipit:

PR #2629, opened 2021-09-30
Accurate exchange rate KEEP/covKEEP with 3 decimals

r-czajkowski:
Accurate exchange rate KEEP/covKEEP with 3 decimals.

obraz

obraz

obraz

### DIFF
--- a/solidity/dashboard/src/components/coverage-pools/ExchangeRate.jsx
+++ b/solidity/dashboard/src/components/coverage-pools/ExchangeRate.jsx
@@ -1,0 +1,42 @@
+import React from "react"
+import { Keep } from "../../contracts"
+import TokenAmount from "../TokenAmount"
+
+const BaseExchangeRate = ({
+  covToken,
+  amount,
+  htmlTag = "div",
+  className = "",
+}) => {
+  const Tag = htmlTag
+  return (
+    <Tag className={className}>
+      {`1 ${covToken.symbol}`} = ~
+      <TokenAmount
+        amount={amount}
+        decimalsToDisplay={3}
+        wrapperClassName="flex-inline"
+        amountClassName=""
+        symbolClassName=""
+      />
+    </Tag>
+  )
+}
+
+export const CoveragePoolV1ExchangeRate = ({
+  covToken,
+  collateralToken,
+  covTotalSupply,
+  totalValueLocked,
+  ...restProps
+}) => (
+  <BaseExchangeRate
+    {...restProps}
+    covToken={covToken}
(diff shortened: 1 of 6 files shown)


solidity/dashboard/src/components/coverage-pools/ExchangeRate.jsx
```diff
@@ -0,0 +1,42 @@
...
+  covToken,
+  collateralToken,
+  covTotalSupply,
+  totalValueLocked,
+  ...restProps
+}) => (
+  <BaseExchangeRate
+    {...restProps}
+    covToken={covToken}
+    amount={Keep.coveragePoolV1.estimatedBalanceFor(
```
michalsmiarowski, line 36: Can we round the value returned from `estimatedBalanceFor`? 

The problem with the function is that it can return value that has more than 18 decimal places. For example:

1.000000000000000000000000000345

Because of that `TokenAmount` will print such numer as 1.000 (1.000000000000000000 in the tooltip)

We should cut out any decimals that are less than the smallest KEEP unit when calculating the estimatedBalance. So i would add `decimalPlaces` to the function:

```
return new BigNumber(covKeep)
      .div(totalSupplyOfCovKeep)
      .multipliedBy(tvl)
      .decimalPlaces(KEEP.smallestPrecisionDecimals)
      .toString()
```

solidity/dashboard/src/components/coverage-pools/ExchangeRate.jsx
```diff
@@ -0,0 +1,42 @@
...
+      collateralToken.fromTokenUnit(1).toString(),
```
michalsmiarowski, line 37: Do we actually need `collateralToken` prop here? We are using `KEEP.coveragePoolV1.estimateBalanceFor` one line above.

So either change above line to `collateralToken.coveragePoolV1.estimateBalanceFor` or remove the `collateralToken` prop from `CoveragePoolV1ExchangeRate` and keep it as `KEEP.fromTokenUnit(1).toString()`

michalsmiarowski (changes requested):
Left some comments. Let's fix them and shipt it :shipit:

solidity/dashboard/src/components/coverage-pools/ExchangeRate.jsx
r-czajkowski: The main idea is to be independent of KEEP token because coverage pools can work with any ERC20 token. I've updated a bit the `estimateBalanceFor` method and left a TODO. Please see c49944d.

michalsmiarowski (approved):
LGTM 🚀

PR #2613, opened 2021-09-23
Governable random beacon parameters

tomaszslabon:
#Depends on https://github.com/keep-network/keep-core/pull/2612
This PR adds a contract containing governable parameters that will be needed by Random Beacon v2.
The parameters are taken from [RFC-19](https://github.com/keep-network/keep-core/pull/2585/files#).

Question: How should the parameters be initialised? One option is to initialize them via a contractor (at least those parameters that take long to initialize).  

The unit tests were executed locally and pass, but they do not execute on CI (other PRs should enable this).
The solidity and typescript code were analysed with tools like `solhint`,`prettier`, `eslint` for typescript (configuration for those mechanisms should probably be added by other PRs).

PR #2632, opened 2021-09-30
Random Beacon Governable Parameters v2

nkuba:
This is an alternative implementation of Governable Parameters that uses storing parameters directly in the storage based on the slot position calculated from the parameter name and a prefix.

See original PR: https://github.com/keep-network/keep-core/pull/2613

This implementation requires lower gas for the contract deployment:
```
·----------------------------|---------------------------|-------------|-----------------------------·
|    Solc version: 0.8.6     ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 30000000 gas  │
·····························|···························|·············|······························
|  Methods                                                                                           │
················|············|·············|·············|·············|···············|··············
|  Contract     ·  Method    ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
················|············|·············|·············|·············|···············|··············
|  Deployments               ·                                         ·  % of limit   ·             │
·····························|·············|·············|·············|···············|··············
|  GovernableParameters      ·          -  ·          -  ·    1857858  ·        6.2 %  ·          -  │
·····························|·············|·············|·············|···············|··············
|  RandomBeaconParameters    ·          -  ·          -  ·    1003311  ·        3.3 %  ·          -  │
·····························|·············|·············|·············|···············|··············
|  RandomBeaconParametersV2  ·          -  ·          -  ·     575090  ·        1.9 %  ·          -  │
·----------------------------|-------------|-------------|-------------|---------------|-------------·
```

Unfortunately, gas estimate of reading the parameter is higher than in the original implementation:
```
estimated gas for V1: 23474
estimated gas for V2: 25246
```


solidity/random-beacon/contracts/RandomBeaconParametersV2.sol
```diff
@@ -0,0 +1,112 @@
+pragma solidity ^0.8.6;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+contract RandomBeaconParametersV2 is Ownable {
```
pdyraga, line 5: This contract could be a good replacement for `RandomBeaconGovernance` introduced in #2613 to reduce the boilerplate code. I see two problems though: 
- we have two governance delays; some parameters are secured with 24h and some others with 2 weeks governance delay,
- some parameters are validated at the moment the change is requested.

Is it something we can have in this version without introducing significant complexity? I take the delay is doable but validation could be problematic 🤔

pdyraga: Actually, if we were doing no validation and used the same delay for all parameters, we'd be able to eliminate `RandomBeaconGovernance` altogether and use [time lock controller from OpenZeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/TimelockController.sol) but I do not think we have this comfort, given [this](https://github.com/keep-network/keep-core/pull/2613#issuecomment-927813679) conversation.

PR #2633, opened 2021-10-04
Fix TBTCv2/Saddle pool

r-czajkowski:
This PR fixes staking/withdrawing to/from the TBTCv2/Saddle pool. It also fixes a bug that dapp displays data from the new pool in the TBTCv1/Saddle card- both the new and the retired Saddle farms show the same rewards even though a user only has funds in the new one.


solidity/dashboard/src/pages/liquidity/LiquidityPage.jsx
```diff
@@ -117,36 +117,22 @@ const LiquidityPage = ({ headerTitle }) => {
...
   }, [isConnected, keepTokenBalance.value, hideBanner, isBannerVisible])
 
-  const addLpTokens = (
-    wrappedTokenBalance,
-    liquidityPairContractName,
-    pool,
-    awaitingPromise
-  ) => {
+  const addLpTokens = (poolId, wrappedTokenBalance, awaitingPromise) => {
+    const liquidityPairData = LIQUIDITY_REWARD_PAIRS[poolId]
```
michalsmiarowski, line 121: Although I really like the name of this const I think we should be consistent in the app. I think we usualy named it as `liquidityRewardPair`.

So either change this to `liquidityPairData` -> `liquidityRewardPair`, or change all occurrences of `liquidityRewardPair` -> `liquidityPairData`.

I like `liquidityPariData` name more, but the first case is faster to fix. The `liquidityPairData` is also reasonable variable name since it is singular from of `LIQUIDITY_REWARD_PAIRS`. I'll left it to you to decide.

michalsmiarowski (changes requested):
Looking great! Tested locally and it works fantastic.

Left one comment about variable naming convention.

solidity/dashboard/src/pages/liquidity/LiquidityPage.jsx
r-czajkowski: e34c63a

michalsmiarowski (approved):
👍

PR #2634, opened 2021-10-04
[Backport] Fix TBTCv2/Saddle pool

r-czajkowski:
Backport of: #2633 

This PR fixes staking/withdrawing to/from the TBTCv2/Saddle pool. It also fixes a bug that dapp displays data from the new pool in the TBTCv1/Saddle card- both the new and the retired Saddle farms show the same rewards even though a user only has funds in the new one.


michalsmiarowski (approved):
🐎

PR #2630, opened 2021-09-30
Random Beacon v2 DKG

nkuba:

PR #2637, opened 2021-10-04
Random Beacon v2 relay request and relay entry 

lukasz-zimnoch:
Depends on: #2630 

Early draft. Work in progress...


solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -0,0 +1,69 @@
...
+
+// TODO: Documentation
+// TODO: Unit tests
+library Relay {
+    struct Request {
+        uint256 id;
+        uint256 startBlock;
+        uint256 groupIndex;
+        bytes previousEntry;
+    }
```
pdyraga, line 12: Regardless if we decide to use a library for `Relay`, we need to revisit types used in this struct. `uint256` for `groupIndex` feels too much, I'd also argue `id` and `startBlock` can probably be reduced, see https://fravoll.github.io/solidity-patterns/tight_variable_packing.html.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -93,9 +94,23 @@ contract RandomBeacon {
...
+        return 1;
+    }
+
+    function requestRelayEntry(bytes calldata previousEntry) external {
+        // TODO: Require fee paid to the maintenance pool.
+        // TODO: Is nonReentrant needed?
+        relay.requestEntry(selectGroup(), previousEntry);
+    }
+
+    function submitRelayEntry(bytes calldata entry) external {
```
pdyraga, line 108: I am not against using the library to better organize the code but we must ensure this function is as cheap as only possible given that we do not reimburse for submitting relay entry and there is no special reward for that. Let's keep it in mind, especially when we'll start working on more interactions (DKG, timeouts, slashing).

solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -0,0 +1,69 @@
...
+        self.requestCount++;
+
+        self.currentRequest = Request(
+            self.requestCount,
+            block.number,
+            groupIndex,
+            previousEntry
+        );
+
+        // TODO: Emit event.
```
pdyraga, line 38: Bear in mind events will have to be re-declared in the contract.

solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    function submitEntry(
+        Data storage self,
+        bytes calldata entry
+    ) internal {
+        require(isRequestInProgress(self), "No relay request in progress");
+        require(!isRequestTimedOut(self), "Relay request timed out");
+
+        // TODO: Verify entry and submitter eligibility.
```
pdyraga, line 48: We will need to have access to group data and to the slashing logic. This part can get unwieldy from the library. I would probably start with having this code in the contract but if you have an idea how to set up the interactions and make it gas-effective, that's fine.

solidity/random-beacon/contracts/libraries/Relay.sol
lukasz-zimnoch: Yep, I'm aware and I'm thinking about emitting events just from the contract and not from the library at all. WDYT?

solidity/random-beacon/contracts/libraries/Relay.sol
lukasz-zimnoch: Yeah, this is still an open topic. I Will play with that and we'll see.

PR #2639, opened 2021-10-05
Bump the dashboard version to v1.16.0-pre

r-czajkowski:
Ref: #2626 

The v1.15.1 has been relesed.


infrastructure/kube/keep-prd/keep-dapp-token-dashboard-deployment.yaml
```diff
@@ -21,6 +21,6 @@ spec:
     spec:
       containers:
       - name: keep-dapp-token-dashboard
-        image: keepnetwork/token-dashboard:v1.14.0
+        image: keepnetwork/token-dashboard:v1.15.1
```
michalinacienciala, line 24:
```suggestion
        image: keepnetwork/token-dashboard:v1.15.2
```
I see that [the latest dashboard version on Mainnet is `v1.15.2`](https://github.com/keep-network/keep-core/releases), we can update this PR to reflect that.

michalinacienciala (changes requested):

PR #2640, opened 2021-10-05
Fix KEEP-only pool

r-czajkowski:
In the 09a87bc7ddfdeb70bcd64574048999bda31379d7 commit we didn't take
into account that KEEP-only pool component uses the same functions to
deposit/withdrawal. This commit will adjust these functions to the
changes made in the above commit.


michalsmiarowski (approved):
Looking good! Let's ship it 🚀

PR #2641, opened 2021-10-05
[Backport] Fix KEEP-only pool

r-czajkowski:
Backport of: #2640 

In the 09a87bc commit we didn't take
into account that KEEP-only pool component uses the same functions to
deposit/withdrawal. This commit will adjust these functions to the
changes made in the above commit.


michalsmiarowski (approved):
👍

PR #2642, opened 2021-10-05
Workflows for beacon v2

dimpar:
- Adding basic workflow setup for beacon v2 which includes: building, testing, linting, slithering, dry-run deploy, deployment on testnent and publishing a package to npm.
- Some minor Readme changes around new dev structure

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -1,0 +1,200 @@
+name: Solidity Random Beacon
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  push:
+    branches:
+      - main
+    paths:
+      - "solidity/random-beacon/**"
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: 'Environment for workflow execution'
+        required: false
+        default: 'dev'
+      upstream_builds:
+        description: 'Upstream builds'
+        required: false
+      upstream_ref:
+        description: 'Git reference to checkout (e.g. branch name)'
+        required: false
+        default: 'main'
+
+jobs:
+  contracts-detect-changes:
+    runs-on: ubuntu-latest
+    outputs:
+      path-filter: ${{ steps.filter.outputs.path-filter }}
+    steps:
+      - uses: actions/checkout@v2
+        if: github.event_name == 'pull_request'
+
+      - uses: dorny/paths-filter@v2
+        if: github.event_name == 'pull_request'
+        id: filter
+        with:
+          filters: |
+            path-filter:
+              - './solidity/random-beacon/**'
+
+  contracts-lint:
+    needs: contracts-detect-changes
+    if: |
+      github.event_name == 'push'
+        || needs.contracts-detect-changes.outputs.path-filter == 'true'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./solidity/random-beacon
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "14.x"
+
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Lint
+        run: yarn lint
+  
+  contracts-slither:
+    needs: contracts-detect-changes
+    if: |
+      github.event_name == 'push'
+        || needs.contracts-detect-changes.outputs.path-filter == 'true'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./solidity/random-beacon
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "14.x"
+          cache: "yarn"
+          cache-dependency-path: solidity/random-beacon/yarn.lock
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8.5
+
+      - name: Install Solidity
+        env:
+          SOLC_VERSION: 0.8.6 # according to solidity.version in hardhat.config.js
+        run: |
+          pip3 install solc-select
+          solc-select install $SOLC_VERSION
+          solc-select use $SOLC_VERSION
+
+      - name: Install Slither
+        env:
+          SLITHER_VERSION: 0.8.0
+        run: pip3 install slither-analyzer==$SLITHER_VERSION
+
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Run Slither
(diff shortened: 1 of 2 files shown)


.github/workflows/contracts.yml
```diff
@@ -0,0 +1,200 @@
...
+      - name: Install Slither
+        env:
+          SLITHER_VERSION: 0.8.0
+        run: pip3 install slither-analyzer==$SLITHER_VERSION
+
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Run Slither
+        run: slither --hardhat-artifacts-directory build .
```
michalinacienciala, line 104: I see that the command is [the same as in `tbtc-v2` project](https://github.com/keep-network/tbtc-v2/blob/44b78bc6c63caa9736509999501a1c182a3650d5/.github/workflows/contracts.yml#L237) and that in [the workflow for `coverage-pools`](https://github.com/keep-network/coverage-pools/blob/3ac340960dbb3729da3285fa8142cdf3bdaa413e/.github/workflows/contracts.yml#L232) we don't pass the `--hardhat-artifacts-directory` param (instead we specify `hardhat_artifacts_directory` option in [`slither.config.json`](https://github.com/keep-network/coverage-pools/blob/3ac340960dbb3729da3285fa8142cdf3bdaa413e/slither.config.json#L4)). Are both those approaches valid, or is the particular approach dictated by some maters?

Also, I wonder if we don't need to add the `slither.config.json` file with `detectors_to_exclude` and `filter_paths` options? I can see this configuration in both `coverage-pools` and `tbtc-v2`.

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,200 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./solidity/random-beacon
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "14.x"
```
michalinacienciala, line 120: We can add caching of dependencies, to speed up execution:
```suggestion
          node-version: "14.x"
          cache: "yarn"
          cache-dependency-path: solidity/random-beacon/yarn.lock
```

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,200 @@
...
+      - name: Load environment variables
+        uses: keep-network/ci/actions/load-env-variables@v1
+        with:
+          environment: ${{ github.event.inputs.environment }}
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "14.x"
+          cache: "yarn"
+          cache-dependency-path: solidity/random-beacon/yarn.lock
```
michalinacienciala, line 173:
```suggestion
          cache-dependency-path: solidity/random-beacon/yarn.lock
          registry-url: "https://registry.npmjs.org"
```

If we're using `NODE_AUTH_TOKEN` env variable in the `Publish to npm` step, we need to specify the `registry-url` property in the step setting up the node.
Otherwise publication of the package will fail with error like this:
```
 '@keep-network/random-beacon@0.0.1-ropsten.0' is not in the npm registry
```

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,200 @@
...
+    defaults:
+      run:
+        working-directory: ./solidity/random-beacon
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Load environment variables
+        uses: keep-network/ci/actions/load-env-variables@v1
+        with:
+          environment: ${{ github.event.inputs.environment }}
```
michalinacienciala, line 167: We can get rid of this step (I know it's present in workflows for `tbtc-v2` and `coverage-pools`, but when I look at them now, I think they're redundant there as well).

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,200 @@
...
+
+      - uses: dorny/paths-filter@v2
+        if: github.event_name == 'pull_request'
+        id: filter
+        with:
+          filters: |
+            path-filter:
+              - './solidity/random-beacon/**'
+
+  contracts-lint:
```
michalinacienciala, line 43: Could we also add job for running prettier? Or combine lint & prettier into `format`?

michalinacienciala, line 43: Could we also add job for running prettier? Or combine lint & prettier into `format`?

michalinacienciala (changes requested):
I haven't finished the whole review, but I'm posting what I have found so far. Tomorrow I will need to continue and take a look at README. I also need to investigate the reason behind checks not executing (probably something about clash of the job names between different workflows).

.github/workflows/contracts.yml
dimpar: same as https://github.com/keep-network/keep-core/pull/2642#discussion_r723089328

.github/workflows/contracts.yml
dimpar: Sure, done in ea4b4d5

.github/workflows/contracts.yml
dimpar: I think both approaches are fine, it just depends on how many flags we want to pass. If it was just one flag, I guess the inline approach is fine.  But `detectors_to_exclude` also might be handy, so I moved params to `slither.config.json`. Good point.

.github/workflows/contracts.yml
dimpar: Removed ea4b4d5

README.adoc
```diff
@@ -76,32 +76,36 @@ keep-core/
   Dockerfile
   main.go, *.go
   docs/
-  solidity-v1/ <1>
-  cmd/ <2>
-  pkg/ <3>
+  solidity <1>
```
michalinacienciala, line 79:
```suggestion
  solidity/ <1>
```

.github/workflows/contracts.yml
michalinacienciala: I can't see this change here, I think it was added accidentally to `contracts-lint` node setup step

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,200 @@
...
+    defaults:
+      run:
+        working-directory: ./solidity/random-beacon
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "14.x"
+          registry-url: "https://registry.npmjs.org"
```
michalinacienciala, line 58: This line in not necessary (although it doesn't do any harm).

PR #2643, opened 2021-10-06
`ExchangeRate` component updates

r-czajkowski:
This PR updates the `ExchangeRate` component based on the users' usability studies.

Main changes:
* always keep 3 decimal places in exchange rate no matter what (so for example `1.000`; `1.100`; `1.110`; `1.111`)
* remove tilda completely,


michalsmiarowski (approved):
Looking good! Tested locally and it was working perfectly! :shipit:

PR #2618, opened 2021-09-27
Code analysis and formatting

dimpar:
Depends on https://github.com/keep-network/keep-core/pull/2612

Adding code formating tools and static code analyzers.

Todo:
- [ ] add a linter for `.ts` files 
- [ ] setup GH workflows for these tools

PR #2655, opened 2021-10-13
Fixing linting problems

dimpar:
This PR fixes existing linting/formatting issues in:
- [x] linting `.ts` files
- [x] linting `.json` files
- [x] linting `.sol` files

Depends on: https://github.com/keep-network/keep-core/pull/2618


pdyraga:
Should we `.git-blame-ignore-revs` the changes from linter/prettier?

dimpar:
Slithering work will be done in a separate PR. Turns out we have more work around that area, since beaconV1 didn't run through `slither` and now we have a bunch of `slither` issues when moving code from `beaconV1` to `beacon`.

dimpar:
> Should we `.git-blame-ignore-revs` the changes from linter/prettier?

Done in 0606146

PR #2671, opened 2021-10-26
Random Beacon v2 banning inactive operators from sortition pool rewards

lukasz-zimnoch:
Depends on: #2664 
Depends on: https://github.com/keep-network/sortition-pools/pull/125

This changeset provides the logic for banning inactive operators from sortition pool rewards. Each operator makes a gas deposit upon pool registration and that deposit is released if the given operator needs to be banned from the rewards during entry submission thus making the submission call cheaper gas-wise.

By the way, this PR leverages https://github.com/keep-network/keep-core/pull/2684 and https://github.com/keep-network/sortition-pools/pull/131 and switch relay entry unit tests `RandomBeacon.Relay.test.ts` to use a real sortition pool. That change allows making gas usage estimates which are closer to the real ones.


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -548,6 +589,39 @@ contract RandomBeacon is Ownable {
...
+            /* solhint-disable not-rely-on-time */
+            // slither-disable-next-line reentrancy-benign
+            punishedOperators[operator] = block.timestamp + punishmentDuration;
+            /* solhint-enable not-rely-on-time */
+
+            // Perform operations on sortition pool only in case the operator
+            // is registered. Otherwise, the punishment set above will prevent
+            // the operator to join the pool in future.
+            if (sortitionPool.isOperatorInPool(operator)) {
+                gasStation.releaseGas(operator);
```
dimpar, line 618: How do we know that the total gas refund of `gasStation.releaseGas(operator)` covers used gas in `sortitionPool.removeOperator(operator)`?

solidity/random-beacon/contracts/libraries/GasStation.sol
```diff
@@ -0,0 +1,57 @@
...
+    /// @notice Sets the memory slots tied to the given address to the value
+    ///         passed as function argument.
+    /// @param addr Address which the deposit is tied to.
+    /// @param value Value which should be inserted to the memory slots.
+    function setDeposit(
+        Data storage self,
+        address addr,
+        uint256 value
+    ) internal {
+        for (uint256 i = 0; i < depositSize; i++) {
```
dimpar, line 53: `depositSize` is set to 1 which makes this for loop unnecessary. But I take that `depositSize = 1` can be changed later and that's why you added a for loop? Do you think this param should be modifiable in case gas increases for operators removal from the sortition pool?

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -187,6 +198,7 @@ contract RandomBeacon is Ownable {
         IERC20 _tToken,
         IStaking _staking
     ) {
+        // TODO: RandomBeacon must be the owner of the sortition pool.
```
dimpar, line 201: I take that the sortition pool is a 1:1 relation with the 'project' that uses it? ie. `random-beacon`<->`sortition pool`, `ecdsa`<->`different sortion pool`?

dimpar (commented):
Reviewed contracts only. Will look at tests a bit later

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -233,12 +237,49 @@ describe("RandomBeacon - Relay", () => {
...
-                    await expect(tx)
-                      .to.emit(sortitionPool, "OperatorsRemoved")
-                      .withArgs([member16.address, member17.address])
+                    // Two members should be kicked out from the pool.
+                    expect(await sortitionPool.operatorsCount()).to.be.equal(
+                      constants.groupSize - 2
+                    )
+
+                    // Member 16 should be kicked out from the pool, blocked
+                    // against re-joining for 2 weeks, and his gas deposit
```
dimpar, line 255: > (...) blocked against re-joining for 2 weeks (...)

Is there an easy way to verify that? Can we test when this member tries to be reinserted back to the sortition pool before 2 weeks pass?

lukasz-zimnoch: It is already tested. See tests of `registerOperator` method.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -519,12 +558,14 @@ contract RandomBeacon is Ownable {
...
         external
     {
-        relay.submitEntry(
+        address[] memory punishedMembers = relay.submitEntry(
             submitterIndex,
             entry,
             groups.getGroup(relay.currentRequest.groupId)
         );
 
+        punishOperators(punishedMembers, 2 weeks);
```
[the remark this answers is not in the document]
lukasz-zimnoch: The RFC did not mention about making it governable 🤷 I think that's a good question for @pdyraga

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -548,6 +589,39 @@ contract RandomBeacon is Ownable {
...
+            punishedOperators[operator] = block.timestamp + punishmentDuration;
+            /* solhint-enable not-rely-on-time */
+
+            // Perform operations on sortition pool only in case the operator
+            // is registered. Otherwise, the punishment set above will prevent
+            // the operator to join the pool in future.
+            if (sortitionPool.isOperatorInPool(operator)) {
+                gasStation.releaseGas(operator);
+                // TODO: Is it possible to kick out operator when pool is locked?
+                sortitionPool.removeOperator(operator);
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, after spending some time in the sortition pool code I think we can theoretically block the entry submission if we try to kick out an operator who has already left the sortition pool. Your proposition sounds good. The only thing to consider is the fact that we plan to use `sortitionPool.removeOperators` to remove multiple operators in one call. In that case, if we surround that call with try-catch and we fail during the removal of one operator, the rest of operators avoid the punishment in that case. Maybe we should handle such a case at the sortition pool level? Curious about your thoughts @dimpar and @pdyraga.

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Well, we don't know at this point. I'm also not sure we should care if it covers the whole gas used in `sortitionPool.removeOperator(operator)`. The goal of the gas station pattern used here is to make `submitRelayEntry` call as cheap as possible for the submitter, especially in case that transaction needs to remove multiple group members from the sortition pool by the way. Currently, we reserve one slot of memory during operator registration. Once we have a real sortition pool in tests (or even better - system tests) we will be able to assess gas consumption and try to optimize by, for instance, reserving more slots upon operator registration. At the same time, we will need to make sure operator registration doesn't become too expensive due to the gas deposit.

solidity/random-beacon/contracts/libraries/GasStation.sol
lukasz-zimnoch: That's the original code from the gas station we used to have in the sortition pool (removed recently). I guess we may be interested with reserving more than one slot (see https://github.com/keep-network/keep-core/pull/2671#discussion_r741280350) so this is why this loop is here. Ideally, we would make this param governable but it's not that simple because we would need to handle existing deposits that would still use the old slot value.

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Yep, see https://github.com/keep-network/sortition-pools/pull/125 to get more context.

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
dimpar: Or we can try to capture the "id" of that failed operator and retry starting from the next id? Although it might be too much and moving that check to the sortition pool might be a solution.

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: Ok, thx for the explanation. Can we add a comment somewhere that this will need to be adjusted? Maybe inside the `GasStation` contract?

solidity/random-beacon/contracts/libraries/GasStation.sol
dimpar: Yea, makes sense, I would just add a comment saying that this one slot should be adjusted based on the real sortition pool or system test.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -562,6 +608,42 @@ contract RandomBeacon is Ownable {
...
+
+            // Perform operations on sortition pool only in case the operator
+            // is registered. Otherwise, the punishment set above will prevent
+            // the operator to join the pool in future.
+            if (sortitionPool.isOperatorInPool(operator)) {
+                gasStation.releaseGas(operator);
+            }
+        }
+
+        // TODO: Is it possible to kick out operator when pool is locked?
```
pdyraga, line 642: *bonk* 😬

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
lukasz-zimnoch: I think the recent changes introduced in `punishOperators` method solve the problem we discuss here. Only actual pool members are passed to `sortitionPool.removeOperators` so there is no risk we would try to remove an operator who already left the pool and we'll revert in the result. Just keep in mind the current implementation of `punishOperators` can be suboptimal and we will be able (probably) to make it cheaper in the near future once all pending changes developed in the sortition pool codebase are ready.

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/210861513bf978a98bd737e3b90a27ca838f0d02

solidity/random-beacon/contracts/libraries/GasStation.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/210861513bf978a98bd737e3b90a27ca838f0d02

pdyraga:
What are the gas costs? Ideally, can you provide the gas costs with/without gas station for optimistic and pessimistic cases?

lukasz-zimnoch:
> What are the gas costs? Ideally, can you provide the gas costs with/without gas station for optimistic and pessimistic cases?

Those are the results with gas station.
```
·--------------------------------------------------------------------------------------|---------------------------|-------------|-----------------------------·
|                                 Solc version: 0.8.6                                  ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 30000000 gas  │
·······················································································|···························|·············|······························
|  Methods                                                                                                                                                     │
···························|···························································|·············|·············|·············|···············|··············
|  Contract                ·  Method                                                   ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
···························|···························································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  registerOperator                                         ·     135060  ·     343813  ·     170743  ·          204  .          -  |
...........................|...........................................................|.............|.............|.............|...............|..............
|  RandomBeaconStub        ·  requestRelayEntry                                        ·      81338  ·     137678  ·     118811  ·           42  ·          -  │
···························|···························································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  submitRelayEntry                                         ·     252440  ·    4296639  ·    1436658  ·           41  ·          -  │
···························|···························································|·············|·············|·············|···············|··············
```


Regarding `submitRelayEntry` the optimistic case (no slashing, no removing from pool) is around `250k`. The worst case (slashing and removing 63 members from pool) is about `4.3M`... We can reserve gas more slots during `registerOperator` but I'm not sure it will help and `registerOperator` is already quite expensive. Seems like the only way to optimize here is waiting for the moment when soritition pool API will allow to remove operator from rewards and not from the pool itself.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -562,6 +647,59 @@ contract RandomBeacon is Ownable {
...
+    ///       are actually in the pool to avoid revert. Two loops are needed
+    ///       because Solidity doesn't allow for memory arrays and mappings.
+    function punishOperators(uint32[] memory ids, uint256 punishmentDuration)
+        internal
+    {
+        address[] memory operators = sortitionPool.getIDOperators(ids);
+        uint256 operatorsInPoolCount = 0;
+
+        for (uint256 i = 0; i < operators.length; i++) {
+            if (sortitionPool.isOperatorInPool(operators[i])) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: We are not interested about the total operators in the pool count. We are interested in which operators from our list are actually in the pool. It's important to remove only those operators to avoid reverting upon removal.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -562,6 +647,59 @@ contract RandomBeacon is Ownable {
...
+                operatorsInPoolCount++;
+            }
+        }
+
+        uint32[] memory operatorsInPoolIDs = new uint32[](operatorsInPoolCount);
```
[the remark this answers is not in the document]
lukasz-zimnoch: Most likely we'll get rid of the `removeOperators` soon. Removing operators is very expensive and will be probably switched to making those operators ineligible for rewards instead. The change will be done in a follow-up PR once the sortition pool supports that.

pdyraga:
> > What are the gas costs? Ideally, can you provide the gas costs with/without gas station for optimistic and pessimistic cases?
> 
> Those are the results with gas station.
> 
> ```
> ·--------------------------------------------------------------------------------------|---------------------------|-------------|-----------------------------·
> |                                 Solc version: 0.8.6                                  ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 30000000 gas  │
> ·······················································································|···························|·············|······························
> |  Methods                                                                                                                                                     │
> ···························|···························································|·············|·············|·············|···············|··············
> |  Contract                ·  Method                                                   ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
> ···························|···························································|·············|·············|·············|···············|··············
> |  RandomBeaconStub        ·  registerOperator                                         ·     135060  ·     343813  ·     170743  ·          204  .          -  |
> ...........................|...........................................................|.............|.............|.............|...............|..............
> |  RandomBeaconStub        ·  requestRelayEntry                                        ·      81338  ·     137678  ·     118811  ·           42  ·          -  │
> ···························|···························································|·············|·············|·············|···············|··············
> |  RandomBeaconStub        ·  submitRelayEntry                                         ·     252440  ·    4296639  ·    1436658  ·           41  ·          -  │
> ···························|···························································|·············|·············|·············|···············|··············
> ```
> 
> Regarding `submitRelayEntry` the optimistic case (no slashing, no removing from pool) is around `250k`. The worst case (slashing and removing 63 members from pool) is about `4.3M`... We can reserve more gas slots during `registerOperator` but I'm not sure it will help and `registerOperator` is already quite expensive. Seems like the only way to optimize here is waiting for the moment when sortition pool API will allow to remove the operator from rewards and not from the pool itself.

250k is still a lot given there is no reimbursement for this function. Do you know what is the cost of just BLS verification + writing necessary data compared to the cost of other checks (eligibility, timeout)?

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
dimpar: > We are not interested about the total operators in the pool count. We are interested in which operators from our list are actually in the pool.

yes, what I meant is to pass an array of `operators` and make the sortition pool do the counting instead of calling a contract for every operator on our side. But if the plan is to get rid of `removeOperators` then I guess this comment won't be relevant.

solidity/random-beacon/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "2.0.0-dev",
   "description": "Keep Random Beacon",
   "scripts": {
+    "clean": "rm -rf build && rm -rf cache && rm -rf typechain",
```
tomaszslabon, line 6: `hardhat clean` seems to do exactly that, so maybe:
`"clean": "hardhat clean"` ?

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -371,14 +401,44 @@ contract RandomBeacon is Ownable {
...
+        sortitionPool.insertOperator(operator);
+    }
+
+    /// @notice Updates the sortition pool status of the caller.
+    function updateOperatorStatus() external {
+        sortitionPool.updateOperatorStatus(
+            sortitionPool.getOperatorID(msg.sender)
+        );
+
+        // If the operator has been removed from the sorition pool during the
```
tomaszslabon, line 437: sorition -> sortition

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -562,6 +647,59 @@ contract RandomBeacon is Ownable {
...
+
+        uint32[] memory operatorsInPoolIDs = new uint32[](operatorsInPoolCount);
+        uint256 j = 0;
+
+        for (uint256 i = 0; i < operators.length; i++) {
+            address operator = operators[i];
+
+            // Set the punishment regardless the operator is actually
+            // registered in the sortition pool. If the operator is not in pool
+            // at the moment, the punishment will prevent they to join during
```
tomaszslabon, line 684: prevent they to join -> prevent them from joining

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -524,12 +584,32 @@ contract RandomBeacon is Ownable {
...
+        uint256 currentRequestId = relay.currentRequest.id;
+
+        Groups.Group memory group = groups.getGroup(
+            relay.currentRequest.groupId
         );
 
+        (uint32[] memory inactiveMembers, uint256 slashingAmount) = relay
+            .submitEntry(submitterIndex, entry, group);
+
+        punishOperators(inactiveMembers, 2 weeks);
```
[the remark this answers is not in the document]
lukasz-zimnoch: I think passing it as a parameter is justified as the calling code decides who it punishes and for how long.

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, I'd hold to making any changes in sortition pools until we have the changes around rewards blocking here.

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/a400d36b149453ad8ffae6a8d739236f50841758

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/a400d36b149453ad8ffae6a8d739236f50841758

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/a400d36b149453ad8ffae6a8d739236f50841758

solidity/random-beacon/package.json
lukasz-zimnoch: Good point! See https://github.com/keep-network/keep-core/pull/2671/commits/aacaaa1e1e096f4c24d77dba19b35bb917aa4d19

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -524,12 +584,32 @@ contract RandomBeacon is Ownable {
...
+
+        if (slashingAmount > 0) {
+            address[] memory groupMembers = sortitionPool.getIDOperators(
```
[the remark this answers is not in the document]
lukasz-zimnoch: That's not the same call. Here we get addresses of **all** group members for slashing. In `punishOperators` we get addresses of **punished** members.

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -67,7 +67,7 @@ describe("RandomBeacon - Relay", () => {
   let signersAddresses: Address[]
```
dimpar, line 67: It seems that we use `member`, `operator` & `signer` across the code to describe the same thing and it might be confusing.  I think it's harder to "merge" `member`&`operator` since ex. sortition pool uses `operator` and in groups lib we use `member`. But can we at least try to replace `signer` with `member`?

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -708,7 +856,7 @@ describe("RandomBeacon - Relay", () => {
   async function assertMembersEligible(members: number[], groupSize: number) {
     for (let i = 0; i < members.length; i++) {
       expect(
-        // eslint-disable-next-line no-await-in-loop,@typescript-eslint/no-unused-expressions
+        // eslint-disable-next-line no-await-in-loop
```
dimpar, line 859: I think it can be moved at the top of the file do disable for all for-loops

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
dimpar: Makes sense, but it's a bit itchy because a) as for now `punishOperators(inactiveMembers, 2 weeks);` is being triggered only from 1 place, b) isn't `2 weeks` a const and even if we call `punishOperators(inactiveMembers, 2 weeks);` from a different place in our code it would be the same value?

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
dimpar: ah, I see, you're right

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
lukasz-zimnoch: This function will be called from DKG as well. Not sure about the value though. If it turns out to be the same we can always extract to const.

solidity/random-beacon/test/RandomBeacon.Pool.test.ts
```diff
@@ -21,67 +25,154 @@ describe("RandomBeacon - Pool", () => {
...
+    beforeEach(async () => {
+      // Operator is registered and gas deposit is made.
+      await stakingStub.setStake(operator.address, constants.minimumStake)
+      await randomBeacon.connect(operator).registerOperator()
+      operatorID = await sortitionPool.getOperatorID(operator.address)
+
+      // Simulate the operator became ineligible.
+      await stakingStub.setStake(operator.address, 0)
+
+      tx = await randomBeacon.connect(operator).updateOperatorStatus()
```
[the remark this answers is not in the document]
lukasz-zimnoch: See  https://github.com/keep-network/keep-core/pull/2671/commits/fecc531e0fd1fc57f38758f2df95b261b9d496ba

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
lukasz-zimnoch: https://github.com/keep-network/keep-core/pull/2671/commits/2291bf7a4f9e8c987eee300f1d9302c243d790bc

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
lukasz-zimnoch: Yeah, I don't like it either but there is no ideal solution here. Even `member` instead `signer` is not justified in all contexts because an operator becomes a member after group creation. But, your proposition sounds good for this use case and I changed accordingly: https://github.com/keep-network/keep-core/pull/2671/commits/d70c7ef883c278ed71d22275b8ec0b2e27bdf845

lukasz-zimnoch:
> > > What are the gas costs? Ideally, can you provide the gas costs with/without gas station for optimistic and pessimistic cases?
> > 
> > 
> > Those are the results with gas station.
> > ```
> > ·--------------------------------------------------------------------------------------|---------------------------|-------------|-----------------------------·
> > |                                 Solc version: 0.8.6                                  ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 30000000 gas  │
> > ·······················································································|···························|·············|······························
> > |  Methods                                                                                                                                                     │
> > ···························|···························································|·············|·············|·············|···············|··············
> > |  Contract                ·  Method                                                   ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
> > ···························|···························································|·············|·············|·············|···············|··············
> > |  RandomBeaconStub        ·  registerOperator                                         ·     135060  ·     343813  ·     170743  ·          204  .          -  |
> > ...........................|...........................................................|.............|.............|.............|...............|..............
> > |  RandomBeaconStub        ·  requestRelayEntry                                        ·      81338  ·     137678  ·     118811  ·           42  ·          -  │
> > ···························|···························································|·············|·············|·············|···············|··············
> > |  RandomBeaconStub        ·  submitRelayEntry                                         ·     252440  ·    4296639  ·    1436658  ·           41  ·          -  │
> > ···························|···························································|·············|·············|·············|···············|··············
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Regarding `submitRelayEntry` the optimistic case (no slashing, no removing from pool) is around `250k`. The worst case (slashing and removing 63 members from pool) is about `4.3M`... We can reserve more gas slots during `registerOperator` but I'm not sure it will help and `registerOperator` is already quite expensive. Seems like the only way to optimize here is waiting for the moment when sortition pool API will allow to remove the operator from rewards and not from the pool itself.
> 
> 250k is still a lot given there is no reimbursement for this function. Do you know what is the cost of just BLS verification + writing necessary data compared to the cost of other checks (eligibility, timeout)?

The cost of BLS verification + writing necessary data + check if entry in progress and there is no timeout is `227908`.

lukasz-zimnoch:
The most expensive parts of `submitRelayEntry` are BLS verification which is about `130k` and getting group details from storage which is around `60k`.

dimpar:
No more comments from my side. Curious about sortition pool changes that are referenced in the 2 comments and leaving them open for now.

solidity/random-beacon/contracts/libraries/GasStation.sol
```diff
@@ -0,0 +1,62 @@
...
+// ▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+//
+//                           Trust math, not hardware.
+
+pragma solidity ^0.8.6;
+
+library GasStation {
+    struct Data {
+        // Map holding info about all deposits: address => (slot => value).
+        mapping(address => mapping(uint256 => uint256)) gasDeposits;
```
pdyraga, line 20: Probably not something for this PR and we need to see how things will work in `sortition-pool/Rewards` contract but this pattern might not be the most efficient one in case we need to iterate and delete.

Another option worth experimenting with would be to just allocate one big slot when operator joins the pool (even in an array) and then if a certain gas cost was exceeded when calling `banRewards` of sortition pool, `pop()` one slot. This approach assumes there are more honest and properly functioning operators joining the pool than operators who needs to be eliminated from rewards. cc @eth-r

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
pdyraga: Should be governable.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -199,6 +229,7 @@ contract RandomBeacon is Ownable {
         IERC20 _tToken,
         IRandomBeaconStaking _staking
     ) {
+        // TODO: RandomBeacon must be the owner of the sortition pool.
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, but we don't have deployment scripts yet so I left it here to not forget about it.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -186,7 +201,22 @@ contract RandomBeacon is Ownable {
...
+        uint256 indexed requestId,
+        uint256 slashingAmount,
+        address[] groupMembers
+    );
+
+    event RelayEntrySubmissionFailureSlashingOccurred(
+        uint256 indexed requestId,
+        uint256 slashingAmount,
+        address[] groupMembers
+    );
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/b4939253e2da5ca48dfc87821f1ad9534e55e32c

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -371,14 +401,44 @@ contract RandomBeacon is Ownable {
...
+
+    /// @notice Updates the sortition pool status of the caller.
+    function updateOperatorStatus() external {
+        sortitionPool.updateOperatorStatus(
+            sortitionPool.getOperatorID(msg.sender)
+        );
+
+        // If the operator has been removed from the sortition pool during the
+        // status update, release its gas deposit.
+        if (!sortitionPool.isOperatorInPool(msg.sender)) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: And this is why that line is here. If the operator was removed during `updateOperatorStatus` we release their gas slot reserved upon their registration. Do you mean we should release the operator's gas slot only during entry submission when we eliminate that operator from rewards?

solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -287,7 +277,7 @@ library Relay {
 
         uint64 currentRequestId = self.currentRequest.id;
 
-        emit RelayEntryTimedOut(currentRequestId);
+        emit RelayEntryTimedOut(currentRequestId, self.currentRequest.groupId);
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/077e3e51cfb3e43e29449e16030fe39a8257b969

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/commit/c641b9a980079e157586ee62dc32e7274af15980

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -541,15 +621,20 @@ contract RandomBeacon is Ownable {
...
 
-        staking.slash(
-            relay.relayEntrySubmissionFailureSlashingAmount,
+        emit RelayEntrySubmissionFailureSlashingOccurred(
+            relay.currentRequest.id,
+            slashingAmount,
             groupMembers
         );
 
+        staking.slash(slashingAmount, groupMembers);
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/ce09ee03883963986e023543d4d50c595b02a7aa

solidity/random-beacon/contracts/libraries/GasStation.sol
lukasz-zimnoch: Ok, I will keep it in mind once we start optimizing gas costs for relay entry.

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Ok, but I will address it in separate PR and just leave a TODO here. I don't want to inflate this PR with the governance setter.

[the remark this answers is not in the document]
lukasz-zimnoch: https://github.com/keep-network/keep-core/commit/fac1bdf8d7286533ef0b5df82431ff3744d608e2

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/223619641ed245e3a41ab5213ddddf74d0fa1a33

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/223619641ed245e3a41ab5213ddddf74d0fa1a33

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -122,6 +128,15 @@ contract RandomBeacon is Ownable {
...
     Callback.Data internal callback;
+    GasStation.Data internal gasStation;
+
+    // Other parameters
+
+    /// @notice List of operators who are prohibited to join the sortition
+    ///         pool within a specific period of time. The key of the mapping
+    ///         is operator address. The value is a timestamp when the
+    ///         punishment ends.
+    mapping(address => uint256) public punishedOperators;
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/223619641ed245e3a41ab5213ddddf74d0fa1a33

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -371,14 +401,44 @@ contract RandomBeacon is Ownable {
...
+        uint256 punishmentDeadline = punishedOperators[operator];
+
+        // slither-disable-next-line incorrect-equality
+        require(
+            /* solhint-disable-next-line not-rely-on-time */
+            punishmentDeadline == 0 || block.timestamp > punishmentDeadline,
+            "Operator has an active punishment"
+        );
+
+        delete punishedOperators[operator];
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2671/commits/223619641ed245e3a41ab5213ddddf74d0fa1a33

lukasz-zimnoch:
@pdyraga @dimpar all comments addressed.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -569,6 +646,31 @@ contract RandomBeacon is Ownable {
...
+            // TODO: Do we need the operator to re-deposit gas once
+            //       current punishment is completed to use it for
+            //       future ones?
+            gasStation.releaseGas(operators[i]);
+        }
+
+        // TODO: Once `banRewards` is implemented on the pool side, make sure
+        //       it does not have an unexpected revert instruction which will
+        //       block entry submission. For example, an operator leaves
+        //       the pool just before it gets banned.
```
pdyraga, line 670: For this reason, I'd wrap the entire block when `punishOperators` and `staking.slash` are called in a try-catch. I don't think we should trust these contracts never revert, especially that staking contract is upgradeable. Not necessarily in this PR though.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -374,14 +403,33 @@ contract RandomBeacon is Ownable {
         );
     }
 
-    /// @notice Registers caller in the sortition pool.
-    function registerMemberCandidate() external {
+    /// @notice Registers the caller in the sortition pool.
+    /// @dev Creates a gas deposit tied to the operator address. The gas
+    ///      deposit is released when the operator is banned for sortition pool
```
pdyraga, line 408: I think it should be _banned from_.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -548,15 +619,21 @@ contract RandomBeacon is Ownable {
...
-        staking.slash(
-            relay.relayEntrySubmissionFailureSlashingAmount,
+        emit RelayEntryTimeoutSlashed(
+            relay.currentRequest.id,
+            slashingAmount,
             groupMembers
         );
 
+        // TODO: Revisit whether the notifier should receive 5% of the slashing amount.
+        staking.seize(slashingAmount, 5, msg.sender, groupMembers);
```
pdyraga, line 635: In T staking contract, this 5 means 5% of the notification reward (long story short: notifiers are rewarded from a separate pool that is funded from slashed tokens, and they are rewarded right away while slashing happens in a separate TX), see [here](https://github.com/threshold-network/solidity-contracts/blob/fa83b98566b7c786a499fc438d97b4dd3eee0293/contracts/staking/TokenStaking.sol#L1172-L1174).

This percentage should be governable and should be between [0,100].

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -569,6 +646,31 @@ contract RandomBeacon is Ownable {
         }
     }
 
+    /// @notice Punishes the given operators by banning their sortition pool rewards.
+    /// @dev By the way, this function releases gas deposits made by operators
+    ///      during their registration. See `registerOperator` function. This
+    ///      action makes punishments cheaper gas-wise.
+    /// @param ids IDs of punished operators.
+    /// @param punishmentDuration Duration of the punishment period in seconds.
+    function punishOperators(uint32[] memory ids, uint256 punishmentDuration)
```
pdyraga, line 655: So maybe just `banFromRewards`?

solidity/random-beacon/contracts/test/SortitionPoolStub.sol
```diff
@@ -2,20 +2,40 @@
 
 pragma solidity ^0.8.6;
 
-import "@keep-network/sortition-pools/contracts/SortitionTree.sol";
+import "./SortitionTreeStub.sol";
 import "../RandomBeacon.sol";
 
 // Stub contract used in tests
-contract SortitionPoolStub is ISortitionPool, SortitionTree {
+contract SortitionPoolStub is ISortitionPool {
```
pdyraga, line 9: I'd really love to eventually get rid of this stub completely. cc @nkuba @dimpar @tomaszslabon

solidity/random-beacon/test/fixtures/index.ts
```diff
@@ -1,18 +1,21 @@
...
 } from "../../typechain"
+import { to1e18 } from "../functions"
 
 export const constants = {
   groupSize: 64,
   groupThreshold: 33,
   signatureThreshold: 48, // groupThreshold + (groupSize - groupThreshold) / 2
   offchainDkgTime: 72, // 5 * (1 + 5) + 2 * (1 + 10) + 20
+  minimumStake: to1e18(100000),
+  poolWeightDivisor: 2000,
```
pdyraga, line 18: I think it's `to1e18(1)`, please confirm with the version currently deployed on mainnet for ECDSA stakers.

solidity/random-beacon/test/RandomBeacon.Pool.test.ts
```diff
@@ -1,18 +1,22 @@
-import { ethers, waffle } from "hardhat"
-import { expect } from "chai"
+/* eslint-disable @typescript-eslint/no-unused-expressions */
 
-import { Signer, Contract } from "ethers"
-import { randomBeaconDeployment } from "./fixtures"
+import { ethers, waffle, helpers } from "hardhat"
+import { expect } from "chai"
+import { ContractTransaction } from "ethers"
```
pdyraga, line 5: Let's cleanup linter warnings from this file.

solidity/random-beacon/test/RandomBeacon.Pool.test.ts
```diff
@@ -21,67 +25,94 @@ describe("RandomBeacon - Pool", () => {
...
+      await randomBeacon.connect(operator).updateOperatorStatus()
+    })
+
+    it("should remove operator from the pool", async () => {
+      expect(await sortitionPool.isOperatorInPool(operator.address)).to.be.false
+    })
+
+    it("should release the gas deposit if operator was removed from pool during the update", async () => {
+      expect(await randomBeacon.hasGasDeposit(operator.address)).to.be.false
+    })
```
pdyraga, line 88: We are missing some `context()` here, `updateOperatorStatus()` does not remove anyone from the pool just because it's called. In this case, the eligible stake went down to 0.

pdyraga (approved):
Left a bunch of comments but this is good enough to be merged, especially since we have a long list of PRs depending on this one. Let's address the comments in follow-up PRs.

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed in https://github.com/keep-network/keep-core/pull/2707/commits/9647cd8d01a54a0f6f1a646406a4a42acb5a4c7c as part of https://github.com/keep-network/keep-core/pull/2707

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Done in https://github.com/keep-network/keep-core/pull/2707/commits/eb70b61db31d2351aad51fd88b3ea905449aaaea as part of https://github.com/keep-network/keep-core/pull/2707

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Done in https://github.com/keep-network/keep-core/pull/2707/commits/c97dba585f021ce3e8ec7fb37e2665a4f7577541 as part of #2707

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Done in https://github.com/keep-network/keep-core/pull/2707/commits/c97dba585f021ce3e8ec7fb37e2665a4f7577541 as part of #2707

solidity/random-beacon/contracts/test/SortitionPoolStub.sol
lukasz-zimnoch: Posted a TODO about that in https://github.com/keep-network/keep-core/pull/2707/commits/c97dba585f021ce3e8ec7fb37e2665a4f7577541 as part of #2707

solidity/random-beacon/test/fixtures/index.ts
lukasz-zimnoch: Done in https://github.com/keep-network/keep-core/pull/2707/commits/c97dba585f021ce3e8ec7fb37e2665a4f7577541 as part of #2707

solidity/random-beacon/test/RandomBeacon.Pool.test.ts
lukasz-zimnoch: Done in https://github.com/keep-network/keep-core/pull/2707/commits/c97dba585f021ce3e8ec7fb37e2665a4f7577541 as part of #2707

solidity/random-beacon/test/RandomBeacon.Pool.test.ts
lukasz-zimnoch: Done in https://github.com/keep-network/keep-core/pull/2707/commits/c97dba585f021ce3e8ec7fb37e2665a4f7577541 as part of #2707

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Done in https://github.com/keep-network/keep-core/pull/2707/commits/e0abf511a5bb37282f4f1eb56f13078de1b56e21 as part of #2707

PR #2679, opened 2021-11-01
Clarification renames of DKG result properties, documentation improvements

pdyraga:
Set of follow-up updates after merging #2676 and #2677. No more changes
planned in this PR, the rest of follow-up comments will be addressed in separate PRs.

I strongly encourage to review this PR commit-by-commit.

d717c08 addresses [follow-up comment](https://github.com/keep-network/keep-core/pull/2676#pullrequestreview-793582051) from #2676:

> Second, I think we should rename `misbehaved` to `misbehavedMembersIndices`
and clearly state it should be a set of member indices in range [1, 64]. We do have
`submitterMemberIndex` and `signingMemberIndices` so let's keep it consistent. 

`signingMemberIndices` field has been renamed to `signingMembersIndices` to
clearly indicate indices come from multiple members and are not indices of
a single member.

`misbehaved` field been renamed to `misbehavedMembersIndices` to keep it aligned
with `signingMembersIndices` and `submitterMemberIndex`.

Documentation has been updated to clearly state all these fields should use
values from `[1, 64]` range.

b8761c9 addresses [follow-up comment](https://github.com/keep-network/keep-core/pull/2677#discussion_r739670066) from #2677: `members` are no longer addresses;
these are unique identifiers of operators.

32159d2 addresses [follow-up comment](https://github.com/keep-network/keep-core/pull/2676#pullrequestreview-793582051) from #2676:

> Finally, it is crucial to note that `Groups.setGroupMembers` CHANGES THE ORDER
of members in the group if `misbehaved` is non-empty, compared to what was
outputted by sortition pool. This is very important information to clients and needs to
be reflected in the documentation.

Added a note to `setGroupMembers` that the member indexes may change.
It is very important to understand the final member indexes will be
different from the ones outputted by the group selection protocol in
case at least one group member was misbehaving.
The final group member indexes should be obtained post-DKG.

### DIFF
--- a/solidity/random-beacon/contracts/libraries/Groups.sol
+++ b/solidity/random-beacon/contracts/libraries/Groups.sol
@@ -70,6 +70,13 @@ library Groups {
 
     /// @notice Sets addresses of members for the group eliminating members at
     ///         positions pointed by the misbehaved array.
+    ///
+    ///         NOTE THAT THIS FUNCTION CHANGES ORDER OF MEMBERS IN THE GROUP
+    ///         IF THERE IS AT LEAST ONE MISBEHAVED MEMBER
+    ///
+    ///         The final group members indexes should be obtained post-DKG
+    ///         and they may differ from the ones outputted by the group
+    ///         selection protocol.
     /// @param group The group storage.
     /// @param members Group member addresses as outputted by the group selection
     ///        protocol.
--- a/solidity/random-beacon/contracts/libraries/Groups.sol
+++ b/solidity/random-beacon/contracts/libraries/Groups.sol
@@ -33,14 +33,15 @@ library Groups {
     /// @param groupPubKey Generated candidate group public key
     /// @param members Addresses of candidate group members as outputted by the
     ///        group selection protocol.
-    /// @param misbehaved Array of misbehaved (disqualified or inactive) group
-    ///        members indices; Indices reflect positions of members in the group,
-    ///        as outputted by the group selection protocol.
+    /// @param misbehavedMembersIndices Array of misbehaved (disqualified or
+    ///        inactive) group members indices; Indices reflect positions of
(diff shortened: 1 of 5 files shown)


solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -69,28 +70,33 @@ library Groups {
...
+        uint8[] calldata misbehavedMembersIndices
     ) private {
         group.members = members;
 
         // Iterate misbehaved array backwards, replace misbehaved
         // member with the last element and reduce array length
-        uint256 i = misbehaved.length;
+        uint256 i = misbehavedMembersIndices.length;
         while (i > 0) {
             // group member indices start from 1, so we need to -1 on misbehaved
```
dimpar, line 98: There are a couple of `misbehaved` words left in the comments in this contract. Thoughts on replacing them with `misbehavedMembersIndices` as well?

pdyraga: Please see https://github.com/keep-network/keep-core/pull/2679/commits/ab7ee6c947d7fc67f7c881aa1944d1330275718e. I changed just one, I think other occurrences are better the way they are. E.g. "iterate misbehaved array backwards" in the comment to the code is easier to read and follow than "iterate misbehavedMembersIndices array backwards". :wink:

solidity/random-beacon/test/utils/dkg.ts
```diff
@@ -81,7 +81,7 @@ async function signDkgResult(
   startBlock: number
 ): Promise<{
   members: number[]
-  signingMemberIndices: number[]
+  signingMembersIndices: number[]
```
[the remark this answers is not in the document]
pdyraga: Good catch, please see ab7ee6c.

dimpar (approved):

PR #2686, opened 2021-11-04
Adding group status

dimpar:
Added notion of an expired, terminated, and stale group. This functionality was moved from beacon-v1 with some adjustments to make things work with a new random beacon group.
Additionally, all the group timestamps were changed to block numbers. This is for
calculation purposes since we use values from the other "relay" lib which use block height.

Some related PRs around group status:
- https://github.com/keep-network/keep-core/pull/815
- https://github.com/keep-network/keep-core/pull/982
- https://github.com/keep-network/keep-core-ghsa-m4mj-gfw9-8gj5/pull/2


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -221,6 +221,9 @@ contract RandomBeacon is Ownable {
         relay.setRelayEntrySubmissionEligibilityDelay(10);
         relay.setRelayEntryHardTimeout(5760); // ~24h assuming 15s block time
         relay.setRelayEntrySubmissionFailureSlashingAmount(1000e18);
+
+        groups.setRelayEntryTimeout(relay.relayEntryTimeout());
+        groups.setGroupLifetime(groupLifetime);
```
pdyraga, line 226: We no longer need `groupLifetime` to be a field of `RandomBeacon` and `updateGroupCreationParameters` governance function no longer works. We need similar approach as for `updateRelayEntryParameters` and keep `groupLifetime` only in `Groups`.

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -14,17 +14,28 @@ library Groups {
...
         // Holds keccak256 hashes of group public keys in the order of registration.
         bytes32[] groupsRegistry;
+        // Group ids that were active but failed creating a relay entry.
+        uint64[] activeTerminatedGroups;
         // TODO: Remember about decreasing the counter in case of expiration or termination.
         uint64 activeGroupsCount;
+        // Points to the first active group, it is also the expired groups counter.
+        uint64 expiredGroupOffset;
+        // Lifetime of a group in blocks after which group expires.
+        uint256 groupLifetime;
```
pdyraga, line 36: `RandomBeacon` defines `groupLifetime` as `@notice Group lifetime in seconds`, we need to make sure it's consistent.

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -14,17 +14,28 @@ library Groups {
 
     struct Group {
         bytes groupPubKey;
-        uint256 activationTimestamp;
+        uint256 activationBlockNumber;
```
pdyraga, line 17: Why did we switch to blocks? I know it's how v1 operates but were there any other reasons for that? It feels that >= 2 weeks group lifetime is not prone to any miner manipulation and expressing group lifetime in seconds is more aligned with undelegation time and reward periods.

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -14,17 +14,28 @@ library Groups {
...
+        bool terminated;
     }
 
     struct Data {
         // Mapping of keccak256 hashes of group public keys to groups details.
         mapping(bytes32 => Group) groupsData;
         // Holds keccak256 hashes of group public keys in the order of registration.
         bytes32[] groupsRegistry;
+        // Group ids that were active but failed creating a relay entry.
+        uint64[] activeTerminatedGroups;
```
pdyraga, line 30: If the group was terminated when it was active but now it is expired, it is not included here, right? If yes, I think it's worth stating explicitly in the comment above.

pdyraga (commented):
Just very initial thoughts, more to come later.

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: The trigger was the calculation of a stale group status [here](https://github.com/keep-network/keep-core/blob/groups-expiration/solidity/random-beacon/contracts/libraries/Groups.sol#L216). Relay entry timeout is in blocks and I couldn't add sec + blocks. The result was changing activation timestamp and a group lifetime to blocks.
I think having everything in blocks might be less human-friendly but helps stay consistent across libraries and calculation purposes.

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: Moved to Groups.sol fd01258

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: Sure, done as part of fd01258

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: Sure, it is already described [here](https://github.com/keep-network/keep-core/blob/groups-expiration/solidity/random-beacon/contracts/libraries/Groups.sol#L294) but also added a comment above b72dafa

dimpar:
For reviewers, tests are failing and looking at them now

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -6,27 +6,43 @@ pragma solidity ^0.8.6;
...
         uint64 activeGroupsCount;
+        // Points to the first active group, it is also the expired groups counter.
+        uint64 expiredGroupOffset;
+        // Group lifetime in blocks. When a group reached its lifetime, it
+        // is no longer selected for new relay requests but may still be
+        // responsible for submitting relay entry if relay request assigned
+        // to that group is still pending.
+        uint256 groupLifetime;
+        // Calculated in the Relay.sol lib.
+        uint256 relayEntryTimeout;
```
lukasz-zimnoch, line 37: Bear in mind this value may change. This is because both `relayEntrySubmissionEligibilityDelay` and `relayEntryHardTimeout` are governable. I see we're now setting `relayEntryTimeout` only in `RandomBeacon` constructor which is not enough.

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -148,29 +127,186 @@ library Groups {
...
+                // it will overflow this array and throw an error.
+                i--;
+            }
+        }
+    }
+
+    /// @notice Terminates group with the provided index. Reverts if the group
+    ///         is already terminated.
+    /// @param  groupId Index in the groupRegistry array.
+    function terminateGroup(Data storage self, uint64 groupId) internal {
```
lukasz-zimnoch, line 187: Can we add `terminateGroup` call in `RandomBeacon.reportRelayEntryTimeout` and then address TODOs in `RandomBeacon.Relay.test.ts` regarding group termination?

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -148,29 +127,186 @@ library Groups {
...
         returns (uint64)
     {
-        // TODO: Revisit and include candidate, terminated and expired groups
-        return self.activeGroupsCount;
-        // TODO: Subtract expired and terminated groups
-        // .sub(self.expiredGroupOffset).sub(
-        //     self.activeTerminatedGroups.length)
+        expireOldGroups(self);
+
+        require(numberOfActiveGroups(self) > 0, "No active groups");
```
lukasz-zimnoch, line 221: If I'm not wrong this `require` combined with the logic living in `RandomBeacon.reportRelayEntryTimeout` can block the beacon in certain circumstances. Suppose the following scenario:

1. We have two groups A and B
2. Group A is almost expired, group B handles a relay request
3. Group B fails to provide the entry and the hard timeout is hit
4. Someone wants to report the timeout and calls `RandomBeacon.reportRelayEntryTimeout`
5. During the call ^, group B gets terminated and `groups.numberOfActiveGroups() > 0` condition is true so we want to enter `relay.retryOnEntryTimeout` but before we need to select the new group 
6. Unfortunately, group A exceeded its lifetime so during the `groups.selectGroup` it gets expired during `expireOldGroups` call
7. We revert in `require(numberOfGroups(self) > 0, "No active groups");`

In that case, nobody will be able to report the entry timeout. Please double-check because maybe I miss something but if that scenario is possible, we need to avoid reverting during `selectGroup`.

lukasz-zimnoch (commented):
The first pass, probably more to come later.

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -148,29 +127,186 @@ library Groups {
...
+        internal
+        view
+        returns (bool)
+    {
+        // TODO: Can a group be considered as "stale" if "expiredGroupOffset" was
+        //       not moved forward (group selection wasn't triggered)?
+        //       In other words can it be stale when it's not officially expired?
+        return
+            groupStaleTime(self, self.groupsRegistry[groupId]) < block.number;
     }
```
lukasz-zimnoch, line 310: Are those methods used anywhere?

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: I've already branched off of this PR on my local and planning on creating a new PR that handles reporting. My plan for this PR is to focus on introducing group statuses and later address the reporting part.

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: Not for now, but I'm making ground for future use: rewards and reporting unauthorized signing. Also, I [confirmed](https://discordapp.com/channels/893441201628405760/893441585881161738/905397883782701106) on Discord that we'll need all 4 statuses: active, terminated, expired, and stale.

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -148,29 +127,186 @@ library Groups {
...
+    /// @notice Terminates group with the provided index. Reverts if the group
+    ///         is already terminated.
+    /// @param  groupId Index in the groupRegistry array.
+    function terminateGroup(Data storage self, uint64 groupId) internal {
+        require(
+            !isGroupTerminated(self, groupId),
+            "Group has been already terminated"
+        );
+        self.groupsData[self.groupsRegistry[groupId]].terminated = true;
+        self.activeTerminatedGroups.push();
```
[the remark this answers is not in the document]
dimpar: Sure c1eac04

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -148,29 +127,186 @@ library Groups {
...
+        self.activeGroupsCount--;
+
+        // Sorting activeTerminatedGroups in ascending order so a non-terminated
```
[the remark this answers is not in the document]
dimpar: 60b9c4c

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: Ok, let's add some block numbers to your scenario to see it clearly. What we have here strictly for group selection is 1:1 from beacon v1. There might be some diff how we call it upstream, but from what I can see it is very similar to v1 as well.
- say the relay hard timeout is at block 10
- group B expires at block 15
- group A expires at block 10
4) now at block# 10 and someone called `RandomBeacon.reportRelayEntryTimeout`
5) B is terminated
6) group A (at block 10) is expired
7) revert
Is this correct?

If so, I guess we have at least 3 possible routes:
a) this is what we want and revert is valid, cause there are no active groups.
b) use `if` instead of `require`
c) count the number of actual active groups on the fly "ahead" of time and enter the `cleanupOnEntryTimeout` instead of `retryOnEntryTimeout`. The way it works now is that say all the groups technically expire at block 10. Now it is block # 100. And even though they are all expired "on paper", before we call the 'selectGroup' they are all marked as "active".

#pdyraga thoughts?

lukasz-zimnoch: Yep, that's correct. I see it's from v1 but in v1 we had a very similar problem that blocked the beacon. Regarding proposed solutions:
- Option "a" will block the beacon as we won't be able to report a timeout due to permanent reverts
- Option "b" also won't work. We could return `0` as result of `selectGroup` in case there are no groups but `0` is a valid `groupId`
- Option "c" seems the most promising. Is there an easy way to count the groups "ahead"?

@pdyraga Have you any other ideas?

pdyraga: I think c) but isn't it all about calling `expireOldGroups` in `reportRelayEntryTimeout` after the group is terminated so that `numberOfGroups(self) == 0` one line later?

lukasz-zimnoch: Yep, that's an option but we probably need to decouple `expireOldGroups` from `selectGroup` as well. It depends on the gas cost since `submitRelayEntry` already exceeds our desired limit.

pdyraga: I am still a bit out of context with this code so pardon my potential ignorance, but my understanding is that `selectGroup` in most cases happens in `requestRelayEntry`, so it is unrelated to `submitRelayEntry` gas cost. Another occurrence should be `reportRelayEntryTimeout` transaction that needs to potentially select another group. What did I miss?

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -148,29 +127,188 @@ library Groups {
...
+                        .activeTerminatedGroups[j + 1];
+                }
+                // Resizing the array length by 1. The last element was copied
+                // over in the loop above to an index "second to last". This is
+                // why we can safely remove it from here.
+                self.activeTerminatedGroups.pop();
+                // At this point the length of activeTerminatedGroups[] was shrinked
+                // by 1. We need to adjust 'i'th counter by 1 as well, otherwise
+                // it will overflow this array and throw an error.
+                i--;
```
pdyraga, line 179: What if `i=0`? Say we have `activeTerminatedGroups = [10, 15, 50, 90]` and `self.expiredGroupOffset = 11`.

solidity/random-beacon/contracts/libraries/Groups.sol
lukasz-zimnoch: No, you are right. That's me who mixed up things, sorry 🤦

dimpar: I think what Piotr suggested is the easiest, yet slipped my radar. So `expireOldGroups()` right after the `terminateGroup()` in `reportRelayEntryTimeout`

```
 function reportRelayEntryTimeout() external {
        (..)

        groups.terminateGroup()
        groups.expireOldGroups()

        if (groups.numberOfActiveGroups() > 0) {
            (..)
        } else {
            (..)
        }
    }
```

And yes, we have 1 another place that calls `selectGroup()` in `requestRelayEntry()`. We can leave `selectGroup()` as is or decouple `expireOldGroups()` and move to `requestRelayEntry()`. I think both are fine.

lukasz-zimnoch: Up to you 👍

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: I realized I didn't move the tests from https://github.com/keep-network/keep-core-ghsa-m4mj-gfw9-8gj5/pull/2/files After I'm done with that, I'll have an answer if that's a bug or not.

dimpar: It was a bug. The tests didn't catch it in v1, because the contracts were written in Solidity 0.5.17 which "allowed" going below 0. I checked it on Remix and `0-1` resulted in uin256 max number. Whereas in Solidity 0.8.6 it reverts. Tests that were moved from the security repo branch also confirmed that on my local. Fix + tests are here b57f948

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: Added `groups.expireOldGroups()` + tests f874263. The rest is untouched.

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -148,29 +127,194 @@ library Groups {
...
+    ///         operation timeout are both in the past.
+    function isStaleGroup(Data storage self, bytes memory groupPubKey)
+        internal
+        view
+        returns (bool)
+    {
+        // TODO: Can a group be considered as "stale" if "expiredGroupOffset" was
+        //       not moved forward which means that group selection wasn't triggered?
+        //       In other words can it be stale when it's not officially expired?
+        return groupStaleTime(self, keccak256(groupPubKey)) < block.number;
```
pdyraga, line 298: I suggest we completely remove the concept of a stale group. Then, we do not need to pass `relayEntryTimeout` to this library. For `reportUnauthorizedSigning` let's assume no key can ever leak. 

This will simplify everything and no key ever leaked is a fair expectation.

solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -290,13 +290,25 @@ library Relay {
...
+    /// @notice Calculates a relay entry timeout.
+    /// @return Relay entry timeout in blocks.
+    function relayEntryTimeout(Data storage self)
+        internal
+        view
+        returns (uint256)
+    {
+        return
+            (dkgGroupSize * self.relayEntrySubmissionEligibilityDelay) +
+            self.relayEntryHardTimeout;
```
pdyraga, line 311: If we get rid of a stale group concept, we no longer need this function, right?

solidity/random-beacon/test/RandomBeacon.Parameters.test.ts
```diff
@@ -43,51 +42,79 @@ describe("RandomBeacon - Parameters", () => {
...
-      let tx
-      beforeEach(async () => {
-        tx = await randomBeacon
-          .connect(governance)
-          .updateRelayEntryParameters(
-            relayRequestFee,
-            relayEntrySubmissionEligibilityDelay,
-            relayEntryHardTimeout,
-            callbackGasLimit
+      context("when updating relay params", () => {
```
pdyraga, line 45: Looks like some merge conflict resolution problem:
updateRelayEntryParameters -> ... -> when updating relay params
🤔

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -638,6 +638,47 @@ describe("RandomBeacon - Relay", () => {
         }
       )
 
+      context(
+        "when timeout happens after all the group terminate or expire",
```
pdyraga, line 642: Isn't it covered by the next context: when no active groups exist after timeout is reported? Line 682.

pdyraga (commented):
Made a high-level review, leaving a detailed one to @lukasz-zimnoch.

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -148,29 +127,194 @@ library Groups {
...
+            block.number
+        ) {
+            Group memory group = self.groupsData[
+                self.groupsRegistry[self.expiredGroupOffset]
+            ];
+            // Decrement only for non-terminated groups, because terminateGroup()
+            // function handles 'activeGroupsCount' counter for terminated groups.
+            if (!group.terminated) {
+                self.activeGroupsCount--;
+            }
```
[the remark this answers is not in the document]
dimpar: >  I do not see us doing this check in v1, why is it needed?

It is needed because just like you said in your second comment we used `self.groupRegistry.length - self.expiredGroupOffset - self.activeTerminatedGroups.length`. If a group was terminated and then it was expired, we do not want to decrement the `self.activeGroupsCount` counter 2 times for the same group no matter if it's `terminated` or `expired`. I tried to use the original approach that was used in v1, but it bite me in other tests and the reason is that it is coupled with `expiredGroupOffset`. In other words, the only place where we move `expiredGroupOffset` is `expireOldGroups()`, so every time we call `activeTerminatedGroups()` without `expireOldGroups()` it may return an outdated result, cause the blocks already moved forward.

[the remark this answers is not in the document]
pdyraga: I am still not following. I think we shouldn't have `self.activeGroupsCount` stored as a field and incrementing/decrementing it. Instead, we should have a function and compute it on the fly as `self.groupRegistry.length - self.expiredGroupOffset - self.activeTerminatedGroups.length`.

In `expireOldGroups` we first move the offset and in the same function, in the same transaction, we update `activeTerminatedGroups` based on the updated offset. When `expireOldGroups` finished, we should not have expired group in `activeTerminatedGroups` so I think `self.groupRegistry.length - self.expiredGroupOffset - self.activeTerminatedGroups.length` holds.

> activeTerminatedGroups() without expireOldGroups() it may return an outdated result, cause the blocks already moved forward.

Correct, and that's why no one but `Groups` library should use `activeTerminatedGroups`. This is similar situation as with `activeGroupsCount()` - it returns the number of groups based on the current stored state, and the stored state may be outdated. The group is not actually expired until the offset is updated on the chain.

Can you give some concrete example when it does not work? This would help me understand what the problem is.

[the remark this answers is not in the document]
dimpar: > Can you give some concrete example when it does not work?

We have two tests that are failing if we change a counter with what we had in v1, but they are failing for the same reason. It boils down to at what point we consider a group to be "active". We have 2 functions: `addCandidateGroup` & `activateCandidateGroup` and in the current approach, we consider it to be active when `activateCandidateGroup()` is called. In `addCandidateGroup()` we populate `groupsData` map & `groupsRegistry` array, but increment the counter in `activateCandidateGroup()`.
Here are the 2 tests that will fail [here](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/test/Groups.test.ts#L597) and [here](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/test/Groups.test.ts#L695). Take a look at the `beforeEach` where we add and activate groups. In the v1 approach, we analyzed if the group was active or not just by looking at the `groupsRegistry` and we didn't have a notion of separately activating a group. And that's the difference.
`activateCandidateGroup()` is used by `approveDkgResult()`. Not entirely sure how it can impact the functionality down the road if we decide to change it.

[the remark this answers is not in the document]
dimpar: Actually, we have another test that will [here](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts#L172). It expects "current state is not IDLE" but was reverted with "not awaiting genesis"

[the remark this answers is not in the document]
pdyraga: I think we could solve it by one additional condition in `numberOfActiveGroups()`. After we evaluate the number with `self.groupRegistry.length - self.expiredGroupOffset - self.activeTerminatedGroups.length`, we need to check if the last group in `self.groupRegistry` has non-zero activation field. If not, we decrease the result by one.

I think this is safer than maintaining the count separately and making sure it's properly updated in all cases. Also, we don't have to read the whole structure when expiring groups. With several groups to expire this could get expensive.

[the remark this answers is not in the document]
dimpar: I had to add a couple of extra checks to use a v1 approach, but overall the idea of checking if the last group is active makes sense. Also had to change tests, that added a couple of groups and activated only the last one, which from my knowledge is not possible. eb46440

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: Ok, it will certainly simplify things. The stale functionality was used along with rewards in v1. If the rewards structure takes this into account, then I think we are fine. 921f30b

solidity/random-beacon/contracts/libraries/Relay.sol
dimpar: Yup, removed 921f30b

solidity/random-beacon/test/RandomBeacon.Parameters.test.ts
dimpar: dc75911

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
dimpar: I wanted to cover the scenario described in [this comment](https://github.com/keep-network/keep-core/pull/2686#discussion_r750128611). Reworded this context, if you have a better idea, let's do it. 07d528f

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -268,7 +262,7 @@ contract RandomBeacon is Ownable {
         // TODO: revisit all initial values
         callbackGasLimit = 200e3;
         groupCreationFrequency = 10;
-        groupLifetime = 2 weeks;
```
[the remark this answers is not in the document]
dimpar: Sure, 65a44fc

dimpar:
@pdyraga @lukasz-zimnoch all the comments are addressed

lukasz-zimnoch (approved):

PR #2690, opened 2021-11-05
DKG: rewards for result submission and timeout notification

tomaszslabon:
This PR modifies `random-beacon` so that it [pays rewards to a DKG result submitter]( https://github.com/keep-network/keep-core/blob/96d328de2c5c54ab7ef88c06c40a198ceba8d2d4/solidity/random-beacon/contracts/RandomBeacon.sol#L454)(but only after the result was approved, not on the result submission itself) and [DKG result timeout notifier](https://github.com/keep-network/keep-core/blob/96d328de2c5c54ab7ef88c06c40a198ceba8d2d4/solidity/random-beacon/contracts/RandomBeacon.sol#L443)

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -457,8 +457,8 @@ contract RandomBeacon is Ownable {
         groups.activateCandidateGroup();
         dkg.complete();
 
-        // TODO: Handle DQ/IA
-        // TODO: Release a rewards to DKG submitter.
+        // TODO: Handle DQ/IA. Should result submitter receive
+        //       reward if they failed to call this function?
         // TODO: Unlock sortition pool
     }
 
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -69,6 +69,7 @@ contract RandomBeacon is Ownable {
     using Groups for Groups.Data;
     using Relay for Relay.Data;
     using Callback for Callback.Data;
+    using SafeERC20 for IERC20;
 
     // Constant parameters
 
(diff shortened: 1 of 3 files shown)


solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -435,11 +439,11 @@ library DKG {
 
     /// @notice Cleans up state after DKG completion.
     /// @dev Should be called after DKG times out or a result is approved.
-    modifier cleanup(Data storage self) {
-        _;
+    function cleanup(Data storage self) internal {
```
tomaszslabon, line 442: I decided to make `cleanup` a function instead of modifier, because we need to pay the reward to the block submitter before we clean up (cleanup zeroes the submitter address stored in DKG data and we need that address to pay the reward). After this change `cleanup` is executed in the very end of `approveDkgResult` and `notifyDkgTimeout`.

nkuba: What if we rename the function to `complete`? We already call `dkg.start()` function, so maybe it would make sense to call `dkg.complete()`?

tomaszslabon: It makes sense, done in https://github.com/keep-network/keep-core/pull/2690/commits/aea62227d52367e34dc04941ae806ddc3a42faff.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -434,21 +435,28 @@ contract RandomBeacon is Ownable {
...
     ///         submitted result is finished. Considers the submitted result as
-    ///         valid and completes the group creation by activating the candidate
-    ///         group.
+    ///         valid, pays reward to the result submitter and completes the
+    ///         group creation by activating the candidate group.
     function approveDkgResult() external {
         dkg.approveResult();
-
+        // Pay the DKG result submission reward.
+        relay.tToken.safeTransfer(
```
lukasz-zimnoch, line 454: And same here.

tomaszslabon: Done in https://github.com/keep-network/keep-core/pull/2690/commits/443a64b2e12a7577f113498f5c40e0bcab5a19de.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -434,21 +435,28 @@ contract RandomBeacon is Ownable {
...
 
-    /// @notice Notifies about DKG timeout.
+    /// @notice Notifies about DKG timeout. Pays the sortition pool unlocking
+    ///         reward to the notifier.
     function notifyDkgTimeout() external {
         dkg.notifyTimeout();
-
-        // TODO: Pay a reward to the caller.
+        // Pay the sortition pool unlocking reward.
+        relay.tToken.safeTransfer(msg.sender, sortitionPoolUnlockingReward);
```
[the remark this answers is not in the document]
tomaszslabon: Done in https://github.com/keep-network/keep-core/pull/2690/commits/443a64b2e12a7577f113498f5c40e0bcab5a19de.

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -1048,12 +1092,17 @@ describe("RandomBeacon - Group Creation", () => {
 
             context("with challenge period passed", async () => {
               let tx: ContractTransaction
+              let initialSubmitterBalance: BigNumber
```
pdyraga, line 1095: Not a const? 🤔

tomaszslabon: I think I cannot use const here.
If something is `const` it needs to be initialised on declaration, so I would have to call:
```
 const initialSubmitterBalance = await testToken.balanceOf(
      submitterAddress1
 )
```

If I call it inside `beforeEach`, it won't be visible in `it` testcases.
If I call it in `context`, outside `beforeEach`, it won't work either.

pdyraga (approved):

PR #2691, opened 2021-11-05
Random Beacon v2 sortition pool locking

lukasz-zimnoch:
Depends on: #2671 
Depends on: https://github.com/keep-network/sortition-pools/pull/132

The sortition pool should be locked upon group creation start and unlocked either during result approval or DKG timeout.


lukasz-zimnoch:
Ready for review. Keeping it draft until #2671 lands on main.

pdyraga:
Is there a possibility we can make this PR not depending on #2671? The change looks pretty straightforward while the one from #2671 will require more work and I think there is no point in blocking this one. The merge shouldn't be that problematic I think.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -611,8 +621,17 @@ contract RandomBeacon is Ownable {
...
         if (relay.requestCount % groupCreationFrequency == 0) {
-            // TODO: Once implemented, invoke:
-            // createGroup(uint256(keccak256(entry)));
+            // Just in case, make sure the group creation can be triggered to
+            // avoid unexpected revert. There is a possibility entries are
+            // submitted very quickly and if the group creation frequency
+            // is small enough, group creations may overlap.
+            if (dkg.currentState() == DKG.State.IDLE) {
+                // slither-disable-next-line reentrancy-events
+                createGroup(uint256(keccak256(entry)));
```
pdyraga, line 630: [RFC 19](https://github.com/keep-network/keep-core/blob/main/docs/rfc/rfc-19-random-beacon-v2.adoc) says:

>  Group creation start transaction should be embedded into relay request transaction and should lock a sortition pool. From this moment, no operator can enter or leave the pool. Once a new relay entry appears on the chain, all off-chain clients should perform group selection optimistically, using the new entry value and a view sortition pool function call. 

So the flow should be:
- Relay entry is requested. If `relay.requestCount % groupCreationFrequency == 0` and if `dkg.currentState() == DKG.State.IDLE`, we begin DKG by locking the pool.
- When relay entry is submitted, we request the creation of a new group with that relay entry. 
- DKG timeout starts counting from the moment relay entry is published.

Seed for group creation must not be known at the moment pool is locked to protect against the manipulation of sortition pool to get favorable group composition. With the current code, seed is not published to the chain but it is know - it sits in the mempool.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -611,8 +621,17 @@ contract RandomBeacon is Ownable {
...
+            } else {
+                // slither-disable-next-line reentrancy-events
+                emit DkgStartFailed();
```
pdyraga, line 633: I would not consider it as any kind of a failure, it's a normal situation.

pdyraga (changes requested):

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Yeah, this event is just in case and can make debugging easier. Do you think we should get rid of it?

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Good point! I misunderstood the RFC. Will change.

solidity/random-beacon/contracts/RandomBeacon.sol
pdyraga: Yeah, I'd remove it to not confuse anyone. It's also some additional cost - really small but why pay it given this situation is normal. 🤷🏻

lukasz-zimnoch:
> Is there a possibility we can make this PR not depending on #2671? The change looks pretty straightforward while the one from #2671 will require more work and I think there is no point in blocking this one. The merge shouldn't be that problematic I think.

#2671 introduce real SP in relay tests so it was easier to stack this PR on top than implementing tests using the stub and then change to real pool.

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: https://github.com/keep-network/keep-core/pull/2691/commits/eb221288d15eb715257eac977f68bb68efc20f5e

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: I have refactored the solution see: https://github.com/keep-network/keep-core/pull/2691/commits/f0a27d9d5da5eeca5344ca02271cbcbab02280d2. I'm still working on tests though. I think embedding pool lock/unlock logic within DKG state machine is the safest option. Keeping it separate seems error-prone because we can lock the pool forever or unlock it during group creation.

lukasz-zimnoch: Tests are ready as well: https://github.com/keep-network/keep-core/pull/2691/commits/efbe99ef5b55ea48841c74e94dc52202a7cbccc8

lukasz-zimnoch:
@pdyraga Ready for another shot.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -71,6 +71,8 @@ library DKG {
         // Group creation is not in progress. It is a state set after group creation
         // completion either by timeout or by a result approval.
         IDLE,
+        // Group creation is warmed up and sortition pool is locked.
+        WARMED_UP,
```
pdyraga, line 75: Maybe `AWAITING_SEED`? This is what is actually happening: DKG locked the pool and is awaiting seed from the random beacon when it's in this state.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -148,24 +150,40 @@ library DKG {
...
+                    }
                 }
             }
         }
     }
 
-    function start(Data storage self, uint256 seed) internal {
+    /// @notice Warms up the group creation process by locking the sortition
+    ///         pool.
+    function warmUp(Data storage self) internal {
```
pdyraga, line 172: Maybe `lockState()` or `lockSortitionPool()`? I think this would be easier to follow from the perspective of `RandomBeacon` contract. That contract also uses the pool (join / update operator), so being explicit that it's about to be locked makes sense.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -443,16 +449,21 @@ contract RandomBeacon is Ownable {
...
 
+        createGroupWarmUp();
+
         createGroup(
             uint256(keccak256(abi.encodePacked(genesisSeed, block.number)))
         );
     }
 
+    /// @notice Warms up the group creation process.
+    function createGroupWarmUp() internal {
```
pdyraga, line 460: Can't we call `dkg.warmUp()` (or `dkg.lockState()` - see other comment) directly? We are not really creating anything here :shrug:

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -571,6 +581,15 @@ contract RandomBeacon is Ownable {
         relay.requestEntry(groupId);
 
         callback.setCallbackContract(callbackContract);
+
+        // If current request should trigger group creation and group creation
+        // is not currently in progress we should start the warm up process.
```
pdyraga, line 586: `trigger group creation and group creation`? 😆 

```suggestion
        // If the current request should trigger group creation we need to lock
        // DKG state (lock sortition pool) to prevent operators from changing its
        // state before relay entry is known. That entry will be used as a group
        // selection seed.
```

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -608,9 +627,10 @@ contract RandomBeacon is Ownable {
             staking.slash(slashingAmount, groupMembers);
         }
 
-        if (relay.requestCount % groupCreationFrequency == 0) {
-            // TODO: Once implemented, invoke:
-            // createGroup(uint256(keccak256(entry)));
+        // If group creation is warmed up, that means the we should start
+        // the actual group creation process.
+        if (dkg.currentState() == DKG.State.WARMED_UP) {
```
pdyraga, line 632: `DKG.State.AWAITING_SEED` would look really pretty here. 😉

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -445,5 +463,17 @@ library DKG {
         delete self.submittedResultHash;
         delete self.resultSubmitter;
         delete self.submittedResultBlock;
+        self.sortitionPool.unlock();
+    }
+
+    /// @notice Cancels the warm up of group creation process by unlocking
+    ///         the sortition pool.
+    function coolDown(Data storage self) internal {
```
pdyraga, line 471: Maybe `notifySeedTimedOut` and put this function next to `notifyTimeout`?

pdyraga (commented):
First-round of comments. Still reviewing, more to come.

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: We need to call it directly in tests via `publicCreateGroupWarmUp`. Calling `dkg.warmUp` directly in the stub causes a compilation error.

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -1159,7 +1175,11 @@ describe("RandomBeacon - Group Creation", () => {
 
           await mineBlocks(params.dkgResultChallengePeriodLength)
 
-          await randomBeacon.approveDkgResult()
+          tx = await randomBeacon.approveDkgResult()
```
pdyraga, line 1178: `tx` is not used and I believe we can also get rid of `let tx: ContractTransaction`

pdyraga: Or... was the intention to add a simple `it("should work` to this describe given that we changed `it("should succeed"` into `beforeEach`? I think adding check about this event is a good move. We could check there some really basic stuff, e.g. if `DkgResultApproved` and `GroupActivated` were emitted.

solidity/random-beacon/contracts/libraries/DKG.sol
lukasz-zimnoch: Yeah, good idea. I had no idea for a good nomenclature here. Thanks for the inspiration! Will change as you proposed.

solidity/random-beacon/contracts/RandomBeacon.sol
pdyraga: Let's then name it the same as `dkg.` function named, document in `@dev` we need it for tests and move to the bottom of `internal` functions.

And one day, let's get rid of the stub exposing `publicCreateGroupWarmUp` completely... 🙏🏻 🤞🏻

solidity/random-beacon/contracts/test/SortitionPoolStub.sol
```diff
@@ -8,6 +8,7 @@ import "../RandomBeacon.sol";
 // Stub contract used in tests
 contract SortitionPoolStub is ISortitionPool {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yep I'm aware but this should be done at once to avoid build failures 🤷‍♂️ I posted a TODO (in https://github.com/keep-network/keep-core/pull/2707) informing this stub is deprecated and real SP should be used.

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -574,34 +663,92 @@ describe("RandomBeacon - Relay", () => {
...
+            // expect(await randomBeacon.isRelayRequestInProgress()).to.be.false
+          })
+
+          it("should cool down group creation", async () => {
+            // TODO: Uncomment those assertions once termination is implemented.
+            // expect(await randomBeacon.getGroupCreationState()).to.be.equal(
+            //   dkgState.IDLE
+            // )
+            // expect(await sortitionPool.isLocked()).to.be.false
+          })
```
pdyraga, line 708: To make it simpler and do not repeat checks that are not related to the warmup (awaiting seed) state, I'd do:

```
when no active groups exist after timeout is reported:
  ... put all the checks here
when no active groups exist after timeout is reported and DKG is awaiting seed:
  ... just assertions about: a) changing DKG state; b) unlocking pool
```

pdyraga: And I believe assertions a) b) can be now implemented and do not require any uncomment TODO.

solidity/random-beacon/contracts/test/SortitionPoolStub.sol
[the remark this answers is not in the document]
pdyraga: Yeah, it's not something we can do in this PR. I'll just keep complaining in every single PR until it's done so that we can all feel the pain lol.

solidity/random-beacon/contracts/libraries/DKG.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2691/commits/a29a5ad1f24125d53b37cef20129d6c5977764be

solidity/random-beacon/contracts/libraries/DKG.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2691/commits/a29a5ad1f24125d53b37cef20129d6c5977764be

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Done as part of https://github.com/keep-network/keep-core/pull/2691/commits/a29a5ad1f24125d53b37cef20129d6c5977764be

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Done as part of https://github.com/keep-network/keep-core/pull/2691/commits/a29a5ad1f24125d53b37cef20129d6c5977764be

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2691/commits/a29a5ad1f24125d53b37cef20129d6c5977764be

solidity/random-beacon/contracts/libraries/DKG.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2691/commits/a29a5ad1f24125d53b37cef20129d6c5977764be

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
lukasz-zimnoch: Regarding the latter, it was not an intention. I just refactored that case to use `beforeEach` and be easily extensible with new `it` assertions. The basic stuff you mentioned is already checked in earlier cases above so checking them again seems redundant. Regarding usused `tx`, see: https://github.com/keep-network/keep-core/pull/2691/commits/56ead30afdaad1d6fce5f6a6d54e61ced13cbddb

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -1045,6 +1051,10 @@ describe("RandomBeacon - Group Creation", () => {
                   .to.emit(randomBeacon, "GroupActivated")
                   .withArgs(0, groupPublicKey)
               })
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/commit/56ead30afdaad1d6fce5f6a6d54e61ced13cbddb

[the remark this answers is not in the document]
pdyraga: One more: I think we should also assert it for `describe("challengeDkgResult"`

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
pdyraga: That "should suceed" we had before was a kind of a shortcut: we don't want to repeat all previous assertions but we also want to have an explicit assertion this transaction passes at `startBlock + dkgTimeout - 1` and is supposed to do everything else as in the case tested in the earlier context.

This is also tested here because `beforeEach` would fail but... I think we miss an explicit "should work" assertion that at least suggests we expect exactly the same behavior as before. I do agree repeating the same tests does not make sense but I'd add that "should work" with at least one or two events checked so that we are not only focused on unlocking the pool but also saying the result needs to be accepted at that time, just like in any other case.

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2691/commits/1797b660235df348f267aabb9e9e929e07a88e15. However, assertions must be commented until we have group termination implemented. If the group is not terminated we enter in the `groups.numberOfActiveGroups() > 0` branch in `reportRelayEntryTimeout`. Once we have termination we can enter in the `else` and test the behavior of `notifySeedTimedOut`

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
lukasz-zimnoch: Done as part of https://github.com/keep-network/keep-core/pull/2691/commits/2a790b469e3d5dd9e857523f9be5b6f67ac3e9bc

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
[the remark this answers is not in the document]
lukasz-zimnoch: Done as part of https://github.com/keep-network/keep-core/pull/2691/commits/2a790b469e3d5dd9e857523f9be5b6f67ac3e9bc

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -348,6 +366,17 @@ library DKG {
...
 
+    /// @notice Notifies about the seed was not delivered and restores the
+    ///         initial DKG state (IDLE).
+    function notifySeedTimedOut(Data storage self) internal {
+        require(
+            currentState(self) == State.AWAITING_SEED,
+            "current state is not AWAITING_SEED"
+        );
+
+        self.sortitionPool.unlock();
```
pdyraga, line 377: `emit DkgSeedTimedOut`?

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -575,6 +662,10 @@ describe("RandomBeacon - Relay", () => {
           // relayEntryHardTimeout`.
           await mineBlocks(64 * 10 + 5760)
 
+          // Simulate group creation is already in progress.
+          await (randomBeacon as RandomBeaconStub).publicDkgLockState()
+          await (randomBeacon as RandomBeaconStub).publicCreateGroup()
```
pdyraga, line 667: Why do we need to call these here? `publicCreateGroup` seems not to be needed and we are calling `publicDkgLockState` in the context for "and DKG is awaiting seed" at line 700.

lukasz-zimnoch: Done in https://github.com/keep-network/keep-core/pull/2691/commits/166a6c8d38f900aa6d0e2df4c72243d870c3e25a

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -451,8 +459,6 @@ contract RandomBeacon is Ownable {
     /// @notice Creates a new group.
     /// @param seed Seed for DKG.
     function createGroup(uint256 seed) internal {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Done in https://github.com/keep-network/keep-core/pull/2691/commits/166a6c8d38f900aa6d0e2df4c72243d870c3e25a

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -148,24 +150,40 @@ library DKG {
...
         }
     }
 
-    function start(Data storage self, uint256 seed) internal {
+    /// @notice Locks the sortition pool and starts awaiting for the
+    ///         group creation seed.
+    function lockState(Data storage self) internal {
         require(currentState(self) == State.IDLE, "current state is not IDLE");
 
+        self.sortitionPool.lock();
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2691/commits/a425d466399a238a7897507eb4e00ba8061009f5

solidity/random-beacon/contracts/libraries/DKG.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2691/commits/a425d466399a238a7897507eb4e00ba8061009f5

lukasz-zimnoch:
@pdyraga Everything addressed.

pdyraga (approved):

PR #2695, opened 2021-11-09
Random Beacon v2 DKG challenge verification and slashing

lukasz-zimnoch:
Depends on: https://github.com/keep-network/sortition-pools/pull/133

This pull request introduces actual DKG challenge verification. The verification code calls the sortition pool to select the group using the DKG execution seed and computes a hash from the group members list. Then, it compares that hash with the hash computed from the submitted group members list. In case those hashes differ, the DKG challenge is deemed justified. A justified challenge causes the slashing of the members who signed the malicious result and triggers a notifier reward payment.


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -460,12 +475,31 @@ contract RandomBeacon is Ownable {
...
+
+        // The notifier should receive 5% of the total slashing amount. The
+        // contract doesn't make a direct transfer but an allowance instead
+        // as there is no guarantee the contract has a sufficient token
+        // balance at the moment. Making a transfer without funds would cause
+        // a revert of the challenge transaction.
+        tToken.safeIncreaseAllowance(
+            msg.sender,
+            (maliciousMembers.length * slashingAmount) / 20
+        );
```
lukasz-zimnoch, line 502: @pdyraga Can you confirm this slashing and rewarding logic is correct? The RFC says " the notifier receives 5% from the slashed amount". If I understand correctly, slashing will burn 100% of tokens so the `RandomBeacon` must pay the reward from its balance. The reward amount is 5% of the total (?) slashing amount. Also, we can't transfer directly because there is no guarantee we have enough balance so making just an allowance seems the right option. We can also make the transfer only if the balance is good and the allowance otherwise.

pdyraga: Nope, this is not correct. We should use `seize` function of the staking contract, see [here](https://github.com/threshold-network/solidity-contracts/blob/fa83b98566b7c786a499fc438d97b4dd3eee0293/contracts/staking/TokenStaking.sol#L893-L900). It will seize tokens and pay reward to `notifier` address. We do not need to care about `RandomBeacon` balance because the balance of this contract is just the maintenance pool and is not used for the reward of relay entry timeout or malicious DKG result notifier.

lukasz-zimnoch: Yep, I wasn't aware of `seize`. Will fix.

lukasz-zimnoch: And fixed: https://github.com/keep-network/keep-core/pull/2695/commits/e0e765b52541dc79d61274f6ca8f09675096bfbc

lukasz-zimnoch:
@pdyraga Ready for another look!

pdyraga:
@lukasz-zimnoch Can you solve merge conflicts? Would love to have the code in the final form before the review.

lukasz-zimnoch:
> @lukasz-zimnoch Can you solve merge conflicts? Would love to have the code in the final form before the review.

Done @pdyraga

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -208,6 +214,12 @@ contract RandomBeacon is Ownable {
         address indexed challenger
     );
 
+    event DkgResultMaliciousSlashed(
```
pdyraga, line 217: I'd rather say `DkgMaliciousResultSlashed` because it's the result that was malicious, not the slashing 😉

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -566,12 +577,21 @@ contract RandomBeacon is Ownable {
...
         groups.popCandidateGroup();
 
-        // TODO: Implement slashing
+        emit DkgResultMaliciousSlashed(
+            resultHash,
+            slashingAmount,
+            maliciousMembersAddresses
+        );
+
+        staking.seize(slashingAmount, 5, msg.sender, maliciousMembersAddresses);
```
pdyraga, line 594: This 5% should be governable, just like `relayEntryTimeoutNotificationRewardMultiplier` is governable. 5% in the RFC was suggested before we knew how T staking contract should look like. T staking contract does not give that 5% from `slashingAmount` just like KEEP staking contract was doing but instead, it is giving it from a special notifier reward pool.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -566,12 +577,21 @@ contract RandomBeacon is Ownable {
     ///         It removes a candidate group that was previously registered with
     ///         the DKG result submission.
     function challengeDkgResult() external {
-        // TODO: Determine parameters required for DKG result challenges.
-        dkg.challengeResult();
+        bytes32 resultHash = dkg.submittedResultHash;
```
pdyraga, line 580: `resultHash` is only used in the event, do we need to cache it this way? Shouldn't we inline the call where we emit `DkgResultMaliciousSlashed`?

lukasz-zimnoch: Yep, we need it. Some cleanup is done within `dkg.challengeResult()`.

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2695/commits/f0db676b0854612c89aabe9532493efb07dbd660

solidity/random-beacon/contracts/RandomBeacon.sol
pdyraga: Can we have challengeResult returning the hash? I think this is cleaner - we do not have to look inside about what information DKG library stores and when it is cleared up.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -303,75 +295,86 @@ library DKG {
...
             )
         );
 
         bytes memory current; // Current signature to be checked.
-
         bool[] memory usedMemberIndices = new bool[](groupSize);
+        signingMembers = new uint32[](signaturesCount);
+        address[] memory membersAddresses = sortitionPool.getIDOperators(
+            result.members
+        );
```
pdyraga, line 351: We do not need to fetch the addresses of all members, addresses of signing group members should be enough. Do we know what is the gas cost difference when working with a real sortition pool? Fetching just signing member addresses will require at least one iteration through `signingMembersIndices` and building an array with their identifiers but maybe it's still cheaper than doing more SREADs on the sortition pool side.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -202,15 +209,7 @@ library DKG {
...
-        verify(
-            self,
-            result.submitterMemberIndex,
-            result.groupPubKey,
-            result.misbehavedMembersIndices,
-            result.signatures,
-            result.signingMembersIndices,
-            result.members
-        );
+        uint32[] memory signingMembers = verify(self, result);
```
pdyraga, line 212: @nkuba Was there any particular reason it was done this way? Gas costs maybe?

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -257,6 +271,12 @@ describe("RandomBeacon - Group Creation", () => {
 
           context("when dkg result was challenged", async () => {
             beforeEach(async () => {
+              // Make the challenge justified by forcing the pool to return
+              // an other actual group than the submitted one.
```
pdyraga, line 275: s/an other actual group/another group ?

I think this will be easier to reason about

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: Sure: https://github.com/keep-network/keep-core/pull/2695/commits/7c9f436788770bf52471df6721aa35ce6aa6f937

solidity/random-beacon/contracts/libraries/DKG.sol
lukasz-zimnoch: We don't have a real sortition pool plugged-in yet to make a fair comparison. Maybe we can leave a TODO and check this while doing the optimization work?

solidity/random-beacon/contracts/libraries/DKG.sol
lukasz-zimnoch: Can't compare gas usage as the original version errors out with `Stack too deep` at compilation stage 🤷

pdyraga: Problem solved then.

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -1558,23 +1659,60 @@ describe("RandomBeacon - Group Creation", () => {
...
+              groupPublicKey,
+              actualGroupMembers,
+              startBlock,
+              noMisbehaved
+            )
+          })
+
+          it("should revert with 'unjustified challenge' error", async () => {
+            await expect(randomBeacon.challengeDkgResult()).to.be.revertedWith(
+              "unjustified challenge"
```
pdyraga, line 1685: We need to lower or upper-case all revert messages one day. Not something for this PR, just complaining as always.

solidity/random-beacon/contracts/libraries/DKG.sol
pdyraga: Fine for me, let's leave a TODO.

pdyraga:
What is the gas cost of `submitDkgResult`, `approveDkgResult`, and `challengeDkgResult` compared to their versions in `main`?

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2695/commits/2f4e1b12c7674db2500142e70b384527b5f4afd7

solidity/random-beacon/contracts/libraries/DKG.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2695/commits/52a1623ca0747be3944ad5e3fe060e3d9b6b05cf

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -152,6 +152,14 @@ contract RandomBeacon is Ownable {
     ///         operator affected.
     uint256 public relayEntryTimeoutNotificationRewardMultiplier;
 
+    /// @notice Percentage of the DKG malicious result notification reward
```
pdyraga, line 155: Actually, this should say "Percentage of the staking contract malicious behavior notification reward...", same for `relayEntryTimeoutNotificationRewardMultiplier`. Can we fix it? 🙏🏻

lukasz-zimnoch: Sure: https://github.com/keep-network/keep-core/pull/2695/commits/0d6f0ab25225ac5f2407f3a0d07b2b922fd33c89

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2695/commits/0d6f0ab25225ac5f2407f3a0d07b2b922fd33c89

lukasz-zimnoch:
> What is the gas cost of `submitDkgResult`, `approveDkgResult`, and `challengeDkgResult` compared to their versions in `main`?



> What is the gas cost of `submitDkgResult`, `approveDkgResult`, and `challengeDkgResult` compared to their versions in `main`?

| Method               | Before (average) | After (average) |
|----------------------|------------------|-----------------|
| `submitDkgResult`    | 1182702          | 1727923         |
| `approveDkgResult`   | 144961           | 164681          |
| `challengeDkgResult` | 79012            | 512847          |

Bear in mind `challengeDkgResult` is just a starting point in `main`.

pdyraga:
I am fine with `challengeDkgResult` being more expensive but the cost difference for `submitDkgResult` is too high. Assuming 115 Gwei gas price today - and it's not the highest gas price nowadays - plus $4237 ETH, it was $576 to submit DKG result and now it is $841 😬 I think 1.1M was already too high.

What makes it so expensive? Storing seed, the hash of submitted members, or computing that hash? Instead of storing the hash, we can read information from `Groups` library in `RandomBeacon.challengeDkgResult` and pass it to `DKG` library. This way, non-happy-path challenger pays and the happy path is less expensive. Though I don't know if this is the problematic piece.

lukasz-zimnoch:
> I am fine with `challengeDkgResult` being more expensive but the cost difference for `submitDkgResult` is too high. Assuming 115 Gwei gas price today - and it's not the highest gas price nowadays - plus $4237 ETH, it was $576 to submit DKG result and now it is $841 😬 I think 1.1M was already too high.
> 
> What makes it so expensive? Storing seed, the hash of submitted members, or computing that hash? Instead of storing the hash, we can read information from `Groups` library in `RandomBeacon.challengeDkgResult` and pass it to `DKG` library. This way, non-happy-path challenger pays and the happy path is less expensive. Though I don't know if this is the problematic piece.

We can't get group members from `Groups` library. This is because members are set through `setGroupMembers` which changes the group members order in case there is at least one misbehaved member. But this is not the problematic part. A bigger problem is storing IDs of members who signed the result. We need to store it and use it in case a justified challenge emerges because we need to know who signed the malicious result and should be slashed. Slashing all the group will be cheaper but not fair against honest members. Anyway, `submitDkgResult` is expensive even without changes from this PR and it is still growing (min stake check in progress) so simple optimizations may not be enough. I'm afraid a bigger refactoring will be required.

pdyraga:
> > I am fine with `challengeDkgResult` being more expensive but the cost difference for `submitDkgResult` is too high. Assuming 115 Gwei gas price today - and it's not the highest gas price nowadays - plus $4237 ETH, it was $576 to submit DKG result and now it is $841 😬 I think 1.1M was already too high.
> > What makes it so expensive? Storing seed, the hash of submitted members, or computing that hash? Instead of storing the hash, we can read information from `Groups` library in `RandomBeacon.challengeDkgResult` and pass it to `DKG` library. This way, non-happy-path challenger pays and the happy path is less expensive. Though I don't know if this is the problematic piece.
> 
> We can't get group members from `Groups` library. This is because members are set through `setGroupMembers` which changes the group members order in case there is at least one misbehaved member. But this is not the problematic part. A bigger problem is storing IDs of members who signed the result (64 `uint32` in the worst case). We need to store it and use it in case a justified challenge emerges because we need to know who signed the malicious result and should be slashed. Slashing all the group will be cheaper but not fair against honest members. Anyway, `submitDkgResult` is expensive even without changes from this PR and it is still growing (min stake check in progress) so simple optimizations may not be enough. I'm afraid a bigger refactoring will be required.

Let's then not store anything and expect the challenger/approver to pass the required data as parameters. We have `submittedResultHash` stored and it's a hash of the entire result. We do not need to store `submittedResultSigningMembers` and`submittedResultMisbehavedMembers`. If needed, the challenger can even pass the entire `Dkg.Result`. We just need to ensure events emitted during `submitDkgResult` have all the information required so that the challenger can assemble parameters. Then, in `challengeDkgResult`, we hash everything, and if the hash matches, use values provided by the challenger.

pdyraga (approved):
`main` now:
```
···························|·······························································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  approveDkgResult                                             ·     121950  ·     171117  ·     131209  ·           72  ·          -  │
···························|·······························································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  challengeDkgResult                                           ·      68986  ·      82666  ·      81653  ·           27  ·          -  │
···························|·······························································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  submitDkgResult                                              ·     593915  ·     947537  ·     845238  ·          126  ·          -  │
···························|·······························································|·············|·············|·············|···············|··············
```

with this change:
```
···························|································································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  approveDkgResult                                              ·     203923  ·     252442  ·     213151  ·           72  ·          -  │
···························|································································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  challengeDkgResult                                            ·     987440  ·    1004540  ·     992580  ·           35  ·          -  │
···························|································································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  submitDkgResult                                               ·     731782  ·    1057792  ·     996681  ·          130  ·          -  │
···························|································································|·············|·············|·············|···············|··············
```

151443 difference for `submitDkgResult` is a lot. I do not know what is the reason but I think computing the hash of `Result`, especially `signatures` could cost a lot. I would consider extracting `signatures` out of the `Result` and providing them to `submitDkgResult` separately given that I think they are not needed in any other function, are already validated, and do not have to be included in the result hash.

Let's also update this doc:
```
// `\x19Ethereum signed message:\n${keccak256(groupPubKey,misbehaved,startBlock)}`
bytes signatures;
```
as it is no longer up to date.

I think this change is good and needs some optimization love later. Merging to unblock everything else.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -110,13 +108,15 @@ library DKG {
...
     event DkgResultSubmitted(
         bytes32 indexed resultHash,
+        uint256 indexed seed,
+        uint256 submitterMemberIndex,
         bytes indexed groupPubKey,
-        address indexed submitter
+        uint8[] misbehavedMembersIndices,
+        bytes signatures,
+        uint256[] signingMembersIndices,
+        uint32[] members
```
pdyraga, line 119: Let's also add a note that the final order of members in the group might be different after DKG completes. See `Groups.setGroupMembers`. Clients should `getGroup` after DKG to see what is their final group member ID in that group.

solidity/random-beacon/.solhintignore
```diff
@@ -1 +1,2 @@
-node_modules/
\ No newline at end of file
+node_modules/
+contracts/hardhat-dependency-compiler/
```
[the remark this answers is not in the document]
pdyraga: Done in https://github.com/keep-network/keep-core/pull/2734

solidity/random-beacon/contracts/libraries/DKG.sol
pdyraga: Done in https://github.com/keep-network/keep-core/pull/2734

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -1073,29 +1186,45 @@ describe("RandomBeacon - Group Creation", () => {
...
+
             await expect(
-              randomBeacon.connect(submitter).approveDkgResult()
+              randomBeacon.connect(submitter).approveDkgResult(stubDkgResult)
             ).to.be.revertedWith("current state is not CHALLENGE")
           })
         })
 
-        context("with dkg result submitted", async () => {
+        context("with malicious dkg result submitted", async () => {
```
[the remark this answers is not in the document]
pdyraga: Done in https://github.com/keep-network/keep-core/pull/2734

PR #2684, opened 2021-11-02
Used real sortition pools in tests

tomaszslabon:
This PR replaces `SortitionPoolStub` with the real `SortitionPool` in tests dealing with a lot of interactions between `RandomBeacon` and `SortitionPool`:

- function `randomBeaconDeployment` can now deploy either a stub version of `sortition-pools` or the real one (depending on the argument)
- made dependency `@keep-network/sortition-pools`  compile with the `hardhat-dependency-compiler` plugin
- used the real `sortition-pools` in tests `RandomBeacon.Pool.test.ts`

PR #2699, opened 2021-11-11
Use real sortition pools in DKG tests

nkuba:
This PR is a follow-up of https://github.com/keep-network/keep-core/pull/2684 to enable usage of the actual Sortition Pools implementation in Group Creation (DKG) tests.

It depends on some changes introduced in https://github.com/keep-network/keep-core/pull/2671/.

Closes https://keepnetwork.atlassian.net/browse/CP-152


tomaszslabon (commented):
@nkuba 
Looks good (we just need to remove the unused function `testDeployment`), but has conflicts. Could we solve them?

pdyraga (approved):

PR #2701, opened 2021-11-12
Update liquidity cards

r-czajkowski:
This PR updates cards on the Liquidity page.

Main changes:
- zero out KEEP-only pool APY,
- convert the TBTC/ETH pool to a `incentives removed` pool.


michalsmiarowski (approved):
LGTM 🚢

PR #2706, opened 2021-11-17
DKG: check stake of the submitter and operators

tomaszslabon:
#Depends on https://github.com/keep-network/keep-core/pull/2708
This PR adds a check of the DKG submitter's stake.

When a DKG result is submitted, the submitter's stake is compared with the `minimumAuhorization` governable parameter and if it's below it, the transaction is reverted.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -42,6 +42,15 @@ interface IRandomBeaconStaking {
         address notifier,
         address[] memory operators
     ) external;
+
+    function stakes(address operator)
+        external
+        view
+        returns (
+            uint96 tStake,
+            uint96 keepInTStake,
+            uint96 nuInTStake
+        );
 }
 
 /// @title Keep Random Beacon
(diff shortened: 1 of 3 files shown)


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -73,6 +73,15 @@ interface IRandomBeaconStaking {
...
     ) external;
+
+    function stakes(address operator)
+        external
+        view
+        returns (
+            uint96 tStake,
+            uint96 keepInTStake,
+            uint96 nuInTStake
+        );
```
pdyraga, line 84: T staking contract does not expose this function. We will probably have to consult the sortition pool about the authorization.

pdyraga (changes requested):

solidity/random-beacon/contracts/RandomBeacon.sol
tomaszslabon: It seems it does:
https://github.com/threshold-network/solidity-contracts/blob/32d5689e1e3c101c74f09102a5c5173d6aac7969/contracts/staking/TokenStaking.sol#L1052

pdyraga: The function exposed in T staking contract is to check the whole stake delegated, not the authorization the random beacon can slash. We will have to add such a function, this is part of https://github.com/keep-network/keep-core/issues/2715 work.

PR #2708, opened 2021-11-19
Governable minimum authorization amount

dimpar:
Adding functionality to update the minimum authorization amount.


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -146,6 +146,11 @@ contract RandomBeacon is Ownable {
     ///         operator affected.
     uint256 public relayEntryTimeoutNotificationRewardMultiplier;
 
+    /// @notice This amount is required to protect against griefing the staking
+    ///         contract and individual applications are allowed to require higher
+    ///         minimum stakes if necessary.
+    uint96 public minimumStake;
```
pdyraga, line 152: I think we should call it `minimumAuthorization` to be precise. `RandomBeacon`, from T staking contract perspective, is an application that will have to implement `authorizationIncreased`, `authorizationDecreaseRequested`, `involuntaryAuthorizationDecrease` [from `IApplication`](https://github.com/threshold-network/solidity-contracts/blob/main/contracts/staking/IApplication.sol). 

Also, from [T staking contract RFC](https://github.com/threshold-network/solidity-contracts/blob/main/docs/rfc-1-staking-contract.adoc#authorizing-an-application-1):

> The application being authorized is notified via a callback every single time new authorization happens and the application may decide to revert the transaction if individual requirements of the application are not met. For example, application may require a certain minimum authorized amount.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -277,6 +282,8 @@ contract RandomBeacon is Ownable {
         // TODO: Revisit if initial value of 2 weeks is enough.
         sortitionPoolRewardsBanDuration = 2 weeks;
         relayEntryTimeoutNotificationRewardMultiplier = 5;
+        // TODO: Revisit the initial value.
+        minimumStake = 100e3;
```
pdyraga, line 286: Let's set it to 100k T; we need e18 here.

solidity/random-beacon/contracts/RandomBeaconGovernance.sol
```diff
@@ -808,6 +814,38 @@ contract RandomBeaconGovernance is Ownable {
...
+
+    /// @notice Finalizes the minimum stake amount update process.
+    /// @dev Can be called only by the contract owner, after the governance
+    ///      delay elapses.
+    function finalizeMinimumStakeUpdate()
+        external
+        onlyOwner
+        onlyAfterGovernanceDelay(
+            minimumStakeChangeInitiated,
+            STANDARD_PARAMETER_GOVERNANCE_DELAY
```
pdyraga, line 839: I think we should consider it a critical parameter. A lower minimum stake means one can get more seats in the group and slashing amount for not submitting entry on time will be much higher.

pdyraga: Let's also remember to list this parameter on the list above `CRITICAL_PARAMETER_GOVERNANCE_DELAY`.

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: Sure, done in 9b2d771

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: 7325bfd

solidity/random-beacon/contracts/RandomBeaconGovernance.sol
dimpar: 3c6c981

> Let's also remember to list this parameter on the list above CRITICAL_PARAMETER_GOVERNANCE_DELAY

Sorry, not sure what you mean by that

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -477,6 +484,11 @@ contract RandomBeacon is Ownable {
         }
     }
 
+    /// @notice Updates the minimum authorization amount.
+    function updateMinimumAuthorization(uint96 minAuthorization) external {
```
pdyraga, line 488: `onlyOwner` is missing

pdyraga: We should add a test for this function to `RandomBeacon.Parameters.test.ts`: revert + update.

pdyraga: I'd also put them next to other update functions (`updateRelayEntryParameters`, `updateGroupCreationParameters`, `updateDkgParameters`, `updateRewardParameters`). Maybe right after ctor? It feels like a major `update*` function.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -146,6 +146,11 @@ contract RandomBeacon is Ownable {
     ///         operator affected.
     uint256 public relayEntryTimeoutNotificationRewardMultiplier;
 
+    /// @notice This amount is required to protect against griefing the staking
+    ///         contract and individual applications are allowed to require higher
+    ///         minimum authorization if necessary.
```
pdyraga, line 151: Hmm... I could have missed something but how does it protect against griefing the staking contract? 🤔 

Minimum stake in T staking contract does protect against it because it ensures no dust stakes can be used to block operator addresses: imagine someone trying to stake and some malicious party frontrunning them with 1e-18 T. But I don't see how this value protects against it. It's just a minimum authorization we require and it needs to be high enough so that slashing for malicious DKG result or relay entry timeout can be executed - worth mentioning it in docs.

solidity/random-beacon/contracts/RandomBeaconGovernance.sol
pdyraga: This list: https://github.com/keep-network/keep-core/blob/3c6c9817deae42caf5edd94a29565b3bd7673015/solidity/random-beacon/contracts/RandomBeaconGovernance.sol#L77-L81

dimpar: ok, done in a4541b7

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: ff369a8

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -290,6 +297,14 @@ contract RandomBeacon is Ownable {
         relay.setRelayEntrySubmissionFailureSlashingAmount(1000e18);
     }
 
+    /// @notice Updates the minimum authorization amount.
+    function updateMinimumAuthorization(uint96 minAuthorization)
```
pdyraga, line 300: I'd rather use `_minimumAuthorization` as a parameter name. That's what we do when we have naming conflict in the rest of `update*` functions and in the constructor.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -290,6 +297,14 @@ contract RandomBeacon is Ownable {
...
+        external
+        onlyOwner
+    {
+        minimumAuthorization = minAuthorization;
+    }
```
pdyraga, line 305: We should emit an event and assert it is emitted in `updateMinimumAuthorization` context in `RandomBeacon.Parameters.test.ts`.

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: changed in 36d98c5

I looked at T Staking contract and based my prev description on it. But yes, I see now that it's different.

solidity/random-beacon/test/RandomBeaconGovernance.test.ts
```diff
@@ -1963,4 +1968,119 @@ describe("RandomBeaconGovernance", () => {
...
+      })
+    })
+
+    context("when the governance delay has not passed", () => {
+      it("should revert", async () => {
+        await randomBeaconGovernance
+          .connect(governance)
+          .beginMinimumAuthorizationUpdate(123)
+
+        await helpers.time.increaseTime(24 * 13 * 60 * 60) // 13 days
```
pdyraga, line 2041: Nitpicking but I'd do `13 * 24 * 60 * 60` for consistency with the rest of 13 days assertions and given that 13 at the beginning is IMO easier to follow.

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: I'm a bit confused actually, who can call this function? I thought it's governance through the RandomBeaconGovernance contract. If yes, won't we duplicate the event which is in `beginMinimumAuthorizationUpdate` and later in `MinimumAuthorizationUpdated`?

pdyraga: We may change the owner one day and we still want to be able to locate when the given parameter changed. That's why emit another event in `RandomBeacon`.

dimpar: fd5ff63

solidity/random-beacon/test/RandomBeaconGovernance.test.ts
dimpar: 8c14b8b

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: 8c14b8b

pdyraga (approved):

PR #2709, opened 2021-11-22
Allow anyone to approve the DKG result if the submitter is late

tomaszslabon:
This PR modifies the random beacon so that anyone can call the `approveDkgResult` function and receive the DKG submission reward if the submitter does not do that on time.

The submitter has `resultSubmissionEligibilityDelay` blocks after the end of the challenge period to approve the result.
If they are late, anyone can approve the result and receive the reward.  

Additional modification: when the balance of the maintenance pool is not enough, the reward submitter receives whatever tokens there are on the maintenance pool.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -22,6 +22,7 @@ import "./libraries/Groups.sol";
 import "./libraries/Callback.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/utils/math/Math.sol";
 
 /// @title Sortition Pool contract interface
 /// @notice This is an interface with just a few function signatures of the
@@ -87,8 +88,8 @@ contract RandomBeacon is Ownable {
     using Groups for Groups.Data;
     using Relay for Relay.Data;
     using Callback for Callback.Data;
-    using SafeERC20 for IERC20;
     using GasStation for GasStation.Data;
+    using SafeERC20 for IERC20;
 
     // Constant parameters
 
@@ -532,14 +533,23 @@ contract RandomBeacon is Ownable {
         dkg.complete();
     }
 
-    /// @notice Approves DKG result. Can be called after challenge period for the
-    ///         submitted result is finished. Considers the submitted result as
-    ///         valid, pays reward to the result submitter, bans misbehaved group
-    ///         members from the sortition pool rewards and completes the group
-    ///         creation by activating the candidate group.
+    /// @notice Approves DKG result. Can be called after challenge period for
+    ///         the submitted result is finished. For the first
+    ///         `resultSubmissionEligibilityDelay` blocks after the end of the
+    ///         challenge period can be called only by the DKG result submitter.
+    ///         After that can be called by anyone. Considers the submitted
+    ///         result as valid, pays reward to the approver, bans misbehaved
+    ///         group members from the sortition pool rewards and completes the
+    ///         group creation by activating the candidate group.
     function approveDkgResult() external {
         dkg.approveResult();
-        tToken.safeTransfer(dkg.resultSubmitter, dkgResultSubmissionReward);
+
+        uint256 maintenancePoolBalance = tToken.balanceOf(address(this));
+        uint256 rewardToPay = Math.min(
+            maintenancePoolBalance,
+            dkgResultSubmissionReward
+        );
+        tToken.safeTransfer(msg.sender, rewardToPay);
 
         if (dkg.submittedResultMisbehavedMembers.length > 0) {
             banFromRewards(
@@ -550,14 +560,8 @@ contract RandomBeacon is Ownable {
 
         groups.activateCandidateGroup();
         dkg.complete();
-
-        // TODO: Should result submitter receive reward if they failed to call
-        //       this function?
-        // TODO: Ensure this function is as cheap as possible and it is
-        //       profitable for the DKG result submitter to call it. Consider
-        //       giving the submitter some time to approve the result and if
-        //       don't, pay the submitter's reward to anybody who calls this
-        //       function.
+        // TODO: Check if this function is cheap enough and it will be
+        //       profitable for the DKG result submitter to call it.
         // TODO: Unlock sortition pool
     }
 
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -533,14 +533,14 @@ contract RandomBeacon is Ownable {
         dkg.complete();
     }
 
-    /// @notice Approves DKG result. Can be called after challenge period for
-    ///         the submitted result is finished. For the first
+    /// @notice Approves DKG result. Can be called when the challenge period for
+    ///         the submitted result is finished. Considers the submitted result
+    ///         as valid, pays reward to the approver, bans misbehaved group
+    ///         members from the sortition pool rewards, and completes the group
+    ///         creation by activating the candidate group. For the first
     ///         `resultSubmissionEligibilityDelay` blocks after the end of the
     ///         challenge period can be called only by the DKG result submitter.
-    ///         After that can be called by anyone. Considers the submitted
-    ///         result as valid, pays reward to the approver, bans misbehaved
-    ///         group members from the sortition pool rewards and completes the
-    ///         group creation by activating the candidate group.
+    ///         After that time, can be called by anyone.
     function approveDkgResult() external {
         dkg.approveResult();
 
--- a/solidity/random-beacon/contracts/libraries/DKG.sol
+++ b/solidity/random-beacon/contracts/libraries/DKG.sol
@@ -394,12 +394,12 @@ library DKG {
         self.sortitionPool.unlock();
     }
 
-    /// @notice Approves DKG result. Can be called after challenge period for
-    ///         the submitted result is finished. For the first
-    ///         `resultSubmissionEligibilityDelay` blocks after the end of the
-    ///         challenge period can be called only by the DKG result submitter.
-    ///         After that can be called by anyone. Considers the submitted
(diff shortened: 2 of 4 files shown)


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -532,14 +533,23 @@ contract RandomBeacon is Ownable {
...
-    ///         members from the sortition pool rewards and completes the group
-    ///         creation by activating the candidate group.
+    /// @notice Approves DKG result. Can be called after challenge period for
+    ///         the submitted result is finished. For the first
+    ///         `resultSubmissionEligibilityDelay` blocks after the end of the
+    ///         challenge period can be called only by the DKG result submitter.
+    ///         After that can be called by anyone. Considers the submitted
+    ///         result as valid, pays reward to the approver, bans misbehaved
+    ///         group members from the sortition pool rewards and completes the
+    ///         group creation by activating the candidate group.
```
pdyraga, line 543: Let's rejigger this doc a little to first focus on the effect and then describe authorization rules:

> Approves DKG result. Can be called when the challenge period for the submitted result is finished. Considers the submitted result as valid, pays reward to the approver, bans misbehaved group members from the sortition pool rewards, and completes the group creation by activating the candidate group. For the first `resultSubmissionEligibilityDelay` blocks after the end of the challenge period can be called only by the DKG result submitter. After that time, can be called by anyone. Caller of this function receives DKG result submission reward.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -550,14 +560,8 @@ contract RandomBeacon is Ownable {
...
-
-        // TODO: Should result submitter receive reward if they failed to call
-        //       this function?
-        // TODO: Ensure this function is as cheap as possible and it is
-        //       profitable for the DKG result submitter to call it. Consider
-        //       giving the submitter some time to approve the result and if
-        //       don't, pay the submitter's reward to anybody who calls this
-        //       function.
+        // TODO: Check if this function is cheap enough and it will be
+        //       profitable for the DKG result submitter to call it.
```
pdyraga, line 564: What is the gas cost of `submitDkgResult` and `approveDkgResult`? Maybe it is something we should no longer be concerned about and the only challenge is to set `dkgResultSubmissionReward` high enough.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -394,9 +394,12 @@ library DKG {
...
 
-    /// @notice Approves DKG result. Can be called after challenge period for the
-    ///         submitted result is finished. Considers the submitted result as
-    ///         valid and completes the group creation.
+    /// @notice Approves DKG result. Can be called after challenge period for
+    ///         the submitted result is finished. For the first
+    ///         `resultSubmissionEligibilityDelay` blocks after the end of the
+    ///         challenge period can be called only by the DKG result submitter.
+    ///         After that can be called by anyone. Considers the submitted
+    ///         result as valid and completes the group creation.
```
pdyraga, line 402: Just like for its `RandomBeacon` contract counterpart, let's reorganize documentation to first focus on the effect of calling this function and then on authorization rules:

> Approves DKG result. Can be called when the challenge period for the submitted result is finished. Considers the submitted result as valid, pays reward to the approver, bans misbehaved group members from the sortition pool rewards, and completes the group creation by activating the candidate group. For the first `resultSubmissionEligibilityDelay` blocks after the end of the challenge period can be called only by the DKG result submitter. After that time, can be called by anyone.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -410,6 +413,15 @@ library DKG {
...
         );
 
+        require(
+            msg.sender == self.resultSubmitter ||
+                block.number >
+                self.submittedResultBlock +
+                    self.parameters.resultChallengePeriodLength +
+                    self.parameters.resultSubmissionEligibilityDelay,
+            "Only the DKG result submitter can approve the result at this moment"
+        );
```
pdyraga, line 423: Nitpicking be we could extract `self.submittedResultBlock + self.parameters.resultChallengePeriodLength` into a variable. I think it should be a bit cheaper in case the optimizer does not take care about it:

```
uint256 challengePeriodEnd = self.submittedResultBlock + self.parameters.resultChallengePeriodLength;

require(block.number > challengePeriodEnd, ...);

require(msg.sender == self.resultSubmitter || block.number > challengePeriodEnd + self.parameters.resultSubmissionEligibilityDelay, ...);

File: solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts [hunk 5]
```diff
@@ -245,7 +247,7 @@ describe("RandomBeacon - Group Creation", () => {
           context("when dkg result was approved", async () => {
             beforeEach(async () => {
               await mineBlocks(params.dkgResultChallengePeriodLength)
-              await randomBeacon.approveDkgResult()
+              await randomBeacon.connect(thirdParty).approveDkgResult()
```
pdyraga, line 250: I think we messed up roles in this test and it's time to clear them up. We use `thirdParty` as a role of DKG result submitter and as a role of some random address, outside of the signing group that is e.g. challenging DKG result. In fact, `thirdParty` is the signing group member and DKG result submitter so tests are checking something else than what we think they do. 

See that:
```
const signers = await registerOperators(
  contracts.randomBeacon as RandomBeacon,
  (await getUnnamedAccounts()).slice(1, 1 + constants.groupSize)
)
```

but later:
```
thirdParty = await ethers.getSigner((await getUnnamedAccounts())[1])
```

What I think should happen is:
```
const signers = await registerOperators(
  contracts.randomBeacon as RandomBeacon,
  (await getUnnamedAccounts()).slice(1, 1 + constants.groupSize)
)

// (...)

thirdParty = await ethers.getSigner((await getUnnamedAccounts())[0])
submitter = await ethers.getSigner((await getUnnamedAccounts())[1])
notSubmitter = await ethers.getSigner((await getUnnamedAccounts())[55])
```

and then we need to inspect all test cases where `thirdParty` is used and either leave it there or replace it with `submitter`. For example, in `approveDkgResult` tests, the scenario of `with challenge period passed -> called by a third party` does not really use a third party.

cc @nkuba @dimpar @lukasz-zimnoch for potential merge conflict resolution

solidity/random-beacon/test/utils/groups.ts
```diff
@@ -13,14 +13,19 @@ export async function createGroup(
...
     startBlock,
-    noMisbehaved
+    noMisbehaved,
+    submitterIndex
   )
+
   await mineBlocks(params.dkgResultChallengePeriodLength)
-  await randomBeacon.approveDkgResult()
+  await randomBeacon
+    .connect(await ethers.getSigner(signers[submitterIndex - 1].address))
```
pdyraga, line 29: Not in this PR but thoughts on adding signer to `Operator` struct:
https://github.com/keep-network/keep-core/blob/fcfb88fd1352e9defd549c21b2809cba4d98ee29/solidity/random-beacon/test/utils/sortitionpool.ts#L9
?
cc @nkuba 

I see we often have to use `ethers.getSigner` in our tests.

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -1319,7 +1321,125 @@ describe("RandomBeacon - Group Creation", () => {
           await assertDkgResultCleanData(randomBeacon)
         })
       })
+
+      context("when the approver is different than the submitter", async () => {
```
pdyraga, line 1325: I'd consider moving these tests under `with off-chain dkg time passed -> with dkg result submitted -> wth challenge period passed` context. Note that we already have there `called by a third party` block of unit tests. We really need to have there two: `when called by the DKG result submitter` and `when called by a third party`.

pdyraga (changes requested):

solidity/random-beacon/contracts/RandomBeacon.sol
tomaszslabon: Done in https://github.com/keep-network/keep-core/pull/2709/commits/2c0a29290112e95b5921f6456cb4adb3e98846d6.

solidity/random-beacon/contracts/libraries/DKG.sol
tomaszslabon: Done in https://github.com/keep-network/keep-core/pull/2709/commits/2c0a29290112e95b5921f6456cb4adb3e98846d6.

solidity/random-beacon/contracts/libraries/DKG.sol
tomaszslabon: Done in https://github.com/keep-network/keep-core/pull/2709/commits/c226e3798c730dd504a46897d6aeb74d265a4e54.

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
tomaszslabon: Done in https://github.com/keep-network/keep-core/pull/2709/commits/ca694fea2c0632d082454df62c51df640d3bb23d.

solidity/random-beacon/test/utils/groups.ts
tomaszslabon: I agree, we could add `signer` to the `Operator` structure.
Should we work on it now, or add it to the backlog, so we can do this once we have all the solidity tasks done?

pdyraga: Let's backlog it, it's definitely nothing on the top of our priority list.

tomaszslabon: Done

pdyraga (approved):

solidity/random-beacon/contracts/RandomBeacon.sol
pdyraga: Awesome. Then I do not think we need to worry about it. We can keep that TODO and double-check once we finish all the other work, no need to block this PR on it.

Issue #2715, opened 2021-11-22
Integrate T staking contract with RandomBeacon


pdyraga:
T staking contract works differently than KEEP staking contract in terms of the eligible stake. Every application authorized by T DAO can have its separate risk ceiling set by the stake authorizer. The application - which is `RandomBeacon` contract in our case - controls what is the minimum authorization and when it can be decreased.

`RandomBeacon` needs to implement [`IApplication` interface](https://github.com/threshold-network/solidity-contracts/blob/main/contracts/staking/IApplication.sol) and update `SortitionPool` every time authorization increases and decreases. `SortitionPool` does no longer need to consult staking contract and `RandomBeacon` - which is the contract owner - can provide the authorized stake amount. The relationship should be: `T TokenStaking <-> RandomBeacon <-> SortitionPool`.

One important issue that needs to be taken into account is how to make malicious players' lives harder when they try to earn rewards but are not responsive in the network. https://github.com/keep-network/sortition-pools/pull/130 assumes all sortition pool members are getting rewards and the scenario we need to avoid is when someone joins the pool, earns rewards, then gets selected to work and does nothing; as a result, they are banned from rewards for the next few weeks but they can exit the pool, change operator address and start earning rewards again. 

The simplest solution seems to be an introduction of a governable stake deauthorization delay. Deauthorization delay should be longer or at least equal to the time for which inactive operator is banned from sortition pool rewards.

Deauthorization delay also solves problems with group expiration: operator needs to stay around and respond for at least the time the group they are a member of is active.

PR #2719, opened 2021-11-23
Added governable authorization decrease delay parameter

pdyraga:
Refs #2715

This parameter says what delay needs to pass between the time
authorization decrease is requested and the time that request
gets approved. It will protect against free-riders earning rewards
and not being active in the network.

Moved `minimumAuthorization` parameter to `Authorization` library
next to `authorizationDecreaseDelay`. The library will contain
all functions related to updating authorization and sortition pool.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeaconGovernance.sol
+++ b/solidity/random-beacon/contracts/RandomBeaconGovernance.sol
@@ -848,7 +848,7 @@ contract RandomBeaconGovernance is Ownable {
     {
         emit MinimumAuthorizationUpdated(newMinimumAuthorization);
         // slither-disable-next-line reentrancy-no-eth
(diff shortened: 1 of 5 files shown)


solidity/random-beacon/contracts/RandomBeaconGovernance.sol
```diff
@@ -1047,6 +1097,18 @@ contract RandomBeaconGovernance is Ownable {
             );
     }
 
+    function getRemainingAuthorizationDecreaseDelayUpdateTime()
```
tomaszslabon, line 1100: Minor thing, could we add a description to this function if the other functions have it?

pdyraga: Good point but please let me address it in a followup PR - I will have one tomorrow and don't want to rebase everything again 😬

PR #2720, opened 2021-11-24
Expect 33 signatures on the DKG result

pdyraga:
Instead of expecting 48 signatures on the result we now expect just
33 which is the group threshold. Also, we expect that at least 90%
of the group members were active during the DKG. This change allows
to go down with DKG `submitDKGResult` gas costs. For 64 signatures,
the average gas cost is 1124329. For 33 signatures the average gas
cost is 815350.

Expecting more signatures than 33 does not make the protocol any more
secure. Even if we have 64 signatures supporting the result, an
adversary must not corrupt 33+1 members.

### DIFF
--- a/solidity/random-beacon/contracts/libraries/DKG.sol
+++ b/solidity/random-beacon/contracts/libraries/DKG.sol
@@ -91,16 +91,18 @@ library DKG {
     /// @dev Size of a group in the threshold relay.
     uint256 public constant groupSize = 64;
 
-    /// @dev Minimum number of group members needed to interact according to the
-    ///      protocol to produce a relay entry.
+    /// @dev The minimum number of group members needed to interact according to
+    ///      the protocol to produce a relay entry. The adversary can not learn
+    ///      anything about the key as long as it does not break into
+    ///      groupThreshold+1 of members.
     uint256 public constant groupThreshold = 33;
 
-    /// @dev The minimum number of signatures required to support DKG result.
-    ///      This number needs to be at least the same as the signing threshold
-    ///      and it is recommended to make it higher than the signing threshold
-    ///      to keep a safety margin for in case some members become inactive.
-    uint256 public constant signatureThreshold =
-        groupThreshold + (groupSize - groupThreshold) / 2;
+    /// @dev The minimum number of active and properly behaving group members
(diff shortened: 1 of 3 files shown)


lukasz-zimnoch (approved):

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -91,16 +91,18 @@ library DKG {
     /// @dev Size of a group in the threshold relay.
     uint256 public constant groupSize = 64;
 
-    /// @dev Minimum number of group members needed to interact according to the
-    ///      protocol to produce a relay entry.
+    /// @dev The minimum number of group members needed to interact according to
+    ///      the protocol to produce a relay entry. The adversary can not learn
+    ///      anything about the key as long as it does not break into
+    ///      groupThreshold+1 of members.
     uint256 public constant groupThreshold = 33;
```
nkuba, line 98: Shouldn't we leave a constant named `signatureThreshold` but just change its value to 33? We could remove `groupThreshold` constant then.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -91,16 +91,18 @@ library DKG {
     /// @dev Size of a group in the threshold relay.
     uint256 public constant groupSize = 64;
 
-    /// @dev Minimum number of group members needed to interact according to the
-    ///      protocol to produce a relay entry.
+    /// @dev The minimum number of group members needed to interact according to
+    ///      the protocol to produce a relay entry. The adversary can not learn
```
nkuba, line 95: The doc is a little bit confusing as it mentions `relay entry` but is used for DKG result signatures number check.

```
    /// @dev The minimum number of group members needed to interact according to
    ///      the protocol to provide signatures for the DKG result. The adversary can not learn (...)

nkuba (commented):
Better late than never 😉

solidity/random-beacon/contracts/libraries/DKG.sol
nkuba: 498949806b6f01ec9320d377ce1de1657e592080

PR #2721, opened 2021-11-24
Use @keep-network/hardhat-helpers

nkuba:
Common utils functions are defined in @keep-network/hardhat-helpers
package and used across our codebase. Here we switch to use these
helpers instead duplicating the code.


pdyraga (approved):
Yes, please.

Issue #2725, opened 2021-11-24
Separate function to inform about failed heartbeat


pdyraga:
The original idea presented in [RFC-19](https://github.com/keep-network/keep-core/blob/main/docs/rfc/rfc-19-random-beacon-v2.adoc) assumes inactive and misbehaving operators are marked as ineligible for rewards in the same transaction in which DKG result or relay entry is submitted. Although this approach is an ideal one, the problem is the gas cost. First, checking the eligibility adds some gas to the transaction. Second, the function marking operators as ineligible costs between 55122-244621 with 118288 on average as implemented in https://github.com/keep-network/sortition-pools/pull/130, and the entire operation marking operators as ineligible will be even more expensive given that it is needed to compute who should be marked as such and transform from IDs to addresses. This will become even more problematic in tBTC v2 when we will have to operate on groups much larger than in the random beacon, so the worst-case scenario is even more expensive.

We attempted to mitigate the cost by introducing a gas station pattern and requiring every operator joining the pool to make a deposit. This approach is not enough though - one operator being in the pool all the time can be marked as ineligible for rewards several times without re-filling the deposit. Also, the gas cost overhead of computing ineligible members is still high.

In tBTC v2 we will have an additional mechanism of heartbeats (https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#26-heartbeats): if the group failed a heartbeat, signers can sign a message saying who was inactive during the heartbeat and these folks will be eliminated from rewards.

The proposition is:
- have a separate pool for DKG submitter rewards, funded from relay request fees
- have a separate pool of notifier rewards, funded by DAO (in the end, the DAO saves money on ineligible operators)
- we should check eligibility of signer in DKG submit transaction and ensure the reward is high enough so that DKG result submitter is a winner and not a looser
- we should mark as ineligible for rewards those who were IA/DQ during off-chain DKG in the approve result tx. We allow for 8 misbehaving operators maximum, the cost of marking as ineligible should be around 10k each.
- we do not care about eligibility for relay entry at all
- we add a function allowing a group to report failed heartbeat; notifier will be rewarded from the notifier rewards pool
- no single operator can call this function on their own - enough signers from the group need to agree it makes sense to call it
- we may add heartbeats to the random beacon and mark people as ineligible if they fail to participate in signing (e.g. disabled their node for a long time) but do nothing with the group
[...]

PR #2729, opened 2021-11-25
Remove relay entry eligibility checks

lukasz-zimnoch:
Refs: #2725
Depends on: https://github.com/keep-network/keep-core/pull/2726

This pull request removes the current eligibility check from `submitRelayEntry` along with the code responsible for banning operators who missed their submission turn. This is because those checks won't be needed with the new approach using `notifyFailedHeartbeat`.


solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -70,7 +82,15 @@ library Relay {
...
 
-    event RelayEntrySubmitted(uint256 indexed requestId, bytes entry);
+    event RelayEntrySubmitted(
+        uint256 indexed requestId,
+        bytes entry,
+        uint256 submissionBlock,
+        uint256 eligibilityDelay,
+        uint256 requestStartBlock,
+        uint256 submitterIndex,
+        uint32[] groupMembers
```
pdyraga, line 92: The cost of emitting event with this field will be significant. Given that the function reporting inactivity can only be called before the next request comes in, I think we may not need all fields in this event. Some information is available on-chain.

lukasz-zimnoch: Fetching that on-chain is even more expensive.

lukasz-zimnoch: > Given that the function reporting inactivity can only be called before the next request comes in

That's not true. According to what we discussed, this function can be called until the next relay entry is submitted, not just the request.

pdyraga: > Fetching that on-chain is even more expensive.

I was thinking about fetching it off-chain and providing to the notify function but I haven't had a chance to look at the rest of the code so unsure if this is feasible and what are the consequences.

lukasz-zimnoch: Yeah in that case I think we can just emit `groupId`. I'll check that.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -245,6 +245,8 @@ contract RandomBeacon is Ownable {
 
     event CallbackFailed(uint256 entry, uint256 entrySubmittedBlock);
 
+    event RewardsBanned(uint32[] ids, uint256 duration);
+
     event BanRewardsFailed(uint32[] ids);
```
pdyraga, line 250: Let's drop these two events. We should assume sortition pool never reverts on marking operators ineligible for rewards and even if it does, it is fine to revert DKG submit transaction because not submitting the result does not lead to slashing.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -775,10 +777,12 @@ contract RandomBeacon is Ownable {
     /// @param ids IDs of banned operators.
     /// @param banDuration Duration of the ban period in seconds.
     function banFromRewards(uint32[] memory ids, uint256 banDuration) internal {
```
pdyraga, line 779: `approveDkgResult` is the only place where this function is called. Can we get rid of `banFromRewards` and just call `sortitionPool.banRewards(ids, banDuration)`, without try-catch, from `submitDkgResult` directly?

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -59,14 +60,13 @@ describe("RandomBeacon - Relay", () => {
   // In the invalid entry scenario `(blsData.nextGroupSignature % 64) + 1`
   // gives 3 so that  member needs to submit the wrong relay entry.
   const invalidEntryFirstEligibleMemberIndex = 3
+  const ineligibleOperatorNotifierReward = to1e18(200)
```
pdyraga, line 63: No longer used.

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -253,262 +255,163 @@ describe("RandomBeacon - Relay", () => {
   })
 
   describe("submitRelayEntry", () => {
+    let requestStartBlock: number
```
pdyraga, line 258: I think it is no longer used.

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -253,262 +255,163 @@ describe("RandomBeacon - Relay", () => {
...
-                      .withArgs(1, blsData.groupSignature)
-                  })
-
-                  it("should terminate the relay request", async () => {
-                    expect(await randomBeacon.isRelayRequestInProgress()).to.be
-                      .false
-                  })
-                }
-              )
+                receipt = await tx.wait()
```
pdyraga, line 289: Looks like no longer used.

pdyraga (changes requested):

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2729/commits/ef1075cc3260578686539f43fba5bd11fc6b5a0a

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2729/commits/ef1075cc3260578686539f43fba5bd11fc6b5a0a

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2729/commits/ef1075cc3260578686539f43fba5bd11fc6b5a0a

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2729/commits/ef1075cc3260578686539f43fba5bd11fc6b5a0a

solidity/random-beacon/contracts/RandomBeacon.sol
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/2729/commits/ef1075cc3260578686539f43fba5bd11fc6b5a0a

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -11,6 +11,7 @@ import { expect } from "chai"
 import { BigNumber, ContractTransaction } from "ethers"
 import type { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers"
 import type { Address } from "hardhat-deploy/types"
+import { describe } from "mocha"
```
pdyraga, line 14: Is it needed?

PR #2731, opened 2021-11-25
Update eligible submitter index counting for DKG

nkuba:
The eligible submitter index should be counted starting with a variable index. Up to now, it was always starting with the first member.

The way for calculating the eligible submitter was introduced to the Relay entry library in https://github.com/keep-network/keep-core/pull/2637 and later exposed to a separate library in
https://github.com/keep-network/keep-core/pull/2726.

Refs https://github.com/keep-network/keep-core/issues/2725

Note that this PR changes the way the eligible member index is calculated. Storing and handling inactive members will be implemented in a separate PR.


solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -205,6 +205,24 @@ library DKG {
         );
         require(!hasDkgTimedOut(self), "dkg timeout already passed");
 
+        (uint256 firstEligibleIndex, uint256 lastEligibleIndex) = Submission
+            .getEligibilityRange(
+                self.seed,
```
lukasz-zimnoch, line 210: According to the RFC, the "seed" for eligibility range calculations should be the hash of new group public key (see https://github.com/keep-network/keep-core/blob/main/docs/rfc/rfc-19-random-beacon-v2.adoc#22-fees-and-rewards) not the seed of the DKG itself. Parameter name in `getEligibilityRange` is a little bit confusing.

lukasz-zimnoch: Try/catch works only for failed external calls and contract creation calls (source: https://docs.soliditylang.org/en/latest/control-structures.html#try-catch)

nkuba: Nice catch! 6bf451d14b890020408dd3a06873a5950a845f47

pdyraga:
Once potential conflicts are resolved, can we compare gas usage before/after?

nkuba:
> Once potential conflicts are resolved, can we compare gas usage before/after?

Before:
```
|  RandomBeaconStub        ·  submitDkgResult                             ·     361842  ·     636556  ·     570165  ·           30  ·          -  │
```

After:
```
|  RandomBeaconStub        ·  submitDkgResult                             ·     362681  ·     637395  ·     547094  ·           53  ·          -  │
```

pdyraga:
@nkuba May I suggest we merge #2471 first? #2471 is pretty small and will be able to use real `sortitionPool.setRewardIneligibility` in this PR then.

lukasz-zimnoch:
@nkuba Can you solve conflicts? I want to do a final pass here.

lukasz-zimnoch (approved):

PR #2733, opened 2021-11-26
Clarified there can only one candidate group - description and tests

tomaszslabon:
This PR:
- adds a description to the `addCandidateGroup` function that clarifies we can have at most one candidate (not activated) group.
- fixes some typos
- updates confusing test descriptions suggesting we can have more than one candidate group
- removes tests where there are two candidate groups (the library allows it, but the code using the library does not).  

Note: `addCandidateGroup` is called in `submitDkgResult`.
[Here](https://github.com/keep-network/keep-core/blob/001ca97953b23a16954e6ebae26f7e07e2c00532/solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts#L959) is the test checking we cannot `submitDkgResult` when there is another one in progress.

### DIFF
--- a/solidity/random-beacon/test/Groups.test.ts
+++ b/solidity/random-beacon/test/Groups.test.ts
@@ -549,7 +549,7 @@ describe("Groups", () => {
         const groupPublicKey1 = "0x0001"
         const groupPublicKey2 = "0x0002"
 
-        context("when both groups are candidate", async () => {
+        context("when candidate group added after activation", async () => {
           let tx: ContractTransaction
 
           beforeEach(async () => {
@@ -652,7 +652,7 @@ describe("Groups", () => {
         const groupPublicKey1 = groupPublicKey
         const groupPublicKey2 = groupPublicKey
 
-        context("when both groups are candidate", async () => {
+        context("when the first group is active", async () => {
           let tx: ContractTransaction
 
           beforeEach(async () => {
@@ -722,13 +722,11 @@ describe("Groups", () => {
 
         it("should remove registered group", async () => {
           const groupsRegistry = await groups.getGroupsRegistry()
-
           expect(groupsRegistry).to.be.lengthOf(0)
         })
 
         it("should not update stored group data", async () => {
           const storedGroup = await groups.getGroup(groupPublicKey)
-
           expect(storedGroup.groupPubKey).to.be.equal(groupPublicKey)
           expect(storedGroup.activationBlockNumber).to.be.equal(0)
           expect(storedGroup.members).to.be.deep.equal(members)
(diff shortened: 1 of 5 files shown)


solidity/random-beacon/test/Groups.test.ts
```diff
@@ -761,57 +759,10 @@ describe("Groups", () => {
...
-        context("when both groups are candidate", async () => {
-          let tx: ContractTransaction
-
-          beforeEach(async () => {
-            await groups.addCandidateGroup(
-              groupPublicKey1,
-              members1,
-              noMisbehaved
-            )
-            await groups.addCandidateGroup(
```
tomaszslabon, line 773: This cannot happen.

solidity/random-beacon/test/Groups.test.ts
```diff
@@ -862,57 +813,6 @@ describe("Groups", () => {
...
```
tomaszslabon, line 879: Removed because this cannot happen.

solidity/random-beacon/test/Groups.test.ts
```diff
@@ -652,7 +652,7 @@ describe("Groups", () => {
         const groupPublicKey1 = groupPublicKey
         const groupPublicKey2 = groupPublicKey
 
-        context("when both groups are candidate", async () => {
+        context("when the first group is active", async () => {
```
pdyraga, line 655: I find the test cases in this context highly confusing. We are testing `activateCandidateGroup` and the scenario I believe we are dealing with is:
- group A added as candidate
- group A challenged
- group B added as candidate with the same pub key as A
- group B activated

The fact B overwrites A is checked in `addCandidateGroup` test, at line 408: `with the same group public key.
The fact we cannot add another candidate group when one with the same pub key is active is checked in `addCandidateGroup` test, at line 318: `with the same group public key`.

So here, I think what we want to check is if activation of B is possible. Something like:

when two groups are registered -> with the same group public key -> when the first group was challenged and replaced -> it should not be possible to activate the second group

tomaszslabon: So you want to test:
* add group A as a candidate
* pop group A
* call `activateCandidateGroup` (it fails)
?

pdyraga: - Add group A as candidate
- Pop group A
- Add group B as candidate (with the same pub key)
- Call activate on B and ensure it works

I think it's the only valid path for "when two groups are registered -> with the same group public key"

tomaszslabon: Done in https://github.com/keep-network/keep-core/commit/fc955c3c1376e50df055de8ba33b8906c04b8e1b.

I basically used this flow and checked if it succeeded:
```
              await groups.addCandidateGroup(
                groupPublicKey1,
                members1,
                noMisbehaved
              )

              await groups.popCandidateGroup()

              await groups.addCandidateGroup(
                groupPublicKey2,
                members2,
                noMisbehaved
              )

              tx = await groups.activateCandidateGroup()             
```

pdyraga (approved):

PR #2734, opened 2021-11-26
Moved DKG result signature validation to challengeDkgResult 

pdyraga:
The previous version was:

In `submitDkgResult` we:
- perform static validation of result (lengths, ranges, sizes)
- check if t+1 signatures signing the result are valid,
- check if all operators who signed the result are in the sortition pool,
- if signatures are not valid or at least one operator is not in the sortition
  pool, the result is rejected.

Then, there is some time to challenge the result and the result is not accepted
until the challenge period is over. The sortition pool is locked for the entire
challenge period.

In `challengeDkgResult` - called by anyone only when something is wrong with the
result - we:
- call sortition pool and check if group members passed in `submitDkgResult` are
  what the sortition pool returns,
- if the expected member list does not match the actual member list, DKG result
  is rejected and all operators who signed the result are slashed.

What we do now:

In `submitDkgResult`:
- check if the submitter is in the sortition pool.

In `challengeDkgResult`:
- perform static validation of the result (lengths, sizes, ranges)
- call sortition pool and check if group members passed in `submitDkgResult`
  are what the sortition pool expects,
- check if t+1 signatures signing the result are valid, and if they come from
  sortition pool members,
- if anything is wrong with the result, slash DKG result submitter.

The previous approach felt wrong for the number of reasons:
- It was expensive!!! More than 1M of gas for a group size of 64 members to
  `submitDkgResult`. This will only get worse for tBTC v2 and a group size of
  100 members. With the alternative approach proposed, the cost should be 
  on average less than 600k.
- Checking if someone is in the sortition pool in `submitDkgResult` is not
  enough. One operator can have multiple signer seats, so we should check the
  stake of the operator. If they provided `S` signatures, we should check that
  their stake is `S*min_stake`. Think of the case when one operator claims they
  have all 64 member seats and we just check if that operator is in the
  sortition pool... It does not make sense, we should check if they have
  `64*min_stake` but it only makes `submitDkgResult` more expensive...
- ...At the same time, we want the minimum stake to be governable and we want to
  be able to lower and increase it. In case the minimum stake was increased, it
  affects only new operators joining the network. Someone not having the minimum
  stake can still be in the sortition pool and be selected by the sortition pool.
  Checking the minimum stake breaks in this case.
- See that we are duplicating some work: in `submitDkgResult` we check if
  members are in the sortition pool and in `challengeDkgResult` we check if they
  were selected by the sortition pool. Challenge checks their sortition pool
  membership implicitly.

In the new approach, minor censorship seems to be the only problem.
We should be fine with accepting this risk if the challenge period is long
enough and the reward for challenging is hefty. We need to move into
submit-challenge-accept model if we want to lower the cost and if miners can
censor transactions for more than 6h, we are not the only ones who have a
problem and this feels like something that needs to be solved on the blockchain
level, not in our protocol. Of course this will all be governable so if we
find a better solution in Q4, we will upgrade.

I have also addressed [this](https://github.com/keep-network/keep-core/pull/2695#discussion_r756127603) and [this](https://github.com/keep-network/keep-core/pull/2695#discussion_r756028278) review comments from #2695.

Gas consumption before (min/max/avg):
```
···························|··································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  approveDkgResult                ·     184343  ·     218862  ·     193086  ·           73  ·          -  │
···························|··································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  challengeDkgResult              ·     984278  ·    1004278  ·     991234  ·           35  ·          -  │
···························|··································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  submitDkgResult                 ·     755531  ·    1057737  ·    1001734  ·          131  ·          -  │
···························|··································|·············|·············|·············|···············|··············
```

Gas consumption after (min/max/avg):
```
···························|··································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  approveDkgResult                ·     184980  ·     219535  ·     193801  ·           73  ·          -  │
···························|··································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  challengeDkgResult              ·    1224991  ·    1277027  ·    1228541  ·           35  ·          -  │
···························|··································|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  submitDkgResult                 ·     339478  ·     636580  ·     591268  ·          131  ·          -  │
···························|··································|·············|·············|·············|···············|··············
```

### DIFF
--- a/solidity/random-beacon/contracts/DKGValidator.sol
+++ b/solidity/random-beacon/contracts/DKGValidator.sol
@@ -108,12 +108,6 @@ contract DKGValidator {
                     return (false, "Corrupted misbehaved members indices");
                 }
             }
-            uint8 last = misbehavedMembersIndices[
-                misbehavedMembersIndices.length - 1
-            ];
-            if (last < 1 || last > groupSize) {
-                return (false, "Corrupted misbehaved members indices");
-            }
         }
 
         // Each signature needs to be 65 bytes long and signatures need to be
@@ -149,12 +143,6 @@ contract DKGValidator {
         }
         for (uint256 i = 1; i < signingMembersIndices.length; i++) {
             if (signingMembersIndices[i - 1] >= signingMembersIndices[i]) {
-                return (false, "Corrupted signing member indices");
-            }
-            uint256 last = signingMembersIndices[
-                signingMembersIndices.length - 1
-            ];
-            if (last < 1 || last > groupSize) {
                 return (false, "Corrupted signing member indices");
             }
         }
--- a/solidity/random-beacon/contracts/DKGValidator.sol
+++ b/solidity/random-beacon/contracts/DKGValidator.sol
@@ -93,10 +93,15 @@ contract DKGValidator {
             return (false, "Too many members misbehaving during DKG");
         }
         if (misbehavedMembersIndices.length > 1) {
+            if (
+                misbehavedMembersIndices[0] < 1 ||
+                misbehavedMembersIndices[misbehavedMembersIndices.length - 1] >
+                64
+            ) {
+                return (false, "Corrupted misbehaved members indices");
+            }
             for (uint256 i = 1; i < misbehavedMembersIndices.length; i++) {
                 if (
-                    misbehavedMembersIndices[i - 1] < 1 ||
-                    misbehavedMembersIndices[i - 1] > groupSize ||
                     misbehavedMembersIndices[i - 1] >=
                     misbehavedMembersIndices[i]
                 ) {
@@ -136,12 +141,14 @@ contract DKGValidator {
 
         // Signing member indices needs to be unique, between [1,64], and sorted
         // in ascending order.
+        if (
+            signingMembersIndices[0] < 1 ||
+            signingMembersIndices[signingMembersIndices.length - 1] > 64
+        ) {
+            return (false, "Corrupted signing member indices");
+        }
         for (uint256 i = 1; i < signingMembersIndices.length; i++) {
-            if (
-                signingMembersIndices[i - 1] < 1 ||
-                signingMembersIndices[i - 1] > groupSize ||
-                signingMembersIndices[i - 1] >= signingMembersIndices[i]
-            ) {
+            if (signingMembersIndices[i - 1] >= signingMembersIndices[i]) {
                 return (false, "Corrupted signing member indices");
             }
             uint256 last = signingMembersIndices[
--- a/solidity/random-beacon/contracts/DKGValidator.sol
+++ b/solidity/random-beacon/contracts/DKGValidator.sol
@@ -1,0 +1,223 @@
+// SPDX-License-Identifier: MIT
+
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "./libraries/BytesLib.sol";
+import "./libraries/DKG.sol";
+
+pragma solidity ^0.8.6;
+
+/// @title DKG result validator
+/// @notice DKGValidator allows performing a full validation of DKG result,
+///         including checking the format of fields in the result, declared
+///         selected group members, and signatures of operators supporting the
+///         result. The operator submitting the result should perform the
+///         validation using a free contract call before submitting the result
+///         to ensure their result is valid and can not be challenged. All other
+///         network operators should perform validation of the submitted result
+///         using a free contract call and challenge the result if the
+///         validation fails.
+contract DKGValidator {
+    using BytesLib for bytes;
+    using ECDSA for bytes32;
+
+    /// @dev Size of a group in the threshold relay.
+    uint256 public constant groupSize = 64;
+
+    /// @dev The minimum number of group members needed to interact according to
+    ///      the protocol to produce a relay entry. The adversary can not learn
+    ///      anything about the key as long as it does not break into
+    ///      groupThreshold+1 of members.
+    uint256 public constant groupThreshold = 33;
+
+    /// @dev The minimum number of active and properly behaving group members
+    ///      during the DKG needed to accept the result. This number is higher
+    ///      than `groupThreshold` to keep a safety margin for members becoming
+    ///      inactive after DKG so that the group can still produce a relay
+    ///      entry.
+    uint256 public constant activeThreshold = 58; // 90% of groupSize
+
+    SortitionPool public sortitionPool;
+
+    constructor(SortitionPool _sortitionPool) {
+        sortitionPool = _sortitionPool;
+    }
+
+    /// @notice Performs a full validation of DKG result, including checking the
+    ///         format of fields in the result, declared selected group members,
+    ///         and signatures of operators supporting the result.
+    /// @param seed seed used to start the DKG and select group members
+    /// @param startBlock DKG start block
+    /// @return isValid true if the result is valid, false otherwise
+    /// @return errorMsg validation error message; empty for a valid result
+    function validate(
+        DKG.Result calldata result,
+        uint256 seed,
+        uint256 startBlock
+    ) external view returns (bool isValid, string memory errorMsg) {
+        (bool hasValidFields, string memory error) = validateFields(result);
+        if (!hasValidFields) {
+            return (false, error);
+        }
+
+        if (!validateSignatures(result, startBlock)) {
+            return (false, "Invalid signatures");
+        }
+
+        if (!validateGroupMembers(result, seed)) {
+            return (false, "Invalid group members");
+        }
+
+        return (true, "");
+    }
+
+    /// @notice Performs a static validation of DKG result fields: lengths,
+    ///         ranges, and order of arrays.
+    /// @return isValid true if the result is valid, false otherwise
+    /// @return errorMsg validation error message; empty for a valid result
+    function validateFields(DKG.Result calldata result)
+        public
+        view
+        returns (bool isValid, string memory errorMsg)
+    {
+        // Group public key needs to be 128 bytes long.
+        if (result.groupPubKey.length != 128) {
+            return (false, "Malformed group public key");
+        }
+
+        // The number of misbehaved members can not exceed the threshold.
+        // Misbehaved member indices needs to be unique, between [1,64],
+        // and sorted in ascending order.
+        uint8[] calldata misbehavedMembersIndices = result
+            .misbehavedMembersIndices;
+        if (groupSize - misbehavedMembersIndices.length < activeThreshold) {
+            return (false, "Too many members misbehaving during DKG");
+        }
+        if (misbehavedMembersIndices.length > 1) {
+            for (uint256 i = 1; i < misbehavedMembersIndices.length; i++) {
+                if (
+                    misbehavedMembersIndices[i - 1] < 1 ||
+                    misbehavedMembersIndices[i - 1] > groupSize ||
+                    misbehavedMembersIndices[i - 1] >=
+                    misbehavedMembersIndices[i]
+                ) {
+                    return (false, "Corrupted misbehaved members indices");
+                }
+            }
+            uint8 last = misbehavedMembersIndices[
+                misbehavedMembersIndices.length - 1
+            ];
+            if (last < 1 || last > groupSize) {
+                return (false, "Corrupted misbehaved members indices");
+            }
+        }
+
+        // Each signature needs to be 65 bytes long and signatures need to be
+        // provided.
+        uint256 signaturesCount = result.signatures.length / 65;
+        if (result.signatures.length == 0) {
+            return (false, "No signatures provided");
+        }
+        if (result.signatures.length % 65 != 0) {
+            return (false, "Malformed signatures array");
+        }
+
+        // We expect the same amount of signatures as the number of declared
+        // group member indices that signed the result.
+        uint256[] calldata signingMembersIndices = result.signingMembersIndices;
+        if (signaturesCount != signingMembersIndices.length) {
+            return (false, "Unexpected signatures count");
+        }
+        if (signaturesCount < groupThreshold) {
+            return (false, "Too few signatures");
+        }
+        if (signaturesCount > groupSize) {
+            return (false, "Too many signatures");
+        }
+
+        // Signing member indices needs to be unique, between [1,64], and sorted
+        // in ascending order.
+        for (uint256 i = 1; i < signingMembersIndices.length; i++) {
+            if (
+                signingMembersIndices[i - 1] < 1 ||
+                signingMembersIndices[i - 1] > groupSize ||
+                signingMembersIndices[i - 1] >= signingMembersIndices[i]
+            ) {
+                return (false, "Corrupted signing member indices");
+            }
+            uint256 last = signingMembersIndices[
+                signingMembersIndices.length - 1
+            ];
+            if (last < 1 || last > groupSize) {
+                return (false, "Corrupted signing member indices");
+            }
+        }
+
+        return (true, "");
+    }
+
+    /// @notice Performs validation of group members as declared in DKG
+    ///         result against group members selected by the sortition pool.
+    /// @param seed seed used to start the DKG and select group members
+    /// @return true if group members matches; false otherwise
+    function validateGroupMembers(DKG.Result calldata result, uint256 seed)
+        public
+        view
+        returns (bool)
+    {
+        // Compute the actual group members hash by selecting actual members IDs
+        // based on seed used for current DKG execution.
+        bytes32 actualGroupMembersHash = keccak256(
+            abi.encodePacked(
+                sortitionPool.selectGroup(groupSize, bytes32(seed))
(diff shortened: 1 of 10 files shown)


solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -472,33 +461,20 @@ library DKG {
...
-        bytes32 actualGroupMembersHash = keccak256(
-            abi.encodePacked(
-                self.sortitionPool.selectGroup(groupSize, bytes32(self.seed))
-            )
-        );
+        bool areSignaturesValid = hasValidSignatures(self, result);
+        bool areMembersValid = true;
+        if (areSignaturesValid) {
+            areMembersValid = hasValidGroupMembers(self, result);
+        }
```
pdyraga, line 468: We talked with @tomaszslabon that it would be good to wrap this part into a try-catch in case the signature validation fails for whatever reason. In `submitDkgResult` we only validate length, and no that the signature has correct format, for example, if it is prefixed with `\x19Ethereum signed message:\n`

lukasz-zimnoch: Try/catch works only for failed external calls and contract creation calls (source: https://docs.soliditylang.org/en/latest/control-structures.html#try-catch)

pdyraga: Yep, this will get expensive. We need to check how `hash.toEthSignedMessageHash().recover(...)` behaves. If the behavior is well-defined and we can protect against garbage data we might be able to avoid it.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -245,116 +259,92 @@ library DKG {
...
-
+        // Signing member indices needs to be unique, between [1,64], and sorted
+        // in ascending order.
+        for (uint256 i = 1; i < signingMembersIndices.length; i++) {
             require(
-                !usedMemberIndices[memberIndex - 1],
-                "Duplicate member index"
+                signingMembersIndices[i - 1] >= 1 &&
+                    signingMembersIndices[i - 1] <= groupSize &&
+                    signingMembersIndices[i - 1] < signingMembersIndices[i],
```
tomaszslabon, line 339: I think we don't need to check that every single element in `singningMembersIndices` is `>=1` and `<=groupSize`.
Checking that `signingMembersIndices[0] >= 1` and signingMembersIndices[signingMembersIndices.length-1]<= groupSize` should be enough, because we also ensure that  `signingMembersIndices` are sorted.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -245,116 +259,92 @@ library DKG {
...
             "Too many members misbehaving during DKG"
         );
-
-        if (result.misbehavedMembersIndices.length > 1) {
-            for (
-                uint256 i = 1;
-                i < result.misbehavedMembersIndices.length;
-                i++
-            ) {
+        if (misbehavedMembersIndices.length > 1) {
```
tomaszslabon, line 295: I think we could rewrite it like this:
```       
if (misbehavedMembersIndices.length > 1) {
    require(
        misbehavedMembersIndices[0] >= 1 &&
            misbehavedMembersIndices[
                misbehavedMembersIndices.length - 1
            ] <=
            groupSize,
        "Corrupted misbehaved members indices"
    );

    for (uint256 i = 0; i < misbehavedMembersIndices.length - 1; i++) {
        require(
            misbehavedMembersIndices[i] <
                misbehavedMembersIndices[i + 1],
            "Corrupted misbehaved members indices"
        );
    }
}
```

If we ensure the elements are sorted, then we just need to check the left-most and right-most values are `>=1 && <= groupSize`.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -472,33 +461,20 @@ library DKG {
...
+        bool areSignaturesValid = hasValidSignatures(self, result);
+        bool areMembersValid = true;
+        if (areSignaturesValid) {
+            areMembersValid = hasValidGroupMembers(self, result);
+        }
 
         require(
-            keccak256(abi.encodePacked(result.members)) !=
-                actualGroupMembersHash,
+            !areMembersValid || !areSignaturesValid,
```
tomaszslabon, line 471: It should probably be:
```
require(areMembersValid && areSignaturesValid, "unjustified challenge");
```

solidity/random-beacon/contracts/libraries/DKG.sol
pdyraga: Very good catch!

solidity/random-beacon/contracts/libraries/DKG.sol
pdyraga: Yep, very good catch!

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -472,33 +401,30 @@ library DKG {
...
+                errorMsg
+            );
+        } catch {
+            // if the validation reverted we consider the DKG result as invalid
+            emit DkgResultChallenged(
+                self.submittedResultHash,
+                msg.sender,
+                "validation reverted"
+            );
+        }
```
pdyraga, line 423:
```
INFO:Detectors:
DKG.challengeResult(DKG.Data,DKG.Result).isValid (contracts/libraries/DKG.sol#406) is a local variable never initialized
DKG.challengeResult(DKG.Data,DKG.Result).errorMsg (contracts/libraries/DKG.sol#406) is a local variable never initialized
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#uninitialized-local-variables
INFO:Detectors:
DKG.challengeResult(DKG.Data,DKG.Result) (contracts/libraries/DKG.sol#383-439) ignores return value by self.dkgValidator.validate(result,self.seed,self.startBlock) (contracts/libraries/DKG.sol#404-423)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#unused-return
INFO:Detectors:
Variable 'DKG.challengeResult(DKG.Data,DKG.Result).isValid (contracts/libraries/DKG.sol#406)' in DKG.challengeResult(DKG.Data,DKG.Result) (contracts/libraries/DKG.sol#383-439) potentially used before declaration: isValid (contracts/libraries/DKG.sol#407)
Variable 'DKG.challengeResult(DKG.Data,DKG.Result).errorMsg (contracts/libraries/DKG.sol#406)' in DKG.challengeResult(DKG.Data,DKG.Result) (contracts/libraries/DKG.sol#383-439) potentially used before declaration: DkgResultChallenged(self.submittedResultHash,msg.sender,errorMsg) (contracts/libraries/DKG.sol#411-415)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#pre-declaration-usage-of-local-variables
```

I am confused about this warning. 🤔

solidity/random-beacon/contracts/libraries/DKG.sol
pdyraga: Extracted to a separate contract.

solidity/random-beacon/test/DKGValidator.test.ts
```diff
@@ -0,0 +1,350 @@
...
+  )) as DKGValidator
+  await dkgValidator.deployed()
+
+  return {
+    sortitionPool,
+    dkgValidator,
+  }
+}
+
+describe("DKGValidator", () => {
```
pdyraga, line 36: This is not testing all possible paths - we haven't had the validation code 100% covered 😬 @tomaszslabon is working on it separately.

solidity/random-beacon/contracts/DKGValidator.sol
```diff
@@ -0,0 +1,230 @@
...
+    ///         and signatures of operators supporting the result.
+    /// @param seed seed used to start the DKG and select group members
+    /// @param startBlock DKG start block
+    /// @return isValid true if the result is valid, false otherwise
+    /// @return errorMsg validation error message; empty for a valid result
+    function validate(
+        DKG.Result calldata result,
+        uint256 seed,
+        uint256 startBlock
+    ) external view returns (bool isValid, string memory errorMsg) {
```
tomaszslabon, line 56: We don't assign anything to `isValid` and `errorMsg`, so maybe:
```
external view returns (bool , string memory) {
....
}

File: solidity/random-beacon/contracts/DKGValidator.sol [hunk 8]
```diff
@@ -0,0 +1,230 @@
...
+            }
+            for (uint256 i = 1; i < misbehavedMembersIndices.length; i++) {
+                if (
+                    misbehavedMembersIndices[i - 1] >=
+                    misbehavedMembersIndices[i]
+                ) {
+                    return (false, "Corrupted misbehaved members indices");
+                }
+            }
+            uint8 last = misbehavedMembersIndices[
```
tomaszslabon, line 111: We already know that `last` is <= 64, because we checked it before the for loop.
We also know that it's >= 1.

solidity/random-beacon/contracts/DKGValidator.sol
```diff
@@ -0,0 +1,230 @@
...
+                misbehavedMembersIndices.length - 1
+            ];
+            if (last < 1 || last > groupSize) {
+                return (false, "Corrupted misbehaved members indices");
+            }
```
lukasz-zimnoch: We do the same for `signingMembersIndices`. Maybe it would be worth extracting this block to separate function and improve readability? Something like `validateMembersIndices` in #2736

solidity/random-beacon/contracts/DKGValidator.sol
```diff
@@ -0,0 +1,230 @@
...
+        uint8[] calldata misbehavedMembersIndices = result
+            .misbehavedMembersIndices;
+        if (groupSize - misbehavedMembersIndices.length < activeThreshold) {
+            return (false, "Too many members misbehaving during DKG");
+        }
+        if (misbehavedMembersIndices.length > 1) {
+            if (
+                misbehavedMembersIndices[0] < 1 ||
+                misbehavedMembersIndices[misbehavedMembersIndices.length - 1] >
+                64
```
lukasz-zimnoch, line 99: Maybe `groupSize` instead of just `64`?

solidity/random-beacon/contracts/DKGValidator.sol
```diff
@@ -0,0 +1,230 @@
...
+        }
+        if (signaturesCount > groupSize) {
+            return (false, "Too many signatures");
+        }
+
+        // Signing member indices needs to be unique, between [1,64], and sorted
+        // in ascending order.
+        if (
+            signingMembersIndices[0] < 1 ||
+            signingMembersIndices[signingMembersIndices.length - 1] > 64
```
lukasz-zimnoch, line 140: Maybe `groupSize` instead of just `64`?

solidity/random-beacon/contracts/DKGValidator.sol
```diff
@@ -0,0 +1,230 @@
...
+                misbehavedMembersIndices.length - 1
+            ];
+            if (last < 1 || last > groupSize) {
+                return (false, "Corrupted misbehaved members indices");
+            }
+        }
+
+        // Each signature needs to be 65 bytes long and signatures need to be
+        // provided.
+        uint256 signaturesCount = result.signatures.length / 65;
```
lukasz-zimnoch, line 115: Maybe we can extract `65` to a constant and name it, for example, `signatureByteSize`?

solidity/random-beacon/contracts/DKGValidator.sol
```diff
@@ -0,0 +1,230 @@
...
+            )
+        );
+
+        address[] memory membersAddresses = sortitionPool.getIDOperators(
+            result.members
+        );
+
+        bytes memory current; // Current signature to be checked.
+
+        uint256 signaturesCount = result.signatures.length / 65;
```
lukasz-zimnoch, line 202: If we decide to extract `signatureBytesSize` we can use it here.

solidity/random-beacon/contracts/DKGValidator.sol
```diff
@@ -0,0 +1,230 @@
...
+        for (uint256 i = 0; i < signaturesCount; i++) {
+            uint256 memberIndex = result.signingMembersIndices[i];
+
+            current = result.signatures.slice(65 * i, 65);
```
lukasz-zimnoch, line 206: If we decide to extract `signatureBytesSize` we can use it here.

solidity/random-beacon/contracts/DKGValidator.sol
```diff
@@ -0,0 +1,230 @@
...
+            signingMembersIndices[0] < 1 ||
+            signingMembersIndices[signingMembersIndices.length - 1] > 64
+        ) {
+            return (false, "Corrupted signing member indices");
+        }
+        for (uint256 i = 1; i < signingMembersIndices.length; i++) {
+            if (signingMembersIndices[i - 1] >= signingMembersIndices[i]) {
+                return (false, "Corrupted signing member indices");
+            }
+            uint256 last = signingMembersIndices[
```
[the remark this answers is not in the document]
pdyraga: Yep!

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: I was thinking about it but at the same time, we know what these fields represent if we name them - we can even use their names in docs.

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: Yep! Fixed in d88c7633d26b16fa1f8b72d9e14f013b9d437016

solidity/random-beacon/contracts/DKGValidator.sol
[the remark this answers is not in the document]
pdyraga: Fixed in d88c7633d26b16fa1f8b72d9e14f013b9d437016.

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: `misbehavedMembersIndices` is `uint8[]` and `signingMembersIndices` is `uint256[]` (for gas efficiency reasons) :(

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: https://github.com/keep-network/keep-core/pull/2734/commits/5bc95364647baabb9b6c7951805b9fb2681dcdfb

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: https://github.com/keep-network/keep-core/pull/2734/commits/5bc95364647baabb9b6c7951805b9fb2681dcdfb

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: https://github.com/keep-network/keep-core/pull/2734/commits/5bc95364647baabb9b6c7951805b9fb2681dcdfb

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: https://github.com/keep-network/keep-core/pull/2734/commits/5bc95364647baabb9b6c7951805b9fb2681dcdfb

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: https://github.com/keep-network/keep-core/pull/2734/commits/5bc95364647baabb9b6c7951805b9fb2681dcdfb

solidity/random-beacon/contracts/DKGValidator.sol
lukasz-zimnoch: Ah ok. But, can't we convert `signingMembersIndices` to `uint8[]` as well? Max index is `64` and we could save on gas in the same way?

solidity/random-beacon/contracts/libraries/DKG.sol
pdyraga: Ignored checks in ab1e64985bde809a60c4e3152ce50eba0d56efba and reported the problem: https://github.com/crytic/slither/issues/982.

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: I am pretty sure it was something about gas cost. If we make then `uint8[]`, the compiler converts to `uint256` when we use them. Let's compare the gas cost once we merge this PR and change (or not).

lukasz-zimnoch (approved):

PR #2737, opened 2021-11-29
Updated `sortition-pools` dependency to the most recent one

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/2715

See https://github.com/keep-network/sortition-pools/pull/135
See https://github.com/keep-network/sortition-pools/pull/136
See https://github.com/keep-network/sortition-pools/pull/137

This is a preparation for T staking contract integration but will also allow
to use `selectGroup` for challenges given that in the most recent version this
function is not protected by `onlyOwner` modifier.

Sortition pool does not longer expose `isEligibleFunction`. Everyone who
has the minimum required authorization for Random Beacon on T staking contract
will be able to join the pool. Removed `RandomBeacon.isEligible`.

I have also updated `RandomBeacon.registerOperator` and
`RandomBeacon.updateOperatorStatus`. `SortitionPool` contract does no longer
work with staking contract and it requires the contract owner to pass the
current authorization amount. These functions will be reworked during the
integration with T staking contract. For now, I am just making everything
to compile and work with the most recent `sortition-pools` version.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -493,7 +493,7 @@ contract RandomBeacon is Ownable {
 
         sortitionPool.insertOperator(
             operator,
-            staking.eligibleStake(msg.sender, address(this))
+            staking.eligibleStake(operator, address(this))
         );
     }
 
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -42,6 +42,11 @@ interface IRandomBeaconStaking {
         address notifier,
         address[] memory operators
     ) external;
+
+    function eligibleStake(address operator, address operatorContract)
+        external
+        view
+        returns (uint256);
 }
 
 /// @title Keep Random Beacon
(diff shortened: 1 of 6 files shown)


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -486,23 +491,20 @@ contract RandomBeacon is Ownable {
             "Operator is already registered"
         );
 
-        sortitionPool.insertOperator(operator);
+        sortitionPool.insertOperator(
+            operator,
+            staking.eligibleStake(msg.sender, address(this))
```
dimpar, line 496: in line 487 we define `operator = msg.sender`. Should we use `operator` here for readibility as well? Or remove operator and use msg.sender instead

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -486,23 +491,20 @@ contract RandomBeacon is Ownable {
...
+        );
     }
 
     /// @notice Updates the sortition pool status of the caller.
     function updateOperatorStatus() external {
         sortitionPool.updateOperatorStatus(
```
dimpar, line 502: Did we remove `sortitionPool.getOperatorID()` function from SP?

solidity/random-beacon/contracts/RandomBeacon.sol
pdyraga: Good point, done in https://github.com/keep-network/keep-core/pull/2737/commits/d42244fa2b145b5d2d07a154c6a0970233b71ca4.

solidity/random-beacon/contracts/RandomBeacon.sol
pdyraga: It is still there but `updateOperatorStatus` accepts `address` now.

pdyraga:
Rebased after https://github.com/keep-network/keep-core/pull/2740 landed in `main`. Linter build should be green now 🤞🏻

dimpar (approved):

PR #2744, opened 2021-12-01
Integrate Tally wallet into keep dashboard

greg-nagy:
Closes https://github.com/tallycash/tally-extension/issues/470

## Constraints
Currently we don't have a published version of the wallet, so the integration can only be tested against a local build of the wallet
- See the quickstart for the extension repo.
- The fix for depends on this PR https://github.com/tallycash/extension/pull/701
- We support only mainnet currently
- You need to enter an address to use the integration

### Pictures of the demo working
https://cln.sh/ARBd9F - gif of the feature working 

CleanShot 2021-12-01 at 12 14 50
CleanShot 2021-12-01 at 12 13 34
CleanShot 2021-12-01 at 12 15 30
CleanShot 2021-12-01 at 12 16 30


solidity-v1/dashboard/src/hooks/useAutoConnect.js
```diff
@@ -41,29 +41,56 @@ const useAutoConnect = () => {
...
+        } else if (walletAddressFromUrl && !connector) {
+          const explorerModeConnector = new ExplorerModeConnector()
+          openModal(MODAL_TYPES.ExplorerMode, {
+            connectAppWithWallet,
+            connector: explorerModeConnector,
+            address: walletAddressFromUrl,
+            connectWithWalletOnMount: true,
+          })
+        }
+      })
```
michalsmiarowski, line 68: In both ifs we have basically the same code but with different connector. We should extract if to a separate method that receive connector as an argument. For example:

`const connectWallet = (connector) = (...) `

and then just:
```suggestion
connectWallet(tallyInjectedConnector)
```

solidity-v1/dashboard/src/hooks/useAutoConnect.js
```diff
@@ -41,29 +41,56 @@ const useAutoConnect = () => {
...
+    }
```
michalsmiarowski, line 93: Similar as above

solidity-v1/dashboard/src/connectors/tally-injected.js
```diff
@@ -0,0 +1,76 @@
+import { AbstractConnector } from "./abstract-connector"
+import { WALLETS } from "../constants/constants"
+import BigNumber from "bignumber.js"
+import { UserRejectedConnectionRequestError } from "./utils"
+
+class TallyInjectedConnector extends AbstractConnector {
+  constructor() {
+    super(WALLETS.TALLY.name)    
+    this.provider = window.tally
+  }
```
michalsmiarowski, line 10: `TallyInjectedConnector` and `MetaMaskInjectedConnector` have basically the same code. The only difference is that it uses different provider and we set up `autoRefreshOnNetworkChange` to false for MetaMask. 

What I propose it to keep it as 'InjectedConnector' and extend both 'TallyInjectedConnector' and 'MetaMaskInjectedConnector' with it.

```suggestion
class InjectedConnector extends AbstractConnector {
  constructor(walletName, provider) {
    super(walletName)    
    this.provider = provider
  }
```

And then:
```
class TallyInjectedConnector extends InjectedConnector {
  constructor() {
    super(WALLETS.TALLY.name, window.tally)
  }
```

Note: The code was not tested.

Thoughts?

solidity-v1/dashboard/src/connectors/tally-injected.js
```diff
@@ -0,0 +1,76 @@
...
+    this.emit("disconnect")
+  }
+
+  _onChainChanged = (chainId) => {
+    this.emit("chainChanged", chainId)
+  }
+
+  enable = async () => {
+    if (!this.provider) {
+      throw new Error("window.tally provider not found")
```
michalsmiarowski, line 26: If you agree with above then we should probably change the error text here. Maybe we could just use "provider not found"?

Also we should make sure that we set `autoRefresOnNetworkChange`  to false for MetaMask only so we would probably have to do some proper changes to that:

```
if (walletName === WALLETS.NAME.METAMASK) {
      // https://docs.metamask.io/guide/ethereum-provider.html#ethereum-autorefreshonnetworkchange
      this.provider.autoRefreshOnNetworkChange = false
}
```

michalsmiarowski (changes requested):
Left some comments too look at 😉 

Also let's run `npm format:fix` before commit to get rid of lint and prettier warnings.

solidity-v1/dashboard/src/connectors/tally-injected.js
greg-nagy: https://github.com/keep-network/keep-core/pull/2744/commits/3248bcf4a251f24f0c6e09faa4402b92f67cb471

solidity-v1/dashboard/src/connectors/tally-injected.js
greg-nagy: https://github.com/keep-network/keep-core/pull/2744/commits/3248bcf4a251f24f0c6e09faa4402b92f67cb471

solidity-v1/dashboard/src/hooks/useAutoConnect.js
greg-nagy: https://github.com/keep-network/keep-core/pull/2744/commits/0c22c3d078d822867fc8de0c83eeeab309a01192

solidity-v1/dashboard/src/hooks/useAutoConnect.js
```diff
@@ -42,59 +42,38 @@ const useAutoConnect = () => {
...
+        const explorerModeConnector = new ExplorerModeConnector()
+        openModal(MODAL_TYPES.ExplorerMode, {
+          connectAppWithWallet,
+          connector: explorerModeConnector,
+          address: walletAddressFromUrl,
+          connectWithWalletOnMount: true,
+        })
+      }
+    })
+  }
```
michalsmiarowski: I think we should move this function inside the `useEffect()` function or refactor it to `useCallback()` with proper dependencies. I think the first approach would be the best because we are not using this function outside of the `useEeffect`

For more infor please see [Can I call separate function in useEffect?](https://stackoverflow.com/a/61918728) and [Define a function inside useEffect or outside?](https://stackoverflow.com/a/61528022)

solidity-v1/dashboard/src/connectors/tally-injected.js
```diff
@@ -0,0 +1,10 @@
+import { WALLETS } from "../constants/constants"
+import { InjectedConnector } from "./injected-connector"
+
+class TallyInjectedConnector extends InjectedConnector {
+  constructor() {
+    super(WALLETS.TALLY.name, window.tally)
```
michalsmiarowski, line 6: @greg-nagy  I might be wrong, but it looks like something with tally extension is not working right, because the tally provider is not injected after refreshing the page 🤔 It works fine when i rebuild the project, but when i manually refresh the page the `window.tally` is `undefined`.

There is also strange behaviour when i disconnect from Tally and connect to it again - it will automatically refresh the page.

michalsmiarowski (changes requested):
Some more comments to look at before the merge 😄

solidity-v1/dashboard/src/connectors/tally-injected.js
greg-nagy: @michalsmiarowski thanks, I noticed the same. Will get back with fixes soon.

greg-nagy: We have made some changes so I don't think this will be a problem. But let's test again today.

solidity-v1/dashboard/src/connectors/injected-connector.js
```diff
@@ -50,7 +56,7 @@ class InjectedConnector extends AbstractConnector {
   }
 
   disconnect = async () => {
-    // window.ethereum injected by MetaMask does not provide a method to
+    // window.tally injected by Tally does not provide a method to
```
michalsmiarowski, line 59: This applies to both MetaMask and Tally so we should probably change this comment to something like:

```suggestion
    // window.ethereum injected by MetaMask and window.tally injected by Tally does not provide a method to
```

greg-nagy: done

solidity-v1/dashboard/src/hooks/useAutoConnect.js
```diff
@@ -27,11 +30,11 @@ const useAutoConnect = () => {
   const isExactRoutePath = useIsExactRoutePath()
   const [injectedTried, setInjectedTried] = useState(false)
 
-  const isWalletFromUrlSameAsInMetamask = useCallback(
-    (metamaskAccounts) => {
+  const isWalletFromUrlSameAsInProvider = useCallback(
+    (providerAccount) => {
```
michalsmiarowski, line 34: We are passing here array of accounts returned from `getAccounts()` method so I think we should keep it as plural:

```suggestion
    (providerAccounts) => {
```

solidity-v1/dashboard/src/hooks/useAutoConnect.js
```diff
@@ -27,11 +30,11 @@ const useAutoConnect = () => {
...
       return (
         walletAddressFromUrl &&
-        metamaskAccounts.some((account) =>
+        providerAccount.some((account) =>
```
michalsmiarowski, line 37: Same as above:

```suggestion
        providerAccounts.some((account) =>
```

solidity-v1/dashboard/src/hooks/useAutoConnect.js
```diff
@@ -41,37 +44,46 @@ const useAutoConnect = () => {
...
+          const explorerModeConnector = new ExplorerModeConnector()
+          openModal(MODAL_TYPES.ExplorerMode, {
+            connectAppWithWallet,
+            connector: explorerModeConnector,
+            address: walletAddressFromUrl,
+            connectWithWalletOnMount: true,
+          })
+        }
+      })
+    }
```
michalsmiarowski, line 72: Right now Explorer Mode does not work as it supposed to work - when we change the wallet address in the url for the one that is not connected to neither MetaMask nor Tally, then nothing happens (and it should connect to Explorer Mode in such case)

The reason of such behavious lies in this function. The `connector` variable that is the argument of `connectWallet` function is not the same  as `connector` used in `else if` statement. The first one is the actual connector that we created for MetaMask/Tally, and the seconds one is the actual used connector that is saved on Redux side. You can see in the deleted lines above that we used `injected` and `connector` as two different variables.

So we have to change those names to different ones. I think we can just change the first `connector` variable to `injectedConnector` but I'm open to suggestion if you have better idea for the name or want to rename the other one.

```suggestion
    const connectWallet = (injectedConnector) => {
      injectedConnector.getAccounts().then((accounts) => {
        setInjectedTried(true)
        if (
          (!isEmptyArray(accounts) && isExactRoutePath) ||
          isWalletFromUrlSameAsInProvider(accounts)
        ) {
          connectAppWithWallet(injectedConnector, false).catch((error) => {
            // Just log an error, we don't want to do anything else.
            console.log(
              "Eager injected connector cannot connect with the dapp:",
              error.message
            )
          })
        } else if (walletAddressFromUrl && !connector) {
          const explorerModeConnector = new ExplorerModeConnector()
          openModal(MODAL_TYPES.ExplorerMode, {
            connectAppWithWallet,
            connector: explorerModeConnector,
            address: walletAddressFromUrl,
            connectWithWalletOnMount: true,
          })
        }
      })
    }
```

solidity-v1/dashboard/src/components/modal/wallets/TallyModal.jsx
```diff
@@ -0,0 +1,44 @@
+import React from "react"
+import { ModalHeader } from "../Modal"
+import { SelectedWalletModal } from "./SelectedWalletModal"
+import * as Icons from "../../Icons"
+import { withBaseModal } from "../withBaseModal"
+
+const rejectedConnectionErrorMsg =
+  "You rejected the connection request in Tally. Please close and try again, and click confirm connection in Tally window."
```
michalsmiarowski, line 8: I think too many `and`a are used here 😅 I would change it to:

```suggestion
  "You rejected the connection request in Tally. Please close and try again, then click confirm connection in Tally window."
```

Thoughts?

michalsmiarowski: If you agree then also change the same thing in `MetaMaskModal.jsx`

michalsmiarowski (changes requested):
Tested it and both mentioned issues (Tally refreshing after relogging and Tally not cooperating well with MetaMask) appear to be fixed! Great job! 🔥 

There is another issue that i found thought and it's with Explorer Mode. You can find details below.

solidity-v1/dashboard/src/hooks/useAutoConnect.js
greg-nagy: This is awesome! This fix is 🔥

I just typed it in and it worked! 💪

michalsmiarowski (approved):
Looking good! Let's go! 🚀

PR #2749, opened 2021-12-02
Updating initial values of RandomBeacon parameters

pdyraga:
Spreadsheet with cost calculations: https://docs.google.com/spreadsheets/d/1JgRCj6srYCHANA8tmgR1XgvsE3eFrYzwSiaTRGkUI5M/edit?usp=sharing

### Initial callback gas limit set to 50k

This is enough to store relay entry, the block at which the entry
was produced and emit an event.

### Initial entry/DKG result submission eligibility delay set to 20 blocks 

Default mining check interval in the client is 60 sec which is about
4 Ethereum blocks. 20 blocks eligibility delay is a bit more
operator-friendly leaving space for at least 4 gas price bumps.
This is especially important for DKG result submitter who needs
to approve the result to claim the reward.

It also means that the time before soft relay entry timeout is
hit, increased from 160 to 320 minutes.

### DKG result challenge period increased to ~48h

There is no validation in `submitDkgResult` function and we need to
ensure malicious DKG results are caught and eliminated no matter
what. The best approach would be to validate everything in `submitDkgResult`
function but then, the cost of group creation would increase
significantly. 48h should be enough to report the problem even if
Ethereum network hard-forked.

### Group creation frequency set to 5, group lifetime set to ~10 weeks

Creating a group is really expensive. The total gas cost of `submitDkgResult`
followed by `approveDkgResult` is 780849 on average and 526488 minimum.

Assuming the average gas consumption, the cost in ETH is:
100Gwei gas price: 0.0780849 ETH
200Gwei gas price: 0.1561698 ETH
500Gwei gas price: 0.3904245 ETH
1000Gwei gas price: 0.780849 ETH

tBTC is expected to create a new wallet every one week and tBTC will
be the only user of the random beacon initially. It means that with
10 weeks group lifetime and group created every 5 requests, we have
a new group created every 5 weeks and every time, tBTC needs to
provide 20% of the total DKG cost.

### DKG result submission reward set to 1000 T

Assuming the average gas consumption, the cost in ETH is:
100Gwei gas price: 0.0780849 ETH
200Gwei gas price: 0.1561698 ETH
500Gwei gas price: 0.3904245 ETH
1000Gwei gas price: 0.780849 ETH

Assuming 200Gwei gas price and T:ETH ratio same as KEEP:ETH ratio
today (2021-12-02), 1 ETH = 5500 KEEP, the reward in T based on
avg gas cost and 200Gwei gas price should be 858.9339 T.

Setting it to 1000 T to keep it rounded and have some safety margin to
make the reward more attractive

### Initial relay request fee set to 200 T

DKG result submission reward is 1000 T, DKG frequency is 5, so
relay request fee needs to be 1000/5 = 200.

### Sortition pool unlocking reward set to 100 T

The total cost of notifyDkgTimeout is 86883 on average and 85216
minimum.

Assuming the average gas consumption, the cost in ETH is:
100Gwei gas price: 0.0086883 ETH
200Gwei gas price: 0.0173766 ETH
500Gwei gas price: 0.0434415 ETH
1000Gwei gas price: 0.086883 ETH

Assuming 200Gwei gas price and T:ETH ratio same as KEEP:ETH ratio
today (2021-12-02), 1 ETH = 5500 KEEP, the reward in T based on
avg gas cost and 200Gwei gas price should be 95.5713 T.

Setting it to 100 T to keep it rounded and have some safety margin
to make the reward more attractive.

### Updated initial reward multiplier values

Average gas costs are:

- `reportUnauthorizedSigning`: 540505
- `challengeDkgResult`:        1227610
- `reportRelayEntryTimeout`:   407314

Setting reward multipliers proportionally

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -318,7 +318,7 @@ contract RandomBeacon is Ownable {
 
         // TODO: revisit all initial values
         callbackGasLimit = 50000;
-        groupCreationFrequency = 10;
+        groupCreationFrequency = 5;
 
         dkgResultSubmissionReward = 0;
         sortitionPoolUnlockingReward = 0;
@@ -341,7 +341,7 @@ contract RandomBeacon is Ownable {
         relay.setRelayEntryHardTimeout(5760); // ~24h assuming 15s block time
         relay.setRelayEntrySubmissionFailureSlashingAmount(1000e18);
 
-        groups.setGroupLifetime(80640); // ~2weeks assuming 15s block time
+        groups.setGroupLifetime(403200); // ~10 months assuming 15s block time
     }
 
     /// @notice Updates the values of authorization parameters.
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -326,8 +326,8 @@ contract RandomBeacon is Ownable {
         maliciousDkgResultSlashingAmount = 50000e18;
         unauthorizedSigningSlashingAmount = 100e3 * 1e18;
         sortitionPoolRewardsBanDuration = 2 weeks;
-        relayEntryTimeoutNotificationRewardMultiplier = 5;
-        unauthorizedSigningNotificationRewardMultiplier = 5;
+        relayEntryTimeoutNotificationRewardMultiplier = 40;
+        unauthorizedSigningNotificationRewardMultiplier = 50;
         dkgMaliciousResultNotificationRewardMultiplier = 100;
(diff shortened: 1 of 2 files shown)


pdyraga:
I looked at Lukasz's change from https://github.com/keep-network/keep-core/pull/2749/commits/62537140578e93ca293a932a6a94a79cb2393d77 and it looks good to me!

pdyraga:
Changes from https://github.com/keep-network/keep-core/pull/2749/commits/ed0de53d996970a5210ee8fc3176ace12299b7d5 and https://github.com/keep-network/keep-core/pull/2749/commits/628692fc640b9e873697cc54dcb57a37301615e9 also look good.

lukasz-zimnoch (approved):

PR #2750, opened 2021-12-02
Entry gas optimisation

eth-r:


solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -36,19 +37,24 @@ library Relay {
     struct Data {
         // Total count of all requests.
         uint64 requestCount;
-        // Previous entry value.
-        bytes previousEntry;
         // Data of current request.
-        Request currentRequest;
```
pdyraga, line 42: We still have `Request` struct declared in this file. Can we get rid of it?

solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -103,13 +109,15 @@ library Relay {
 
         uint64 currentRequestId = ++self.requestCount;
 
-        self.currentRequest = Request(
+        self.currentRequestID = currentRequestId;
+        self.currentRequestGroupID = groupId;
+        self.currentRequestStartBlock = uint64(block.number);
```
pdyraga, line 114: This will... truncate?

> We did not introduce checks for explicit type conversions from larger to smaller types, because we think that such checks might be unexpected. We would love to get your input on that matter!

https://blog.soliditylang.org/2020/10/28/solidity-0.8.x-preview/

On the other hand:

> Explicit type conversions will always truncate and never cause a failing assertion with the exception of a conversion from an integer to an enum type.

https://docs.soliditylang.org/en/v0.8.10/control-structures.html#checked-or-unchecked-arithmetic

But I think the one from Solidity docs is about `unchecked`.

solidity/random-beacon/contracts/libraries/BLS.sol
```diff
@@ -47,12 +47,25 @@ library BLS {
...
 
+        return
+            _verify(
+                AltBn128.g2Unmarshal(publicKey),
+                AltBn128.g1Unmarshal(message),
+                _signature
+            );
+    }
+
+    function _verify(
```
pdyraga, line 58: Do we need to underscore it? The compiler should resolve overloaded versions correctly I think.

solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -256,7 +272,33 @@ library Relay {
 
         return
             isRequestInProgress(self) &&
-            block.number > self.currentRequest.startBlock + _relayEntryTimeout;
+            block.number > self.currentRequestStartBlock + _relayEntryTimeout;
+    }
+
+    /// @notice Returns whether a relay entry request is currently in progress.
+    /// @return True if there is a request in progress. False otherwise.
+    function _isRequestInProgress(Data memory self)
```
pdyraga, line 280: Can we put it next to `function isRequestInProgress(Data storage self)` version to make it clear we have two and the difference is `storage` vs `memory`?

solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -256,7 +272,33 @@ library Relay {
...
+        internal
+        view
+        returns (bool)
+    {
+        return self.currentRequestID != 0;
+    }
+
+    /// @notice Returns whether the current relay request has timed out.
+    /// @return True if the request timed out. False otherwise.
+    function _hasRequestTimedOut(Data memory self)
```
pdyraga, line 290: Same question about putting it next to `function hasRequestTimedOut(Data storage self)`

pdyraga: If this has a very marginal gas diff, maybe we should get rid of the duplicate code?

solidity/random-beacon/contracts/libraries/BLS.sol
eth-r: Some tests break if I don't underscore it: `test/BLS.test.ts:18:30 - error TS2339: Property 'verify' does not exist on type 'BLS'.`

pdyraga: Ethers can not resolve the correct function, we'd need to provide a signature as a string, e.g. `randomBeacon["getGroup(bytes)"]`.  Let's leave the underscore, it's not worth our pain in tests.

solidity/random-beacon/contracts/libraries/Relay.sol
eth-r: It will truncate, but in practice it would take a few billion years to reach the limit of `uint64` in block numbers so it doesn't matter unless I'm seriously mistaken about something

pdyraga:
Before vs After:
```
···························|························|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  requestRelayEntry     ·     103131  ·     176571  ·     158653  ·           45  ·          -  │
···························|························|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  submitRelayEntry      ·     194627  ·     490189  ·     300537  ·           20  ·          -  │
···························|························|·············|·············|·············|···············|··············


···························|························|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  requestRelayEntry     ·      78800  ·     152240  ·     134322  ·           45  ·          -  │
···························|························|·············|·············|·············|···············|··············
|  RandomBeaconStub        ·  submitRelayEntry      ·     192493  ·     488134  ·     298427  ·           20  ·          -  │
···························|························|·············|·············|·············|···············|··············
```

pdyraga:
Before
 ·--------------------------|-------------·
 |  Contract Name           ·  Size (KB)  │
 ···························|··············
 |  RandomBeacon            ·     21.397  │
 ···························|··············
 |  RandomBeaconStub        ·     23.239  │
 ···························|··············

After
 ·--------------------------|-------------·
 |  Contract Name           ·  Size (KB)  │
 ···························|··············
 |  RandomBeacon            ·     22.164  │
 ···························|··············
 |  RandomBeaconStub        ·     24.132  │
 ···························|··············

pdyraga (approved):

PR #2753, opened 2021-12-03
Modified validation of group members: iterate instead of hashing

tomaszslabon:
This PR modifies the way group members are validated during DKG result challenge: instead of calculating hash, group members are iterated and compared one by one.

After this change `challengeDkgResult` seems to be cheaper in the optimistic case, but slightly more expensive in the pessimistic.
It is also faster in case the lengths of received and actual members are different. 

Iterating:
- optimistic (first operator mismatch): 1142513
- pessimistic (last operator mismatch): 1167251

Hashing (note the cost the same in both cases):
- optimistic: 1160070
- pessimistic: 1160070

Note: I could only test it via `challengeDkgResult` function and not dedicated `DKGValidator` tests (for some reason gas reporter doesn't show the output from `DKGValidator` - maybe because those are `view` functions?)

### DIFF
--- a/solidity/random-beacon/contracts/DKGValidator.sol
+++ b/solidity/random-beacon/contracts/DKGValidator.sol
@@ -175,18 +175,20 @@ contract DKGValidator {
         view
         returns (bool)
     {
-        // Compute the actual group members hash by selecting actual members IDs
-        // based on seed used for current DKG execution.
-        bytes32 actualGroupMembersHash = keccak256(
-            abi.encodePacked(
-                sortitionPool.selectGroup(groupSize, bytes32(seed))
-            )
+        uint32[] calldata resultMembers = result.members;
+        uint32[] memory actualGroupMembers = sortitionPool.selectGroup(
+            groupSize,
+            bytes32(seed)
         );
-
-        // TODO: check what is more efficient - computing hash or iterating
-        return
-            keccak256(abi.encodePacked(result.members)) ==
-            actualGroupMembersHash;
+        if (resultMembers.length != actualGroupMembers.length) {
+            return false;
+        }
+        for (uint256 i = 0; i < resultMembers.length; i++) {
+            if (resultMembers[i] != actualGroupMembers[i]) {
+                return false;
+            }
+        }
+        return true;
     }
 
     /// @notice Performs validation of signatures supplied in DKG result.


dimpar:
I used the following snippet to see the gas usage when running [validate test](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/test/DKGValidator.test.ts#L102)

```
uint256 startGas = gasleft();
(code)
uint256 gasUsed = startGas - gasleft();
console.log("gasUsed", gasUsed);
```

For loop approach
- all members valid: `641258` 
- randomly shuffling members (invalid): `616000`

Hashing approach
- all members valid: `633595`
- randomly shuffling members (invalid): `633595`

Diff between for-loop & hashing when members are valid is 7663gas. I take that in most cases the result should contain valid members, so we probably should leave the current implementation as-is. It will make sense to replace it when we deal in most cases with the invalid members.

tomaszslabon:
> I used the following snippet to see the gas usage when running [validate test](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/test/DKGValidator.test.ts#L102)
> 
> ```
> uint256 startGas = gasleft();
> (code)
> uint256 gasUsed = startGas - gasleft();
> console.log("gasUsed", gasUsed);
> ```
> 
> For loop approach
> 
> * all members valid: `641258`
> * randomly shuffling members (invalid): `616000`
> 
> Hashing approach
> 
> * all members valid: `633595`
> * randomly shuffling members (invalid): `633595`
> 
> Diff between for-loop & hashing when members are valid is 7663gas. I take that in most cases the result should contain valid members, so we probably should leave the current implementation as-is. It will make sense to replace it when we deal in most cases with the invalid members.

I wasn't sure about it either, but it is called in the `challengeDkgResult`, so it should be called when the caller thinks there is something wrong with the DKG result (and the members can be in wrong order).

dimpar:
> validate test

Revisiting this PR, cause it also propagates to ECDSA wallet implementation. My prev. comment around valid members and the gas usage actually doesn't apply here, because if the members are valid then the dkg challenge will be reverted and 0 gas is used.

In 64 members array when the invalid index is somewhere between [0-50] it will be cheaper for iteration. If members[0] (most optimistic) is invalid, then we save around 17k gas. For an invalid member at members[50] the gas usage is almost the same as for hashing approach.

In 100 members array the threshold is the same, index 50, so the probability for iteration and hashing approach is the same in terms of gas usage. I would leave hashing cause it's less code.

So it makes sense to merge this PR in random beacon for 64 members, but for ECDSA with 100 members I'd leave hashing. WDYT @pdyraga @tomaszslabon

dimpar (approved):

PR #2754, opened 2021-12-03
Handle TODOs in Random Beacon tests

nkuba:
This PR handles TODOs that were left in the Random Beacon tests.


tomaszslabon:
@nkuba 
There is one more TODO in tests:
https://github.com/keep-network/keep-core/blob/8e3689ba3aeb9a5d08a71c79282c521f85ee1d2b/solidity/random-beacon/test/Groups.test.ts#L599

I'm not entirely sure what the purpose of that TODO is, but I think we could get rid of it.
I think we test everything there is to test about `activateCandidateGroup` in that testcase.

nkuba:
> @nkuba There is one more TODO in tests:
> 
> https://github.com/keep-network/keep-core/blob/8e3689ba3aeb9a5d08a71c79282c521f85ee1d2b/solidity/random-beacon/test/Groups.test.ts#L599
> 
> I'm not entirely sure what the purpose of that TODO is, but I think we could get rid of it. I think we test everything there is to test about `activateCandidateGroup` in that testcase.

Good catch @tomaszslabon ! I removed the TODO.

tomaszslabon:
I don't have any further comments, just wanted to ask somebody (e.g. @pdyraga) if [calling genesis](https://github.com/keep-network/keep-core/blob/45c444c77221c009420ed9b4702fc776690ed8ef/solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts#L247) is the proper way to restart DKG after DKG timeout notification.

tomaszslabon (approved):

tomaszslabon:
@nkuba I think we could merge it. We just need to resolve the conflicts.

PR #2762, opened 2021-12-06
Optimize tests execution time

lukasz-zimnoch:
Currently, the execution of the entire test suite of the Random Beacon v2 takes about 25 minutes. This is because we use a BDD test structure with a frequent usage of `beforeEach` hook which often executes long-running actions (e.g. `mineBlocks`) repeatedly. We can optimize the execution time by doing some optimizations in the test structure. 

First, we introduce a snapshotting mechanism (`snapshot.ts`) we used to have in v1. It can create a snapshot of the chain state using `evm_snapshot` call and restore it with `evm_revert`. This way we have a tool that allows us to avoid unnecessary setup calls and just jump to a previously saved snapshot containing the desired state. At this point, it's worth noting that Waffle's fixtures behave in a similar way, i.e. make a snapshot of the loaded state using `evm_snapshot` and restore it upon next calls with `evm_revert` but we can't use them with our test structure because of the `evm_snapshot` and `evm_evert` limitations ([source](https://github.com/trufflesuite/ganache-cli-archive#custom-methods)). Long story short, this mechanism can't restore arbitrary snapshots multiple times because each snapshot and all subsequent ones created after are deleted upon `evm_revert` call. To mitigate this problem and use snapshotting with our test structure, `snapshot.ts` stores the snapshots in a stack so multiple nested snapshots can be used seamlessly.

To leverage the full power of snapshots, we also need to replace almost all `beforeEach` hooks we use to set up state across test contexts. Instead of that, we should prefer `before` hooks combined with snapshots. In other words, all constructions like this:
```
beforeEach(async () => {
    // Setup instructions.
})
```
should be now replaced with:
```
before(async () => {
    await createSnapshot()

    // Setup instructions.
})

after(async () => {
    await restoreSnapshot()
})
```
This is functionally the same as the `beforeEach` hook but leverages the snapshotting mechanism to speed up test execution.

Finally, we decrease some contract parameters used in tests that have an impact on test execution time. Parameters like `relayEntryHardTimeout` or `groupLifeTime` are often used with `mineBlocks` to reach the desired chain state. Big values of those parameters have a negative impact on test execution time.

All changes presented in this PR have a positive impact on the test execution time (only test suites with execution time > 30 seconds have been optimized):
| Test suite                 | Before     | After      |
|----------------------------|------------|------------|
| RandomBeacon.Relay.test.ts | 12 minutes | 32 seconds |
| RandomBeacon.GroupCreation.test.ts | 6 minutes | 60 seconds |
| RandomBeacon.Callback.test.ts | 32 seconds | 8 seconds |
| RandomBeaconGovernance.test.ts | 49 seconds | 9 seconds |
| DKGValidator.test.ts | 46 seconds | 7 seconds |

All tests take about 3 minutes now, comparing to around 25 minutes before.


solidity/random-beacon/test/utils/snapshot.ts
```diff
@@ -0,0 +1,15 @@
+import { ethers } from "hardhat"
```
nkuba, line 1: Can we add the code to https://github.com/keep-network/hardhat-helpers, so we can use it in other repos?

lukasz-zimnoch: Opened https://github.com/keep-network/hardhat-helpers/pull/12. I'll use it here once it lands on main.

tomaszslabon:
@lukasz-zimnoch 
This looks really good. There are some conflicts though, probably with [my PR](https://github.com/keep-network/keep-core/pull/2755) where are made revert messages uppercase, but they should be easy to solve.

solidity/random-beacon/test/utils/snapshot.ts
lukasz-zimnoch: Done in https://github.com/keep-network/keep-core/pull/2762/commits/da8599aad54b44eeff66ae764a5376d37576f185

solidity/random-beacon/test/RandomBeacon.Callback.test.ts
```diff
@@ -74,26 +63,23 @@ describe("RandomBeacon - Callback", () => {
...
 
   describe("requestRelayEntry", () => {
-    beforeEach(async () => {
+    before(async () => {
+      await createSnapshot()
+
       await approveTestToken()
     })
 
+    after(async () => {
```
nkuba, line 75: Would it work if we create a snapshot in the `before` block, store the snapshot ID, and then in `afterEach` we would restore a snapshot with a specific ID instead of poping the stack?

lukasz-zimnoch: Nope, as I mentioned in the PR description, `evm_snapshot` and `evm_revert` have some limitations. If you use `evm_revert` with a specific ID, this snapshot is restored but the ID is removed and can't be reused. Also, all IDs created after the removed ID are removed as well. This is why stacking snapshots is the only possible way if we want to use multiple snapshots and fit into our nested test structure.

solidity/random-beacon/test/fixtures/index.ts
```diff
@@ -27,16 +27,16 @@ export const dkgState = {
 }
 
 export const params = {
-  relayRequestFee: 0,
+  relayRequestFee: to1e18(100),
   relayEntrySubmissionEligibilityDelay: 10,
-  relayEntryHardTimeout: 5760,
+  relayEntryHardTimeout: 100,
```
nkuba, line 32: Do the parameters' values affect tests duration? I'd love to keep the values as close as possible to the actually used in the mainnet contracts deployement.

solidity/random-beacon/test/RandomBeacon.Callback.test.ts
nkuba: 😿 
🤷🏻‍♂️

solidity/random-beacon/test/fixtures/index.ts
lukasz-zimnoch: Yep, as I mentioned in the description, the higher value of such a param, the higher `mineBlocks` execution time. This is especially visible in scenarios related to relay entry timeout. This can't be easily optimized with snapshotting because each `mineBlocks` must be invoked at least one time.

tomaszslabon (approved):

PR #2766, opened 2021-12-13
Authorizations: integration with T staking contract

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/2715

Depends on https://github.com/threshold-network/solidity-contracts/pull/40
Depends on https://github.com/keep-network/sortition-pools/pull/142/
Depends on https://github.com/keep-network/sortition-pools/pull/143

This Pull Request is still a work in progress 🏗️ but it presents the general
approach to the problem. Hence the draft status. 👷🏻 

# Overview

This PR proposes an alternative approach to integrating with T staking contract
as discussed in [this](https://github.com/keep-network/keep-core/pull/2747#discussion_r765839410) comment. It requires that authorization increase and authorization
decrease requests will only be performed at the time sortition pool is not locked.

This should be easy to implement in the UI and if we set the parameters right,
there should be significantly more time when the pool is not locked compared to
the time it is locked.

# Functionality

## `authorizationIncreased`

Used by T staking contract to inform the random beacon that the
authorized amount for the given operator increased. Can only be
called when the sortition pool is not locked. Increases in-pool
weight and rewards weight in the pool proportionally to the
authorized stake amount immediately. Reverts if the sortition
pool is locked or if the authorization amount is below the
minimum.

## `authorizationDecreaseRequested`

Used by T staking contract to inform the random beacon that the
authorization decrease for the given operator has been regitered.
Can only be called when the sortition pool is not locked.

If the authorization is decreased to the amount higher than the
minimum authorization, in-pool weight and rewarded stake are
reduced in the sortition pool immediately so that the operator,
during the autorization decrease delay period, can not be
selected to new groups with stake authorization amount higher
then the one to which it is deauthorizing to. Authorized stake
amount remains the same until deauthorization request is
approved.

If the operator is requesting authorization decrease to zero,
in pool-weight is reduced to zero immediately and operator is
removed from the sortition pool. Rewarded stake stays at the
same level as before. Operator, during the authorization
decrease delay period, can not be selected to new groups but
it should still earn rewards based on their last authorized stake
until authorization decrease request is approved.

Overwrites pending authorization decrease request if the one
pending is for the amount higher than the minimum authorization.

Reverts if authorization decrease to 0 is pending.
Reverts if authorization decrease is requested to a non-zero value
below the minimum authorization amount.
Reverts if the sortition pool is locked.

## `approveAuthorizationDecrease`

Approves the previously registered authorization decrease
request. Reverts if authorization decrease delay has not passed
yet or if the auhorization decrease was not requested for the
given operator.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -952,7 +952,7 @@ contract RandomBeacon is Ownable {
     ///         execute slashing for providing a malicious DKG result or when
     ///         a relay entry times out.
     function minimumAuthorization() external view returns (uint96) {
-        return authorization.minimumAuthorization;
+        return authorization.parameters.minimumAuthorization;
     }
 
     /// @notice Delay in seconds that needs to pass between the time
@@ -960,7 +960,7 @@ contract RandomBeacon is Ownable {
     ///         gets approved. Protects against free-riders earning rewards and
     ///         not being active in the network.
     function authorizationDecreaseDelay() external view returns (uint64) {
-        return authorization.authorizationDecreaseDelay;
+        return authorization.parameters.authorizationDecreaseDelay;
     }
 
     /// @return Flag indicating whether a relay entry request is currently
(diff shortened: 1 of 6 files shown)


pdyraga:
Replaced by https://github.com/keep-network/keep-core/pull/2903.

Issue #2756, opened 2021-12-06
DKG: optimize gas cost of storing group members


pdyraga:
# Problem

DKG on-chain process for random beacon v2 is performed in two steps: the result is submitted by an eligible group member, and after the challenge period is over, the result is getting approved. Optionally, in case the result submitted is found malicious by off-chain maintainers, anyone can submit a challenge, and the submitted result is getting discarded.

For a group size of 64 members, the cost of gas (min-max-avg) is:
```
|  approveDkgResult     ·     201748  ·     247436  ·     213813  │
|  challengeDkgResult   ·     325448  ·    1178122  ·    1019890  │
|  submitDkgResult      ·     339955  ·     637080  ·     567450  │
```

The same code, without any modifications, for a group size of 100 members, gives the following gas cost:
```
|  approveDkgResult     ·     239940  ·     285597  ·     252059  |
|  RandomBeaconStub     ·     439904  ·    1818890  ·    1564474  |
|  RandomBeaconStub     ·     433635  ·     821141  ·     742949  |
```

For a group size of 64 members, the average total gas cost of a happy path DKG is ~780k gas total which is a significant expense. That's why in #2749, we proposed to create a new beacon group every 5 weeks. In tBTC v2, we expect that a new wallet will be created every one week and 995k average total gas cost of a happy path of 100 group members DKG is a pill hard to swallow.

# Proposed solution

One of the most significant costs is storing group member identifiers on-chain in `Groups.addCandidateGroup->setGroupMembers`. Without storing group members on-chain, the cost of `submitDkgResult` goes down by 192k gas, to `374620` on average for a group size of 64 members, and goes down by 326k gas, to `416845` for a group size of 100 members. This gives 588k and 669k average total DKG gas cost respectively. 

[...]

PR #2768, opened 2021-12-15
Optimizing gas cost by removing group members storage

dimpar:
Ref https://github.com/keep-network/keep-core/issues/2756

Storing all members on-chain is expensive. The idea behind this change is to store a hash of dkg group members that actively participated in dkg.

For a group size of 64 members, the cost of gas (min-max-avg) is:

Before:
```
|  approveDkgResult     ·     185764  ·     247364  ·     206239  │
|  challengeDkgResult   ·     333425  ·    1175723  ·     977292  │
|  submitDkgResult      ·     340003  ·     637152  ·     544857  │
|  submitRelayEntry     ·     192481  ·     488134  ·     246530  │
```

Cost of `setGroupMembers()` when running tests in `Groups.test.ts -> addCandidateGroup` context 
- no misbehaved: ~224210
- with misbehaved: ~226353

After:
```
|  approveDkgResult     ·     185742  ·     247342  ·     206217  │
|  challengeDkgResult   ·     333403  ·    1175701  ·     978441  │
|  submitDkgResult      ·     306979  ·     479157  ·     394890  │
|  submitRelayEntry     ·     214082  ·     495202  ·     266195  │
```

Cost of `hashGroupMembers()` when running tests in `Groups.test.ts -> addCandidateGroup` context
- no misbehaved: ~31975
- with misbehaved: ~62058

Cost of: `keccak256(abi.encode(members))`: ~30189


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -728,7 +730,7 @@ contract RandomBeacon is Ownable {
 
         if (slashingAmount > 0) {
             address[] memory groupMembers = sortitionPool.getIDOperators(
-                group.members
+                members
```
nkuba, line 733: Please remember about checks if provided members list corresponds to the hash stored on DKG result approval.

dimpar: Sure, added checks

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -22,7 +22,8 @@ library Groups {
     struct Group {
         bytes groupPubKey;
         uint256 activationBlockNumber;
-        uint32[] members;
+        // Hash of active group members identifiers array (excluding IA/DQ members).
+        bytes32 activeMembersHash;
```
nkuba, line 26: As this is declared in the `Group` struct context this could be `membersHash`
```suggestion
        bytes32 membersHash;
```

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -58,17 +59,12 @@ library Groups {
...
-    ///        members in the group, as outputted by the group selection
-    ///        protocol.
+    /// @param activeMembers Addresses of active candidate group members as outputted
+    ///        by the group selection protocol.
     function addCandidateGroup(
         Data storage self,
         bytes calldata groupPubKey,
-        uint32[] calldata members,
-        uint8[] calldata misbehavedMembersIndices
+        uint32[] calldata activeMembers
```
nkuba, line 67:
```suggestion
        uint32[] calldata groupMembers
```

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -87,8 +83,7 @@ library Groups {
         // candidate group was already registered before and popped.
         Group storage group = self.groupsData[groupPubKeyHash];
         group.groupPubKey = groupPubKey;
-
-        setGroupMembers(group, members, misbehavedMembersIndices);
+        group.activeMembersHash = keccak256(abi.encode(activeMembers));
```
nkuba, line 86:
```suggestion
        group.membersHash = keccak256(abi.encode(groupMembers));
```

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -342,41 +337,4 @@ library Groups {
 
         return shiftedIndex;
     }
```
nkuba, line 339: I would leave `setGroupMembers` untouched in this PR.

solidity/random-beacon/contracts/libraries/Heartbeat.sol
```diff
@@ -66,14 +66,13 @@ library Heartbeat {
         SortitionPool sortitionPool,
         FailureClaim calldata claim,
         Groups.Group storage group,
-        uint256 nonce
+        uint256 nonce,
+        uint32[] calldata members
```
nkuba, line 70:
```suggestion
        uint32[] calldata groupMembersIds
```

solidity/random-beacon/contracts/libraries/Heartbeat.sol
```diff
@@ -66,14 +66,13 @@ library Heartbeat {
         SortitionPool sortitionPool,
         FailureClaim calldata claim,
         Groups.Group storage group,
-        uint256 nonce
+        uint256 nonce,
```
nkuba, line 69: We need to update doc above to reflect the function signature.

solidity/random-beacon/contracts/libraries/Heartbeat.sol
```diff
@@ -66,14 +66,13 @@ library Heartbeat {
...
+        uint32[] calldata members
     ) external view returns (uint32[] memory failedMembers) {
-        uint256 groupSize = group.members.length;
-
         // Validate failed members indices. Maximum indices count is equal to
```
nkuba, line 72: Aren't we missing list vs hash validation here?

solidity/random-beacon/contracts/test/GroupsStub.sol
```diff
@@ -16,14 +16,9 @@ contract GroupsStub {
 
     function addCandidateGroup(
         bytes calldata groupPubKey,
-        uint32[] calldata members,
-        uint8[] calldata misbehavedMembersIndices
+        uint32[] calldata activeMembers
```
nkuba, line 19: This change is a subject to another PR I believe.

solidity/random-beacon/contracts/DKGValidator.sol
```diff
@@ -189,6 +195,42 @@ contract DKGValidator {
...
+            uint256 misbehavedMemberArrayPosition = result
+                .misbehavedMembersIndices[i - 1] - 1;
+            for (uint256 j = 0; j < result.activeMembers.length; j++) {
+                // active members array should not contain any misbehaved members
+                if (
+                    result.members[misbehavedMemberArrayPosition] ==
+                    result.activeMembers[j]
+                ) {
+                    return false;
+                }
```
pdyraga, line 227: I do not think this validation rule is correct. One operator could have multiple seats in a group. It is possible that one member from that operator misbehaved and another did not. Then, we should expect only one of them to be removed from the group. That's how it worked in v1 and that's how off-chain DKG is implemented.

solidity/random-beacon/contracts/libraries/Heartbeat.sol
pdyraga: It is validated earlier, in `RandomBeacon.notifyFailedHeartbeat`: 

https://github.com/keep-network/keep-core/blob/d830ff39a27cbc0657f780ba16f5356aaf7964c7/solidity/random-beacon/contracts/RandomBeacon.sol#L905-L908

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -82,6 +82,9 @@ library DKG {
         // Identifiers of candidate group members as outputted by the group
         // selection protocol.
         uint32[] members;
+        // Identifiers of candidate group members as outputted by the group
+        // selection protocol excluding IA/DQ members
+        uint32[] activeMembers;
```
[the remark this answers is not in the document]
pdyraga: What is the gas difference in `submitDkgResult` when we need to filter out group members compared to the savings we are getting in this approach and the gas cost of the challenge?

I would like to see concrete numbers before we decide on this change, especially on 64 and 100 members group with minimum, average, and maximum possible number of misbehaving members.

It is possible this change gives great results but it is also possible that the challenge cost could get increased significantly, we need to know that for sure.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -717,18 +717,27 @@ contract RandomBeacon is Ownable {
 
     /// @notice Creates a new relay entry.
     /// @param entry Group BLS signature over the previous entry.
-    function submitRelayEntry(bytes calldata entry) external {
+    /// @param activeMembers Active group member ids array.
+    function submitRelayEntry(
+        bytes calldata entry,
+        uint32[] calldata activeMembers
```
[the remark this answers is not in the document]
pdyraga: 100% agreed. Adding one more point: in case of `submitRelayEntry`, `reportRelayEntryTimeout`, and similar, the additional parameter should be named just `groupMembers`.

Would love us to first asses the impact before we actually decide on this change. This code was already audited.

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -122,14 +125,15 @@ library DKG {
         uint8[] misbehavedMembersIndices,
         bytes signatures,
         uint256[] signingMembersIndices,
-        uint32[] members
+        bytes32 membersHash
```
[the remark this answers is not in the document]
pdyraga: The last time I checked, emitting `uint32[] members` in an event had a significant gas cost, it does not come for free and it is possible some savings we got from the change in this PR come from this change.

As much as I'd like to make it cheaper, I do not know if there is an alternative if we do not want to execute view functions on a previous chain state to know who is in the group.

solidity/random-beacon/contracts/libraries/DKG.sol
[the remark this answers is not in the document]
pdyraga: One advantage of doing the change proposed in this PR I see is that - depending on the implementation in the challenge DKG result function in the validator - we could preserve the order of members, as returned by the sortition pool and just filter out misbehaved members instead of changing the order of group members (see the node in `setGroupMembers` function).

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -760,12 +769,19 @@ contract RandomBeacon is Ownable {
     }
 
     /// @notice Reports a relay entry timeout.
-    function reportRelayEntryTimeout() external {
+    /// @param activeMembers Active group member ids array.
+    function reportRelayEntryTimeout(uint32[] calldata activeMembers) external {
         uint64 groupId = relay.currentRequestGroupID;
+        Groups.Group memory group = groups.getGroup(groupId);
```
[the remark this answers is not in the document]
dimpar: Interesting, looks like it is cheaper to use `storage` for a local struct var.

`Groups.Group memory group = groups.getGroup(groupId);` gas: (min-max-avg: 417387, 430320, 424182)

`Groups.Group storage group = groups.getGroup(groupId);` gas: (min-max-avg: 405745,  418678, 412540)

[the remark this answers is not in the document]
pdyraga: It is not copying the structure into the memory but works like a pointer.

solidity/random-beacon/contracts/DKGValidator.sol
dimpar: Ok, for simplicity say the group size is 10. We have:
`misbehavedMembersIndices = [1,7,3]`
`members = [2,4,1,5,2,8,7,9,5,4]`

`members[1] = 4`
`members[7] = 9`
`members[3] = 5`

These ^ are the ids of misbehaved members. Ex:
- `activeMembers = [2,3,1,2,8,7,9,6,8]` -> invalid active members, `9` misbehaved
- `activeMembers = [2,3,1,2,8,7,4,6,8]` -> invalid active members, `4` misbehaved
- `activeMembers = [2,3,1,4,8,7,4,6,8]` -> invalid active members. `4` id is present twice, but it is the same member id that misbehaved.
- `activeMembers = [2,3,1,6,8,7,6,8]` -> valid active members, no misbehaved

In this function we do not remove anything, it returns `true/false` if the `activeMembers` array `valid/invalid. Can you pls point where's a flaw here, so I can understand it better using this example?

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: @nkuba I'm not sure if we can leave it untouched based on the proposed solution in #2756 
> Instead of storing group member identifiers as uint32[] on-chain, we could store a hash of group member identifiers array and expect uint32[] identifier array to be submitted for every function requiring it.

and later
> `DkgResultApproved` should contain uint32[] members field that should have the final value (after eliminating misbehaving IA/DQ members)

My understanding is that we need to filter out IA/DQ before we set a hash of members that participated in dkg. `DkgResultApproved` will also emit members without IA/DQ members. In the following functions: `submitRelayEntry`, `reportRelayEntryTimeout`, `reportUnauthorizedSigning`, `notifyFailedHeartbeat` one of the params will be an array of members that participated in dkg that was emitted in `DkgResultApproved`.

Here's what this function can look like after the change: 1c429b3 I didn't test it yet, but the idea is to hash only those members that participated in dkg (without IA/DQ)

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: I see two problems with the current implementation.

First one:
```
for (uint256 i = 0; i < result.misbehavedMembersIndices.length; i++) {
  // group member indices start from 1, so we need to -1 on misbehaved
  uint256 misbehavedMemberArrayPosition = result
    .misbehavedMembersIndices[i - 1] - 1;
```

What is `result.misbehavedMembersIndices[i - 1]` for `i = 0`?

I am surprised tests are green because Solidity documentation says:

> By default, arithmetic is always “checked”, which mean that if the result of an operation falls outside the value range of the type, the call is reverted through a failing assertion. 

https://docs.soliditylang.org/en/v0.8.10/types.html#integers

Similar example is in [this article](https://blog.soliditylang.org/2020/10/28/solidity-0.8.x-preview/) in *Actual Checked Arithmetic* section.

It is possible we do not have it properly covered in unit tests but I'd be surprised if that's the case given that it should fail on the simplest case with even just one misbehaved member. 🤔 



Second problem:

The current implementation is going through `misbehavedMemberIndices`, turns them into operator identifiers, and checks if these operator identifiers are not present in `activeMembers`.

This is not how `setGroupMembers` worked and it is not how DKG protocol is designed and implemented.

Using your example:

```
members = [2,4,1,5,2,8,7,9,5,4]
misbehavedMembersIndices = [1,7,3]
```
so:
```
 1 2 3 4 5 6 7 8 9 10
[2,4,1,5,2,8,7,9,5,4]
 x   x       x
```

Note that `misbehavedMemberIndices` are 1-indexed, just like member IDs and that's why we need to `-1`, see [here](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/contracts/libraries/DKG.sol#L65-L68).

Active members (without changing the order) should be: `[4,5,2,8,9,5,4]`.

@nkuba to double-check

dimpar: Just to note, this potential change is meant to be implemented in a separate PR which was agreed on Discord and in some [comments](https://github.com/keep-network/keep-core/pull/2768#discussion_r773103929). In this discussion, I just wanted to make sure if we want to explore this further and if there are some major flaws.

As for the first problem, there are some background findings by me after I opened this PR. One of them is test coverage of this verification step. I've changed a bunch of tests in this PR but overlooked covering this particular function. It will fail just like you said, and `misbehavedMembersIndices[i - 1] - 1;` should be probably replaced with `misbehavedMembersIndices[i] - 1;` (TBD in a separate PR).

As for the second problem, I've skipped `-1` for simplicity, just wanted to discuss a general idea. Implementation includes that. From your example, I see that what I was missing is if `activeMembers` is trimmed/replaced with some valid members like `[4,5,2,8,9]` (`[5,4]` are missing), then the verification would still pass, but it shouldn't. The `activeMembers` should be exactly the subset of `members` minus those that misbehaved, not more and not less, correct?

solidity/random-beacon/contracts/RandomBeacon.sol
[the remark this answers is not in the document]
dimpar: Yea, makes sense. All the points 1-4 seem correct. I've renamed these vars here: 4c847d3

solidity/random-beacon/contracts/libraries/DKG.sol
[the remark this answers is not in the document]
dimpar: Yes, I'll try to extract gas usage in a separate PR for comparison.

solidity/random-beacon/contracts/libraries/DKG.sol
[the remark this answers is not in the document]
dimpar: Yes, the fewer events we have, the cheaper it is. I will try to track the exact cost of emitting such an event but need to fix the tests first.

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: done

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: done

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: done

solidity/random-beacon/contracts/libraries/Heartbeat.sol
dimpar: done

solidity/random-beacon/contracts/libraries/Heartbeat.sol
dimpar: Added docs for members

solidity/random-beacon/contracts/libraries/Heartbeat.sol
dimpar: We validate it upstream just like Piotr mentioned

solidity/random-beacon/contracts/test/GroupsStub.sol
dimpar: agreed

solidity/random-beacon/contracts/DKGValidator.sol
pdyraga: Let me rephrase: we are marking group members with the particular indexes as misbehaving. We are not marking operators with the particular ids as misbehaving. That's the difference. 

In other words, it is possible that the operator 0xABC with identifier (from sortition pool) 1337 could be selected to the group at three positions: 1, 50, 64, and it is misbehaving at positions 1 and 50. It means that we should remove operator 1337 at positions 1 and 50 but leave them at position 64.

dimpar: Right, I think I understood you. 
```
members = [2,4,1,5,2,8,7,9,5,4]
operator 0xABC -> sortition pool id 1337 was selected to the group at positions [1,7,3]
but this operator is misbehaving at [1,7] 
```
marking [1,7] as misbehaved but leaving an index [3]
```
 1 2 3 4 5 6 7 8 9 10
[2,4,1,5,2,8,7,9,5,4]
 x           x
```
the active members array should be exactly:
```
[4,1,5,2,8,9,5,4]
```

solidity/random-beacon/contracts/libraries/DKG.sol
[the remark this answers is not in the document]
nkuba: It's crucial for the challenge submission to know exactly what was submitted as DKG Result. A DKG result submitter can provide any set of members, which would be needed for a challenge. Knowing just the hash wouldn't be enough, the challenger should pull the members' list from the DKG result submission transaction input parameters, which adds complexity. We should first ensure that it will be possible to obtain all the required information from the transaction, if they are not emitted in the event.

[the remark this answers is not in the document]
nkuba: The emitted event should have a transaction hash, we could get the transaction by the hash and try decoding input parameters. It's more complicated than emitting data in an event, but possible. This way we could reduce costs even more by cutting out other parameters from events, that duplicate transaction's input paramters.

solidity/random-beacon/contracts/libraries/Heartbeat.sol
nkuba: Great! Can we add a comment inline here to clarify?

solidity/random-beacon/contracts/DKGValidator.sol
nkuba: Please also note the specific way how `setGroupMembers` removes entries by replacing them with poped from the end of the list.
So in the example above it would result with:
```
[4,4,1,5,2,8,5,9]
```

solidity/random-beacon/contracts/libraries/DKG.sol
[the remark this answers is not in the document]
dimpar: That's true, we need group members during the challenge step [here](https://github.com/keep-network/keep-core/blob/group-members-storage-optimization/solidity/random-beacon/contracts/DKGValidator.sol#L188). We do emit all the members in `DkgResultApproved` event, maybe we can emit members (excluding IA/DQ) in `DkgResultSubmitted`, but emit only `hash(members)` in the `DkgResultApproved` event. I'll see what's the cost of that.

solidity/random-beacon/contracts/DKGValidator.sol
dimpar: Yes, I noticed, but I thought we can pass members that participated in dkg in order that we specify, no? I know that the `hash([4,1,5,2,8,9,5,4]) != hash([4,4,1,5,2,8,5,9])`, but can't we filter all members off-chain by just removing misbehaved, so the original order is preserved? Additional shuffling of members adds complexity on-chain, which most likely adds gas. Unless I'm not aware of something and it has to be in the order you've mentioned.

solidity/random-beacon/contracts/libraries/DKG.sol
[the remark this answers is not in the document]
dimpar: gas used for `DkgResultSubmitted` with `hash(result.members)` for group size 64: 44980

gas used for `DkgResultSubmitted` with `result.members` for group size 64: 61065

The diff is 16k and having in mind that it would be easier to read the data from event rather than the input params from the DKG result submission, I think we should leave `uint32[] members`.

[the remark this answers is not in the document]
dimpar: Here's the change 2b583e9 What I'm not 100% sure about is if we should emit members hash with or without misbehaved members in `DkgResultApproved` event. Thoughts? For submitters or callers that would pass members hash in the functions param, it would be easier, but on our side, we'll add gas to filter out misbehaved members. On the other hand, if the second PR lands in main which deals with already filtered out members, then we won't be needing filtering out members here.

solidity/random-beacon/contracts/libraries/Heartbeat.sol
dimpar: fa1bd7f

dimpar:
@nkuba Ready for another look. I've also updated the gas usage table in the PR description before and after the change. cc @pdyraga

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -232,7 +232,8 @@ contract RandomBeacon is Ownable {
 
     event DkgResultApproved(
         bytes32 indexed resultHash,
-        address indexed approver
+        address indexed approver,
+        bytes32 membersHash
```
pdyraga, line 236: Why do we need `membersHash` in this event?

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -717,26 +718,35 @@ contract RandomBeacon is Ownable {
...
+            group.membersHash == keccak256(abi.encode(groupMembers)),
+            "Invalid group members"
+        );
+
         uint256 slashingAmount = relay.submitEntry(entry, group.groupPubKey);
 
         if (slashingAmount > 0) {
-            address[] memory groupMembers = sortitionPool.getIDOperators(
-                group.members
+            address[] memory activeOperators = sortitionPool.getIDOperators(
```
pdyraga, line 740: Why `activeOperators` and not just `groupMemberIds`? `groupMembers` does not necessarily contain only active group members.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -717,26 +718,35 @@ contract RandomBeacon is Ownable {
 
     /// @notice Creates a new relay entry.
     /// @param entry Group BLS signature over the previous entry.
-    function submitRelayEntry(bytes calldata entry) external {
+    /// @param groupMembers Group member ids that participated in dkg (excluding IA/DQ).
+    function submitRelayEntry(
```
pdyraga, line 722: Given that the cost of `submitRelayEntry` increased by `21k`, can we add an alternative `submitRelayEntry` function that accepts no `groupMembers` and can be called only before the soft timeout? 

Let's see what is overall gas cost of this change to the system.

dimpar: Yes, I've actually started doing it as part of this PR in commit 61977c2 but reverted in beb7377 since there were already a lot of changes in the current PR. Any objections if I open a separate PR for `submitRelayEntry` split?

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: Added to verify members against `DkgResultSubmitted`, but you're right, we have `resultHash` which include members. Looks like it will require a bit more lines of code to verify dkg members for contract callers if needed, but we save gas on not hashing members and not emitting them. 075d32c

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: [This](https://github.com/keep-network/keep-core/pull/2768#discussion_r773098164) comment
> So the list selected by the sortition pool (1) could be: selectedMembers or selectedOperators or operators, especially that in the Sortition Pool we use a term operator for the parties.
2.a. Could be activeOperators
2.b. Could be misbehavedOperators.

Ok, I thought they are active. Since we're calling sortition pool then maybe we can name them `operators`?

solidity/random-beacon/contracts/RandomBeacon.sol
pdyraga: Is it that much work? If we want it to be a separate PR, let's do not merge this one until we have full clarity about the costs. That is, merge two PRs one after another.

dimpar: Draft PR is [here](https://github.com/keep-network/keep-core/pull/2772). I still need to modify a couple of tests there, but we can compare the gas usage.

nkuba:
Can we update the PR description? I believe this part no longer holds:
>Moreover, filtering active group members can be also done off-chain to
simplify things on-chain. In terms of verification, it can be done
during the challenge phase on-chain against all members and misbehaved
members that can be verified beforehand against the sortition pool.

solidity/random-beacon/contracts/RandomBeacon.sol
nkuba: We're getting IDs for `groupMembers`, so naming it `groupMemberIds` makes sense to me.

pdyraga: ☝🏻 they are not selected by the sortition pool; the sortition pool is just keeping track of IDs

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -343,40 +344,46 @@ library Groups {
...
-        uint256 i = misbehavedMembersIndices.length;
-        while (i > 0) {
+        if (misbehavedMembersIndices.length > 0) {
+            // members that generated a group signing key
+            uint32[] memory groupMembers = new uint32[](
+                members.length - misbehavedMembersIndices.length
+            );
+            uint256 k = 0; // misbehaved members counter
+            uint256 j = 0; // group members counter
             // group member indices start from 1, so we need to -1 on misbehaved
```
nkuba, line 369:
```suggestion
            // misbehaved members indices start from 1, so we need to -1 on misbehaved
```

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -343,40 +344,46 @@ library Groups {
...
+                if (i != misbehavedMemberArrayPosition) {
+                    groupMembers[j] = members[i];
+                    j++;
+                } else if (k < misbehavedMembersIndices.length - 1) {
+                    k++;
+                    misbehavedMemberArrayPosition =
+                        misbehavedMembersIndices[k] -
+                        1;
+                }
+            }
```
nkuba, line 382: Would this work fine if we get rid of `misbehavedMemberArrayPosition` like in suggestion below?
```suggestion
            for (uint256 i = 0; i < members.length; i++) {
                if (i != misbehavedMembersIndices[k] - 1) {
                    groupMembers[j] = members[i];
                    j++;
                } else if (k < misbehavedMembersIndices.length - 1) {
                    k++;
                }
            }
```

solidity/random-beacon/test/Groups.test.ts
```diff
@@ -62,91 +62,92 @@ describe("Groups", () => {
 
           expect(storedGroup.groupPubKey).to.be.equal(groupPublicKey)
           expect(storedGroup.activationBlockNumber).to.be.equal(0)
-          expect(storedGroup.members).to.be.deep.equal(members)
+          expect(storedGroup.membersHash).to.be.equal(
+            keccak256(defaultAbiCoder.encode(["uint32[]"], [members]))
```
nkuba, line 66: Could this hash calculation be pulled to a util function? E.g. in `test/utils/groups.ts`.

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -1685,11 +1739,30 @@ describe("RandomBeacon - Relay", () => {
...
             },
-            1
+            1,
+            membersIDs
           ) // Initial nonce is `0`.
         ).to.be.revertedWith("Invalid nonce")
       })
     })
+
+    context("when active members are invalid", () => {
```
nkuba, line 1749:
```suggestion
    context("when group members are invalid", () => {
```

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -772,10 +799,22 @@ describe("RandomBeacon - Relay", () => {
         await expect(
           randomBeacon
             .connect(notifier)
-            .reportUnauthorizedSigning(notifierSignature, 0)
+            .reportUnauthorizedSigning(notifierSignature, 0, membersIDs)
         ).to.be.revertedWith("Invalid signature")
       })
     })
+
+    context("when active members are not valid", () => {
```
nkuba, line 807:
```suggestion
    context("when group members are invalid", () => {
```

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -669,9 +687,18 @@ describe("RandomBeacon - Relay", () => {
...
-        await expect(randomBeacon.reportRelayEntryTimeout()).to.be.revertedWith(
-          "Relay request did not time out"
-        )
+        await expect(
+          randomBeacon.reportRelayEntryTimeout(membersIDs)
+        ).to.be.revertedWith("Relay request did not time out")
+      })
+    })
+
+    context("when active members are invalid", () => {
```
nkuba, line 696:
```suggestion
    context("when group members are invalid", () => {
```

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -416,10 +417,21 @@ describe("RandomBeacon - Relay", () => {
             await expect(
               randomBeacon
                 .connect(submitter)
-                .submitRelayEntry(blsData.nextGroupSignature)
+                .submitRelayEntry(blsData.nextGroupSignature, membersIDs)
             ).to.be.revertedWith("Invalid entry")
           })
         })
+
+        context("when active members are not valid", () => {
```
nkuba, line 425:
```suggestion
        context("when group members are invalid", () => {
```

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -270,6 +270,8 @@ library DKG {
         self.submittedResultHash = keccak256(abi.encode(result));
         self.submittedResultBlock = block.number;
 
+        // To recreate the members that took part in dkg, the result.members array
+        // should be filtered from result.misbehavedMembersIndices
```
nkuba, line 274:
```suggestion
        // To recreate the members that actively took part in dkg, the members array
        // should be filtered out from misbehavedMembersIndices.
```

solidity/random-beacon/contracts/RandomBeacon.sol
nkuba: Actually, aren't we converting IDs to addresses here? Shouldn't the variable be called `groupMembersAddresses`?

solidity/random-beacon/contracts/libraries/DKG.sol
dimpar: e538ed6

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
dimpar: e538ed6

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
dimpar: e538ed6

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
dimpar: e538ed6

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
dimpar: e538ed6

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: e538ed6

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -829,21 +852,21 @@ contract RandomBeacon is Ownable {
 
         groups.terminateGroup(groupId);
 
-        address[] memory groupMembers = sortitionPool.getIDOperators(
-            groups.getGroup(groupId).members
+        address[] memory activeOperators = sortitionPool.getIDOperators(
```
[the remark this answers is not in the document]
dimpar: e538ed6

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -760,25 +769,32 @@ contract RandomBeacon is Ownable {
...
+
+        require(
+            group.membersHash == keccak256(abi.encode(groupMembers)),
+            "Invalid group members"
+        );
         uint256 slashingAmount = relay
             .relayEntrySubmissionFailureSlashingAmount;
-        address[] memory groupMembers = sortitionPool.getIDOperators(
-            groups.getGroup(groupId).members
+        address[] memory activeOperators = sortitionPool.getIDOperators(
```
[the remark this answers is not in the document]
dimpar: e538ed6

solidity/random-beacon/contracts/DKGValidator.sol
```diff
@@ -217,7 +217,7 @@ contract DKGValidator {
             signingMemberIds[i] = result.members[signingMembersIndices[i] - 1];
         }
 
-        address[] memory signingMemberAddresses = sortitionPool.getIDOperators(
+        address[] memory selectedOperators = sortitionPool.getIDOperators(
```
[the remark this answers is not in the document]
dimpar: e538ed6

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: Yup 08c8a4b

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -343,40 +344,46 @@ library Groups {
...
     /// @param group The group storage.
     /// @param members Group member addresses as outputted by the group selection
     ///        protocol.
     /// @param misbehavedMembersIndices Array of misbehaved (disqualified or
     ///        inactive) group members. Indices reflect positions
     ///        of members in the group as outputted by the group selection
-    ///        protocol.
-    function setGroupMembers(
+    ///        protocol. Indices must be in ascending order.
+    function setGroupMembersHash(
```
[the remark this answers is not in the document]
dimpar: Gas wise it's the same, I checked. But I agree, it looks cleaner if we just return the hash. 08c8a4b

solidity/random-beacon/test/Groups.test.ts
dimpar: 5f5de44

dimpar:
@nkuba comments are addressed

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: e538ed6

solidity/random-beacon/test/system/e2e.test.ts
```diff
@@ -98,6 +98,7 @@ describe("System -- e2e", () => {
 
   context("when testing a happy path with 15 relay requests", () => {
     let groupPubKeyCounter = 0
+    const activeGroupMembers = []
```
nkuba, line 101:
```suggestion
    const groupMembers = []
```

solidity/random-beacon/test/system/e2e.test.ts
```diff
@@ -167,6 +174,7 @@ describe("System -- e2e", () => {
             txSubmitRelayEntry.blockNumber,
             noMisbehaved
           )
+          activeGroupMembers.push(dkgResult.members)
```
nkuba, line 177: I don't understand fully what this test does. Could you please clarify, why are we pushing new entries to the `activeGroupMembers` array? With each loop, we're adding more entries, shouldn't we just set the array completely `activeGroupMembers = dkgResult.members`?

dimpar: We have 15 requests and for every 5th request, there's a new dkg that happens. In total, we'll have 3 groups that can be chosen to participate in dkg. It mimics the group selection logic which is implemented in `selectGroup()`. In this example, say 12th request will choose a group number 1 and we need to keep track of all the groups (and members inside) so we can pass those members in `submitRelayEntry()`. That's why we have a `groupMembers` array here.

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -1076,7 +1080,9 @@ describe("RandomBeacon - Group Creation", () => {
 
               expect(storedGroup.groupPubKey).to.be.equal(groupPublicKey)
               expect(storedGroup.activationBlockNumber).to.be.equal(0)
-              expect(storedGroup.members).to.be.deep.equal(dkgResult.members)
+              expect(storedGroup.membersHash).to.be.equal(
+                hashUint32Array(dkgResult.members)
+              )
 
               await restoreSnapshot()
             })
```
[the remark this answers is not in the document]
dimpar: Yes, we can challenge the dkg result, except if we pass a non-ascending order it won't be accepted here. It will revert with `array out of bounds`. This code works and assumes the [ascending order](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/contracts/libraries/DKG.sol#L66). 
I've added a test that verifies `array out of bounds` msg. Dkg challange for non-ordered misbehaved array wasn't touched in this PR and there are already tests that cover it.

solidity/random-beacon/test/system/e2e.test.ts
dimpar: 0f9ceae

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -1076,7 +1080,9 @@ describe("RandomBeacon - Group Creation", () => {
 
               expect(storedGroup.groupPubKey).to.be.equal(groupPublicKey)
               expect(storedGroup.activationBlockNumber).to.be.equal(0)
-              expect(storedGroup.members).to.be.deep.equal(dkgResult.members)
+              expect(storedGroup.membersHash).to.be.equal(
+                hashUint32Array(dkgResult.members)
+              )
 
               await restoreSnapshot()
```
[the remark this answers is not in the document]
dimpar: aebc2c9

solidity/random-beacon/test/RandomBeacon.Callback.test.ts
```diff
@@ -1,6 +1,9 @@
+/* eslint-disable no-await-in-loop */
```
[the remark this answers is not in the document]
dimpar: d02b1f0

solidity/random-beacon/test/Groups.test.ts
```diff
@@ -62,91 +63,82 @@ describe("Groups", () => {
...
-          it("should filter out misbehaved members", async () => {
             const expectedMembers = [...members]
-            expectedMembers[0] = expectedMembers.pop()
+            expectedMembers.splice(0, 1)
 
             expect(
-              (await groups.getGroup(groupPublicKey)).members
-            ).to.be.deep.equal(expectedMembers)
+              (await groups.getGroup(groupPublicKey)).membersHash
+            ).to.be.equal(hashUint32Array(expectedMembers))
```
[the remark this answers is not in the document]
dimpar: 6ec7bbf

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -343,40 +344,40 @@ library Groups {
...
+    ///         hashing.
     /// @param members Group member addresses as outputted by the group selection
     ///        protocol.
     /// @param misbehavedMembersIndices Array of misbehaved (disqualified or
     ///        inactive) group members. Indices reflect positions
     ///        of members in the group as outputted by the group selection
-    ///        protocol.
-    function setGroupMembers(
-        Group storage group,
+    ///        protocol. Indices must be in ascending order.
```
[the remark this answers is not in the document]
dimpar: d9cbd48

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -22,7 +22,8 @@ library Groups {
     struct Group {
         bytes groupPubKey;
         uint256 activationBlockNumber;
-        uint32[] members;
+        // Hash of group members identifiers array (excluding IA/DQ members).
```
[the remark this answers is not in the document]
dimpar: d9cbd48

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -760,25 +769,32 @@ contract RandomBeacon is Ownable {
...
-    function reportRelayEntryTimeout() external {
+    /// @param groupMembers Group member ids that participated in dkg (excluding IA/DQ).
+    function reportRelayEntryTimeout(uint32[] calldata groupMembers) external {
         uint64 groupId = relay.currentRequestGroupID;
+        Groups.Group storage group = groups.getGroup(groupId);
+
+        require(
+            group.membersHash == keccak256(abi.encode(groupMembers)),
+            "Invalid group members"
+        );
```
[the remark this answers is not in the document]
dimpar: d9cbd48

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -1254,41 +1260,92 @@ describe("RandomBeacon - Group Creation", () => {
...
+                    groupPublicKey
+                  )
+
+                  const expectedMembers = [...dkgResult.members]
+                  expectedMembers.splice(1, 1) // index -1
+                  expectedMembers.splice(7, 1) // index -2 (cause expectedMembers already shrinked)
+                  expectedMembers.splice(8, 1) // index -3
+                  expectedMembers.splice(26, 1) // index -4
+                  expectedMembers.splice(55, 1) // index -5
+                  expectedMembers.splice(58, 1) // index -6
```
pdyraga, line 1313: We could add a comment above that says misbehaved indices are `[2, 9, 11, 30, 60, 64]`. This could help to understand the magic happening here.

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -88,7 +89,7 @@ library Groups {
         Group storage group = self.groupsData[groupPubKeyHash];
         group.groupPubKey = groupPubKey;
 
-        setGroupMembers(group, members, misbehavedMembersIndices);
+        group.membersHash = hashGroupMembers(members, misbehavedMembersIndices);
```
pdyraga, line 92: Worth clarifying in `@param misbehavedMembersIndices` docs for `addCandidateGroup`, what is the expected ordering of elements.

solidity/random-beacon/contracts/libraries/Heartbeat.sol
```diff
@@ -56,24 +56,26 @@ library Heartbeat {
...
     ///         `FailureClaim` struct documentation. Reverts if verification
     ///         fails.
+    /// @dev Group members hash is validated upstream in
+    ///      RandomBeacon.notifyFailedHeartbeat()
     /// @param sortitionPool Sortition pool used by the application performing
     ///        claim verification.
     /// @param claim Failure claim.
     /// @param group Group raising the claim.
     /// @param nonce Current nonce for group used in the claim.
+    /// @param members Group member ids that participated in dkg (excluding IA/DQ).
```
pdyraga, line 66: Same comment to the documentation of this param as in `RandomBeacon.sol`.

solidity/random-beacon/contracts/libraries/Heartbeat.sol
```diff
@@ -107,8 +109,8 @@ library Heartbeat {
             )
         ).toEthSignedMessageHash();
 
-        address[] memory groupMembers = sortitionPool.getIDOperators(
-            group.members
+        address[] memory selectedOperators = sortitionPool.getIDOperators(
```
pdyraga, line 112: I would call them `groupMembersAddresses`. These are not operators that would be selected by the sortition pool right now and these could not be the operators selected by the sortition pool in the past in case someone misbehaved during DKG. We are calling sortition pool to return operator addresses based on their identifiers, that's all.

pdyraga (commented):
Solidity looks good to me, I was not reviewing tests. Leaving it in @nkuba's hands.

solidity/random-beacon/contracts/libraries/Groups.sol
```diff
@@ -22,7 +22,8 @@ library Groups {
     struct Group {
         bytes groupPubKey;
         uint256 activationBlockNumber;
-        uint32[] members;
+        // Keccak256 hash of group members identifiers array (excluding IA/DQ members).
```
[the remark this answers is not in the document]
dimpar: Not sure if I understand the 2nd sentence: "Group members do not include operators selected by sortition pool *but* misbehaving during DKG." Is the following sentence has the same meaning to you? "Group members do not include operators selected by sortition pool that misbehaved during DKG."?

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -717,26 +717,35 @@ contract RandomBeacon is Ownable {
 
     /// @notice Creates a new relay entry.
     /// @param entry Group BLS signature over the previous entry.
-    function submitRelayEntry(bytes calldata entry) external {
+    /// @param groupMembers Group member ids that participated in dkg (excluding IA/DQ).
```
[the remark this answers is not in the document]
dimpar: Sure 3589a32

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -270,6 +270,8 @@ library DKG {
         self.submittedResultHash = keccak256(abi.encode(result));
         self.submittedResultBlock = block.number;
 
+        // To recreate the members that actively took part in dkg, the members array
+        // should be filtered out from misbehavedMembersIndices.
```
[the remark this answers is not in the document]
dimpar: Makes sense, but you can't put a comment inside the event. I put it above the event.
8ee733d

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
dimpar: e6ba9ef

solidity/random-beacon/contracts/libraries/Groups.sol
dimpar: bf32449

solidity/random-beacon/contracts/libraries/Heartbeat.sol
dimpar: 3589a32

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -717,26 +717,35 @@ contract RandomBeacon is Ownable {
...
+            "Invalid group members"
+        );
+
         uint256 slashingAmount = relay.submitEntry(entry, group.groupPubKey);
 
         if (slashingAmount > 0) {
-            address[] memory groupMembers = sortitionPool.getIDOperators(
-                group.members
+            address[] memory groupMembersIds = sortitionPool.getIDOperators(
+                groupMembers
```
nkuba, line 740: I just noticed that those are addresses that we are converting to, we need a rename.

```suggestion
            address[] memory groupMembersAddresses = sortitionPool.getIDOperators(
                groupMembers
```

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -760,25 +769,33 @@ contract RandomBeacon is Ownable {
...
+            group.membersHash == keccak256(abi.encode(groupMembers)),
+            "Invalid group members"
+        );
+
         uint256 slashingAmount = relay
             .relayEntrySubmissionFailureSlashingAmount;
-        address[] memory groupMembers = sortitionPool.getIDOperators(
-            groups.getGroup(groupId).members
+        address[] memory groupMembersIds = sortitionPool.getIDOperators(
+            groupMembers
```
nkuba, line 785:
```suggestion
        address[] memory groupMembersAddresses = sortitionPool.getIDOperators(
            groupMembers
```

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -829,21 +853,21 @@ contract RandomBeacon is Ownable {
 
         groups.terminateGroup(groupId);
 
-        address[] memory groupMembers = sortitionPool.getIDOperators(
-            groups.getGroup(groupId).members
+        address[] memory groupMembersIds = sortitionPool.getIDOperators(
+            groupMembers
```
nkuba, line 857:
```suggestion
        address[] memory groupMembersAddresses = sortitionPool.getIDOperators(
            groupMembers
```

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -122,7 +124,7 @@ library DKG {
         uint8[] misbehavedMembersIndices,
         bytes signatures,
         uint256[] signingMembersIndices,
-        uint32[] members
+        uint32[] selectedMembers
```
nkuba, line 127:
```suggestion
        uint32[] selectedMembersIds
```

solidity/random-beacon/contracts/libraries/Groups.sol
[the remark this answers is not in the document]
nkuba: > Not sure if I understand the 2nd sentence: "Group members do not include operators selected by sortition pool _but_ misbehaving during DKG." Does the following sentence have the same meaning to you? "Group members do not include operators selected by sortition pool that misbehaved during DKG."?

Sounds the same to me.

solidity/random-beacon/contracts/test/RandomBeaconStub.sol
```diff
@@ -24,15 +24,14 @@ contract RandomBeaconStub is RandomBeacon {
         return callback;
     }
 
-    function roughlyAddGroup(
-        bytes calldata groupPubKey,
-        uint32[] calldata members
-    ) external {
+    function roughlyAddGroup(bytes calldata groupPubKey, bytes32 membersHash)
```
nkuba, line 27:
```suggestion
    function roughlyAddGroup(bytes calldata groupPubKey, bytes32 groupMembersHash)
```

solidity/random-beacon/test/RandomBeacon.Callback.test.ts
```diff
@@ -42,6 +43,9 @@ const fixture = async () => {
 }
 
 describe("RandomBeacon - Callback", () => {
+  const dkgSeed: BigNumber = BigNumber.from(
```
nkuba, line 46: Is it used anywhere?

solidity/random-beacon/test/RandomBeacon.Callback.test.ts
```diff
@@ -63,6 +68,7 @@ describe("RandomBeacon - Callback", () => {
     testToken = contracts.testToken as TestToken
     callbackContract = contracts.callbackContractStub as CallbackContractStub
     callbackContract1 = contracts.callbackContractStub1 as CallbackContractStub
+    membersIDs = signers.map((member) => member.id)
```
nkuba, line 71: Nice!

nkuba (changes requested):
I'm sorry for noticing this that late but those are mostly just inconsistent naming comments.

solidity/random-beacon/contracts/libraries/Groups.sol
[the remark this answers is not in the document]
dimpar: 37fe0b3

solidity/random-beacon/contracts/test/RandomBeaconStub.sol
dimpar: 7e2de71

solidity/random-beacon/contracts/libraries/Heartbeat.sol
```diff
@@ -56,24 +56,26 @@ library Heartbeat {
...
     /// @param nonce Current nonce for group used in the claim.
+    /// @param members Identifiers of group members.
     /// @return failedMembers Identifiers of members who failed the heartbeat.
     function verifyFailureClaim(
         SortitionPool sortitionPool,
         FailureClaim calldata claim,
         Groups.Group storage group,
-        uint256 nonce
+        uint256 nonce,
+        uint32[] calldata members
```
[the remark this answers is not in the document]
dimpar: 7e2de71

solidity/random-beacon/contracts/libraries/DKG.sol
dimpar: 7e2de71

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: 7e2de71

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: 7e2de71

solidity/random-beacon/test/RandomBeacon.Callback.test.ts
dimpar: removed 2bb94a9

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: 7e2de71

dimpar:
> I'm sorry for noticing this that late but those are mostly just inconsistent naming comments.

Thx for checking! All comments are addressed

nkuba:
LGTM, I believe all @pdyraga comments were also addressed. I'll merge this PR tomorrow morning, giving Piotr a chance to veto.

PR #2771, opened 2021-12-22
Block MM connection when Tally is injected by default

michalsmiarowski:
There was a bug with Tally and MetaMask - if "use Tally as default wallet"
option in Tally was checked, then the app automatically connected to Tally 
wallet when we picked MetaMask, but displayed the MetaMask logo next to user 
address, which might be confused for some users.
We decided to block the possibility to connect to MM in such case and print
proper message, that tells user to uncheck the "use Tally as default wallet"
option if he wants to connect to MetaMask.

image


r-czajkowski (approved):
LGTM!

PR #2773, opened 2021-12-29
Initial ECDSA Wallet Contracts project configuration

nkuba:
This is an initial configuration for the ECDSA Wallet Contracts project.
The configuration is based on https://github.com/keep-network/keep-core/tree/main/solidity/random-beacon project.

Refs https://github.com/keep-network/keep-core/issues/2758

Issue #2758, opened 2021-12-06
ECDSA wallet DKG


pdyraga:
Depends on #2756 

Important: This issue is only about DKG, signature requests are out of scope.

For tBTC v2, we need tBTC bridge to create ECDSA wallets in the network. We should use the same, submit-challenge/approve approach for DKG as in random beacon v2 except that instead of a BLS key, the group should submit an ECDSA public key. We will start in `solidity/ecdsa` with `WalletFactory` and `Wallet` contracts. We should be able to import and reuse random beacon `DKG` and `DKGValidator` parametrizing group size, group threshold, and active threshold. I strongly suggest we do not try to make group size and threshold parameters of the create wallet function but hardcode them in the wallet as constants and assume a change of the group size is a hard-fork event of tBTC bridge component.

The total gas cost of DKG (submit+approve) is of paramount importance and after the changes from #2756 it should not exceed 669k. For a wallet created once a week, this is still a lot and should be optimized aggressively when working on this code. For example, we need to validate what is the difference between cloning `Wallet` contract and using a single `Wallet` contract dispatching signing requests by virtual wallet ID, in a similar way how signing requests are dispatched to groups in `RandomBeacon`.

Reusing `Groups` library from the random beacon is probably not a good idea at this point given that ECDSA wallets will have completely other lifecycle properties than beacon groups and failing heartbeat means something completely else. If we end up duplicating `Groups` library code now (with no expiration/termination/selection), this is fine.

Wallet API should be designed as a part of the work on this issue, two example options are presented below. The first one feels way more elegant but we need to take into account gas costs - to be decided when working on this issue.

```
contract WalletFactory {
  function createNewWallet(): Wallet
}

contract Wallet {
  function submitDkgResult(DKG.Result)
  function approveDkgResult(DKG.Result)
  function challengeDkgResult(DKG.Result)

  // later:
  function sign(bytes32 digest)
}
```

```
contract Wallet {
  function createNew(): uint // wallet ID
  function submitDKGResult(uint walletID, DKG.Result);
  function approveDkgResult(uint walletID, DKG.Result)
  function challengeDkgResult(uint walletID, DKG.Result)

  // later:
  function sign(uint walletID, bytes32 digest)
}
```

nkuba:
[...]

PR #2774, opened 2021-12-29
GH Workflow configuration for ECDSA contracts

nkuba:
This PR contains workflow configuration for ECDSA contracts located in `solidity/ecdsa` directory.

It also renames the existing random beacon workflow for consistency.

Depends on https://github.com/keep-network/keep-core/pull/2773

### DIFF
--- a/.github/workflows/contracts-ecdsa.yml
+++ b/.github/workflows/contracts-ecdsa.yml
@@ -1,0 +1,203 @@
+name: Solidity ECDSA
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  push:
+    branches:
+      - main
+    paths:
+      - "solidity/ecdsa/**"
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Environment for workflow execution"
+        required: false
+        default: "dev"
+      upstream_builds:
+        description: "Upstream builds"
+        required: false
+      upstream_ref:
+        description: "Git reference to checkout (e.g. branch name)"
+        required: false
+        default: "main"
+
+jobs:
+  contracts-detect-changes:
+    runs-on: ubuntu-latest
+    outputs:
+      path-filter: ${{ steps.filter.outputs.path-filter }}
+    steps:
+      - uses: actions/checkout@v2
+        if: github.event_name == 'pull_request'
+
+      - uses: dorny/paths-filter@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/contracts-ecdsa.yml
```diff
@@ -0,0 +1,203 @@
...
+          cache: "yarn"
+          cache-dependency-path: solidity/ecdsa/yarn.lock
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8.5
+
+      - name: Install Solidity
+        env:
+          SOLC_VERSION: 0.8.6 # according to solidity.version in hardhat.config.js
```
michalinacienciala, line 94:
```suggestion
          SOLC_VERSION: 0.8.9 # according to solidity.version in hardhat.config.ts
```
Same change can be done in the random beacon workflow.

.github/workflows/contracts-ecdsa.yml
```diff
@@ -0,0 +1,203 @@
...
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "14.x"
+          cache: "yarn"
+          cache-dependency-path: solidity/ecdsa/yarn.lock
+
+      - name: Install dependencies
+        run: yarn install --network-concurrency 1
```
michalinacienciala, line 62: Have you tried just `yarn install`?
`--network-concurrency 1` was used as a workaround for failing installation in `random-beacon`, but maybe we wan't get the failure here in `ecdsa`. `--network-concurrency 1` may slow down installation a bit, so it may be worth to remove it if it's not needed.
Same applies to other places in the file where we use --network-concurrency 1`.

michalinacienciala (commented):
Two small comments left. 
Also, it may be good idea to do a dry-run of the workflow before merging - what do you think? To do that we coud temporarily edit the workflow, add `--dry-run` flag in the publish step and later reverse the change / change git hiistory.

michalinacienciala:
Oh, and we should remember to set the `contracts-deployment-testnet` job as required for PR merge after we merge this PR.

.github/workflows/contracts-ecdsa.yml
nkuba: 🤦🏻 copy-paste 🤦🏻

michalinacienciala (approved):
LGTM!

PR #2775, opened 2021-12-29
ECDSA Wallet DKG initial implementation

nkuba:


solidity/ecdsa/contracts/WalletFactory.sol
```diff
@@ -0,0 +1,115 @@
...
+    event DkgSeedTimedOut();
+
+    event WalletCreated(address walletAddress);
+
+    constructor(SortitionPool _sortitionPool, DKGValidator _dkgValidator) {
+        dkg.init(_sortitionPool, _dkgValidator);
+    }
+
+    // TODO: Revisit to implement mechanism for a fresh wallet creation.
+    // Consider who the caller will be: anyone or onlyOwner?
```
dimpar, line 69: [Here](https://github.com/keep-network/keep-core/issues/2758) it says "For tBTC v2, we need tBTC bridge to create ECDSA wallets in the network." so I take it's tBTC bridge only.

dimpar:
Before we go too deep, do you think it makes sense to focus on contracts + tests in this PR and leave the deployment scripts for a separate PR?

solidity/ecdsa/contracts/DKGValidator.sol
```diff
@@ -0,0 +1,252 @@
...
+///         to ensure their result is valid and can not be challenged. All other
+///         network operators should perform validation of the submitted result
+///         using a free contract call and challenge the result if the
+///         validation fails.
+contract DKGValidator {
+    using BytesLib for bytes;
+    using ECDSA for bytes32;
+
+    /// @dev Size of a group in the threshold relay.
+    uint256 public constant groupSize = 64; // TODO: Confirm value
```
dimpar, line 42: Maybe we can use the initial values that are specified in [RFC here](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#group-size)? Group size: 100, groupThreshold: 51

nkuba: e951472af5c037e734fd8d507401abec29974088

dimpar:
@nkuba What do you think about changing the structure of contracts a bit to something like this:
```
import "./libraries/Wallets.sol";
import "@thesis/solidity-contracts/contracts/clone/CloneFactory.sol";

contract WalletManager is CloneFactory {
	registerOperator() external;
	updateOperatorStatus() external;

	submitDkgResult(DKG.Result calldata dkgResult) external;
	approveDkgResult(DKG.Result calldata dkgResult) external;
	challengeDkgResult(DKG.Result calldata dkgResult) external;

	createNewWallet() external {
		wallets.addWallet()
	};
}

---------------------------------

library Wallets {

	struct Data {
		wallets[]
	}

	struct Wallet {
		(...)
	}
	
	addWallet(Wallet wallet) internal {
		wallets.push(wallet)
	}

	(... other wallets functions)
}
```

The above structure is similar to what we have in random beacon. I think *WalletFactory* does too much for being a factory and its name suggests that it creates new wallets which is fine but also does a bunch of other things.

nkuba:
> @nkuba What do you think about changing the structure of contracts a bit to something like this:
> 
> ```
> import "./libraries/Wallets.sol";
> import "@thesis/solidity-contracts/contracts/clone/CloneFactory.sol";
> 
> contract WalletManager is CloneFactory {
> 	registerOperator() external;
> 	updateOperatorStatus() external;
> 
> 	submitDkgResult(DKG.Result calldata dkgResult) external;
> 	approveDkgResult(DKG.Result calldata dkgResult) external;
> 	challengeDkgResult(DKG.Result calldata dkgResult) external;
> 
> 	createNewWallet() external {
> 		wallets.addWallet()
> 	};
> }
> 
> ---------------------------------
> 
> library Wallets {
> 
> 	struct Data {
> 		wallets[]
> 	}
> 
> 	struct Wallet {
> 		(...)
> 	}
> 	
> 	addWallet(Wallet wallet) internal {
>                 Wallet wallet = Wallet(clonedWalletAddress)
> 		wallets.push(wallet)
> 	}
> 
> 	(... other wallets functions)
> }
> ```
> 
> The above structure is similar to what we have in random beacon. I think _WalletFactory_ does too much for being a factory and its name suggests that it creates new wallets which is fine but also does a bunch of other things.

`WalletFactory` handles the creation of a new Wallet instance all the way through the DKG up to the Wallet activation on DKG result approval. IMHO it makes sense to implement it in `WalletFactory` contract.
What I imagine we add in the future is a `WalletManager` contract that will handle the wallets' lifecycle, so requesting `WalletFactory` for a new wallet creation, wallet selection and other state changes for the wallet e.g. wallet termination.

nkuba:
The current gas costs:
```
·-------------------------------------------|---------------------------|-------------|-----------------------------·
|            Solc version: 0.8.9            ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 30000000 gas  │
············································|···························|·············|······························
|  Methods                                                                                                          │
··················|·························|·············|·············|·············|·············|················
|  Contract       ·  Method                 ·  Min        ·  Max        ·  Avg        ·  # calls    ·  eur (avg)    │
··················|·························|·············|·············|·············|·············|················
|  StakingStub    ·  increaseAuthorization  ·      44858  ·      44870  ·      44869  ·         67  ·            -  │
··················|·························|·············|·············|·············|·············|················
|  WalletFactory  ·  approveDkgResult       ·     197524  ·     205585  ·     199820  ·         15  ·            -  │
··················|·························|·············|·············|·············|·············|················
|  WalletFactory  ·  challengeDkgResult     ·     194180  ·     215640  ·     198472  ·          5  ·            -  │
··················|·························|·············|·············|·············|·············|················
|  WalletFactory  ·  registerOperator       ·     205948  ·     342687  ·     210639  ·         67  ·            -  │
··················|·························|·············|·············|·············|·············|················
|  WalletFactory  ·  requestNewWallet       ·          -  ·          -  ·      84630  ·          9  ·            -  │
··················|·························|·············|·············|·············|·············|················
|  WalletFactory  ·  submitDkgResult        ·     453308  ·     573356  ·     492522  ·         46  ·            -  │
·-----------------|-------------------------|-------------|-------------|-------------|-------------|---------------·
```

solidity/ecdsa/contracts/WalletFactory.sol
```diff
@@ -0,0 +1,270 @@
...
+        Wallet latestWallet = wallets[wallets.length - 1];
+        require(
+            latestWallet.activationBlockNumber() == 0,
+            "The latest registered wallet was already activated"
+        );
+
+        (bytes32 maliciousResultHash, uint32 maliciousSubmitter) = dkg
+            .challengeResult(dkgResult);
+
+        wallets.pop();
```
dimpar, line 204: Don't we need to destroy the wallet contract as well? `selfdestruct(address(wallet))`;

And another question if we delete wallets that were valid but its age / active period has passed?

nkuba:
> Before we go too deep, do you think it makes sense to focus on contracts + tests in this PR and leave the deployment scripts for a separate PR?

Deployment scripts are required for the tests, as we use fixtures based on the deployed contracts.

pdyraga:
```
|  WalletFactory  ·  submitDkgResult        ·     453308  ·     573356  ·     492522  ·         46  ·            -  │
```

I know it's still early but after the changes from #2768, shouldn't the cost be around `306979  ·     479157  ·     394890` ? What makes this difference?

solidity/ecdsa/contracts/WalletFactory.sol
nkuba: > Don't we need to destroy the wallet contract as well? selfdestruct(address(wallet));

I'm not sure we need to do this. The destruct would have to be called from the `Wallet` contract, and I'm not sure if it will work with a clone/proxy patter properly. `selfdestruct` transfers ETH from the contract so we would need to make sure we handle such a scenario.

> And another question if we delete wallets that were valid but its age / active period has passed?

I guess we won't delete them but change their state.

nkuba:
> ```
> |  WalletFactory  ·  submitDkgResult        ·     453308  ·     573356  ·     492522  ·         46  ·            -  │
> ```
> 
> I know it's still early but after the changes from #2768, shouldn't the cost be around `306979 · 479157 · 394890` ? What makes this difference?

Here is `submitDkgResult` gas costs breakdown: 
```
dkg.submitResult                = 272110
+ createClone                   = 313313
+ wallets.push                  = 357616
+ wallet.init+hashGroupMembers  = 440060
+ emit WalletCreated            = 455295
```

dimpar (commented):
Also another proposition, but I realize it might be too late taking into account the post audit for random beacon. Anyway, I see that some libs are almost identical copy ex. DKG.sol. Maybe we can extract them to a separate repo and import them to /random-beacon and /ecdsa? Not only we can save time on reviews, but also execution of tests should take less time.

solidity/ecdsa/test/Wallet.test.ts
```diff
@@ -1 +1,305 @@
...
+    masterWallet,
+    cloneFactory,
+    deployer,
+    owner,
+    thirdParty,
+  }
+}
+
+describe("Wallet", () => {
+  const membersIdsHash = keccak256(ethers.utils.toUtf8Bytes("MEMBERS IDS"))
```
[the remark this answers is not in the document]
nkuba: a462af361ecd94edaad706b607100f4f663c2668

solidity/ecdsa/test/fixtures/index.ts_backup
```diff
@@ -0,0 +1,190 @@
+import { Contract } from "ethers"
```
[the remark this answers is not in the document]
nkuba: a834603b84ecb080e23d83d874910551a3e9b853

solidity/ecdsa/contracts/WalletFactory.sol
dimpar: ok, I thought that we can save some gas on selfdestruct, but after some testing it doesn't look like we save anything. So let's leave it as is.

solidity/ecdsa/contracts/WalletFactory.sol
```diff
@@ -0,0 +1,306 @@
...
+        returns (uint256);
+}
+
+contract WalletFactory is CloneFactory, Ownable {
+    using DKG for DKG.Data;
+
+    // Libraries data storages
+    DKG.Data internal dkg;
+
+    Wallet[] public wallets;
```
dimpar, line 39: Will there be a case when we need to iterate this array? If no, how about we change the array to a map? I did some experiments map vs. array:

```
Wallet[] public wallets;
mapping(bytes32 => Wallet) public walletsByPubKeyHash; // alternative with mapping

(...)
wallets.push(wallet) // gas: 44297
walletsByPubKeyHash[wallet.publicKeyHash()] = wallet; // gas: 23243

wallets.pop(); // gas: 6147
delete walletsByPubKeyHash[keccak256(dkgResult.groupPubKey)]; // gas: 5466
```

It is not that significant when removing a wallet especially it won't be that frequent, but when adding we have ~21k of difference in favor of mapping. What do you think? btw. sorry for the late comment, I could've proposed this earlier.

dimpar:
Probably nitpicking, but in a lot of tests we don't start a context description with `context("when something happens")` but just `context("something")`. When running all the tests it reads:
```
something
 it should revert
```

but in BDD the proper way is:
```
when something happens
 it should revert
```

Here's the [source](https://mochajs.org/#bdd)

solidity/ecdsa/test/WalletFactory.DKG.test.ts
```diff
@@ -0,0 +1,2668 @@
...
+              // })
+            })
+          })
+
+          context("at the end of challenge period", async () => {
+            before(async () => {
+              await createSnapshot()
+
+              await mineBlocksTo(
+                resultSubmissionBlock +
```
dimpar, line 2287: I noticed that tests run very slowly even with the snapshots. The difference between the current block and the blocks to mine in this particular case is `11519` which is a very big number. In other places we have similar cases. Do you think we can stub dkg parameters and lower them? They won't reflect the real world, but at the end we can implement e2e test(s) that will have params similar to a real world.
Execution time on my box just for `Wallet.test.ts` + `WalletFactory.DKG.test.ts` is already >13min

solidity/ecdsa/contracts/DKGValidator.sol
```diff
@@ -0,0 +1,251 @@
...
+    {
+        // Compute the actual group members hash by selecting actual members IDs
+        // based on seed used for current DKG execution.
+        bytes32 actualGroupMembersHash = keccak256(
+            abi.encodePacked(
+                sortitionPool.selectGroup(groupSize, bytes32(seed))
+            )
+        );
+
+        // TODO: check what is more efficient - computing hash or iterating
```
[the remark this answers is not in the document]
nkuba: Can we remove it as part of #2779 from the random-beacon's library and then we will just copy-paste it here? I want the `DKGValidator.sol` file to be a copy-paste with just a few crucial changes.

solidity/ecdsa/test/WalletFactory.DKG.test.ts
```diff
@@ -0,0 +1,2667 @@
...
+
+        context("with dkg result not submitted", async () => {
+          it("should revert with 'Current state is not CHALLENGE' error", async () => {
+            await expect(
+              walletFactory.challengeDkgResult(stubDkgResult)
+            ).to.be.revertedWith("Current state is not CHALLENGE")
+          })
+        })
+
+        context("with malicious dkg result submitted", async () => {
```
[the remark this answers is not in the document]
nkuba: Nice catch! ba79f28ac12773e4fc69e82f478de02b0bad8608

solidity/ecdsa/contracts/DKGValidator.sol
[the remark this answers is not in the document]
dimpar: The thing is that if we leave 64 group members in the random-beacon then it is more efficient gas wise to use the iterative approach implemented in [this PR](https://github.com/keep-network/keep-core/pull/2753). Only if we have 100members then gas on average will be the same for hashing vs. iterative, but hashing is more concise and cleaner imo. So I'd rather merge #2753, next #2779 but at this point, it will be already changed to an iterative approach, so if we want to leave hashing here, the copy-pasting 1:1 won't work.

Unless at some point we'll extract common libs to a separate repo like I mentioned in [here](https://github.com/keep-network/keep-core/pull/2775#pullrequestreview-852945168), then I guess we can temporarily leave the iterative approach in /ecdsa as well.

nkuba:
Current gas usages:
```
·----------------------------------------------|---------------------------|-------------|-----------------------------·
|             Solc version: 0.8.9              ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 30000000 gas  │
···············································|···························|·············|······························
|  Methods                                                                                                             │
·····················|·························|·············|·············|·············|···············|··············
|  Contract          ·  Method                 ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
·····················|·························|·············|·············|·············|···············|··············
|  CloneFactoryStub  ·  createClone            ·          -  ·          -  ·      64548  ·            1  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  StakingStub       ·  increaseAuthorization  ·      44858  ·      44870  ·      44869  ·           67  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  Wallet            ·  activate               ·          -  ·          -  ·      48137  ·            2  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  Wallet            ·  init                   ·          -  ·          -  ·      93465  ·            7  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  Wallet            ·  sign                   ·          -  ·          -  ·      27742  ·            2  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  WalletFactory     ·  approveDkgResult       ·     194645  ·     202706  ·     196912  ·           16  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  WalletFactory     ·  challengeDkgResult     ·     392052  ·    1714665  ·    1475685  ·           17  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  WalletFactory     ·  notifyDkgTimeout       ·          -  ·          -  ·      46552  ·            1  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  WalletFactory     ·  registerOperator       ·     205948  ·     342687  ·     210639  ·           67  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  WalletFactory     ·  requestNewWallet       ·          -  ·          -  ·      84652  ·           13  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  WalletFactory     ·  submitDkgResult        ·     431276  ·     553290  ·     471859  ·           95  ·          -  │
·····················|·························|·············|·············|·············|···············|··············
|  Deployments                                 ·                                         ·  % of limit   ·             │
···············································|·············|·············|·············|···············|··············
|  CloneFactoryStub                            ·          -  ·          -  ·     135786  ·        0.5 %  ·          -  │
·----------------------------------------------|-------------|-------------|-------------|---------------|-------------·
```

solidity/ecdsa/contracts/WalletFactory.sol
```diff
@@ -0,0 +1,315 @@
...
+    ///         `resultSubmissionEligibilityDelay` blocks after the end of the
+    ///         challenge period can be called only by the DKG result submitter.
+    ///         After that time, can be called by anyone.
+    /// @dev It transfers Wallet's ownership to the Wallet Factory owner.
+    /// @param dkgResult Result to approve. Must match the submitted result
+    ///        stored during `submitDkgResult`.
+    function approveDkgResult(DKG.Result calldata dkgResult) external {
+        uint32[] memory misbehavedMembers = dkg.approveResult(dkgResult);
+
+        emit WalletActivated(address(wallet));
```
dimpar, line 226: I think we can remove this event and save gas. Inside of dkg.approveResult we already have `emit DkgResultApproved(submittedResultHash, sender)`. And during the `submitDkgResult()` we `emit WalletCreated(walletAddress, submittedResultHash)` and based on hash we can fetch the wallet address. 
I agree that it is more readable to emit the address of an approved wallet, but because we pay extra attention on the gas usage, I think we can get rid of WalletActivated event.

solidity/ecdsa/test/WalletFactory.DKG.test.ts
```diff
@@ -0,0 +1,2769 @@
...
+                await expect(tx)
+                  .to.emit(walletFactory, "WalletActivated")
+                  .withArgs(wallet.address)
+              })
+
+              it("should unlock the sortition pool", async () => {
+                await expect(await sortitionPool.isLocked()).to.be.false
+              })
+
+              it("should transfer wallet ownership to wallet registry")
```
dimpar, line 1792: Are there any plans to implement this one? If yes, can we add a todo? We have in total 3 "its" like this in this file.

dimpar (commented):
I think it's a good foundation for the next PRs. I don't have any more comments.

nkuba:
Closing in favor of https://github.com/keep-network/keep-core/pull/2792.

PR #2779, opened 2022-01-04
Passing members hash

dimpar:
Refs https://github.com/keep-network/keep-core/issues/2756

A hash of group members that actively took part in DKG can be calculated off-chain and passed to the contract's `submitDKGResult()`. This way we skip filtration of the misbehaved members. However, this change requires additional validation during the DKG challenge phase to verify that hash. The reasoning is that the dkg challenge will be done less frequently compared to dkg result submission which overall should save gas. 

### Branch `main`:
```
|  approveDkgResult                ·     185804  ·     247404  ·     205493  │
|  challengeDkgResult              ·     333403  ·    1175701  ·     978235  │
|  submitDkgResult                 ·     306979  ·     479157  ·     395381  │
|  submitRelayEntry (no slashing)  ·     192118  ·     241745  ·     208999  │
|  submitRelayEntry (slashing)     ·     214170  ·     495263  ·     354717  │
```

Bytecode:
```
 |  RandomBeacon            ·     22.507  │
 |  RandomBeaconStub        ·     23.855  │
```

### Current branch `passing-members-hash`:
```
|  approveDkgResult                ·     186422  ·     247907  ·     206113   │
|  challengeDkgResult              ·     336904  ·    1180374  ·     982299   │
|  submitDkgResult                 ·     297301  ·     421691  ·     383470   │
|  submitRelayEntry (no slashing)  ·     192140  ·     241767  ·     209021   |
|  submitRelayEntry (slashing)     ·     214192  ·     495285  ·     354739   │
```

Bytecode:
```
 |  RandomBeacon            ·     22.117  │
 |  RandomBeaconStub        ·     23.465  │
```

### Diff in gas (min-max-avg):
```
|  challengeDkgResult              ·            ·    +60273   ·             │
|  submitDkgResult                 ·     -9678  ·    -57466   ·     -11911  │
```

### Diff max in gas: current branch vs. main (min-max-avg):
```
|  challengeDkgResult: +84500   │ more costly for challengeDkgResult()
|  submitDkgResult:    -90442   | less costly for submitDkgResult()
```

### DIFF
--- a/solidity/random-beacon/contracts/libraries/DKG.sol
+++ b/solidity/random-beacon/contracts/libraries/DKG.sol
@@ -82,6 +82,9 @@ library DKG {
         // Identifiers of candidate group members as outputted by the group
         // selection protocol.
         uint32[] members;
+        // Keccak256 hash of group members identifiers that actively took part
+        // in DKG (excluding IA/DQ members).
+        bytes32 membersHash;
     }
 
     /// @notice States for phases of group creation. The states doesn't include
@@ -391,9 +394,14 @@ library DKG {
     /// @dev Can be called during a challenge period for the submitted result.
     /// @param result Result to challenge. Must match the submitted result
     ///        stored during `submitResult`.
+    /// @param groupMembersHash Challenged group members hash.
     /// @return maliciousResultHash Hash of the malicious result.
     /// @return maliciousSubmitter Identifier of the malicious submitter.
-    function challengeResult(Data storage self, Result calldata result)
+    function challengeResult(
+        Data storage self,
+        Result calldata result,
+        bytes32 groupMembersHash
+    )
         external
         returns (bytes32 maliciousResultHash, uint32 maliciousSubmitter)
     {
@@ -417,7 +425,12 @@ library DKG {
         // https://github.com/crytic/slither/issues/982
         // slither-disable-next-line unused-return
         try
-            self.dkgValidator.validate(result, self.seed, self.startBlock)
+            self.dkgValidator.validate(
+                result,
+                self.seed,
+                self.startBlock,
+                groupMembersHash
+            )
         returns (
             // slither-disable-next-line uninitialized-local,variable-scope
             bool isValid,
--- a/solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
+++ b/solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
@@ -741,6 +741,7 @@ describe("RandomBeacon - Group Creation", () => {
                 startBlock,
                 noMisbehaved,
                 firstEligibleSubmitterIndex,
+                undefined,
                 constants.groupThreshold
               ))
             })
@@ -2187,6 +2188,91 @@ describe("RandomBeacon - Group Creation", () => {
             await expect(
               randomBeacon.challengeDkgResult(stubDkgResult)
             ).to.be.revertedWith("Current state is not CHALLENGE")
+          })
+        })
+
+        context("with invalid members hash submitted", async () => {
+          let dkgResultHash: string
+          let dkgResult: DkgResult
+          let members: Operator[]
+          let membersIds: number[]
+
+          before(async () => {
+            await createSnapshot()
+
+            members = await selectGroup(sortitionPool, genesisSeed)
+            membersIds = members.map((m) => m.id)
+          })
+
+          after(async () => {
+            await restoreSnapshot()
+          })
+
+          context("when misbehaved members are present", async () => {
+            it("should emit DkgResultChallenged event", async () => {
+              const actualMisbehavedIndices = [1, 7, 14, 42]
+              const invalidMisbehavedIndices = [3, 8, 15, 41]
+              const invalidMembersHash = hashDKGMembers(
+                membersIds,
+                invalidMisbehavedIndices
+              )
+
+              ;({ dkgResult, dkgResultHash } =
+                await signAndSubmitArbitraryDkgResult(
+                  randomBeacon,
+                  groupPublicKey,
+                  members,
+                  startBlock,
+                  actualMisbehavedIndices,
+                  undefined,
+                  invalidMembersHash
+                ))
+
+              const tx = await randomBeacon
+                .connect(thirdParty)
+                .challengeDkgResult(dkgResult)
+
+              await expect(tx)
+                .to.emit(randomBeacon, "DkgResultChallenged")
+                .withArgs(
+                  dkgResultHash,
+                  await thirdParty.getAddress(),
+                  "Invalid members hash"
+                )
+            })
+          })
+
+          context("when misbehaved members are not present", async () => {
+            it("should emit DkgResultChallenged event", async () => {
+              const invalidMisbehavedIndices = [3]
+              const invalidMembersHash = hashDKGMembers(
+                membersIds,
+                invalidMisbehavedIndices
+              )
+
+              ;({ dkgResult, dkgResultHash } =
+                await signAndSubmitArbitraryDkgResult(
+                  randomBeacon,
+                  groupPublicKey,
+                  members,
+                  startBlock,
+                  noMisbehaved,
+                  undefined,
+                  invalidMembersHash
+                ))
+
+              const tx = await randomBeacon
+                .connect(thirdParty)
+                .challengeDkgResult(dkgResult)
+
+              await expect(tx)
+                .to.emit(randomBeacon, "DkgResultChallenged")
+                .withArgs(
+                  dkgResultHash,
+                  await thirdParty.getAddress(),
+                  "Invalid members hash"
+                )
+            })
           })
         })
 
--- a/solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
+++ b/solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
@@ -21,6 +21,7 @@ import {
   DkgResult,
   noMisbehaved,
   signAndSubmitUnrecoverableDkgResult,
+  hashDKGMembers,
 } from "./utils/dkg"
 import { registerOperators, Operator } from "./utils/operators"
 import { selectGroup, hashUint32Array } from "./utils/groups"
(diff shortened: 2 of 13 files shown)


nkuba:
I like the idea of filtering the members off-chain, but I'm afraid of introducing the changes now, after the audit. I'm curious about @pdyraga stand.

dimpar:
> > I like the idea of filtering the members off-chain, but I'm afraid of introducing the changes now, after the audit. I'm curious about @pdyraga stand.
> 
> Let's first merge #2772 and #2780. Then, let's compare changes in this PR against `main` from the perspective of gas costs and contract size for 64 and 100 group members. We will have a full picture then.

Results are ready in the description for 64 members. It is harder to compare 100 members, because we have bunch of tests that are written for 64 members and they fail. This is not a one number change case, but for sure the numbers for gas will be even higher than for 64 members. The tests that did pass produced these results for 100 members:

Branch: `main`

```
submitDkgResult    ·   345905  ·     542948  ·     428666
challengeDkgResult ·           ·    1500531  ·    
```

Branch: `passing-members-hash` current

```
submitDkgResult    ·  330874  ·     452494  ·     409014
challengeDkgResult ·          ·    1507884  ·    
```

pdyraga:
Can we implement just two tests:
- happy path
- challenge failing on the last validation step

execute them for 64 and 100 members and compare? We do not need to have all tests passing, we need the happy path which should be at least 99% of cases and the worst case for challenge.

dimpar:
> Can we implement just two tests:
> 
> * happy path
> * challenge failing on the last validation step
> 
> execute them for 64 and 100 members and compare? We do not need to have all tests passing, we need the happy path which should be at least 99% of cases and the worst case for challenge.

## branch `passing-members-hash`

Note that when you run *all* tests, 'min gas' is sort of misleading for `challengeDkgResult`, cause it will show you `333403` of gas only when it hit "validation reverted" [here](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/contracts/libraries/DKG.sol#L441), which is not a "happy path" of dkg challenge. All other executions consumed > 1.1M of gas. So when the function executes without any exceptions thrown, it will always consume >1.1M gas.

### 64 members

`challengeDkgResult()` when members hash is invalid (last step most expensive) `1224317`

`submitDkgResult()` happy path: `416714`
`submitDkgResult()` max: `421691`

### 100 members

`challengeDkgResult()` when members hash is invalid (last step most expensive) `1575953`

`submitDkgResult()` happy path: `445716`

## branch `main`

### 64 members

`challengeDkgResult()` when group members are invalid (last step most expensive) `1164044`

`submitDkgResult()` happy path: `426392`
`submitDkgResult()` max: `479157`


### 100 members

`challengeDkgResult()` when group members are invalid (last step most expensive) `1491453`

`submitDkgResult()` happy path: `460436`

solidity/random-beacon/test/utils/dkg.ts
```diff
@@ -112,19 +116,27 @@ export async function signAndSubmitArbitraryDkgResult(
     submitterIndex = firstEligibleIndex(ethers.utils.keccak256(groupPublicKey))
   }
 
+  let membersHash: string
+  if (!groupMembersHash) {
+    membersHash = hashDKGMembers(members, misbehavedIndices)
+  } else {
+    membersHash = groupMembersHash
+  }
```
nkuba, line 124: Can it be simplified?

```suggestion
  if (!groupMembersHash) {
    groupMembersHash = hashDKGMembers(members, misbehavedIndices)
  }
```

solidity/random-beacon/test/utils/dkg.ts
```diff
@@ -252,3 +267,24 @@ export async function signDkgResult(
 
   return { members, signingMembersIndices, signaturesBytes }
 }
+
+// Creates a members hash that actively participated in dkg
+export function hashDKGMembers(
+  members: number[],
+  misbehavedMembersIndices: number[]
+) {
```
nkuba, line 275: Let's fix `Missing return type on function` warning.
```suggestion
): string {
```

solidity/random-beacon/test/utils/dkg.ts
```diff
@@ -252,3 +267,24 @@ export async function signDkgResult(
 
   return { members, signingMembersIndices, signaturesBytes }
 }
+
+// Creates a members hash that actively participated in dkg
+export function hashDKGMembers(
```
nkuba, line 272: I know it's just a utils function, but maybe it would be worth adding a unit test for it? Especially to confirm misbehaved members filtering.

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -1318,35 +1320,6 @@ describe("RandomBeacon - Group Creation", () => {
                 })
               }
             )
```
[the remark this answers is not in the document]
dimpar: I removed it because of two reasons:
- after some thought, it feels that we checked the behavior of array implementation itself, but not the functionality of the function. The only validation aspect of it was an indication that we're aware of a such scenario
- when it was used to calucate hash during the dkg submission we didn't have any other checks. But now since the members hashing moved to the dkg challenge, we have a test that validates misbehaved members order [here](https://github.com/keep-network/keep-core/blob/passing-members-hash/solidity/random-beacon/test/DKGValidator.test.ts#L407). If the misbehaved members are unordered, it won't even hit the `validateMembersHash`

solidity/random-beacon/test/utils/dkg.ts
dimpar: it can't be simplified just like in the suggestion above, because eslint blocks reassignment of the param, but we can do a hybrid approach like here 0040b76

solidity/random-beacon/test/utils/dkg.ts
dimpar: 449a226

solidity/random-beacon/test/utils/dkg.ts
dimpar: Added 78be281

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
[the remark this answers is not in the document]
nkuba: That's cool we still got it covert indirectly. I just wanted to make sure that we have a test to verify that it will blow up in case of mixed order.

solidity/random-beacon/test/Utils.test.ts
```diff
@@ -0,0 +1,75 @@
+import { ethers } from "hardhat"
```
nkuba, line 1: Could we move the file to `solidity/random-beacon/test/utils/dkg.test.ts`.
It feels to me that placing it next to all the contract tests may cause a confusion.

solidity/random-beacon/test/Utils.test.ts
```diff
@@ -0,0 +1,75 @@
...
+import { expect } from "chai"
+import { hashDKGMembers } from "./utils/dkg"
+
+describe("utils/dkg", () => {
+  context("hashDKGMembers", () => {
```
nkuba, line 6: If we move the file as proposed in the previous comment we could reduce one level.

```suggestion
describe("hashDKGMembers", () => {
```

solidity/random-beacon/test/Utils.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    const members = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+    context("when there are no misbehaved members", () => {
```
nkuba, line 9: We're missing `it` blocks. We could rename the contexts to e.g. 
`it("calculates hash when there are no misbehaved members", () => {`

solidity/random-beacon/test/Utils.test.ts
```diff
@@ -0,0 +1,75 @@
+import { ethers } from "hardhat"
+import { expect } from "chai"
+import { hashDKGMembers } from "./utils/dkg"
+
+describe("utils/dkg", () => {
+  context("hashDKGMembers", () => {
+    const members = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```
nkuba, line 7: Could we make it more difficult by changing the values not to be `0 ... 9`? We expect indices in `misbehavedMembers` array, so I don't want to confuse them.

```suggestion
    const members = [100, 101, 102, 103, 104, 105, 106, 107, 108, 109]
```

solidity/random-beacon/test/Utils.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+      const actualHash = hashDKGMembers(members, misbehavedMembers)
+      const expectedHash = ethers.utils.keccak256(
+        ethers.utils.defaultAbiCoder.encode(["uint32[]"], [expectedMembers])
+      )
+
+      expect(expectedHash).to.be.equal(actualHash)
+    })
+
+    context("when a misbehaved member is not present in members array", () => {
```
nkuba, line 63: What will happen when `misbehavedMembers = [0]`?

dimpar: image

fixed in 2ca346e

solidity/random-beacon/test/Utils.test.ts
dimpar: 2ca346e

solidity/random-beacon/test/Utils.test.ts
dimpar: 2ca346e

solidity/random-beacon/test/Utils.test.ts
dimpar: 2ca346e

dimpar:
@nkuba I agree with all the recent comments around dkg.test.ts Thery were addressed in a single commit here: 2ca346e

nkuba (approved):

nkuba:
@pdyraga want to give it a final look?

solidity/random-beacon/contracts/libraries/DKG.sol
```diff
@@ -82,6 +82,9 @@ library DKG {
         // Identifiers of candidate group members as outputted by the group
         // selection protocol.
         uint32[] members;
```
nkuba, line 84: I'm sorry I missed it before, shouldn't this be renamed to `selectedOperators`?

nkuba: Actually, this may require a broader naming cleanup. So we assume that operators are selected from the sortition pool to participate in a DKG. Once operators complete the DKG successfully they become group members. 

I propose the following naming convention:

|old name|new name|
|:---:|:---:|
|`members`|`selectedOperators`|
|`submitterMemberIndex`|`submitterOperatorIndex`|
|`misbehavedMembersIndices`|`misbehavedOperatorsIndices`|
|`signingMembersIndices`|`signingOperatorsIndices`|
|`membersHash`|`groupMembersHash`|

Refs our previous conversation https://github.com/keep-network/keep-core/pull/2768#discussion_r773098164

nkuba: CC: @pdyraga @dimpar

dimpar: > Once operators complete the DKG successfully they become group members.

What do you consider a successful DKG completion? Is it upon dkg submission or when the challenge period is passed and dkg is approved?

nkuba: > > Once operators complete the DKG successfully they become group members.
> 
> What do you consider a successful DKG completion? Is it upon dkg submission or when the challenge period is passed and dkg is approved?

submitDkgResult + approveDkgResult

PR #2780, opened 2022-01-05
Gas-optimized version of submitRelayEntry when soft timeout has not passed

pdyraga:
Depends on https://github.com/keep-network/keep-core/pull/2768

# Background

In case the soft timeout has not passed yet, there is no point to calculate
the slashing factor in `submitRelayEntry`. Also, because there is no slashing,
passing group members is not necessary and computing group members hash
to compare it with the value stored in groups registry is not necessary.

# Details

This is a competing approach to the one in #2772 aiming at making minimal
changes to the codebase. It's up to @nkuba and @dimpar to decide with
which approach we want to continue.

Gas usage is pretty much the same but the contract has a little bit less bytecode
in this version:

In #2772:
```
 |  RandomBeacon            ·     (was: 22.630) now 22.549  │
 ···························|··············
 |  RandomBeaconStub        ·     (was: 24.180) now 23.897  │
```

Here:
```
 |  RandomBeacon            ·     22.406  │
 ···························|··············
 |  RandomBeaconStub        ·     23.956  │
```

This will be important for changes in https://github.com/keep-network/keep-core/issues/2715 where we will need to add more code
to `RandomBeacon` contract.

Worth noting, with two versions of `submitEntry` we can go even lower by making
the slasheable version `public` and linking the contract as external. Skipping it
until it is really needed to do not complicate the deployment:

```
 |  RandomBeacon            ·     22.331  │
 ···························|··············
 |  RandomBeaconStub        ·     23.881  │
```

# Outstanding work

This changeset is still a draft. The list of things that still need to be done include:
- [x] fixing tests for delay factor calculations,
- [x] adding more tests for the new `submitRelayEntry2` function,
- [x] renaming `submitRelayEntry2` to just `submitRelayEntry` and dealing with
  overloaded calls in unit tests.

### DIFF
--- a/solidity/random-beacon/test/RandomBeacon.Relay.test.ts
+++ b/solidity/random-beacon/test/RandomBeacon.Relay.test.ts
@@ -277,7 +277,7 @@ describe("RandomBeacon - Relay", () => {
     })
   })
 
-  describe("submitRelayEntry happy path", () => {
+  describe("submitRelayEntry before the soft timeout - happy path", () => {
     before(async () => {
       await createSnapshot()
 
@@ -393,7 +393,7 @@ describe("RandomBeacon - Relay", () => {
     })
   })
 
-  describe("submitRelayEntry after the soft timeout", () => {
+  describe("submitRelayEntry before and after the soft timeout", () => {
     before(async () => {
       await createSnapshot()
 
--- a/solidity/random-beacon/test/RandomBeacon.Relay.test.ts
+++ b/solidity/random-beacon/test/RandomBeacon.Relay.test.ts
@@ -277,7 +277,7 @@ describe("RandomBeacon - Relay", () => {
     })
   })
 
-  describe("submitRelayEntry before the soft timeout - happy path", () => {
+  describe("submitRelayEntry(bytes)", () => {
     before(async () => {
       await createSnapshot()
 
@@ -393,7 +393,7 @@ describe("RandomBeacon - Relay", () => {
     })
   })
 
-  describe("submitRelayEntry before and after the soft timeout", () => {
+  describe("submitRelayEntry(bytes,uint32[])", () => {
     before(async () => {
       await createSnapshot()
 
--- a/solidity/random-beacon/test/RandomBeacon.Relay.test.ts
+++ b/solidity/random-beacon/test/RandomBeacon.Relay.test.ts
@@ -277,7 +277,7 @@ describe("RandomBeacon - Relay", () => {
     })
   })
 
-  describe("submitRelayEntry", () => {
+  describe("submitRelayEntry happy path", () => {
     before(async () => {
       await createSnapshot()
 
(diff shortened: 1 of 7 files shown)


dimpar:
Just a note, there's also [this](https://github.com/keep-network/keep-core/pull/2779) follow-up PR to #2772 that can reduce the gas usage and byte code. Gas and byte code size is in the PR description.

The obvious downside is that it's a post-audit change.

dimpar:
## Gas usage and byte code size from #2772

Gas:
```
|  submitRelayEntry (no slashing)  ·     191316  ·     240943  ·     208197  │
|  submitRelayEntry (slashing)     ·     213726  ·     494962  ·     354344  │
```

Byte code:
```
 |  RandomBeacon            ·     22.549  │
 |  RandomBeaconStub        ·     23.897  │
```

vs. current PR

Gas:
```
|  submitRelayEntry (no slashing)  ·     192118  ·     241745  ·     208999  | 
|  submitRelayEntry (slashing)     ·     214170  ·     495263  ·     354717  | 
```

Byte code:
```
 |  RandomBeacon            ·     22.507  │
 |  RandomBeaconStub        ·     23.855  │
```

## Diff #2772 vs. current PR

Gas:
```
|  submitRelayEntry (no slashing)  ·     +802  ·     +802  ·     +802  | 
|  submitRelayEntry (slashing)     ·     +444  ·     +301  ·     +373  | 
```

Byte code:
```
 |  RandomBeacon            ·     -0.042  │
 |  RandomBeaconStub        ·     -0.042  │
```

## To sum up: 
In 2772 we spend less gas during the relay entry submission, but the way it's implemented uses more bytes in cotract size. The other difference is that in 2772 we get rid of slashing factor and calculating just the slashing amount (hence the gas difference). Although in this PR we have less changes in the contract, which we need to take into account since this is a post audit change. 

Regarding the byte code size, we can go even lower per [this comment](https://github.com/keep-network/keep-core/pull/2780#issuecomment-1005897656)

dimpar:
After discussing with @nkuba both approaches that were implemented in #2772 and in this PR we decided to go with this PR and close 2772. Even though 2772 uses less gas, but it introduces more changes to the codebase which we would like to reduce taking into account the audit was already done.

solidity/random-beacon/test/RandomBeacon.Relay.test.ts
```diff
@@ -277,7 +277,7 @@ describe("RandomBeacon - Relay", () => {
     })
   })
 
-  describe("submitRelayEntry", () => {
+  describe("submitRelayEntry before the soft timeout - happy path", () => {
```
nkuba, line 280: I was confused with the test naming. I'm wondering whether it wouldn't be better to name the tests for overlapping function names like:
```
describe("submitRelayEntry(bytes)", () => {
```
and
```
describe("submitRelayEntry(bytes,uint32[])", () => {
```

It would be consistent with other tests as we always use the function name in `describe` message.

pdyraga: I would definitely rename to include the signature just like @nkuba suggests.

dimpar: dfe4320

nkuba (approved):

PR #2781, opened 2022-01-06
💰 Add new TBTCV2 + Saddle Meta V2 pool

michalsmiarowski:
Ref: https://github.com/keep-network/keep-ecdsa/pull/948

This PR adds new liquidity card with TBTCV2 + Saddle Meta V2 pool

Main changes:
- adds new liquidity card - TBTCV2 + Saddle Meta V2 pool
- fix `More Info` link in Inactive cards
- widen the banner on top of Liquidity page to make it more in line with Liquidity Cards
- removes unnecessary `incentivesRemoved` prop from Liquidity Cards


r-czajkowski (approved):
LGTM! Tested locally and I was able to deposit LP tokens and exit from the pool. I ran the dapp locally with mainnet contracts and seems to work but I did not test deposit/exit.

r-czajkowski (approved):

PR #2791, opened 2022-01-20
Ecdsa wallet slashing

dimpar:

Depends on https://github.com/keep-network/keep-core/pull/2775

Slashing submitter of a malicious dkg result during the dkg challenge call.


dimpar:
Closing in favor of https://github.com/keep-network/keep-core/pull/2797 The reason is that the base concept has changed around how we create ecdsa wallets. See https://github.com/keep-network/keep-core/pull/2775#issuecomment-1020271811

PR #2795, opened 2022-01-25
Migration and staking bugfixes 🐛 ❌ 

michalsmiarowski:
Some overall bug fixes 🐛 ❌  for the T Migration and keep staking

Changes:
- fix percentage calculations for the progress bar in `staked` tab when there is pending undelegation
- creates `useCurrentDate` hook, that returns current date in unix timestamp updated every seconds
- semi-fix of the bug that caued the need to refresh the page after executing a transaction that updates the status of Delegation or Undelegation. Now, you don't have to refresh the page, but you still have to go to another page and then comeback to the previous one to fetch the data again
- fix updating staked value after recovering tokens, which was not updated at all because of a typo in action name
- display ">99%" and "<1%" in AllocationProgressBar when there is really small amount of (for example) unstaked grant tokens compared to unstaked wallet tokens
- rename `recover tokens` occurences in dapp to `claim tokens`
- add `upgrade to t` button to threshold modals
- improve text of the claiming tokens modal
- fix `readyToRelease` calculation after doing a stake from grant tokens (it causes to reduce the value of stake to 0 until user refresh the page)
- restyle `Token Grants` page and fix displaying staked tokens in circular progress bar there
- remove `ECDSA` and `Random Beacon` flags from `Staked` tile (for now)
- restyle Delegations and Undelegations table (add the possibility to center the whole tab, add info if the stake is from Wallet or from Granted Tokens)
- restyle resource tooltip and use more info icon everywhere
- restyle Total Value Locked tooltip
- install `react-tooltip` library
- add tooltips to allocation progress bars
- add margin to AddToCalnedar button so it looks better in success modal when undelegating the stake

### DIFF
--- a/solidity-v1/dashboard/package-lock.json
+++ b/solidity-v1/dashboard/package-lock.json
@@ -32032,6 +32032,22 @@
         }
       }
     },
+    "react-tooltip": {
+      "version": "4.2.21",
+      "resolved": "https://registry.npmjs.org/react-tooltip/-/react-tooltip-4.2.21.tgz",
+      "integrity": "sha512-zSLprMymBDowknr0KVDiJ05IjZn9mQhhg4PRsqln0OZtURAJ1snt1xi5daZfagsh6vfsziZrc9pErPTDY1ACig==",
+      "requires": {
+        "prop-types": "^15.7.2",
+        "uuid": "^7.0.3"
+      },
+      "dependencies": {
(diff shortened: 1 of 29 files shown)


r-czajkowski (dismissed):

r-czajkowski (approved):
LGTM! 👍

PR #2796, opened 2022-01-25
Threshold integration

r-czajkowski:
Ref: #2798
Ref: #2799 
Ref: #2803
Ref: #2807

This is an integration PR. It adds support for moving the legacy KEEP stake to T network.

PR #2799, opened 2022-01-27
Update dashboard testnet GH workflow

r-czajkowski:
Ref: #2796 

Resolve `@threshold-network/solidity-contracts` dependency version on
the CI side.

### DIFF
--- a/.github/workflows/dashboard-v1-testnet.yml
+++ b/.github/workflows/dashboard-v1-testnet.yml
@@ -74,7 +74,8 @@ jobs:
               @keep-network/keep-ecdsa \
               @keep-network/tbtc \
               @keep-network/tbtc-v2 \
-              @keep-network/coverage-pools
+              @keep-network/coverage-pools \
+              @threshold-network/solidity-contracts
 
       - name: Get upstream packages' versions
         if: github.event_name == 'workflow_dispatch'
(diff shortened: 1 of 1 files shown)


.github/workflows/dashboard-v1-testnet.yml
```diff
@@ -88,6 +89,7 @@ jobs:
             tbtc-contracts-version = github.com/keep-network/tbtc/solidity#version
             tbtc-v2-contracts-version = github.com/keep-network/tbtc-v2/solidity#version
             coverage-pools-version = github.com/keep-network/coverage-pools#version
+            threshold-contracts-version = github.com/threshold-network/solidity-contracts#version
```
michalinacienciala, line 92: I'm 90% sure that we need to do some modifications in https://github.com/keep-network/ci/blob/main/config/config.json and https://github.com/keep-network/ci/tree/main/actions/upstream-builds-query to make this work. I will verify that and create the needed PRs in the `ci` repo.

michalinacienciala: Also, we should modify the actions in the `ci` repo to make the `@threshold-network/solidity-contracts` automatically deploy somewhere after `keep-core` and before `coverage-pools` (see https://coda.io/d/Building-Keep_d-fmEgBNFVH/Current-CI-process_su1ww#_luqjk). I will work on that as well.

michalinacienciala (approved):
Approving, `threshold-network` module is now incorporated in the CI flow, so we can use it when building dashboard.

PR #2801, opened 2022-01-27
ECDSA Wallet: Integrate with Random Beacon

nkuba:
This PR integrates DKG with the Random Beacon.

When a new wallet is requested we lock DKG state and issue a request to the Random Beacon for the new relay entry. DKG is not started yet.

Once a relay entry is produced we expect the Random Beacon to execute a callback function, which will forward the relay entry to the DKG as a seed. This is a moment when DKG really starts.

We introduced a timeout notification mechanism for the seed delivery to unlock the DKG if the Random Beacon takes too long to produce the entry.

It is expected that the request to the Random Beacon for a new relay entry is not reverted. It's very important for the DKG to have a fresh seed.


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -50,14 +50,23 @@ interface IRandomBeaconStaking {
         returns (uint256);
 }
 
+/// @title Random Beacon interface
+interface IRandomBeacon {
```
dimpar, line 54: 👍

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -143,6 +157,14 @@ contract WalletRegistry is Ownable {
         );
     }
 
+    function __beaconCallback(uint256 _randomRelayEntry, uint256 blockNumber)
```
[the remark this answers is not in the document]
nkuba: I'm not sure if emitting an event here would be beneficial. Do you have any use case for it in mind?

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -151,8 +173,15 @@ contract WalletRegistry is Ownable {
 
         dkg.lockState();
         dkg.start(
-            uint256(keccak256(abi.encodePacked(relayEntry, block.number)))
+            uint256(keccak256(abi.encodePacked(randomRelayEntry, block.number)))
```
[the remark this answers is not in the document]
nkuba: It may be possible that subsequent DKG executions are started with the same `randomRelayEntry`, (e.g. when the beacon doesn't produce a new relay entry). We should start DKG with a unique seed anyways, so we're adding the `block.number`.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -151,8 +173,15 @@ contract WalletRegistry is Ownable {
...
         );
+
+        // FIXME: What if relay entry is currently under creation and this request
+        // will fail? What if it will fail constantly due to a bad timing, do we
+        // want to pull a relay entry requested by someone else?
+        try randomBeacon.requestRelayEntry(this) {} catch {
```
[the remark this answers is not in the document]
nkuba: I guess we don't want to block a wallet creation because of an error returned from the Random Beacon.

solidity/random-beacon/contracts/RandomBeacon.sol
nkuba: I wasn't sure if this change to `random-beacon` can be included in this PR or it should go into a separate PR. These are separate projects in a monorepo.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -96,17 +101,26 @@ contract WalletRegistry is Ownable {
         IERC20 _tToken,
         IWalletStaking _staking,
         DKGValidator _dkgValidator,
+        IRandomBeacon _randomBeacon,
         address _walletOwner
     ) {
         sortitionPool = _sortitionPool;
         tToken = _tToken;
         staking = _staking;
+        randomBeacon = _randomBeacon;
```
nkuba, line 110: @dimpar @pdyraga Thoughts on having `randomBeacon` governable?

solidity/ecdsa/contracts/WalletRegistry.sol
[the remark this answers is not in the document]
dimpar: Only informational use case about updating the relay entry and at what block. I don't think it's a "must have" event, so yea, we can leave as is or consider adding it if doesn't cost too much gas.

solidity/ecdsa/contracts/WalletRegistry.sol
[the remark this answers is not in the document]
dimpar: Ok, this comment also depends on [the question](https://github.com/keep-network/keep-core/pull/2801#discussion_r795657136) of whether we can create a new wallet when the random beacon fails to provide a new entry. Unless you're sure that we can do that?

solidity/ecdsa/contracts/WalletRegistry.sol
dimpar: I take that it might be useful in case of an update of `randomBeacon`?

solidity/ecdsa/test/WalletRegistry.RandomBeacon.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      await waffle.loadFixture(walletRegistryFixture))
+
+    const randomBeaconMock = await deployMockContract(
+      deployer,
+      (
+        await deployments.getArtifact("RandomBeacon")
+      ).abi
+    )
+
+    await walletRegistry.updateRandomBeacon(randomBeaconMock.address)
```
dimpar, line 42: At the end we'll need to update a random beacon contract through governance, cause 'updateRandomBeacon()` will be ownable ie. `walletRegistryGovernance.connect(governance).updateRandomBeacon(newRandomBeacon.address)`. Can we add a TODO here?

solidity/ecdsa/contracts/WalletRegistry.sol
pdyraga: Yes, we need to be able to upgrade it.

solidity/ecdsa/contracts/WalletRegistry.sol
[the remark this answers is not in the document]
pdyraga: `requestNewWallet` should revert if the beacon is busy. We do not want to create a new wallet with the previous known entry and maintainers should request for a new wallet only when the beacon is ready to provide a new entry.

solidity/ecdsa/contracts/WalletRegistry.sol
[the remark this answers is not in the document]
pdyraga: We can't use the previous entry/block number to select operators to a new wallet, it increases the attack surface. Relay entry must not be known at the time a new wallet is requested.

solidity/ecdsa/contracts/WalletRegistry.sol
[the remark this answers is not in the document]
pdyraga: We will have the event emitted in the context of TX: https://github.com/keep-network/keep-core/blob/c9e57b21666ae090cf422b920944557759c15ab0/solidity/random-beacon/contracts/libraries/Relay.sol#L174

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -135,17 +137,25 @@ contract WalletRegistry is Ownable {
...
+    }
+
     // TODO: Update to governable params
     function updateDkgParams(
+        uint256 newSeedTimeout,
         uint256 newResultChallengePeriodLength,
         uint256 newResultSubmissionTimeout,
         uint256 newSubmitterPrecedencePeriodLength
     ) external {
+        dkg.setSeedTimeout(newSeedTimeout);
```
dimpar, line 158: Can we merge first 2800 with the gov params? This entire function will be gone and replaced by the ownable `updateDkgParams`. Everytime a new gov params is introduced it produces conflicts with 2800 PR which creates double work, here and there.

solidity/ecdsa/test/WalletRegistry.RandomBeacon.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      })
+
+      after(async () => {
+        await restoreSnapshot()
+
+        resetMock(randomBeacon)
+      })
+
+      it("should revert", async () => {
+        // FIXME: For some reason this check doesn't work with the expected error message
```
dimpar, line 60: Not very familiar with a smock lib (but I like the mocking lib idea) but can it be related to a wrong param, ie shouldn't we use `requestRelayEntry.whenCalledWith(giveWalletRegistryAddress)` and the reason for not failing is becuase of the addresses in the params mismatch?

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -135,17 +137,25 @@ contract WalletRegistry is Ownable {
         maliciousDkgResultNotificationRewardMultiplier = 100;
 
         dkg.init(_sortitionPool, _dkgValidator);
+        dkg.setSeedTimeout(1440); // ~6h assuming 15s block time // TODO: Verify value
         dkg.setResultChallengePeriodLength(11520); // ~48h assuming 15s block time
         dkg.setResultSubmissionTimeout(100 * 20); // TODO: Verify value
         dkg.setSubmitterPrecedencePeriodLength(20); // TODO: Verify value
     }
 
+    // TODO: Update to governable params
```
pdyraga, line 146: When implementing governance mechanism for it, please do not apply a delay.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -181,17 +191,28 @@ contract WalletRegistry is Ownable {
 
     /// @notice Requests a new wallet creation.
     /// @dev Can be called only by the owner of wallets.
-    ///      It starts a new DKG process.
+    ///      It locks the DKG and request a new random relay entry. It expects
+    ///      that the DKG process will be started once a new random relay entry
+    ///      gets generated.
     function requestNewWallet() external {
         require(msg.sender == walletOwner, "Caller is not the Wallet Owner");
```
pdyraga, line 198: Non-blocking this PR but why not making it a modifier?

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -181,17 +191,28 @@ contract WalletRegistry is Ownable {
...
-        dkg.start(
-            uint256(keccak256(abi.encodePacked(relayEntry, block.number)))
+
+        randomBeacon.requestRelayEntry(this);
+    }
+
+    /// @notice A callback that is executed once a new random relay entry gets
+    ///         generated. It starts the DKG process.
+    /// @dev Can be called only by the random beacon contract.
+    /// @param randomRelayEntry Random relay entry.
```
pdyraga, line 208: Relay entry is always random, why not just `relayEntry`?

solidity/ecdsa/contracts/libraries/DKG.sol
```diff
@@ -33,6 +34,8 @@ library DKG {
     using ECDSA for bytes32;
 
     struct Parameters {
+        // Time in blocks during which a seed is expected to be delivered.
+        uint256 seedTimeout;
```
pdyraga, line 38: We need to make it crystal clear in documentation that the time when we await for a seed is not included as a time of DKG because DKG has not yet started - it needs a seed to start - and the DKG timeout does _not_ include the time the contract waits for the seed.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -50,14 +50,23 @@ interface IRandomBeaconStaking {
...
+    /// @notice Creates a request to generate a new relay entry, which will
+    ///         include a random number (by signing the previous entry's
```
pdyraga, line 56: Relay entry is the random number.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -19,6 +19,8 @@ import "./libraries/Wallets.sol";
 import "./DKGValidator.sol";
 import "@keep-network/sortition-pools/contracts/SortitionPool.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
+import {IRandomBeacon} from "@keep-network/random-beacon/contracts/RandomBeacon.sol";
+import {IRandomBeaconConsumer} from "@keep-network/random-beacon/contracts/libraries/Callback.sol";
```
pdyraga, line 23: Non-blocking this PR but we could extract `IRandomBeaconConsumer` interface to its own `*.sol` file. Importing an interface from a library that is supposed to be internal to the random beacon code is a code smell.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -181,17 +191,28 @@ contract WalletRegistry is Ownable {
...
-            uint256(keccak256(abi.encodePacked(relayEntry, block.number)))
+
+        randomBeacon.requestRelayEntry(this);
+    }
+
+    /// @notice A callback that is executed once a new random relay entry gets
+    ///         generated. It starts the DKG process.
+    /// @dev Can be called only by the random beacon contract.
+    /// @param randomRelayEntry Random relay entry.
+    function __beaconCallback(uint256 randomRelayEntry, uint256) external {
```
pdyraga, line 209: What is the gas cost of this function? Are we sure it's going to fit under the random beacon's callback gas limit? Could be worth to mention it in `@dev` doc for this function.

solidity/ecdsa/contracts/libraries/DKG.sol
```diff
@@ -264,32 +278,23 @@ library DKG {
...
-    // slither-disable-next-line dead-code
-    function notifySeedTimedOut(Data storage self) internal {
-        require(
-            currentState(self) == State.AWAITING_SEED,
-            "Current state is not AWAITING_SEED"
-        );
-
-        // TODO: Verify if enough time passed.
+    function notifySeedTimeout(Data storage self) internal {
+        require(hasSeedTimedOut(self), "seed awaiting has not timed out");
```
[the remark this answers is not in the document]
pdyraga: Also, let's be consistent and start with a capital letter.

pdyraga (commented):
I was looking only at `*.sol`. Looks elegant 👌🏻

solidity/ecdsa/contracts/WalletRegistry.sol
nkuba: In tests, it's `77307` on avarage.

nkuba: Currently, RandomBeacon's `callbackGasLimit` is set to `50000`.

nkuba: :untada:

solidity/ecdsa/test/WalletRegistry.WalletCreation.test.ts
```diff
@@ -2818,103 +3121,244 @@ describe("WalletRegistry - Wallet Creation", async () => {
...
+    })
+
+    context("with new wallet requested", async () => {
+      let requestNewWalletStartBlock: number
+
+      before(async () => {
+        await createSnapshot()
+        const tx = await walletRegistry.connect(walletOwner).requestNewWallet()
+
+        requestNewWalletStartBlock = tx.blockNumber
```
dimpar: We can use `tx.blockNumber` directly instead of `requestNewWalletStartBlock` in the test below. I don't think we need to create a new var for this.

solidity/ecdsa/contracts/libraries/DKG.sol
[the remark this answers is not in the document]
nkuba: 3e6f203cff426537ce854aaa2f754419ecfd1170

solidity/ecdsa/contracts/WalletRegistry.sol
nkuba: There's just one place where we use it. If we use it in more places later we should add a modifier.

solidity/ecdsa/contracts/WalletRegistry.sol
nkuba: e01090f05ccca3ddf50dfda58ca92a17f817a688

solidity/ecdsa/contracts/WalletRegistry.sol
nkuba: I don't like that the interface sits in the library's file. But extracting it seems inconsistent with other interfaces that are defined in `RandomBeacon.sol`. How about we extract all the interfaces to a `interfaces/` directory, separate `*.sol` file each?

solidity/ecdsa/contracts/WalletRegistry.sol
nkuba: 9a5c5f1fc2d7886f35bf31629f89cc459c23e456

solidity/ecdsa/contracts/WalletRegistry.sol
nkuba: bd5485dde517d972244713b8eef7d8a6f78ff268

solidity/ecdsa/test/WalletRegistry.WalletCreation.test.ts
dimpar: `requestNewWalletStartBlock` is used only in one place. Instead of `await mineBlocksTo(requestNewWalletStartBlock + dkgTimeout + 1)` we can do `await mineBlocksTo(tx.blockNumber + dkgTimeout + 1)`. In both cases we extract from tx, the only difference is that we introduce a new var as it is coded now. However, it's a test which is no biggy in terms of the gas costs, so the bottom line it's a personal preference. We certainly can leave as is.

nkuba: Yeah, now I see, I was looking at another test case, where this variable is used twice. But for me, it makes sense for consistency to leave it as is, as we have a similar approach in other tests where we define `requestNewWalletStartBlock`.

solidity/ecdsa/contracts/WalletRegistry.sol
pdyraga: There is a difference between what's in the internal library and what's in the contract `WalletRegistry` is supposed to have an address to. Two options are equally good for me from the external perspective:
- Add `api/` directory, have there `IRandomBeacon.sol` and `IRandomBeaconConsumer.sol`
- Inside `RandomBeacon.sol` have `IRandomBeacon` and `IRandomBeaconConsumer` defined.

However, the second option sucks from `Callback.sol`'s perspective because it imports something from a file it was supposed to be imported from.

So I'd go with `api` (not `interfaces`, let's make it concise) and separate files as you suggested.

solidity/ecdsa/contracts/WalletRegistry.sol
pdyraga: Just double-checking, is 56k enough when the callback is called for the very first time? There is a gas cost difference when we modify a value that is already a non-default one and when we change a value from a default one to something else.

solidity/ecdsa/contracts/WalletRegistry.sol
nkuba: 15022421ca49ac41811cd6b9d26cb9b160aa0d80

solidity/ecdsa/contracts/WalletRegistry.sol
nkuba: We'll be always setting a value over a default (`0`). For the first time, the struct is initialized with a default, and before the next run we clean up the seed with `delete` https://github.com/keep-network/keep-core/blob/15022421ca49ac41811cd6b9d26cb9b160aa0d80/solidity/ecdsa/contracts/libraries/EcdsaDkg.sol#L512.

solidity/ecdsa/contracts/EcdsaDkgValidator.sol
```diff
@@ -24,10 +24,11 @@ pragma solidity ^0.8.9;
 
 import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import "@keep-network/random-beacon/contracts/libraries/BytesLib.sol";
-import "./libraries/DKG.sol";
+import "@keep-network/sortition-pools/contracts/SortitionPool.sol";
```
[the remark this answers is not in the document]
nkuba: Yes, we have a reference to `SortitionPool`. I'm not sure why the build hadn't failed without this import.

solidity/ecdsa/test/fixtures/index.ts
```diff
@@ -21,6 +21,8 @@ export const constants = {
   groupThreshold: 51,
   minimumStake: to1e18(100000),
   poolWeightDivisor: to1e18(1),
+  governanceDelayStandard: 43200, // 12 hours
+  governanceDelayCritical: 1210000, // 2 weeks
```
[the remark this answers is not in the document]
nkuba: 890e9b40b4942815b6a5f1eee1961ef1107c5099

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -125,21 +125,26 @@ contract WalletRegistry is Ownable {
...
 
+    event RandomBeaconUpgraded(address randomBeacon);
+
     event WalletOwnerUpdated(address walletOwner);
 
     constructor(
         SortitionPool _sortitionPool,
         IWalletStaking _staking,
-        DKGValidator _dkgValidator,
+        EcdsaDkgValidator _EcdsaDkgValidator,
```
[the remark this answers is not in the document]
nkuba: 8084171d2fe846484a09a8f82bacca96d28680e6

solidity/ecdsa/test/WalletRegistry.RandomBeacon.test.ts
```diff
@@ -0,0 +1,239 @@
...
+          await walletRegistry.connect(walletOwner).requestNewWallet()
+        })
+
+        after(async () => {
+          await restoreSnapshot()
+        })
+
+        // The exact value was noted from a test execution and is used as a reference
+        // for all future executions.
+        it("should not exceed 77308", async () => {
```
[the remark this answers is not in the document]
nkuba: We got two types of tests in this file. Here we call the `__calback` function directly in a transaction, so it costs more than 56k as it adds ~20k as the transaction execution cost. Another way is like it will be used in real-life so executed inside another transaction so calling just the `__calback` function inside another function call will cost ~56k.

solidity/ecdsa/contracts/EcdsaDkgValidator.sol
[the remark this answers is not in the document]
dimpar: I've cleaned cache and removed the build folder without SortitionPool.sol and it compiles. `EcdsaDkg.sol` imports `SortitionPool` and here we import `EcdsaDkg`, maybe that's why.

dimpar:
@nkuba New changes look good to me, I have no more comments.

solidity/random-beacon/contracts/api/IRandomBeacon.sol
```diff
@@ -0,0 +1,26 @@
...
+//                           Trust math, not hardware.
+
+pragma solidity ^0.8.9;
+
+import "./IRandomBeaconConsumer.sol";
+
+/// @title Random Beacon interface
+interface IRandomBeacon {
+    /// @notice Creates a request to generate a new relay entry, which will
+    ///         include a random number (by signing the previous entry's
```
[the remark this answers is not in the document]
nkuba: 43e6b44194ccff1417b01a216b0fbacea7d13b4c

PR #2802, opened 2022-01-31
Ignore `external` directory in git and prettier

michalinacienciala:
An `external/npm/@keep-network/keep-core/artifacts` directory is created
every time the `threshold-network/solidity-contracts` dependency is
installed. This is due to a workaround in
`threshold-network/solidity-contracts` which allows us to use contracts
which come from different repos and share the same name. Creation of
the `external` folder is a by-product of that. We don't need to track
changes in that folder or run prettier on its files.

### DIFF
--- a/solidity-v1/dashboard/.gitignore
+++ b/solidity-v1/dashboard/.gitignore
@@ -7,6 +7,9 @@
 # production
 /build
 
+# generated
+/external
+
 # misc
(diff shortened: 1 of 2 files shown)


michalinacienciala:
Closing. Opened https://github.com/keep-network/keep-core/pull/2803 instead.

PR #2804, opened 2022-01-31
tBTC reward allocation 2021-12-21 -> 2022-01-22

r-czajkowski:
Adding tBTC rewards merkle tree computed in keep-network/keep-ecdsa#950 to KEEP Token Dashboard.

This PR also adjusts the `tBTC Rewards` page to [the proposal](https://forum.keep.network/t/proposal-to-extend-stakedrop-rewards-for-an-additional-4-months/351) that extends stakedrop rewards for an additional 4 months. We will have two more allocations so to simplify calculating the period of the reward we hardcoded the `merkleRoot` in the `ExtendedECDSARewards.intervalsDates` array. After generating the new merkle tree we must add the `merkleRoot` hash to a given interval.


michalsmiarowski (approved):
LGTM 🚀

PR #2807, opened 2022-02-01
Display `transaction is pending` modal

r-czajkowski:
While a transaction is pending we should display a modal indicating that
fact with a link to the transaction details.


michalsmiarowski (approved):
Wonderful code ❗

Issue #2824, opened 2022-02-14
ETH refunds - ECDSA and random beacon


pdyraga:
# Overview 

This issue aims at implementing reimbursements for random beacon and ECDSA operator contracts, including implementing the reimbursement ETH pool, as proposed in https://github.com/keep-network/tbtc-v2/pull/115.

Please treat the idea presented here as just a proposition as the reality may require some adjustments.

# Reimbursements

## Random Beacon

| Function                           | ETH reimbursed? | Slashing? | Can anyone call? |
|------------------------------------|-----------------|-----------|------------------|
| submitDkgResult + approveDkgResult | Yes             | No        | No               |
| notifyDkgTimeout                   | Yes             | No        | Yes              |
| submitRelayEntry                   | Yes             | No        | In practice, No  |
| notifyFailedHeartbeat              | Yes             | No        | Yes              |
| challengeDkgResult                 | No              | Yes       | Yes              |
| reportRelayEntryTimeout            | No              | Yes       | Yes              |
| reportUnauthorizedSigning          | No              | Yes       | Yes              |

## ECDSA

| Function                           | ETH reimbursed? | Slashing? | Can anyone call? |
|------------------------------------|-----------------|-----------|------------------|
| submitDkgResult + approveDkgResult | Yes             | No        | No               |
| notifyDkgTimeout                   | Yes             | No        | Yes              |
| notifySeedTimeout                  | Yes             | No        | Yes              |
| challengeDkgResult                 | No              | Yes       | Yes              |

# Smart contracts

## Reimbursement Pool

Although it would be nice to have separate bookkeeping per application, we want the mechanism to be as cheap as possible and not require a lot of operational work. `ReimbursementPool` will be a single contract holding ETH received from the DAO, managing the gas price ceiling, and authorization of contracts eligible to use that ETH for reimbursements. It should also allow the DAO to pull ETH in case an upgrade is needed. This contract will be located in `keep-core`.

```
contract ReimbursementPool is Ownable {
  /// @notice Static gas cost of submitting a TX.
  uint256 staticGasCost;

  /// @notice Max refundable gas price. Protects against malicious
  ///         operator-miners.
  uint256 maxGasPrice;

  modifier onlyAuthorized {
    require(isAuthorized[msg.sender], "Not authorized for a refund");
  }

  function refund(uint256 gasSpent, address spender) external onlyAuthorized {
    // add constant 
  }

  function authorize(address _contract) external onlyOwner {
[...]

PR #2826, opened 2022-02-15
ETH reimbursement pool for RFC-6

dimpar:
This PR implements [the ETH pool described in RFC-6](https://github.com/keep-network/tbtc-v2/blob/7b3cb19f6860cfc32677b15d5de643a9ee0f0f51/docs/rfc/rfc-6.adoc#32-eth-pool) and [this issue](https://github.com/keep-network/keep-core/issues/2824). ETH pool is used for transaction reimbursements executed on authorized contracts by the DAO.


solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,111 @@
...
+    function withdrawAll(address to) external onlyOwner {
+        require(address(this).balance > 0, "Nothing to withdraw");
+
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool sent, ) = to.call{value: address(this).balance}("");
+        require(sent, "Failed to send Ether");
+    }
+
+    /// @notice Receive ETH
+    receive() external payable {}
```
dimpar, line 110: Should we limit this to DAO only?

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,127 @@
...
+    /// @notice Authorized contracts that can interact with the reimbursment pool.
+    ///         Authorization can be granted and removed by the governance.
+    mapping(address => bool) public isAuthorized;
+
+    /// @notice Static gas of submitting a transaction.
+    uint256 public staticGas;
+
+    /// @notice Max gas price used to reimburse a transaction submitter. Protects
+    ///         against malicious operator-miners.
+    uint256 public maxGasPrice;
```
nkuba, line 30: Could we initialize the values? Either assign them here or in a constructor.

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,127 @@
...
+        uint256 gasPrice = tx.gasprice < maxGasPrice
+            ? tx.gasprice
+            : maxGasPrice;
+
+        uint256 refundAmount = (gasSpent + staticGas) * gasPrice;
+
+        /* solhint-disable avoid-low-level-calls */
+        // slither-disable-next-line low-level-calls
+        (bool sent, ) = receiver.call{value: refundAmount}("");
+        /* solhint-enable avoid-low-level-calls */
```
nkuba, line 56: Would this work?

```suggestion
        // slither-disable-next-line low-level-calls
        (bool sent, ) = receiver.call{value: refundAmount}("");
```

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,127 @@
...
+
+        uint256 gasPrice = tx.gasprice < maxGasPrice
+            ? tx.gasprice
+            : maxGasPrice;
+
+        uint256 refundAmount = (gasSpent + staticGas) * gasPrice;
+
+        /* solhint-disable avoid-low-level-calls */
+        // slither-disable-next-line low-level-calls
+        (bool sent, ) = receiver.call{value: refundAmount}("");
```
nkuba, line 55: Would it be worth emitting an event that a refund was made?

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,127 @@
...
+        /* solhint-enable avoid-low-level-calls */
+        require(sent, "Failed to refund Ether");
+    }
+
+    /// @notice Authorize a contract that can interact with this reimbursment pool.
+    ///         Can be authorized by the governance only.
```
nkuba, line 61: Could we replace `governance` with `owner` across this contract?

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,127 @@
...
+            ? tx.gasprice
+            : maxGasPrice;
+
+        uint256 refundAmount = (gasSpent + staticGas) * gasPrice;
+
+        /* solhint-disable avoid-low-level-calls */
+        // slither-disable-next-line low-level-calls
+        (bool sent, ) = receiver.call{value: refundAmount}("");
+        /* solhint-enable avoid-low-level-calls */
+        require(sent, "Failed to refund Ether");
```
nkuba, line 57: We need to remember to catch this error when we use this function, so we don't end up with a failing transaction. Maybe it would be worth adding a note about it in `@dev` docs?

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,127 @@
...
+        require(
+            address(this).balance >= amount,
+            "Insufficient contract balance"
+        );
+        require(receiver != address(0), "Receiver's address cannot be zero");
+
+        /* solhint-disable avoid-low-level-calls */
+        // slither-disable-next-line low-level-calls
+        (bool sent, ) = receiver.call{value: amount}("");
+        /* solhint-enable avoid-low-level-calls */
```
nkuba, line 107: Can the disable be simplified?
```suggestion
        // slither-disable-next-line low-level-calls
        (bool sent, ) = receiver.call{value: amount}("");
```

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,127 @@
...
+    function withdrawAll(address receiver) external onlyOwner {
+        require(address(this).balance > 0, "Nothing to withdraw");
+        require(receiver != address(0), "Receiver's address cannot be zero");
+
+        /* solhint-disable avoid-low-level-calls */
+        // slither-disable-next-line low-level-calls
+        (bool sent, ) = receiver.call{value: address(this).balance}("");
+        /* solhint-enable avoid-low-level-calls */
+        require(sent, "Failed to send Ether");
+    }
```
nkuba, line 123: Do we really need this function? If the answer is yes, maybe we could simply call the `withdraw`?

```suggestion
    function withdrawAll(address receiver) external onlyOwner {
        withdraw(address(this).balance, receiver);
    }
```

solidity/random-beacon/contracts/ReimbursmentPool.sol
nkuba: IMHO it's good this way.

nkuba (changes requested):

solidity/random-beacon/contracts/ReimbursmentPool.sol
pdyraga: Not sure if it's worth it given we need to keep in mind the extra gas cost of the wrapping code. If we ever need any stats, we can simply find all out-transfers from this contract.

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,127 @@
...
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+contract ReimbursementPool is Ownable, ReentrancyGuard {
+    /// @notice Authorized contracts that can interact with the reimbursment pool.
+    ///         Authorization can be granted and removed by the governance.
+    mapping(address => bool) public isAuthorized;
+
+    /// @notice Static gas of submitting a transaction.
```
pdyraga, line 25: Not only the TX, also all the wrapping code that must be executed for the refund.

solidity/random-beacon/contracts/ReimbursmentPool.sol
pdyraga: I think we need it so that we do not have to race to call `withdraw(uint256 amount, address receiver)` with transactions using the pool. See that it fails if the amount does not match.

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,127 @@
...
+    uint256 public staticGas;
+
+    /// @notice Max gas price used to reimburse a transaction submitter. Protects
+    ///         against malicious operator-miners.
+    uint256 public maxGasPrice;
+
+    event StaticGasUpdated(uint256 staticGas);
+
+    event MaxGasPriceUpdated(uint256 maxGasPrice);
```
pdyraga, line 34: Small note, we could name them `newStaticGas` and `newMaxGasPrice` to clearly state these are the new values.

solidity/random-beacon/contracts/ReimbursmentPool.sol
pdyraga: I would ignore the result of `.call` in this function. For externally owned accounts `call` should always work. If the caller is a smart contract, it's that smart contract's responsibility to make sure Ether can be received. If there is some problem with receiving Ether, we do not need to revert, especially since we would revert at the very end, when the most complex and expensive logic has been already executed and paid for before `refund` is called. If we go this way, let's make sure to explain it in `@dev` section of docs.

solidity/random-beacon/contracts/ReimbursmentPool.sol
pdyraga: If someone wants to donate, there is no reason for stopping them.

pdyraga (commented):
Was looking only at `*.sol`.

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: Actually, thinking a bit more about it, do we need to make static gas var updateable? The cost of the transaction is constant which is 21000 & the cost of the entire refund function is around 16500. Total 37500. Gas amount will stay the same unless we change the refund function in a new contract, where we'll need to adjust the gas costs accordingly. So I'm thinking of making only the `maxGasPrice` updatable.

pdyraga: Opcode cost can change, an example is https://eips.ethereum.org/EIPS/eip-2929. If we want to be able to adjust without redeploying `ReimbursmentPool`, we need to have this value updatable.

dimpar: Ah, in that case, it makes sense to make a static gas cost updatable. 👍

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: Yup, good idea [3138b17](https://github.com/keep-network/keep-core/pull/2826/commits/3138b1714cfab6d81ce0403c4b1b50d987fe73c3)

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: I tried, but it didn't

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: This is a refund function that will be usually called in already heavy gas-wise functions, so adding extra should be probably avoided.

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: [11dc5a9](https://github.com/keep-network/keep-core/pull/2826/commits/11dc5a9fe83218df832396f364040bfd680dc0fa)

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: I'm on the fence between ignoring and wrapping a call in a try-catch on the caller side. The best practice is to have a `require` and if something goes south then a try-catch will inform about the error. I think this approach sort of "decouples" this contract and doesn't care who's going to call it. On the other hand, try-catch will add extra costs and at least for now, we know what contracts will call the `ReimbursmentPool` contract. Anyway, I removed the result of the `.call`, but also open for alternatives. [063786d](https://github.com/keep-network/keep-core/pull/2826/commits/063786d85371ffa16d1c170106990fd94acf1061)

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: Didn't work, solhint is still complaining.

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: Simplified a bit [59b67ee](https://github.com/keep-network/keep-core/pull/2826/commits/59b67ee770f0031653bdc57cda01165da6c9e48c)

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: [3b8beed](https://github.com/keep-network/keep-core/pull/2826/commits/3b8beedcb07572f2c104749bb30ad0b6b16c11d1)

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,391 @@
...
+        to: reimbursementPool.address,
+        value: ethers.utils.parseEther("10.0"), // Send 10.0 ETH
+      })
+    })
+
+    afterEach(async () => {
+      await restoreSnapshot()
+    })
+
+    context("when withdrawing all the funds as a non owner", () => {
```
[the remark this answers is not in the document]
dimpar: part of [3169069](https://github.com/keep-network/keep-core/pull/2826/commits/3169069e0d8d41d3f953811ae1a6f6bc99e68e08)

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,391 @@
...
+            .connect(thirdParty)
+            .withdraw(
+              ethers.utils.parseEther("2.0"),
+              contractToAuthorize.address
+            )
+        ).to.be.revertedWith("Ownable: caller is not the owner")
+      })
+    })
+
+    context("when widhrawing funds as an owner", () => {
```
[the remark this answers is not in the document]
dimpar: part of [3169069](https://github.com/keep-network/keep-core/pull/2826/commits/3169069e0d8d41d3f953811ae1a6f6bc99e68e08)

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,391 @@
...
+      it("should revert", async () => {
+        await expect(
+          reimbursementPool
+            .connect(thirdParty)
+            .refund(ethers.utils.parseEther("2.0"), thirdParty.address)
+        ).to.be.revertedWith("Contract is not authorized for a refund")
+      })
+    })
+
+    context("when contract is authorized", () => {
```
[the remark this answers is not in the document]
dimpar: part of [3169069](https://github.com/keep-network/keep-core/pull/2826/commits/3169069e0d8d41d3f953811ae1a6f6bc99e68e08)

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,391 @@
...
+        await expect(
+          reimbursementPool
+            .connect(thirdParty)
+            .unauthorize(contractToAuthorize.address)
+        ).to.be.revertedWith("Ownable: caller is not the owner")
+      })
+    })
+
+    context("when the caller is the owner", () => {
+      it("should authorize a contract", async () => {
```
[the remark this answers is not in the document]
dimpar: part of [3169069](https://github.com/keep-network/keep-core/pull/2826/commits/3169069e0d8d41d3f953811ae1a6f6bc99e68e08)

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,132 @@
...
+        // slither-disable-next-line low-level-calls,unchecked-lowlevel
+        receiver.call{value: refundAmount}("");
+        /* solhint-enable avoid-low-level-calls */
+    }
+
+    /// @notice Authorize a contract that can interact with this reimbursment pool.
+    ///         Can be authorized by the owner only.
+    /// @param _contract Authorized contract.
+    function authorize(address _contract) external onlyOwner {
+        isAuthorized[_contract] = true;
```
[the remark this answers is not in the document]
dimpar: part of [3169069](https://github.com/keep-network/keep-core/pull/2826/commits/3169069e0d8d41d3f953811ae1a6f6bc99e68e08)

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,132 @@
...
+    /// @param receiver An address where ETH is sent.
+    function withdrawAll(address receiver) external onlyOwner {
+        withdraw(address(this).balance, receiver);
+    }
+
+    /// @notice Withdraws ETH amount from this pool which are sent to a given
+    ///         address. Can be set by the owner only.
+    /// @param amount Amount to withdraw from the pool.
+    /// @param receiver An address where ETH is sent.
+    function withdraw(uint256 amount, address receiver) public onlyOwner {
```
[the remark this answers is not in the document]
dimpar: I'm not sure if there's much value to it because:
- we have `requires` that will tell if the transaction fails, especially this one: `require(sent, "Failed to send Ether")`
- it will be executed by the gov which can see right away if the funds were received at the provided address
- it adds extra cost

Unless you have a specific case in mind when it's needed?

solidity/random-beacon/test/ReimbursmentPool.test.ts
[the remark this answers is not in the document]
nkuba: One more test case please: `refund` when the contract's balance is not high enough

solidity/random-beacon/contracts/ReimbursmentPool.sol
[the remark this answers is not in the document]
nkuba: It can be useful for maintainers/operators to know that some movement happened in the contract and there is a risk that transactions may not be reimbursable. 
`withdraw` is not a transaction that is expected to be executed often, so I don't think we should be worried about adding an event here.

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,468 @@
...
+const ZERO_ADDRESS = ethers.constants.AddressZero
+const { createSnapshot, restoreSnapshot } = helpers.snapshot
+const { provider } = waffle
+const fixture = async () => reimbursmentPoolDeployment()
+
+describe("ReimbursementPool - Pool", () => {
+  let owner: SignerWithAddress
+  let thirdParty: SignerWithAddress
+  let refundee: SignerWithAddress
+  let contractToAuthorize: SignerWithAddress
```
nkuba, line 18:
```suggestion
  let authorizedContract: SignerWithAddress
```

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,468 @@
...
+          )
+          expect(refundeeBalanceDiff).to.be.gt(
+            ethers.utils.parseUnits("73000", "gwei")
+          )
+
+          expect(refundeeBalanceDiff).to.be.lt(
+            ethers.utils.parseUnits("146000", "gwei")
+          )
+
+          await expect(tx).not.to.emit(reimbursementPool, "SendingEtherFailed")
```
nkuba, line 287: Let's extract this check to a separate `it` test case.

dimpar: [b2781e1](https://github.com/keep-network/keep-core/pull/2826/commits/b2781e16230fd4935ec19337a6b5cd5792a60526)

solidity/random-beacon/test/ReimbursmentPool.test.ts
dimpar: Changed to a more general `thirdPartyContract`, cause `authorizedContract` also won't fit in the `authorize` tests, ie. it's not yet authorized. [b2781e1](https://github.com/keep-network/keep-core/pull/2826/commits/b2781e16230fd4935ec19337a6b5cd5792a60526)

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -281,6 +318,38 @@ describe("ReimbursementPool - Pool", () => {
...
+          await expect(
+            reimbursementPool
+              .connect(contractToAuthorize)
+              .refund(50000, ZERO_ADDRESS)
+          ).to.be.revertedWith("Receiver's address cannot be zero")
+        })
+      })
+
+      context("when no funds available in the pool", () => {
+        it("should emit SendingEtherFailed event", async () => {
```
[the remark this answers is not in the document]
dimpar: [b2781e1](https://github.com/keep-network/keep-core/pull/2826/commits/b2781e16230fd4935ec19337a6b5cd5792a60526)

solidity/random-beacon/contracts/ReimbursmentPool.sol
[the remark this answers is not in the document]
dimpar: [b2781e1](https://github.com/keep-network/keep-core/pull/2826/commits/b2781e16230fd4935ec19337a6b5cd5792a60526)

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,505 @@
...
+          .to.be.true
+
+        await expect(tx)
+          .to.emit(reimbursementPool, "thirdPartyContract")
+          .withArgs(thirdPartyContract.address)
+      })
+    })
+  })
+
+  describe("unuthorize", () => {
```
nkuba, line 422: We should have a before block where we first authorize the address.

nkuba (changes requested):

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    event MaxGasPriceUpdated(uint256 newMaxGasPrice);
+
+    event SendingEtherFailed(uint256 refundAmount, address receiver);
+
+    event thirdPartyContract(address thirdPartyContract);
+
+    event UnthirdPartyContract(address unthirdPartyContract);
+
+    event WithdrawnFunds(uint256 withdrawnAmount, address receiver);
```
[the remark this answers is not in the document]
dimpar: [c03c038](https://github.com/keep-network/keep-core/pull/2826/commits/c03c038461a772771d7ae100b372f9c81d46c230)

solidity/random-beacon/test/ReimbursmentPool.test.ts
dimpar: good point [c03c038](https://github.com/keep-network/keep-core/pull/2826/commits/c03c038461a772771d7ae100b372f9c81d46c230)

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    event StaticGasUpdated(uint256 newStaticGas);
+
+    event MaxGasPriceUpdated(uint256 newMaxGasPrice);
+
+    event SendingEtherFailed(uint256 refundAmount, address receiver);
+
+    event thirdPartyContract(address thirdPartyContract);
+
+    event UnthirdPartyContract(address unthirdPartyContract);
```
pdyraga, line 42: Why not `AuthorizeContractForRefund` or `DeauthorizeContractForRefund`? Alternatively just `AuthorizeContract` or `DeauthorizeContract`?

dimpar: Uh, that's clearly a global scope find-replace error, thx for catching it. They were named previously  `AuthorizeContract` and `UnauthorizeContract`. Changed in [28daf23](https://github.com/keep-network/keep-core/pull/2826/commits/28daf23db05be69b14c19e60676812fe7a831cf5)

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,517 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions */
+
+import { ethers, waffle, helpers } from "hardhat"
+import { expect } from "chai"
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers"
+import type { ContractTransaction } from "ethers"
+import { reimbursmentPoolDeployment, params } from "./fixtures"
+import type { ReimbursementPool } from "../typechain"
```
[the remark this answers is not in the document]
dimpar: I did a `yarn format:fix` and haven't noticed any warnings, so not sure what are you referring to?

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,517 @@
...
+
+const ZERO_ADDRESS = ethers.constants.AddressZero
+const { createSnapshot, restoreSnapshot } = helpers.snapshot
+const { provider } = waffle
+const fixture = async () => reimbursmentPoolDeployment()
+
+describe("ReimbursementPool - Pool", () => {
```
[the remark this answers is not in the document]
dimpar: Removed

solidity/random-beacon/test/fixtures/index.ts
```diff
@@ -47,6 +47,8 @@ export const params = {
   unauthorizedSigningSlashingAmount: to1e18(100000),
   minimumAuthorization: to1e18(100000),
   authorizationDecreaseDelay: 0,
+  reimbursmentPoolStaticGas: 37500,
```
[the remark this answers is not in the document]
dimpar: Static gas cost includes the execution of code in the `refund()` + transaction cost. This number was captured at the moment of implementing the tests a while ago. After addressing some of the comments the avg. gas bumped to ~41k for the `refund()`. The static gas will be passed in the constructor during the deployment and should be probably revisited then.

```
gas for refund (min, max, avg) 40433 | 41893 | 41075  
```

I don't think we can precisely tell what is the cost of the `refund()`. We have min, max and avg gas usage depending on the situation. If we assume that the avg. is 41k then at most a caller will be net-negative around 800-900gas. Assuming the gas is 100gwei that is around 0,00008 ETH. Or can be a bit net-positive if a user hits a min gas path.

More integration tests between ECDSA Wallet registry / random-beacon and Reimbursement pool will be added in the next PRs where the refunds are checked.

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,517 @@
...
+  let reimbursementPool: ReimbursementPool
+
+  // prettier-ignore
+  before(async () => {
+    [owner, thirdParty, thirdPartyContract, refundee] =
+      await ethers.getSigners()
+  })
+
+  beforeEach("load test fixture", async () => {
+    const contracts = await waffle.loadFixture(fixture)
```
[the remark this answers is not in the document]
dimpar: changed

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,517 @@
...
+        )
+
+        expect(reimbursementPoolBalance).to.be.equal(
+          ethers.utils.parseEther("1.0")
+        )
+      })
+    })
+
+    context("when the owner funds a reimbursment pool", () => {
+      it("should withdraw entire ETH balance", async () => {
```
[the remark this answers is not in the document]
dimpar: copy-pasta error, changed.

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,517 @@
...
+        expect(reimbursementPoolBalance).to.be.equal(
+          ethers.utils.parseEther("1.0")
+        )
+      })
+    })
+  })
+
+  describe("withdrawAll", () => {
+    beforeEach(async () => {
+      await createSnapshot()
```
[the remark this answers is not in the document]
dimpar: removed `beforeEach`

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,517 @@
...
+            .refund(50000, refundee.address)
+
+          const refundeeBalanceAfter = await provider.getBalance(
+            refundee.address
+          )
+          const refundeeBalanceDiff = refundeeBalanceAfter.sub(
+            refundeeBalanceBefore
+          )
+          expect(refundeeBalanceDiff).to.be.gt(
+            ethers.utils.parseUnits("73000", "gwei")
```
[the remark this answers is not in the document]
dimpar: I've added more descriptions. The idea of the lower range at the moment of writing tests was to check if the gas is greater than 73000 assuming the gas price was 1gwei. It was just to show it's not 0, but I see that it might be a bit confusing, so I changed it just to `>0`. The higher range is calculated based on: `gas*tx.gasprice` which is a bit less than 145k. The precise number in the other test is based on the `maxGasPrice` which is set to 1gwei for testing purposes.

solidity/random-beacon/test/ReimbursmentPool.test.ts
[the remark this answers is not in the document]
pdyraga: image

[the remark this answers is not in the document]
dimpar: Strange, can't recreate it on my local. I removed `node_modules, typechain, build, cache` folders and do the `yarn lint`. No warnings. I'm on node 14.9.0. @nkuba do you see the same warnings on your box?

solidity/random-beacon/test/ReimbursmentPool.test.ts
```diff
@@ -0,0 +1,511 @@
...
+          .to.be.true
+
+        await expect(tx)
+          .to.emit(reimbursementPool, "AuthorizeContract")
+          .withArgs(thirdPartyContract.address)
+      })
+    })
+  })
+
+  describe("unuthorize", () => {
```
nkuba, line 416:
```suggestion
  describe("unauthorize", () => {
```

nkuba (changes requested):

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,150 @@
...
+    /// @dev Ignoring the result of sending ETH to a receiver is made on purpose.
+    ///      For EOA receiving ETH should always work. If a receiver is a smart
+    ///      contract, then we do not want to fail a transaction, because in some
+    ///      cases the refund is done at the very end of multiple calls where all
+    ///      the previous calls were already paid off. It is a receiver's smart
+    ///      contract resposibility to make sure it can receive ETH.
+    /// @dev Only authorized contracts are allowed calling this function.
+    /// @param gasSpent Gas spent on a transaction that needs to be reimbursed.
+    /// @param receiver Address where the reimbursment is sent.
+    function refund(uint256 gasSpent, address receiver) external nonReentrant {
```
pdyraga, line 64: What is the gas cost difference with/without non-reentrant? All approved contracts are supposed to be trusted.

solidity/random-beacon/contracts/ReimbursmentPool.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    event StaticGasUpdated(uint256 newStaticGas);
+
+    event MaxGasPriceUpdated(uint256 newMaxGasPrice);
+
+    event SendingEtherFailed(uint256 refundAmount, address receiver);
+
+    event AuthorizeContract(address thirdPartyContract);
+
+    event UnauthorizeContract(address thirdPartyContract);
```
pdyraga, line 42: All other events are named using the past tense. Can we do the same for these two?

solidity/random-beacon/test/fixtures/index.ts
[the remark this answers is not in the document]
pdyraga: It would be good to establish this number before merging this PR. `ReimbursmentPool` code should not change in the future so there is no better moment than now, especially that as you say, we will have more tests added in the following PRs and we should be able to validate our choice this way.

I think the strategy should be: if we can't use a precise number, we establish one that is returning a tiny bit more than actually spent. If we see a frontrunning in the network, we will lower it and everyone will receive a tiny bit less than what they have actually spent.

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: With `nonReentrant`: 40433  ·      41893  ·      41075 
Without:  38083  ·      39543  ·      38725

If per prev. comments we take `max`, then the diff is `3168`. 

Yes, there are supposed to be trusted, but what if this trusted contract has a bug that can enable reentrancy here? The diff is not that small, but also not too big compared to the overall cost and having in mind the risk of draining the pool maybe we should leave the `nonReentrant` modifier in place?

solidity/random-beacon/contracts/ReimbursmentPool.sol
dimpar: [ffab18c](https://github.com/keep-network/keep-core/pull/2826/commits/ffab18c52e13342a15d6195aa3c8dfd848c05da5)

solidity/random-beacon/test/ReimbursmentPool.test.ts
dimpar: [cbc5623](https://github.com/keep-network/keep-core/pull/2826/commits/cbc5623a574ef8563b12801b3ead13da9b05b9e8)

pdyraga (approved):

nkuba (approved):

PR #2827, opened 2022-02-16
Adjust Random Beacon DKG after the changes in ECDSA DKG

nkuba:
Refs: https://github.com/keep-network/keep-core/issues/2816

This PR implements changes introduced in the ECDSA project that are common to both projects.

- [x] emit the entire DKG result in DkgResultSubmitted
- [x] register the group only when DKG result was accepted
- [x] wrap calls to TokenStaking.seize in a try-catch when challenging the result,
- [x] prefix duplicated contract names (DKG -> BeaconDkg, DKGValidator -> BeaconDkgValidator)
- [x] ensure consistent code formatting with ecdsa project

This PR doesn't cover reimbursements in ETH instead of T token.

PR #2840, opened 2022-02-22
Removing slashing factor

dimpar:
Depends on https://github.com/keep-network/keep-core/pull/2827

This [PR aims](https://github.com/keep-network/keep-core/pull/2827#issuecomment-1047732716) to simplify slashing amount calculation. The core idea of how it is calculated [stays the same](https://github.com/keep-network/keep-core/pull/2772#discussion_r776784840) but the difference is in the implementation, ie. removing the slashing factor.

Before:
```
Bytes size for RandomBeacon: 21.644
Gas for submitRelayEntry(): 214122  ·   495215  ·   354669 
```

After:
```
Bytes size for RandomBeacon: 21.606
Gas for submitRelayEntry(): 213826  ·   495062  ·   354444  
```


nkuba (commented):
I'm so sorry, I just noticed I have pending review comments that I've never submitted!

solidity/random-beacon/contracts/libraries/Relay.sol
```diff
@@ -131,15 +131,10 @@ library Relay {
         Data storage self,
         bytes calldata entry,
         bytes storage groupPubKey
-    ) internal returns (uint256 slashingAmount) {
+    ) internal returns (uint256) {
```
nkuba, line 134: Could we return a `uint96` value? This is expected by [`TokenStaking.seize`](https://github.com/threshold-network/solidity-contracts/blob/702aa5f12242e885e7b0c2b3228bb788439d30a8/contracts/staking/TokenStaking.sol#L1125) and [`TokenStaking.slash`](https://github.com/threshold-network/solidity-contracts/blob/702aa5f12242e885e7b0c2b3228bb788439d30a8/contracts/staking/TokenStaking.sol#L1110) functions.

dimpar: yea, we can do here. It will add some gas, but it needs to be done somewhere anyway [221a3ce](https://github.com/keep-network/keep-core/pull/2840/commits/221a3ceac74de426ff4c22e55b1c08e007da3405)

solidity/random-beacon/contracts/test/RelayStub.sol
```diff
@@ -21,7 +21,15 @@ contract RelayStub {
         relay.currentRequestStartBlock = uint64(block.number);
     }
 
-    function getSlashingFactor() external view returns (uint256) {
-        return relay.getSlashingFactor();
+    function setRelayEntrySubmissionFailureSlashingAmount(
+        uint256 relayEntrySubmissionFailureSlashingAmount
```
nkuba, line 25:
```suggestion
        uint96 relayEntrySubmissionFailureSlashingAmount
```

solidity/random-beacon/contracts/test/RelayStub.sol
```diff
@@ -21,7 +21,15 @@ contract RelayStub {
...
+    ) external {
+        relay.setRelayEntrySubmissionFailureSlashingAmount(
+            relayEntrySubmissionFailureSlashingAmount
+        );
+    }
+
+    function calculateSlashingAmount() external view returns (uint256) {
```
nkuba, line 32:
```suggestion
    function calculateSlashingAmount() external view returns (uint96) {
```

nkuba (changes requested):
I left two comments that should be fixed in this PR. Besides that, we should update the slashing amounts to be `uint96` across the code, but this is already handled in https://github.com/keep-network/keep-core/pull/2903/commits/6539ba1906c8b890baee9e853ed7c20601e2eec8.

solidity/random-beacon/contracts/test/RelayStub.sol
dimpar: [5d73243](https://github.com/keep-network/keep-core/pull/2840/commits/5d7324365bf0f0b811f6237a487cf8479e1a76a7)

solidity/random-beacon/contracts/test/RelayStub.sol
dimpar: [5d73243](https://github.com/keep-network/keep-core/pull/2840/commits/5d7324365bf0f0b811f6237a487cf8479e1a76a7)

nkuba (approved):

PR #2842, opened 2022-02-23
Fix creating a contract instance

r-czajkowski:
Fix creating a contract instance from the hardhat artifact. In some
cases, the artifact doesn't include `receipt` object and
`transactionHash` so we need to set the `deployedAtBlock` to `1`. This
is not an ideal solution because we should fetch past events from the
deployment block- some node providers (eg. Alchemy) block requests if
the block range is greater than 2k.


michalsmiarowski (approved):
Looking great  🌊 🚢 🌊

PR #2845, opened 2022-02-23
Dashboard docker image build updates

nkuba:
This PR changes docker image build process for Token Dashboard that came up during the v1.17.0 release.

`@threshold-network/solidity-contracts` dependency has `postinstall` script that works perfectly fine locally but fails inside a docker image building. We don't need to execute this script in docker as it's used for the deployment of contracts, but here we are expecting final artifacts to be provided. This is a solution proposed by @r-czajkowski 🙇🏻‍♂️ 

Another change we add here is docker image build executed as part of GitHub actions, to confirm that the build works and won't surprise us on mainnet release.

### DIFF
--- a/.github/workflows/dashboard-v1-mainnet.yml
+++ b/.github/workflows/dashboard-v1-mainnet.yml
@@ -2,9 +2,9 @@ name: Token Dashboard / Mainnet
 
 on:
   pull_request:
-    branches: [ main ]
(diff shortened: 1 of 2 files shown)


r-czajkowski (approved):
LGTM! 👍

michalinacienciala (approved):

PR #2846, opened 2022-02-24
Fix displaying undelegated stakes in app->threshold

michalsmiarowski:
There was a bug when the stake that was undelegated and recovered displayed in
the tables in Applications->Threshold page. This PR fixes that.

Main changes:
- fixed a bug that caused undelegated and recovered stakes to be displayed in 
Application -> Threshold page
- updates the threshold auth data after doing an undelegation so this way user 
don't have to refresh the page to see the actual changes


r-czajkowski (approved):
👍

Issue #2847, opened 2022-02-24
Critical parameters delay in governance contracts does not really make sense


pdyraga:
Refs: #2835 
2-week critical governance parameter does not make sense for ECDSA `WalletRegistry` given that wallets will have to remain active for months. We could swallow it for the `RandomBeacon` but for `WalletRegistry` 2 weeks does not make any sense. On the other hand, increasing the critical parameters update delay to several months is also suboptimal.

One option would be to eliminate governance contracts completely and just deploy TimeLock contract. On the other hand, TimeLock updates are more complex to set up than simple calls to the existing governance contracts, and the structure of `update*` functions in `RandomBeacon` and `WalletRegistry` is more complicated.

What I would do is to introduce just one governable delay for parameter updates and eliminate the separation between critical/standard parameters given it creates a false assumption of safety.

Issue #2834, opened 2022-02-22
Integrate T staking contract with ECDSA Wallet Registry


pdyraga:
Refs #2715

T staking contract works differently than KEEP staking contract in terms of the eligible stake. Every application authorized by T DAO can have its separate risk ceiling set by the stake authorizer. The application - which is `WalletRegistry` contract in our case - controls what is the minimum authorization and when it can be decreased.

`WalletRegistry` needs to implement [`IApplication` interface](https://github.com/threshold-network/solidity-contracts/blob/main/contracts/staking/IApplication.sol) and update `SortitionPool` every time authorization increases and decreases. `SortitionPool` does no longer need to consult staking contract and `WalletRegistry` - which is the contract owner - can provide the authorized stake amount. The relationship should be: `T TokenStaking <-> WalletRegistry <-> SortitionPool`.

Just like in the case of #2715, we first need to decide how authorization increase and decrease mechanism should work, especially how rewards are allocated during undelegation period.

One additional requirement is to figure out how slashing should work. That is, if tBTC v2 `Bridge` (wallet owner) wants to slash someone, is it doing it via `WalletRegistry`? If not, then how?

PR #2848, opened 2022-02-24
Authorization parameters for WalletRegistry contract

pdyraga:
Refs: #2834 

Introduced two authorization-related parameters in the WalletRegistry contract:
- minimum authorization
- authorization decrease delay

`Authorization` library will be further developed to enable integration with Threshold `TokenStaking` contract. Changes introducing governable parameters are included separately in this PR to keep the diff size sane.

Side note: after looking at `WalletRegistryGovernance`, I've decided to open https://github.com/keep-network/keep-core/issues/2847 to improve how governance contract works for Random Beacon and ECDSA.

### DIFF
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -14,6 +14,7 @@
 
 pragma solidity ^0.8.9;
 
+import "./libraries/EcdsaAuthorization.sol";
 import "./libraries/EcdsaDkg.sol";
 import "./libraries/Wallets.sol";
 import "./EcdsaDkgValidator.sol";
@@ -39,10 +40,12 @@ interface IWalletStaking {
 }
 
 contract WalletRegistry is IRandomBeaconConsumer, Ownable {
+    using EcdsaAuthorization for EcdsaAuthorization.Data;
     using EcdsaDkg for EcdsaDkg.Data;
     using Wallets for Wallets.Data;
 
     // Libraries data storages
+    EcdsaAuthorization.Data internal authorization;
     EcdsaDkg.Data internal dkg;
     Wallets.Data internal wallets;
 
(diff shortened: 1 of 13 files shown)


solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -170,6 +178,28 @@ contract WalletRegistry is IRandomBeaconConsumer, Ownable {
...
     }
 
+    /// @notice Updates the values of authorization parameters.
+    /// @dev Can be called only by the contract owner, which should be the
+    ///      wallet registry governance contract. The caller is responsible for
+    ///      validating parameters.
+    /// @param _minimumAuthorization New minimum authorization amount
+    /// @param _authorizationDecreaseDelay New authorization decrease delay in
+    ///        seconds
+    function updateAuthorizationParameters(
```
nkuba, line 188: We should initialize the values for deployment in the constructor.

pdyraga: Good point, see the two recent commits please.

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -447,6 +542,33 @@ contract WalletRegistryGovernance is Ownable {
...
+        view
+        returns (uint256)
+    {
+        return
+            getRemainingChangeTime(
+                minimumAuthorizationChangeInitiated,
+                CRITICAL_PARAMETER_GOVERNANCE_DELAY
+            );
+    }
+
```
dimpar, line 559: Can we add a comment for this function as well?

pdyraga: Yes, good catch! Please see https://github.com/keep-network/keep-core/pull/2848/commits/c68ce675c83f6061a92380a3f3723dbaa1e1c91b

dimpar (approved):

nkuba (approved):

Issue #2833, opened 2022-02-22
Integrate WalletRegistry with tBTC Bridge contract


pdyraga:
tBTC `Bridge` contract that is the wallet owner can [`requestNewWallet` on the `WalletRegistry` contract](https://github.com/keep-network/keep-core/blob/acc1f639936a10dbd9a3bb3610fb74ea3a12f5b8/solidity/ecdsa/contracts/WalletRegistry.sol#L280). Currently, when DKG result is approved [with a call to `approveDkgResult`](https://github.com/keep-network/keep-core/blob/acc1f639936a10dbd9a3bb3610fb74ea3a12f5b8/solidity/ecdsa/contracts/WalletRegistry.sol#L347), tBTC `Bridge` is not informed a new wallet has been created.

We need to integrate `WalletRegistry` with tBTC v2 `Bridge`. The easiest option seems to add a callback to `Bridge` that would execute as a part of `approveDkgResult` transaction.

Separately from the callback, we need to figure out how wallet ID should be represented and how wallet pubkey should be stored so that fraud proofs are possible (see https://github.com/keep-network/tbtc-v2/pull/133) and so that it is possible to validate the wallet pubkey hash [is known and belongs to an active wallet](https://github.com/keep-network/tbtc-v2/blob/73fcd4832ad1a1cea43d9826b49f634453337f24/solidity/contracts/bridge/Bridge.sol#L575) when processing the sweep proof.

PR #2854, opened 2022-02-28
ECDSA Wallets: Callback to the Wallet Owner after wallet creation

nkuba:
Wallet Owner has to be notified when a requested wallet gets created. We handle this with a callback function that submits to the Wallet Owner a keccak256 hash of the wallet's public key. 

It is required that the Wallet Owner contract implements IWalletOwner interface:
```solidity
interface IWalletOwner {
    /// @notice Callback function executed once a new wallet is created.
    /// @dev Should be callable only by the Wallet Registry.
    /// @param publicKeyHash Keccak256 hash of the wallet's public key. It is
    ///        considered an unique wallet identifier.
    function __notifyWalletCreated(bytes32 publicKeyHash) external;
}
```

Refs https://github.com/keep-network/keep-core/issues/2833

### DIFF
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -398,7 +398,12 @@ contract WalletRegistry is IRandomBeaconConsumer, Ownable {
         //slither-disable-next-line redundant-statements
         misbehavedMembers;
 
-        try walletOwner.notifyEcdsaWalletCreated(publicKeyHash) {} catch {
+        try
+            walletOwner.notifyEcdsaWalletCreated(
+                publicKeyHash,
+                dkgResult.groupPubKey
+            )
+        {} catch {
             // Should never happen but we want to ensure a non-critical path
             // failure from an external contract does not stop the DKG to complete.
             // slither-disable-next-line reentrancy-events
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -184,9 +184,12 @@ contract WalletRegistry is IRandomBeaconConsumer, Ownable {
     ///      wallet registry governance contract. The caller is responsible for
     ///      validating parameters.
     /// @param _randomBeacon Random Beacon address.
-    function upgradeRandomBeacon(address _randomBeacon) external onlyOwner {
-        randomBeacon = IRandomBeacon(_randomBeacon);
-        emit RandomBeaconUpgraded(_randomBeacon);
+    function upgradeRandomBeacon(IRandomBeacon _randomBeacon)
+        external
+        onlyOwner
+    {
+        randomBeacon = _randomBeacon;
+        emit RandomBeaconUpgraded(address(_randomBeacon));
     }
 
     /// @notice Updates the values of authorization parameters.
@@ -275,16 +278,17 @@ contract WalletRegistry is IRandomBeaconConsumer, Ownable {
     /// @notice Updates the values of the wallet parameters.
     /// @dev Can be called only by the contract owner, which should be the
     ///      wallet registry governance contract. The caller is responsible for
-    ///      validating parameters.
+    ///      validating parameters. The wallet owner has to implement `IWalletOwner`
+    ///      interface.
     /// @param _walletOwner New wallet owner address.
-    function updateWalletOwner(address _walletOwner) external onlyOwner {
+    function updateWalletOwner(IWalletOwner _walletOwner) external onlyOwner {
         require(
-            _walletOwner != address(0),
+            address(_walletOwner) != address(0),
             "Wallet owner address cannot be zero"
         );
 
-        walletOwner = IWalletOwner(_walletOwner);
-        emit WalletOwnerUpdated(_walletOwner);
+        walletOwner = _walletOwner;
+        emit WalletOwnerUpdated(address(_walletOwner));
     }
 
     /// @notice Registers the caller in the sortition pool.
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -110,8 +110,6 @@ contract WalletRegistry is IRandomBeaconConsumer, Ownable {
         bytes32 indexed dkgResultHash
     );
 
-    event WalletOwnerNotificationFailed(bytes32 indexed publicKeyHash);
-
     event DkgMaliciousResultSlashed(
         bytes32 indexed resultHash,
         uint256 slashingAmount,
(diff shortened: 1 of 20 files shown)


solidity/ecdsa/contracts/api/IWalletOwner.sol
```diff
@@ -0,0 +1,23 @@
...
+//                           Trust math, not hardware.
+
+pragma solidity ^0.8.9;
+
+interface IWalletOwner {
+    /// @notice Callback function executed once a new wallet is created.
+    /// @dev Should be callable only by the Wallet Registry.
+    /// @param publicKeyHash Keccak256 hash of the wallet's public key. It is
+    ///        considered an unique wallet identifier.
+    function notifyWalletCreated(bytes32 publicKeyHash) external;
```
nkuba, line 22: I was also thinking about naming this function `__newWalletCreatedCallback` to be consistent with Random Beacon's `__beaconCallback` function.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -390,6 +398,12 @@ contract WalletRegistry is IRandomBeaconConsumer, Ownable {
         //slither-disable-next-line redundant-statements
         misbehavedMembers;
 
+        try walletOwner.notifyEcdsaWalletCreated(publicKeyHash) {} catch {
```
nkuba, line 401: Do we want to introduce a gas ceiling for this call similar to what we did in the random beacon callback? This function will be called on the wallet owner contract which we expect to be safe to be called and managed by governance. Introducing a gas limit would introduce complexity and increase the contract size.

michalinacienciala:
Not directly related to the scope of this PR, but maybe you could fix typo in the [L374 of WalletRegistry.sol](https://github.com/keep-network/keep-core/blob/main/solidity/ecdsa/contracts/WalletRegistry.sol#L374), since you're doing changes in that file? I think instead `resultSubmissionTimeout` it should be `submitterPrecedencePeriodLength`...

solidity/ecdsa/contracts/WalletRegistry.sol
pdyraga: I don't think we should have the gas limit for the reasons you described. Additionally, two more comments:

1. I do not think we should wrap the notification with a try-catch. Not completing DKG is not a slasheable event and wallet owner is a trusted contract. If the wallet owner was not notified about the wallet created, the entire DKG was for nothing. `notifyEcdsaWalletCreated` should never fail but even if there was a temporary problem in the contract, we could let the `approveDkgResult` to fail and have it submitted one more time later, once that temporary problem in the wallet owner contract gets resolved.
2. Instead of passing `publicKeyHash`, why don't we pass the wallet public key as a parameter? Right now, calling `notifyEcdsaWalletCreated` requires the `Bridge` to call `getWalletPublicKey` on the registry and contracts are interacting back and forth.

solidity/ecdsa/contracts/api/IWalletOwner.sol
pdyraga: Yeah, not a bad idea. This convention is not used in `IApplication` interface calling application contracts from T `TokenStaking` but at least contracts inside `keep-network` org would use the same convention.

nkuba: 80a617cf1729668f0389a9c2d9018f467e675169

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -165,7 +168,27 @@ contract WalletRegistryGovernance is Ownable {
...
+    ///      walletOwner has not been set before. It doesn't enforce a governance
+    ///      delay for the initial update. Any subsequent updates should be performed
+    ///      with beginWalletOwnerUpdate/finalizeWalletOwnerUpdate with respect
+    ///      of a governance delay.
+    /// @param _walletOwner The Wallet Owner's address
+    function initializeWalletOwner(address _walletOwner) external onlyOwner {
+        require(
+            address(walletRegistry.walletOwner()) == address(0),
+            "Wallet Owner already initialized"
+        );
```
pdyraga, line 185: It should be possible to update the wallet owner, let's not handcuff ourselves. It may happen we may need to upgrade `Bridge` contract address.

solidity/ecdsa/test/WalletRegistry.WalletOwner.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  describe("approveDkgResult", async () => {
+    let dkgResult: DkgResult
+    let submitter: SignerWithAddress
+
+    before(async () => {
+      await createSnapshot()
+
+      await walletOwner.__ecdsaWalletCreatedCallback.reverts(
+        "wallet owner internal error"
+      )
```
pdyraga, line 47: Why do we need it here?

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
nkuba: There is a standard path for `walletOwner` update with `beginWalletOwnerUpdate` and `finalizeWalletOwnerUpdate`.
`initializeWalletOwner` is here as we cannot set the initial value in the `WalletRegistry` constructor due to the cyclic dependency between contracts.

solidity/ecdsa/test/WalletRegistry.WalletOwner.test.ts
nkuba: I have no idea how it got here 😕  I'll remove it!

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
pdyraga: FWIW, we will have to revisit the initialization of this contract. We have more logic in `WalletRegistry` now and I think it should be deployed behind a transparent upgradeable proxy. If so, some constructor parameters will need to be moved to initialize function.

nkuba: Sure, let's just make sure we register it in the backlog.

solidity/ecdsa/test/WalletRegistry.WalletOwner.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        await walletOwner.__ecdsaWalletCreatedCallback.reset()
+      })
+
+      it("should revert", async () => {
+        // FIXME: For some reason this check doesn't work with the expected error message
+        // await expect(tx).to.be.revertedWith("wallet owner internal error")
+        await expect(tx).to.be.reverted
+      })
+
+      it("should call random beacon", async () => {
```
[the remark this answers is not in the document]
nkuba: 96e216a828ee8e3d79a99b22e65a6be498c60549

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
pdyraga: https://github.com/keep-network/keep-core/issues/2861 and https://github.com/keep-network/tbtc-v2/issues/161.

pdyraga (approved):

PR #2855, opened 2022-02-28
Update threshold auth data after staking

michalsmiarowski:
Update threshold auth data after creating a delegation, so it updates 
immediately without refreshing

### DIFF
--- a/solidity-v1/dashboard/src/reducers/threshold-authorization.js
+++ b/solidity-v1/dashboard/src/reducers/threshold-authorization.js
@@ -139,7 +139,7 @@ const addStakeToAuthData = (
     },
     isStakedToT: false,
     isFromGrant: _isFromGrant,
-    canBeMovedToT: _isFromGrant,
+    canBeMovedToT: !_isFromGrant,
   }
 
   const updatedAuthData = [...authData, newStake]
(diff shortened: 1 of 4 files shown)


solidity-v1/dashboard/src/reducers/threshold-authorization.js
```diff
@@ -105,6 +112,38 @@ const updateThresholdAuthData = (authData, { operatorAddress }) => {
...
+    beneficiaryAddress,
+    operatorAddress,
+    stakeAmount,
+    contract: {
+      contractName: AUTH_CONTRACTS_LABEL.THRESHOLD_TOKEN_STAKING,
+      operatorContractAddress: operatorContractAddress,
+      isAuthorized: false,
+    },
+    isStakedToT: false,
+    isFromGrant: isFromGrant ? isFromGrant : false,
```
r-czajkowski, line 139: Can we `!!isFromGrant` or `Boolean(isFromGrant)`?

solidity-v1/dashboard/src/reducers/threshold-authorization.js
```diff
@@ -105,6 +112,38 @@ const updateThresholdAuthData = (authData, { operatorAddress }) => {
...
+    canBeMovedToT: false,
```
r-czajkowski, line 140: The stake created from liquid tokens can be moved to T network. So we should `canBeMovedToT: isFromGrant`.

r-czajkowski (changes requested):

solidity-v1/dashboard/src/reducers/threshold-authorization.js
r-czajkowski: You are 💯 right! 🦅 👁️

PR #2858, opened 2022-03-07
ECDSA Wallets: Store uncompressed public key

nkuba:
We want to store a public key of the wallet so it can be used by the wallet owner (Bridge) for transactions handling. The Bridge will use both compressed and uncompressed public keys. We store the longer one (uncompressed) that can be easily converted to the compressed format.

We also renamed `publicKeyHash` to walletID as this reflects better what this property is. Using the name `publicKeyHash` was confusing in the external owner contract as there may be another key hash in use, but this is an implementation detail of what hashing function we use for identification.

Please notice that gas costs went up:

BEFORE:
|  WalletRegistry.approveDkgResult | 210844 | 225949 | 214789 │

AFTER:
|  WalletRegistry.approveDkgResult | 257700 | 257940 | 257830 |

Depends on https://github.com/keep-network/keep-core/pull/2854
Refs: https://github.com/keep-network/tbtc-v2/pull/159

### DIFF
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -502,7 +502,7 @@ contract WalletRegistry is IRandomBeaconConsumer, IWalletRegistry, Ownable {
         view
         returns (bytes memory)
     {
-        return wallets.registry[walletID].publicKey;
+        return wallets.getWalletPublicKey(walletID);
     }
 
     /// @notice Checks if a wallet with the given ID is registered.
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -107,7 +107,7 @@ contract WalletRegistry is IRandomBeaconConsumer, IWalletRegistry, Ownable {
     event DkgSeedTimedOut();
 
     event WalletCreated(
-        bytes32 indexed publicKeyHash,
+        bytes32 indexed walletID,
         bytes32 indexed dkgResultHash
     );
 
@@ -391,18 +391,19 @@ contract WalletRegistry is IRandomBeaconConsumer, IWalletRegistry, Ownable {
     function approveDkgResult(EcdsaDkg.Result calldata dkgResult) external {
         uint32[] memory misbehavedMembers = dkg.approveResult(dkgResult);
 
-        bytes32 publicKeyHash = keccak256(dkgResult.groupPubKey);
+        bytes32 walletID = wallets.addWallet(
+            dkgResult.membersHash,
+            dkgResult.groupPubKey
+        );
 
-        wallets.addWallet(dkgResult.membersHash, publicKeyHash);
-
-        emit WalletCreated(publicKeyHash, keccak256(abi.encode(dkgResult)));
+        emit WalletCreated(walletID, keccak256(abi.encode(dkgResult)));
 
         // TODO: Disable rewards for misbehavedMembers.
         //slither-disable-next-line redundant-statements
         misbehavedMembers;
 
         walletOwner.__ecdsaWalletCreatedCallback(
-            publicKeyHash,
+            walletID,
             dkgResult.groupPubKey
         );
 
@@ -483,24 +484,36 @@ contract WalletRegistry is IRandomBeaconConsumer, IWalletRegistry, Ownable {
         return dkg.hasDkgTimedOut();
     }
 
-    function getWallet(bytes32 publicKeyHash)
+    function getWallet(bytes32 walletID)
         external
         view
         returns (Wallets.Wallet memory)
     {
-        return wallets.registry[publicKeyHash];
+        return wallets.registry[walletID];
     }
 
-    /// @notice Checks if a wallet with the given public key hash is registered.
-    /// @param publicKeyHash Wallet's public key hash.
-    /// @return True if wallet is registered, false otherwise.
-    function isWalletRegistered(bytes32 publicKeyHash)
+    /// @notice Gets public key of a wallet with a given wallet ID.
+    ///         The public key is returned in an uncompressed format as a 64-byte
+    ///         concatenation of X and Y coordinates.
+    /// @param walletID ID of the wallet.
+    /// @return Uncompressed public key of the wallet.
+    function getWalletPublicKey(bytes32 walletID)
         external
         view
-        returns (bool)
+        returns (bytes memory)
     {
-        return wallets.isWalletRegistered(publicKeyHash);
+        return wallets.registry[walletID].publicKey;
     }
+
+    /// @notice Checks if a wallet with the given ID is registered.
+    /// @param walletID Wallet's ID.
+    /// @return True if wallet is registered, false otherwise.
+    function isWalletRegistered(bytes32 walletID) external view returns (bool) {
+        return wallets.isWalletRegistered(walletID);
+    }
+
+    // TODO: Add function to close the Wallet so the members are notified that
+    // they no longer need to track the wallet.
 
     /// @notice Retrieves dkg parameters that were set in DKG library.
     function dkgParameters()
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -14,6 +14,7 @@
 
 pragma solidity ^0.8.9;
 
+import "./api/IWalletRegistry.sol";
 import "./api/IWalletOwner.sol";
 import "./libraries/EcdsaAuthorization.sol";
 import "./libraries/EcdsaDkg.sol";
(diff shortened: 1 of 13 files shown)


solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -482,25 +484,37 @@ contract WalletRegistry is IRandomBeaconConsumer, Ownable {
...
 
+    /// @notice Checks if a wallet with the given ID is registered.
+    /// @param walletID Wallet's ID.
+    /// @return True if wallet is registered, false otherwise.
+    function isWalletRegistered(bytes32 walletID) external view returns (bool) {
+        return wallets.isWalletRegistered(walletID);
+    }
+
+    // TODO: Add function to close the Wallet so the members are notified that
+    // they no longer need to track the wallet.
```
pdyraga, line 516: In this PR or separately? If separately, please capture it in the issue and add it to [the project](https://github.com/orgs/keep-network/projects/31).

nkuba: https://github.com/keep-network/keep-core/issues/2864

solidity/ecdsa/contracts/api/IWalletRegistry.sol
```diff
@@ -0,0 +1,30 @@
...
+//   ▓▓▓▓▓▓   ▀▓▓▓▓▓▓▄     ▐▓▓▓▓▓▓     ▓▓▓▓▓   ▓▓▓▓▓▓     ▓▓▓▓▓   ▐▓▓▓▓▓▌
+// ▓▓▓▓▓▓▓▓▓▓ █▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+// ▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+//
+//                           Trust math, not hardware.
+
+pragma solidity ^0.8.9;
+
+interface IWalletRegistry {
+    /// @notice Requests a new wallet creation.
```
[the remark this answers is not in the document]
nkuba: 0c94c4a68005d14d181b373bb152656dec2ea1ff

solidity/ecdsa/contracts/libraries/Wallets.sol
```diff
@@ -14,22 +14,19 @@
...
     struct Wallet {
-        // TODO: Verify if we want to store the whole public key or having
-        // just the public key hash is enough.
         // Keccak256 hash of group members identifiers array. Group members do not
         // include operators selected by the sortition pool that misbehaved during DKG.
         bytes32 membersIdsHash;
-        bytes32 digestToSign;
+        // Uncompressed ECDSA public key (64-byte long) as a concatenation of X
+        // and Y coordinates.
+        bytes publicKey;
```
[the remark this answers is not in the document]
nkuba: > We should store the key in a compressed form 

We need both compressed and uncompressed public keys format in the Bridge to validate transactions and signatures. It's easier to convert the key from uncompressed to compressed format than the other way around. As it would require elliptic curve operations.

> use the remaining space for uint64 wallet creation date. This will be needed to manage the wallet lifecycle. Please look at wallet_creation_period and wallet_max_age parameters in [TBTCv2 RFC-1](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#26-wallet-lifecycle).

We probably will want to store it in the Bridge next to the other wallet details as the Bridge will manage the wallet lifecycle as the owner.

> Have you tried to replace it with two bytes32 slots

Nice catch, it reduced cost for about 24k (fe77e91f7f5a5f81a0bc199fdc989810eb08970f)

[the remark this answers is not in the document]
pdyraga: Is the gas consumption in the PR description up-to-date?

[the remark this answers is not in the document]
nkuba: It is now.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -390,18 +391,19 @@ contract WalletRegistry is IRandomBeaconConsumer, Ownable {
...
-        emit WalletCreated(publicKeyHash, keccak256(abi.encode(dkgResult)));
+        emit WalletCreated(walletID, keccak256(abi.encode(dkgResult)));
 
         // TODO: Disable rewards for misbehavedMembers.
         //slither-disable-next-line redundant-statements
         misbehavedMembers;
 
         walletOwner.__ecdsaWalletCreatedCallback(
-            publicKeyHash,
+            walletID,
```
pdyraga, line 406: Would it make sense to return `(walletID bytes32, publicKeyX bytes32, publicKeyY bytes32)` from `wallets.addWallet` and pass `publicKeyX bytes32, publicKeyY bytes32` separately to `__ecdsaWalletCreatedCallback`?

I am looking at https://github.com/keep-network/tbtc-v2/pull/159 and I see we split `bytes` public key into x and y coordinate when compressing the key:
```
        bytes32 x = uncompressedPublicKey.slice32(0);
        bytes32 y = uncompressedPublicKey.slice32(3);
```

If `Bridge` needs separate coordinates, then maybe we can pass them to save gas on slicing?

solidity/ecdsa/test/WalletRegistry.Wallets.test.ts
```diff
@@ -20,6 +23,183 @@ describe("WalletRegistry - Wallets", async () => {
...
+                  test.publicKey
+                )
+              ).to.be.revertedWith(test.expectedError)
+            })
+          })
+        })
+      })
+    })
+
+    context("with wallet registered", async () => {
```
pdyraga, line 146: Non-blocking but worth keeping in mind: we do not need `async` for context fn in the majority of cases.

solidity/ecdsa/test/WalletRegistry.Wallets.test.ts
```diff
@@ -47,8 +227,54 @@ describe("WalletRegistry - Wallets", async () => {
...
+        await expect(
+          walletRegistry.getWalletPublicKey(formatBytes32String("NON EXISTING"))
+        ).to.be.revertedWith("Wallet with given ID has not been registered")
+      })
+    })
+
+    context("with wallet registered", async () => {
+      const walletPublicKey = ecdsaData.group1.publicKey
+
+      // TODO: Add tests to cover zero padding
```
pdyraga, line 248: In this PR or separately?

solidity/ecdsa/contracts/WalletRegistry.sol
nkuba: Yes, we can do that.

solidity/ecdsa/test/WalletRegistry.Wallets.test.ts
nkuba: If it doesn't cause any problems I like that we keep it consistent and don't try to figure out where it should be placed and where it's not needed. IMHO it's more important for `await` with expects where it's not always needed but if we miss it in a place where it should be added then we will end up with a false-positive test result.

pdyraga: I might be missing something, can you give me one example in our codebase of where it is really needed?

solidity/ecdsa/contracts/WalletRegistry.sol
pdyraga: ... and I would leave `getWalletPublicKey` as-is because that's what we need for [`CheckBitcoinSigs` in `bitcoin-spv`](https://github.com/keep-network/bitcoin-spv/blob/solidity/v3.3.0-solc-0.8/solidity/contracts/CheckBitcoinSigs.sol#L20) that will be used by TBTC `Bridge` for fraud proofs.

solidity/ecdsa/test/WalletRegistry.Wallets.test.ts
nkuba: Do you mean an example of `await` with `expect`? Here we have it, but if it wasn't provided we would not catch a test failure: https://github.com/keep-network/keep-core/blob/b6768d9bd326f4fe67bb189fafd27b5aff641c0c/solidity/ecdsa/test/WalletRegistry.Wallets.test.ts#L173-L177

I've seen numerous of times situations where we were having `expect` without an `await` where it was really necessary. This is why I prefer to add `await` to all the `expect`s, even if it's redundant.

solidity/ecdsa/test/WalletRegistry.Wallets.test.ts
nkuba: bfb8b9b7ad89e3dd0aa61af4837f7f545c75be98

pdyraga (approved):

PR #2860, opened 2022-03-08
Reimbursements in Random Beacon from the ETH pool

dimpar:
Part of [RFC-6](https://github.com/keep-network/tbtc-v2/blob/7b3cb19f6860cfc32677b15d5de643a9ee0f0f51/docs/rfc/rfc-6.adoc)
Issue: https://github.com/keep-network/keep-core/issues/2824

This PR should've been split into smaller pieces, but this was clear only after the fact. We have 3 main goals here:

1. Refunding ETH to notifiers and dkg submitters in RandomBeacon from the ETH reimbursement pool. This was implemented in a similar fashion to what we have in the WalletRegisty contract. All of the refunds are calculated on the fly and the gas offset is added to specific functions. The offset is added because only what's inside of the function can be calculated on the fly and it's not taking into account things like function params which can have a heavy load too. Gas offsets were chosen by running tests and fine-tuned with a try/error approach to make the refund as precise as possible. Extra attention requires `submitDkgResult`+`approveDkgResult` cause it's a 2 step refund. The avg. gas of `submitDkgResult` is hardcoded to the contract and then added in the `approvedDkgResult` call. Only the approver gets the full refund.
All of the refunds are slightly net-positive for the callers, but <0.002ETH.

2. Removing some of the rewards functionality in favor of refunding ETH for transactions from the Reimbursable Pool described in 1)

3. Authorizing and unathorizing contracts/callers that can request a new relay entry.


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -528,6 +561,57 @@ contract RandomBeacon is IRandomBeacon, Ownable {
         );
     }
 
+    /// @notice Updates the DKG result submission gas.
+    /// @dev Can be called only by the contract owner, which should be the
+    ///      wallet registry governance contract. The caller is responsible for
```
pdyraga, line 566: Not the wallet registry governance ser.

```suggestion
    ///      random beacon governance contract. The caller is responsible for
```

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -151,27 +152,42 @@ contract RandomBeacon is IRandomBeacon, Ownable {
     ///         operator affected.
     uint256 public dkgMaliciousResultNotificationRewardMultiplier;
 
+    /// @notice Calculated gas cost for submitting a DKG result. This will
+    ///         be refunded as part of the DKG approval process. It is in the
+    ///         submitter's interest to not skip his priority turn on the approval,
+    ///         otherwise the refund of the DKG submission will be refunded to
+    ///         other member that will call the DKG approve function.
```
pdyraga, line 159: Let's be a bit more specific about what member we are talking about. If that's also the case for `WalletRegistry.sol`, let's update both in this PR:
```suggestion
    ///         another group member that will call the DKG approve function.
```

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -151,27 +152,42 @@ contract RandomBeacon is IRandomBeacon, Ownable {
...
+    uint256 public dkgResultSubmissionGas = 235000;
+
+    // @notice Gas is meant to balance the DKG approval's overall cost. Can be
```
pdyraga, line 162:
```suggestion
    // @notice Gas meant to balance the DKG approval's overall cost. Can be
```

Same for other offsets below.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -151,27 +152,42 @@ contract RandomBeacon is IRandomBeacon, Ownable {
...
     // Other parameters
 
     /// @notice Stores current operator inactivity claim nonce for given group.
     ///         Each claim is made with an unique nonce which protects
     ///         against claim replay.
     mapping(uint64 => uint256) public inactivityClaimNonce; // groupId -> nonce
 
+    /// @notice Authorized contracts that can request a relay entry without
+    ///         sending any fees.
+    mapping(address => bool) public authorizedContracts;
```
pdyraga, line 183: So maybe we should rename it to `authorizedRequesters`? This will be more explicit about what they are authorized for.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -307,6 +323,10 @@ contract RandomBeacon is IRandomBeacon, Ownable {
         address[] groupMembers
     );
 
+    event AuthorizedContract(address contractToAuthorize);
+
+    event UnauthorizedContract(address contractToUnauthorize);
```
pdyraga, line 328: Looking at other events the convention is that we are putting the verb at the end, so they should be `ContractAuthorized` and `ContractUnauthorized`. If we decide to rename `authorizedContracts` mapping to `authorizedRequesters`, then I would rename them to `RequesterAuthorized` and `RequesterDeauthorized`.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -528,6 +561,57 @@ contract RandomBeacon is IRandomBeacon, Ownable {
...
+        emit AuthorizedContract(_contract);
+    }
+
+    /// @notice Unauthorize contract which can request a relay entry without
+    ///         paying fees. Unauthorization can be done by the owner only.
+    function unauthorizeContract(address _contract) external onlyOwner {
+        delete authorizedContracts[_contract];
+
+        emit UnauthorizedContract(_contract);
+    }
```
pdyraga, line 613: The current approach is *clearer* but if we need to save some bytecode, I would consider doing a similar trick as we did in tBTC v2 contract where we have [a function with a boolean parameter flipping the `isVaultTrusted` boolean flag](https://github.com/keep-network/tbtc-v2/blob/22c72cf5ede18da64ead99741c6cf8b544d4715a/solidity/contracts/bridge/Bridge.sol#L500-L503). This would be something like `setRequesterStatus(address requester, boolean isTrusted)`.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -942,8 +1055,9 @@ contract RandomBeacon is IRandomBeacon, Ownable {
     ///         of group members, operators of members deemed as inactive are
     ///         banned for sortition pool rewards for duration specified by
     ///         `sortitionPoolRewardsBanDuration` parameter. The sender of
-    ///         the claim must be one of the claim signers. This function
-    ///         can be called only for active and non-terminated groups.
+    ///         the claim must be one of the claim signers. The sender is
+    ///         refunded from the Reimbursement Pool. This function can be
+    ///         called only for active and non-terminated groups.
```
pdyraga, line 1060: Should we add a similar sentence to the documentation of other functions supporting refunds? Alternatively, do _not_ add it here.

pdyraga (commented):
First round! @dimpar please merge the recent `main` to this PR and then address the comments. Once it's done, I'll do a second pass. In case there are still problems with `RandomBeacon` contract size, I will try to help.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -353,7 +383,10 @@ contract RandomBeacon is IRandomBeacon, Ownable {
         dkg.setSubmitterPrecedencePeriodLength(20); // TODO: Verify value
 
         relay.initSeedEntry();
-        relay.setRelayRequestFee(200e18);
+        // total gas of submit+approve ~780849. Assume gas price is 200gwei
+        // total dkg gas cost: 0.0000002 * 780849 = 0.1561698
+        // there need to be 5 requests to trigger dkg, so 0.156 / 5 = ~0.031 ETH need to be provided to break even.
+        relay.setRelayRequestFee(31e15); // TODO: revisit. Do we want to add a markup?
```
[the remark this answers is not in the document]
dimpar: > The parameters we took into account when calculating the relay request fee may change so the fee should be alterable

The relay request fee is changeable and can be updated by the governance in the current implementation. 

> On the other hand, we have problems with the contract size and adding one more public function makes it only worse. I would vote for eliminating the fee completely and just reverting in requestRelayEntry if !authorizedContracts[msg.sender]

Sure, I thought it's a must-have functionality which is described in the issue [here](https://github.com/keep-network/keep-core/issues/2824). But if we're okay removing it, then yes, it reduces the size.
We can certainly require only `authorizedContracts[msg.sender]`, but let's keep in mind that having a separate contract that will charge a fee (if needed) will increase the overall cost for the relay entry.

Request fee is removed here [f11d300](https://github.com/keep-network/keep-core/pull/2860/commits/f11d300c16a4c67a2530d827ba9549a8e908f338)

Tests will be fixed in a different commit, cause now they won't run because of the contract size limit errro

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -151,27 +152,42 @@ contract RandomBeacon is IRandomBeacon, Ownable {
     ///         operator affected.
     uint256 public dkgMaliciousResultNotificationRewardMultiplier;
```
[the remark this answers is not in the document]
dimpar: Yea, good one [b3e2053](https://github.com/keep-network/keep-core/pull/2860/commits/b3e20536d1474d12173162d3f4cd795ee88c0ab6)

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -528,6 +561,57 @@ contract RandomBeacon is IRandomBeacon, Ownable {
...
+    ///      wallet registry governance contract. The caller is responsible for
+    ///      validating parameters.
+    /// @param _dkgResultSubmissionGas DKG result submission gas
+    function updateDkgResultSubmissionGas(uint256 _dkgResultSubmissionGas)
+        external
+        onlyOwner
+    {
+        dkgResultSubmissionGas = _dkgResultSubmissionGas;
+        emit DkgResultSubmissionGasUpdated(_dkgResultSubmissionGas);
+    }
```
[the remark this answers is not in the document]
dimpar: Sure [79994e1](https://github.com/keep-network/keep-core/pull/2860/commits/79994e16930b4422f42e88e0a53512af55566ddd)

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: You sure? I think the [correct form](https://www.collinsdictionary.com/dictionary/english/be-meant-to#:~:text=phrase,to%20announce%20it%20to%20everyone.) is "be meant to"

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: Sure, done in [9984f7a](https://github.com/keep-network/keep-core/pull/2860/commits/9984f7abdedd261570a5c6958299dc51c9807ffd)

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: I removed the sentence, cause other docs are smaller, and mentioning just about reimbursement pool raises a question about why we mention that and not other stuff as well.

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: It's gone now, replaced by `RequesterAuthorizationUpdated` event

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: Done [dd9ad55](https://github.com/keep-network/keep-core/pull/2860/commits/dd9ad55739999e3d5bdc88bdc68b84b1c20367f9)

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: [dd9ad55](https://github.com/keep-network/keep-core/pull/2860/commits/dd9ad55739999e3d5bdc88bdc68b84b1c20367f9)

solidity/random-beacon/contracts/RandomBeacon.sol
dimpar: [dd9ad55](https://github.com/keep-network/keep-core/pull/2860/commits/dd9ad55739999e3d5bdc88bdc68b84b1c20367f9)

solidity/ecdsa/test/utils/inactivity.ts
```diff
@@ -16,7 +16,7 @@ export async function signOperatorInactivityClaim(
 }> {
   const messageHash = ethers.utils.solidityKeccak256(
     ["uint256", "bytes", "uint8[]", "bool"],
-    [nonce, groupPubKey, inactiveMembersIndices, failedHeartbeat]
+    [nonce, groupPubKey, inactiveMembersIndices, operatorInactivity]
```
pdyraga, line 19: Why are we changing it? The last component is a boolean flag whether heartbeat failed:
https://github.com/keep-network/keep-core/blob/8dc2df00236d2a09f038187c83fd8441aa731d63/solidity/ecdsa/contracts/libraries/EcdsaInactivity.sol#L40-L50

solidity/random-beacon/contracts/api/IRandomBeacon.sol
```diff
@@ -21,5 +21,7 @@ interface IRandomBeacon {
     /// @notice Creates a request to generate a new relay entry. Requires a
     ///         request fee denominated in T token.
     /// @param callbackContract Beacon consumer callback contract.
-    function requestRelayEntry(IRandomBeaconConsumer callbackContract) external;
+    function requestRelayEntry(IRandomBeaconConsumer callbackContract)
+        external
+        payable;
```
pdyraga, line 26: No longer `payable`.

dimpar: [a371249](https://github.com/keep-network/keep-core/pull/2860/commits/a371249e124beb881c6deeb436cbc866f8650939)

solidity/ecdsa/test/utils/inactivity.ts
dimpar: [6ae3265](https://github.com/keep-network/keep-core/pull/2860/commits/6ae32656cd66c38c0d03beb3525bfe89e9fe5a03)

solidity/random-beacon/contracts/RandomBeacon.sol
pdyraga: Right but what I am trying to say is that we are describing this particular field. So maybe just: *Gas that is meant to...*"?

solidity/random-beacon/contracts/RandomBeaconGovernance.sol
```diff
@@ -1057,6 +884,157 @@ contract RandomBeaconGovernance is Ownable {
         newRelayEntrySubmissionFailureSlashingAmount = 0;
     }
 
+    /// @notice Begins the dkg result submission gas update process.
```
dimpar: dkg -> DKG

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -217,18 +210,6 @@ describe("RandomBeacon - Group Creation", () => {
                 "Not awaiting genesis"
               )
             })
-
-            it("should refund ETH", async () => {
```
dimpar: This test validates the refunded ETH after `.approveDkgResult(dkgResult)` call done in `before(..)`. In every `approveDkgResult` call we do refund ETH to the caller. `submitterInitialBalance` returned from `await signAndSubmitCorrectDkgResult()` is needed to calculate the difference between the initial caller's balance and what it has after the refund.

It is true that we have also tests in `approveDkgResult` but each refund scenario returns a slightly different result. This is how the offset was chosen by running different tests and picking a number that works for many scenarios.

Same comment for the other "should refund ETH" tests that were removed in this commit.

dimpar:
Some ideas that can save ~0.1KB is using a private function instead of an `onlyStakingContract` modifier and placing that private function inside the code that used a modifier before. The other one is using a shorter error message in the `require`s.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -614,7 +617,7 @@ contract WalletRegistry is
 
         dkg.complete();
 
-        // Refunds msg.sender's ETH for dkg result submission & dkg approval
+        // Refund msg.sender's ETH for dkg result submission & dkg approval
```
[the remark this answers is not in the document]
pdyraga: 1ba3e7536999e8fd9dd27380a44c20ad0c6a5583

solidity/random-beacon/test/RandomBeaconGovernance.test.ts
```diff
@@ -3398,6 +2846,51 @@ describe("RandomBeaconGovernance", () => {
...
+        await expect(
+          randomBeaconGovernance
+            .connect(thirdParty)
+            .setRequesterAuthorization(thirdPartyContract.address, true)
+        ).to.be.revertedWith("Ownable: caller is not the owner")
+
+        await expect(
+          randomBeaconGovernance
+            .connect(thirdParty)
+            .setRequesterAuthorization(thirdPartyContract.address, false)
```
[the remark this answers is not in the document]
pdyraga: I wanted to make sure there is no path allowing to deauthorize a contract if someone is not the governance. That is, no matter the value of the second parameter, the call will fail for a non-governance caller. But if you think it's too much I can remove the second assertion.

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
pdyraga: Hmm... I would then improve test cases for `approveDkgResult` in case we need to cover more scenarios. If we place assertions for refunding ETH for functions that are not supposed to refund ETH, this is getting confusing.

Also, please see https://github.com/keep-network/keep-core/pull/2860/commits/701e0686698c306cd38929dc71f43ad384aefbbe.

Two tests are still failing (one added in the referenced commit), I think we may either need to adjust the offset or give up on the precision.

pdyraga:
@dimpar OK, I am done here, no more comments/change requests from my end other than these two failing unit tests that may require an update to the gas offset (cc @nkuba) or to give up on precision. If you would like to change anything in my commits, please go ahead and push changes. I'll do a final pass on your signal.

pdyraga (approved):

PR #2862, opened 2022-03-09
ECDSA: Use the same governance delay for all parameters

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/2847

2-week critical governance parameter does not make sense for ECDSA `WalletRegistry`
given that wallets will have to remain active for months. We could swallow it for
the `RandomBeacon` but for `WalletRegistry` 2 weeks does not make any sense and 2
months is not enough.

One option would be to eliminate governance contracts completely and just deploy
`TimeLock` contract. On the other hand, `TimeLock` updates are more complex to set
up than simple calls to the existing governance contract, and the structure of
update* functions in `WalletRegistry` is more complicated.

Here we introduce just one governable delay for parameter updates and eliminate
the separation between critical/standard parameters given it creates a false
assumption of safety. There is also a function allowing to update the governance
delay.

### DIFF
--- a/solidity/ecdsa/contracts/WalletRegistryGovernance.sol
+++ b/solidity/ecdsa/contracts/WalletRegistryGovernance.sol
@@ -24,6 +24,9 @@ import {IRandomBeacon} from "@keep-network/random-beacon/contracts/api/IRandomBe
 /// @notice Owns the `WalletRegistry` contract and is responsible for updating
 ///         its governable parameters in respect to the governance delay.
 contract WalletRegistryGovernance is Ownable {
+    uint256 public newGovernanceDelay;
+    uint256 public governanceDelayChangeInitiated;
+
     address public newWalletOwner;
     uint256 public walletOwnerChangeInitiated;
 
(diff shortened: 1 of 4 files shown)


dimpar:
The implementation looks good, but just want to confirm one thing. Say we have:
- current delay: 1 week
- initiate the delay upgrade to 4 weeks
- the next day a gov. calls to update some gov param, ex. authorization
- after 1 week passes, the delay is updated to 4 weeks
- but now, the authorization can be updated after 3 weeks and 1 day

DAO should be aware that if anything is updated within the `delay` upgrade process, then the timeframe will be different for another gov. param which was initiated during the `delay` upgrade (for both increase and decrease the delay).

nkuba (approved):

pdyraga:
> 

Correct. I think it makes to leave it this way to do not add complexity. It's not that big deal in my opinion. The governance may decide to delay the approval anyway. This contract is enforcing no change is applied sooner than the delay but it's not enforcing that it is applied as soon as the delay ends.

PR #2863, opened 2022-03-09
Let to transfer WalletRegistry ownership to someone else

pdyraga:
~~Depends on https://github.com/keep-network/keep-core/pull/2862; Keeping as a draft until #2862 is merged.~~

`WalletRegistryGovernance` contract controls governable parameters of
`WalletRegistry` contract. If it ever happens that the ownership of
`WalletRegistry` must be transferred to someone else, functionality to
transfer the ownership is now available.

### DIFF
--- a/solidity/ecdsa/contracts/WalletRegistryGovernance.sol
+++ b/solidity/ecdsa/contracts/WalletRegistryGovernance.sol
@@ -27,6 +27,9 @@ contract WalletRegistryGovernance is Ownable {
     uint256 public newGovernanceDelay;
     uint256 public governanceDelayChangeInitiated;
 
+    address public newWalletRegistryOwner;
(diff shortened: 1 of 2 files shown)


nkuba (dismissed):

dimpar:
It might happen that if some gov param will be initiated for an update during owner transfer, then two separate owners will be involved. The first one is the one that initiated it, and the second one will finalize it. If that's ok, then the code lgtm.

pdyraga:
> It might happen that if some gov param will be initiated for an update during owner transfer, then two separate owners will be involved. The first one is the one that initiated it, and the second one will finalize it. If that's ok, then the code lgtm.

Correct. This contract is enforcing no change is applied sooner than the delay but it's not enforcing that it is applied as soon as the delay ends.

dimpar (approved):

PR #2865, opened 2022-03-11
Beacon: Use the same governance delay for all parameters

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/2847
See https://github.com/keep-network/keep-core/pull/2862

2-week critical governance parameter dit not make sense for ECDSA `WalletRegistry`
given that wallets have to remain active for months. We could swallow it for
the `RandomBeacon` but it is handcuffing the governance - to make it work,
no group should live longer than 2 weeks. Given that ECDSA governance has only
one delay, beacon governance is updated for consistency.

Another option was to eliminate governance contracts completely and just deploy
`TimeLock` contract. However, `TimeLock` updates are more complex to set up than
simple calls to the existing governance contract, and the structure of `update*`
functions in `RandomBeacon` is more complicated.

Further work, outside of this PR to make the diff size sane:
- allow to update the governance delay
- allow transferring RandomBeacon ownership somewhere else


dimpar (approved):

dimpar:
lgtm and I can merge unless @nkuba wants to take a look?

Issue #2839, opened 2022-02-22
ECDSA wallet heartbeats


pdyraga:
Just like in the case of the Random Beacon, ECDSA wallet members should be able to sign a message notifying some operators failed a heartbeat. One of the reasons could be, for example, missing their round to submit a result to the chain.

Additionally, there should be a way to submit a failed of weak heartbeat info and notify tBTC Bridge about it so that it can request the wallet to start moving funds somewhere else.

See https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#heartbeat

PR #2868, opened 2022-03-14
Mark operators misbehaving during ECDSA DKG as ineligible for rewards

### LINKED ISSUE #2857: Disable rewards for misbehaved members in ECDSA wallet
When approving `dkg.approveResult(dkgResult)` in `approveDkgResult()` we need to disable rewards for the misbehaved members.

pdyraga:
Closes #2857

Group members who were misbehaving during DKG (inactive or disqualified) are marked as ineligible for rewards for a certain, governable, period of time.

Added governable parameter to `WalletRegistry` indicating that time. The new governable parameter will also be used for reward ineligibility period for operators who missed their turn during DKG result submission or that failed a heartbeat (this is covered in https://github.com/keep-network/keep-core/issues/2839).

### DIFF
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -163,10 +163,12 @@ contract WalletRegistry is IRandomBeaconConsumer, IWalletRegistry, Ownable {
 
         // TODO: Implement governance for the parameters
         // TODO: revisit all initial values
+        sortitionPoolRewardsBanDuration = 2 weeks;
 
         // slither-disable-next-line too-many-digits
         authorization.setMinimumAuthorization(400000e18); // 400k T
         authorization.setAuthorizationDecreaseDelay(5184000); // 60 days
+        
         maliciousDkgResultSlashingAmount = 50000e18;
         maliciousDkgResultNotificationRewardMultiplier = 100;
 
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -72,6 +72,11 @@ contract WalletRegistry is IRandomBeaconConsumer, IWalletRegistry, Ownable {
     ///         operator affected.
     uint256 public maliciousDkgResultNotificationRewardMultiplier;
 
+    /// @notice Duration of the sortition pool rewards ban imposed on operators
+    ///         who missed their turn for DKG result submission or who failed
+    ///         a heartbeat.
+    uint256 public sortitionPoolRewardsBanDuration;
+
     // External dependencies
 
     SortitionPool public immutable sortitionPool;
@@ -129,7 +134,8 @@ contract WalletRegistry is IRandomBeaconConsumer, IWalletRegistry, Ownable {
     );
 
     event RewardParametersUpdated(
-        uint256 maliciousDkgResultNotificationRewardMultiplier
+        uint256 maliciousDkgResultNotificationRewardMultiplier,
+        uint256 sortitionPoolRewardsBanDuration
     );
 
     event SlashingParametersUpdated(uint256 maliciousDkgResultSlashingAmount);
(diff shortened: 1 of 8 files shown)


solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -72,6 +72,11 @@ contract WalletRegistry is IRandomBeaconConsumer, IWalletRegistry, Ownable {
     ///         operator affected.
     uint256 public maliciousDkgResultNotificationRewardMultiplier;
 
+    /// @notice Duration of the sortition pool rewards ban imposed on operators
+    ///         who missed their turn for DKG result submission or who failed
+    ///         a heartbeat.
+    uint256 public sortitionPoolRewardsBanDuration;
```
dimpar, line 78: Shouldn't we set the initial value here or in the constructor?

solidity/ecdsa/test/WalletRegistry.WalletCreation.test.ts
```diff
@@ -1869,14 +1874,14 @@ describe("WalletRegistry - Wallet Creation", async () => {
...
+            it("should ban misbehaved operators from sortition pool rewards", async () => {
+              const now = await helpers.time.lastBlockTime()
+              const expectedUntil = now + params.sortitionPoolRewardsBanDuration
 
-            //   await expect(tx)
-            //     .to.emit(sortitionPool, "IneligibleForRewards")
-            //     .withArgs(misbehavedIds, expectedUntil)
-            // })
+              await expect(tx)
+                .to.emit(sortitionPool, "IneligibleForRewards")
```
dimpar, line 1882: What do you think about adding a test when there are no misbehaved members and the event was not emitted?

solidity/ecdsa/contracts/WalletRegistry.sol
pdyraga: Yeah, we should. Fixed in https://github.com/keep-network/keep-core/pull/2868/commits/2642b128b76751959c527cf17ae5ba75fcefa70e.

solidity/ecdsa/test/WalletRegistry.WalletCreation.test.ts
pdyraga: Added in https://github.com/keep-network/keep-core/pull/2868/commits/6f96d7d22d08da1df35d67ff4a415b2f3c2380be. I attached to the existing `context` to do not introduce one more nesting level.

solidity/ecdsa/test/fixtures/index.ts
```diff
@@ -40,6 +40,7 @@ export const params = {
   dkgResultChallengePeriodLength: 10,
   dkgResultSubmissionTimeout: 30,
   dkgSubmitterPrecedencePeriodLength: 5,
+  sortitionPoolRewardsBanDuration: 604800, // 7 days
```
dimpar, line 43: I take that this value is for testing purposes only and doen't have to match what we have in the WalletRegistry contract 14 days?

solidity/ecdsa/test/utils/dkg.ts
```diff
@@ -56,6 +56,7 @@ export async function signAndSubmitCorrectDkgResult(
   signers: Operator[]
   dkgResult: DkgResult
   dkgResultHash: string
+  members: number[]
```
dimpar, line 59: `dkgResult` already has `members` field and we can use that in tests.

dimpar:
no more comments

solidity/ecdsa/test/utils/dkg.ts
pdyraga: Yeah, very good point. I reverted this change in https://github.com/keep-network/keep-core/pull/2868/commits/8d88718b92b1c98cbe26feccfc2cd8a107138e42.

solidity/ecdsa/test/fixtures/index.ts
pdyraga: I do not think we have finalized this value. But yeah, it's probably better to test with the same value that is used in the constructor. See https://github.com/keep-network/keep-core/pull/2868/commits/f87b1f440905744c17066a32aec8845c350f88f2, please.

solidity/ecdsa/test/WalletRegistry.WalletCreation.test.ts
```diff
@@ -1835,7 +1845,8 @@ describe("WalletRegistry - Wallet Creation", async () => {
 
               await mineBlocksTo(startBlock + dkgTimeout - 1)
 
-              let submitter
+              let members: number[]
```
dimpar, line 1848: This guy is not used anymore. And also the 'misbehavedIds' has not type, we can declare it as number[]

pdyraga: See https://github.com/keep-network/keep-core/pull/2868/commits/c1d223c805be6558e25c7ae997187f54c409cb34 ser.

dimpar (approved):

PR #2872, opened 2022-03-17
PlainTransferEscrowBeneficiary for PhasedEscrow and BatchedPhasedEscrow

pdyraga:
`PlainTransferEscrowBeneficiary` is a beneficiary contract that can
receive a withdrawal order from `PhasedEscrow` and `BatchedPhasedEscrow`
contracts. It immediately sends the received tokens to the recipient
set by the beneficiary deployer.

The recipient does not have to implement any custom functions, it can
be a smart contract or EOA - beneficiary is executing a plain ERC20
transfer.

### DIFF
--- a/solidity-v1/contracts/PhasedEscrow.sol
+++ b/solidity-v1/contracts/PhasedEscrow.sol
@@ -183,6 +183,30 @@ interface IStakingPoolRewards {
     function notifyRewardAmount(uint256 amount) external;
 }
 
+/// @title PlainTransferEscrowBeneficiary
+/// @notice  PlainTransferEscrowBeneficiary is a beneficiary contract that can
+///          receive a withdrawal order from PhasedEscrow and BatchedPhasedEscrow
+///          contracts. It immediately sends the received tokens to the recipient
+///          set by the beneficiary deployer.
+///          The recipient does not have to implement any custom functions, it can
+///          be a smart contract or EOA - beneficiary is executing a plain ERC20
+///          transfer.
+contract PlainTransferEscrowBeneficiary is Ownable, IBeneficiaryContract {
+    using SafeERC20 for IERC20;
+
+    IERC20 public token;
+    address public recipient;
+
+    constructor(IERC20 _token, address _recipient) public {
+        token = _token;
+        recipient = _recipient;
+    }
+
+    function __escrowSentTokens(uint256 amount) external onlyOwner {
+        token.safeTransfer(recipient, amount);
+    }
+}
+
 /// @title StakingPoolRewardsEscrowBeneficiary
 /// @notice A beneficiary contract that can receive a withdrawal phase from a
 ///         PhasedEscrow contract. Immediately stakes the received tokens on a
--- a/solidity-v1/test/TestPhasedEscrow.js
+++ b/solidity-v1/test/TestPhasedEscrow.js
@@ -24,11 +24,15 @@ const BeaconBackportRewardsEscrowBeneficiary = contract.fromArtifact(
 const BeaconRewardsEscrowBeneficiary = contract.fromArtifact(
   "BeaconRewardsEscrowBeneficiary"
 )
+const PlainTransferEscrowBeneficiary = contract.fromArtifact(
+  "PlainTransferEscrowBeneficiary"
+)
 
 const BeaconRewards = contract.fromArtifact("BeaconRewards")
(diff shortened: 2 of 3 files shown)


solidity-v1/test/TestPhasedEscrow.js
```diff
@@ -388,6 +392,33 @@ describe("PhasedEscrow", () => {
...
+  describe("when withdrawing to a PlainTransferEscrowBeneficiary", () => {
+    const baseBalance = 200000000
+    const transferAmount = 9100000
+
+    before(async () => {
+      phasedEscrow = await PhasedEscrow.new(token.address, { from: owner })
+      const amount = web3.utils.toBN(baseBalance)
+      await token.transfer(phasedEscrow.address, amount, { from: owner })
+
+      rewardsContract = await TestSimpleReceiver.new({ from: owner })
```
nkuba, line 404:
```suggestion
      recipientContract = await TestSimpleReceiver.new({ from: owner })
```

solidity-v1/test/TestPhasedEscrow.js
```diff
@@ -388,6 +392,33 @@ describe("PhasedEscrow", () => {
...
+      rewardsBeneficiary = await PlainTransferEscrowBeneficiary.new(
```
nkuba, line 405:
```suggestion
      escrowBeneficiary = await PlainTransferEscrowBeneficiary.new(
```

solidity-v1/test/TestPhasedEscrow.js
```diff
@@ -810,6 +841,73 @@ describe("BatchedPhasedEscrow", () => {
       )
     })
   })
+
+  describe("when withdrawing to a PlainTransferEscrowBeneficiary", () => {
+    const baseBalance = 200000000
+    const transferAmount = 9100000
+
+    let rewardsContract
+    let rewardsBeneficiary
```
nkuba, line 850:
```suggestion
    let recipientContract
    let escrowBeneficiary
```

solidity-v1/contracts/PhasedEscrow.sol
```diff
@@ -183,6 +183,30 @@ interface IStakingPoolRewards {
...
 
+/// @title PlainTransferEscrowBeneficiary
+/// @notice  PlainTransferEscrowBeneficiary is a beneficiary contract that can
+///          receive a withdrawal order from PhasedEscrow and BatchedPhasedEscrow
+///          contracts. It immediately sends the received tokens to the recipient
+///          set by the beneficiary deployer.
+///          The recipient does not have to implement any custom functions, it can
+///          be a smart contract or EOA - beneficiary is executing a plain ERC20
+///          transfer.
+contract PlainTransferEscrowBeneficiary is Ownable, IBeneficiaryContract {
```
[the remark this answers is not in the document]
pdyraga: There is no point in changing that because it's probably the last time we deploy a beneficiary. And the last time we'll use any beneficiary will be in April.

Anyway, I think keeping them together makes sense as they are very closely related.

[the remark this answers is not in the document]
pdyraga: I think `PlainTransfer*` reflects what is the purpose of this beneficiary. All beneficiaries are forwarding tokens but only this one is doing it via plain ERC20 transfer, without calling any contract functions on the other end.

solidity-v1/test/TestPhasedEscrow.js
pdyraga: It is not going to work because of how `assertRewards` is constructed:

https://github.com/keep-network/keep-core/blob/c5f3acf0658f341fe8656219693b761001485033/solidity-v1/test/TestPhasedEscrow.js#L435

solidity-v1/test/TestPhasedEscrow.js
pdyraga: It's not going to work because of how `assertRewards` is constructed:
https://github.com/keep-network/keep-core/blob/c5f3acf0658f341fe8656219693b761001485033/solidity-v1/test/TestPhasedEscrow.js#L456

solidity-v1/test/TestPhasedEscrow.js
pdyraga: Here I can do the rename but I kept it aligned with `assertRewards` earlier. WDYT?

nkuba (approved):

PR #2876, opened 2022-03-21
Callback for wallet's hearbeat failure

lukasz-zimnoch:
Refs: #2839
Refs: https://github.com/keep-network/tbtc-v2/issues/139

Here we expose `__ecdsaWalletHeartbeatFailedCallback` function in the `IWalletOwner` interface that will be used to notify the wallet owner about a heartbeat failure detected on a specific wallet.

### DIFF
--- a/solidity/ecdsa/contracts/api/IWalletOwner.sol
+++ b/solidity/ecdsa/contracts/api/IWalletOwner.sol
@@ -25,4 +25,16 @@ interface IWalletOwner {
         bytes32 publicKeyX,
         bytes32 publicKeyY
     ) external;
+
+    /// @notice Callback function executed once a wallet heartbeat failure
+    ///         is detected.
+    /// @dev Should be callable only by the Wallet Registry.
+    /// @param walletID Wallet's unique identifier.
+    /// @param publicKeyY Wallet's public key's X coordinate.
+    /// @param publicKeyY Wallet's public key's Y coordinate.
+    function __ecdsaWalletHeartbeatFailedCallback(
+        bytes32 walletID,
+        bytes32 publicKeyX,
+        bytes32 publicKeyY
+    ) external;
 }


solidity/ecdsa/contracts/api/IWalletOwner.sol
```diff
@@ -25,4 +25,16 @@ interface IWalletOwner {
...
+    ///         is detected.
+    /// @dev Should be callable only by the Wallet Registry.
+    /// @param walletID Wallet's unique identifier.
+    /// @param publicKeyY Wallet's public key's X coordinate.
+    /// @param publicKeyY Wallet's public key's Y coordinate.
+    function __ecdsaWalletHeartbeatFailedCallback(
+        bytes32 walletID,
+        bytes32 publicKeyX,
+        bytes32 publicKeyY
+    ) external;
```
nkuba, line 39: Isn't `walletID` enough? I'd expect we call this function once a wallet is already registered and the Owner was notified about it by `__ecdsaWalletCreatedCallback`.

lukasz-zimnoch: Nope, it isn't. We index wallets by 20-byte public key hash in the Bridge. If we pass just the `walletID`, we can't fetch the wallet from storage.

nkuba: Would it be beneficial if we store walletID -> 20-byte hash in a Bridge on wallet creation and use it for a heartbeat failure? Would such mapping be used anywhere else besides this scenario?

lukasz-zimnoch: I don't think so since it will increase the gas cost of wallet creation significantly by `22284` and add costs upon reading. For now, reading from such a mapping would be done just in this scenario but there is no guarantee we won't use it somewhere else. However, wallet creation will occur much more often than heartbeat failures so I think we should optimize the former.

On the other hand, what's the exact problem with passing the uncompressed public key? If this is about reading too much on the ECDSA registry side, maybe we can get rid of passing the `walletID` instead? Actually, the Bridge doesn't use it apart from putting it to wallet-related events. I understand we want to have a kind of cross-contract wallet ID but looks like the uncompressed PK fits that role in a better way as it can be easily handled by both Bridge and ECDSA registry.

pdyraga: I would think more about whose responsibility is to handle heartbeats. Is it `WalletRegistry` or the `Bridge`? If it is the responsibility of `WalletRegistry` then I think we have no other choice than to leave this signature as-is. If it is the responsibility of the `Bridge`, then `WalletRegistry` needs to expose just one function to close a wallet and release off-chain operators from keeping the key material.

I think we should also improve the naming. In `RandomBeacon` we have `notifyFailedHeartbeat` that lets to mark operators as ineligible for rewards without closing a group. I think we may need something similar for ECDSA in case someone is consistently not participating in heartbeats but we are still above the safety threshold. To clearly distinguish these two situations, I would rename `RandomBeacon.notifyFailedHeartbeat` to `RandomBeacon.notifyInactiveOperators` and use *heartbeat failure* name only for wallet-wide events.

pdyraga: Thinking more about it, I think the version proposed by @lukasz-zimnoch makes sense. If `WalletRegistry` is responsible for identifying inactive operators, it should also be responsible for wallet-wide heartbeat failures. `Bridge` should be just about bridging and wallet internals should be handled elsewhere.

I kinda like the parameters of `__ecdsaWalletCreatedCallback` are 1:1 with the parameters of `__ecdsaWalletHeartbeatFailedCallback` and in this model, (x,y) becomes just a secondary key.

@lukasz-zimnoch I would still like to rename `RandomBeacon.notifyFailedHeartbeat` to `notifyInactiveOperators` and introduce `notifyInactiveOperators` in `WalletRegistry` as well.

pdyraga (approved):
Holding with merge to give you a few hours to chime in @nkuba

solidity/ecdsa/contracts/api/IWalletOwner.sol
nkuba: > @lukasz-zimnoch I would still like to rename RandomBeacon.notifyFailedHeartbeat to notifyInactiveOperators and introduce notifyInactiveOperators in WalletRegistry as well.

Shouldn't we rename `__ecdsaWalletHeartbeatFailedCallback` to `__ecdsaWalletInactiveOperatorsCallback` then?

lukasz-zimnoch: > and use heartbeat failure name only for wallet-wide events.

If I understood Piotr correctly, we shouldn't because that's a wallet-related event.

pdyraga: Yes, this is wallet-related callback. Bridge does not need to know about operators marked as inactive due to inactivity (e.g. they missed their submission round or were inactive during the heartbeat for weeks but the heartbeat is still above the critical threshold)

PR #2878, opened 2022-03-23
Reverting random-beacon dependency to >2.0.0-dev <2.0.0-ropsten

dimpar:
Depends on https://github.com/keep-network/keep-core/pull/2856

Reverting the random-beacon dependency back to `>2.0.0-dev <2.0.0-ropsten`. The reason behind this change is that in 2856 we added `Reimbursable.sol` contract which lives in `random-beacon` project. In the `package.json,` the random-beacon dependency is on the `>2.0.0-dev <2.0.0-ropsten` version which doesn't include `Reimbursable.sol` and it broke the build. A commit was added that referenced the local `random-beacon` just for 2856 to pass the build. Here we set it back to `>2.0.0-dev <2.0.0-ropsten` version.

### DIFF
--- a/solidity/ecdsa/package.json
+++ b/solidity/ecdsa/package.json
@@ -62,7 +62,7 @@
     "typescript": "^4.5.4"
   },
   "dependencies": {
-    "@keep-network/random-beacon": ">2.0.0-dev <2.0.0-ropsten",
+    "@keep-network/random-beacon": ">2.0.0-dev. <2.0.0-ropsten",
     "@keep-network/sortition-pools": "^2.0.0-dev.0",
     "@openzeppelin/contracts": "^4.4.1",
     "@threshold-network/solidity-contracts": ">1.1.0-dev <1.1.0-ropsten"
--- a/solidity/ecdsa/package.json
+++ b/solidity/ecdsa/package.json
@@ -62,7 +62,7 @@
     "typescript": "^4.5.4"
   },
   "dependencies": {
-    "@keep-network/random-beacon": ">2.0.0-dev.* <2.0.0-ropsten",
+    "@keep-network/random-beacon": ">2.0.0-dev <2.0.0-ropsten",
     "@keep-network/sortition-pools": "^2.0.0-dev.0",
     "@openzeppelin/contracts": "^4.4.1",
     "@threshold-network/solidity-contracts": ">1.1.0-dev <1.1.0-ropsten"
--- a/solidity/ecdsa/package.json
+++ b/solidity/ecdsa/package.json
@@ -62,7 +62,7 @@
     "typescript": "^4.5.4"
   },
   "dependencies": {
-    "@keep-network/random-beacon": "../random-beacon",
+    "@keep-network/random-beacon": ">2.0.0-dev <2.0.0-ropsten",
     "@keep-network/sortition-pools": "^2.0.0-dev.0",
     "@openzeppelin/contracts": "^4.4.1",
     "@threshold-network/solidity-contracts": ">1.1.0-dev <1.1.0-ropsten"
--- a/solidity/ecdsa/package.json
+++ b/solidity/ecdsa/package.json
@@ -62,7 +62,7 @@
     "typescript": "^4.5.4"
   },
   "dependencies": {
-    "@keep-network/random-beacon": ">2.0.0-dev. <2.0.0-ropsten",
+    "@keep-network/random-beacon": ">2.0.0-dev.* <2.0.0-ropsten",
     "@keep-network/sortition-pools": "^2.0.0-dev.0",
     "@openzeppelin/contracts": "^4.4.1",
     "@threshold-network/solidity-contracts": ">1.1.0-dev <1.1.0-ropsten"
--- a/solidity/ecdsa/yarn.lock
+++ b/solidity/ecdsa/yarn.lock
@@ -612,10 +612,10 @@
     "@openzeppelin/upgrades" "^2.7.2"
     openzeppelin-solidity "2.4.0"
 
-"@keep-network/random-beacon@>2.0.0-dev <2.0.0-ropsten":
-  version "2.0.0-dev.0"
-  resolved "https://registry.yarnpkg.com/@keep-network/random-beacon/-/random-beacon-2.0.0-dev.0.tgz#ea5f450256ea42c842c22fce39611fdb77b0bb49"
-  integrity sha512-8bjljuCp9ZpGAuWtdVOkYT/HslAbnoI3DZdNbhCfp9BGPtWQa32XvOHHu25oqY9xj4IgKURoNqGhSbgb4+m5ww==
+"@keep-network/random-beacon@>2.0.0-dev. <2.0.0-ropsten":
+  version "2.0.0-dev.2"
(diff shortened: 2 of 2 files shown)


pdyraga:
@dimpar For future-us, can you please explain in the PR description what was the reason for this change? I want to copy it to the merge commit. Let's also rebase it over the most recent `main` and undraft. I've just merged #2856.

solidity/ecdsa/package.json
```diff
@@ -62,7 +62,7 @@
     "typescript": "^4.5.4"
   },
   "dependencies": {
-    "@keep-network/random-beacon": ">2.0.0-dev <2.0.0-ropsten",
+    "@keep-network/random-beacon": ">2.0.0-dev. <2.0.0-ropsten",
```
michalinacienciala, line 65: Hmm, I wonder why we need this `.`? Did this help you with regenerating the `yarn.lock`?
When I tried this locally having the dot did not make any difference when I run `yarn upgrade @keep-network/random-beacon` - my `yarn.lock` stayed with `random-beacon resolved to `2.0.0-dev.0` version...

dimpar: For ">2.0.0-dev <2.0.0-ropsten":
- removed `node_modules`, removed `yarn.lock`
- run `yarn` and it resolved to `2.0.0-dev.0`

For ">2.0.0-dev. <2.0.0-ropsten":
- did the same as above and it resolved to `2.0.0-dev.2`

dimpar: The build here also passed with a `.`. It didn't before without a `.`

nkuba: This is strange. We use the syntax without the dot across all our projects and it works fine.

nkuba (changes requested):

solidity/ecdsa/package.json
dimpar: I know, but it didn't work without a `.` neither on my local, nor here in CI

michalinacienciala: I often use https://semver.npmjs.com/ to see which packages fit semver range.
With range `>2.0.0-dev <2.0.0-ropsten` it only highlights `2.0.0-dev.0` as fitting the range. With `>2.0.0-dev. <2.0.0-ropsten` none seem to fit the range.
image
This is very strange...

nkuba: I believe what solved the problem in CI was updating the `yarn.lock`. For the first commit, the build failed because the dependency version for `@keep-network/random-beacon` was set to `2.0.0-dev.0`. CI started passing because you updated the version in `yarn.lock`, not because you added the dot.

For your local environment, I'm not sure why the version wasn't resolved correctly, maybe it was because of yarn cache or something? From what I've seen `yarn upgrade` resolves the latest version correctly for the range we define usually.

dimpar: https://classic.yarnpkg.com/en/docs/dependency-versions/#toc-x-ranges

After the `.` if you leave empty it resolves to `*`.

dimpar: I did a clean clone `git clone` of `main` branch of `keep-core`, so no cache no nothing. I changed
```
"@keep-network/random-beacon": "../random-beacon"
```

to 

```
"@keep-network/random-beacon": ">2.0.0-dev <2.0.0-ropsten"
```

and it didn't resolve to 2.0.0-dev.2, but stayed as -dev.0

when I changed to
```
"@keep-network/random-beacon": ">2.0.0-dev.* <2.0.0-ropsten"
```
it resolved to -dev.2 

I'm not married to this solution, so if you have a better idea let's explore, but I'm not sure how to fix it differently.

nkuba: > https://classic.yarnpkg.com/en/docs/dependency-versions/#toc-x-ranges
> 
> After the `.` if you leave empty it resolves to `*`.


It may be yarn specific, I'm not sure it works the same way with npm.
If we decide to use it it would be probably better to replace the range with `"@keep-network/random-beacon": "2.0.0-dev.*"` if it works correctly.

But still, I'm not sure I'd like to change the approach.

nkuba: `2.0.0-dev.0` was tagged as `latest`. I updated the `latest` tag to point to `2.0.0.-dev.2` and I believe now it resolves to this version. Could you please confirm?

dimpar: > It may be yarn specific, I'm not sure it works the same way with npm.

idk, but we use `yarn` in this project. There had to be probably a major problem with `yarn` to make a decision to go back with `npm`. Besides `npm` might also have some things that won't be compatible with `yarn` or some other dependency manager. In either case, we don't know that for sure. Sometimes it's hard to be very generic.

> If we decide to use it it would be probably better to replace the range with "@keep-network/random-beacon": "2.0.0-dev.*" if it works correctly.

https://github.com/keep-network/keep-core/pull/2878/commits/269f249995f23025e425ae6b2f07c632fc19ac26

> But still, I'm not sure I'd like to change the approach.

I'd also rather just leave as-is and move on with other things, but `>2.0.0-dev <2.0.0-ropsten` didn't work for me.

nkuba: I tested how it resolves with another package that has a lot more versions:
`yarn add "@keep-network/keep-core"@">1.8.0-pre <1.8.0-rc"`
And it resolved all fine to `1.8.0-pre.16`.

dimpar: > `2.0.0-dev.0` was tagged as `latest`. I updated the `latest` tag to point to `2.0.0.-dev.2` and I believe now it resolves to this version. Could you please confirm?

Yes, that thing worked. [23876a4](https://github.com/keep-network/keep-core/pull/2878/commits/23876a44e9f7cce01db27d3195062de5407022d5)

michalinacienciala (approved):

michalinacienciala:
GH actions are misbehaving again (https://www.githubstatus.com/)... Once maintenance is finished, we may need to amend the latest commit to kick off the ckecks again.

michalinacienciala:
@dimpar, I published `random-beacon@2.0.0-dev.3` package and can see that the `yarn upgrade @keep-network/random-beacon` still resolves to `random-beacon@2.0.0-dev.2` (that's the package tagged as `latest`).
I'm currently considering what is the best approach to tackle that. We may end up with changing the `random-beacon` dependency in `solidity/ecdsa/package.json` this way:
```
  "dependencies": {
    "@keep-network/random-beacon": "development",
```
We can merge the #2878 in the current form (once @nkuba approves the changes) or we can wait until we have a clear picture how to best handle versioning/tagging.

Btw, on a different note: shouldn't the version of `random-beacon` package in `solidity/random-beacon/package.json` be bumped up as part of https://github.com/keep-network/keep-core/pull/2856? I think that introducing new contracts qualifies for a reason to bump up the version...

dimpar:
> I published random-beacon@2.0.0-dev.3 package and can see that the yarn upgrade @keep-network/random-beacon still resolves to random-beacon@2.0.0-dev.2 (that's the package tagged as latest).

FWIW `>2.0.0-dev.*` resolved to `-dev.3` which is the latest now.

dimpar:
closing in favor of https://github.com/keep-network/keep-core/pull/2892?

PR #2882, opened 2022-03-23
Merge gas-related updates into one function in WalletRegistry

pdyraga:
Followup of #2880

Gas-related parameters are now updated using a single function on
`WalletRegistry`. This approach lets us save on the contract size and is
consistent with other functions updating various `WalletRegistry`
parameters.

I have also renamed `dkgApprovalGasOffset` to `dkgResultApprovalGasOffset`.
Given that the functions are `submitDkgResult` and  `approveDkgResult`.

I have also took an opportunity and cleaned up a bunch of TODOs that are
no longer valid

### DIFF
--- a/solidity/ecdsa/contracts/EcdsaDkgValidator.sol
+++ b/solidity/ecdsa/contracts/EcdsaDkgValidator.sol
@@ -58,7 +58,6 @@ contract EcdsaDkgValidator {
 
     /// @dev Size in bytes of a public key produced by group members during the
     /// the DKG. The length assumes uncompressed ECDSA public key.
-    // TODO: Investigate if we could use compressed public keys of 33 bytes length.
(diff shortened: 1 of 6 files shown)


dimpar:
If you search for "DkgApproval" it will show you results in the `WalletRegistryGovernance` & `WalletRegistry.WalletCreation.test.ts`. It can be changed to "DkgResultApproval" as well.

Other than ^ it looks good to me.

pdyraga:
> If you search for "DkgApproval" it will show you results in the `WalletRegistryGovernance` & `WalletRegistry.WalletCreation.test.ts`. It can be changed to "DkgResultApproval" as well.
> 
> Other than ^ it looks good to me.

Done in the last commit, https://github.com/keep-network/keep-core/pull/2882/commits/b5bdc294c3c4bfdd4ce75119f19842e2d1307188.

PR #2883, opened 2022-03-23
Replaced Random Beacon failed heartbeat with operator inactivity claim

pdyraga:
The goal of the inactivity claim is to have the group punish operators
who are constantly inactive and not contributing to the work by
excluding them from sortition pool rewards.

The goal of the wallet heartbeat failure - present only in ECDSA - is to
transfer funds to another wallet from the one who failed the heartbeat.

Inactivity claim in ECDSA will be only about excluding operators from
rewards and will not require moving funds.

For this reason, we need to keep these two actions separate and properly
named:
- inactivity claim: exclude operators from rewards
- heartbeat failure: move funds to another safe wallet (just ECDSA)

I went ahead and removed the logic for rewarding operators who
submitted the claim from a special pool of rewards. This action should
refund ETH to the submitter and this will be done in #2860. I did not
want to work on renaming something that soon will be removed anyway.

### DIFF
--- a/solidity/random-beacon/README.adoc
+++ b/solidity/random-beacon/README.adoc
@@ -238,14 +238,14 @@ within the governable eligibility period, `(new_entry % group_size) + 1` and
 so on.
 
 If some group members are notoriously ignoring their duty, the group can vote on
-failed <<heartbeats,heartbeat>> notification for these operators.
+<<inactivity,inactivity>> notification for these operators.
 
 T rewards will be distributed continuously to all operators registered in the beacon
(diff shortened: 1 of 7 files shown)


lukasz-zimnoch (approved):

lukasz-zimnoch:
Some linter errors prevent against merging @pdyraga

pdyraga:
Ooops, my fault. Should become green after the last commit!

lukasz-zimnoch (approved):

PR #2892, opened 2022-03-25
Use `development` tag to configure dependency between packages

michalinacienciala:
Some time ago we temporarily changed `@keep-network/random-beacon`
dependency in `solidity/ecdsa/package.json` from
`>2.0.0-dev <2.0.0-ropsten` to `../random-beacon`. This was only a
tempoorary change, to allow for passing of checks in PR2856. Now we want
to set the dependency to again point to the latest package meant for the
development environment. We discovered, that using
`>2.0.0-dev <2.0.0-ropsten` range to do that is not ideal (if one of the
packages in the range is tagged 'latest', but there are newer packages
which seem to fit the range, those newer packages won't be recognize as
fitting the range). As an alternative, we propose to refer to the latest
package by using following dependency configuration:
```
"@keep-network/random-beacon": "development"
```
This will always point us to the package tagged `deployment`. What we
also want to make sure, is to publish under this tag only the packages
which were built on the `main` branch.

### DIFF
--- a/.github/workflows/npm-ecdsa.yml
+++ b/.github/workflows/npm-ecdsa.yml
@@ -48,6 +48,13 @@ jobs:
           commit: ${{ github.sha }}
 
       - name: Publish package
+        if: github.ref == 'main'
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: npm publish --access=public --tag=development
+
+      - name: Publish package
+        if: github.ref != 'main'
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npm publish --access=public
--- a/.github/workflows/npm-random-beacon.yml
+++ b/.github/workflows/npm-random-beacon.yml
@@ -43,6 +43,13 @@ jobs:
           commit: ${{ github.sha }}
 
       - name: Publish package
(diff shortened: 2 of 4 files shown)


michalinacienciala:
We could merge this PR instead of https://github.com/keep-network/keep-core/pull/2878.

michalinacienciala:
I just need to get the checks passing, once GH Actions will be working normally again...

.github/workflows/npm-ecdsa.yml
```diff
@@ -48,6 +48,13 @@ jobs:
           commit: ${{ github.sha }}
 
       - name: Publish package
+        if: github.ref == 'main'
```
nkuba, line 51: Do we need these `main` checks? We already have it handled by https://github.com/keep-network/keep-core/blob/b09fcc41f8211ef6ea8451ecc646b4e733a70207/.github/workflows/npm-ecdsa.yml#L4-L6

So the workflow will be executed only on push to `main`.

nkuba (changes requested):

.github/workflows/npm-ecdsa.yml
michalinacienciala: It can be also executed manually on any branch, that's why I added this clause.

nkuba (approved):

PR #2894, opened 2022-03-25
Ledger packages migration

michalsmiarowski:
The package that we use for Ledger - `@ledgerhq/hw-transport-u2f` have been deprecated.

We migrate from `@ledgerhq/hw-transport-u2f` to the newer and safer `@ledgerhq/hw-transport-webusb`

Mozilla Firefox does not support WebUSB at this moment so we remove the possibility to log in directly from Ledger Live on Firefox (and other browsers expect Chrome).

More info: [https://github.com/LedgerHQ/ledgerjs/blob/master/docs/migrate_webusb.md](https://github.com/LedgerHQ/ledgerjs/blob/master/docs/migrate_webusb.md)

Main changes:
- install `@ledgerhq/hw-transport-webusb` and uninstall deprecated `@ledgerhq/hw-transport-u2f` package
- install `react-device-detect` library which help with detecting what device and what browser does user use
- display proper message in Ledger Login window, that informs user that are viewing the dashboard from different browser than Chrome, that it's impossible to login straight from Ledger and that they should either use Chrome or log in through Ledger Live Bridge on MetaMask

Ledger login modal on Firefox:
image

Testing:

To test the Ledger log in locally install packages from mainnet and run the dashboard with HTTP=true flag:
`HTTPS=true npm start`

### DIFF
--- a/solidity-v1/dashboard/package-lock.json
+++ b/solidity-v1/dashboard/package-lock.json
@@ -2884,6 +2884,76 @@
         }
       }
     },
+    "@ledgerhq/hw-transport-webusb": {
+      "version": "6.24.1",
+      "resolved": "https://registry.npmjs.org/@ledgerhq/hw-transport-webusb/-/hw-transport-webusb-6.24.1.tgz",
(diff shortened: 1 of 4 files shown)


r-czajkowski (approved):
LGTM! :shipit:

PR #2900, opened 2022-03-28
Define library properties in Parameters struct

nkuba:
To reduce contract size and better organize the code we define parameters in structures inside libraries.

What is surprising those changes didn't reduce the contract size 😒 

```
 ·--------------------------|-------------·-------------|
 |  Contract Name           · BEFORE (KB) ·  AFTER (KB) │
 ···························|···························|
 |  RandomBeacon            ·     21.526  ·     21.706  │
 ···························|···························|
 |  RandomBeaconGovernance  ·     24.166  ·     24.569  │
 ·--------------------------|-------------·-------------|
```

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -720,7 +720,7 @@ contract RandomBeacon is IRandomBeacon, Ownable {
 
         relay.requestEntry(groupId);
 
-        fundDkgRewardsPool(msg.sender, relay.relayRequestFee);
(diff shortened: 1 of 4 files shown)


pdyraga:
> What is surprising those changes didn't reduce the contract size 😒

Library does not have its own storage, all of it lands in the contract :) Although it looks from the code perspective something is extracted, on the storage level it is not.

nkuba:
> > What is surprising those changes didn't reduce the contract size 😒
> 
> Library does not have its own storage, all of it lands in the contract :) Although it looks from the code perspective something is extracted, on the storage level it is not.

I'm not sure I get your comment. I didn't want to extract the storage but reduce contract size by replacing couple of functions with one.

PR #2905, opened 2022-03-29
`notifyOperatorInactivity` for Wallet Registry

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/2839

The function lets the wallet signing group operators notify about operators who are
inactive in the group. Using this function, a majority of the wallet signing
group can decide about punishing specific group members who constantly fail
to do their job. If the provided claim is proved to be valid and signed by
a sufficient number of group members, operators of members deemed as inactive are
banned from sortition pool rewards for the duration specified by
`sortitionPoolRewardsBanDuration` parameter. The sender of the claim must be one
of the claim signers. This function can be called only for registered wallets.

TODO (in a separate PR):
- [ ] ETH reimbursement for `notifyOperatorInactivity`

### DIFF
--- a/solidity/ecdsa/contracts/libraries/Wallets.sol
+++ b/solidity/ecdsa/contracts/libraries/Wallets.sol
@@ -37,11 +37,11 @@ library Wallets {
 
     /// @notice Registers a new wallet.
     /// @dev Uses a public key hash as a unique identifier of a wallet.
-    /// @param membersIdsHash Keccak256 hash of group members identifiers array.
-    /// @param publicKey Uncompressed public key.
-    /// @return walletID Wallet's ID.
-    /// @return publicKeyX Wallet's public key's X coordinate.
-    /// @return publicKeyY Wallet's public key's Y coordinate.
+    /// @param membersIdsHash Keccak256 hash of group members identifiers array
+    /// @param publicKey Uncompressed public key
+    /// @return walletID Wallet's ID
+    /// @return publicKeyX Wallet's public key's X coordinate
+    /// @return publicKeyY Wallet's public key's Y coordinate
     function addWallet(
         Data storage self,
         bytes32 membersIdsHash,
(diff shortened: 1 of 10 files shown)


solidity/ecdsa/contracts/libraries/Wallets.sol
```diff
@@ -81,11 +81,24 @@ library Wallets {
         return self.registry[walletID].publicKeyX != bytes32(0);
     }
 
+    /// @notice Returns Keccak256 hash of the wallet signing group members
+    ///         identifiers array. Group members do not include operators
+    ///         selected by the sortition pool that misbehaved during DKG.
+    /// @param walletID ID of the wallet
+    /// @return Wallet signing group members hash
+    function getWalletMembersIdsHash(Data storage self, bytes32 walletID)
```
lukasz-zimnoch, line 86: Worth documenting that this method will return `bytes32(0)` if `walletID` refers to an unknown wallet.

solidity/ecdsa/contracts/libraries/Wallets.sol
```diff
@@ -81,11 +81,24 @@ library Wallets {
...
+    }
+
     /// @notice Gets public key of a wallet with a given wallet ID.
     ///         The public key is returned in an uncompressed format as a 64-byte
     ///         concatenation of X and Y coordinates.
-    /// @param walletID ID of the wallet.
-    /// @return Uncompressed public key of the wallet.
+    /// @param walletID ID of the wallet
+    /// @return Uncompressed public key of the wallet
     function getWalletPublicKey(Data storage self, bytes32 walletID)
```
lukasz-zimnoch, line 99: Worth documenting that this method will revert if `walletID` refers to an unknown wallet.

lukasz-zimnoch (approved):
Just two non-blocking comments.

solidity/ecdsa/contracts/libraries/Wallets.sol
pdyraga: Done in #2908, commit https://github.com/keep-network/keep-core/pull/2908/commits/b9898f4d6eace62316d9bb512a94559d7f4f96a0.

solidity/ecdsa/contracts/libraries/Wallets.sol
pdyraga: Done in #2908, commit https://github.com/keep-network/keep-core/pull/2908/commits/b9898f4d6eace62316d9bb512a94559d7f4f96a0.

PR #2906, opened 2022-03-29
Remove functions from RandomBeaconStub

nkuba:
The functions are not necessary for the RandomBeaconStub as we can use
existing functions to obtain required data.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -1009,7 +1009,7 @@ contract RandomBeacon is IRandomBeacon, Ownable {
     /// @notice Locks the state of group creation.
     /// @dev This function is meant to be used by test stubs which inherits
     ///      from this contract and needs to lock the DKG state arbitrarily.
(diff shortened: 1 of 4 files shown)


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -1009,7 +1009,7 @@ contract RandomBeacon is IRandomBeacon, Ownable {
     /// @notice Locks the state of group creation.
     /// @dev This function is meant to be used by test stubs which inherits
     ///      from this contract and needs to lock the DKG state arbitrarily.
-    function dkgLockState() internal {
+    function dkgLockState() external {
```
pdyraga, line 1012: Is this correct? Anyone can now lock `RandomBeacon` state even though no DKG is in progress.

pdyraga (changes requested):

pdyraga (approved):

pdyraga:
Slither build is failing because of https://github.com/crytic/slither/issues/1067. Ignoring for now.

PR #2907, opened 2022-03-30
Implemented `WalletRegistry.closeWallet` logic

### LINKED ISSUE #2864: Let the owner close a wallet
Wallet Registry should be notified by the wallet owner about a wallet termination and notify the members that they no longer need to monitor the wallet.

Refs https://github.com/keep-network/tbtc-v2/issues/139

pdyraga:
Closes #2864
Depends on #2905; Keeping it as a draft until #2905 is merged.

So far this function was just a stub and now the implementation is there. The
logic is pretty simple - we emit an event and remove the wallet from the
registry.

The idea behind removing the wallet from the registry is that we do not want to
allow reporting inactivity for a closed wallet and at the same time, we do not
want to maintain separate wallet states in `WalletRegistry` - it's happening
already in tBTCv2 `Bridge` contract.

### DIFF
--- a/solidity/ecdsa/contracts/libraries/Wallets.sol
+++ b/solidity/ecdsa/contracts/libraries/Wallets.sol
@@ -67,6 +67,18 @@ library Wallets {
         self.registry[walletID].publicKeyY = publicKeyY;
     }
 
+    /// @notice Deletes wallet with the given ID from the registry. Reverts
+    ///         if wallet with the given ID has not been registered or if it
+    ///         has already been closed.
+    function deleteWallet(Data storage self, bytes32 walletID) internal {
+        require(
+            isWalletRegistered(self, walletID),
+            "Wallet with the given ID has not been registered"
+        );
+
+        delete self.registry[walletID];
+    }
+
     /// @notice Checks if a wallet with the given ID is registered.
     /// @param walletID Wallet's ID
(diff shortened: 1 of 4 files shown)


solidity/ecdsa/contracts/libraries/Wallets.sol
```diff
@@ -67,6 +67,18 @@ library Wallets {
...
+    /// @notice Deletes wallet with the given ID from the registry. Reverts
+    ///         if wallet with the given ID has not been registered or if it
+    ///         has already been closed.
+    function deleteWallet(Data storage self, bytes32 walletID) internal {
+        require(
+            isWalletRegistered(self, walletID),
+            "Wallet with the given ID has not been registered"
+        );
+
+        delete self.registry[walletID];
```
lukasz-zimnoch, line 79: I'm wondering whether deleting is enough here. What if someone submits a deleted wallet pubkey via `approveDkgResult` and registers it again? I know the result must be signed by multiple members but still. Shouldn't we store an additional piece of info that will prevent re-adding an already deleted wallet?

lukasz-zimnoch: Yep that sounds good!

lukasz-zimnoch (approved):

PR #2908, opened 2022-03-30
ECDSA heartbeat failure handling

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/2839

`notifyOperatorInactivity` function allows to signal about single operators
being inactive as well as to signal wallet-wide heartbeat failures that are
propagated to the wallet owner who should begin the procedure of moving
responsibilities to another wallet given that the wallet that failed the heartbeat
may soon not be able to function and provide new signatures.


solidity/ecdsa/contracts/libraries/EcdsaPubKey.sol
```diff
@@ -0,0 +1,38 @@
...
+// ▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+//
+//                           Trust math, not hardware.
+
+pragma solidity ^0.8.9;
+
+/// @notice Auxiliary library providing functions to convert between bytes32
+///         public key X and Y coordinates and uncompressed public key
+///         bytes array format.
+library EcdsaPubKey {
```
nkuba, line 20: Do we need to introduce this library? The functions look super simple and it seems like we could live without them. How does it affect execution gas costs?

nkuba (approved):

solidity/ecdsa/contracts/libraries/EcdsaPubKey.sol
pdyraga: It should not affect gas costs because internal functions of not-externally-linked libraries are just inlined by the compiler. I was not sure about having this library but doing `[:32]` `[32:]` in the code outside of `Wallets` felt wrong. I am open for removing it if you think it is too much.

nkuba: > It should not affect gas costs because internal functions of not-externally-linked libraries are just inlined by the compiler.

Yeah, I know that the functions got inlined, but isn't it adding something to the general cost that we will call an internal function in between?

pdyraga: When a function is inlined, its body should be inlined, it should not require additional calls. But... I compared gas costs and there are ~200 units less consumed when the library is not used. Hard to say why, maybe it's not how solc does the inlining or maybe there is something more. Anyway, this library could be too much - removed it in https://github.com/keep-network/keep-core/pull/2908/commits/7c7a8a8284f0e24a72fe48ab4de9b9369afaa8b7.

solidity/ecdsa/contracts/libraries/EcdsaPubKey.sol
```diff
@@ -0,0 +1,38 @@
...
+    ///         public key bytes.
+    function getX(bytes calldata publicKey) internal pure returns (bytes32) {
+        return bytes32(publicKey[:32]);
+    }
+
+    /// @notice Extract Y public key coordinate from uncompressed, concatenated
+    ///         public key bytes.
+    function getY(bytes calldata publicKey) internal pure returns (bytes32) {
+        return bytes32(publicKey[32:]);
+    }
```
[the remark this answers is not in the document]
pdyraga: This library was removed in https://github.com/keep-network/keep-core/pull/2908/commits/7c7a8a8284f0e24a72fe48ab4de9b9369afaa8b7. See [this](https://github.com/keep-network/keep-core/pull/2908#discussion_r839283566) discussion please.

lukasz-zimnoch (approved):
LGTM! Leaving honors to @nkuba

nkuba (approved):

PR #2903, opened 2022-03-29
Integrate T TokenStaking with RandomBeacon

nkuba:
Refs: https://github.com/keep-network/keep-core/pull/2850
Closes: https://github.com/keep-network/keep-core/issues/2715
Depends on: https://github.com/keep-network/keep-core/pull/2906

This PR integrates Random Beacon with [Threshold Network `TokenStaking` contract](https://github.com/threshold-network/solidity-contracts/blob/v1.1.0/contracts/staking/TokenStaking.sol).

It implements functions to let the operator join the sortition pool and update its status in the sortition pool. `StakingStub` contract has been removed and all tests operate on a real T `TokenStaking` now.

The changes were already implemented in ECDSA Wallets in https://github.com/keep-network/keep-core/pull/2850 and this PR is based on it.

### Operators To Staking Providers Mapping

Please note that due to 8e9b2a48227bc5830b3b508c9729403bb32ce250 a gas cost of executing functions that call seize and slash grew significantly:

```
        method            |   min   |   max   |   avg
-------------------------------------------------------
submitRelayEntry          | 213867  | 3693768 | 1760490 
reportRelayEntryTimeout   | 3641465 | 3659466 | 3654018
reportUnauthorizedSigning |    -    |    -    | 3774342
```

PR #2909, opened 2022-03-31
ECDSA Wallets: Use unique accounts for staking and operator registration roles in tests

nkuba:
We need to make our tests more robust by using unique addresses for each role that is related to staking and operator registration. 
This requires configuring a hardhat network with more than 500 accounts so each address is unique and doesn't overlap.
Thanks to that the tests are more reliable and reduce the possibility to give false-positive results.

In the Random Beacon we introduce this change in https://github.com/keep-network/keep-core/pull/2903.

### DIFF
--- a/solidity/ecdsa/test/utils/operators.ts
+++ b/solidity/ecdsa/test/utils/operators.ts
@@ -5,10 +5,14 @@ import { ethers } from "hardhat"
 // eslint-disable-next-line import/no-cycle
 import { params } from "../fixtures"
 
-import type { Address } from "hardhat-deploy/types"
-import type { BigNumber } from "ethers"
+import type { BigNumber, BigNumberish } from "ethers"
 import type { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers"
-import type { WalletRegistry, T } from "../../typechain"
+import type {
+  WalletRegistry,
+  T,
+  SortitionPool,
+  TokenStaking,
+} from "../../typechain"
 
 export type OperatorID = number
 export type Operator = {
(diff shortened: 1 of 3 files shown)


solidity/ecdsa/test/utils/operators.ts
```diff
@@ -18,55 +22,50 @@ export type Operator = {
 
 export async function registerOperators(
   walletRegistry: WalletRegistry,
-  tToken: T,
-  addresses: Address[],
+  t: T,
+  numberOfOperators: number,
+  unnamedSignersOffset = 0,
```
pdyraga, line 27: Shouldn't it be 10 by default?

nkuba: 🤷🏻 Yeah, we could do that.

pdyraga: If we don't do that, then one of the operators/staking providers/beneficiaries/authorizers uses the same address as deployer, third party etc. Correct?

nkuba: It won't overlap with the deployer as it's a named account. We define offset based on the use case in tests, for example here:
https://github.com/keep-network/keep-core/blob/e691e2406f0be1ed4b89a52e9f7ae783cb49a7e0/solidity/ecdsa/test/fixtures/index.ts#L77-L96

We pull one unnamed signer for `thirdParty` hence we define offset as `1`.

pdyraga: Ah, I see now. Let's leave it as-is then.

pdyraga (approved):

solidity/ecdsa/test/utils/operators.ts
nkuba: Please see if https://github.com/keep-network/keep-core/pull/2910 helps anything.

pdyraga: it does!

PR #2914, opened 2022-04-06
Notify operator inactivity refund

dimpar:
Refs: https://github.com/keep-network/keep-core/issues/2839

Refunding ETH for calling `notifyOperatorInactivity`. Gas offset was set to `85000` with a try&error approach. The notifier will be a bit net-positive, but less than `0.002ETH` when the gas price is 200 gwei. In reality, it will be even lower because tests use some slack.
Gas offset is a gov param and can be adjusted according to the current market conditions.


pdyraga:
👀

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -76,10 +76,16 @@ contract WalletRegistry is
     ///         other member that will call the DKG approve function.
     uint256 public dkgResultSubmissionGas = 275000;
 
-    // @notice Gas meant to balance the DKG result approval's overall cost. Can
-    //         be updated by the governace based on the current market conditions.
+    /// @notice Gas is meant to balance the DKG result approval's overall cost.
```
pdyraga, line 79: Should it be "Gas is meant to..." or "Gas meant to"? My understanding is that the second is correct, given that we are explaining what `dkgResultApprovalGasOffset` is for.

```suggestion
    /// @notice Gas meant to balance the DKG result approval's overall cost.
```

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -71,6 +71,9 @@ contract WalletRegistryGovernance is Ownable {
     address payable public newReimbursementPool;
     uint256 public reimbursementPoolChangeInitiated;
 
+    uint256 public notifyOperatorInactivityGasOffsetChangeInitiated;
+    uint256 public newNotifyOperatorInactivityGasOffset;
```
pdyraga, line 75: Can we place them above `newReimbursementPool`, right after DKG result approval gas offset fields?

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -172,6 +175,14 @@ contract WalletRegistryGovernance is Ownable {
     );
     event ReimbursementPoolUpdated(address reimbursementPool);
 
+    event NotifyOperatorInactivityGasOffsetUpdateStarted(
+        uint256 notifyOperatorInactivityGasOffset,
+        uint256 timestamp
+    );
+    event NotifyOperatorInactivityGasOffsetUpdated(
+        uint256 notifyOperatorInactivityGasOffset
+    );
```
pdyraga, line 184: Can we place them above `ReimbursementPoolUpdateStarted`, right after events for DKG approval gas offset?

solidity/ecdsa/deploy/02_deploy_reimbursement_pool.ts
```diff
@@ -14,6 +18,11 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
     log: true,
   })
 
+  await deployerSigner.sendTransaction({
+    to: ReimbursementPool.address,
+    value: ethers.utils.parseEther("100.0"), // Send 100.0 ETH
+  })
```
pdyraga, line 24: Can we do it only if `deployments.getNetworkName() === "hardhat"`?

For mainnet deployment, we definitely not want to send 100 ETH from the deployer account.

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
```diff
@@ -2418,4 +2419,143 @@ describe("WalletRegistryGovernance", async () => {
...
+      it("should revert", async () => {
+        await expect(
+          walletRegistryGovernance
+            .connect(thirdParty)
+            .beginNotifyOperatorInactivityGasOffsetUpdate(100)
+        ).to.be.revertedWith("Ownable: caller is not the owner")
+      })
+    })
+
+    context("when the caller is the owner and value is correct", () => {
```
pdyraga, line 2434:
```suggestion
    context("when the caller is the owner", () => {
```

We do not really validate the value.

pdyraga (approved):
Some non-blocking comments. @dimpar please let me know what you think and I can address them in my PR - I need to add `withdrawIneligibleRewards` to `WalletRegistry`.

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
dimpar: Sure

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
dimpar: Sure

solidity/ecdsa/deploy/02_deploy_reimbursement_pool.ts
dimpar: Yea, good catch, certainly it was added only for the testing hardhat purposes.

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
dimpar: Right, copy pasta from other tests.

dimpar:
> 

Sure, I've replied to your comments. I can create a new PR addressing them or also totally fine if you address them in your PR. Thanks

PR #2919, opened 2022-04-07
RandomBeacon: withdrawRewards

pdyraga:
See https://github.com/keep-network/sortition-pools/pull/156

Previously, `SortitionPool.withdrawRewards` could be called directly and
the beneficiary address was fetched from the staking contract based on
the operator address.

This is no longer going to work - T `TokenStaking` contract does not store
operators but staking providers - operator is now a per-application role.

We now expect `SortitionPool.withdrawReward` call to be proxied through the
application `RandomBeacon` contract that will fetch the beneficiary address
from T `TokenStaking`.

This PR is also adding unit tests for `RandomBeacon.withdrawIneligibleRewards`.
This function was not covered with unit tests previously.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -562,7 +562,7 @@ contract RandomBeacon is IRandomBeacon, IApplication, Ownable {
     }
 
     /// @notice Withdraw rewards for the given staking provider to their
-    ///         beneficiary address Reverts if staking provider has not
+    ///         beneficiary address. Reverts if staking provider has not
     ///         registered the operator address.
     function withdrawRewards(address stakingProvider) external {
         address operator = stakingProviderToOperator(stakingProvider);
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -561,6 +561,16 @@ contract RandomBeacon is IRandomBeacon, IApplication, Ownable {
         );
     }
 
+    /// @notice Withdraw rewards for the given staking provider to their
+    ///         beneficiary address Reverts if staking provider has not
+    ///         registered the operator address.
+    function withdrawRewards(address stakingProvider) external {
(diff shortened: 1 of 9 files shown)


solidity/random-beacon/package.json
```diff
@@ -30,7 +30,7 @@
     "prepack": "tsc -p tsconfig.export.json && hardhat export-artifacts export/artifacts"
   },
   "dependencies": {
-    "@keep-network/sortition-pools": "^2.0.0-pre.6",
+    "@keep-network/sortition-pools": "^2.0.0-pre.7",
```
dimpar, line 33: Non blocking, but shouldn't it be auto updated since we have `^`? If it's not updating then we might have a similiar issue that we had when we put a `random-beacon` dependency in `ecdsa`. cc @nkuba

dimpar (approved):

solidity/random-beacon/package.json
pdyraga: I changed it because the code requires _at least_ `2.0.0-pre.7` now, it's not going to compile with `2.0.0-pre.6`. `^2.0.0-pre.7` is reflecting it: _at least `2.0.0-pre.7`, please_.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -561,6 +561,16 @@ contract RandomBeacon is IRandomBeacon, IApplication, Ownable {
...
     }
 
+    /// @notice Withdraw rewards for the given staking provider to their
+    ///         beneficiary address. Reverts if staking provider has not
+    ///         registered the operator address.
+    function withdrawRewards(address stakingProvider) external {
+        address operator = stakingProviderToOperator(stakingProvider);
+        require(operator != address(0), "Unknown operator");
+        (, address beneficiary, ) = staking.rolesOf(stakingProvider);
+        sortitionPool.withdrawRewards(operator, beneficiary);
```
nkuba, line 571: Would it make sense to add a requirement that `beneficiary != address(0)`? Maybe it should be a part of `sortitionPool.withdrawRewards`?

PR #2920, opened 2022-04-08
ECDSA: withdraw sortition pool rewards

pdyraga:
See https://github.com/keep-network/sortition-pools/pull/156

~~Depends on https://github.com/keep-network/keep-core/pull/2919; I am planning no more changes in the code but keeping it as a draft until #2919 is merged.~~

Previously, `SortitionPool.withdrawRewards` could be called directly and
the beneficiary address was fetched from the staking contract based on
the operator address.

This is no longer going to work - T `TokenStaking` contract does not store
operators but staking providers - operator is now a per-application role.

We now expect `SortitionPool.withdrawReward` call to be proxied through the
application `WalletRegistry` contract that will fetch the beneficiary address
from T `TokenStaking`.

This PR is also implementing `withdrawIneligibleRewards` function
on the `WalletRegistry`.

`SortitionPool` allows the owner to withdraw rewards that would normally
be allocated to operators in case these operators were marked as ineligible
for rewards. This needs to be called by the `SortitionPool` owner, which
here is the `WalletRegistry`.

### DIFF
--- a/solidity/ecdsa/deploy/01_deploy_sortition_pool.ts
+++ b/solidity/ecdsa/deploy/01_deploy_sortition_pool.ts
@@ -13,7 +13,7 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
 
   const SortitionPool = await deployments.deploy("SortitionPool", {
     from: deployer,
-    args: [TokenStaking.address, T.address, POOL_WEIGHT_DIVISOR],
+    args: [T.address, POOL_WEIGHT_DIVISOR],
     log: true,
   })
 
--- a/solidity/ecdsa/deploy/01_deploy_sortition_pool.ts
+++ b/solidity/ecdsa/deploy/01_deploy_sortition_pool.ts
@@ -8,7 +8,6 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
 
   const POOL_WEIGHT_DIVISOR = to1e18(1) // TODO: Update value
 
-  const TokenStaking = await deployments.get("TokenStaking")
   const T = await deployments.get("T")
 
   const SortitionPool = await deployments.deploy("SortitionPool", {
--- a/solidity/ecdsa/test/WalletRegistry.Rewards.test.ts
+++ b/solidity/ecdsa/test/WalletRegistry.Rewards.test.ts
@@ -4,6 +4,7 @@ import { expect } from "chai"
 import { walletRegistryFixture } from "./fixtures"
 import ecdsaData from "./data/ecdsa"
 import { createNewWallet } from "./utils/wallets"
+import { signOperatorInactivityClaim } from "./utils/inactivity"
 
 import type { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers"
 import type { FakeContract } from "@defi-wonderland/smock"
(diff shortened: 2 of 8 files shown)


solidity/ecdsa/test/WalletRegistry.Rewards.test.ts
```diff
@@ -0,0 +1,174 @@
...
+        expect(await tToken.balanceOf(beneficiary)).to.equal(0)
+        await walletRegistry.withdrawRewards(stakingProvider)
+        expect(await tToken.balanceOf(beneficiary)).to.be.gt(0)
+      })
+    })
+  })
+
+  describe("withdrawIneligibleRewards", () => {
+    const inactiveMembersIndices = [1, 5, 10]
+    const heartbeatFailed = false
```
dimpar, line 107: Not biggy, but now we sort of have 2 phrases to describe the same thing. Before we only used "heartbeat failures" and then we started using "operator inactivity". How about we rename it to "isOperatorActive"?

dimpar (dismissed):

solidity/ecdsa/test/WalletRegistry.Rewards.test.ts
pdyraga: It is not the same thing. Both `WalletRegistry` and `RandomBeacon` has a function to `notifyOperatorInactivity`. In both cases, this function is checking signatures of group members supporting the claim, and marking inactive operators as not eligible for rewards.

Only `WalletRegistry` accepts an additional parameter to the inactivity claim: a boolean flag whether the heartbeat failed:
https://github.com/keep-network/keep-core/blob/8dc2df00236d2a09f038187c83fd8441aa731d63/solidity/ecdsa/contracts/libraries/EcdsaInactivity.sol#L40-L50

Operator inactivity is a minor event leading to marking operators as ineligible for rewards.

Failed heartbeat (due to too many inactive operators) is a major event and leads to notification of tBTC `Bridge` that triggers the procedure of moving funds to another ECDSA wallet.

pdyraga:
#2919 is merged so I am undrafting this PR.

solidity/ecdsa/test/WalletRegistry.Rewards.test.ts
dimpar: Ah, I see.. so we use "operator inactivity" when talking about a single operator, but "failed heartbeat" when many.

solidity/ecdsa/deploy/01_deploy_sortition_pool.ts
```diff
@@ -13,7 +13,7 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
 
   const SortitionPool = await deployments.deploy("SortitionPool", {
     from: deployer,
-    args: [TokenStaking.address, T.address, POOL_WEIGHT_DIVISOR],
+    args: [T.address, POOL_WEIGHT_DIVISOR],
```
dimpar, line 16: `TokenStaking` on line 11 can be removed now

pdyraga: Good catch. Fixed in https://github.com/keep-network/keep-core/pull/2920/commits/f5da4a4ae51e314233bf7acf450fa15fcb5fac26.

dimpar (approved):

nkuba (approved):

PR #2921, opened 2022-04-08
WalletRegistry and Reimbursable ownership/governance updates

nkuba:
~~Depends on https://github.com/keep-network/keep-core/pull/2922~~

In this PR we modify the `WalletRegistry` to make it ready for becoming upgradable.

We removed `Ownable` with `Governable` which is a simpler solution and is consistent with `TokenStaking` contract from threshold-network.

Also, we implement `onlyReimbursableAdmin` from `Reimbursable` contract (https://github.com/keep-network/keep-core/pull/2922).

### DIFF
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -239,6 +239,11 @@ contract WalletRegistry is
         _;
     }
 
+    modifier onlyReimbursableAdmin() override {
+        require(governance == msg.sender, "Caller is not the governance");
+        _;
+    }
+
     constructor(
         SortitionPool _sortitionPool,
         IStaking _staking,
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -21,6 +21,7 @@ import "./libraries/EcdsaDkg.sol";
 import "./libraries/Wallets.sol";
 import "./libraries/EcdsaInactivity.sol";
 import "./EcdsaDkgValidator.sol";
+import "./Governable.sol";
 
 import "@keep-network/sortition-pools/contracts/SortitionPool.sol";
 import "@keep-network/random-beacon/contracts/api/IRandomBeacon.sol";
@@ -36,7 +37,7 @@ contract WalletRegistry is
     IWalletRegistry,
     IRandomBeaconConsumer,
     IApplication,
-    Ownable,
+    Governable,
     Reimbursable
 {
     using EcdsaAuthorization for EcdsaAuthorization.Data;
@@ -266,6 +267,8 @@ contract WalletRegistry is
         dkg.setResultChallengePeriodLength(11520); // ~48h assuming 15s block time
         dkg.setResultSubmissionTimeout(100 * 20);
         dkg.setSubmitterPrecedencePeriodLength(20);
+
+        _transferGovernance(msg.sender);
     }
 
     /// @notice Used by staking provider to set operator address that will
@@ -393,13 +396,13 @@ contract WalletRegistry is
     }
 
     /// @notice Updates address of the Random Beacon.
-    /// @dev Can be called only by the contract owner, which should be the
+    /// @dev Can be called only by the contract guvnor, which should be the
     ///      wallet registry governance contract. The caller is responsible for
     ///      validating parameters.
     /// @param _randomBeacon Random Beacon address.
     function upgradeRandomBeacon(IRandomBeacon _randomBeacon)
         external
-        onlyOwner
+        onlyGovernance
     {
         randomBeacon = _randomBeacon;
         emit RandomBeaconUpgraded(address(_randomBeacon));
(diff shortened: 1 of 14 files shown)


solidity/ecdsa/test/WalletRegistryGovernance.test.ts
```diff
@@ -357,24 +357,24 @@ describe("WalletRegistryGovernance", async () => {
       })
 
       it("should not transfer the ownership", async () => {
```
pdyraga, line 359:
```suggestion
      it("should not transfer the governance", async () => {
```
?

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
```diff
@@ -441,39 +441,39 @@ describe("WalletRegistryGovernance", async () => {
...
             .connect(governance)
-            .finalizeWalletRegistryOwnershipTransfer()
+            .finalizeWalletRegistryGovernanceTransfer()
         })
 
         after(async () => {
           await restoreSnapshot()
         })
 
         it("should transfer wallet registry ownership", async () => {
```
pdyraga, line 459:
```suggestion
        it("should transfer wallet registry governance", async () => {
```
?

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -84,11 +84,11 @@ contract WalletRegistryGovernance is Ownable {
     );
     event GovernanceDelayUpdated(uint256 governanceDelay);
 
-    event WalletRegistryOwnershipTransferStarted(
+    event WalletRegistryGovernanceTransferStarted(
         address newWalletRegistryOwner,
```
pdyraga, line 88:
```suggestion
        address newWalletRegistryGovernance,
```

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -84,11 +84,11 @@ contract WalletRegistryGovernance is Ownable {
...
         uint256 timestamp
     );
-    event WalletRegistryOwnershipTransferred(address newWalletRegistryOwner);
+    event WalletRegistryGovernanceTransferred(address newWalletRegistryOwner);
```
pdyraga, line 91:
```suggestion
    event WalletRegistryGovernanceTransferred(address newWalletRegistryGovernance);
```

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -29,7 +29,7 @@ contract WalletRegistryGovernance is Ownable {
     uint256 public governanceDelayChangeInitiated;
 
     address public newWalletRegistryOwner;
```
pdyraga, line 31:
```suggestion
    address public newWalletRegistryGovernance;
```

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -262,9 +262,9 @@ contract WalletRegistryGovernance is Ownable {
         newGovernanceDelay = 0;
     }
 
-    /// @notice Begins the wallet registry ownership transfer process.
+    /// @notice Begins the wallet registry governance transfer process.
     /// @dev Can be called only by the contract owner.
-    function beginWalletRegistryOwnershipTransfer(
+    function beginWalletRegistryGovernanceTransfer(
         address _newWalletRegistryOwner
```
pdyraga, line 268:
```suggestion
        address _newWalletRegistryGovernance
```

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -430,13 +438,16 @@ contract WalletRegistry is
     ///      validating parameters. The wallet owner has to implement `IWalletOwner`
```
pdyraga, line 441: At line 439 should say "only by the contract guvnor"

solidity/ecdsa/test/Governable.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    ;({ deployer, governance } = await ethers.getNamedSigners())
+    ;[thirdParty] = await ethers.getUnnamedSigners()
+
+    const GovernableFactory: GovernableImpl__factory =
+      await ethers.getContractFactory("GovernableImpl", deployer)
+    governable = await GovernableFactory.deploy()
+  })
+
+  describe("constructor", () => {
+    it("sets governance to default zero address", async () => {
```
pdyraga, line 28: Keep in mind we now use `it("should...` pattern everywhere

solidity/ecdsa/test/Governable.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    describe("when called by a third party", () => {
+      it("succeeds", async () => {
+        const newGovernance = ethers.Wallet.createRandom().address
+
+        await governable
+          .connect(thirdParty)
+          ._transferGovernanceExposed(newGovernance)
+
+        expect(await governable.governance()).to.be.equal(newGovernance)
+      })
```
pdyraga, line 174: I think we should delete these two cases. `_transferGovernance` can be called only by the contract itself. If we want to test that it works, let's leave the deployer case because that's what is happening in the constructor case. Or even better, remove "when called by..." altogether because only the contract can call it.

pdyraga (approved):

pdyraga:
Left some non-blocking comments that can be addressed in a follow-up PR. No reason to block here, especially given this change needs to be integrated with `RandomBeacon`.

nkuba:
> Left some non-blocking comments that can be addressed in a follow-up PR. No reason to block here, especially given this change needs to be integrated with `RandomBeacon`.

Thanks, I posted updates in https://github.com/keep-network/keep-core/pull/2948

PR #2922, opened 2022-04-08
Remove Ownable from Reimbursable

nkuba:
Reimbursable is an abstract contract with a function `updateReimbursementPool` that should be able to be executed only by a specific role. 

Using Ownable for this contract was not necessary, especially since this causes problems when using upgradable implementations, where we should use `OwnableUpgradable` instead of `Ownable`.

With the `onlyReimbursableAdmin` virtual modifier we leave it to the implementation to check the role according to their needs.

### DIFF
--- a/solidity/random-beacon/contracts/Reimbursable.sol
+++ b/solidity/random-beacon/contracts/Reimbursable.sol
@@ -14,10 +14,9 @@
 
 pragma solidity ^0.8.9;
 
-import "@openzeppelin/contracts/access/Ownable.sol";
 import "./ReimbursementPool.sol";
 
-abstract contract Reimbursable is Ownable {
+abstract contract Reimbursable {
     ReimbursementPool public reimbursementPool;
 
     event ReimbursementPoolUpdated(address newReimbursementPool);
@@ -28,9 +27,13 @@ abstract contract Reimbursable is Ownable {
         reimbursementPool.refund(gasStart - gasleft(), receiver);
     }
 
+    modifier onlyReimbursableAdmin() virtual {
+        _;
+    }
+
     function updateReimbursementPool(ReimbursementPool _reimbursementPool)
         external
-        onlyOwner
+        onlyReimbursableAdmin
     {
         emit ReimbursementPoolUpdated(address(_reimbursementPool));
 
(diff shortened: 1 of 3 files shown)


solidity/random-beacon/contracts/Reimbursable.sol
```diff
@@ -28,9 +27,13 @@ abstract contract Reimbursable is Ownable {
         reimbursementPool.refund(gasStart - gasleft(), receiver);
     }
 
+    modifier onlyReimbursableAdmin() virtual {
+        _;
+    }
```
pdyraga, line 32: I would hold on until #2860 is merged to see if `RandomBeacon` contract size allows for this kind of flexibility. If not, I would be also fine with having `Reimbursable` extend upgradeable `Ownable`.

dimpar: It looks like the contract size allows for this change. 
`main` branch is currently at `24.394`. I did a little bit of testing in RandomBeacon and after the changes, it would be around `24.220`

I think we should open a draft PR (which we'll need to do anyway if this PR lands in `main`) that branches off of this branch and add the `onlyReimbursableAdmin` modifier and probably `Governable.sol` and see how it impacts the RandomBeacon contract size. Something similar to what we have in https://github.com/keep-network/keep-core/pull/2921 then we can decide if we should merge the changes proposed here or not.

solidity/random-beacon/contracts/Reimbursable.sol
```diff
@@ -28,9 +27,13 @@ abstract contract Reimbursable is Ownable {
         reimbursementPool.refund(gasStart - gasleft(), receiver);
     }
 
+    modifier onlyReimbursableAdmin() virtual {
```
dimpar, line 30: Why not `onlyReimbursableGovernance`? Are we expecting to have a different "admin" other than governance?

pdyraga: I am pretty sure the assumption was that the contract inheriting from `Reimbursable` does not need to inherit from `Governable`. I will leave this thread open so that @nkuba can clarify.

pdyraga (approved):
LGTM. I am going to merge it to unblock the rest of work, especially the work on the RandomBeacon part.

solidity/random-beacon/contracts/Reimbursable.sol
pdyraga: Done in https://github.com/keep-network/keep-core/pull/2947

PR #2926, opened 2022-04-20
WalletRegistry deployed as upgradeable proxy

nkuba:
Depends on https://github.com/keep-network/keep-core/pull/2921

This PR implements the upgradable deployment of the _WalletRegistry_ contract.
The work here is very much based on the PR for _TokenStaking_ upgradable deployment (https://github.com/threshold-network/solidity-contracts/pull/62).

For upgradable deployments, we use [@keep-network/hardhat-helpers](https://github.com/keep-network/hardhat-helpers/) library that uses [openzeppelin-upgrades](https://github.com/OpenZeppelin/openzeppelin-upgrades) plugin under the hood.

## Constructor, Initializer

We moved most of the operations from the constructor to an initializer, which is a [requirement of the upgradable contracts](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#initializers).
The constructor was left to handle the initialization of immutable variables that are included in the contract's bytecode and is safe to use for this purpose.

## Gaps

We added gaps (`__gap`) to reserve storage slots for possible future upgrades. The gaps were added to base contracts and structs that hold data for libraries. There was no need to add gaps to structs that are stored in mappings, as mapping values are [stored in different slots](https://docs.soliditylang.org/en/v0.8.13/internals/layout_in_storage.html#mappings-and-dynamic-arrays).

Read more:
- https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
- https://forum.openzeppelin.com/t/storage-gaps-in-libraries-and-structs/28171

## Gas Usage, Refunds

Since the upgradable proxy pattern adds to the transaction's gas consumption, we had to tweak the values related to gas offsets. 

We couldn't use `refundable` modifier for `notifySeedTimeout` and `notifyDkgTimeout`, as the `staticGas` that is defined for `ReimbursementPool` doesn't cover the increased cost due to the proxy calls. Hence we added offset for `notifySeedTimeout`. 
For `notifyDkgTimeout` the situation is more interesting, as this transaction recovers more gas by cleaning up stored data with dkg complete. To not refund the caller too much, we subtract the gas that is expected to be recovered with the cleanup.

Additionally, we had to increase the Random Beacon's callback gas limit, to cover the callback execution with a proxy contract. 

## Initializable

We use OpenZeppelin's implementation of the `Initializable` contract. The latest contract version exposes `reinitializer` function that should be used in new versions of the upgradable contract. Unfortunately, this feature is available right now only in an RC package `v4.6.0-rc.0`. We use this package version and hope that a final version is released by the time we deploy our contracts.

## ProxyAdmin

`ProxyAdmin` is a contract that is used to manage upgrades of an upgradable proxy contract. The deployments scripts were updated to transfer the ownership of the `ProxyAdmin` to the ESDM (Emergency Security Developer Multisig). In the future, once a DAO is established we want to replace `ProxyAdmin` with a [`ProxyAdminWithDeputy`](https://github.com/threshold-network/solidity-contracts/blob/main/contracts/governance/ProxyAdminWithDeputy.sol).


solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -99,8 +104,8 @@ contract WalletRegistry is
 
     // External dependencies
 
-    SortitionPool public immutable sortitionPool;
-    IStaking public immutable staking;
+    SortitionPool public sortitionPool;
+    IStaking public staking;
```
cygnusv, line 108: If you want, you can still use immutable variables, as well as using a constructor to initialize them. See the `constructorArgs` option in https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades

nkuba: Thanks for the comment! Initially, I was thinking that using both constructor and initializer will be too confusing, so I decided to use just one. But to be honest upgradable contracts are already complex and confusing so this constructor+initializer won't add much complexity 😉 
I updated the code in 35d49261b56857c35df4830cb7f6cc7de11d77f0

dimpar:
We have more test failures than what CI reports. I was able to see them when I removed `yarn.lock` locally and rebuilt everything.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -244,20 +251,29 @@ contract WalletRegistry is
...
-        sortitionPool = _sortitionPool;
-        staking = _staking;
+    ) external initializer {
         randomBeacon = _randomBeacon;
         reimbursementPool = _reimbursementPool;
 
         // TODO: revisit all initial values
 
+        dkgResultSubmissionGas = 272_000;
+        dkgResultApprovalGasOffset = 63_000;
```
dimpar, line 274: I did some testing and I think we should actually increase it to `79k`. As a rule of thumb it is better for callers to be slightly net-positive rather than net-negative. If we lower it to `63k` then we have 3 tests failing and the callers would have to pay around 0.003ETH when the gas price is 200 gwei. However, if we increase this gas offset to 79k, then in a couple of scenarios we would need to increase the payout in `WalletRegistry.WalletCreation.test.ts`:
- line: 1784 from 0.002 to 0.0045
- line: 1636 from 0.001 to 0.0035

So it's a tradeoff, but we want to avoid for callers to be net-negative.

nkuba: b1b364fbca8f46fdca935d745c50608a606e1648

nkuba: In 8a49ccf7b41dca75aa36c2ba9bafbfb0c7601a87 I added verification of gas usage for transactions. It showed that value for `submitDkgResult` could be higher as the tests for `submitDkgResult` have higher gas consumption than `272 000`.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -244,20 +251,29 @@ contract WalletRegistry is
...
-    ) {
-        sortitionPool = _sortitionPool;
-        staking = _staking;
+    ) external initializer {
         randomBeacon = _randomBeacon;
         reimbursementPool = _reimbursementPool;
 
         // TODO: revisit all initial values
 
+        dkgResultSubmissionGas = 272_000;
```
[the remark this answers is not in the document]
nkuba: I like the underscores, as they improve the code readability very much. Especially where we've got millions and thousands for gas parameters it reduces confusion. I updated just a bit of the code, we can change the rest in another PR or when touching the related places.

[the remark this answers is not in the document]
pdyraga: We need to update all the other places in the code to follow underscore notation. I was planning to do it as a part of https://github.com/keep-network/keep-core/issues/2933 and https://github.com/keep-network/keep-core/issues/2932.

solidity/ecdsa/deploy/02_deploy_reimbursement_pool.ts
```diff
@@ -3,11 +3,11 @@ import type { HardhatRuntimeEnvironment } from "hardhat/types"
 import type { DeployFunction } from "hardhat-deploy/types"
 
 const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
-  const { getNamedAccounts, deployments } = hre
+  const { getNamedAccounts, ethers, deployments } = hre
   const { deployer } = await getNamedAccounts()
 
-  const staticGas = 40800 // gas amount consumed by the refund() + tx cost
-  const maxGasPrice = 500000000000 // 500 gwei
+  const staticGas = 48_000 // gas amount consumed by the refund() + tx cost
```
dimpar, line 9: Why change to 48k? `40800` was chosen as an average gas amount for the `ReimbursementPool.refund()` function. If we run the tests in `ReimbursmentPool.test.ts` and look at the returned transaction `gasUsed` from the `refund()` then the gas used is between 40433 - 41893. The higher-end is when on the rare occasion something goes south with the refund. Having said that, 48k seems to be too high. We want callers to break even as much as we can or make them just slightly net-positive, but not incentivize them with additional ETH from the Reimbursement pool.
Also, changing it to 48k. breaks around 10 tests in the `random-beacon`. We need to be careful with this number, cause we have one Reimbursement pool for `random-beacon`, `ecdsa` and possibly for other contracts like maintainer proxy.
I think we should leave it as-is and tweak the gas offsets instead.

solidity/ecdsa/contracts/libraries/EcdsaAuthorization.sol
```diff
@@ -31,18 +31,30 @@ library EcdsaAuthorization {
         // authorization decrease amount is small, significant, or if it is
         // a decrease to zero.
         uint64 authorizationDecreaseDelay;
+        // Reserved storage space in case we need to add more variables.
+        // See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+        // slither-disable-next-line unused-state
+        uint256[18] __gap;
```
nkuba, line 37: @cygnusv I'm curious about your opinion, whether it makes sense to add gaps to structs that are used as a storage for libraries?

nkuba: Posted a question on OpenZeppelin's forum - [link](https://forum.openzeppelin.com/t/storage-gaps-in-libraries-and-structs/28171).

solidity/ecdsa/deploy/02_deploy_reimbursement_pool.ts
nkuba: I changed the value as we've got a very simple function `notifySeedTimeout` that uses `refundable` modifier:
https://github.com/keep-network/keep-core/blob/a98a1e4d38c3c4c3093284663f6a650101a54a0f/solidity/ecdsa/contracts/WalletRegistry.sol#L688-L690

If I leave `staticGas = 40800` a test for refund fails as the submitter's balance is negative it requires bumping the static gas up to around 4800 to let the test pass.

`notifySeedTimeout` is a very basic function, does it make sense to add an offset to it? Calling the function via proxy probably added the ~7k gas. If we don't need to modify the `staticGas = 40800`, then we would need to add offset to each function called in an upgradable contract.

dimpar: > Calling the function via proxy probably added the ~7k gas

ok, now it adds up. This is specific for `WalletRegistry`, and `ReimbursementPool` is an independent contract that doesn't care about the details of the callee. It will only reimburse for the gas that it consumes, and if a caller contract uses more gas on top, then it should be passed to the `refund` function. This way any contract can count on `ReimbursementPool` for the almost exact refund.

nkuba: Please see b5a155457096170982f028e0d683cbd2eb2c9973

solidity/ecdsa/test/WalletRegistry.Deployment.test.ts
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions */
+import { deployments, ethers, upgrades, helpers } from "hardhat"
```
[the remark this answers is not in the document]
nkuba: 112adbc614285c7427dc3939a802343075e14c76

solidity/ecdsa/deploy/09_transfer_proxy_admin_ownership.ts
```diff
@@ -0,0 +1,24 @@
...
+  const { ethers, getNamedAccounts, upgrades, deployments } = hre
+  const { esdm } = await getNamedAccounts()
+  const { deployer } = await ethers.getNamedSigners()
+
+  deployments.log(`transferring ProxyAdmin ownership to ${esdm}`)
+
+  // TODO: Once a DAO is established we want to switch to ProxyAdminWithDeputy and
+  // use the DAO as the proxy admin owner and ESDM as the deputy. Until then we
+  // use ESDM as the owner of ProxyAdmin contract.
+  const newProxyAdminOwner = esdm
```
nkuba, line 15: @pdyraga is this correct? Who should be responsible for upgrading the WalletRegistry contract?

pdyraga: We need to discuss it with the community but I would like to avoid transferring the ownership to DAO until the system is fully launched and confirmed to be stable. In the first weeks or months, this has to be developer-owned. I think it's fine to assume this will be ESDM initially.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -78,15 +84,28 @@ contract WalletRegistry is
...
+    ///         current market conditions.
+    uint256 internal _notifySeedTimeoutGasOffset;
+
+    /// @notice Gas that is meant to balance the notification of a DKG protocol
+    ///         execution timeout. It can be updated by the governance based on the
+    ///         current market conditions.
+    /// @dev The value is subtracted for the refundable gas calculation, as the
+    ///      DKG timeout notification transaction recovers some gas when cleaning
+    ///      up the storage.
+    uint256 internal _notifyDkgTimeoutNegativeGasOffset;
```
pdyraga, line 107: It could - theoretically - happen that after some hardfork, and opcode pricing changes, we will need a positive offset so it could make sense to use a signed integer. On the other hand, this is:
1. unlikely to happen
2. DKG timeout is not called often - hopefully never
3. we can always upgrade this contract

nkuba: That's a great catch! I also like that in the same message you provided reasoning for not changing the code. Lazy me appreciates that!

solidity/ecdsa/contracts/libraries/EcdsaDkg.sol
```diff
@@ -71,6 +75,10 @@ library EcdsaDkg {
         bytes32 submittedResultHash;
         // Block number from the moment of the DKG result submission.
         uint256 submittedResultBlock;
+        // Reserved storage space in case we need to add more variables.
+        // See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+        // slither-disable-next-line unused-state
+        uint256[11] __gap;
```
pdyraga, line 81: Continuing the thread about gaps: Everything in `Data`, including`Parameters`, take 12 storage slots. I think we should have `uint256[38] __gap;` but only at the end of `Data`. I would remove the gap in `Parameters` and in case we need to add more parameters, I would solve it by `ParametersV2` struct. I think this approach is more flexible because we have the freedom to use our 38 storage slots in any way we want.

solidity/ecdsa/contracts/libraries/Wallets.sol
```diff
@@ -22,12 +22,19 @@ library Wallets {
...
     }
 
     struct Data {
         // Mapping of keccak256 hashes of wallet public keys to wallet details.
         // Hash of public key is considered an unique wallet identifier.
         mapping(bytes32 => Wallet) registry;
+        // Reserved storage space in case we need to add more variables.
+        // See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+        // slither-disable-next-line unused-state
+        uint256[19] __gap;
```
pdyraga, line 37: 49?

pdyraga: I tested it, and what is interesting, switching to 50-element gaps and eliminating them from parameter structs, reduces the contract size.

Now vs altered:
```
 |  WalletRegistry               ·     24.357  │
 |  WalletRegistry               ·     24.196  │
```

pdyraga (commented):
More to come.

solidity/ecdsa/contracts/test/upgrades/WalletRegistryV2.sol
```diff
@@ -0,0 +1,1174 @@
...
+        staking = _staking;
+    }
+
+    /// @dev Initializes upgradable contract on deployment.
+    function initialize(
+        DKGValidator _ecdsaDkgValidator,
+        IRandomBeacon _randomBeacon,
+        ReimbursementPool _reimbursementPool
+    ) external initializer {
+        // TEST: Removed initializer of V1
```
pdyraga, line 288: We do not need to remove it, right? The `initializer` should protect against calling this function another time. Are we trying to get some bytecode back this way? Could be worth dropping a comment on why it is removed.

solidity/ecdsa/contracts/test/upgrades/WalletRegistryV2.sol
```diff
@@ -0,0 +1,1174 @@
+// This is a contract used to test WalletRegistry upgradeability. It is a copy
+// of WalletRegistry contract with some differences marked with `TEST:` comments.
```
pdyraga, line 2: Non-critical but there were some changes in `WallegRegistry` along the way:
- `authorizationParameters()` function is missing
- `authorizationDecreaseDelay()` is still here but was removed in the original implementation
- `updateAuthorizationParameters` is missing two new parameters
- `AuthorizationParametersUpdated` event is missing a parameter
- `_dkgResultSubmissionGas` field doc lost one line for some reason

Would be great to sync them again at the end.

Same comment to two other copies of `WalletRegistry.sol`.

solidity/ecdsa/contracts/test/upgrades/WalletRegistryV2.sol
```diff
@@ -0,0 +1,1139 @@
+// This is a contract used to test WalletRegistry upgradeability. It is a copy
```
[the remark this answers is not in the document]
pdyraga: I personally like it here. The first line attracts attention.

solidity/ecdsa/contracts/test/upgrades/WalletRegistryV2Invalid.sol
```diff
@@ -0,0 +1,1189 @@
+// This is a contract used to test WalletRegistry upgradeability. It is a copy
```
pdyraga, line 1: Do you have an idea of a better name for this file? `WalletRegistryV2MissingSlot.sol` is also *Invalid* and it would be great to be explicit about what exactly is *invalid* in this file.

Some ideas:
- `WalletRegistryV2TooEarlySlot.sol`
- `WalletRegistryV2MisplacedNewSlot.sol`

solidity/ecdsa/contracts/test/RandomBeaconStub.sol
```diff
@@ -7,14 +7,14 @@ import "@keep-network/random-beacon/contracts/api/IRandomBeaconConsumer.sol";
...
-//  This implementation is used to test callback's gas limit only. In most tests
+// This implementation is used to test callback's gas limit only. In most tests
 // we use smock's FakeContract of IRandomBeacon.
 contract RandomBeaconStub is IRandomBeacon {
     using Callback for Callback.Data;
     Callback.Data internal callback;
 
     // This value has to reflect the one set in the Random Beacon contract!
-    uint256 public callbackGasLimit = 56000;
+    uint256 public callbackGasLimit = 64000;
```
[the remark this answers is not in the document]
pdyraga: Ah, I see we changed it in this PR. Is it because the callback is executed via proxy and it's more expensive? 8k increase is a lot 🤔

solidity/ecdsa/deploy/09_transfer_proxy_admin_ownership.ts
pdyraga: Reading it one more time, yes - the comment in TODO is correct. By the time we deploy tBTC v2 the DAO should be fully deployed including [`ProxyAdminWithDeputy`](https://github.com/threshold-network/solidity-contracts/blob/main/contracts/governance/ProxyAdminWithDeputy.sol). `ProxyAdminWithDeputy` should give us the functionalities we need.

solidity/ecdsa/deploy/10_deploy_proxy_admin_with_deputy.ts
```diff
@@ -0,0 +1,27 @@
+import type { HardhatRuntimeEnvironment } from "hardhat/types"
```
pdyraga, line 1: I think we could theoretically reuse `ProxyAdminWithDeputy` we will use for the staking contract, because there is no state in the `ProxyAdmin` specific to proxy or implementation contract. On the other hand, it feels cleaner to have a separate admin per proxy contract. In case we deploy another application contract with another type of proxy, we may need a separate `ProxyAdminWithDeputy` anyway. See https://github.com/threshold-network/solidity-contracts/pull/21.

cc @cygnusv

solidity/ecdsa/deploy/10_deploy_proxy_admin_with_deputy.ts
```diff
@@ -0,0 +1,27 @@
...
+import type { DeployFunction } from "hardhat-deploy/types"
+import type { ProxyAdmin } from "../typechain"
+
+const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
+  const { ethers, getNamedAccounts, upgrades, deployments } = hre
+  const { esdm } = await getNamedAccounts()
+
+  const ProxyAdminWithDeputy = await deployments.get("ProxyAdminWithDeputy")
```
pdyraga, line 9: I am confused. Is this script deploying or trying to fetch an already deployed instance? The script is named `10_deploy*` so I'd expect to see `await deployments.deploy` instead of `awaits deployments.get`.

solidity/ecdsa/hardhat.config.ts
```diff
@@ -74,7 +74,9 @@ const config: HardhatUserConfig = {
       // we use higher gas price for tests to obtain more realistic results
       // for gas refund tests than when the default hardhat ~1 gwei gas price is
       // used
-      gasPrice: 200000000000, // 200 gwei
+      gasPrice: 200000000000, // 200 gwei,
+      // Ignore contract size on test deployment to hardhat network.
```
pdyraga, line 78: Let's make it explicit in the comment size of which contracts we are ignoring. We are not interested in ignoring the size of all contracts, just staking (already deployed, problem solved by decreasing the number of the optimizer runs) and test contracts.

solidity/ecdsa/test/WalletRegistry.Inactivity.test.ts
```diff
@@ -176,7 +189,8 @@ describe("WalletRegistry - Inactivity", () => {
                         subsequentInactiveMembersIndices,
                         groupThreshold,
                         (signatures) => signatures,
-                        (signingMembersIndices) => signingMembersIndices
+                        (signingMembersIndices) => signingMembersIndices,
+                        1_210_000
```
pdyraga, line 193: It is not entirely clear to me why we assert the amount of gas used. Do we want to make sure this does not get too expensive? What about other functions? Or maybe there is some other reason?

solidity/ecdsa/test/fixtures/index.ts
```diff
@@ -60,7 +60,7 @@ export const walletRegistryFixture = deployments.createFixture(
     operators: Operator[]
     reimbursementPool: ReimbursementPool
   }> => {
-    await deployments.fixture(["WalletRegistry"])
+    await deployments.fixture(["WalletRegistryTransferGovernance"])
```
pdyraga, line 63: This will execute all deployments steps up to 08_, included. Correct? If so, why don't we execute up to 09_, included?

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
```diff
@@ -2714,6 +2716,288 @@ describe("WalletRegistryGovernance", async () => {
...
+        tx = await walletRegistryGovernance
+          .connect(governance)
+          .beginNotifySeedTimeoutGasOffsetUpdate(100)
+      })
+
+      after(async () => {
+        await restoreSnapshot()
+      })
+
+      it("should not update the operator inactivity gas offset", async () => {
```
pdyraga, line 2745:
```suggestion
      it("should not update the notify seed timeout gas offset", async () => {
```

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
```diff
@@ -2714,6 +2716,288 @@ describe("WalletRegistryGovernance", async () => {
...
+          tx = await walletRegistryGovernance
+            .connect(governance)
+            .finalizeNotifySeedTimeoutGasOffsetUpdate()
+        })
+
+        after(async () => {
+          await restoreSnapshot()
+        })
+
+        it("should update the operator inactivity gas offset", async () => {
```
pdyraga, line 2836:
```suggestion
        it("should update the notify seed timeout gas offset", async () => {
```

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
```diff
@@ -2714,6 +2716,288 @@ describe("WalletRegistryGovernance", async () => {
...
+        tx = await walletRegistryGovernance
+          .connect(governance)
+          .beginNotifyDkgTimeoutNegativeGasOffsetUpdate(100)
+      })
+
+      after(async () => {
+        await restoreSnapshot()
+      })
+
+      it("should not update the operator inactivity gas offset", async () => {
```
pdyraga, line 2886:
```suggestion
      it("should not update the notify DKG timeout negative gas offset", async () => {
```

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
```diff
@@ -2714,6 +2716,288 @@ describe("WalletRegistryGovernance", async () => {
...
+          tx = await walletRegistryGovernance
+            .connect(governance)
+            .finalizeNotifyDkgTimeoutNegativeGasOffsetUpdate()
+        })
+
+        after(async () => {
+          await restoreSnapshot()
+        })
+
+        it("should update the operator inactivity gas offset", async () => {
```
pdyraga, line 2977:
```suggestion
        it("should update the notify DKG timeout negative gas offset", async () => {
```

solidity/ecdsa/test/WalletRegistry.Upgrade.test.ts
```diff
@@ -0,0 +1,336 @@
...
+
+          await expect(
+            upgradeProxy("WalletRegistry", "WalletRegistryV2Invalid", {
+              factoryOpts: {
+                libraries: { EcdsaInactivity: EcdsaInactivity.address },
+                signer: proxyAdminOwner,
+              },
+              proxyOpts: {
+                constructorArgs: [AddressZero, AddressZero],
+                unsafeAllowLinkedLibraries: true,
```
pdyraga, line 46: If we decide to change to `unsafeAllow` (see my comment to `04_deploy_wallet_registry.ts`), then let's also remember to update this file. It's best to have the calls 1:1, just in case.

solidity/ecdsa/test/WalletRegistry.Upgrade.test.ts
```diff
@@ -0,0 +1,336 @@
...
+              },
+            })
+          ).to.be.rejectedWith(
+            Error,
+            "Deleted `_maliciousDkgResultNotificationRewardMultiplier`"
+          )
+        })
+      })
+    })
```
pdyraga, line 74: Non-blocking this PR, but I would love to see similar tests with a library.

I know, it should work the same. But just in case.

solidity/ecdsa/test/WalletRegistry.Upgrade.test.ts
```diff
@@ -0,0 +1,336 @@
...
+        expect(newWalletRegistry.address).equal(walletRegistry.address)
+      })
+
+      it("should not update governance", async () => {
+        expect(await walletRegistry.governance()).equal(
+          walletRegistryGovernance.address
+        )
+      })
+
+      it("should reinitialize existing immutable variable", async () => {
```
pdyraga, line 135: I think the name of this test is not correct. The value of an immutable variable is stored in the bytecode of the implementation contract. Nothing in the storage of the proxy or implementation contract is updated. Especially, nothing is reinitialized (`initializer` / `reinitializer` modifiers).

Maybe: "should use the new value of the immutable variable"?

pdyraga: To add more context, the very next test name is correct because the random beacon address is passed as a parameter to `initializeV2` function which is `reinitializer(2)`.

solidity/ecdsa/test/WalletRegistry.Upgrade.test.ts
pdyraga: Oh yes, I now see a TODO at line 287! ❤️

solidity/ecdsa/test/WalletRegistry.Upgrade.test.ts
```diff
@@ -0,0 +1,336 @@
...
+      })
+
+      it("should execute updated function logic", async () => {
+        await expect(walletRegistry.notifyDkgTimeout()).to.be.revertedWith(
+          "nice try, but no"
+        )
+      })
+    })
+
+    describe("when a contract gets upgraded during DKG", () => {
```
pdyraga, line 181: 🚀 👌🏻

solidity/ecdsa/contracts/test/RandomBeaconStub.sol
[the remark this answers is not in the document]
nkuba: Yes, exactly what you wrote.

solidity/ecdsa/contracts/libraries/EcdsaDkg.sol
```diff
@@ -46,6 +46,10 @@ library EcdsaDkg {
         // approve it. Once this period ends and the submitter have not approved
```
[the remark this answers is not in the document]
nkuba: c13fc04abfc89b2b42f5dbe1615743a406a70db7

solidity/ecdsa/contracts/libraries/EcdsaInactivity.sol
```diff
@@ -52,6 +52,8 @@ library EcdsaInactivity {
         // in range [1, groupMembers.length], unique, and sorted in ascending
```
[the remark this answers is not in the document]
nkuba: c13fc04abfc89b2b42f5dbe1615743a406a70db7

solidity/ecdsa/deploy/04_deploy_wallet_registry.ts
```diff
@@ -24,35 +25,38 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
...
-      EcdsaInactivity: EcdsaInactivity.address,
+    factoryOpts: {
+      signer: await ethers.getSigner(deployer),
+      libraries: {
+        EcdsaInactivity: EcdsaInactivity.address,
+      },
+    },
+    proxyOpts: {
+      constructorArgs: [SortitionPool.address, TokenStaking.address],
+      unsafeAllowLinkedLibraries: true,
```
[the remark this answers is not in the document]
nkuba: 8ef9325a39f175182163c51d3d43feec8bf0948f

solidity/ecdsa/test/WalletRegistry.Inactivity.test.ts
nkuba: We use them to track gas usage changes, to bring our attention in case the gas usage changes significantly. It is also helpful when estimating gas offsets for reimbursements configuration.

solidity/ecdsa/test/fixtures/index.ts
nkuba: We can actually skip a tag reference and just do `await deployments.fixture()` to get a snapshot of the whole deployment.

nkuba: ae8e9f6dd680a5cf1d89206485fd3d0b6d0e31b4

solidity/ecdsa/test/WalletRegistry.Upgrade.test.ts
nkuba: 8ef9325a39f175182163c51d3d43feec8bf0948f

solidity/ecdsa/contracts/libraries/EcdsaAuthorization.sol
```diff
@@ -71,6 +71,10 @@ library EcdsaAuthorization {
         //                   authorizationDecreaseDelay
         //
         uint64 authorizationDecreaseChangePeriod;
+        // Reserved storage space in case we need to add more variables.
+        // See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+        // slither-disable-next-line unused-state
+        uint256[17] __gap;
```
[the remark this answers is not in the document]
nkuba: I decided on 20 as 50 seemed way too much for our structures. I can make it 50, for consistency and to reduce confusion. No problemo!

I don't like skipping a gap for `Parameters`. Why should we create a `ParametersV2` in the future as we could just use a gap that we planned ahead? On the other hand that would mean we have 50 (in ParametersV2) + 50 (in Data) slots reserved, which also doesn't make sense.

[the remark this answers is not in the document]
nkuba: e74e17d368484b95abbe2c2d4ae4b6a57d97a2b7

solidity/ecdsa/contracts/libraries/EcdsaDkg.sol
nkuba: e74e17d368484b95abbe2c2d4ae4b6a57d97a2b7

nkuba: Let's not include what's inside the Parameters structure to calculate a gap for `Data`, same as you calculated a gap for `EcdsaAuthorization` in https://github.com/keep-network/keep-core/pull/2926#discussion_r862463999


So here we would have 1 slot for Parameters and 8 for other elements, so the gap would be `41`.

solidity/ecdsa/contracts/libraries/Wallets.sol
nkuba: e0f9ea4fb0d90a13b88e6de5038e5e75fcb51e38

solidity/ecdsa/contracts/libraries/EcdsaAuthorization.sol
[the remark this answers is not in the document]
pdyraga: > I don't like skipping a gap for Parameters. Why should we create a ParametersV2 in the future as we could just use a gap that we planned ahead? On the other hand that would mean we have 50 (in ParametersV2) + 50 (in Data) slots reserved, which also doesn't make sense.

Personally, I like none of it. It is choosing the lesser evil, having *some* well-defined pattern and sticking to it. None of it is ideal. Having one gap, always adding up to 50, at the end of the storage space, is the simplest what we can do and the least error-prone. We will avoid holes inside library storage at least.

solidity/ecdsa/contracts/libraries/EcdsaDkg.sol
pdyraga: I did include what's in `Parameters` structure to calculate a gap for `EcdsaAuthorization`. `EcdsaAuthorization.Parameters` occupy one storage slot because storage is tightly packed inside a structure. We have 224 bits there, so it's one EVM word. We even have a space for another `uint32` there.

I do think it makes sense to include what's inside `Parameters`. This way, each library has 50 storage slots consistently.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -244,15 +267,20 @@ contract WalletRegistry is
...
 
-    constructor(
-        SortitionPool _sortitionPool,
-        IStaking _staking,
+    /// @dev Used to initialize immutable variables only, use `initialize` function
+    ///      for upgradable contract initialization on deployment.
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor(SortitionPool _sortitionPool, IStaking _staking) {
+        sortitionPool = _sortitionPool;
+        staking = _staking;
```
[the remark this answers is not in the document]
nkuba: 81cf5547f7654b08244ba8290994d42ea2e17827

solidity/ecdsa/contracts/test/upgrades/WalletRegistryV2.sol
nkuba: Yeah, I think we could even remove the `initialize` function completely.

nkuba: b43df059dbd52682c5e8ce8f42467afb2be3e358

solidity/ecdsa/contracts/test/RandomBeaconStub.sol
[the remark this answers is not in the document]
pdyraga: Oh my. I had to test it myself, and indeed, even 62k is no longer enough.

solidity/ecdsa/contracts/test/upgrades/WalletRegistryV2Invalid.sol
nkuba: 6212e48870a16033e4917476a03dda1bb25afd19

solidity/ecdsa/contracts/test/upgrades/WalletRegistryV2.sol
nkuba: 52b0dd92ddda423c56c3ebddebabba913499ea98

solidity/ecdsa/deploy/10_deploy_proxy_admin_with_deputy.ts
nkuba: Good catch, this had to be a mistake.

solidity/ecdsa/deploy/10_deploy_proxy_admin_with_deputy.ts
nkuba: 5bfc5285996d172a5bb03f1cc2beb91eda60f163

solidity/ecdsa/deploy/10_deploy_proxy_admin_with_deputy.ts
nkuba: 5bfc5285996d172a5bb03f1cc2beb91eda60f163

solidity/ecdsa/contracts/libraries/EcdsaDkg.sol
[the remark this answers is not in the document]
nkuba: 3818e9e144d8f8749a3da1d0bf992895c043a093

solidity/ecdsa/contracts/libraries/EcdsaInactivity.sol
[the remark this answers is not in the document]
nkuba: 3818e9e144d8f8749a3da1d0bf992895c043a093

solidity/ecdsa/contracts/test/upgrades/WalletRegistryV2.sol
```diff
@@ -0,0 +1,1174 @@
...
+        require(governance == msg.sender, "Caller is not the governance");
+        _;
+    }
+
+    /// @dev Used to initialize immutable variables only, use `initialize` function
+    ///      for upgradable contract initialization on deployment.
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor(SortitionPool _sortitionPool, IStaking _staking) {
+        sortitionPool = _sortitionPool;
+        staking = _staking;
```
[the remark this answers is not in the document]
nkuba: 52b0dd92ddda423c56c3ebddebabba913499ea98

solidity/ecdsa/contracts/test/upgrades/WalletRegistryV2.sol
nkuba: 52b0dd92ddda423c56c3ebddebabba913499ea98

solidity/ecdsa/deploy/04_deploy_wallet_registry.ts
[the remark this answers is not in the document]
nkuba: 8ef9325a39f175182163c51d3d43feec8bf0948f

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
nkuba: 4305d2984b3d2b6de204eb71a6dd43c1dbe55691

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
nkuba: 4305d2984b3d2b6de204eb71a6dd43c1dbe55691

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
nkuba: 4305d2984b3d2b6de204eb71a6dd43c1dbe55691

solidity/ecdsa/test/WalletRegistryGovernance.test.ts
nkuba: 4305d2984b3d2b6de204eb71a6dd43c1dbe55691

solidity/ecdsa/contracts/libraries/EcdsaDkg.sol
nkuba: > I did include what's in Parameters structure to calculate a gap for EcdsaAuthorization. EcdsaAuthorization.Parameters occupy one storage slot because storage is tightly packed inside a structure. We have 224 bits there, so it's one EVM word. We even have a space for another uint32 there.

Wow, you're even more clever than I thought! 😉 

But.. I don't like getting that much into detail to calculate the gap. Should we analyze the type of each element in the struct? What about [mappings and dynamic arrays](https://docs.soliditylang.org/en/v0.8.13/internals/layout_in_storage.html#mappings-and-dynamic-arrays) - should they be included when calculating the gap or not? IMHO it looks like we're overthinking it.

I can agree to include elements of `Parameters` struct in the gap in `Data`, but each element with the same wage, no matter how they are packed in slots, so `EcdsaAuthorization.Parameters` will subtract `3` from the gap.

nkuba: And we should subtract mappings from the gap as well. Each mapping is `- 1`. Just to simplify the operation of calculating the gap.

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -669,18 +697,118 @@ contract WalletRegistryGovernance is Ownable {
...
+        newNotifyDkgTimeoutNegativeGasOffset = _newNotifyDkgTimeoutNegativeGasOffset;
+        notifyDkgTimeoutNegativeGasOffsetChangeInitiated = block.timestamp;
+        emit NotifyDkgTimeoutNegativeGasOffsetUpdateStarted(
+            _newNotifyDkgTimeoutNegativeGasOffset,
+            block.timestamp
+        );
+        /* solhint-enable not-rely-on-time */
+    }
+
+    /// @notice Finalizes the notify DKG timeout gas offset update process.
```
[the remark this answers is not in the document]
nkuba: 55a570255c06107468171388011682f71a1d69fc

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -669,18 +697,118 @@ contract WalletRegistryGovernance is Ownable {
...
+            notifyDkgTimeoutNegativeGasOffset
+        );
+        notifySeedTimeoutGasOffsetChangeInitiated = 0;
+        newNotifySeedTimeoutGasOffset = 0;
+    }
+
+    /// @notice Begins the notify DKG timeout gas offset update process.
+    /// @dev Can be called only by the contract owner.
+    /// @param _newNotifyDkgTimeoutNegativeGasOffset New DKG timeout gas notification
+    ///        gas offset
```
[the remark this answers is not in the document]
nkuba: 55a570255c06107468171388011682f71a1d69fc

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -1131,6 +1259,32 @@ contract WalletRegistryGovernance is Ownable {
...
+    function getRemainingNotifySeedTimeoutGasOffsetUpdateTime()
+        external
+        view
+        returns (uint256)
+    {
+        return
+            getRemainingChangeTime(notifySeedTimeoutGasOffsetChangeInitiated);
+    }
+
+    /// @notice Get the time remaining until the DKG timeout gas offset
```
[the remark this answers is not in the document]
nkuba: 55a570255c06107468171388011682f71a1d69fc

solidity/ecdsa/hardhat.config.ts
nkuba: 55a570255c06107468171388011682f71a1d69fc

solidity/ecdsa/test/WalletRegistry.Upgrade.test.ts
nkuba: 96cae7abe3b6d5b63e53817845ecb5a2a10c641e

solidity/ecdsa/contracts/libraries/EcdsaDkg.sol
pdyraga: > But.. I don't like getting that much into detail to calculate the gap. Should we analyze the type of each element in the struct?

Yes, this is giving us gas savings when reading and writing. That's why we do not use `uint256` everywhere but go down to `uint64` and `uint32`. We not only need to analyze the type of each element but also how they are ordered so that tightly packing is possible.

> What about [mappings and dynamic arrays](https://docs.soliditylang.org/en/v0.8.13/internals/layout_in_storage.html#mappings-and-dynamic-arrays) - should they be included when calculating the gap or not?

> And we should subtract mappings from the gap as well. Each mapping is - 1. Just to simplify the operation of calculating the gap.

We need to assume each mapping and dynamic array (including a string) occupies one storage slot. This how we calculate the gap in `EcdsaAuthorization`. 

Examples from OpenZeppelin:
- https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v4.5.2/contracts/finance/PaymentSplitterUpgradeable.sol
- https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v4.5.2/contracts/token/ERC777/ERC777Upgradeable.sol
- https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v4.5.2/contracts/token/ERC20/ERC20Upgradeable.sol
- https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v4.5.2/contracts/finance/PaymentSplitterUpgradeable.sol

nkuba: > > But.. I don't like getting that much into detail to calculate the gap. Should we analyze the type of each element in the struct?
> 
> Yes, this is giving us gas savings when reading and writing. That's why we do not use `uint256` everywhere but go down to `uint64` and `uint32`. We not only need to analyze the type of each element but also how they are ordered so that tightly packing is possible.

I know that typing and ordering is crucial for storage slots consumption. What I mean is that we don't need to go into such details when calculating the gap. We could simplify it and assume that each element subtracts 1 from the gap (50). As we did for mappings.

pdyraga: A gap is a collection of 256-bit elements, so it's a collection of EVM words. 

Gap has 50 elements by default, so 50 EVM 256-bit words. 

If we subtract always one element, no matter how much storage space is actually occupied, for `Parameters{uint64, uint64, uint128}`, we would end up with 1 EVM word used by the storage and 47 EVM words in a gap. 1 + 47 != 50, so this is inconsistent. 


> As we did for mappings.

Mapping and array do occupy one storage slot, that's why we subtracted 1 EVM word (or 256 bits). True, those are dynamic data types, and elements are written into random slots. But dynamic data types will always be such special cases and that one storage slot we subtracted is actually occupied. Worth noting, where actually data lands is computed based on the one slot occupied by the element in the storage we are discussing.

nkuba: You won 😸 80895aab2cbac1f8df04b209f7ce64ba01cd93db

pdyraga (approved):

PR #2934, opened 2022-04-20
Random Beacon: restricted function mutability, added zero-address and zero-value checks to constructors

pdyraga:
Restricted mutability of `validateFields` and `validateMembersHash` in
`BeaconDkgValidator` from `view` to `pure`.

Restricted mutability of `validateMembersIndices` in `BeaconInactivity` from
`view` to `pure`.

Added zero-address and zero-value checks to constructors.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -1404,13 +1404,12 @@ contract RandomBeacon is IRandomBeacon, IApplication, Ownable, Reimbursable {
         return sortitionPool.isOperatorInPool(operator);
     }
 
-    /// @notice Selects a new group of operators based on the provided seed.
+    /// @notice Selects a new group of operators. Can only be called when DKG
+    ///         is in progress and the pool is locked.
     ///         At least one operator has to be registered in the pool,
     ///         otherwise the function fails reverting the transaction.
-    /// @param seed Number used to select operators to the group.
     /// @return IDs of selected group members.
-    function selectGroup(bytes32 seed) external view returns (uint32[] memory) {
-        // TODO: Read seed from RandomBeaconDkg
-        return sortitionPool.selectGroup(DKG.groupSize, seed);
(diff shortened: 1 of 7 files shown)


pdyraga:
Adding unit tests, should be ready soon.

solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -369,6 +369,18 @@ contract RandomBeacon is IRandomBeacon, IApplication, Ownable, Reimbursable {
         staking = _staking;
         reimbursementPool = _reimbursementPool;
 
+        require(
+            address(_sortitionPool) != address(0),
```
dimpar, line 373: How about "Zero-address sortition pool", "Zero-address T Token" etc? If a revert happens, then the error message is more descriptive of where exactly it happened.

Same comment in all the places in this PR where we have "Zero-address ..."

solidity/random-beacon/test/RandomBeacon.Constructor.test.ts
```diff
@@ -0,0 +1,145 @@
...
+          BeaconInactivity: inactivity.address,
+        },
+      }
+    )
+  })
+
+  describe("constructor", () => {
+    context("when all passed addresses are valid", () => {
+      it("should work", async () => {
+        await RandomBeacon.deploy(
```
dimpar, line 64: How about 
```
        await expect(
          RandomBeacon.deploy(
            sortitionPool.address,
            tToken,
            staking,
            dkgValidator,
            reimbursementPool
          )
        ).to.not.be.reverted
```

pdyraga: Smart.

solidity/random-beacon/contracts/RandomBeacon.sol
pdyraga: The same message is used to save on the contract size. If we use another messages, we will end up with much more bytecode.

dimpar (approved):

Issue #2935, opened 2022-04-20
Revisit common interface with PRE and other Threshold applications: Random Beacon and ECDSA


pdyraga:
See https://github.com/threshold-network/solidity-contracts/issues/89 and https://github.com/threshold-network/solidity-contracts/issues/88. We need to agree on the interface and adjust (or not) `RandomBeacon` and `WalletRegistry` contracts.

PR #2936, opened 2022-04-21
Random Beacon: read DKG seed from chain instead of passing it

### LINKED ISSUE #2929: RandomBeacon selectGroup should read the seed from DKG library
https://github.com/keep-network/keep-core/blob/62c219ad8fc9d79e4e69710608d5fb7f71038c01/solidity/random-beacon/contracts/RandomBeacon.sol#L1395-L1403

Ideally, off-chain clients should be able to call `RandomBeacon.selectGroup()` without having to worry about the seed value.

pdyraga:
Closes #2929

Instead of requiring operators to pass the value that is already
available on-chain, we just read this value from DKG library storage.

### DIFF
--- a/solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
+++ b/solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
@@ -2721,6 +2721,41 @@ describe("RandomBeacon - Group Creation", () => {
       await restoreSnapshot()
     })
   })
+
+  describe("selectGroup", async () => {
+    context("when dkg was not triggered", async () => {
+      before(async () => {
+        await createSnapshot()
+      })
+
+      after(async () => {
+        await restoreSnapshot()
+      })
+
+      it("should revert", async () => {
+        await expect(randomBeacon.selectGroup()).to.be.revertedWith(
+          "Sortition pool unlocked"
+        )
+      })
+    })
+
+    context("when dkg was triggered", async () => {
+      before(async () => {
+        await createSnapshot()
+
+        await genesis(randomBeacon)
+      })
+
+      after(async () => {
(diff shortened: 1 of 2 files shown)


pdyraga:
I've opened this draft PR so that I can branch off from it for the work on #2935 and see the final contract size. We need to add unit tests, to `selectGroup` function, leaving it up to you @dimpar.

dimpar:
> I've opened this draft PR so that I can branch off from it for the work on #2935 and see the final contract size. We need to add unit tests, to `selectGroup` function, leaving it up to you @dimpar.

Tests are added

solidity/random-beacon/test/RandomBeacon.GroupCreation.test.ts
```diff
@@ -2721,6 +2721,41 @@ describe("RandomBeacon - Group Creation", () => {
...
+      })
+
+      after(async () => {
+        await restoreSnapshot()
+      })
+
+      it("should select a group", async () => {
+        const selectedGroup = await randomBeacon.selectGroup()
+        expect(selectedGroup.length).to.eq(constants.groupSize)
+      })
```
pdyraga, line 2756: Can we also assert it is the same group that would be returned by the sortition pool if we provided there DKG seed manually?

dimpar: [2e95d35](https://github.com/keep-network/keep-core/pull/2936/commits/2e95d354d3be25ae8ece0448ce352e62c2dd80ce)

pdyraga:
LGTM, GitHub does not let me approve this PR so leaving it up to you @dimpar

dimpar (approved):

PR #2937, opened 2022-04-22
Update sortition pools unlocks and DKG complete for notifySeedTimeout

nkuba:
After `notifySeedTimeout` is called we want to unlock the sortition pool and cleanup DKG state. In this PR we make it consistent for both ECDSA and Random Beacon.

`sortitionPool.unlock()` is called as part of `dkg.complete` function.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -1085,6 +1085,7 @@ contract RandomBeacon is IRandomBeacon, IApplication, Ownable, Reimbursable {
             // avoid blocking the future group creation.
             if (dkg.currentState() == DKG.State.AWAITING_SEED) {
                 dkg.notifySeedTimedOut();
+                dkg.complete();
             }
         }
     }
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -777,7 +777,6 @@ contract RandomBeacon is IRandomBeacon, IApplication, Ownable, Reimbursable {
     /// @notice Notifies about DKG timeout.
     function notifyDkgTimeout() external refundable(msg.sender) {
         dkg.notifyTimeout();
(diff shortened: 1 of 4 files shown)


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -1085,6 +1085,7 @@ contract RandomBeacon is IRandomBeacon, IApplication, Ownable, Reimbursable {
             // avoid blocking the future group creation.
             if (dkg.currentState() == DKG.State.AWAITING_SEED) {
                 dkg.notifySeedTimedOut();
+                dkg.complete();
```
pdyraga, line 1088: Why don't we call `complete()` from `notifySeedTimedOut`? Documentation to `BeaconDkg.notifySeedTimedOut` says "Notifies about the seed was not delivered and restores the initial DKG state (IDLE)" but it's in fact doing nothing other than checking the state:

https://github.com/keep-network/keep-core/blob/9675fe9f51cb4c63a73ae55f106a3454b3fbf643/solidity/random-beacon/contracts/libraries/BeaconDkg.sol#L277-L286

pdyraga: Same about `notifyTimeout()`. All occurrences of `dkg.notifyTimeout()` and `dkg.notifySeedTimeout()` in `RandomBeacon` and `WalletRegistry` are followed by `dkg.complete()`.

nkuba: Yeah, I was thinking about it. I'm glad you brought it up! I'll update the code.

nkuba: 91ce31d5b57fb98164be6a891310ad6527afdf05

pdyraga (approved):
Contract size went down (`main` vs this branch):
```
 |  RandomBeacon            ·     24.458  ·               │
 |  RandomBeacon            ·     24.394  ·       -0.064  │
```

Issue #2870, opened 2022-03-15
Unit test for ECDSA DKG challenge and TokenStaking contract failing on slashing


pdyraga:
See [this](https://github.com/keep-network/keep-core/pull/2850#discussion_r822739137) conversation.

When working on https://github.com/keep-network/keep-core/pull/2850 and replacing `StakingStub` with a real T `TokenStaking` for tests, one unit test covering `challangeDkgResult` had to be removed.  `challengeDkgResult` does not revert if slashing in `TokenStaking` does revert. It makes sense because it is better to do not slash than to let malicious DKG result be accepted. Testing this behavior is complicated though. 

From one side we are trying to test with a real `TokenStaking` contract and from the other, we are trying to force it to behave in a way it is never behaving.

Using Smock mocks requires setting up half of the world that is already set up in `index.ts`.
Altering `index.ts` and making it way more complicated for one test feels wrong. 
Replacing `TokenStaking` with some stub extending `TokenStaking` for all tests sucks as well - the entire point is to test against a real contract instead of hoping the inheriting contract does not modify some behavior.

lionakhnazarov:
Closing as stale. Legacy v1 ECDSA/TokenStaking test note, superseded on main

PR #2942, opened 2022-04-22
WalletRegistry.seize function implemented

### LINKED ISSUE #2927: Slashing capabilities for WalletRegistry wallet owner
The wallet owner of `WalletRegistry` should be able to slash/seize tokens. This is needed to implement wallet frauds in tBTC `Bridge`.

May require https://github.com/threshold-network/solidity-contracts/pull/87

pdyraga:
Closes #2927

Allows the wallet owner to add all signing group members of the wallet with the given ID to the slashing queue of the staking contract. The notifier will receive reward per each group member from the staking contract notifiers treasury. The reward is scaled by the `rewardMultiplier` provided as a parameter.

This function will be used by tBTC Bridge contract to slash wallet signing group members.

TODO:
- [ ] Some time ago, hardhat gas reporter plugin stopped providing the report. For this reason, the gas cost of calling the new function is unknown at the moment.
- [ ] `@threshold-network/solidity-contracts` dependency should be updated to version `>1.2.0-dev <1.2.0-ropsten` to include the gas cost fix for `TokenStaking` from https://github.com/threshold-network/solidity-contracts/pull/87. This is blocked by a problem with running tests after the upgrade:
  ```
  Error: ERROR processing /Users/piotr/git/keep-network/keep-core/solidity/ecdsa/node_modules/@threshold-network/solidity-contracts/export/deploy/00_resolve_keep_registry.js:
  TypeError: Cannot read properties of undefined (reading 'length')
  ```

### DIFF
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -774,6 +774,58 @@ contract WalletRegistry is
         );
     }
 
+    /// @notice Allows the wallet owner to add all signing group members of the
+    ///         wallet with the given ID to the slashing queue of the staking .
+    ///         contract. The notifier will receive reward per each group member
+    ///         from the staking contract notifiers treasury. The reward is
+    ///         scaled by the `rewardMultiplier` provided as a parameter.
+    /// @param amount Amount of tokens to seize from each signing group member
+    /// @param rewardMultiplier Fraction of the staking contract notifiers
+    ///        reward the notifier should receive; should be between [0, 100]
+    /// @param notifier Address of the misbehavior notifier
+    /// @param walletID ID of the wallet
+    /// @param walletMembersIDs Identifiers of the wallet signing group members
+    /// @dev Requirements:
+    ///      - The expression `keccak256(abi.encode(walletMembersIDs))` must
+    ///        be exactly the same as the hash stored under `membersIdsHash`
+    ///        for the given `walletID`. Those IDs are not directly stored
+    ///        in the contract for gas efficiency purposes but they can be
+    ///        read from appropriate `DkgResultSubmitted` and `DkgResultApproved`
+    ///        events.
+    ///      - `rewardMultiplier` must be between [0, 100]
+    function seize(
+        uint96 amount,
+        uint256 rewardMultiplier,
+        address notifier,
+        bytes32 walletID,
+        uint32[] calldata walletMembersIDs
+    ) external onlyWalletOwner {
+        bytes32 memberIdsHash = wallets.getWalletMembersIdsHash(walletID);
+        require(
+            memberIdsHash == keccak256(abi.encode(walletMembersIDs)),
+            "Invalid wallet members identifiers"
+        );
+
+        address[] memory groupMembersAddresses = sortitionPool.getIDOperators(
+            walletMembersIDs
+        );
+        address[] memory stakingProvidersAddresses = new address[](
+            walletMembersIDs.length
+        );
+        for (uint256 i = 0; i < groupMembersAddresses.length; i++) {
+            stakingProvidersAddresses[i] = operatorToStakingProvider(
+                groupMembersAddresses[i]
+            );
+        }
+
+        staking.seize(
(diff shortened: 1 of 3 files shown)


solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -774,6 +774,58 @@ contract WalletRegistry is
...
+    /// @param walletMembersIDs Identifiers of the wallet signing group members
+    /// @dev Requirements:
+    ///      - The expression `keccak256(abi.encode(walletMembersIDs))` must
+    ///        be exactly the same as the hash stored under `membersIdsHash`
+    ///        for the given `walletID`. Those IDs are not directly stored
+    ///        in the contract for gas efficiency purposes but they can be
+    ///        read from appropriate `DkgResultSubmitted` and `DkgResultApproved`
+    ///        events.
+    ///      - `rewardMultiplier` must be between [0, 100]
+    function seize(
```
nkuba, line 796: Shouldn't we add this function to the `IWalletRegistry`?

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -774,6 +774,58 @@ contract WalletRegistry is
...
+        address[] memory stakingProvidersAddresses = new address[](
+            walletMembersIDs.length
+        );
+        for (uint256 i = 0; i < groupMembersAddresses.length; i++) {
+            stakingProvidersAddresses[i] = operatorToStakingProvider(
+                groupMembersAddresses[i]
+            );
+        }
+
+        staking.seize(
```
nkuba, line 821: Shouldn't we add a try/catch just like we do for other `staking.seize` and `staking.slash` calls in `WalletRegistry` and `RandomBeacon`? 
If not we should be clear in the documentation that the function may revert and when calling it from the TBTCv2 we should double-check if a revert from the staking contract is acceptable for the use case.

solidity/ecdsa/test/WalletRegistry.Slashing.test.ts
```diff
@@ -0,0 +1,152 @@
...
+            .mul(rewardMultiplier)
+            .div(100)
+
+          const receivedReward = notifierBalanceAfter.sub(notifierBalanceBefore)
+
+          expect(receivedReward).to.equal(
+            perMemberReward.mul(constants.groupSize)
+          )
+        })
+      })
```
nkuba, line 149: Can we add a test case to validate that the transaction reverts if `staking.seize` fails?

solidity/ecdsa/contracts/WalletRegistry.sol
pdyraga: We should. I'll add it in another PR, good catch.

solidity/ecdsa/contracts/WalletRegistry.sol
pdyraga: I wanted to leave this responsibility to the calling code. That should should know best if it wants to revert or not. Agreed, it should be super-clear in the docs. I'll clarify it.

solidity/ecdsa/test/WalletRegistry.Slashing.test.ts
pdyraga: Yes and no - my favourite answer. See #2870.

solidity/ecdsa/contracts/WalletRegistry.sol
pdyraga: https://github.com/keep-network/keep-core/pull/2942/commits/ebde33349de2502cfda46df0073daadd9320de4f

pdyraga:
This PR is ready for another chance @nkuba.

solidity/ecdsa/test/WalletRegistry.Slashing.test.ts
nkuba: Maybe it would be worth adding at least a placeholder so we know that when the referenced issue is fixed we need to implement a test?

nkuba (approved):

PR #2946, opened 2022-04-23
Use Governable contract from random-beacon module

pdyraga:
~~Depends on #2945; keeping it as a draft until #2945 is not merged~~

To do not duplicate exactly the same code, `Governable` is declared next
to `Reimbursable` in `random-beacon` module and we now import it from
`ecdsa` module.

~~There is a temporary workaround in `package.json` to import from
`../random-beacon` instead of the development tag. This will have to be
changed back once #2945 moving files to `random-beacon` gets merged and a
new package gets published. See commit d99ab6057242c156d07784147387b973a9cb8fb8.~~


solidity/ecdsa/package.json
```diff
@@ -62,7 +62,7 @@
     "typescript": "^4.5.4"
   },
   "dependencies": {
-    "@keep-network/random-beacon": "development",
+    "@keep-network/random-beacon": "../random-beacon",
```
nkuba, line 65: Want to leave it as is or set it back to `development` once an updated package is published?

pdyraga: Waiting for a new package to be published, I will switch it in this PR.

pdyraga: Switched back.

pdyraga:
Undrafted @nkuba

nkuba (approved):

PR #2947, opened 2022-04-23
RandomBeacon contract switched to use Governable

pdyraga:
~~Depends on #2946~~

`RandomBeacon` is now fully aligned with `WalletRegistry` when it comes to
how `Governable` and `Reimbursable` abstract contracts are used. This change
also allowed to reduce `RandomBeacon` contract size by a tiny bit.

Before/after:
```
 |  RandomBeacon   ·     24.352  │
 |  RandomBeacon   ·     24.333  │
```


nkuba (approved):

nkuba:
LGTM, slither job is failing, but it should be solved with https://github.com/keep-network/keep-core/pull/2949

Issue #2930, opened 2022-04-20
WalletRegistry selectGroup should read the seed from DKG library


pdyraga:
https://github.com/keep-network/keep-core/blob/62c219ad8fc9d79e4e69710608d5fb7f71038c01/solidity/ecdsa/contracts/WalletRegistry.sol#L956-L964

Ideally, off-chain clients should be able to call `WalletRegistry.selectGroup()` without having to worry about the seed value.

PR #2950, opened 2022-04-25
Reading dkg seed from the DKG library

dimpar:
Closes https://github.com/keep-network/keep-core/issues/2930

DKG seed can be read from the DKG library and there's no need passing it
as a parameter.

### DIFF
--- a/solidity/ecdsa/test/WalletRegistry.WalletCreation.test.ts
+++ b/solidity/ecdsa/test/WalletRegistry.WalletCreation.test.ts
@@ -3455,7 +3455,7 @@ describe("WalletRegistry - Wallet Creation", async () => {
     })
   })
 
-  describe.only("selectGroup", async () => {
+  describe("selectGroup", async () => {
     context("when dkg was not triggered", async () => {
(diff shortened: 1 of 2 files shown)


solidity/ecdsa/test/WalletRegistry.WalletCreation.test.ts
```diff
@@ -3454,6 +3454,52 @@ describe("WalletRegistry - Wallet Creation", async () => {
       })
     })
   })
+
+  describe.only("selectGroup", async () => {
```
pdyraga, line 3458: Oooonlyyyyyyy!

pdyraga (approved):

solidity/ecdsa/test/WalletRegistry.WalletCreation.test.ts
pdyraga: https://github.com/keep-network/keep-core/pull/2950/commits/d9b0278b37f90059c60f3d2afe51314bab396907

PR #2951, opened 2022-04-25
Lowering the size of RandomBeacon contract: parameter function refactoring

pdyraga:
~~Depends on #2947. Keeping as a draft until #2947 is not merged but I plan no changes to the code so it is ready for review!~~

This PR allows to go down with `RandomBeacon` contract size from `24.333 KB` to `24.130 KB`.

This is cleaning up the structure but most importantly, should add enough space for some reward-related functions (to be done in a separate PR).

It is grouping parameters together so that instead of exposing them as separate view functions, they are grouped into:
- `relayEntryParameters()`
- `groupCreationParameters()`
- `rewardParameters()`
- `slashingParameters()`
- `gasParameters()`

No changes in the logic of how `RandomBeacon` works.

I recommend reviewing this PR commit-by-commit.

### DIFF
--- a/solidity/random-beacon/contracts/Governable.sol
+++ b/solidity/random-beacon/contracts/Governable.sol
@@ -20,6 +20,8 @@ pragma solidity ^0.8.9;
 ///      function in a child contract.
 abstract contract Governable {
     // Governance of the contract
+    // The variable should be initialized by the implementing contract.
+    // slither-disable-next-line uninitialized-state
     address public governance;
 
     event GovernanceTransferred(address oldGovernance, address newGovernance);
--- a/solidity/random-beacon/contracts/RandomBeaconGovernance.sol
+++ b/solidity/random-beacon/contracts/RandomBeaconGovernance.sol
@@ -685,12 +685,18 @@ contract RandomBeaconGovernance is Ownable {
         emit SortitionPoolRewardsBanDurationUpdated(
             newSortitionPoolRewardsBanDuration
         );
+        (
+            ,
+            uint256 relayEntryTimeoutNotificationRewardMultiplier,
+            uint256 unauthorizedSigningNotificationRewardMultiplier,
+            uint256 dkgMaliciousResultNotificationRewardMultiplier
+        ) = randomBeacon.rewardParameters();
         // slither-disable-next-line reentrancy-no-eth
         randomBeacon.updateRewardParameters(
             newSortitionPoolRewardsBanDuration,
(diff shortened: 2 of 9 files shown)


solidity/random-beacon/contracts/Governable.sol
```diff
@@ -20,6 +20,8 @@ pragma solidity ^0.8.9;
 ///      function in a child contract.
 abstract contract Governable {
     // Governance of the contract
+    // The variable should be initialized by the implementing contract.
+    // slither-disable-next-line uninitialized-state
```
pdyraga, line 24: cc @nkuba

solidity/random-beacon/contracts/RandomBeaconGovernance.sol
```diff
@@ -484,10 +499,20 @@ contract RandomBeaconGovernance is Ownable {
         onlyAfterGovernanceDelay(groupCreationFrequencyChangeInitiated)
     {
         emit GroupCreationFrequencyUpdated(newGroupCreationFrequency);
+        (
+            ,
+            uint256 groupLifetime,
+            uint256 dkgResultChallengePeriodLength,
+            uint256 dkgResultSubmissionTimeout,
+            uint256 dkgSubmitterPrecedencePeriodLength
+        ) = randomBeacon.groupCreationParameters();
```
beaushinkle, line 508: Ah, this was the point of all of the refactorings in `RandomBeacon.sol`. Rather than call the contract a bunch of time for individual fields, we can call the fields in bundles

pdyraga: Most importantly, this saves us bytecode space. Every public field and function occupies more bytecode than internal field or function. We are very close to the limit and we really need this space for `availableRewards` view function that should be added to Threshold `IApplication`.

dimpar (approved):

solidity/random-beacon/contracts/RandomBeaconGovernance.sol
```diff
@@ -1026,11 +1135,17 @@ contract RandomBeaconGovernance is Ownable {
         emit RelayEntrySubmissionGasOffsetUpdated(
             newRelayEntrySubmissionGasOffset
         );
+        (
+            uint256 dkgResultSubmissionGas,
+            uint256 dkgResultApprovalGasOffset,
+            uint256 notifyOperatorInactivityGasOffset,
+
```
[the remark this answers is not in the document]
pdyraga: Agreed but that's what the linter rules currently enforce. It was not added by me, I even tried to delete it.

PR #2956, opened 2022-04-26
Updated beacon `@threshold-network/solidity-contract` dependency

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/2935

Dependency updated to `>1.2.0-dev <1.2.0-ropsten`, with a fix on slashing
gas costs and changes to `IApplication` interface.

After upgrading the dependency, I had to fix the unit tests for `RandomBeacon`
authorizations to take into account the change of the behavior in the staking contract.
This was done in the commit 9d061aab160e93b7b0a5424b881a3b32ac54a102.

Note that the same fix was done for `WalletRegistry` in https://github.com/keep-network/keep-core/pull/2942.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeacon.sol
+++ b/solidity/random-beacon/contracts/RandomBeacon.sol
@@ -345,8 +345,6 @@ contract RandomBeacon is IRandomBeacon, IApplication, Governable, Reimbursable {
         address indexed operator
     );
 
-    event RewardsWithdrawn(address indexed stakingProvider, uint96 amount);
-
     /// @dev Assigns initial values to parameters to make the beacon work
     ///      safely. These parameters are just proposed defaults and they might
     ///      be updated with `update*` functions after the contract deployment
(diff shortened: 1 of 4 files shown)


solidity/random-beacon/contracts/RandomBeacon.sol
```diff
@@ -345,8 +345,6 @@ contract RandomBeacon is IRandomBeacon, IApplication, Governable, Reimbursable {
         address indexed operator
     );
 
-    event RewardsWithdrawn(address indexed stakingProvider, uint96 amount);
```
nkuba, line 348: Just to confirm, do we really want to remove this event? Why we don't need it anymore?

nkuba (approved):
One non-blocking question

solidity/random-beacon/contracts/RandomBeacon.sol
pdyraga: It is no longer needed because it is [now declared in `IApplication` interface](https://github.com/threshold-network/solidity-contracts/blob/47268908a1aa92ab215b76e5f26730519643b663/contracts/staking/IApplication.sol#L25-L26) and is a part of [`withdrawRewards` function contract](https://github.com/threshold-network/solidity-contracts/blob/47268908a1aa92ab215b76e5f26730519643b663/contracts/staking/IApplication.sol#L31-L32). After updating `@threshold-network/solidity-contracts` dependency, the compilation failed because `RandomBeacon` extends `IApplication` and that event is already defined in `IApplication`. See https://github.com/threshold-network/solidity-contracts/pull/95 please.

Issue #2955, opened 2022-04-26
Fix gas reporter for beacon and ecdsa


pdyraga:
Hardhat gas reporter plugin report is empty for `beacon` and `ecdsa` modules.

For `ecdsa` it was fixed in https://github.com/keep-network/keep-core/pull/2944 but because of some other change, it stopped working again. I *think* the reason could be something around https://github.com/threshold-network/solidity-contracts/pull/92 or other change in `@threshold-network/solidity-contracts` deployment scripts because it stopped working after updating `@threshold-network/solidity-contracts` dependency from `">1.1.0-dev <1.1.0-ropsten"` to `">1.2.0-dev <1.2.0-ropsten"` in #2942.

For `beacon`, gas reporter starts working after commenting out `task(TASK_TEST, "Runs mocha tests").setAction` in `hardhat.config.ts` but the root cause is probably somewhere else, and this is just a manifestation.

nkuba:
The problems are caused by a bug in the gas reporter plugin, please see https://github.com/cgewecke/hardhat-gas-reporter/issues/86#issuecomment-1114700353 for more details.

PR #2958, opened 2022-04-28
Fixing gas reporter for random-beacon

dimpar:
Refs https://github.com/keep-network/keep-core/issues/2955

Combination of overwriting the hardhat's "test" task and requiring test fixtures `require("./test/fixtures")` resulted in hiding the gas report after the tests run. Instead of overwriting a "test" task, we can also create a new task that checks the count of needed accounts and run this check before the main "test" task.

Another addition to this PR is fetching the current market data from the Coinmarketcap. The gas report will show the current price for executing specific functions and the cost of contract deployments based on the live data. The free tier API key is good for 10000 requests per month. Execution: `COINMARKETCAP_API_KEY=<key> yarn test` For testing purposes please see [8374135](https://github.com/keep-network/keep-core/pull/2958/commits/837413537b26ed3b7ac0acc2caf5d496d477a8b7) ;)

### DIFF
--- a/solidity/random-beacon/hardhat.config.ts
+++ b/solidity/random-beacon/hardhat.config.ts
@@ -133,22 +133,21 @@ const config: HardhatUserConfig = {
   },
   gasReporter: {
     currency: "USD",
-    coinmarketcap: process.env.COINMARKETCAP_API_KEY
-  }
+    coinmarketcap: process.env.COINMARKETCAP_API_KEY,
+  },
 }
 
-task("check-accounts-count", "Checks accounts count")
-  .setAction(async () => {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires,global-require
-    const { constants } = require("./test/fixtures")
+task("check-accounts-count", "Checks accounts count").setAction(async () => {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires,global-require
+  const { constants } = require("./test/fixtures")
 
-    if (testConfig.operatorsCount < constants.groupSize4) {
-      throw new Error(
-        "not enough accounts predefined for configured group size: " +
-          `expected group size: ${constants.groupSize} ` +
-          `number of predefined accounts: ${testConfig.operatorsCount}`
-      )
-    }
-  });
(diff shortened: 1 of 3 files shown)


solidity/random-beacon/hardhat.config.ts
```diff
@@ -133,21 +131,23 @@ const config: HardhatUserConfig = {
...
+  },
 }
 
-task(TASK_TEST, "Runs mocha tests").setAction(async (args, hre, runSuper) => {
+task("check-accounts-count", "Checks accounts count").setAction(async () => {
   // eslint-disable-next-line @typescript-eslint/no-var-requires,global-require
   const { constants } = require("./test/fixtures")
 
-  if (testConfig.operatorsCount < constants.groupSize) {
+  if (testConfig.operatorsCount < constants.groupSize4) {
```
nkuba, line 144:
```suggestion
  if (testConfig.operatorsCount < constants.groupSize) {
```

solidity/random-beacon/package.json
```diff
@@ -15,7 +15,8 @@
   "scripts": {
     "clean": "hardhat clean",
     "build": "hardhat compile",
-    "test": "hardhat test",
+    "check-accounts-count": "hardhat check-accounts-count",
```
nkuba, line 18: We don't need to define this script. It's enough that we included task execution in `test` script.
```suggestion
```

nkuba (changes requested):

solidity/random-beacon/hardhat.config.ts
dimpar: oops, fat finger fixed in [b7d2f6e](https://github.com/keep-network/keep-core/pull/2958/commits/b7d2f6e3a86b5b3161deab98a8a2cb33c599b0f2)

solidity/random-beacon/package.json
dimpar: right, removed in [fcbfa94](https://github.com/keep-network/keep-core/pull/2958/commits/fcbfa946e9f1c860de440c12e5eeb05a24965322)

nkuba (approved):

PR #2960, opened 2022-04-28
Introduced authorization decrease change period parameter to the beacon

pdyraga:
~~Depends on #2959~~
See https://github.com/threshold-network/solidity-contracts/pull/99

Introduced authorization decrease change period parameter to the beacon

This value protect against malicious operators who manipulate
their weight by overwriting authorization decrease request, and
lowering or increasing their eligible stake this way.

Authorization decrease change period is the time period before the authorization
decrease delay end, during which the authorization decrease request can be
overwritten.

When the request is overwritten, the authorization decrease delay is
reset.

For example, if `authorizationDecraseChangePeriod` is set to 4
days, `authorizationDecreaseDelay` is set to 14 days, and someone
requested authorization decrease, it means they can not
request another decrease for the first 10 days. After 10 days pass,
they can request again and overwrite the previous authorization
decrease request. The delay time will reset for them and they
will have to wait another 10 days to alter it and 14 days to
approve it.

If set to a value equal to `authorizationDecreaseDelay, it means
that authorization decrease request can be always overwritten.
If set to zero, it means authorization decrease request can not
be overwritten until the delay end, and one needs to wait for the entire
authorization decrease delay to approve their decrease or to alter it.

```
(1) authorization decrease requested timestamp
(2) from this moment authorization decrease request can be
    overwritten
(3) from this moment authorization decrease request can be
    approved, assuming it was NOT overwritten in (2)

  (1)                            (2)                        (3)
 --x------------------------------x--------------------------x---->
   |                               \________________________/
   |                             authorizationDecreaseChangePeriod
    \______________________________________________________/
                   authorizationDecreaseDelay
```

### DIFF
--- a/solidity/random-beacon/test/RandomBeacon.Authorization.test.ts
+++ b/solidity/random-beacon/test/RandomBeacon.Authorization.test.ts
@@ -15,6 +15,7 @@ import type {
   TokenStaking,
   T,
   IApplication,
+  RandomBeaconGovernance,
 } from "../typechain"
 
 const { mineBlocks } = helpers.time
@@ -27,10 +28,12 @@ const MAX_UINT64 = ethers.BigNumber.from("18446744073709551615") // 2^64 - 1
 describe("RandomBeacon - Authorization", () => {
   let t: T
   let randomBeacon: RandomBeacon
+  let randomBeaconGovernance: RandomBeaconGovernance
   let sortitionPool: SortitionPool
   let staking: TokenStaking
 
   let deployer: SignerWithAddress
+  let governance: SignerWithAddress
 
   let owner: SignerWithAddress
   let stakingProvider: SignerWithAddress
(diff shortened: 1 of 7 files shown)


pdyraga:
#2959 got merged so I am undrafting this one

solidity/random-beacon/test/RandomBeacon.Authorization.test.ts
```diff
@@ -710,26 +716,219 @@ describe("RandomBeacon - Authorization", () => {
...
+              authorizationDecreaseDelay,
+              authorizationDecreaseChangePeriod,
+            } = await randomBeacon.authorizationParameters()
+            expect(authorizationDecreaseDelay).to.equal(
+              authorizationDecreaseChangePeriod
             )
-          ).to.be.equal(deauthorizingSecond)
+          })
+
+          context("when delay passed", () => {
```
nkuba, line 738: Can we add one more test under `"when change period is equal delay"` to cover a case where the `requestAuthorizationDecrease` is called without any `increaseTime`?

solidity/random-beacon/test/RandomBeacon.Authorization.test.ts
```diff
@@ -710,26 +716,219 @@ describe("RandomBeacon - Authorization", () => {
...
+              expect(
+                await randomBeacon.pendingAuthorizationDecrease(
+                  stakingProvider.address
+                )
+              ).to.be.equal(deauthorizingSecond)
+            })
+          })
+        })
+
+        context("when change period is zero", () => {
```
nkuba, line 795: Can we have two test cases under `"when change period is zero"`:
1. `requestAuthorizationDecrease` is called right away without any time increase
1. `requestAuthorizationDecrease` is called after the delay period

nkuba (approved):
Left non-blocking comments that can be taken care of in a follow-up.

PR #2966, opened 2022-05-01
update pre setup link

Leibniz137:
"that was an interim link that is no longer current" -maclane
image

see: https://discord.com/channels/866378471868727316/870383405936820224/967854416814284810


michalsmiarowski (approved):
LGTM! 👍

PR #2967, opened 2022-05-02
Document system flows

michalinacienciala:
In this PR we're adding the diagram documenting the interactions between
contracts in v2 of the system.
The diagram was created using the TikZ package of LaTeX. Having it in the
text form allows to keep the diagram  version-controlled and reviewable.
The PNG version of the diagram was rendered using
`imagemagick` tool. PDF version can be generated by
running `pdflatex system-flow.tikz`.

We place the diagram in the `./docs/diagrams` folder and move the v1-related docs
to `./docs-v1`.


michalinacienciala:
@Piotr, let me know if you also wish to store the `*.pdf` version of the diagram.

michalinacienciala:
Also, I movved all the dosc previously stored in `./docs` to `./docs-v1`. This includes among others the `rfc/rfc-19-random-beacon-v2.adoc` file. Should I create an `rfc` folder in `./docs` and move that file there?

michalinacienciala:
@pdyraga , if you don't need this PR to be merged yet, keep it open, there is couple of improvements I would like to apply. They shouldn't significantly affect the generated graphic, but should make the diagram easier to edit.

michalinacienciala:
@pdyraga, ready for review

.github/workflows/docs.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
```
pdyraga, line 5: We could also rename the job from `Docs` to `Docs v1` or something similar.

pdyraga (approved):

.github/workflows/docs.yml
michalinacienciala: Good point, see https://github.com/keep-network/keep-core/pull/2997.

PR #2981, opened 2022-05-12
ECDSA: Updated `@openzeppelin/contracts-upgradeable` to 4.6.0

pdyraga:
Non-RC version is now available. Updating from `4.6.0-rc.0` to `4.6.0`. Also, removed `@openzeppelin/contracts` dependency given we now use `@openzeppelin/contracts-upgradeable` in imports across ECDSA Solidity module code.

Refs https://github.com/keep-network/tbtc-v2/issues/152


pdyraga:
I will do the same in `tbtc-v2` but need to have this PR merged first so that I can update `@keep-network/ecdsa` dependency in `tbtc-v2`.

lukasz-zimnoch (approved):

PR #2982, opened 2022-05-12
Updated @threshold-network/solidity-contract to use 'development' tag

pdyraga:
This way, the most recent version of the package will be installed.
It should also help us with resolving the compilation bug that was fixed
in https://github.com/threshold-network/solidity-contracts/pull/97.

See https://github.com/keep-network/keep-core/pull/2981


michalinacienciala (approved):

pdyraga:
Reverted in https://github.com/keep-network/keep-core/pull/2985.

PR #2989, opened 2022-05-13
Deployments improvements

nkuba:
This PR contains a couple of improvements to the deployment scripts:
- deploy `SortitionPool` in `ecdsa` as `EcdsaSortitionPool` as we will have a separate pool for the `RandomBeacon` and we will name it `BeaconSortitionPool`
- don't transfer ownership of `ProxyAdmin` if the desired address is already an owner of the contract
- use the same compiler settings as in `@threshold-network/solidity-contracts` for `TokenStaking` compilation, with this change we won't hit the gas size limit as we will reflect how the contract was build originaly


pdyraga (approved):

nkuba:
Closing this PR as we cannot add all the changes in one shot due to the dependencies between `ecdsa` and `random-beacon` projects.

PR #2991, opened 2022-05-16
Deployments updates ecdsa

nkuba:
This PR contains a couple of improvements to the deployment scripts:
- deploy `SortitionPool` in `ecdsa` as `EcdsaSortitionPool` as we will have a separate pool for the `RandomBeacon` and we will name it `BeaconSortitionPool`
- don't transfer ownership of `ProxyAdmin` if the desired address is already an owner of the contract
- use the same compiler settings as in `@threshold-network/solidity-contracts` for `TokenStaking` compilation, with this change we won't hit the gas size limit as we will reflect how the contract was build originaly

Refs https://github.com/keep-network/keep-core/pull/2990
Refs https://github.com/keep-network/keep-core/pull/2989
Refs https://github.com/keep-network/tbtc-v2/issues/250


solidity/ecdsa/deploy/08_transfer_proxy_admin_ownership.ts
```diff
@@ -0,0 +1,34 @@
...
+  // use the DAO as the proxy admin owner and ESDM as the deputy. Until then we
+  // use ESDM as the owner of ProxyAdmin contract.
+  const newProxyAdminOwner = esdm
+
+  const proxyAdmin = await upgrades.admin.getInstance()
+
+  const currentOwner = await proxyAdmin.owner()
+
+  // The `@openzeppelin/hardhat-upgrades` plugin deploys a single ProxyAdmin
+  // per network. We don't want to transfer the ownership if the owner is already
```
pdyraga, line 21: Does it mean we'll have a single proxy admin for `WalletRegistry` and `Bridge`?

pdyraga (approved):

solidity/ecdsa/deploy/08_transfer_proxy_admin_ownership.ts
nkuba: No, as those are separate hardhat projects.

Issue #3005, opened 2022-06-02
Regenerate contract bindings


pdyraga:
We need to regenerate smart contract bindings in [`pkg/chain/gen`](https://github.com/keep-network/keep-core/tree/main/pkg/chain/gen) to include v2 smart contracts:
- Threshold Network `TokenStaking`
- `SortitionPool`
- `RandomBeacon`
- `WalletRegistry`
-  TBTCv2 `Bridge`

Part of this work requires finding a workaround to get rid of the old v1 contracts without breaking the build. One potential option could be to replace `ethereum.go` implementation with a fake one.

Dependencies and Makefiles need to be updated to make `go generate` use the v2 code.

nkuba:
Right now we generate ABIs for contracts in the [go bindings generator with `solc`](https://github.com/keep-network/keep-core/blob/2966277e6fe24cb2348d1d00d3fb1327482e796e/pkg/chain/v2/common/Makefile#L30-L31). What if we do the ABIs generation with hardhat in the contracts project and expose the ABI files in the package to be consumed by the Go generator?

nkuba:
@pdyraga do we want to add the `MaintainerProxy` to the list? It was suggested by @dimpar in https://github.com/keep-network/keep-core/pull/3016#issuecomment-1167283716.

PR #3016, opened 2022-06-07
Generate Go bindings for V2 contracts

nkuba:
Closes https://github.com/keep-network/keep-core/issues/3005
Depends on https://github.com/keep-network/keep-common/pull/88
Depends on https://github.com/keep-network/tbtc-v2/pull/376

TODO:
- [x] generate bindings for V2 contracts
- [x] fix generator to handle structs as parameters correctly
- [x] update keep-common dependency once https://github.com/keep-network/keep-common/pull/88 is merged
- [x] generate and commit bindings
- [x] fix build
- [x] fix docker image build
- [x] build go from artifacts

## Introduction

This PR implements Go bindings generation for Solidity contracts:
- Threshold Network `TokenStaking`
- `SortitionPool`
- `RandomBeacon`
- `WalletRegistry`
-  TBTCv2 `Bridge`

## SortitionPool

Please note that bindings for `SortitionPool` are generated separately for `random-beacon` and `ecdsa` as those projects can use different dependency versions.

## Contracts Artifacts

This PR changes the approach to what we consider the source for code generation. 
Previously we were generating Go bindings from solidity contract sources which was causing a lot of headaches in regards to setting up `solc` compiler just for this task. 
What is more, there are problems with `solc` support on machines that run on Apple M1 processors, which some folks in the team transitioned to.
Instead of generating ABIs from sources, we use contracts artifacts that are being published as part of NPM packages. We download the artifacts to `_contracts` directory. The directory name is prefixed with the underscore (`_`), so it is [ignored by the go tool](https://pkg.go.dev/cmd/go#hdr-Package_lists_and_patterns).

## V1 Contracts

We will be removing parts of V1-related code from the client bit by bit. In this PR we remove the generation of the binding for V1 contracts. We leave the already generated code and use it in the docker image building process. This simplifies the docker build and reduces the build time.

## Execution

Ensure `npm` at least v7.x is installed and run:
```sh
go generate ./...
```


dimpar:
Are you sure we don't miss anything in `go.mod` to make `go generate ./...` work? I had the following error:

Screenshot 2022-06-27 at 12 47 24


I did the suggested commands above and then it worked.

lukasz-zimnoch:
> Are you sure we don't miss anything in `go.mod` to make `go generate ./...` work? I had the following error:
> 
> Screenshot 2022-06-27 at 12 47 24
> 
> I did the suggested commands above and then it worked.

Do you use Go `1.18`?

dimpar:
> > Are you sure we don't miss anything in `go.mod` to make `go generate ./...` work? I had the following error:
> > Screenshot 2022-06-27 at 12 47 24
> > I did the suggested commands above and then it worked.
> 
> Do you use Go `1.18`?

go version go1.18.3 darwin/amd64

pkg/chain/common/gen/Makefile
```diff
@@ -0,0 +1,45 @@
...
+# Expects two arguments:
+# 1 - dependencies mapping for contracts import
+# 2 - solidity source file path
+solc_generate = solc $(1) \
+		--allow-paths ${solidity_dir} \
+		--overwrite \
+		--abi \
+		-o abi $(2)
+
+all: gen_contract_go
```
dimpar, line 24: When I run `make` or `make all` from `pkg/chain` I get:
`make: *** No rule to make target "all".  Stop.`

When I run `make` or `make all` from a given dir like `pkg/chain/random-beacon/gen` I get: 
```
solc  @openzeppelin/=/node_modules/@openzeppelin/ @keep-network/=/node_modules/@keep-network/ @threshold-network/=/node_modules/@threshold-network/ @thesis/=/node_modules/@thesis/ --allow-paths  --overwrite --abi -o abi  
No input files given. If you wish to use the standard input please specify "-" explicitly.
```

Do you have similar errors? Is `make` command supported?

nkuba: You need to specify `SOLIDITY_DIR` env variable.
Like here:
https://github.com/keep-network/keep-core/blob/f60b46da4018840e4ff3ee63a3bd5e3e59837395/pkg/chain/random-beacon/gen/gen.go#L3

dimpar:
> This PR implements Go bindings generation for Solidity contracts:
Threshold Network TokenStaking
SortitionPool
RandomBeacon
WalletRegistry
TBTCv2 Bridge

What about `MaintainerProxy` from TBTCv2? Maintainers are the willing off-chain clients and I think we need bindings for this contract as well.

nkuba:
> > This PR implements Go bindings generation for Solidity contracts:
> > Threshold Network TokenStaking
> > SortitionPool
> > RandomBeacon
> > WalletRegistry
> > TBTCv2 Bridge
> 
> What about `MaintainerProxy` from TBTCv2? Maintainers are the willing off-chain clients and I think we need bindings for this contract as well.

I'm not sure the client is expected to support the TBTCv2 maintenance at this stage. I routed your question to the parent issue https://github.com/keep-network/keep-core/issues/3005.

nkuba:
The build is finally green. I'm looking forward to your reviews @lukasz-zimnoch @dimpar @pdyraga.

pdyraga:
👀

pkg/chain/common/gen/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+# Function to run solidity sources comilation.
+# Expects two arguments:
+# 1 - dependencies mapping for contracts import
+# 2 - solidity source file path
+solc_generate = solc $(1) \
+		--allow-paths ${solidity_dir} \
+		--overwrite \
+		--abi \
+		-o abi $(2)
```
pdyraga, line 22: Having a specific solc version installed globally is getting problematic. Running `brew install solidity` as in our [`macos-setup.sh`](https://github.com/keep-network/keep-core/blob/b872c35aa405b4e857fba7317aa6bd23137f5a01/scripts/macos-setup.sh#L10-L13) (which is by the way outdated now), installs the most recent version 0.8.15 that does not allow to compile 0.8.9 code. One option is to use a published binary or a tool like [`solc-select`](https://github.com/crytic/solc-select) but binaries are not available for M1 and [it will take some time](https://github.com/ethereum/solidity/issues/12291) until they will be published. It is [no longer that easy](https://stackoverflow.com/questions/3987683/homebrew-install-specific-version-of-formula) to install brew formula from the specific version of a file on GitHub and it requires doing a local tap.

There are two options - we can change `pragma solidity 0.8.9;` to `pragma solidity ^0.8.9;` in all imported projects. I did it locally for the sake of testing this PR on solc 0.8.15 and it works. This is not a good solution though because eventually we want to fix these versions for deployment, tag the specific version, and import the tagged version from Go client.

Another option is to use `hardhat build` to generate ABIs. This feels like a good solution to me because one needs to have `npm` and `yarn` anyway to install the dependencies.

How hard would it be to switch from `solc` to `yarn build` here?

pdyraga: We synced with @nkuba about this problem and the idea is to not use `solc` at all. From the Go code perspective, we are interested only in ABI. We do not want to `yarn install` Solidity projects and execute builds. Instead, we want to `wget` packages published to the NPM registry and read ABI from there. It's Solidity project's responsibility to generate ABIs and publish a package.

@nkuba will try to implement the new solution today. By the end of the day, we'll decide if we merge this PR and update the code separately to unblock the rest of the work or if we are able to implement the new solution quickly in this PR.

nkuba: I updated the PR with the new approach.

pkg/chain/common/gen/Makefile
nkuba: Please try again after the latest changes I pushed.

dimpar: Now it works.

dimpar:
> > > Are you sure we don't miss anything in `go.mod` to make `go generate ./...` work? I had the following error:
> > > Screenshot 2022-06-27 at 12 47 24
> > > I did the suggested commands above and then it worked.
> > 
> > 
> > Do you use Go `1.18`?
> 
> go version go1.18.3 darwin/amd64

After the recent changes, it works now.

pdyraga:
👀

pdyraga (approved):

Issue #3019, opened 2022-06-10
Group life time


dimpar:
There's something wrong with how we update the group lifetime. In `RandomBeacon` we [set 259200 blocks](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/contracts/RandomBeacon.sol#L462) which are ~45days but in the `RandomBeaconGovernance` we have a check [1 < groupLifetime < 14 days](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/contracts/RandomBeaconGovernance.sol#L542) 
The first problem is that we compare apples to oranges ie. blocks and time. And the second potential problem is that the range is between 1 and 14 days, whereas it is initially set to 45 days. 

The group should not live longer than the authorization decrease delay.

Refs https://github.com/keep-network/keep-core/issues/3011

PR #3020, opened 2022-06-10
Flagging code that needs to be fixed

dimpar:
Closes https://github.com/keep-network/keep-core/issues/3019

TBD: What is the correct group lifetime range?


dimpar:
closing in favor of https://github.com/keep-network/keep-core/pull/3018

PR #3023, opened 2022-06-14
Remove specific stakes from t auth data

michalsmiarowski:
Removes stakes that are cancelled (while in initialization period) from
threshold authorization table. Also disable the `Authorize and Stake` button for
the stakes that are in initialization period (and adds tooltip informing about
why it is disabled).

AUTHORIZE AND STAKE


r-czajkowski (approved):
🚢

Issue #3011, opened 2022-06-03
Modify client installation scripts


pdyraga:
Modify `scripts/install.sh` to adjust it to Threshold staking mechanism. This also requires developing code to stake with T.

The installation script should:
0. Build the client.
1. Deploy Threshold Token Staking.
2. Deploy `RandomBeacon`.
3. Stake into T.
4. Increase authorization on `RandomBeacon` contract for the client. 

Let's also do a cleanup and delete Celo-specific installation scripts.

If needed, modify `start.sh` as well.

Installation script PR: https://github.com/keep-network/keep-core/pull/3026

Depends on:
- https://github.com/keep-network/keep-core/pull/3018
- https://github.com/threshold-network/solidity-contracts/pull/108

Related PR:
- https://github.com/keep-network/keep-core/pull/3033
- https://github.com/keep-network/tbtc-v2/pull/365

PR #3026, opened 2022-06-16
'install.sh' script to work on local Geth

dimpar:
Refs https://github.com/keep-network/keep-core/issues/3011

`install.sh` script will:
- unlock Geth accounts
- install yarn dependencies
- deploy contracts on local Geth
- build a client

`initialize.sh` script will:
- mint & approve `T` tokens
- stake `T` tokens
- increase authorization for `RandomBeacon`
- register operator

Tested on
```
Geth Version: 1.10.19-unstable
Go Version: go1.18.3
```

### How to run scripts
The basic execution is presented below. For more options please see `--help`

`./scripts/install.sh` - you need to run it only once

`./scripts/initialize.sh --stake-owner 0x1C373bdfA9Dc7CA00A4ed123C43E68A9706527E3` - for each client you need to set a `--stake-owner` with a new address. In other words if you need to setup 3 clients you need to run it 3 times:
- `./scripts/initialize.sh --stake-owner <address1>`
- `./scripts/initialize.sh --stake-owner <address2>`
- `./scripts/initialize.sh --stake-owner <address3>`
In this basic execution the default values for `provider, operator, beneficiary, and authorizer` will be equal to `stake owner`. If you need to set different addresses, you can do so with argument options. See `--help`

### How to run Geth

Example config for `.zshrc` to run `Geth` locally:
```
alias cleangeth="rm -rf $GETH_DATA_DIR/geth"
alias initgeth="cd ~/go-ethereum && ./build/bin/geth --datadir=$GETH_DATA_DIR init $GETH_DATA_DIR/genesis.json"
alias cleangethdb="cd ~/go-ethereum && ./build/bin/geth removedb"

alias rungeth="cleangethdb && cleangeth && initgeth && cd ~/go-ethereum && ./build/bin/geth --port 3000 --networkid 1101 --identity 'somerandomidentity' \
       --ws --ws.addr '127.0.0.1' --ws.port '8546' --ws.origins '*' --ws.api 'admin, debug, web3, eth, txpool, personal, ethash, miner, net' \
       --http --http.port '8545' --http.addr '127.0.0.1' --http.corsdomain '' --http.api 'admin, debug, web3, eth, txpool, personal, ethash, miner, net' \
       --datadir=$GETH_DATA_DIR --allow-insecure-unlock \
       --miner.etherbase=$GETH_ETHEREUM_ACCOUNT --mine --miner.threads=1"
```

then in terminal run `rungeth`

### Future work
- automate running the `initialize.sh` for all the fetched addresses.
- split `stake` task to 4 different tasks: mint, stake, authorization, and operator registration.

Depends on:
https://github.com/threshold-network/solidity-contracts/pull/108

### DIFF
--- a/solidity/random-beacon/yarn.lock
+++ b/solidity/random-beacon/yarn.lock
@@ -1334,9 +1334,9 @@
     "@openzeppelin/contracts" "^4.1.0"
 
 "@threshold-network/solidity-contracts@development":
-  version "1.2.0-dev.14"
-  resolved "https://registry.yarnpkg.com/@threshold-network/solidity-contracts/-/solidity-contracts-1.2.0-dev.14.tgz#2e7104e41fdf261696aec8dc3a94de7c77be6a53"
-  integrity sha512-9OBoYMelzzVN1MKswRCtdTmwRuz3JvDHFRJo8jJuSiHeINHWbvAYQyz0KkF4qEhvLmbMxsRfO82jQcmN+LjqKg==
+  version "1.2.0-dev.15"
+  resolved "https://registry.yarnpkg.com/@threshold-network/solidity-contracts/-/solidity-contracts-1.2.0-dev.15.tgz#3415579a3be37aed452656a91124c8bd575df569"
+  integrity sha512-kETXCaWVho0uTh8Dtq3pAewlbw37tfqzFQRwh2cHjiQ78xtKpj1JZ+VbsAOnsPuLqqfCLHZmfWvK2dLfLaeFZQ==
   dependencies:
     "@keep-network/keep-core" ">1.8.1-dev <1.8.1-pre"
     "@openzeppelin/contracts" "~4.5.0"


.gitignore
```diff
@@ -53,3 +53,6 @@ export.json
 
 # TypeScript
 typechain/
+
+# Contract artifacts
+solidity/random-beacon/deployments
```
nkuba, line 58: We should declare it in the Random Beacon's file: https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/.gitignore

scripts/install.sh
```diff
@@ -20,40 +20,42 @@ help()
...
 # Parse short options
 OPTIND=1
-while getopts "c:n:mh" opt
+while getopts "c:n:dh" opt
 do
    case "$opt" in
       c ) config_dir_path="$OPTARG" ;;
       n ) network="$OPTARG" ;;
-      m ) contracts_only=true ;;
+      d ) reuse_contracts=true ;;
```
nkuba, line 58: Should this be assigned to `true`, shouldn't it be `$OPTARG`?

scripts/install.sh
```diff
@@ -63,54 +65,50 @@ shift $(expr $OPTIND - 1) # remove options from positional parameters
 # Overwrite default properties
 CONFIG_DIR_PATH=${config_dir_path:-$CONFIG_DIR_PATH_DEFAULT}
 NETWORK=${network:-$NETWORK_DEFAULT}
-CONTRACTS_ONLY=${contracts_only:-false}
+REUSE_CONTRACTS=${reuse_contracts:-false}
+
+echo "reuse contracts::" ${REUSE_CONTRACTS}
```
nkuba, line 70: 💣 ?

scripts/install.sh
```diff
@@ -63,54 +65,50 @@ shift $(expr $OPTIND - 1) # remove options from positional parameters
...
     printf "${LOG_START}Unlocking ethereum accounts...${LOG_END}"
     KEEP_ETHEREUM_PASSWORD=$KEEP_ETHEREUM_PASSWORD \
-        npx truffle exec scripts/unlock-eth-accounts.js --network $NETWORK
+        npx hardhat run scripts/unlock-eth-accounts.ts --network $NETWORK
 fi
 
-printf "${LOG_START}Migrating contracts...${LOG_END}"
-rm -rf build/
+printf "${LOG_START}Building contracts...${LOG_END}"
+# rm -rf build && rm -rf cache && rm -rf typechain
```
nkuba, line 91: Uncomment?

scripts/install.sh
```diff
@@ -20,40 +20,42 @@ help()
...
            "Required only for 'local' network. Default value is 'password'"
    echo -e "\tCONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY: Contracts owner private key on Ethereum. Required for non-local network only"
    echo -e "\nCommand line arguments:\n"
    echo -e "\t--config-dir: Path to keep-core client configuration file(s)"
    echo -e "\t--network: Ethereum network for keep-core client."\
-                        "Available networks and settings are specified in the 'truffle-config.js'"
-   echo -e "\t--contracts-only: Should execute contracts part only."\
-                        "Client installation will not be executed.\n"
+                        "Available networks and settings are specified in the 'hardhat.config.ts'"
+   echo -e "\t--reuse-contracts: The already deployed contracts can be reused to speed up the development. Default is false.\n"\
```
nkuba, line 32: Maybe `--skip-deployment`?

scripts/install.sh
```diff
@@ -63,54 +65,50 @@ shift $(expr $OPTIND - 1) # remove options from positional parameters
...
-    npx truffle migrate --reset --network $NETWORK
+yarn build
 
-printf "${LOG_START}Copying contract artifacts...${LOG_END}"
-rm -rf artifacts
-cp -r build/contracts artifacts
-npm link
+if [ "$REUSE_CONTRACTS" = false ] ; then
+   printf "${LOG_START}Deploying contracts...${LOG_END}"
+   rm -rf deployments/development/
```
nkuba, line 97: Isn't `--reset` flag enough?

scripts/install.sh
```diff
@@ -63,54 +65,50 @@ shift $(expr $OPTIND - 1) # remove options from positional parameters
...
-  done
-
-  printf "${LOG_START}Building keep-core client...${LOG_END}"
-  cd $KEEP_CORE_PATH
-  go generate ./...
-  go build -a -o keep-core .
-fi
+    npx hardhat run scripts/init-contracts.ts --network $NETWORK
+
+printf "${LOG_START}Building keep-core client...${LOG_END}"
```
nkuba, line 108: To maintain the previous functionality of `CONTRACTS_ONLY` flag, maybe we could introduce `--skip-client-build` flag? This would be consistent with `--skip-deployment` flag.

nkuba (changes requested):

scripts/install.sh
dimpar: Sure, we can do this.

Part of: [2dcb842](https://github.com/keep-network/keep-core/pull/3026/commits/2dcb842ce661077ccb2491e425aa08f4112bbdbe)

scripts/install.sh
dimpar: Technically yes and this was added "just in case", cause if we remove `/deployments` then for sure we need new artifacts. Unless you have a strong opinion, I'd maybe comment this out and leave it for now. If we notice that something is cached or not updated in the contracts we can uncomment this and see if that helps. Part of: [2dcb842](https://github.com/keep-network/keep-core/pull/3026/commits/2dcb842ce661077ccb2491e425aa08f4112bbdbe)

scripts/install.sh
dimpar: Sure, I like your name better Part of: [2dcb842](https://github.com/keep-network/keep-core/pull/3026/commits/2dcb842ce661077ccb2491e425aa08f4112bbdbe)

scripts/install.sh
dimpar: Part of: [2dcb842](https://github.com/keep-network/keep-core/pull/3026/commits/2dcb842ce661077ccb2491e425aa08f4112bbdbe)

scripts/install.sh
dimpar: image

scripts/install.sh
dimpar: It worked with the original version as well, but I changed it to `$OPTARG`. Part of: [2dcb842](https://github.com/keep-network/keep-core/pull/3026/commits/2dcb842ce661077ccb2491e425aa08f4112bbdbe)

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
...
+  const stakedAmount = to1e18(1_000_000) // 1MM T
+
+  const deployerSigner = await ethers.getSigner(deployer)
+  await t.connect(deployerSigner).mint(stakeOwner, stakedAmount)
+  const stakeOwnerSigner = await ethers.getSigner(stakeOwner)
+  await t.connect(stakeOwnerSigner).approve(staking.address, stakedAmount)
+
+  // We need to wait until the block is mined and that transactions in the
+  // mempool were executed.
+  await new Promise((f) => setTimeout(f, BLOCK_TIME))
```
nkuba, line 39: We could use [`wait()` function](https://docs.ethers.io/v5/api/providers/types/#providers-TransactionResponse).
```suggestion
  ;(
    await t.connect(stakeOwnerSigner).approve(staking.address, stakedAmount)
  ).wait()
```

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  console.log(
+    `Staking ${stakedAmount.toString()} to the staking provider ${stakingProvider} ...`
+  )
+
+  await staking
+    .connect(stakeOwnerSigner)
+    .stake(stakingProvider, beneficiary, authorizer, stakedAmount)
+
+  await new Promise((f) => setTimeout(f, BLOCK_TIME))
```
nkuba, line 49:
```suggestion
  ;(
    await staking
      .connect(stakeOwnerSigner)
      .stake(stakingProvider, beneficiary, authorizer, stakedAmount)
  ).wait()
```

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
...
+async function init() {
+  const t: T = await helpers.contracts.getContract<T>("T")
+  const staking: TokenStaking =
+    await helpers.contracts.getContract<TokenStaking>("TokenStaking")
+  const randomBeacon: RandomBeacon = await helpers.contracts.getContract(
+    "RandomBeacon"
+  )
+
+  const accounts = await getAccounts()
+  const deployer = accounts[0]
```
nkuba, line 24:
```suggestion
  const { deployer } = await getNamedAccounts()
```

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
...
+  const stakeOwner = accounts[1]
+  const stakingProvider = accounts[2]
```
nkuba, line 26: Maybe instead of getting accounts at specific indices, we could expect the address to be provided to the script as a task arguments?

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions, import/no-extraneous-dependencies */
+import { ethers, helpers } from "hardhat"
```
nkuba, line 2: Instead of a script, we could make this file a [task](https://hardhat.org/guides/create-task). We could expect the stake owner and operator to be provided as arguments.

For example see https://github.com/keep-network/tbtc-v2/tree/main/solidity/tasks

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
...
+const { to1e18 } = helpers.number
+
+const BLOCK_TIME = 5000 // in ms
+
+async function getAccounts() {
+  const provider = new ethers.providers.JsonRpcProvider(process.env.NETWORK)
+  return provider.listAccounts()
+}
+
+async function init() {
```
nkuba, line 15: We could rename the script file from `init-contracts.ts` to `stake.ts` and rename this function, as we're not initializing the contracts but staking.

```suggestion
async function stake() {
```

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
...
+  )
+
+  const accounts = await getAccounts()
+  const deployer = accounts[0]
+  const stakeOwner = accounts[1]
+  const stakingProvider = accounts[2]
+  const beneficiary = stakeOwner
+  const authorizer = stakeOwner
+
+  const stakedAmount = to1e18(1_000_000) // 1MM T
```
nkuba, line 30: The value could be a task's argument.

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  await new Promise((f) => setTimeout(f, BLOCK_TIME))
+
+  console.log(
+    `T balance of the staking contract: ${(
+      await t.balanceOf(staking.address)
+    ).toString()}`
+  )
+
+  const minimumAuthorization = await randomBeacon.minimumAuthorization()
```
nkuba, line 57: Authorization amount also could be a task argument with a default `minimumAuthorization` if not provided.

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  await staking
+    .connect(authorizerSigner)
+    .increaseAuthorization(
+      stakingProvider,
+      randomBeacon.address,
+      minimumAuthorization
+    )
+
+  await new Promise((f) => setTimeout(f, BLOCK_TIME))
```
nkuba, line 72:
```suggestion
  ;(
    await staking
      .connect(authorizerSigner)
      .increaseAuthorization(
        stakingProvider,
        randomBeacon.address,
        minimumAuthorization
      )
  ).wait()
```

nkuba (changes requested):

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const authorizedStaked = await staking.authorizedStake(
+    stakingProvider,
+    randomBeacon.address
+  )
+
+  console.log(
+    `Staked authorization ${authorizedStaked.toString()} was increased for the Random Beacon`
+  )
+}
```
nkuba, line 82: We also need to call `RandomBeacon.registerOperator` function from the staking provider account to register an operator.

nkuba (changes requested):

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
...
+
+init()
+  .then(() => process.exit(0))
+  .catch((error) => {
+    console.error(error.error)
```
nkuba, line 87:
```suggestion
    console.error(error)
```

nkuba (changes requested):

scripts/install.sh
```diff
@@ -63,54 +66,52 @@ shift $(expr $OPTIND - 1) # remove options from positional parameters
...
 NETWORK=${network:-$NETWORK_DEFAULT}
-CONTRACTS_ONLY=${contracts_only:-false}
+SKIP_DEPLOYMENT=${skip_deployment:-false}
+SKIP_CLIENT_BUILD=${skip_client_build:-false}
 
 # Run script
 printf "${LOG_START}Starting installation...${LOG_END}"
 
 printf "Config dir path: $CONFIG_DIR_PATH\n"
 printf "Network: $NETWORK"
```
michalinacienciala, line 76:
```suggestion
printf "Network: $NETWORK\n"
```

scripts/install.sh
```diff
@@ -63,54 +66,52 @@ shift $(expr $OPTIND - 1) # remove options from positional parameters
...
 CONFIG_DIR_PATH=${config_dir_path:-$CONFIG_DIR_PATH_DEFAULT}
 NETWORK=${network:-$NETWORK_DEFAULT}
-CONTRACTS_ONLY=${contracts_only:-false}
+SKIP_DEPLOYMENT=${skip_deployment:-false}
+SKIP_CLIENT_BUILD=${skip_client_build:-false}
 
 # Run script
 printf "${LOG_START}Starting installation...${LOG_END}"
 
 printf "Config dir path: $CONFIG_DIR_PATH\n"
```
michalinacienciala, line 75: Do we use `$CONFIG_DIR_PATH` anywhere other than here? I can't see it being used anywhere.

scripts/install.sh
```diff
@@ -63,54 +66,52 @@ shift $(expr $OPTIND - 1) # remove options from positional parameters
...
+rm -rf build && rm -rf cache && rm -rf typechain
+yarn build
 
-printf "${LOG_START}Copying contract artifacts...${LOG_END}"
-rm -rf artifacts
-cp -r build/contracts artifacts
-npm link
+if [ "$SKIP_DEPLOYMENT" = false ] ; then
+   printf "${LOG_START}Deploying contracts...${LOG_END}"
+   # rm -rf deployments/development/
```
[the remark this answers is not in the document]
dimpar: pls see https://github.com/keep-network/keep-core/pull/3026#discussion_r900165580

scripts/install.sh
```diff
@@ -20,40 +20,43 @@ help()
    echo -e "\nUsage: ENV_VAR(S) $0"\
            "--config-dir <path>"\
            "--network <network>"\
-           "--contracts-only"
+           "--skip-deployment <boolean>"\
+           "--skip-client-build <boolean>"
```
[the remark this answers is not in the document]
dimpar: We need to provide a boolean in the current implementation. However, I've changed some logic [91e9039](https://github.com/keep-network/keep-core/pull/3026/commits/91e9039e501ecd4cca7d27e1475797dc03043ff1) so we can pass these 2 flags without a boolean

scripts/install.sh
```diff
@@ -20,40 +20,43 @@ help()
...
    echo -e "\nEnvironment variables:\n"
    echo -e "\tKEEP_ETHEREUM_PASSWORD: The password to unlock local Ethereum accounts to set up delegations."\
            "Required only for 'local' network. Default value is 'password'"
    echo -e "\tCONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY: Contracts owner private key on Ethereum. Required for non-local network only"
```
michalinacienciala, line 28: We no longer use the `local` network name. Also, I think the variable is also not needed for the `hardhat` network.
Maybe
```suggestion
   echo -e "\tCONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY: Contracts owner private key on Ethereum. Required for non-development and non-hardhat network only"
```
?

michalinacienciala (commented):
First batch of comments (I'm still testing, there might be more)

michalinacienciala:
From time to time, when running the script I get following error in the `Initializing contracts...` step:
```js
Staking 1000000000000000000000000 to the staking provider 0x65849fc91F8289b401e9d8F6D306b814B7F055C3 ...
T balance of the staking contract: 0
Increasing min authorization 40000000000000000000000 for the Random Beacon ...
ProviderError: execution reverted: Not authorizer
    at HttpProvider.request (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/hardhat/src/internal/core/providers/http.ts:49:19)
    at GanacheGasMultiplierProvider.request (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/hardhat/src/internal/core/providers/gas-providers.ts:310:34)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

One time I also got following error:
```js
Staking 1000000000000000000000000 to the staking provider 0x65849fc91F8289b401e9d8F6D306b814B7F055C3 ...
T balance of the staking contract: 0
Increasing min authorization 40000000000000000000000 for the Random Beacon ...
undefined
```

Those two above errors occurred when I was running the `install.sh` with default values (`development` network) or with `--skip-client-build` flag. I also had successfull executions for those commands.

michalinacienciala:
Apart from that I noticed a problem with execution of the script on `hardhat` network - in that case script always fails on `Initializing contracts...` step, printing just `undefined` below the table:
```
Initializing contracts...
No need to generate any newer typings.
 ·--------------------------|-------------|---------------·
 |  Contract Name           ·  Size (KB)  ·  Change (KB)  │
 ···························|·············|················
 |  Address                 ·      0.086  ·               │
 ···························|·············|················
[...]
 ···························|·············|················
 |  VendingMachine          ·      3.451  ·               │
 ·--------------------------|-------------|---------------·
undefined
```
I provided additional logging to the scripts and it looks to me that problem occurs on  `const t: T = await helpers.contracts.getContract<T>("T")` line from `init-contracts.ts`.

solidity/random-beacon/scripts/init-contracts.ts
dimpar: Yea, I see why it might be confusing, but I don't think `stake.ts` & `stake()` fit perfectly here because `stake.ts` suggests that we only execute a `stake` part here, which is not true. We also do authorization and [regestring an operator](https://github.com/keep-network/keep-core/pull/3026#discussion_r901686383). Same with `stake()` which suggests we execute only a `stake()` function. I think renaming to `staking.ts` and `setup()` are more general namings for functionalities around "staking".

solidity/random-beacon/scripts/init-contracts.ts
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions, import/no-extraneous-dependencies */
```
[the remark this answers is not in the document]
dimpar: I removed `no-unused-expressions` but if you try `yarn format:fix` without `import/no-extraneous-dependencies` the Eslint will show you errors.

solidity/random-beacon/scripts/unlock-eth-accounts.ts
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable no-await-in-loop, import/no-extraneous-dependencies */
```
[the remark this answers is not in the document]
dimpar: Try removing it and you'll see that Eslint is complaining

solidity/random-beacon/scripts/init-contracts.ts
dimpar: The reason why I put `error.error` is that displaying just `error` will give you an error stack with a general heading `Error: cannot estimate gas;`. The actual error is behind `error.error`, but if you have a strong opinion on that I changed it as part of [d106884](https://github.com/keep-network/keep-core/pull/3026/commits/d1068848c80f083d6e6dc690fd563253b502fca3)

solidity/random-beacon/scripts/init-contracts.ts
dimpar: I changed it with `await` to make it work.

solidity/random-beacon/scripts/init-contracts.ts
dimpar: I changed it with `await` to make it work.

solidity/random-beacon/.gitignore
```diff
@@ -4,3 +4,6 @@ cache/
 export/
 hardhat-dependency-compiler/
 export.json
+
+# Contract artifacts
+/deployments
```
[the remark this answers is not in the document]
dimpar: done

solidity/random-beacon/scripts/init-contracts.ts
dimpar: done

scripts/install.sh
dimpar: Done

scripts/install.sh
dimpar: Right, it's not used anymore

solidity/random-beacon/scripts/init-contracts.ts
dimpar: Done in [af96f7a](https://github.com/keep-network/keep-core/pull/3026/commits/af96f7a2a297b8f4a474740eb23c9704a9abfe41)

scripts/install.sh
```diff
@@ -63,54 +66,52 @@ shift $(expr $OPTIND - 1) # remove options from positional parameters
...
 printf "Network: $NETWORK"
 
-cd $KEEP_CORE_SOL_PATH
+cd $KEEP_BEACON_SOL_PATH
 
-printf "${LOG_START}Installing NPM dependencies...${LOG_END}"
-npm install
+printf "${LOG_START}Installing YARN dependencies...${LOG_END}"
+# rm -rf node_modules/
+yarn install
```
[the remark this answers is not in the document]
dimpar: Mmm.. do you have an example of such checking somewhere? The main goal of this script is to deploy contracts on `Geth`, set up staking contracts, and build the client. For this, we need to run a local instance of `Geth` cause it's a prerequisite to run everything locally. Also, the error code suggests that something is not right with a server and the first thing that comes up to my mind is to see if the local instance is running.

solidity/random-beacon/scripts/init-contracts.ts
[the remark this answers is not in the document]
nkuba: You could add the `scripts` directory here https://github.com/keep-network/keep-core/blob/9f530727c816bf9472ed2d4c975a53cf801d3531/solidity/random-beacon/.eslintrc#L7

solidity/random-beacon/scripts/init-contracts.ts
nkuba: Should it be `error.message` then? `error.error` printed `undefined` for me.

scripts/install.sh
michalinacienciala: Then I think we can also delete:
https://github.com/keep-network/keep-core/blob/17d3a2cc12aa9bad352fc175ec553c4bbb697590/scripts/install.sh#L13,
https://github.com/keep-network/keep-core/blob/17d3a2cc12aa9bad352fc175ec553c4bbb697590/scripts/install.sh#L21,
https://github.com/keep-network/keep-core/blob/17d3a2cc12aa9bad352fc175ec553c4bbb697590/scripts/install.sh#L42,
https://github.com/keep-network/keep-core/blob/17d3a2cc12aa9bad352fc175ec553c4bbb697590/scripts/install.sh#L56
and modify https://github.com/keep-network/keep-core/blob/17d3a2cc12aa9bad352fc175ec553c4bbb697590/scripts/install.sh#L53.

scripts/install.sh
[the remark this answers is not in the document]
michalinacienciala: I was thinking of something checking list of processes in the system (basing on `ps aux | grep geth`) or something similar to the script that what we had in `local-setup` repo that waited until Geth gets initialized: https://github.com/keep-network/local-setup/blob/main/pre-install.sh). But if adapting this for our purposes would not be something really quick, I'm ok with leaving the code as it is.

solidity/random-beacon/scripts/init-contracts.ts
dimpar: [388f9df](https://github.com/keep-network/keep-core/pull/3026/commits/388f9df64b13b1a3710d0c979e578a7d5fc7f2a9) & [5a48a86](https://github.com/keep-network/keep-core/pull/3026/commits/5a48a8651da2af566d5989040c377573909f9f7f)
I've converted two scripts and made them tasks. However, I'm not sure how much value it adds, to be honest. The same thing we could've accomplished with the existing scripts. As far as I understand tasks are useful when can be used ex. by a third party or can be run standalone. In our case everything is run by `install.sh` and I'm not sure if those tasks will ever be run standalone. Anyway, I might be wrong but just wanted to discuss briefly expectations of what should be put as a "task" and what can be written and run as a "simple script".

solidity/random-beacon/scripts/init-contracts.ts
dimpar: see https://github.com/keep-network/keep-core/pull/3026#discussion_r903739904

solidity/random-beacon/scripts/init-contracts.ts
dimpar: see https://github.com/keep-network/keep-core/pull/3026#discussion_r903739904

solidity/random-beacon/scripts/init-contracts.ts
dimpar: see https://github.com/keep-network/keep-core/pull/3026#discussion_r903739904

solidity/random-beacon/scripts/init-contracts.ts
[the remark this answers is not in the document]
dimpar: [e958c35](https://github.com/keep-network/keep-core/pull/3026/commits/e958c358a9497c463cc5454d082a778395ae5727)

solidity/random-beacon/scripts/init-contracts.ts
dimpar: I've converted this file to a "task". No more catching errors.

dimpar:
> From time to time, when running the script I get following error in the `Initializing contracts...` step:
> 
> ```js
> Staking 1000000000000000000000000 to the staking provider 0x65849fc91F8289b401e9d8F6D306b814B7F055C3 ...
> T balance of the staking contract: 0
> Increasing min authorization 40000000000000000000000 for the Random Beacon ...
> ProviderError: execution reverted: Not authorizer
>     at HttpProvider.request (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/hardhat/src/internal/core/providers/http.ts:49:19)
>     at GanacheGasMultiplierProvider.request (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/hardhat/src/internal/core/providers/gas-providers.ts:310:34)
>     at processTicksAndRejections (internal/process/task_queues.js:95:5)
> ```
> 
> One time I also got following error:
> 
> ```js
> Staking 1000000000000000000000000 to the staking provider 0x65849fc91F8289b401e9d8F6D306b814B7F055C3 ...
> T balance of the staking contract: 0
> Increasing min authorization 40000000000000000000000 for the Random Beacon ...
> undefined
> ```
> 
> Those two above errors occurred when I was running the `install.sh` with default values (`development` network) or with `--skip-client-build` flag. I also had successfull executions for those commands.

It might've been related to hardcoded timeout. We use `wait()` function now to wait for a block to be mined. Pls try again.

dimpar:
> Apart from that I noticed a problem with execution of the script on `hardhat` network - in that case script always fails on `Initializing contracts...` step, printing just `undefined` below the table:
> 
> ```
> Initializing contracts...
> No need to generate any newer typings.
>  ·--------------------------|-------------|---------------·
>  |  Contract Name           ·  Size (KB)  ·  Change (KB)  │
>  ···························|·············|················
>  |  Address                 ·      0.086  ·               │
>  ···························|·············|················
> [...]
>  ···························|·············|················
>  |  VendingMachine          ·      3.451  ·               │
>  ·--------------------------|-------------|---------------·
> undefined
> ```
> 
> I provided additional logging to the scripts and it looks to me that problem occurs on `const t: T = await helpers.contracts.getContract<T>("T")` line from `init-contracts.ts`.

Yea, it seems that Hardhat has a different configuration. I believe that this `install.sh` script and all the tasks can be adjusted, but the main goal of this PR is to make it work on Geth because we test off-chain clients on Geth. I would not configure Hardhat to run with off-chain clients in this PR. If we need to do so in the future, then it can be done in a separate PR so we do not block this one.

scripts/install.sh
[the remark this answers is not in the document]
dimpar: Ok, I've added a simple script to check if Geth is running or not. [c08fbd1](https://github.com/keep-network/keep-core/pull/3026/commits/c08fbd1117259ff04c2dd245059a55cf25acb0e3)
I would leave more sophisticated checks for another PR if needed so we do not block this one.

michalinacienciala:
I've tested the install script on the latest `keep-core@client-installation-scripts` and `solidity-contracts@local-network` code, on `development` envirinment.
Around 50% of executions of `./install.sh` or `./install.sh --skip-client-build` ends for me with:
```
Setting up staking...
Staking 1000000000000000000000000 to the staking provider 0x65849fc91F8289b401e9d8F6D306b814B7F055C3 ...
T balance of the staking contract: 1000000000000000000000000
Increasing min authorization 40000000000000000000000 for the Random Beacon ...
An unexpected error occurred:

ProviderError: authentication needed: password or unlock
    at HttpProvider.request (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/hardhat/src/internal/core/providers/http.ts:49:19)
    at GanacheGasMultiplierProvider.request (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/hardhat/src/internal/core/providers/gas-providers.ts:310:34)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

And around 50% of executions of `./install.sh --skip-deployment` ends for me with:
```
Setting up staking...
Staking 1000000000000000000000000 to the staking provider 0x65849fc91F8289b401e9d8F6D306b814B7F055C3 ...
An unexpected error occurred:

Error: cannot estimate gas; transaction may fail or may require manual gas limit (error={"name":"ProviderError","code":3,"_isProviderError":true,"data":"0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000001a50726f766964657220697320616c726561647920696e20757365000000000000"}, method="estimateGas", transaction={"from":"0x52aA72262C904281c49765499F85a774c4598850","to":"0x6ad23Caa12Bf3EB18fB26968a53c34Ba6Cca9df4","data":"0x5961d5e900000000000000000000000065849fc91f8289b401e9d8f6d306b814b7f055c300000000000000000000000052aa72262c904281c49765499f85a774c459885000000000000000000000000052aa72262c904281c49765499f85a774c459885000000000000000000000000000000000000000000000d3c21bcecceda1000000","accessList":null}, code=UNPREDICTABLE_GAS_LIMIT, version=providers/5.4.5)
    at Logger.makeError (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/@ethersproject/logger/src.ts/index.ts:225:28)
    at Logger.throwError (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/@ethersproject/logger/src.ts/index.ts:237:20)
    at checkError (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/@ethersproject/providers/src.ts/json-rpc-provider.ts:80:16)
    at EthersProviderWrapper.<anonymous> (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/@ethersproject/providers/src.ts/json-rpc-provider.ts:555:20)
    at step (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:48:23)
    at Object.throw (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:29:53)
    at rejected (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:21:65)
    at processTicksAndRejections (internal/process/task_queues.js:95:5) {
  reason: 'cannot estimate gas; transaction may fail or may require manual gas limit',
  code: 'UNPREDICTABLE_GAS_LIMIT',
  error: ProviderError: execution reverted: Provider is already in use
      at HttpProvider.request (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/hardhat/src/internal/core/providers/http.ts:49:19)
      at GanacheGasMultiplierProvider.request (/Users/michalina/workspace2/keep-core/solidity/random-beacon/node_modules/hardhat/src/internal/core/providers/gas-providers.ts:310:34)
      at processTicksAndRejections (internal/process/task_queues.js:95:5),
  method: 'estimateGas',
  transaction: {
    from: '0x52aA72262C904281c49765499F85a774c4598850',
    to: '0x6ad23Caa12Bf3EB18fB26968a53c34Ba6Cca9df4',
    data: '0x5961d5e900000000000000000000000065849fc91f8289b401e9d8f6d306b814b7f055c300000000000000000000000052aa72262c904281c49765499f85a774c459885000000000000000000000000052aa72262c904281c49765499f85a774c459885000000000000000000000000000000000000000000000d3c21bcecceda1000000',
    accessList: null
  }
}
```

dimpar:
> And around 50% of executions of ./install.sh --skip-deployment ends for me with: [..]

You get this error because stake was already created for the given owner. Please look at [this require](https://github.com/threshold-network/solidity-contracts/blob/main/contracts/staking/TokenStaking.sol#L284). If this script passed the first time, it won't pass the second time without ex. restarting the local blockchain. You have [one shot](https://www.youtube.com/watch?v=_Yhyp-_hX2s&t=37s) unless you change the input data for the `stake(..)` function.

dimpar:
> Around 50% of executions of ./install.sh or ./install.sh --skip-client-build ends for me with: [..]

Yea, it looks like that by default `unlock()` unlocks the accounts for 5min which sometimes is not enough time, and by the time we need an unlocked account it was locked again, hence the error. In this [e10974d](https://github.com/keep-network/keep-core/pull/3026/commits/e10974ded2f402d9c595391c58ca2f4b32e16ae9) commit, I changed the unlocking to the no-time limit.

michalinacienciala (approved):
I tested the script locally after the recent changes and it works fine with regards to the scope of this PR. There were some issues (like running the script on `hardhat` network), but they are outside of the scope of this PR (as explained in the comments).
My review was based mostly on results of running the script, not on code analysis, so please don't merge the PR until the code aspect gets reviewed and @nkuba's comments aren't marked as resolved.

pdyraga:
Let's also document how to set up local dev environment in a separate PR. We need something like https://github.com/keep-network/keep-core/tree/main/docs-v1/development, especially Getting Set Up and Building in `docs/` for the new version of the client.

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+  "Stakes T tokens, increases authorization and registers the operator "
+)
+  .addOptionalParam("owner", "Staking owner address", undefined, types.string)
+  .addOptionalParam("provider", "Staking provider", undefined, types.string)
+  .addOptionalParam("operator", "Staking operator", undefined, types.string)
+  .addOptionalParam("amount", "Staking amount", 1_000_000, types.int)
+  .addOptionalParam(
+    "authorization",
+    "Authorization amount",
+    undefined,
```
nkuba, line 16: 40_000 is the default minimum authorization for the beacon, and we should use it here.

```suggestion
    "authorizationAmount",
    "Authorization amount (default: minimumAuthorization)",
    undefined,
```

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+  const accounts = await new ethers.providers.JsonRpcProvider().listAccounts()
+  const stakeOwner = owner || accounts[1]
+  const stakingProvider = provider || accounts[2]
+  const stakingOperator = operator || accounts[3]
+  const beneficiary = stakeOwner
+  const authorizer = stakeOwner
+
+  const stakedAmount = to1e18(amount)
+
+  const deployerSigner = await ethers.getSigner(deployer)
```
nkuba, line 50: We could pull the deployer as signer here. 
```suggestion
  const { deployer } = await helpers.signers.getNamedSigners())
```

With this change we could remove `const { deployer } = await getNamedAccounts()` above.

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+  await (await t.connect(deployerSigner).mint(stakeOwner, stakedAmount)).wait()
+  const stakeOwnerSigner = await ethers.getSigner(stakeOwner)
```
nkuba, line 52:
```suggestion
  const stakeOwner = await ethers.getSigner(args.stakeOwner)
```

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  console.log(
+    `T balance of the staking contract: ${(
+      await t.balanceOf(staking.address)
+    ).toString()}`
+  )
+
+  const stakingAuthorization =
+    authorization || (await randomBeacon.minimumAuthorization())
+  const authorizerSigner = await ethers.getSigner(authorizer)
```
nkuba, line 75:
```suggestion
  const authorizer = await ethers.getSigner(args.authorizer)
```

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  console.log(
+    `Staked authorization ${authorizedStaked.toString()} was increased for the Random Beacon`
+  )
+
+  console.log(
+    `Registering operator ${stakingOperator.toString()} for a staking provider ${stakingProvider.toString()} ...`
+  )
+
+  const stakingProviderSigner = await ethers.getSigner(stakingProvider)
```
nkuba, line 104:
```suggestion
  const stakingProvider = await ethers.getSigner(args.stakingProvider)
```

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+  args: {
+    owner: string
+    provider: string
+    operator: string
+    amount: BigInteger
+    authorization: BigInteger
+  }
+) {
+  const { ethers, helpers, getNamedAccounts } = hre
+  const { owner, provider, operator, amount, authorization } = args
```
nkuba, line 34: Instead of destructuring the `owner`, `provider` etc. from args we could define the signers based on `args.*` as proposed in other comments.

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+  ).wait()
+
+  console.log(
+    `T balance of the staking contract: ${(
+      await t.balanceOf(staking.address)
+    ).toString()}`
+  )
+
+  const stakingAuthorization =
+    authorization || (await randomBeacon.minimumAuthorization())
```
nkuba, line 74:
```suggestion
    to1e18(args.authorization) || (await randomBeacon.minimumAuthorization())
```

solidity/random-beacon/tasks/unlock-eth-accounts.ts
```diff
@@ -0,0 +1,41 @@
...
+import { task } from "hardhat/config"
+
+import type { HttpNetworkConfig } from "hardhat/types"
+
+task("unlock-accounts", "Unlock ethereum accounts").setAction(
+  async (args, hre) => {
+    const { ethers } = hre
+
+    if (hre.network.name === "development") {
+      const password = process.env.KEEP_ETHEREUM_PASSWORD || "password"
```
nkuba, line 11: Could we use task arguments instead of process.env variables?

scripts/install.sh
```diff
@@ -7,110 +7,106 @@ DONE_START='\n\e[1;32m' # new line + bold + green
...
-printf "${LOG_START}Copying contract artifacts...${LOG_END}"
-rm -rf artifacts
-cp -r build/contracts artifacts
-npm link
+if [ "$SKIP_DEPLOYMENT" = false ] ; then
+   printf "${LOG_START}Deploying contracts...${LOG_END}"
+   # rm -rf deployments/development/
 
-printf "${LOG_START}Initializing contracts...${LOG_END}"
+   CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY=$CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY \
```
nkuba, line 94: How is `CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY` used in the process?

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+  const { to1e18 } = helpers.number
+  const t = await helpers.contracts.getContract("T")
+  const staking = await helpers.contracts.getContract("TokenStaking")
+  const randomBeacon = await helpers.contracts.getContract("RandomBeacon")
+  const accounts = await new ethers.providers.JsonRpcProvider().listAccounts()
+  const stakeOwner = owner || accounts[1]
+  const stakingProvider = provider || accounts[2]
+  const stakingOperator = operator || accounts[3]
+  const beneficiary = stakeOwner
+  const authorizer = stakeOwner
```
nkuba, line 46: IMHO we shouldn't pull the accounts from the provider and use those 1, 2, 3 for staking with this script. We should require the addresses of the accounts to be provided as arguments when running this task.

We will have to register multiple nodes, not just one.

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
+import { task, types } from "hardhat/config"
+
+import type { HardhatRuntimeEnvironment } from "hardhat/types"
+
+task(
+  "staking",
```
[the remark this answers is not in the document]
dimpar: I think "stake" suggests that it calls only a `stake()`. I wanted to name it a more general "staking" because we also invoke some functions around staking.

solidity/random-beacon/tasks/staking.ts
pdyraga: I tend to agree with @nkuba. This is supposed to be an abstract task that does not make any assumptions about the accounts available for the provider. All parameters should be supplied from the outside and then, we can reuse this task for testnet deployments as well. Right now we assume that:
- The first account available for the provider is the deployer. This will probably not be true for testnet deployments. The deployer is the account we use in our CI, and this account may not be available outside of CI, or even if it is available outside of CI, then not necessarily in the provisioning environment for testnet, or it may not be the first account there.
- There are at least 4 accounts available for the provider. In the simplest case, we need just one account.

I would always expect the owner to be provided from the outside and if we are going to make any assumptions about defaults, the only assumption I'd make is that if not provided, beneficiary == authorizer == owner. This is the default option suggested in Threshold Dashboard and I see no reason for not using it for a local development environment.

A separate complication I asked about in another comment is that I *think* installation script in the current form allows setting only one client instance. We need more and this was supported by the previous version.

To recap, what I would suggest:
- fetch all accounts available for the provider on `install.sh` level (separate task returning an array?),
- iterate over them, call "mint", "stake", "authorize", "register-operator" for all of them,
- to make it simple for local development, let's assume that stake owner == staking provider == beneficiary == authorizer == operator,
- accept all roles as parameters to individual tasks: this will allow us setting more complicated scenarios on testnet.

Opinions?

scripts/install.sh
```diff
@@ -7,110 +7,106 @@ DONE_START='\n\e[1;32m' # new line + bold + green
...
-      KEEP_CORE_CONFIG_FILE_PATH=$CONFIG_FILE \
-      CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY=$CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY \
-          npx truffle exec scripts/lcl-client-config.js --network $NETWORK
-  done
-
-  printf "${LOG_START}Building keep-core client...${LOG_END}"
-  cd $KEEP_CORE_PATH
-  go generate ./...
-  go build -a -o keep-core .
+    npx hardhat staking --network $NETWORK
```
pdyraga, line 101: Does it mean only one client is supported? We need more - with one client we cannot test network and anything that depends on at least two unique operators participating in the protocol (sortition pool selection, ECDSA key generation retries, ECDSA signing retries, broadcast channel filters, and more).

pdyraga (commented):
I tried testing the script in action but I keep getting:
```
failed to get chainId, falling back on net_version...
An unexpected error occurred:

Error: ERROR processing /Users/piotr/git/keep-network/keep-core/solidity/random-beacon/node_modules/@threshold-network/solidity-contracts/export/deploy/00_resolve_keep_registry.js:
Error: deployed KeepRegistry contract not found
    at Object.<anonymous> (/Users/piotr/git/keep-network/keep-core/solidity/random-beacon/node_modules/@threshold-network/solidity-contracts/export/deploy/00_resolve_keep_registry.js:59:27)
    at step (/Users/piotr/git/keep-network/keep-core/solidity/random-beacon/node_modules/@threshold-network/solidity-contracts/export/deploy/00_resolve_keep_registry.js:33:23)
    at Object.next (/Users/piotr/git/keep-network/keep-core/solidity/random-beacon/node_modules/@threshold-network/solidity-contracts/export/deploy/00_resolve_keep_registry.js:14:53)
    at fulfilled (/Users/piotr/git/keep-network/keep-core/solidity/random-beacon/node_modules/@threshold-network/solidity-contracts/export/deploy/00_resolve_keep_registry.js:5:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

after the unlock phase, when it's time to deploy contracts. Anything extra I should do? I also tried merging-in `main` to this branch but it did not help.

solidity/random-beacon/tasks/staking.ts
dimpar: ok, I can adjust the scripts, but to be clear, providing addresses as parameters is still available in this script and we can do so by: `npx hardhat staking --owner 0x123 --provider 0x234 --operator 0x345 --amount 424242` and it will overwrite the default ones (I assumed that would simplify development). In the `install.sh` we just need to pass these parameters to the `staking.ts`

dimpar: After talking with @nkuba we decided on the following approach:
- `install.sh` can be run once which handles fetching all the libs, unlocks, builds and deploying contracts 
- `initialize.sh` should be run for each client. If you need 3 clients, then this script should be run 3 times. `--stake-owner` param is a mandatory param. Other parameters are optional. 

[acd4aeb](https://github.com/keep-network/keep-core/pull/3026/commits/acd4aeb4f3f0c44ed64babd2070f8f20b88c8632)

scripts/pre-install.sh
```diff
@@ -0,0 +1,17 @@
...
+if [[ $isRunning == 200 || $isRunning == 403 ]]
+then
+    # Service is online
+    echo "Geth is running"
+    exit 0
+else
+    # Service is offline or not working correctly
+    echo "Geth is not running or not working correctly"
+    exit 1
+fi
```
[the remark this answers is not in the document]
dimpar: I would remove it, it's complicating things [85ed3a9](https://github.com/keep-network/keep-core/pull/3026/commits/85ed3a9b9b4fe625704c065bbde94062cc794016). I don't think we're gaining a lot by having this check.

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+  "Stakes T tokens, increases authorization and registers the operator "
+)
+  .addOptionalParam("owner", "Staking owner address", undefined, types.string)
+  .addOptionalParam("provider", "Staking provider", undefined, types.string)
```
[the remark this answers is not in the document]
dimpar: Hardhat doesn't allow for camel cases. It requires optional args to be all lowercase. Since this is a staking file and all the things here are around staking, I assumed it would be fine naming params just "owner", "provider" etc. 

Same comment for the similar comments below.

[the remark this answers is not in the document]
dimpar: Hardhat doesn't allow for camel cases. It requires optional args to be all lowercase. Since this is a staking file and all the things here are around staking, I assumed it would be fine naming params just "owner", "provider" etc. 

Same comment for the similar comments below.

scripts/pre-install.sh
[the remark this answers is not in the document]
nkuba: I agree. This script could fit great in the local-setup scripts, where we also have scripts for running the geth node.

scripts/initialize.sh
```diff
@@ -0,0 +1,114 @@
...
+done
+
+# Parse short options
+OPTIND=1
+while getopts "o:p:d:b:a:s:k:h" opt
+do
+   case "$opt" in
+      o ) stake_owner="$OPTARG" ;;
+      p ) staking_provider="$OPTARG" ;;
+      d ) staking_operator="$OPTARG" ;;
```
nkuba, line 60:
```suggestion
      d ) operator="$OPTARG" ;;
```

scripts/install.sh
```diff
@@ -1,116 +1,102 @@
...
   shift
   case "$arg" in
-    "--config-dir")     set -- "$@" "-c" ;;
-    "--network")        set -- "$@" "-n" ;;
-    "--contracts-only") set -- "$@" "-m" ;;
-    "--help")           set -- "$@" "-h" ;;
-    *)                  set -- "$@" "$arg"
+    "--network")           set -- "$@" "-n" ;;
+    "--stake-owner")       set -- "$@" "-o" ;;
+    "--staking-provider")  set -- "$@" "-p" ;;
```
nkuba, line 40: We don't need these options anymore in the install script, right?

solidity/random-beacon/hardhat.config.ts
```diff
@@ -91,6 +92,11 @@ const config: HardhatUserConfig = {
         : undefined,
       tags: ["tenderly"],
     },
+    development: {
+      url: "http://localhost:8545",
+      chainId: 1101,
+      tags: ["local"],
```
nkuba, line 98: Note that the `local` tag will change to `allowStubs` in https://github.com/keep-network/keep-core/pull/3052.

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,123 @@
+import { task, types } from "hardhat/config"
```
nkuba, line 1: Should we also rename the file to `stake.ts` to reflect the task name?

nkuba, line 1: Should we also rename the file to `stake.ts` to reflect the task name?

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,123 @@
...
+  // is used for development purposes.
+  if (beneficiary === undefined) {
+    beneficiary = owner
+  }
+
+  // Authorizer can equal to the owner if not set otherwise. This simplification
+  // is used for development purposes.
+  if (authorizer === undefined) {
+    authorizer = owner
+  }
```
nkuba, line 68:
```suggestion
  if (!beneficiary) {
    beneficiary = owner
  }

  // Authorizer can equal to the owner if not set otherwise. This simplification
  // is used for development purposes.
  if (!authorizer) {
    authorizer = owner
  }
```

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import type { HardhatRuntimeEnvironment } from "hardhat/types"
+
+task(
+  "staking",
+  "Stakes T tokens, increases authorization and registers the operator "
+)
+  .addOptionalParam("owner", "Staking owner address", undefined, types.string)
+  .addOptionalParam("provider", "Staking provider", undefined, types.string)
+  .addOptionalParam("operator", "Staking operator", undefined, types.string)
```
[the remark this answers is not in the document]
dimpar: Changed

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+  .addOptionalParam("amount", "Staking amount", 1_000_000, types.int)
```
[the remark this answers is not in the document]
dimpar: see https://github.com/keep-network/keep-core/pull/3026#discussion_r911715042

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
+import { task, types } from "hardhat/config"
+
+import type { HardhatRuntimeEnvironment } from "hardhat/types"
+
+task(
+  "staking",
+  "Stakes T tokens, increases authorization and registers the operator "
+)
+  .addOptionalParam("owner", "Staking owner address", undefined, types.string)
```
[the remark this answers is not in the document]
dimpar: changed

solidity/random-beacon/tasks/staking.ts
dimpar: done

solidity/random-beacon/tasks/staking.ts
```diff
@@ -0,0 +1,110 @@
...
+  const staking = await helpers.contracts.getContract("TokenStaking")
+  const randomBeacon = await helpers.contracts.getContract("RandomBeacon")
+  const accounts = await new ethers.providers.JsonRpcProvider().listAccounts()
+  const stakeOwner = owner || accounts[1]
+  const stakingProvider = provider || accounts[2]
+  const stakingOperator = operator || accounts[3]
+  const beneficiary = stakeOwner
+  const authorizer = stakeOwner
+
+  const stakedAmount = to1e18(amount)
```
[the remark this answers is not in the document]
dimpar: see https://github.com/keep-network/keep-core/pull/3026#discussion_r911802296

solidity/random-beacon/tasks/staking.ts
dimpar: https://github.com/keep-network/keep-core/pull/3026#discussion_r911802296

solidity/random-beacon/tasks/staking.ts
dimpar: https://github.com/keep-network/keep-core/pull/3026#discussion_r911802296

solidity/random-beacon/tasks/staking.ts
dimpar: https://github.com/keep-network/keep-core/pull/3026#discussion_r911802296

solidity/random-beacon/tasks/staking.ts
dimpar: done

scripts/install.sh
dimpar: it's removed now

scripts/install.sh
```diff
@@ -7,110 +7,106 @@ DONE_START='\n\e[1;32m' # new line + bold + green
 DONE_END='\n\n\e[0m'    # new line + reset
 
 KEEP_CORE_PATH=$PWD
-KEEP_CORE_SOL_PATH="$KEEP_CORE_PATH/solidity-v1"
+./scripts/pre-install.sh
```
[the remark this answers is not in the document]
dimpar: `pre-install.sh` is removed now

scripts/install.sh
```diff
@@ -7,110 +7,106 @@ DONE_START='\n\e[1;32m' # new line + bold + green
...
+        npx hardhat unlock-accounts --network $NETWORK
 fi
 
-printf "${LOG_START}Migrating contracts...${LOG_END}"
-rm -rf build/
+printf "${LOG_START}Building contracts...${LOG_END}"
 
-CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY=$CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY \
-    npx truffle migrate --reset --network $NETWORK
+rm -rf build && rm -rf cache && rm -rf typechain
```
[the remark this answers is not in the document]
dimpar: `yarn clean` already exists. I think there's an option to overwrite it in the `task` but this should be probably done in a separate PR. I've changed it to `yarn clean && rm -rf cache` cause `yarn clean` doesn't clean the `cache` dir.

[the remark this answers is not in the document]
nkuba: Maybe we could update it in the yarn `clean` script:
https://github.com/keep-network/keep-core/blob/084bfdf1a686d6bd51cd2db1fc07d1308514a1f1/solidity/random-beacon/package.json#L17
So we would have:
`"clean": "hardhat clean && rm -rf cache",`

scripts/initialize.sh
```diff
@@ -0,0 +1,114 @@
...
+   exit 1 # Exit script after printing help
+}
+
+# Transform long options to short ones
+for arg in "$@"; do
+  shift
+  case "$arg" in
+    "--stake-owner")          set -- "$@" "-o" ;;
+    "--staking-provider")     set -- "$@" "-p" ;;
+    "--staking-operator")     set -- "$@" "-d" ;;
```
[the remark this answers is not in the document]
dimpar: done

scripts/initialize.sh
dimpar: done

scripts/install.sh
dimpar: nope

solidity/random-beacon/tasks/staking.ts
dimpar: done, although I don't think it reflects what's inside of this script.

dimpar: https://github.com/keep-network/keep-core/pull/3026#discussion_r911824771

scripts/install.sh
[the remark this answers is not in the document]
dimpar: Yea, I guess that would be the simplest. Wasn't sure what might be the implications, but looks like all should be fine [ae96aae](https://github.com/keep-network/keep-core/pull/3026/commits/ae96aae61df7278d3b89d850c77caf5f87a3249a)

nkuba (approved):
Worked fine for me 🎉  Left some non-blocking comments

PR #3027, opened 2022-06-16
Updated Go version from 1.16 to 1.18

tomaszslabon:
This PR updates the version go Go to `1.18`:
- version `1.18` is required in the `go.mod` file
- used image with version `1.18.3` of Go in the `Dockerfile` used to build and run tests - it is the latest stable release.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.16.7-alpine3.14 AS gobuild
+FROM golang:1.18.3-alpine3.16 AS gobuild
 
 ENV GOPATH=/go \
 	GOBIN=/go/bin \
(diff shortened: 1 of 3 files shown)


lukasz-zimnoch (approved):

Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN apk add --update --no-cache \
 
 COPY --from=ethereum/solc:0.5.17 /usr/bin/solc /usr/bin/solc
 
-RUN go get gotest.tools/gotestsum
+RUN go install gotest.tools/gotestsum@latest
```
pdyraga, line 27: Should we also update:

https://github.com/keep-network/keep-core/blob/b872c35aa405b4e857fba7317aa6bd23137f5a01/scripts/macos-setup.sh#L18-L23

https://github.com/keep-network/keep-core/blob/b872c35aa405b4e857fba7317aa6bd23137f5a01/scripts/ubuntu-setup.sh#L107

?

tomaszslabon: @pdyraga 
We should, see https://github.com/keep-network/keep-core/pull/3035.

Issue #3006, opened 2022-06-02
Register in Random Beacon sortition pool


pdyraga:
The client should check if it is eligible to join the sortition pool and if so, join it.

Reference: https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/contracts/libraries/BeaconAuthorization.sol

Regarding chain-facing implementation, I suggest taking different approach for v2: We should no longer have a single `ethereum.go`. It’s too big and too complicated. Instead, we should have an Ethereum layer per module. In this case, `pkg/sortition/sortition.go` and `pkg/sortition/chain.go` and `pkg/sortition/ethereum.go`.

PR #3028, opened 2022-06-17
Register in Random Beacon's Sortition Pool

nkuba:
Closes: https://github.com/keep-network/keep-core/issues/3006
Depends on: https://github.com/keep-network/keep-core/pull/3016

In this PR we register an operator in the Random Beacon's Sortition Pool.

We introduced a `sortition` package that defines a function for registration in the sortition pool. The package defines an interface for sortition pool-related functions that should be implemented by the contract handle (in this case RandomBeacon, but later we will have to do the same for WalletRegistry).

### DIFF
--- a/pkg/chain/ethereum/_LEGACY_V1
+++ b/pkg/chain/ethereum/_LEGACY_V1
@@ -1,1 +1,1 @@
-This directory contains legacy code from V1. It should be removed.
+This directory contains legacy code from V1. It should be moved to fit the new directories strucutre.
--- a/pkg/chain/ethereum/_LEGACY_V1
+++ b/pkg/chain/ethereum/_LEGACY_V1
@@ -1,1 +1,1 @@
-This directory contains legacy code from V1. It should be moved to fit the new directories strucutre.
+This directory contains legacy code from V1. It should be moved to fit the new directories structure.
--- a/pkg/chain/ethereum/connect.go
+++ b/pkg/chain/ethereum/connect.go
@@ -335,43 +335,11 @@ func (c *Chain) TransactionMutex() *sync.Mutex {
 func fetchChainConfig(c *Chain) (*relaychain.Config, error) {
 	logger.Infof("fetching relay chain config")
 
-	groupSize, err := c.keepRandomBeaconOperatorContract.GroupSize()
-	if err != nil {
-		return nil, fmt.Errorf("error calling GroupSize: [%v]", err)
-	}
-
-	threshold, err := c.keepRandomBeaconOperatorContract.GroupThreshold()
-	if err != nil {
-		return nil, fmt.Errorf("error calling GroupThreshold: [%v]", err)
-	}
-
-	ticketSubmissionTimeout, err :=
-		c.keepRandomBeaconOperatorContract.TicketSubmissionTimeout()
-	if err != nil {
-		return nil, fmt.Errorf(
-			"error calling TicketSubmissionTimeout: [%v]",
-			err,
-		)
-	}
-
-	resultPublicationBlockStep, err := c.keepRandomBeaconOperatorContract.ResultPublicationBlockStep()
-	if err != nil {
-		return nil, fmt.Errorf(
-			"error calling ResultPublicationBlockStep: [%v]",
-			err,
-		)
-	}
-
-	relayEntryTimeout, err := c.keepRandomBeaconOperatorContract.RelayEntryTimeout()
-	if err != nil {
-		return nil, fmt.Errorf("error calling RelayEntryTimeout: [%v]", err)
-	}
-
 	return &relaychain.Config{
-		GroupSize:                  int(groupSize.Int64()),
-		HonestThreshold:            int(threshold.Int64()),
-		TicketSubmissionTimeout:    ticketSubmissionTimeout.Uint64(),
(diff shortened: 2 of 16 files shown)


pkg/chain/ethereum/_LEGACY_V1
```diff
@@ -1 +1 @@
-This directory contains legacy code from V1. It should be removed.
+This directory contains legacy code from V1. It should be moved to fit the new directories structure.
```
pdyraga, line 1: I would clear up this directory and remove everything that is not needed for v2, assuming there is a way to make the compilation pass.

pkg/chain/ethereum/connect.go
```diff
@@ -92,25 +95,24 @@ func connectWithClient(
 		)
 	}
 
-	ec := &ethereumChain{
-		config:           config,
+	c := &Chain{
```
pdyraga, line 98: Two questions:
1. Why not keep it as `ethereumChain` and have `ethereumBeaconChain` and `ethereumEcdsaChain`? I think it's easier to navigate through implementation files when it's implicitly denoted this is Ethereum implementation.
2. Why do we export it?

nkuba: This change was made in the initial implementation approach I've taken for this PR to reuse common Ethereum chain implementation in the random beacon and ecdsa packages, along with keeping it functional for the legacy code.

lukasz-zimnoch:
I'm a bit lost with the chain package structure we aim for. Especially, I don't understand the point of declaring upfront interfaces in the same package as the implementation of that interface. That seems to match more the Java-style than the Go-style interfaces. There are also some issues with that approach, for example:
- `pkg/chain/random-beacon/chain.go`: There is a `Handle` interface declared in the `beacon` package. The client code will use it as `beacon.Handle`. That suggests this is a handle to the beacon instance, not the chain subset used by the random beacon. Looks like there can be more naming-related problems like that once we add more functions and interfaces.
- Such chain interfaces with a default Ethereum implementation living nearby will be likely polluted with chain-specific types soon. Also, if we wanted to add another chain, we would be forced to add the new chain implementation in the same package. That will make the package messy
- The implementation of each chain will be scattered through the module-specific packages. For example, if we want to implement Ethereum, we need to add `ethereum.go` in `pkg/chain/random-beacon`, `pkg/chain/ecdsa` and so on. This doesn't look like a bad idea at the moment but, at the same time, we need to create an additional package for common features and import them from there. That makes the package structure more complicated. There is also a risk that we will need to do more juggling in order to avoid code duplication in the future.
- Module packages containing the domain logic like `pkg/beacon` and future `pkg/ecdsa` will not control their chain interfaces. That makes them dependent on external packages that define those chain interfaces. Also, those external packages contain a default chain implementation that can strongly affect the interface. The direction of dependency seems to be wrong. The intuition says that the domain logic should declare its own chain interface and the external implementation should depend on it.

That said, I'm wondering whether we shouldn't go with something like:
- `pkg/beacon` declares its own `chain.go` where it defines the chain interface needed by that package. Actually, it already has the `pkg/beacon/relay/chain/chain.go` file with the relay chain interface. We can flatten that a little bit and remove the `relay` sub-package.
- `pkg/ecdsa` will declare its own `chain.go` as well.
- We will have a single `pkg/chain/ethereum` package containing the entire Ethereum chain implementation split into several module-specific files like: `beacon.go`, `ecdsa.go` and containing the common features like `stake_monitor.go`, `signing.go` and so on. This is also the place where contract bindings will live. 
- We will remove `pkg/chain/chain.go`. It looks like this is an artificial grouping of the `BlockCounter`, `StakeMonitor` and `Signing` interfaces. We can push those interfaces to separate files in `pkg/chain` package or move them to `pkg/beacon` and `pkg/ecdsa` altogether. 

What are your thoughts guys?

pdyraga:
I think your proposal makes sense but I also see some advantages of keeping chain implementations per module. I think we need to do some more experimentation and then discuss one more time before we decide to go this way or another.

A couple of quick thoughts:

Yes, keeping `ethereum.go` next to `chain.go` inside a module feels weird at first. On the other hand, if we treat the module like a chain-agnostic building block providing some functionality, it could make sense. Say, we have `sortition` module with `sortition.go`, `chain.go`, `ethereum.go` and `algorand.go`. The code outside of `sortition.go` does not care about most of `sortition`'s `chain.go` functions because they are needed only in this context. Also, if there is something sortition-specific that must be included in `ethereum.go` implementation of one function, so that `ethereum.go` is not just a thin layer or read-writes, having it inside the interested module makes sense.

In my opinion, the problem with `ethereum/ethereum.go` is that it's a bag of functions without a context of how they are used and sometimes the context is needed to understand some specific logic inside of `ethereum.go`. For instance, why some call is retried and some other is not?

On the other hand, I think that instantiating one handle and then passing it everywhere is easier than instantiating multiple per-module handles that can even depend on each other. Also, using a set of common functions should be easier with a single handle, as you say.

Regardless of the decision here, I think keeping `local.go` per module makes sense. Usually, there is some logic in `local.go`  needed by the tests of the given module that does not make sense without the context. For example, why do we return an error from `JoinSortitionPool` call if eligible stake is 0 or if the operator <> staking provider mapping is not set. This is very specific to `sortition`.

pdyraga:
Closing in favor of #3056, #3067, #3070, #3077

PR #3029, opened 2022-06-20
ECDSA: Improve validation of new wallet's public key

pdyraga:
A lack of data validation on the public key could allowed the same wallet to be
added multiple times.

When a wallet is added, `self.registry[walletID].publicKeyX` was checked to be
zero in `approveDkgResult`, to ensure the `walletID` was not already used.

However, the publicKeyX of the added wallet was not checked to be non-zero. As
a result, if a wallet was added with its `publicKeyX` equal to zero, the same
wallet could be added multiple times.

The additional operations on the wallet (such as deletion and getters) check
that the wallet is registered by calling `isWalletRegistered`, which checks that
`publicKeyX` is non-zero. As a result, a wallet added with a `publicKeyX` equal
to zero will prevent most of its usage.

This problem has been reported by Trails of Bits during the security review and
is fixed here.

When working on this fix I realized our validation of a wallet's uniqueness
could be improved. Instead of checking it in `approveDkgResult`, when the
non-unique wallet passed the challenge period, we should check it in
`submitDkgResult`. This way, in an extremely unlikely case, the same group is
getting chosen and they produce the same public key, DKG will not be blocked
forever.

This change increases the gas cost of `submitDkgResult` by ~2920 gas units on
average.

Before/after:

```
|  WalletRegistry   ·  submitDkgResult  ·  285138  ·  291630  ·  287484  │
|  WalletRegistry   ·  submitDkgResult  ·  288009  ·  294525  ·  290404  │

|  WalletRegistry   ·  approveDkgResult ·  270041  ·  331267  ·  279228  │
|  WalletRegistry   ·  approveDkgResult ·  269906  ·  331108  ·  279055  │
```


solidity/ecdsa/test/WalletRegistry.WalletCreation.test.ts
```diff
@@ -1077,194 +1162,207 @@ describe("WalletRegistry - Wallet Creation", async () => {
               })
             })
 
-            context("with invalid dkg result submitted", async () => {
```
pdyraga, line 1080: The changes below are a result of adding "...and challenged" to the context description and formatting the code with new indentation.

dimpar (approved):

Issue #3004, opened 2022-06-02
Update libp2p dependency to the most recent version


pdyraga:
In #2621 we did a full update of libp2p to the most recent version as of September 2021. We need to do the same now using the most recent libp2p dependency versions.

Additionally, we need to fix a bug introduced during the work on #2621 manifesting in client tests failing from time to time:
```
=== Failed
=== FAIL: pkg/net/libp2p TestSendReceiveUnicastChannel (2.01s)
2022-05-10T20:02:35.943Z	ERROR	keep-net-libp2p	libp2p/unicast_channel.go:232	couldn't find unmarshaler for type [test/unmarshaler]
2022-05-10T20:02:35.943Z	ERROR	keep-net-libp2p	libp2p/unicast_channel.go:232	couldn't find unmarshaler for type [test/unmarshaler]
2022-05-10T20:02:35.944Z	ERROR	keep-net-libp2p	libp2p/unicast_channel.go:232	couldn't find unmarshaler for type [test/unmarshaler]
    unicast_channel_test.go:122: [peer2Receiver] unexpected number of messages
        actual:   0
        expected: 3
    unicast_channel_test.go:140: [peer2Receiver] expected message [one] not received
    unicast_channel_test.go:140: [peer2Receiver] expected message [two] not received
    unicast_channel_test.go:140: [peer2Receiver] expected message [three] not received
    unicast_channel_test.go:93: unexpected number of `OnUnicastChannelOpened` invocations
        actual:   0
        expected: 1
```

PR #3031, opened 2022-06-21
Fix for unicast channel related error

tomaszslabon:
Refs https://github.com/keep-network/keep-core/issues/3004.
This PR adds a fix for failing unit tests related to unicast channels.

Inside `TestSendReceiveUnicastChannel` we [send messages between peer 1 and peer 2](https://github.com/keep-network/keep-core/blob/333156b595e005322f06de2c8bda45cf822ff87a/pkg/net/libp2p/unicast_channel_test.go#L58).

Then [we retrieve the channel that should be created by the sent messages](https://github.com/keep-network/keep-core/blob/333156b595e005322f06de2c8bda45cf822ff87a/pkg/net/libp2p/unicast_channel_test.go#L66).

If the sent messages are received by peer 2 before the above call, it created the channel properly and sets the unmarshaller to handle the messages. In that case [the call](https://github.com/keep-network/keep-core/blob/333156b595e005322f06de2c8bda45cf822ff87a/pkg/net/libp2p/unicast_channel_test.go#L66) retrieves the channel and the test passes.

If the sent messages are received by peer 2 after the above call, the unmarshaller is not set and [the call](https://github.com/keep-network/keep-core/blob/333156b595e005322f06de2c8bda45cf822ff87a/pkg/net/libp2p/unicast_channel_test.go#L66) creates the channel instead of retrieving it and the channel has no unmarshaller set and the test fails.

The simplest solution is to wait for a moment before retrieving the channel inside `TestSendReceiveUnicastChannel`.

### DIFF
--- a/pkg/net/libp2p/unicast_channel_test.go
+++ b/pkg/net/libp2p/unicast_channel_test.go
@@ -63,6 +63,9 @@ func TestSendReceiveUnicastChannel(t *testing.T) {
 		}
 
 		// Peer 2 get the channel from cache.
+		// Wait for a moment until the messages sent by peer 1 are received by
+		// peer 2, so that the channel is opened and unmarshaller set.
+		time.Sleep(2 * time.Second)
 		peer2Channel, err := provider2.UnicastChannelWith(identity1.id)
 		if err != nil {
 			t.Fatal(err)


pkg/net/libp2p/unicast_channel_test.go
```diff
@@ -63,6 +63,9 @@ func TestSendReceiveUnicastChannel(t *testing.T) {
 		}
 
 		// Peer 2 get the channel from cache.
+		// Wait for a moment until the messages sent by peer 1 are received by
+		// peer 2, so that the channel is opened and unmarshaller set.
+		time.Sleep(2 * time.Second)
```
lukasz-zimnoch, line 68: Why this is exactly 2 seconds?

tomaszslabon: I think another value like 1 s 0r 0,5 s would work as well.
I chose 2 seconds because we already had such a sleep time in `unicast_channel_test.go` and it worked well.

lukasz-zimnoch (approved):

Issue #3013, opened 2022-06-03
Random Beacon DKG test loop


pdyraga:
The client should perform Random Beacon DKG every 100 Ethereum blocks and store the result on disk. It should not submit the result to the chain. This is a test loop for the early milestone.

Issue #3012, opened 2022-06-03
Internal development testnet for Random Beacon and TBTC v2


pdyraga:
We need an internal testnet for development builds. 
We will build our test environment against **Goerli**.

Prerequisites:

- [ ] Geth testnet node with testnet ether mining (geth) (@electropolis)
  - [x] Install [geth v1.10.20](https://geth.ethereum.org/docs/install-and-build/installing-geth) on Virtual Machine and start syncing with Goerli (just PoC). (Flags to use: --goerli, --syncmode=snap, --http, --ws) Consider using ethereum/client-go docker image for simple installation.
  - [x] Run Goerli geth on kubernetes:
    - [x] StatefulSet
    - [x] Persistent storage
    - [x] Expose ports (HTTP, WS, P2P)
    - [x] Internal service
    - [ ] External service (extra task)
    - [x] #3264
- [x] Deploy V1 contracts to testnet manually and publish them to NPM Package Registry to use as hardcoded dependencies in V2 (@nkuba) - `@keep-network/keep-core@1.8.1-goerli.0` https://github.com/keep-network/keep-core/pull/3059
- Manually deployed and published V2 contracts (@nkuba):
  - [x] Threshold Network (`@threshold-network/solidity-contracts@1.2.0-goerli.0`)
  - [x] Keep Random Beacon (`@keep-network/random-beacon@2.0.0-goerli.0`)
  - [x] Keep ECDSA (`@keep-network/ecdsa@2.0.0-goerli.0`)
  - [x] Keep TBTC v2 (`@keep-network/tbtc-v2@0.1.1-goerli.0`)
- [x] Automate deployment of V2 contracts (@michalinacienciala):
  - [x] CI config update: https://github.com/keep-network/ci/pull/28
  - [x] Update of workflows in `solidity-contracts`: https://github.com/threshold-network/solidity-contracts/pull/114
  - [x] Update of workflows in `keep-core`: https://github.com/keep-network/keep-core/pull/3081
  - [x] Update of workflows in `tbtc-v2`: https://github.com/keep-network/tbtc-v2/pull/382
  - [ ] ~~Replace the [stub of the BitcoinRelay contract](https://github.com/keep-network/tbtc-v2/pull/387/commits/17cb5e533008ea14d79e4d1db1b1ed83db3b3d37)~~ - moved to an independent issue (https://github.com/keep-network/tbtc-v2/issues/400)
- [x] Configure contracts references in the client (@nkuba)
[...]

PR #3034, opened 2022-06-24
DKG test loop

lukasz-zimnoch:
Refs: #3013 

This pull request detaches the existing DKG mechanism from Random Beacon v1 contracts and introduces a DKG test loop that is triggered every 500th block and stores the result on disk. That test loop is an interim mechanism that will allow us to integrate with Random Beacon v2 contracts soon. All details about v1 and v2 DKG along with the test loop description are captured below.

### DKG in Random Beacon V1

The DKG v1 flow is as follows:
1. The `KeepRandomBeaconOperator` contract emits the `GroupSelectionStarted` event that contains a seed value
2. The client reacts to the `GroupSelectionStarted` event by starting the ticket submission process using the seed value held by the event.
3. Once the ticket submission ends, the client checks its eligibility by calling the `selectedParticipants` on-chain function.
4. If the client is eligible for the given DKG instance, i.e controls some of the group members, it starts the DKG execution.
5. Once the group public key is generated, the client attempts to publish it to the `KeepRandomBeaconOperator` contract and monitors for `DkgResultSubmittedEvent` event to confirm the result publication.
6. Once the key is published, the client stores the key material on disk and holds the group participation info in an in-memory group cache to be ready for incoming relay entry requests.

### DKG in Random Beacon V2

In Random Beacon V2, things happen differently:
1. The `RandomBeacon` contract emits the `DkgStarted` event that contains a seed value.
2. The client checks its eligibility by calling the `selectGroup` on-chain function exposed by the `RandomBeacon` or directly by the `SortitionPool` contract.
3. If the client is eligible for the given DKG instance, i.e controls some of the group members, it starts the DKG execution.
5. Once the group public key is generated, the client attempts to publish it to the `RandomBeacon` contract and monitors for `DkgResultSubmitted` event to confirm the result publication.
6. Once the key is published, the client stores the key material on disk. However, since v2 uses an optimistic DKG result validation with a challenge mechanism, the client cannot consider the group as fully-operable immediately. It must wait for the result's approval. To handle this, the client must store the key material aside and avoid adding the group participation info to the in-memory group cache right after result publication.

### Interim DKG test loop introduced by this PR

We decided to move from v1 DKG to v2 DKG incrementally. To do so, we made the following steps:
- We removed everything related to the current group selection (`pkg/beacon/relay/groupselection`) and ticket submission logic, as it is no longer relevant for v2. 
- We detached the DKG mechanism from the v1 `KeepRandomBeaconOperator` contract by mocking the contact points between the DKG logic and the contract binding. This resulted in several modifications in the `pkg/beacon/relay/chain/chain.go` interfaces: 
    - The `OnGroupSelectionStarted` event listener was removed and replaced by the `OnDKGStarted` event listener with a temporary implementation that emits the `DkgStarted` event every 500th block. 
    - The `SubmitDKGResult` implementation was removed and replaced with a local implementation that pipes the result directly to the handlers registered by the `OnDKGResultSubmitted` event listener 
    - The `OnDKGResultSubmitted` implementation was removed and replaced with a local implementation that registers a handler for results coming from `SubmitDKGResult`
    - There is a new function `SelectGroup` that replaces the removed group selection logic. It has a temporary implementation that returns a group consisting of `N` group members controlled by the calling operator
- We changed the event deduplicator logic for `DkgStarted` events. It's now a seed-based time cache.


pkg/beacon/relay/event/deduplicator.go
```diff
@@ -24,54 +24,22 @@ type chain interface {
...
-		newGroupSelectionStartBlock > minCurrentGroupCreationEndBlock
-
-	if shouldUpdate {
-		d.currentGroupSelectionStartBlock = newGroupSelectionStartBlock
-		return true
-	}
-
-	return false
-}
-
```
pdyraga, line 74: Quoting from the PR description:

> We removed the event deduplicator for the group selection events as it is no longer relevant for v2. We will need a new logic to handle v2 caveats. This was covered by a TODO in beacon.go

I am not sure if this is the case. As far as I remember, the deduplicator was mostly handling soft forks. We were receiving the same event twice but with a different block number: `n` and then `n+1`. The implementation is quite naive: it checks if the group selection had a chance to finish since the last emitted group selection started event and if not, it was considering the event as a duplicate of the last one.

Although this is naive, and we could improve it by, for example, waiting for N blocks, then checking the last N blocks if the event is included, I think the current implementation should do the work, even for v2.

pkg/beacon/relay/chain/chain.go
```diff
@@ -52,21 +55,10 @@ type RelayEntryInterface interface {
...
-	SubmitTicket(ticket *Ticket) *async.EventGroupTicketSubmissionPromise
-	// GetSubmittedTickets gets the submitted group candidate tickets so far.
-	GetSubmittedTickets() ([]uint64, error)
-	// GetSelectedParticipants returns `GroupSize` slice of addresses of
-	// candidates which have been selected to the currently assembling group.
-	GetSelectedParticipants() ([]StakerAddress, error)
+	// SelectGroup returns the group members for the group generated by
+	// the given seed. This function can return an error if the relay chain's
+	// state does not allow for group selection at the moment.
+	SelectGroup(seed *big.Int) ([]StakerAddress, error)
```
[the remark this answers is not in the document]
pdyraga: Just a note for the future implementation with a real `SortitionPool`: fortunately, [there is a function](https://github.com/keep-network/sortition-pools/blob/b3fcb43e09d9fc2304fa2f65635aec406a6cfe79/contracts/SortitionTree.sol#L78-L97) allowing to fetch the addresses in one shot.

pkg/beacon/relay/node.go
```diff
@@ -43,51 +42,67 @@ func (n *Node) IsInGroup(groupPublicKey []byte) bool {
 	return len(n.groupRegistry.GetGroup(groupPublicKey)) > 0
 }
 
-// JoinGroupIfEligible takes a threshold relay entry value and undergoes the
-// process of joining a group if this node's virtual stakers prove eligible for
-// the group generated by that entry. This is an interactive on-chain process,
-// and JoinGroupIfEligible can block for an extended period of time while it
+// JoinDKGIfEligible takes a seed value and undergoes the process of the
+// distributed key generation if this node's operator proves eligible for
```
pdyraga, line 46:
```suggestion
// distributed key generation if this node's operator proves to be eligible for
```

I think it should be *to be*. It does not come from this PR but we can fix it. Nothing blocking, of course.

pkg/beacon/relay/node.go
```diff
@@ -122,23 +137,39 @@ func (n *Node) JoinGroupIfEligible(
 					return
 				}
 
-				// final broadcast channel name for group is the compressed
-				// public key of the group
+				// The final broadcast channel name for group is the compressed
+				// public key of the group.
 				channelName := hex.EncodeToString(
```
pdyraga, line 142: Can we rename it to `groupPublicKey` or something similar? Even if `groups.go` expects the channel name, we can pass `groupPublicKey` as a channel name and in the context of `node.go` we do not expect the channel but the group public key to be logged.

pkg/beacon/relay/registry/groups_test.go
```diff
@@ -197,6 +221,12 @@ func (mgri *mockGroupRegistrationInterface) OnGroupRegistered(
 	panic("not implemented")
 }
 
+func (mgri *mockGroupRegistrationInterface) IsGroupRegistered(
```
pdyraga, line 224: Is it used anywhere?

pkg/beacon/relay/registry/storage.go
```diff
@@ -2,15 +2,21 @@ package registry
...
 	"sync"
 
 	"github.com/keep-network/keep-common/pkg/persistence"
 
 	"encoding/hex"
 )
 
+const (
+	candidateMembershipPrefix = "candidate_membership"
+	approvedMembershipPrefix  = "approved_membership"
```
pdyraga, line 15: I am thinking if we shouldn't use the [snapshot](https://github.com/keep-network/keep-ecdsa/blob/bd9800d7367aa665b745c7149e1040a3b16deaf0/pkg/registry/storage.go#L14) mechanism from ECDSA node. This is something our operators are familiar with and I think it makes sense here and in ECDSA:
- any key material produced by the node lands in `snapshot`
- when the group is approved, we copy from `snapshot` to `current` (this would be new)

We would need to "snapshot" the state anyway, and I think adding `candidate`/`approved` does not make sense if the snapshot is already there (this is essentially candidate).

Instead of passing the `GroupState` as a parameter, I would "snapshot" the signer as soon as the key material is produced. Then, once the group is approved on-chain, I would copy from the `snapshot` to the `current`. This way, we do not have to worry if the node has been restarted during the challenge period. Also, the "snapshot" would not store the group in the in-memory registry. This is exactly [how it is working](https://github.com/keep-network/keep-ecdsa/blob/bd9800d7367aa665b745c7149e1040a3b16deaf0/pkg/registry/keeps.go#L67-L73) in ECDSA.

pkg/beacon/relay/event/deduplicator.go
lukasz-zimnoch: > As far as I remember, the deduplicator was mostly handling soft forks

Not only soft forks. The event subscription mechanism from contract bindings can replay the event when fetching past events. This happens more often than soft forks and results in an event having the same block.

>  I think the current implementation should do the work, even for v2.

This is something I'm not sure about. In v2 the minimum time between two `DkgStarted` events is `DKG protocol time + publication time + challenge period time`. In theory, we can expose `NotifyGroupCreationStarted` function from the event deduplicator to decide about group creation start. However, at the same time, we must be prepared for `DkgResultChallenged` events that actually restart the process and make the time between two `DkgStarted` events longer. The deduplicator must handle that. This is why I think we need to approach deduplication logic once we have challenges and approvals implemented together.

pkg/beacon/relay/registry/storage.go
lukasz-zimnoch: > This is something our operators are familiar with and I think it makes sense here and in ECDSA:
    - any key material produced by the node lands in snapshot
    - when the group is approved, we copy from snapshot to current (this would be new)

Well, this is not that simple. The snapshot mechanism is write-only. There is no logic in the code to read from it. This is a last-resort mechanism for manual restoration in case something bad happens. What ECDSA does is rather:

- any key material produced by the node lands in snapshot
- when the public key is published on-chain it is just normally stored in the regular storage (snapshot is untouched)

If the client shuts down between those steps, the key would not be published at all. That works well in ECDSA because the time between a snapshot and a regular save is quite short. In beacon v2, the time between result publication and approval (the challenge period) will be much longer so the client restart risk is much higher. If we use snapshot on result publication + store on approval, we don't have automatic restore after the client restarts out of the box. This is a problem because the client must be aware of what candidate groups he is a member of.
In the current approach, we would address this by adding an in-mem cache for candidate groups once we add the result approvals and challenges feature. If we want to go with snapshots, we'll need to make a huge refactor of the entire storage code. This can be done but then, I propose to do it in a separate PR.

pkg/beacon/relay/chain/chain.go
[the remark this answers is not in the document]
lukasz-zimnoch: Yep, there is a comment about the deprecation of `StakerAddress`. I can replace it with `chain.Address` during the cleanup, I plan to do in the `pkg/beacon` package.

pkg/beacon/relay/node.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3034/commits/e49262d3683af4ad25eee0a45e031d7f1377770c

pkg/beacon/relay/node.go
lukasz-zimnoch: See 87e9b770

pkg/beacon/relay/registry/groups_test.go
lukasz-zimnoch: Yep, in `TestUnregisterStaleGroups` and `TestUnregisterStaleGroupsSkipLastGroupCheck`

pkg/chain/ethereum/ethereum.go
```diff
@@ -332,28 +236,66 @@ func (ec *ethereumChain) GetGroupMembers(groupPublicKey []byte) (
 	return stakerAddresses, nil
 }
 
+// TODO: Implement a real DkgStarted event subscription once it is possible
+//       on the contract side. The current implementation generate a fake
+//       event every 100th block where the seed is the keccak256 of the
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3034/commits/185fe7fb81d11d49c63b8e092be375deebdd9af0

pkg/beacon/relay/registry/storage.go
pdyraga: The solution I proposed requires refactoring of storage. The snapshot would still be read-only but we would need to be able to copy from the snapshot.

On a high level, the idea is not to load candidate members to the memory - even a separate cache - and instead copy the directory when a group is approved. I think this is simpler and has the advantage that copy is one disk write operation, we do not need to write in the loop which I think is less risky.

What I do not like in the current implementation is that looking at just `current` directory it is not possible to say which groups were approved and which were not. We are also mixing candidate key shares with approved key shares which is not the end of the world but I think we could make it cleaner.

If we want to treat snapshots as something special we write only one time and never touch them again, I think it's even better because by obeying this rule in the code, we avoid corrupting snapshot shares in case of a programming error. Looking at the storage level, we would have:

```
snapshot/
  0xAAA
  0xBBB
  0xCCC
current/
  candidate_0xAAA/
    membership_1
    membership_21
  0xBBB
    membership_33
    membership_55
  0xCCC
     membership_64
``` 

Then, when `0xAAA` gets approved, we `mv current/candidate_0xAAA current/0xAAA` and we finally land with:
```
snapshot/
  0xAAA
  0xBBB
  0xCCC
current/
  0xAAA/
    membership_1
    membership_21
  0xBBB
    membership_33
    membership_55
  0xCCC
     membership_64
```

Once the directory is renamed, we load all the members from `0xAAA` into the memory.

If we are fine with this layout, I would propose something like:

`groups.go`
```
// RegisterCandidateGroup registers that a candidate group with the given
// public key was successfully created.
func (g *Groups) RegisterCandidateGroup(
	signer *dkg.ThresholdSigner,
	groupId string,
) error {
	g.mutex.Lock()
	defer g.mutex.Unlock()

	groupPublicKey := groupKeyToString(signer.GroupPublicKeyBytes())

	membership := &Membership{
		Signer:      signer,
		ChannelName: groupId,
	}

	err := g.storage.saveCandidate(membership, groupID)
	if err != nil {
		return fmt.Errorf("could not persist membership to the storage: [%v]", err)
	}

	if groupState == GroupStateApproved {
		g.myGroups[groupPublicKey] = append(g.myGroups[groupPublicKey], membership)
	}

	return nil
}

// ApproveGroup registers that a candidate group with the given
// public key was successfully approved.
func (g *Groups) ApproveGroup(
	groupId string,
) error {
	err := g.storage.approveCandidate(groupId)
	if err != nil {
		return fmt.Errorf("could not promote group to approved in the storage: [%v]", err)
	}
     
        loadGroup(groupId)
}
```

`storage.go`
```
func (ps *persistentStorage) saveCandidate(
	membership *Membership,
	groupID string // use this as a directory name
) error {
  // (...)
}

func (ps *persistentStorage) approveCandidate(
	groupID string // use this as a directory name
) error {
  // (...)
}

func (ps *persistentStorage) read(groupID string) (<-chan *Membership, <-chan error) {
 // (...) load a single group
}

// Alternatively, we can combine approveCandidate(groupID string) with read(groupID string):
func (ps *persistentStorage) approveCandidate(
	groupID string // use this as a directory name
) (<-chan *Membership, <-chan error)  {
  // (...)
}
```

However, this is a bit of work and as you say, it's not the best idea to do it in this PR. If we agree with this approach, I'd clean up the code in this PR and prepare for the new version:
1. Get rid of `GroupState` and candidate/approved prefixes.
2. Rename `save` in `storage.go` to `saveCandidate`, accept `groupID string`, and prepend `candidate_` to group identifier when saving.
3. In `groups.go`, have only `RegisterCandidateGroup`.
4. Modify `storage.go` `readAll` to read candidate groups from disk. Make it a temporary change, until we refactor the storage. In a separate PR refactoring the storage, we should clean this up to read only approved groups and temporarily, approve the group right away, after it's saves as a candidate.

WDYT?

pkg/beacon/relay/event/deduplicator.go
pdyraga: > Not only soft forks. The event subscription mechanism from contract bindings can replay the event when fetching past events. This happens more often than soft forks and results in an event having the same block.

Yeah, that's a good point!

The time it takes the DKG protocol to produce the result is 72 blocks ([[1](https://github.com/keep-network/keep-core/blob/9f8d78ec3b7c91938db79dc9cc5882fec8ef49e9/solidity/random-beacon/contracts/libraries/BeaconDkg.sol#L117)], [[2](https://github.com/keep-network/keep-core/blob/9f8d78ec3b7c91938db79dc9cc5882fec8ef49e9/pkg/beacon/relay/gjkr/states.go#L13-L37)]). Subscription monitoring mechanism pulls 100 past blocks [by default](https://github.com/keep-network/keep-common/blob/58daa5270187dea35e7dd3642cc3541ecd45e0f6/pkg/chain/ethereum/ethutil/subscribe_opts.go#L12-L15) and although it could make sense for the relay request subscription, it does not make sense for DKG because after the first 6 blocks of inactivity, the game for the given member is over.

This solution is not ideal but at the same time, it's the simplest one. I am afraid that dealing with challenges will make this code pretty complicated.

I *think* that this should work:
1. Leave the deduplicator for DKG as-is, and removed just `currentGroupSelectionStartBlock`.
2. Make it clear `NotifyGroupSelectionStarted` filters out just for `minGroupCreationDurationBlocks` to the past.
3. When setting up a subscription for DKG, disable subscription monitoring loop. For 6 blocks, we would need to monitor pretty aggressively (every 10 seconds or so).

Does it make sense?

pkg/beacon/relay/registry/storage.go
```diff
@@ -25,15 +31,32 @@ func newStorage(persistence persistence.Handle) storage {
...
 	membershipBytes, err := membership.Marshal()
 	if err != nil {
 		return fmt.Errorf("marshalling of the membership failed: [%v]", err)
 	}
 
-	hexGroupPublicKey := hex.EncodeToString(membership.Signer.GroupPublicKeyBytesCompressed())
+	directory := hex.EncodeToString(membership.Signer.GroupPublicKeyBytesCompressed())
+
+	var membershipPrefix string
+	switch groupState {
```
dimpar, line 46: Not sure what will be the final solution around prefixes, but I think one way of not relying on the `switch` might be something like this:
```
// GroupState determines the given group state.
type GroupState string

// Enum that denotes possible group states.
const (
	GroupStateCandidate GroupState = "candidate"
	GroupStateApproved = "approved"
)
```

and then in the `storage.go` we can replace the `switch` with:

```
membershipPrefix := fmt.Sprintf("%v_%v", groupState, "membership")
```

Just an idea, unless we need integers for the state as our "enums".

pkg/beacon/relay/event/deduplicator.go
pdyraga: We synced with @lukasz-zimnoch and came up with a better solution:
1. When setting up a subscription for DKG, disable subscription monitoring loop. 
2. Use a time-cache-based deduplicator and look at the DKG seed from the event.

pkg/beacon/relay/node.go
lukasz-zimnoch: See 5ac87dfa

pkg/beacon/relay/registry/storage.go
lukasz-zimnoch: See 641466e7

pkg/beacon/relay/chain/chain.go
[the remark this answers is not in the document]
lukasz-zimnoch: There will be a follow-up PR that will add a real-world implementation of that function. I'll consider the rename then.

pkg/beacon/relay/registry/storage.go
pdyraga: Keeping this thread as unresolved so that we can refer it back when reworking the storage.

pkg/beacon/relay/event/deduplicator.go
lukasz-zimnoch: Here it is fbe3062b

pdyraga:
There are some problems with running the client without v1 contracts deployed and initialized.

In #3026 we reworked v1 scripts into v2 approach so it is no longer possible to set up v1 contracts locally on the dev environment. Also, the work we are doing in #3012 will set up testnet based on v2 contracts.

We need to find a way to run a client now. I tried locally and I think we can do some workarounds for M1:
- Disable `ec.initializeBalanceMonitoring(ctx)` in `connect.go`.
- Rework `fetchChainConfig` in `connect.go` to return hardcoded values.
- Disable `stakeMonitor.HasMinimumStake` check in `start.go`.

pkg/beacon/relay/event/deduplicator.go
lukasz-zimnoch: And added a TODO about disabling the monitoring loop 7c122d51

pdyraga:
One more note, I think we shouldn't close issue #3013 with merging this PR. We still need to integrate with Sortition Pool group selection and ensure we can run this loop over the network with multiple clients.

pdyraga:
There is one more problem with v1 contracts not deployed. The DKG errors out with:

```
2022-07-06T13:54:39.750+0200	WARN	keep-dkg	go-log@v1.0.5/log.go:180	[member:22] DKG result publication process failed [failed to initiate new state [could not check if the result is already submitted: [response [[]] was not long enough to interpret while resolving original error [no contract code at given address]]]]
2022-07-06T13:54:39.892+0200	ERROR	keep-relay	relay/node.go:136	failed to execute dkg: [DKG result publication timed out]
```

I would consider disabling `dkgResult.Publish` and `decideMemberFate` in `dkg.go`. We could theoretically short-circuit them but for v2 we need more sophisticated behavior when it comes to the result publication.

lukasz-zimnoch:
> There is one more problem with v1 contracts not deployed. The DKG errors out with:
> 
> ```
> 2022-07-06T13:54:39.750+0200	WARN	keep-dkg	go-log@v1.0.5/log.go:180	[member:22] DKG result publication process failed [failed to initiate new state [could not check if the result is already submitted: [response [[]] was not long enough to interpret while resolving original error [no contract code at given address]]]]
> 2022-07-06T13:54:39.892+0200	ERROR	keep-relay	relay/node.go:136	failed to execute dkg: [DKG result publication timed out]
> ```
> 
> I would consider disabling `dkgResult.Publish` and `decideMemberFate` in `dkg.go`. We could theoretically short-circuit them but for v2 we need more sophisticated behavior when it comes to the result publication.

Here it is https://github.com/keep-network/keep-core/pull/3034/commits/10e12b9ecaaf6b1b60eae7f1ed62f4a12213a51e

pdyraga (approved):

PR #3051, opened 2022-06-30
Move go libp2p bootstrap under keep core

tomaszslabon:
This PR moves `go-libp2p-bootstrap` functionalities under `keep-core`.
The bootstrap functionalities were moved under `libp2p` package.

### DIFF
--- a/pkg/net/libp2p/bootstrap.go
+++ b/pkg/net/libp2p/bootstrap.go
@@ -85,7 +85,13 @@ func Bootstrap(
 	if rt != nil {
 		ctx := goprocessctx.OnClosingContext(proc)
 		if err := rt.Bootstrap(ctx); err != nil {
-			proc.Close()
+			if procErr := proc.Close(); procErr != nil {
+				return nil, fmt.Errorf(
+					"process closing error [%v] while handling bootstrap error [%v]",
+					procErr,
+					err,
+				)
+			}
 			return nil, err
 		}
 	}


pkg/net/libp2p/bootstrap.go
```diff
@@ -0,0 +1,223 @@
...
+		if err := rt.Bootstrap(ctx); err != nil {
+			if procErr := proc.Close(); procErr != nil {
+				return nil, fmt.Errorf(
+					"process closing error [%v] while handling bootstrap error [%v]",
+					procErr,
+					err,
+				)
+			}
+			return nil, err
+		}
```
tomaszslabon, line 96: Notice that due to `client-scan` errors I had to change this fragment from:
```
		if err := rt.Bootstrap(ctx); err != nil {
			proc.Close()
			return nil, err
		}
```
to 
```
		if err := rt.Bootstrap(ctx); err != nil {
			if procErr := proc.Close(); procErr != nil {
				return nil, fmt.Errorf(
					"process closing error [%v] while handling bootstrap error [%v]",
					procErr,
					err,
				)
			}
			return nil, err
		}
```
See https://github.com/keep-network/keep-core/pull/3051/commits/b5711bff7b7367f96e69db6a4ff07572cf5a0f55.

Also I had to change some function and variable name to lowercase to make them unexported.
See https://github.com/keep-network/keep-core/pull/3051/commits/c162488093d2c53cce8a8dde31b47c9d35dc5c3c.

lukasz-zimnoch (approved):

PR #3052, opened 2022-06-30
Update @threshold-network/solidity-contracts dependency in Beacon

nkuba:
Updated @threshold-network/solidity-contracts dependency to the latest
published development package.

### DIFF
--- a/solidity/random-beacon/hardhat.config.ts
+++ b/solidity/random-beacon/hardhat.config.ts
@@ -76,12 +76,17 @@ const config: HardhatUserConfig = {
           testConfig.nonStakingAccountsCount +
           testConfig.stakingRolesCount * testConfig.operatorsCount,
       },
-      tags: ["local"],
       // we use higher gas price for tests to obtain more realistic results
       // for gas refund tests than when the default hardhat ~1 gwei gas price is
       // used
       gasPrice: 200000000000, // 200 gwei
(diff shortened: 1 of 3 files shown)


solidity/random-beacon/hardhat.config.ts
```diff
@@ -76,12 +76,17 @@ const config: HardhatUserConfig = {
           testConfig.nonStakingAccountsCount +
           testConfig.stakingRolesCount * testConfig.operatorsCount,
       },
-      tags: ["local"],
       // we use higher gas price for tests to obtain more realistic results
       // for gas refund tests than when the default hardhat ~1 gwei gas price is
       // used
       gasPrice: 200000000000, // 200 gwei
       allowUnlimitedContractSize: true,
+      tags: ["allowStubs"],
```
dimpar, line 84: This PR needs for testing https://github.com/keep-network/keep-core/pull/3026 but we should probably also update `/ecdsa` with the same tags.

dimpar (approved):

solidity/random-beacon/hardhat.config.ts
nkuba: ECDSA has to happen in a separate PR as we need to update the dependency to the random beacon there.

https://github.com/keep-network/keep-core/pull/3053

PR #3056, opened 2022-07-05
Interface and local implementation for sortition chain

pdyraga:
Refs #3006 

This change is greatly inspired by the work of @nkuba in https://github.com/keep-network/keep-core/pull/3028 but it is just a subset of the original work with some changes introduced.

First, `sortition.Chain` handle interface is introduced. In opposite to v1 approach, where we had a global chain handle, the new proposed approach is to keep a chain handle per module. This should make the chain code less complicated and allow introducing custom behavior to functions, if needed be the logic of the given module.

Second, `chain.Address` type based on `string` is introduced. We need a chain-agnostic implementation of a chain address to use in interfaces. So far, it's been an array of bytes, as presented in `Signing` interface, but `string` seems easier to use in most cases.

Last but not least, a local implementation of `sortition.Chain` handle is introduced.

Also, a set of simple test utils for assertions is added. Most of the time, such simple assertions are all we need, and having them extracted to a single place lets us avoid the assertion error message formatting boilerplate in test files. Long-term, if needed, we may use some more sophisticated assertion library.


pkg/sortition/chain.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/keep-network/keep-core/pkg/chain"
+)
+
+// Chain handle for interaction with the sortition pool contracts.
+type Chain interface {
+	OperatorToStakingProvider() (chain.Address, bool, error)
+	EligibleStake(stakingProvider chain.Address) (*big.Int, error)
+	IsPoolLocked() (bool, error)
+	IsOperatorInPool() (bool, error)
+	JoinSortitionPool() error
```
lukasz-zimnoch, line 15: Non-blocking for this PR but can we document those functions?

lukasz-zimnoch (approved):

PR #3057, opened 2022-07-05
Use contracts stubs only for unit tests

nkuba:
For unit tests, we want to deploy some stubs, but we don't want to use them if we deploy the contracts as dependencies in other projects. To explicitly request for stubs deployment we provide `TEST_USE_STUBS_*` flag and use it in unit tests only.


solidity/random-beacon/package.json
```diff
@@ -16,7 +16,7 @@
   "scripts": {
     "clean": "hardhat clean && rm -rf cache",
     "build": "hardhat compile",
-    "test": "hardhat check-accounts-count && hardhat test",
+    "test": "TEST_USE_STUBS_BEACON=true hardhat check-accounts-count && hardhat test",
```
dimpar, line 19:
```suggestion
    "test": "hardhat check-accounts-count && TEST_USE_STUBS_BEACON=true hardhat test",
```

dimpar: How about we unify the namings when we use contract stubs? What I mean is that in the `tags` we have `allowStubs`, here we have `..USE_STUBS_..`. WDYT if we have something like `tags: ["contractStubs"]` and `"test": "CONTRACT_STUBS_BEACON=true hardhat test" / "CONTRACT_STUBS_ECDSA=true hardhat test"`?

nkuba: To be honest I like the current solution better as the names explain pretty well what they mean.

dimpar:
@nkuba build fails. I think we should split this PR into 2 to make the build pass, ie one for `/random-beacon` changes and another one for `/ecdsa`, because it has a `/random-beacon` dependency in it. `/ecdsa` doesn't see the latest changes in `/random-beacon` and it runs out of gas

nkuba:
> @nkuba build fails. I think we should split this PR into 2 to make the build pass, ie one for `/random-beacon` changes and another one for `/ecdsa`, because it has a `/random-beacon` dependency in it. `/ecdsa` doesn't see the latest changes in `/random-beacon` and it runs out of gas

The checks are passing now.

dimpar (dismissed):

Issue #3010, opened 2022-06-03
Update network firewall to check for non-zero stake in Threshold TokenStaking


pdyraga:
Modify the firewall to check if someone has a stake delegation in the Threshold TokenStaking contract. This step also requires resolving operator-to-staking-provider mapping in `RandomBeacon` and `WalletRegistry`. The resolution and check should be done in the stake monitor.

We do not want to require the minimum stake to avoid problems with scanning all ECDSA wallets in case someone got slashed. (Note, we had this problem in v1.) Known delegation and existing operator-staking-provider mapping mean someone has or had a minimum stake and the minimum authorization required by the application.

Reference: https://github.com/keep-network/keep-core/blob/main/pkg/firewall/firewall.go

PR #3062, opened 2022-07-06
Refactoring of the chain packages

lukasz-zimnoch:
Depends on: #3034 

This pull request refactors the chain package structure to match the expectations of v2. I recommend going commit by commit to grasp the changes in an easy way.

### Planned packages structure for v2

We aim for the following structure of packages:
```
pkg
    beacon
        beacon.go <-- main package file exposing `beacon.Initialize` function
        chain.go  <-- declares `beacon.Chain` interface that sets expectations for the required subset of on-chain functions
    tbtc
        tbtc.go   <-- main package file exposing `tbtc.Initialize` function
        chain.go  <-- declares `tbtc.Chain` interface that sets expectations for the required subset of on-chain functions
    chain
        ethereum
            beacon/gen  <-- beacon contract bindings
            tbtc/gen    <-- tbtc contract bindings
            ethereum.go <-- defines `ethereum.Chain` which is a base non-app-specific chain exposing generic features
            beacon.go   <-- defines `ethereum.BeaconChain` which embeds `ethereum.Chain` and implements `beacon.Chain`
            tbtc.go     <-- defines `ethereum.TbtcChain` which embeds `ethereum.Chain` and implements `tbtc.Chain`
            ...         <-- other files like stake_monitor.go, key.go, etc.
        block_counter.go <-- declares chain-agnostic `chain.BlockCounter` interface
        signing.go       <-- declares chain-agnostic `chain.Singing` interface
        stake_monitor.go <-- declares chain-agnostic `chain.StakeMonitor` interface
        ...              <-- other chain-agnostic cross-application interfaces and types
```
This structure allows each application package (`pkg/beacon` and `pkg/tbtc`) to set and control their own expectations regarding the chain. The `pkg/chain` contains chain implementations that must implement application-specific chain interfaces. This way the chain implementation is decoupled from the application domain code and does not affect it. Chain-agnostic interfaces declared directly in the `pkg/chain` package represents cross-application chain concepts that are shared by all applications. 

### Steps towards the desired structure done in this PR

This PR makes the following steps to reach the desired structure:

1. It renames the legacy `pkg/chain/ethereum` package to `pkg/chain/ethereum_v1` and adjusts all the references
2. Adds the stub of the new Ethereum implementation in `pkg/chain/ethereum` that consists of the base `ethereum.Chain` chain handle placed in `ethereum.go` and app-specific chain handles `ethereum.BeaconChain` and `ethereum.TbtcChain` placed in `beacon.go` and `tbtc.go` respectively.
3. Removes the utility chain functionality to reduce the complexity of chain code
4. Removes the `chain.Handle` interface that was living in `pkg/chain` in favor of small specialized interfaces scattered through separate files like `stake_monitor.go`, `block_counter.go` and so on.
5. Removes the `chain.Staker` thus the entire concept of staker as it no longer relevant for v2
6. Moves the Ethereum key code from the legacy to the new package as it remains unchanged
7. Adds the generic features implementations to the new `pkg/chain/ethereum` package. Those are the block counter, the stake monitor and the signer.

### Next steps that need to be done

To completely transition towards the new structure we still need to:

1. Move v2 contract bindings under `pkg/chain/ethereum`
2. Simplify the `pkg/beacon` structure to have `chain.go` at the root
3. Extend `pkg/chain/ethereum` to contain references to v2 contract bindings and all the required features 
4. Gradually replace usages of the legacy `pkg/chain/ethereum_v1` package in favor of `pkg/chain/ethereum`.


cmd/relay.go
```diff
@@ -1,104 +0,0 @@
-package cmd
```
pdyraga, line 1: Let's add an issue to the last milestone to implement the relay command with a reference to this PR where we temporarily remove it.

pkg/chain/ethereum/tbtc.go
```diff
@@ -0,0 +1,24 @@
+package ethereum
+
+import (
+	"context"
+	"github.com/keep-network/keep-common/pkg/chain/ethereum"
+	"github.com/keep-network/keep-common/pkg/chain/ethereum/ethutil"
+)
+
+// TBTCChain represents a TBTC-specific chain handle.
+type TBTCChain struct {
```
pdyraga, line 10: Maybe `TbtcChain`? Full camel case mode looks better.

pdyraga (commented):
This is really good start. I would also consider moving:
- `pkg/chain/random-beacon`
- `pkg/chain/tbtc-v2`
- `pkg/chain/threshold-network`
- `pkg/chain/ecdsa`

under `ethereum`. But maybe in a separate PR to unblock the other work?

pkg/chain/signing.go
```diff
@@ -0,0 +1,39 @@
...
+	// reason, an error is returned.
+	VerifyWithPublicKey(
+		message []byte,
+		signature []byte,
+		publicKey []byte,
+	) (bool, error)
+
+	// PublicKeyToAddress converts operator's public key to an address
+	// associated with the chain.
+	PublicKeyToAddress(publicKey *operator.PublicKey) ([]byte, error)
```
pdyraga, line 34: In #3056 we added a TODO to `PublicKeyToAddress` and to `PublicKeyBytesToAddress` to return `Address` type. Let's make sure we do not lose this TODO when merging changes from `main`.

pkg/chain/ethereum/stake_monitor.go
```diff
@@ -1,67 +1,21 @@
...
 
-type ethereumStakeMonitor struct {
-	ethereum *ethereumChain
+type stakeMonitor struct {
+	chain *Chain
 }
 
-func (esm *ethereumStakeMonitor) HasMinimumStake(
+// TODO: Real implementation with v2 contracts.
+func (sm *stakeMonitor) HasMinimumStake(
```
pdyraga, line 17: @tomaszslabon Your #3060 will need this function. The proposed approach is: the `firewall` package should define an interface it is expecting, or import `chain.StakeMonitor`. I think the latter is fine for the time being. Per #3010, we no longer want to check `HasMinimumStake` but `HasStakeDelegation` (we should rename this function). And the logic checking staking provider <> operator mapping and the TokenStaking roles should be checked inside `HasStakeDelegated`. 

I think we should check both `RandomBeacon` and `WalletRegistry` for staking provider <> operator mapping. We want to have one network of nodes for both applications.

pkg/chain/ethereum/block_counter.go
```diff
@@ -0,0 +1,8 @@
+package ethereum
+
+import "github.com/keep-network/keep-core/pkg/chain"
+
+// BlockCounter creates a BlockCounter that uses the block number in Ethereum.
+func (c *Chain) BlockCounter() (chain.BlockCounter, error) {
+	return c.blockCounter, nil
```
pdyraga, line 7: What is the idea behind this function? THe block counter comes from `keep-common`, right? Maybe this function should be a part of `ethereum.go`? 🤔

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: See c29d0e53

pkg/chain/ethereum/block_counter.go
lukasz-zimnoch: We could put in `ethereum.go` but wanted to keep it in the same way as others generic features and align with how interfaces are declared in `pkg/chain`. Also, if we want to merge `keep-common` one day, the block counter will live here.

pkg/chain/ethereum/ethereum.go
```diff
@@ -2,458 +2,93 @@ package ethereum
...
+	key          *keystore.Key
+	client       ethutil.EthereumClient
+	blockCounter *ethlike.BlockCounter
+}
+
+// NewChain construct a new instance of the Ethereum chain handle.
+func NewChain(
+	ctx context.Context,
+	config *ethereum.Config,
+	client ethutil.EthereumClient,
```
pdyraga, line 30: This should probably be `*ethclient.Client,` from `github.com/ethereum/go-ethereum/ethclient`. We need to obtain `ChainID` with a call to `chainID, err := client.ChainID(context.Background())` and this is not possible with `ethutil.EthereumClient`.

pdyraga: I did this change in #3067 at commit https://github.com/keep-network/keep-core/pull/3067/commits/012e36247d967791a7f8bbd73ce85f68ff7b5010. Started building on a top of this PR.

pkg/chain/ethereum/beacon.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// NewBeaconChain construct a new instance of the beacon-specific Ethereum
+// chain handle.
+func NewBeaconChain(
+	ctx context.Context,
+	config *ethereum.Config,
+	client ethutil.EthereumClient,
+) (*BeaconChain, error) {
+	chain, err := NewChain(ctx, config, client)
```
pdyraga, line 22: The problem with creating a new instance of the base chain in `New*` function of the application chain is that we will end up with multiple instances of the base chain: one for Random Beacon and one for TBTC. If the base chain holds the transaction mutex - and this feels like the right place - it means we have two mutexes.

To avoid it, I would consider keeping the base chain as a singleton and creating application chains using public functions of the base chain. 

I did this change in https://github.com/keep-network/keep-core/pull/3067/commits/a3ba22abc7ce4b0c75413b261a1e11d12df8361c in the followup PR of #3067.

pdyraga: Please also see https://github.com/keep-network/keep-core/pull/3067/commits/fa755027dde397dda86b1dfd428c4fa271cbf88a that deals with `Connect` function.

PR #3067, opened 2022-07-07
RandomBeacon contract for the Beacon Ethereum chain - sortition functions

pdyraga:
~~Depends on #3062~~

Refs #3006

This PR implements Ethereum chain functions for Random Beacon Sortition Pool. The following functions are implemented for Ethereum:
```
OperatorToStakingProvider() (chain.Address, bool, error)
EligibleStake(stakingProvider chain.Address) (*big.Int, error)
IsPoolLocked() (bool, error)
IsOperatorInPool() (bool, error)
JoinSortitionPool() error
```

Some additional changes were necessary to instantiate `RandomBeacon` and `SortitionPool` contract references:
- The base Ethereum chain needs to provide the chain ID, transaction mutex, nonce manager, and mining waiter.
- We need to ensure there is just one instance of transaction mutex. This mutex is held by the base Ethereum chain so there has to be only one instance of Ethereum base chain. The idea is to call global `Connect` only one time and create all the chain instances only one time. Creating chain instances outside of `Connect` is not possible from the outside of `ethereum` package.
- We do not need to instantiate `ethclient.Client` outside of `Connect`, this is the action that should be executed upon establishing the connection, so `Connect` function is the right place.


pdyraga:
There are no unit tests for Ethereum implementation, as they would require having an Ethereum client instance running. This will be covered by system tests. To test functions during the review, the following simple implementation can be used along with `scripts/install.sh` and `scripts/initialize.sh`:
```
package ethereum

import (
	"context"
	"fmt"
	"testing"

	"github.com/keep-network/keep-common/pkg/chain/ethereum"
)

func TestBeaconChain_KnownOperator(t *testing.T) {
	ctx := context.Background()

	config := ethereum.Config{}

	config.URL = "http://127.0.0.1:8545"
	config.Account.KeyFile = "/Users/piotr/dev/ethereum/data/keystore/UTC--2018-11-01T06-24-04.418691008Z--65ea55c1f10491038425725dc00dffeab2a1e28a"
	config.Account.KeyFilePassword = "password"
	config.ContractAddresses = make(map[string]string)
	config.ContractAddresses["RandomBeacon"] = "0x33a4bB615E56bB7538cFF511C4dc0A66fb1B27F8"

	beacon, _, err := Connect(ctx, config)
	if err != nil {
		t.Fatal(err)
	}

	stakingProvider, ok, err := beacon.OperatorToStakingProvider()
	if err != nil {
		t.Fatal(err)
	}
	fmt.Printf("stakingProvider = %v, ok = %v\n", stakingProvider, ok)

	eligibleStake, err := beacon.EligibleStake("0x65ea55c1f10491038425725dc00dffeab2a1e28a")
	if err != nil {
		t.Fatal(err)
	}
	fmt.Printf("eligibleStake = %v\n", eligibleStake)

	eligibleStake, err = beacon.EligibleStake("0x33a4bB615E56bB7538cFF511C4dc0A66fb1B27F8")
	if err != nil {
		t.Fatal(err)
	}
	fmt.Printf("eligibleStake = %v\n", eligibleStake)

	isPoolLocked, err := beacon.IsPoolLocked()
	if err != nil {
		t.Fatal(err)
	}
	fmt.Printf("isPoolLocked = %v\n", isPoolLocked)

	isOperatorInPool, err := beacon.IsOperatorInPool()
	if err != nil {
		t.Fatal(err)
	}
	fmt.Printf("isOperatorInPool = %v\n", isOperatorInPool)

	err = beacon.JoinSortitionPool()
	if err != nil {
		t.Fatal(err)
	}

	isOperatorInPool, err = beacon.IsOperatorInPool()
	if err != nil {
		t.Fatal(err)
	}
	fmt.Printf("isOperatorInPool = %v\n", isOperatorInPool)
}
```

pkg/chain/ethereum/ethereum.go
```diff
@@ -18,17 +22,75 @@ var logger = log.Logger("keep-chain-ethereum")
...
 }
 
-// NewChain construct a new instance of the Ethereum chain handle.
-func NewChain(
+// Connect creates Random Beacon and TBTC Ethereum chain handles.
+func Connect(
 	ctx context.Context,
-	config *ethereum.Config,
-	client ethutil.EthereumClient,
+	config ethereum.Config,
```
lukasz-zimnoch, line 51: Any particular reason to pass the config as a value and not a pointer?

pdyraga: My reasoning was that we do not want to modify anything in the original config and that the value passed should be read-only. This works better if we pass by value (a copy) instead of a pointer.

PR #3070, opened 2022-07-07
IsOperatorUpToDate and UpdateOperatorStatus functions added to sortition Chain

pdyraga:
Refs #3006 

`IsOperatorUpToDate` and `UpdateOperatorStatus` functions added to sortition `Chain` and implemented for the local and Ethereum chain.

`IsOperatorUpToDate` and `UpdateOperatorStatus` are needed to monitor the status of the operator in the sortition pool. The local Go implementation mirrors the behavior of `BeaconAuthorization` Solidity library.

### DIFF
--- a/pkg/sortition/internal/local.go
+++ b/pkg/sortition/internal/local.go
@@ -18,11 +18,11 @@ type localChain struct {
 	operatorToStakingProvider      map[chain.Address]chain.Address
 	operatorToStakingProviderMutex sync.RWMutex
 
-	// Stake for an operator, as set in the Sortition Pool contract.
+	// Weight of an operator, as set in the Sortition Pool contract.
 	sortitionPool      map[chain.Address]*big.Int
 	sortitionPoolMutex sync.RWMutex
 
-	// Stake for an operator, as set in the Token Staking contract,
+	// Stake for a staking provider, as set in the Token Staking contract,
 	// minus the pending authorization decrease.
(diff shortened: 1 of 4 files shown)


pkg/sortition/internal/local.go
```diff
@@ -88,6 +88,28 @@ func (lc *localChain) IsOperatorInPool() (bool, error) {
...
+
+	stakingProvider, isRegistered := lc.operatorToStakingProvider[lc.operatorAddress]
+	if !isRegistered {
+		return false, errOperatorUnknown
+	}
+
+	eligibleStake, hasStake := lc.eligibleStake[stakingProvider]
+	weight, isInPool := lc.sortitionPool[lc.operatorAddress]
+
+	if !isInPool {
```
dimpar, line 106: Might be slightly easier to read, up to you..
```
        if isInPool {
		return weight.Cmp(eligibleStake) == 0, nil
	}

	return !hasStake || eligibleStake.Cmp(big.NewInt(0)) == 0, nil
```

pdyraga: That's a good idea. I changed it in #3077 to avoid conflicts. See de8cd7075be3a3014829fcd744fbcad426ee0357 please.

lukasz-zimnoch (approved):
LGTM! Leaving it in @dimpar hands as he left some comments

dimpar (approved):

PR #3074, opened 2022-07-08
Cleanup of the `pkg/beacon` package

lukasz-zimnoch:
This pull request is the next step of the package refactoring started in #3062. It simplifies the structure of the `pkg/beacon` and makes some general cleanup. Specific changes are described below. Please review commit by commit to easily understand all changes.

### Remove the intermediary `relay` package

So far, all the beacon code have lived under `pkg/beacon/relay/*`. The intermediary `relay` package turned out to be redundant and just made the package structure more complicated. This PR removes the `relay` package and places everything directly in the root `pkg/beacon` package. This makes sense because the core logic now lives there.

### Embed the `pkg/beacon/relay.go` file into the `pkg/beacon/node.go` file.

There is a `Node` structure that encapsulates the behavior of the beacon node. So far, their code was scattered through two files `relay.go` and `node.go`. The rules behind that split were unclear so we decided to merge them together into a single `node.go` file that defines the `Node` type along with all their methods.

### Make the `Node` type package-private

The aforementioned `Node` type is used internally by the `pkg/beacon` and should never be instantiated manually outside of the `pkg/beacon` package. External clients should use `beacon.Initialize` function. That said, we decided to make the `Node` type package-private. 

### Adjustments around beacon chain interface

So far, the beacon chain interface was living in the `pkg/beacon/relay/chain` package and was divided into two files `chain.go` and `result.go`. Since the `relay` package was removed, the new path is `pkg/beacon/chain` which is more straightforward. To adhere to this change,  all code references that used `relaychain.Interface` are now using `beaconchain.Interface`. Also, the `result.go` file content was merged into the `chain.go` file, and all the nomenclature was adjusted to say about the "beacon chain interface" instead of the "relay chain". The interface was also aligned to be compatible with the new `pkg/chain/ethereum` package that we plan to use for v2. 

Ideally, the beacon chain interface should live directly in the `pkg/beacon` and should be named `beacon.Chain`. Unfortunately, that causes import cycles with the `dkg` and `entry` packages since `pkg/beacon` triggers the logic from there. Making that work would require a bigger refactoring that is described in the last section of this PR (see **Ideas for further refactoring**)

### Remove deprecated `StakerAddress` and replace it with `chain.Address`

Since v2 doesn't use the concept of `staker`, the `StakerAddress` type was deprecated recently. Here we remove it and replace all occurrences with the new `chain.Address` type introduced recently. The `chain.Address` type represents a chain-agnostic address and is a good fit to replace the obsolete `StakerAddress`.

### Ideas for further refactoring

As mentioned above, the best place for the beacon chain interface is the root `pkg/beacon`. However, that approach brings some import cycles. If we transform the `pkg/beacon/chain.Interface` to be `pkg/beacon.Chain`, the `pkg/beacon/dkg` and `pkg/beacon/entry` packages still need to import it. At the same time, the `node` type living in the `pkg/beacon` triggers the `pkg/beacon/dkg` and `pkg/beacon/entry` logic directly. To break the cycle, we need to make the root `pkg/beacon` independent of their sub-packages that hold the implementation of the DKG and entry signing logic. In order to do so, we would need to:
- Move the `dkg.ThresholdSigner` to the `pkg/beacon` package and rename it to `beacon.Signer`. This type seems to belong to the core beacon logic and should not be tied to the specific DKG implementation provided by the `pkg/beacon/dkg` package
- Decouple `pkg/beacon` from `pkg/beacon/dkg` and `pkg/beacon/entry` using the `beacon.DkgExecutor` and `beacon.RelayEntryExecutor` interfaces:

    ```
    package beacon

    // Former `beaconchain.Interface` moved from `pkg/beacon/chain`
    type Chain interface {
       ...
    }

    // Former `dkg.ThresholdSigner` moved from `pkg/beacon/dkg`
    type Signer struct {
       ...
    }
    
    type DkgExecutor interface {
        Execute(chain Chain, channel net.BroadcastChannel, ...) (*Signer, error)
    }

    type RelayEntryExecutor interface {
        Execute(chain Chain, channel net.BroadcastChannel, signer *Signer, ...) error
    }
    ```
- Make `beacon.Intialize` accepting `beacon.DkgExecutor` and `beacon.RelayEntryExecutor` interfaces and use them to trigger DKG and relay entry from within `beacon.node`. Those interfaces would be implemented by `pkg/beacon/dkg` and `pkg/beacon/entry` respectively. Those implementation would be injected in the `cmd/start.go` file.

Those changes will introduce a proper direction of dependencies in the entire `pkg/beacon` package and will allow to easily test the `pkg/beacon` code by providing DKG and entry stubs.


pkg/beacon/beacon.go
```diff
@@ -25,32 +23,36 @@ var logger = log.Logger("keep-beacon")
...
+		return fmt.Errorf("failed to get block counter instance: [%v]", err)
+	}
+
+	signing := beaconChain.Signing()
+
+	groupRegistry := registry.NewGroupRegistry(beaconChain, persistence)
 	groupRegistry.LoadExistingGroups()
 
-	node := relay.NewNode(
+	node := newNode(
```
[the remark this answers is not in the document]
lukasz-zimnoch: Honestly, I think that still makes sense because we handle instance creation in one place. This will be useful once we add unit tests into the mix.

cmd/start.go
```diff
@@ -59,7 +59,7 @@ func init() {
 		}
 }
 
-// Start starts a node; if it's not a bootstrap node it will get the Node.URLs
+// Start starts a node; if it's not a bootstrap node it will get the node.URLs
```
pdyraga, line 62: Should still be `Node.URLs` because it's about the location in config file.

pkg/beacon/beacon.go
```diff
@@ -25,32 +23,36 @@ var logger = log.Logger("keep-beacon")
...
 		blockCounter,
 		chainConfig,
 		groupRegistry,
 	)
 
-	eventDeduplicator := event.NewDeduplicator(relayChain)
+	eventDeduplicator := event.NewDeduplicator(beaconChain)
 
-	node.ResumeSigningIfEligible(relayChain, signing)
+	node.ResumeSigningIfEligible(beaconChain, signing)
```
pdyraga, line 53: `beaconChain` has `signing` inside, so we could pass just one parameter to this function.

pkg/beacon/chain/chain.go
```diff
@@ -52,17 +48,17 @@ type RelayEntryInterface interface {
...
-// pertains to relay group selection activities.
+// GroupSelectionInterface defines the subset of the beacon chain interface that
+// pertains to the group selection activities.
 type GroupSelectionInterface interface {
 	// SelectGroup returns the group members for the group generated by
-	// the given seed. This function can return an error if the relay chain's
+	// the given seed. This function can return an error if the beacon chain's
 	// state does not allow for group selection at the moment.
-	SelectGroup(seed *big.Int) ([]StakerAddress, error)
+	SelectGroup(seed *big.Int) ([]chain.Address, error)
```
pdyraga, line 57: ❤️

pdyraga (approved):
Tested locally with DKG test loop, no problems spotted.

pkg/beacon/beacon.go
lukasz-zimnoch: Yep, we will address that once we work on the relay entry for v2.

cmd/start.go
lukasz-zimnoch: Ah, my IDE's rename feature went too far.

Issue #3008, opened 2022-06-02
Restore reward eligibility in the Random Beacon sortition pool


pdyraga:
In case the operator was marked as ineligible for rewards, the client needs to monitor the ineligibility status and restore the eligibility as soon as possible.

References:

https://github.com/keep-network/sortition-pools/blob/b3fcb43e09d9fc2304fa2f65635aec406a6cfe79/contracts/SortitionPool.sol#L140-L152

https://github.com/keep-network/sortition-pools/blob/b3fcb43e09d9fc2304fa2f65635aec406a6cfe79/contracts/Rewards.sol#L74-L99

PR #3077, opened 2022-07-08
Monitor Random Beacon sortition pool status: join pool and update pool

### LINKED ISSUE #3006: Register in Random Beacon sortition pool

### LINKED ISSUE #3007: Update Random Beacon state
Depends on #3006 

The client needs to monitor the sortition pool state and update it when it gets outdated. A simple loop similar to the one we have in `keep-ecdsa` should do the work.

Reference: https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/contracts/libraries/BeaconAuthorization.sol

pdyraga:
~~Depends on #3070~~

Closes #3006 
Closes #3007 

`MonitorPool` periodically checks the status of the operator in the sortition pool. If the operator is supposed to be in the sortition pool but is not there yet, the function attempts to add the operator to the pool. If the operator is already in the pool and its status is no longer up to date, the function attempts to update the operator's status in the pool.

There was a complication with `ethereum` chain - the v2 implementation does not support v1 `ethereum` chain functions and the v1 implementation does not support v2 functions. For now, both instances are passed when creating a `Node`. In the future, we will decide if we want to port features to v2 gradually or cut the entire v1 chain in one move altogether.


pkg/sortition/sortition_test.go
```diff
@@ -0,0 +1,182 @@
...
+	if os.Getenv("PRINT_LOGS_IN_TEST") == "true" {
+		err := log.SetLogLevel("*", "DEBUG")
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "logger initialization failed: [%v]\n", err)
+			os.Exit(-1)
+		}
+	}
+
+	os.Exit(m.Run())
+}
```
lukasz-zimnoch, line 33: Can we drop a short comment explaining what happens here? Might be hard to grasp at the first sight.

pkg/sortition/internal/local/local.go
```diff
@@ -1,4 +1,4 @@
-package sortition
+package local
```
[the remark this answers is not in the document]
pdyraga: Good point. I did the file rename in https://github.com/keep-network/keep-core/pull/3077/commits/3c68e2cdd87eb4122a7ae070323650e9d848f2a6. The struct rename is a bit more complicated because we import `"github.com/keep-network/keep-core/pkg/chain"` in that file and can't have a structure named `chain`. I experimented with aliasing the package but I think it brought more obfuscation than clarity.

pkg/sortition/sortition_test.go
pdyraga: Good point, done in https://github.com/keep-network/keep-core/pull/3077/commits/05f10d92acbc9560cd398a08b1000b2fa2c6679e.

pdyraga:
All changes pushed, I am undrafting this PR @lukasz-zimnoch

pkg/sortition/internal/local/local.go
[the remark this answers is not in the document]
lukasz-zimnoch: Maybe we should consider exporting the channel struct and have `local.Chain`? That seems to make sense as we have `local.Connect` function that returns that struct outside.

cmd/start.go
```diff
@@ -92,17 +93,22 @@ func Start(c *cli.Context) error {
...
 	}
 
-	chainProvider, err := ethereum_v1.Connect(ctx, config.Ethereum)
+	beaconChain, _, err := ethereum.Connect(ctx, config.Ethereum)
 	if err != nil {
 		return fmt.Errorf("error connecting to Ethereum node: [%v]", err)
 	}
 
-	blockCounter, err := chainProvider.BlockCounter()
+	legacyChainProvider, err := ethereum_v1.Connect(ctx, config.Ethereum)
```
lukasz-zimnoch, line 101: s/`legacyChainProvider`/`chainProviderV1`? Should make more sense as we have `ethereum_v1` package. This will also allow to avoid conflicts with https://github.com/keep-network/keep-core/pull/3076.

pkg/beacon/beacon.go
```diff
@@ -23,21 +25,25 @@ var logger = log.Logger("keep-beacon")
 // internal random beacon implementation. Returns an error if this failed,
 // otherwise enters a blocked loop.
 func Initialize(
+	ctx context.Context,
 	operatorPublicKey *operator.PublicKey,
-	beaconChain beaconchain.Interface,
+	// TODO: Get rid of legacy Ethereum chain
+	legacyBeaconChain beaconchain.Interface,
```
lukasz-zimnoch, line 31: s/`legacyBeaconChain`/`beaconChainV1` ? Exactly same reasons as in `legacyChainProvider`-related comment.

pkg/beacon/beacon.go
```diff
@@ -23,21 +25,25 @@ var logger = log.Logger("keep-beacon")
...
+	// TODO: Accept a generic interface, not just sortition.Chain
+	beaconChain sortition.Chain,
```
lukasz-zimnoch, line 33: PR #3076 uses `beaconchain.Interface` as the type of the new `beaconChain`. Do we agree that `sortition.Chain` should be embedded into `beaconchain.Interface`?

pkg/sortition/sortition.go
```diff
@@ -0,0 +1,118 @@
...
+
+	isOperatorInPool, err := chain.IsOperatorInPool()
+	if err != nil {
+		return err
+	}
+
+	if isOperatorInPool {
+		logger.Info("operator is in the sortition pool")
+	} else {
+		logger.Info("operator is not in the sortition pool")
```
lukasz-zimnoch, line 75: Just to make sure we don't `return` here deliberately. We still want to check the operator status even though the operator is not in the pool right?

pdyraga: Yes. `IsOperatorUpToDate` not only checks whether the operator that is already in pool has up-to-date weight but also if operator is supposed to be in the sortition pool. That is, if someone is *not* in the sortition pool but their eligible stake is non-zero, `IsOperatorUpToDate` returns false.

pdyraga: This flow can be translated to English as:
- Are you in the pool? Yes. Are you up to date?
- Are you in the pool? No. Are you supposed to be in the pool?

pkg/beacon/beacon.go
pdyraga: https://github.com/keep-network/keep-core/pull/3077/commits/dc2cad879bebd36c3c4983ae2493b5e02816ccf8

pkg/sortition/sortition.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	isOperatorUpToDate, err := chain.IsOperatorUpToDate()
+	if err != nil {
+		return err
+	}
+
+	if isOperatorUpToDate {
+		if isOperatorInPool {
+			logger.Infof("sortition pool operator status is up to date")
```
dimpar, line 85: I think we can use just `logger.Info`, not `logger.Infof` cause there's nothing to format here. Same comment for other places like this in this file.

pdyraga: Yeah, good point. Let's fix it when working on #3008.

PR #3092, opened 2022-07-14
Bumping up keep-network/sortition-pools to 2.0.0-pre.13

dimpar:
Bumping to the latest version of the sortition pool


pdyraga:
To add some more context, we need to pull changes from https://github.com/keep-network/sortition-pools/pull/182 so that we can implement rewards eligibility monitoring in #3085. This refs #3008.

pdyraga (approved):

PR #3099, opened 2022-07-20
Added check for a timeout in the DKG loop

tomaszslabon:


pkg/chain/ethereum/beacon.go
```diff
@@ -648,6 +644,17 @@ func (mrb *mockRandomBeacon) SubmitRelayEntry(
...
+func (mrb *mockRandomBeacon) ReportRelayEntryTimeout() error {
+	mrb.currentRequestMutex.Lock()
+	defer mrb.currentRequestMutex.Unlock()
+
+	// Set the current request start block to nil, so that a new relay entry
+	// request can be begin.
+	mrb.currentRequestStartBlock = nil
+
+	return nil
+}
```
pdyraga, line 656: Is the client in the current version calling this function?

tomaszslabon: The client in the current version calls this function.
There is a log that appears every time the relay entry timeout happens:
```
2022-07-26T16:28:06.107+0200    WARN    keep-beacon      go-log@v1.0.5/log.go:180relay entry was not submitted on time, reporting timeout at block [2214]
ReportRelayEntryTimeout called!!!!!!!!!!!!!!!!!!!!!!!!
```

(I temporarily added my own log, just to be sure: ```ReportRelayEntryTimeout called!!!!!!!!!!!!!!!!!!!!!!!!```)

pdyraga:
👀

pkg/chain/ethereum/beacon.go
```diff
@@ -451,22 +451,18 @@ func (mrb *mockRandomBeacon) OnDKGStarted(
 				// Generate an event every 500th block.
 				if block%500 == 0 {
 					mrb.currentDkgMutex.Lock()
-
-					if mrb.currentDkgStartBlock == nil {
```
pdyraga, line 455: We have this check when submitting DKG result:
https://github.com/keep-network/keep-core/blob/a04728e4591d0eede8ee8b712d83489e37470463/pkg/chain/ethereum/beacon.go#L517-L520

Should we remove it as well @lukasz-zimnoch? I do not understand why we need it there. We count on the fact 500 blocks is enough for a DKG to complete now.

lukasz-zimnoch: Yep, we should remove it in that case.

pdyraga (approved):

pkg/chain/ethereum/beacon.go
tomaszslabon: Done in 7d3722f79dd2ae45df78a91a92814c7134eb52b0.

PR #3100, opened 2022-07-21
Restore old keep-core v1 deploy script

michalsmiarowski:
There was a change recently that refactored the `install.sh` script in
such a way that it's deploying the contracts on local geth instead of
ganache. This caused some problem when deploying the contracts locally
for ganache when developing the threshold dApp.

We are temporarly restoring back the old keep-core v1 deploy script as a
new file to get rid of errors in deployment proccess and pretty soon we
will refactor the whole process and move to goerli testnet (instead of
developing the contracts locally).


scripts/install_keep_core_v1.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
nkuba, line 1: Can we rename the file to `install-v1.sh`?

scripts/install_keep_core_v1.sh
```diff
@@ -0,0 +1,98 @@
...
+
+# Overwrite default properties
+CONFIG_DIR_PATH=${config_dir_path:-$CONFIG_DIR_PATH_DEFAULT}
+NETWORK=${network:-$NETWORK_DEFAULT}
+CONTRACTS_ONLY=${contracts_only:-false}
+
+# Run script
+printf "${LOG_START}Starting installation...${LOG_END}"
+
+printf "Config dir path: $CONFIG_DIR_PATH\n"
```
nkuba, line 71: We're not using this path anywhere, right? Can we remove all traces of config dir path?

scripts/install_keep_core_v1.sh
```diff
@@ -0,0 +1,98 @@
...
+      h ) help ;;
+      ? ) help ;; # Print help in case parameter is non-existent
+   esac
+done
+shift $(expr $OPTIND - 1) # remove options from positional parameters
+
+# Overwrite default properties
+CONFIG_DIR_PATH=${config_dir_path:-$CONFIG_DIR_PATH_DEFAULT}
+NETWORK=${network:-$NETWORK_DEFAULT}
+CONTRACTS_ONLY=${contracts_only:-false}
```
nkuba, line 66: It's not used now, can we remove it? Also command args should be updated.

nkuba (changes requested):

scripts/install_keep_core_v1.sh
michalsmiarowski: 9cd8ce2675c6b0ae21fa1157c0eb01d2a76e6f63

scripts/install_keep_core_v1.sh
michalsmiarowski: 881196cad3d854f4e4be3e4bd6cee7e4966b9be9

scripts/install_keep_core_v1.sh
michalsmiarowski: 881196cad3d854f4e4be3e4bd6cee7e4966b9be9

scripts/start_dashboard.sh
```diff
@@ -33,7 +33,7 @@ yarn link
 
 printf "${LOG_START}Migrating contracts for keep-ecdsa...${LOG_END}"
 cd "$KEEP_ECDSA_PATH"
-./scripts/install.sh --network local --contracts-only
+./scripts/install_keep_core_v1.sh --network local --contracts-only
```
nkuba, line 36:
```suggestion
./scripts/install-v1.sh --network local
```

scripts/install-v1.sh
```diff
@@ -0,0 +1,93 @@
...
+# Defaults, can be overwritten by env variables/input parameters
+CONFIG_DIR_PATH_DEFAULT="$KEEP_CORE_PATH/configs"
+NETWORK_DEFAULT="local"
+KEEP_ETHEREUM_PASSWORD=${KEEP_ETHEREUM_PASSWORD:-"password"}
+CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY=${CONTRACT_OWNER_ETH_ACCOUNT_PRIVATE_KEY:-""}
+
+help()
+{
+   echo -e "\nUsage: ENV_VAR(S) $0"\
+           "--config-dir <path>"\
```
nkuba, line 21: We should remove all traces of `--config-dir` and `--contracts-only`.

nkuba (changes requested):

scripts/install-v1.sh
```diff
@@ -0,0 +1,85 @@
...
+  case "$arg" in
+    "--network")        set -- "$@" "-n" ;;
+    "--help")           set -- "$@" "-h" ;;
+    *)                  set -- "$@" "$arg"
+  esac
+done
+
+# Parse short options
+OPTIND=1
+while getopts "c:n:mh" opt
```
nkuba, line 43: We also need to remove short options

```suggestion
while getopts "n:h" opt
```

nkuba (changes requested):

nkuba (approved):
:fire:

PR #3109, opened 2022-07-25
One More Once: Signing Retry Algorithm

beaushinkle:
This PR adds a `EvaluateRetryParticipantsForSigning` function in the new `pkg/ecdsa` package which takes in a list of chain addresses, and returns a subset of those chain addresses subject to a random seed and retry count number. 

We plan on using such a function each time a signing event fails for any reason (time out, malformed signature, etc), in order to randomly select a minimum subset of operators capable of producing a valid signature, thereby avoiding as many malicious operators as possible.

In a near future PR, we'll get a sister function, `EvaluateRetryParticipantsForKeyGeneration` that has a similar API, but is geared for... key generation :)

tagging @lukasz-zimnoch for review


pkg/ecdsa/retry_test.go
```diff
@@ -0,0 +1,106 @@
...
+		t.Error("The seed did not affect the subset generation. All subsets were the same.")
+	}
+}
+
+func invariants(groupMembers []chain.Address, subset []chain.Address, quantity uint, t *testing.T) {
+	isSubset(groupMembers, subset, t)
+	isStable(groupMembers, subset, quantity, t)
+	isLargeEnough(subset, quantity, t)
+	affectedBySeed(groupMembers, subset, quantity, t)
+	affectedByRetryCount(groupMembers, quantity, t)
```
beaushinkle, line 87: What do you think about this testing structure? If you like it, I can add more test cases!

lukasz-zimnoch: Very explanatory, I like it! I will have some minor comments regarding naming but will post them separately.

pkg/ecdsa/retry.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (ba byAddress) Len() int           { return len(ba) }
+func (ba byAddress) Swap(i, j int)      { ba[i], ba[j] = ba[j], ba[i] }
+func (ba byAddress) Less(i, j int) bool { return ba[i] < ba[j] }
+
+func EvaluateRetryParticipantsForSigning(
+	groupMembers []chain.Address,
+	seed int64,
+	retryCount uint,
+	quantity uint,
```
lukasz-zimnoch, line 21: Looks vague at the first sight, maybe we can rename it to, for example, `retryParticipantsCount`?

pkg/ecdsa/retry.go
```diff
@@ -0,0 +1,62 @@
...
+) ([]chain.Address, error) {
+	if int(quantity) > len(groupMembers) {
+		return nil, fmt.Errorf(
+			"Asked for too many seats. %d seats were requested, but there are only %d available.",
```
lukasz-zimnoch, line 25: We typically start the error message with a lowercase due to the error wrapping. Also, we put format parameters into square braces:
```suggestion
			"asked for too many seats. [%d] seats were requested, but there are only [%d] available.",
```

pkg/ecdsa/retry.go
```diff
@@ -0,0 +1,62 @@
...
+			len(groupMembers),
+		)
+	}
+	operatorToSeatCount := make(map[chain.Address]uint)
+	for _, operator := range groupMembers {
+		operatorToSeatCount[operator]++
+	}
+	rand.Seed(seed + int64(retryCount))
+
+	operators := make(byAddress, len(operatorToSeatCount))
```
lukasz-zimnoch, line 36: The `byAddress` type is a comparator and should not be used to denote types:
```suggestion
	operators := make([]chain.Address, len(operatorToSeatCount))
```

pkg/ecdsa/retry.go
```diff
@@ -0,0 +1,62 @@
...
+	i := 0
+	for operator := range operatorToSeatCount {
+		operators[i] = operator
+		i++
+	}
+	sort.Sort(operators)
```
lukasz-zimnoch, line 42: You can apply the comparator this way:
```suggestion
	sort.Sort(byAddress(operators))
```

pkg/ecdsa/retry.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/keep-network/keep-core/pkg/chain"
+)
+
+type byAddress []chain.Address
+
+func (ba byAddress) Len() int           { return len(ba) }
+func (ba byAddress) Swap(i, j int)      { ba[i], ba[j] = ba[j], ba[i] }
+func (ba byAddress) Less(i, j int) bool { return ba[i] < ba[j] }
+
+func EvaluateRetryParticipantsForSigning(
```
lukasz-zimnoch, line 17: Turns out that we are able to determine members who were inactive during the given attempt of DKG. Since inactivity will be much more common than actual malice, I'm wondering whether we can use the inactivity information during the retry algorithm somehow in order to increase the success rate. Any thoughts about that?

pkg/ecdsa/retry_test.go
```diff
@@ -0,0 +1,106 @@
...
+		memberMap[operator] = struct{}{}
+	}
+	for _, operator := range subset {
+		if _, ok := memberMap[operator]; !ok {
+			t.Errorf("Subset member [%s] is not in the operator group.", operator)
+		}
+	}
+}
+
+func testEq(a, b []chain.Address) bool {
```
lukasz-zimnoch, line 22: You can do the same using the built-in `reflect.DeepEqual` function.

pkg/ecdsa/retry_test.go
```diff
@@ -0,0 +1,106 @@
+package ecdsa
```
lukasz-zimnoch, line 1: Can we reorder functions in that file? Typical structure is:
```
//Actual test scenarios
TestABC() { ... }
TestXYZ() { ... }

// Helpers 
func initializeABC() { ... }
```

pkg/ecdsa/retry_test.go
```diff
@@ -0,0 +1,106 @@
...
+	for retryCount := uint(1); retryCount < 30 && allTheSame; retryCount++ {
+		newSubset, _ := EvaluateRetryParticipantsForSigning(groupMembers, int64(72312), retryCount, quantity)
+		allTheSame = allTheSame && testEq(subset, newSubset)
+	}
+	if allTheSame {
+		t.Error("The seed did not affect the subset generation. All subsets were the same.")
+	}
+}
+
+func invariants(groupMembers []chain.Address, subset []chain.Address, quantity uint, t *testing.T) {
```
lukasz-zimnoch, line 82: s/invariants/assertInvariants?

lukasz-zimnoch, line 82: Can we put the `t *testing.T` parameter as the first one? This way we make it clear that function can steer the test execution.

pkg/ecdsa/retry_test.go
lukasz-zimnoch, line 1: Can we add a scenario covering the unhappy path when `int(quantity) > len(groupMembers)`?

pkg/ecdsa/retry.go
```diff
@@ -0,0 +1,62 @@
...
+	groupMembers []chain.Address,
+	seed int64,
```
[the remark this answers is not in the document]
beaushinkle: > This is the same seed as the one used for DKG?

`EvaluateRetryParticipantsForSigning` will need a different seed for each message, and I envision the seed as being the `keccak` of the message. I'll add that to the docstring!

pkg/ecdsa/retry_test.go
beaushinkle: I was reading on https://stackoverflow.com/questions/15311969/checking-the-equality-of-two-slices that a `testEq` method like what I wrote is *wildly* faster than `reflect.DeepEqual`, though if we don't care, I don't care! It's test code, so it's not like it has to be performant

pkg/ecdsa/retry.go
beaushinkle: > due to the error wrapping

ahhh makes a bunch of sense

pkg/ecdsa/retry.go
beaushinkle: just to be clear - this PR is about signing; the follow up will be about keygen!

If you have explicit inactivity information, here's what I would suggest (for both dkg and signing).

Step 1) exclude inactive participants and try to make it work. loop until the previous run failed but had full participation
Step 2) start using the retry algorithm from this PR on the participating operators. if operators start becoming inactive, go back to Step 1.

pkg/ecdsa/retry.go
```diff
@@ -0,0 +1,62 @@
+package ecdsa
```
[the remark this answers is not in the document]
beaushinkle: 41decc31ea6714ff74acf342821d2d29c2a27187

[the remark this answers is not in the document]
beaushinkle: [41decc3](https://github.com/keep-network/keep-core/pull/3109/commits/41decc31ea6714ff74acf342821d2d29c2a27187)

pkg/ecdsa/retry.go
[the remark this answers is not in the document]
beaushinkle: [5a186f2](https://github.com/keep-network/keep-core/pull/3109/commits/5a186f2e1f50326d85b19848616588eb4c4d8282)

pkg/ecdsa/retry.go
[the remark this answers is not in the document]
beaushinkle: [5a186f2](https://github.com/keep-network/keep-core/pull/3109/commits/5a186f2e1f50326d85b19848616588eb4c4d8282)

pkg/ecdsa/retry.go
beaushinkle: [f5040e1](https://github.com/keep-network/keep-core/pull/3109/commits/f5040e1b63d7f44cf9ed88aeb7f08813c1a2ea20)

pkg/ecdsa/retry.go
```diff
@@ -0,0 +1,62 @@
...
+type byAddress []chain.Address
+
+func (ba byAddress) Len() int           { return len(ba) }
+func (ba byAddress) Swap(i, j int)      { ba[i], ba[j] = ba[j], ba[i] }
+func (ba byAddress) Less(i, j int) bool { return ba[i] < ba[j] }
+
+func EvaluateRetryParticipantsForSigning(
+	groupMembers []chain.Address,
+	seed int64,
+	retryCount uint,
```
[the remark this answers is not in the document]
beaushinkle: [9fd11dd6](https://github.com/keep-network/keep-core/pull/3109/commits/9fd11dd6768b4a3605fda124004542dd95ff83a0)

pkg/ecdsa/retry.go
beaushinkle: [efc919723efa5b3cf9081bf691c916d102590e99](https://github.com/keep-network/keep-core/pull/3109/commits/efc919723efa5b3cf9081bf691c916d102590e99)

pkg/ecdsa/retry.go
beaushinkle: [c68a994](https://github.com/keep-network/keep-core/pull/3109/commits/c68a994e6559aea4344ca18b1655f86c619dfed3
)

pkg/ecdsa/retry.go
beaushinkle: [c68a994](https://github.com/keep-network/keep-core/pull/3109/commits/c68a994e6559aea4344ca18b1655f86c619dfed3
)

pkg/ecdsa/retry_test.go
beaushinkle: [6309ed9](https://github.com/keep-network/keep-core/pull/3109/commits/6309ed935475873a8886891e77b35327e040ba93
)

pkg/ecdsa/retry_test.go
beaushinkle: [460c495](https://github.com/keep-network/keep-core/pull/3109/commits/460c495327b77664374955ad6acfc26dff0ca2c6
)

beaushinkle: [e8df085](https://github.com/keep-network/keep-core/pull/3109/commits/e8df085d39bfcca930a4f9ae2a3863237f8ac5e6
)

pkg/ecdsa/retry_test.go
beaushinkle: [0b2fb77](https://github.com/keep-network/keep-core/pull/3109/commits/0b2fb77e7f41b5ecfbb1a957cb94b3955f884412
)

pkg/ecdsa/retry.go
lukasz-zimnoch: Makes sense!

pkg/ecdsa/retry_test.go
lukasz-zimnoch: Up to you! I'm good with both.

pkg/ecdsa/retry/retry_test.go
```diff
@@ -0,0 +1,136 @@
+package retry
```
lukasz-zimnoch, line 1: Just a last comment and I'm ready to merge: we did `s/quantity/retryParticipantsCount` though we still use `quantity` here. Can we align?

pkg/ecdsa/retry_test.go
beaushinkle:
```
== testEq ==
❯ go test
PASS
ok  	github.com/keep-network/keep-core/pkg/ecdsa/retry	0.097s

== reflect.DeepEqual ==
~/keep-core/pkg/ecdsa/retry retry-signing
❯ go test
PASS
ok  	github.com/keep-network/keep-core/pkg/ecdsa/retry	0.113s
```

`0.016s` is not worth the extra lines of code. Changing!

beaushinkle: [7feb73e](https://github.com/keep-network/keep-core/pull/3109/commits/7feb73ef2c3edbf07cfb313a8edc362fc41131b9
)

pkg/ecdsa/retry/retry_test.go
beaushinkle: [d7726da](https://github.com/keep-network/keep-core/pull/3109/commits/d7726da2eb70a3058c269a48fa0d8bbf10ac1187
)

beaushinkle: good catch!

lukasz-zimnoch (approved):

PR #3061, opened 2022-07-06
Görli contracts deployment

nkuba:
Görli became a recommended test network after Ropsten's deprecation notice ([link](https://blog.ethereum.org/2022/06/21/testnet-deprecation/)).

In this PR we update deployment scripts and configuration to support Görli.

The contracts were already deployed to Görli:
`@keep-network/random-beacon@2.0.0-goerli.0`
`@keep-network/ecdsa@2.0.0-goerli.0`

PR #3111, opened 2022-07-25
Support multiple accounts for hardhat test network

nkuba:
We want to be able to configure private keys to multiple accounts for
example on goerli network to be able to run initialization scripts.
With this change we expect the environment property to define one or
more private keys separated with a comma (`,`).

The change is made also to the variable name, to better reflect the
meaning.


solidity/random-beacon/hardhat.config.ts
```diff
@@ -94,8 +94,8 @@ const config: HardhatUserConfig = {
     ropsten: {
       url: process.env.CHAIN_API_URL || "",
       chainId: 3,
-      accounts: process.env.CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY
-        ? [process.env.CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY]
+      accounts: process.env.ACCOUNTS_PRIVATE_KEYS
+        ? process.env.ACCOUNTS_PRIVATE_KEYS.split(",")
```
michalinacienciala, line 98: If we change the name of this property, we should also update it here:
https://github.com/keep-network/keep-core/blob/144e7698646e5d26e21efaadf79d5a564635ceb7/.github/workflows/contracts-random-beacon.yml#L187

michalinacienciala: Also, don't we need to make similar change to the `solidity/ecdsa/hardhat.config.ts`? Currently in the CI we pass the value of `CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY` in both projects by filling it with the same GH secret. If we were to provide comma-separated accounts in that secret, with your changes it would work only for the `random-beacon` deploy job.

michalinacienciala (changes requested):
I left two comments.
Other than that - in PR https://github.com/keep-network/keep-core/pull/3061 we replace `ropsten` network config with `goerli` config. I think Git will be able to handle both changes correctly, but it would be good to double check after the merges if everything is at the right place.

solidity/random-beacon/hardhat.config.ts
nkuba: cfefd399429d6d23d9f872d0a0034be295b21fdf

michalinacienciala (approved):

Issue #3041, opened 2022-06-28
ECDSA DKG test loop


pdyraga:
Depends on #3014 

The client should perform threshold ECDSA DKG every 100 Ethereum blocks and store the result on disk. It should not submit the result to the chain. This is a test loop for the early milestone. The logic should include retry algorithm from #3014.

lukasz-zimnoch:
One thing to improve: https://github.com/keep-network/keep-core/pull/3112#discussion_r941209134

lukasz-zimnoch:
Integration between PRs that must be addressed before closing this issue:
- https://github.com/keep-network/keep-core/pull/3120#discussion_r951203237

PR #3112, opened 2022-07-25
ECDSA DKG test loop: Actual protocol logic

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3041
Depends on: #3093

Here we add the core logic of the ECDSA DKG protocol. This involves adding Binance's TSS keygen implementation to work with the state machine. Specific changes are summarized below.

### TSS keygen on the state machine

The ECDSA DKG starts with two phases:
- `ephemeralKeyPairGenerationState` where each member generates an ephemeral key pair and broadcasts a message containing generated public keys for each other member
- `symmetricKeyGenerationState` where each member receives public keys generated by other members and uses them, along with their own private keys, to generate symmetric keys using the ECDH algorithm. These symmetric keys will secure the P2P communication between members in further phases

This pull request uses the `github.com/binance-chain/tss-lib v1.3.3` library to add the TSS keygen to the ECDSA DKG protocol. The TSS keygen is responsible for all the heavy lifting related to key generation and consists of 3 rounds with a finalization round where the result is prepared. That gives us four additional phases in the ECDSA DKG:
- `tssRoundOneState` where each member generates commitments and Paillier public key then broadcasts them using a single broadcast message
- `tssRoundTwoState` where each member generates shares and de-commitments. The shares are meant to be exchanged as secured P2P messages and this is how the TSS protocol generates them internally. However, we are grouping them into a single broadcast message along with the de-commitments. Each share is encrypted with the respective symmetric key generated with the share's receiver.
- `tssRoundThreeState`where each member produces the Paillier proof and then broadcasts it using a single broadcast message
- `finalizationState` where each member prepares its final TSS result

Once the whole DKG execution is done successfully, each group member has the same group public key as a result. The result also contains private key material required for the TSS signing process. If the DKG execution detects missing messages at any stage, it fails fast as there is no point to continue because the TSS protocol requires full participation at all stages. Inactivity handling and the retry strategy allowing to overcome byzantine faults are still under development and will be attached in one of the follow-up PRs.

### TSS pre-parameters pool

Each group member participating in the TSS keygen needs a set of TSS pre-parameters. Generation of a single pre-parameters set can take a while so we cannot generate them on the fly upon protocol start because that can de-synchronize group members and cause inactivites. The better approach is to generate them upfront and hold them in a pool. Here we implement this approach using the new `tssPreParamsPool` type whose implementation is greatly inspired by https://github.com/keep-network/keep-ecdsa/blob/main/pkg/node/params_pool.go. Since we want to reuse the pool between DKG execution, we are also introducing a stateful `dkg.Executor` type that holds the pool reference and exposes a function allowing to trigger DKG.

### Real sortition pool for group selection

So far each node performing the ECDSA DKG test loop was doing the process only with itself. Here we are integrating with the real sortition pool that allows making a proper group selection that spans over registered operators and thus involves multiple nodes that must communicate over the real network.

### Next steps

The next steps will be covered by the follow-up PRs, probably in this order:

- Handling inactive members and integration of the retry algorithm
- Persistence layer for `pkg/tbtc` package that will allow storing the key material on disk
- ECDSA DKG result submission (using a mock contract for now)


pkg/ecdsa/dkg/gen/pb/message.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+option go_package = "pb";
+package dkg;
+
+message EphemeralPublicKeyMessage {
+    uint32 senderID = 1;
+    map<uint32, bytes> ephemeralPublicKeys = 3;
```
beaushinkle, line 8: should the id here be 2? I know it doesn't matter in terms of the specification, but for readability/confusion

pkg/ecdsa/dkg/member.go
```diff
@@ -126,3 +170,27 @@ func (fm *finalizingMember) Result() *Result {
...
+			"",
+			big.NewInt(int64(groupMemberID)),
+		)
+
+		if memberID == groupMemberID {
+			partyID = newPartyID
+		}
+
+		groupPartiesIDs = append(groupPartiesIDs, newPartyID)
+	}
```
beaushinkle, line 193: since we already know the size of size of `groupPartiesIDs` ahead of time; it will be equal to `len(groupMembersIDs)`, we can save ourselves some allocations:

option 1:
```go
groupPartiesIDs := make([]*tss.PartyID, 0, len(groupMembersIDs))
```

or more efficiently but uglier IMO
```go
	groupPartiesIDs := make([]*tss.PartyID, len(groupMembersIDs))

	for i, groupMemberID := range groupMembersIDs {
		newPartyID := tss.NewPartyID(
			strconv.Itoa(int(groupMemberID)),
			"",
			big.NewInt(int64(groupMemberID)),
		)

		if memberID == groupMemberID {
			partyID = newPartyID
		}

		groupPartiesIDs[i] = newPartyID
	}
```

pkg/ecdsa/dkg/states.go
```diff
@@ -106,16 +121,131 @@ func (skgs *symmetricKeyGenerationState) Receive(msg net.Message) error {
...
+			tros.phaseMessages = append(tros.phaseMessages, phaseMessage)
+		}
+	}
+
+	return nil
+}
+
+func (tros *tssRoundOneState) Next() state.State {
+	//TODO implement me
+	panic("implement me")
```
beaushinkle, line 180: image

pkg/ecdsa/dkg/gen/pb/message.proto
lukasz-zimnoch: Definitely! Looks like I made a typo. See 1ca5fe07

pkg/ecdsa/dkg/member.go
lukasz-zimnoch: That's right! See e0d8501e

pkg/ecdsa/dkg/member.go
```diff
@@ -111,24 +126,185 @@ func (skgm *symmetricKeyGeneratingMember) MarkInactiveMembers(
...
+}
+
+// tssRoundTwoMember represents one member in a distributed key generating
+// group performing the second round of the TSS keygen.
+type tssRoundTwoMember struct {
+	*tssRoundOneMember
+}
+
+// MarkInactiveMembers takes all messages from the previous DKG protocol
+// execution phase and marks all member who did not send a message as IA.
```
beaushinkle, line 188: I think it's worth expanding "IA" to "Inactive" in the docstrings

pkg/ecdsa/dkg/params_pool.go
```diff
@@ -0,0 +1,81 @@
...
+				err,
+			)
+			continue
+		}
+
+		t.logger.Infof(
+			"generated new tss pre parameters, took: [%s], "+
+				"current pool size: [%d]",
+			time.Since(start),
+			len(t.pool)+1,
```
beaushinkle, line 70: why is it len+1 ?

pkg/ecdsa/dkg/params_pool.go
```diff
@@ -0,0 +1,81 @@
...
+		)
+
+		t.pool <- params
+	}
+}
+
+// get returns TSS pre parameters from the pool. It pumps the pool after getting
+// and entry. If the pool is empty it will wait for a new entry to be generated.
```
beaushinkle, line 78: s/and entry/an entry

pkg/ecdsa/dkg/protocol.go
```diff
@@ -115,5 +117,318 @@ func (skgm *symmetricKeyGeneratingMember) isValidEphemeralPublicKeyMessage(
...
+) (*tssRoundTwoMessage, error) {
+	// Use messages from round one to update the local party and advance
+	// to round two.
+	for _, tssRoundOneMessage := range tssRoundOneMessages {
+		senderID := tssRoundOneMessage.SenderID()
+		sortedPartiesIDs := trtm.tssParameters.Parties().IDs()
+		senderPartyIDKey := memberIDToTssPartyIDKey(senderID)
+		senderPartyID := sortedPartiesIDs.FindByKey(senderPartyIDKey)
+
+		_, tssErr := trtm.tssParty.UpdateFromBytes(
```
beaushinkle, line 171: When do we use named errors rather than `err` in the code base? There must be times when we need to keep around two errors simultaneously rather than handling one and overriding the variable.

Here though, it looks like we could call this just `err` since there aren't any other errors in scope

pkg/ecdsa/dkg/params_pool.go
pdyraga: This is before the `t.pool <- params` write. If the buffered channel is full, the write operation is blocked. If `t.pool` has 10 elements buffer, and the generated pre-params are 11th, `t.pool <- params` will block but the pool, as a structure, has 11 pre-params generated. Just holding with adding the 11th one to the buffer.

pkg/ecdsa/dkg/dkg.go
```diff
@@ -23,24 +54,26 @@ func Execute(
...
 
 	member := newMember(
-		logger,
+		e.logger,
 		memberIndex,
 		groupSize,
 		dishonestThreshold,
 		membershipValidator,
+		seed.Text(16), // TODO: Should change on retry.,
+		e.tssPreParamsPool.get(),
```
pdyraga, line 68: For an early version when we'll not have that many participants, we need to ensure the pool size is... large.

pkg/ecdsa/dkg/protocol.go
lukasz-zimnoch: This is because `UpdateFromBytes` doesn't return a pure `error` type but `tss.Error` instead. In several places, I needed to differentiate between them. I decided to do the same in all cases for consistency.

pkg/ecdsa/dkg/dkg.go
lukasz-zimnoch: This is true!

pkg/ecdsa/dkg/dkg.go
```diff
@@ -23,24 +54,26 @@ func Execute(
...
 	registerUnmarshallers(channel)
 
 	member := newMember(
-		logger,
+		e.logger,
 		memberIndex,
 		groupSize,
 		dishonestThreshold,
 		membershipValidator,
+		seed.Text(16), // TODO: Should change on retry.,
```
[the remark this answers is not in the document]
lukasz-zimnoch: This is the `sessionID` and I think each retry is a separate session network-wise.

pkg/ecdsa/dkg/member.go
lukasz-zimnoch: See d6bbdac0

pkg/ecdsa/dkg/params_pool.go
lukasz-zimnoch: See d6bbdac0

lukasz-zimnoch:
Code is ready. Will undraft once all unit tests are ready.

pkg/ecdsa/dkg/dkg.go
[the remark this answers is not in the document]
pdyraga: Right. From `tss-lib` documentation:

> Within your transport, each message should be wrapped with a session ID that is unique to a single run of the keygen, signing or re-sharing rounds. This session ID should be agreed upon out-of-band and known only by the participating parties before the rounds begin. Upon receiving any message, your program should make sure that the received session ID matches the one that was agreed upon at the start.

I see session ID is a concept used in our messages but I don't see it being used anywhere in https://github.com/bnb-chain/tss-lib. Did I miss something? Or is it just a requirement that messages should not be mixed between sessions and we are responsible for making sure this is the case?

pkg/ecdsa/dkg/params_pool_test.go
```diff
@@ -0,0 +1,176 @@
...
+	// Create new pool.
+	tssPool := newTestPool(poolSize)
+
+	if len(tssPool.pool) != 0 {
+		t.Errorf(
+			"invalid init length\nexpected: [%d]\nactual:   [%d]",
+			0,
+			len(tssPool.pool),
+		)
+	}
```
pdyraga, line 24: We could reduce it to `testutils.AssertIntsEqual(t, "init length", 0, len(tssPool.pool))`. Same comment to all other integer assertions in this file.

pkg/tbtc/node.go
```diff
@@ -25,9 +28,16 @@ func newNode(
 	netProvider net.Provider,
 	persistence persistence.Handle,
 ) *node {
+	// TODO: Pass TSS pre-parameters pool config from the outside.
```
pdyraga, line 31: Watch out for the changes from #3090. This PR altered how configuration is passed.

pkg/ecdsa/dkg/member.go
```diff
@@ -111,24 +126,199 @@ func (skgm *symmetricKeyGeneratingMember) MarkInactiveMembers(
 	filter.FlushInactiveMembers()
 }
 
+// initializeTssRoundOne returns a member to perform next protocol operations.
+func (skgm *symmetricKeyGeneratingMember) initializeTssRoundOne() *tssRoundOneMember {
+	tssPartyID, groupTssPartiesIDs := generateTssPartiesIDs(
+		skgm.id,
+		skgm.group.MemberIDs(),
```
pdyraga, line 133: Not in this PR but I would consider renaming `MemberIDs()` to `MemberIndexes()`, `InactiveMemberIDs()` to `InactiveMemberIndexes()` and so on.

`skgm.group.MemberIDs()` is just an array of numbers from 1 to 100 and it makes sense to generate TSS party ID based on them.

For a moment I was confused and thought `skgm.group.MemberIDs()` is a list of identifiers, as outputted by the Sortition Pool, especially that the documentation of `MemberIDs()` says:

> MemberIDs returns IDs of all group members, as initially selected to the group

This would not make sense because we would have TSS party IDs duplicated.

pkg/ecdsa/dkg/member.go
```diff
@@ -111,24 +126,199 @@ func (skgm *symmetricKeyGeneratingMember) MarkInactiveMembers(
...
+	)
+
+	tssParameters := tss.NewParameters(
+		tss.EC(),
```
pdyraga, line 137: Maybe we should call `S256()` to ask for secp256k1 explicitly in case `tss-lib` ever changes the default curve?

pkg/ecdsa/dkg/member.go
```diff
@@ -111,24 +126,199 @@ func (skgm *symmetricKeyGeneratingMember) MarkInactiveMembers(
...
+		tss.NewPeerContext(tss.SortPartyIDs(groupTssPartiesIDs)),
+		tssPartyID,
+		len(groupTssPartiesIDs),
+		skgm.group.DishonestThreshold(),
```
pdyraga, line 141: Are we absolutely sure this function accepts the dishonest threshold and not the honest threshold?

pkg/ecdsa/dkg/params_pool.go
beaushinkle: 👍

pkg/ecdsa/dkg/member.go
beaushinkle: Without digging too deeply into how we're defining `DishonestThreshold`, if, say, we want to represent a 51-of-100, the `threshold` param to `tss.NewParameters` needs to be `50`. If we want to represent a 3-of-3, it needs to be `2`.

Same thing here: https://github.com/keep-network/keep-ecdsa/blob/5031a0a8c52c2495644a2d94efe9a3efd38f2c80/pkg/ecdsa/tss/signing.go#L126

lukasz-zimnoch: So this should be `HonestThreshold-1` right?

beaushinkle: Yeah!

I looked up the definition of DishonestThreshold, and something seems up.

```
// according to the protocol needed to generate a new relay entry.
```

So for a 51-of-100, that reads like it should be 51. Then, DishonestThreshold is:

```
// DishonestThreshold is the maximum number of misbehaving participants for
// which it is still possible to generate a signature.
```
In a 51-of-100, you can have 49 participants misbehave and still generate a sig (so long as the other 51 are all properly behaving), which follow's it's code definition: `cc.GroupSize - cc.HonestThreshold`

In the literature, (and as far as I can tell in tss-lib), this is how the threshold param is defined: https://eprint.iacr.org/2019/114.pdf

> A threshold signature scheme enables n parties to share the power to issue digital signatures
under a single public key. A threshold t is specified such that any subset of t + 1 players can
jointly sign, but any smaller subset cannot.

We want to be super sure. I'll do some tss-lib specific testing but it doesn't have to block this PR. Can always follow up with a PR if `honest-1` was wrong

lukasz-zimnoch: See 6e90f79d

pkg/ecdsa/dkg/member.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3112/commits/c7c413d9e048f82d5f379eb0ce6821f60ac9bfa2

pkg/ecdsa/dkg/member.go
lukasz-zimnoch: Left a TODO: https://github.com/keep-network/keep-core/pull/3112/commits/98929caf210936fcdf43b39e12c7158a050f38f4

pkg/ecdsa/dkg/params_pool_test.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3112/commits/542385852b294c7a650b1a1033eb43861b2fc0e4

pkg/ecdsa/dkg/marshaling_test.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	unmarshaled := &ephemeralPublicKeyMessage{}
+
+	err = pbutils.RoundTrip(msg, unmarshaled)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !reflect.DeepEqual(msg, unmarshaled) {
+		t.Fatalf("unexpected content of unmarshaled message")
```
beaushinkle, line 39: non-blocking

Do you think it's worth adding more detail to the failure message? We could print out the sender id and public keys of both messages, for instance

pkg/ecdsa/dkg/marshaling_test.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+}
+
+func TestFuzzEphemeralPublicKeyMessage_MarshalingRoundtrip(t *testing.T) {
+	for i := 0; i < 10; i++ {
+		var (
+			senderID            group.MemberIndex
+			ephemeralPublicKeys map[group.MemberIndex]*ephemeral.PublicKey
+		)
```
beaushinkle, line 48: for my own understanding, when do you prefer:
```go
var (
  variable1 type1
  variable2 type2
)
```
to 
```go
var variable1 type1
var variable2 type2
```
?

pkg/ecdsa/dkg/marshaling_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+		f := fuzz.New().NilChance(0.1).
+			NumElements(0, 512).
+			Funcs(pbutils.FuzzFuncs()...)
```
beaushinkle, line 52: ah, so `gofuzz` is essentially doing generative testing for us. Seems useful for the retry logic testing too!

pkg/ecdsa/dkg/marshaling_test.go
```diff
@@ -0,0 +1,219 @@
...
+		f.Fuzz(&payload)
+		f.Fuzz(&sessionID)
+
+		message := &tssRoundOneMessage{
+			senderID:  senderID,
+			payload:   payload,
+			sessionID: sessionID,
+		}
+
+		_ = pbutils.RoundTrip(message, &tssRoundOneMessage{})
```
beaushinkle, line 110: I see that `pbutils.RoundTrip` returns an error - do we want to make any assertions on that?

pkg/ecdsa/dkg/marshaling_test.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+}
+
+func TestFuzzTssRoundOneMessage_Unmarshaler(t *testing.T) {
+	pbutils.FuzzUnmarshaler(&tssRoundOneMessage{})
```
beaushinkle, line 115: Looks like this is missing an assertion

pkg/ecdsa/dkg/marshaling_test.go
lukasz-zimnoch: Ideally, we can print all fields of both messages. However, in practice, marshaling/unmarshaling errors often require a debugging session anyway. Printing the content may help sometimes but I think this is not worth the additional complexity related to formatting the failure message in a readable way and so on.

pkg/ecdsa/dkg/marshaling_test.go
lukasz-zimnoch: Basically, this is a matter of personal preference. We don't have project-wide conventions for that. Personally, I try to group related variables/constants together or in all cases when this is justified by better readability.

pkg/ecdsa/dkg/marshaling_test.go
lukasz-zimnoch: The entire idea behind fuzzing is to detect unexpected panics, not necessarily recoverable errors. This is why we initially ignored the potential error here. On the other hand, that's a good point because if we ignore a regular error instead of nuking the test we can omit an unexpected flaw. Will add `t.Fatal(err)` in all `pbutils.RoundTrip` calls.

pkg/ecdsa/dkg/marshaling_test.go
lukasz-zimnoch: Same answer as in https://github.com/keep-network/keep-core/pull/3112#discussion_r937530334.

pkg/ecdsa/dkg/states.go
```diff
@@ -107,35 +134,290 @@ func (skgs *symmetricKeyGenerationState) Receive(msg net.Message) error {
...
+		if trts.member.shouldAcceptMessage(
+			phaseMessage.SenderID(),
+			msg.SenderPublicKey(),
+		) && trts.member.sessionID == phaseMessage.sessionID {
+			trts.phaseMessages = append(trts.phaseMessages, phaseMessage)
+		}
+	}
+
+	return nil
+}
```
beaushinkle, line 285: is this function capable of producing an error?

beaushinkle: ah, we probably need it to return an `error` anyway to make it fit in the `Receive` api

lukasz-zimnoch: Yep, we don't produce an error but we must implement the `state.State` interface.

pkg/ecdsa/dkg/params_pool.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+
+		t.logger.Infof(
+			"generated new tss pre parameters, took: [%s], "+
+				"current pool size: [%d]",
+			time.Since(start),
+			len(t.pool)+1,
+		)
+
+		t.pool <- params
```
beaushinkle, line 73: Oh, because this is `t.pool` is a fixed-size channel at `poolSize`, this blocks when `len(t.pool) == poolSize` until `t.get` is called. That consumes a pre-param from the channel, opens up a slot, then we immediately fill the slot, and start calculating the next one until we get blocked again.

The `for {` style loop infinitely repeats, but doesn't continually execute code because the pool runs out of space and it blocks on that goroutine.

beaushinkle:
Okay! Got Lukasz to walk me through the relevant bits of the codebase today which helped a ton, and then gave this one more pass. Let's send it!

beaushinkle (approved):

pkg/ecdsa/dkg/states.go
```diff
@@ -107,35 +134,290 @@ func (skgs *symmetricKeyGenerationState) Receive(msg net.Message) error {
...
+	// hang forever if something goes wrong. To avoid such a resource leak,
+	// we set a round timeout based on state block duration and an average
+	// block time. The exact duration doesn't need to be super-accurate because
+	// if the timeout is hit, the execution will fail anyway. We just want
+	// to give enough time for round computation and make sure the round
+	// terminates regardless of the result.
+	stateBlocks := tros.DelayBlocks() + tros.ActiveBlocks()
+	stateDuration := 15 * time.Duration(stateBlocks)
+	roundCtx, roundCtxCancel := context.WithTimeout(ctx, stateDuration)
+	defer roundCtxCancel()
```
pdyraga, line 179: I am late to the party but there is something about `roundCtx` that makes me really uncomfortable. 

If the actual block time is less than 15 seconds and someone hangs on `member.tssRoundOne` or any other round execution, they will get out of sync with the rest of the group. They will keep executing the protocol with a delay as if nothing wrong happened. Sure, the other group members should reject messages from the operator that is late, but debugging problems might be a nightmare, especially since one operator can have multiple member seats, each one potentially running into their own out-of-sync problems.

The current design of the state machine assumes `Initiate` is a function that can execute the logic and quickly return. Any work that happens in the background - like retransmission - is done in a separate goroutine limited in time by the context and block waiter that cancels it.

There are two options I can think of we can use instead of creating another context and assuming the block has 15 seconds:
- Trigger time-consuming TSS rounds in a separate goroutine with execution time limited by the `ctx` passed to `Initiate`,
- Rework the state machine to not block on `Initiate` and to cancel the context when the block waiter reaches the desired height.

pdyraga: > Actually, roundCtx is only for avoiding hanging forever in case something goes wrong and the round message cannot be produced. If the actual block time is less than 15 seconds, the entire state duration will be shorter than roundCtx and if the member does not produce the round message quickly, it will fail anyway. 

That's something I am not sure about. I don't think it will fail, it will rather keep executing the state machine, no?

That being said, I do not think this change breaks the protocol but I think we need to make it right (option 1 or 2) before we consider #3041 as done.

lukasz-zimnoch: > That's something I am not sure about. I don't think it will fail, it will rather keep executing the state machine, no?

Actually, due to TSS nature, it will go to the next state but fail really soon since it will miss required messages.

> That being said, I do not think this change breaks the protocol but I think we need to make it right (option 1 or 2) before we consider https://github.com/keep-network/keep-core/issues/3041 as done.

Yep, we will address it before closing 3041

pkg/ecdsa/dkg/member.go
pdyraga: Done in https://github.com/keep-network/keep-core/pull/3428

PR #3115, opened 2022-07-27
 #3012 Add LoadBalancer with static IP

electropolis:
- added `type` to `spec` section in service of kubernetes defining kind of service
 - added LoadBalancer IP which was previously reserved using gcloud command as regional IP


electropolis:
README.md will also appear today

nkuba (approved):

PR #3120, opened 2022-08-01
ECDSA DKG result submission

tomaszslabon:
Refs https://github.com/keep-network/keep-core/issues/3041.

This PR adds functionalities for DKG result submission in Tbtc chain in the similar manner as it was done in the Beacon chain.
Note that it uses a mock wallet registry. 

1) Added functions to Tbtc chain that are needed to perform result signing and submission
2) Added additional ECDSA states to perform result signing and submission
3) Implemented functionalities for publishing the result (function Publish which initiates the process)
4) Added handling of DKG result submission failure

In order to deliver this PR and unlock other PRs, it was decided that unit tests for some functions will be added in a separate PR.
The functions that need unit tests are: 
- `dkgResultSigner.SignResult`
- `dkgResultSigner.VerifySignature`
- `dkgResultSubmitter.SubmitResult`
- `signingMember.VerifyDKGResultSignatures`

TODO:
- test manually


pkg/tecdsa/dkg/dkg.go
```diff
@@ -85,6 +86,101 @@ func (e *Executor) Execute(
...
+	logger log.StandardLogger,
+	memberIndex group.MemberIndex,
+	membershipValidator *group.MembershipValidator,
+	submissionConfig *SubmissionConfig,
+	result *Result,
+	channel net.BroadcastChannel,
+	resultHandler ResultHandler,
+	blockCounter chain.BlockCounter,
+	startBlockHeight uint64,
+) error {
```
lukasz-zimnoch, line 152: Let's follow the same order of parameters as in `Execute`.

pkg/tecdsa/dkg/dkg.go
```diff
@@ -85,6 +86,101 @@ func (e *Executor) Execute(
...
+	Signing() chain.Signing
+	// CalculateDKGResultHash calculates 256-bit hash of DKG result in standard
+	// specific for the chain. Operation is performed off-chain.
+	CalculateDKGResultHash(result *Result) (ResultHash, error)
+}
+
+type ResultHandler interface {
+	ResultSigner
+	ResultSubmitter
+}
```
lukasz-zimnoch, line 138: I think that interface is redundant. Signing and submitting are two separate steps so it makes sense to keep those interfaces separately.

pkg/tecdsa/dkg/member.go
```diff
@@ -256,9 +259,382 @@ func (fm *finalizingMember) MarkInactiveMembers(
...
+	config *SubmissionConfig
+	// Hash of DKG result preferred by the current participant.
+	preferredDKGResultHash ResultHash
+	// Signature over preferredDKGResultHash calculated by the member.
+	selfDKGResultSignature []byte
+}
+
+// SignDKGResult calculates hash of DKG result and member's signature over this
+// hash. It packs the hash and signature into a broadcast message.
+func (sm *signingMember) SignDKGResult(
```
lukasz-zimnoch, line 332: Let's move that to `protocol.go` and cover with unit tests.

pkg/tecdsa/dkg/dkg.go
```diff
@@ -85,6 +86,101 @@ func (e *Executor) Execute(
...
+	// on-chain.
+	IsGroupRegistered(groupPublicKey []byte) (bool, error)
+	// OnDKGResultSubmitted registers a callback that is invoked when an on-chain
+	// notification of a new, valid submitted result is seen.
+	OnDKGResultSubmitted(
+		func(event *ResultSubmissionEvent),
+	) subscription.EventSubscription
+}
+
+type ResultSigner interface {
```
lukasz-zimnoch, line 127: Let's document this interface and move it above `ResultSubmitter` to reflect the order of operations.

pkg/tecdsa/dkg/dkg.go
```diff
@@ -85,6 +86,101 @@ func (e *Executor) Execute(
...
+	// Signing returns the chain's signer.
+	Signing() chain.Signing
+	// CalculateDKGResultHash calculates 256-bit hash of DKG result in standard
+	// specific for the chain. Operation is performed off-chain.
+	CalculateDKGResultHash(result *Result) (ResultHash, error)
```
lukasz-zimnoch, line 132: What do you think about exposing methods like:
```
SignResult(result *Result) (ResultSignature, ResultHash, error)
VerifyResult(result *Result, signature ResultSignature, publicKey []byte) (bool, error)
```
where the `ResultSignature` is just a `byte[]`. Then, we can implement that interface in `pkg/tbtc`

pkg/tecdsa/dkg/dkg.go
```diff
@@ -85,6 +86,101 @@ func (e *Executor) Execute(
...
+		signatures map[group.MemberIndex][]byte,
+	) error
+	// IsGroupRegistered checks if group with the given public key is registered
+	// on-chain.
+	IsGroupRegistered(groupPublicKey []byte) (bool, error)
+	// OnDKGResultSubmitted registers a callback that is invoked when an on-chain
+	// notification of a new, valid submitted result is seen.
+	OnDKGResultSubmitted(
+		func(event *ResultSubmissionEvent),
+	) subscription.EventSubscription
```
lukasz-zimnoch, line 124: I think the only method we need here is:
```
SubmitResult(result *Result, memberIndex group.MemberIndex, signatures map[group.MemberIndex][]byte) error
```
We should implement that interface in `pkg/tbtc`. That implementation should be responsible for submitter eligibility as well.

pkg/tecdsa/dkg/member.go
```diff
@@ -256,9 +259,382 @@ func (fm *finalizingMember) MarkInactiveMembers(
...
+// VerifyDKGResultSignatures verifies signatures received in messages from other
+// group members.
+// It collects signatures supporting only the same DKG result hash as the one
+// preferred by the current member.
+// Each member is allowed to broadcast only one signature over a preferred DKG
+// result hash.
+// The function assumes that the public key presented in the message is the
+// correct one. This key needs to be compared against the one used by network
+// client earlier, before this function is called.
+func (sm *signingMember) VerifyDKGResultSignatures(
```
lukasz-zimnoch, line 374: Let's move that function into `protocol.go` and cover with unit tests.

pkg/tecdsa/dkg/member.go
```diff
@@ -256,9 +259,382 @@ func (fm *finalizingMember) MarkInactiveMembers(
...
+		senderID,
+		senderPublicKey,
+	)
+	isSenderAccepted := sm.group.IsOperating(senderID)
+
+	return !isMessageFromSelf && isSenderValid && isSenderAccepted
+}
+
+// SubmissionConfig contains parameters describing DKG submission process.
+type SubmissionConfig struct {
```
lukasz-zimnoch, line 473: I think that struct won't be needed eventually.

pkg/tecdsa/dkg/member.go
```diff
@@ -256,9 +259,382 @@ func (fm *finalizingMember) MarkInactiveMembers(
...
+
+// SubmittingMember represents a member submitting a DKG result to the
+// blockchain along with signatures received from other group members supporting
+// the result.
+type submittingMember struct {
+	logger log.StandardLogger
+
+	// Represents the member's position for submission.
+	index  group.MemberIndex
+	config *SubmissionConfig
```
lukasz-zimnoch, line 497: Can't we just embed the `*signingMember` here and expose `initializeSubmittingMember` from there to do the conversion just like we do in other DKG states?

pkg/tecdsa/dkg/member.go
```diff
@@ -256,9 +259,382 @@ func (fm *finalizingMember) MarkInactiveMembers(
...
+// already submitted by another member it returns `0`.
+func (sm *submittingMember) SubmitDKGResult(
+	result *Result,
+	signatures map[group.MemberIndex][]byte,
+	resultSubmitter ResultSubmitter,
+	blockCounter chain.BlockCounter,
+	startBlockHeight uint64,
+) error {
+	// TODO: Consider moving the whole implementation to the resultSubmitter
+	//       interface
```
lukasz-zimnoch, line 522: That's right, the logic of submission should be controlled by the application performing DKG (so `pkg/tbtc` in our case).

pkg/tecdsa/dkg/member.go
```diff
@@ -256,9 +259,382 @@ func (fm *finalizingMember) MarkInactiveMembers(
...
+// waitForSubmissionEligibility waits until the current member is eligible to
+// submit a result to the blockchain. First member is eligible to submit straight
+// away, each following member is eligible after pre-defined block step.
+func (sm *submittingMember) waitForSubmissionEligibility(
+	blockCounter chain.BlockCounter,
+	startBlockHeight uint64,
+	blockStep uint64,
+) (<-chan uint64, error) {
+	// T_init + (member_index - 1) * T_step
+	blockWaitTime := (uint64(sm.index) - 1) * blockStep
```
lukasz-zimnoch, line 623: Something to discuss with @pdyraga: Do we want to determine the first submitter in the same way as in beacon DKG?

lukasz-zimnoch (changes requested):
First pass. More to come later.

pkg/tbtc/node.go
```diff
@@ -161,17 +164,82 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
 					return
 				}
 
+				startPublicationBlockHeight := endBlockNumber
```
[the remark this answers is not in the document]
tomaszslabon: Done in 1d575bd64a9d7ee615414a33a0fdfafcfd2b4929.

pkg/tbtc/node.go
```diff
@@ -141,7 +144,7 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
 			memberIndex := index + 1
 
 			go func() {
-				result, _, err := n.dkgExecutor.Execute(
+				result, endBlockNumber, err := n.dkgExecutor.Execute(
```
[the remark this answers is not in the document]
tomaszslabon: Done in 1d575bd64a9d7ee615414a33a0fdfafcfd2b4929.

pkg/tbtc/chain.go
```diff
@@ -59,6 +118,14 @@ type ChainConfig struct {
 	// HonestThreshold is the minimum number of active participants behaving
 	// according to the protocol needed to generate a signature.
 	HonestThreshold int
+	// ResultPublicationBlockStep is the duration (in blocks) that has to pass
+	// before group member with the given index is eligible to submit the
+	// result.
+	// Nth player becomes eligible to submit the result after
+	// T_dkg + (N-1) * T_step
+	// where T_dkg is time for phases 1-12 to complete and T_step is the result
+	// publication block step.
```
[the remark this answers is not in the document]
tomaszslabon: Done in 1d575bd64a9d7ee615414a33a0fdfafcfd2b4929.

pkg/tecdsa/dkg/dkg.go
```diff
@@ -85,6 +86,101 @@ func (e *Executor) Execute(
...
+		channel:       channel,
+		resultHandler: resultHandler,
+		blockCounter:  blockCounter,
+		member: &signingMember{
+			logger:              logger,
+			index:               memberIndex,
+			group:               result.Group,
+			membershipValidator: membershipValidator,
+			config:              submissionConfig,
+		},
```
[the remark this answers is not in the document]
tomaszslabon: Done in dea8a0788aeddd585ecd96529e2edfa770d4b87c.
I added a constructor for the `submittingMember` as well.

pkg/tecdsa/dkg/dkg.go
tomaszslabon: Done in dea8a0788aeddd585ecd96529e2edfa770d4b87c.

pkg/tecdsa/dkg/dkg.go
tomaszslabon: Done in 73582615cd626e5b6665345235e489d5b56f2383.

pkg/tbtc/chain.go
```diff
@@ -24,13 +27,68 @@ type DistributedKeyGenerationChain interface {
...
+}
+
+// GroupSelectionInterface defines the subset of the TBTC chain interface that
+// pertains to the group selection activities.
+type GroupSelectionInterface interface {
+	// SelectGroup returns the group members for the group generated by
+	// the given seed. This function can return an error if the TBTC chain's
+	// state does not allow for group selection at the moment.
+	SelectGroup(seed *big.Int) ([]chain.Address, error)
+}
```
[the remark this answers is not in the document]
tomaszslabon: Done in 12309cfa97b51f0e3901c9854b5cea3b0d589867.

pkg/tbtc/chain.go
```diff
@@ -24,13 +27,68 @@ type DistributedKeyGenerationChain interface {
...
+	// operation.
+	IsStaleGroup(groupPublicKey []byte) (bool, error)
+}
+
+// GroupInterface defines the subset of the TBTC chain interface that pertains
+// specifically to the group management.
+type GroupInterface interface {
+	GroupSelectionInterface
+	GroupRegistrationInterface
 }
```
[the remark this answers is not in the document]
tomaszslabon: Done in 12309cfa97b51f0e3901c9854b5cea3b0d589867.

pkg/tecdsa/dkg/dkg.go
tomaszslabon: Done in c46033aafc6a094ee8380749b1b0ac164db3cb21.

pkg/tecdsa/dkg/dkg.go
tomaszslabon: Done in e3e4bcd23c969076d0a7a55673f4c1dbc8dd38a3. although the interface looks a bit different.

pkg/tecdsa/dkg/member.go
tomaszslabon: Removed in b4b0f607f3967810ebd934fe99602b59ef8e7c80.

pkg/tecdsa/dkg/gen/pb/message.proto
```diff
@@ -25,4 +25,11 @@ message TSSRoundThreeMessage {
     uint32 senderID = 1;
     bytes payload = 2;
     string sessionID = 3;
-}
\ No newline at end of file
+}
+
+message DKGResultHashSignature {
```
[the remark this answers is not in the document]
tomaszslabon: Done in c65086902a608faac1dc6cf7990e33f85f6c74b8.

pkg/tecdsa/dkg/message.go
```diff
@@ -90,3 +90,26 @@ func (trtm *tssRoundThreeMessage) SenderID() group.MemberIndex {
 func (trtm *tssRoundThreeMessage) Type() string {
 	return messageTypePrefix + "tss_round_three_message"
 }
+
+// dkgResultHashSignatureMessage is a message payload that carries a hash of
+// the DKG result and a signature over this hash for a DKG result.
+//
+// It is expected to be broadcast within the group.
+type dkgResultHashSignatureMessage struct {
```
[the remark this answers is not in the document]
tomaszslabon: Done in 7206caae68cc470455f2e4130e78b9b47f4b05ed.

pkg/tecdsa/dkg/message.go
```diff
@@ -90,3 +90,26 @@ func (trtm *tssRoundThreeMessage) SenderID() group.MemberIndex {
...
+
+// SenderID returns protocol-level identifier of the message sender.
+func (drhsm *dkgResultHashSignatureMessage) SenderID() group.MemberIndex {
+	return drhsm.senderID
+}
+
+// Type returns a string describing an dkgResultHashSignatureMessage type for
+// marshaling purposes.
+func (drhsm *dkgResultHashSignatureMessage) Type() string {
+	return messageTypePrefix + "result_hash_signature_message"
```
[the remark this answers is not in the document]
tomaszslabon: Done in 5ac4b4e3962e2731d47b2b96dc6c350076063647.

pkg/tecdsa/dkg/dkg.go
tomaszslabon: Removed in 2c277164427d27dc6bb12c83f88bdb67a359f00f.

pkg/tecdsa/dkg/member.go
tomaszslabon: Done in be88ead76ae30ea4f9dd01d5b2f00342e374b986.

pkg/tecdsa/dkg/member.go
tomaszslabon: Done in e3e4bcd23c969076d0a7a55673f4c1dbc8dd38a3.

pkg/tbtc/chain.go
```diff
@@ -24,13 +27,68 @@ type DistributedKeyGenerationChain interface {
...
+	// member index.
+	SubmitDKGResult(
+		participantIndex group.MemberIndex,
+		dkgResult *dkg.Result,
+		signatures map[group.MemberIndex][]byte,
+	) error
+	// OnDKGResultSubmitted registers a callback that is invoked when an on-chain
+	// notification of a new, valid submitted result is seen.
+	OnDKGResultSubmitted(
+		func(event *dkg.ResultSubmissionEvent),
```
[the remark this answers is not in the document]
tomaszslabon: Done in 310f98d52ea5cc4b00ae02c983b04b74b3c9eba8.

pkg/tecdsa/dkg/dkg.go
```diff
@@ -85,6 +86,101 @@ func (e *Executor) Execute(
...
+// the index of the member who submitted the result and a final public key of
+// the group.
+type ResultSubmissionEvent struct {
+	MemberIndex         uint32
+	GroupPublicKeyBytes []byte
+	Misbehaved          []uint8
+
+	BlockNumber uint64
+}
+
```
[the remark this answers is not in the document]
tomaszslabon: Done in 310f98d52ea5cc4b00ae02c983b04b74b3c9eba8.

pkg/tbtc/chain.go
```diff
@@ -24,13 +27,68 @@ type DistributedKeyGenerationChain interface {
 	// OnDKGStarted registers a callback that is invoked when an on-chain
 	// notification of the DKG process start is seen.
 	OnDKGStarted(
-		func(event *DKGStartedEvent),
+		func(event *dkg.StartedDKGEvent),
```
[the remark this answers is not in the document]
tomaszslabon: Done in 310f98d52ea5cc4b00ae02c983b04b74b3c9eba8.

pkg/tecdsa/dkg/result.go
```diff
@@ -1,10 +1,55 @@
...
+	// Group represents the group state, including members, disqualified,
+	// and inactive members.
+	Group *group.Group
+	// Misbehaved members are all members either inactive or disqualified.
+	// Misbehaved members are represented as a slice of bytes for optimizing
+	// on-chain storage. Each byte is an inactive or disqualified member index.
+	Misbehaved []byte
+	// Group public key bytes generated by the protocol execution, empty if the
+	// protocol failed.
+	GroupPublicKeyBytes []byte
```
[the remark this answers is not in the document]
tomaszslabon: Done in f29f201134e1db2d627c0e481cd2244cf4ea85a4

pkg/tecdsa/dkg/result.go
```diff
@@ -1,10 +1,55 @@
...
+
+	"github.com/keep-network/keep-core/pkg/protocol/group"
+	"github.com/keep-network/keep-core/pkg/tecdsa"
+)
 
 // Result of distributed key generation protocol.
 type Result struct {
+	// Group represents the group state, including members, disqualified,
+	// and inactive members.
+	Group *group.Group
```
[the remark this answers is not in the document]
tomaszslabon: I think we need it.
It is used [when we check is operator is a member of the group](https://github.com/keep-network/keep-core/blob/f29f201134e1db2d627c0e481cd2244cf4ea85a4/pkg/tecdsa/dkg/member.go#L442).

We could remove the `Misbehaved` from the `dkg.Result` structure, though.
We could filter the misbehaved (inactive or disqualified) members on the spot, when this data is needed.
It would of course be slower than caching it, but the code would be cleaner.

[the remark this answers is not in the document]
lukasz-zimnoch: You need it to be a part of the `signingMember` but I'm talking about the `Result` struct here. I think we don't need to return anything else than the `PrivateKeyShare` and `Misbehaved` array. Am I wrong?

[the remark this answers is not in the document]
tomaszslabon: But the `signingMember` gets its `group` field from the `Result`.
But in theory we can remove `group` from the `Result` and return it from `n.dkgExecutor.Execute(...)` call.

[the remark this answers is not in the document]
lukasz-zimnoch: I think we did not understand each other. Let's break it down from the very beginning. DKG consists of two parts:
1. Actual protocol execution that produces the result
2. Result publication

What I mean is that part 1 should return the minimal required amount of information without redundancy. If we return the `Result` that holds key share and misbehaved + the `group` separately, we have redundancy as misbehaved can be inferred from the `group`. This is not something we should do here. If you need the `group` for part 2, we should do the following:

1. Part 1 should return `Result` that contains only `PrivateKeyShare` and `Group`. Nothing else. Misbehaved members can be taken from `Group`
2. Part 2 should take `Result` as parameter. It should use the `Group` to create the signing member. It should build the misbehaved array in the format expected by the on-chain contract.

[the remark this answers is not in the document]
tomaszslabon: Removed the `Misbehaved` in d5d7ef1ed7ed6c7418216ab8728be35c3f5ad4c5.
Calculating the `Misbehaved` from the `Result.Group` could be done in a few ways. 
I opted for adding `MisbehavedMembersBytes()` function to the `Result`, so the misbehaved can be obtained anywhere they are needed. But it could very well be done differently.

pkg/tecdsa/dkg/marshaling.go
```diff
@@ -187,3 +187,39 @@ func unmarshalPublicKeyMap(
 
 	return unmarshalled, nil
 }
+
+// Marshal converts this dkgResultHashSignatureMessage to a byte array suitable
+// for network communication.
+func (d *dkgResultHashSignatureMessage) Marshal() ([]byte, error) {
```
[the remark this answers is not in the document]
tomaszslabon: Doen in 267ec48c7ab327c86809cc49c0e38565e0cd8b3c.

pkg/tecdsa/dkg/member.go
tomaszslabon: Done in 78fdd1ad719aa298a5bc9f10705288e5989d7571.

pkg/tbtc/chain.go
```diff
@@ -24,13 +27,68 @@ type DistributedKeyGenerationChain interface {
...
-type DKGStartedEvent struct {
-	Seed        *big.Int
+// GroupRegistrationEvent represents an event of registering a new group with the
+// given public key.
+// TODO: Adjust to the v2 WalletRegistry contract and rename to GroupRegistered.
+type GroupRegistrationEvent struct {
+	GroupPublicKey []byte
+
 	BlockNumber uint64
 }
```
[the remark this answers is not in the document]
tomaszslabon: Done in 73abf3b4b1ddafad06e1f7b05759e4ff79235ae3.

pkg/tbtc/chain.go
```diff
@@ -24,13 +27,68 @@ type DistributedKeyGenerationChain interface {
...
+type GroupSelectionInterface interface {
+	// SelectGroup returns the group members for the group generated by
+	// the given seed. This function can return an error if the TBTC chain's
+	// state does not allow for group selection at the moment.
+	SelectGroup(seed *big.Int) ([]chain.Address, error)
+}
+
+// GroupRegistrationInterface defines the subset of the TBTC chain interface
+// that pertains to the group registration activities.
+type GroupRegistrationInterface interface {
```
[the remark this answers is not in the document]
tomaszslabon: Removed the interface in 73abf3b4b1ddafad06e1f7b05759e4ff79235ae3.
In beacon, we used this interface for example to check if DKG was already submitted 
(function `IsGroupRegistered(groupPublicKey []byte) (bool, error)`.

If I understand correctly, we should add some function, e.g. `IsDKGAlreadySubmitted` that will do the same for us (check if it's already submitted), right?

pkg/tbtc/node.go
```diff
@@ -173,17 +176,77 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+
 				// TODO: Snapshot the key material before doing on-chain result
 				//       submission.
 
-				// TODO: Submit the result using the chain layer.
-
 				// TODO: The final `signingGroupOperators` may differ from
 				//       the original `selectedSigningGroupOperators`.
 				//       Consider that when integrating the retry algorithm.
 				signer := newSigner(
```
lukasz-zimnoch, line 247: Shouldn't we remove the above TODO as well?

pkg/tbtc/node.go
```diff
@@ -205,3 +268,136 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
 		logger.Infof("not eligible for DKG with seed [0x%x]", seed)
 	}
 }
+
+// decideMemberFate decides what the member will do in case it failed to publish
+// its DKG result. Member can stay in the group if it supports the same group
+// public key as the one registered on-chain and the member is not considered as
+// misbehaving by the group.
+func (n *node) decideMemberFate(
```
lukasz-zimnoch, line 276: Let's be specific: s/decideMemberFate/decideSigningGroupMemberFate. Also, let's remember about the change in the comment.

pkg/tbtc/node.go
```diff
@@ -173,17 +176,77 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
 					return
 				}
 
+				publicationStartBlock := endBlock
+				operatingMemberIDs := result.Group.OperatingMemberIDs()
```
lukasz-zimnoch, line 180: Let's avoid using `ID`: s/operatingMemberIDs/operatingMemberIndexes

pkg/tbtc/node.go
```diff
@@ -153,7 +156,7 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
 				n.protocolLatch.Lock()
 				defer n.protocolLatch.Unlock()
 
-				result, _, err := n.dkgExecutor.Execute(
+				result, endBlock, err := n.dkgExecutor.Execute(
```
lukasz-zimnoch, line 159: s/endBlock/executionEndBlock

pkg/tbtc/node.go
```diff
@@ -205,3 +268,136 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+		return nil, fmt.Errorf(
+			"[member:%v] could not stay in the group because "+
+				"the member is considered as misbehaving",
+			memberIndex,
+		)
+	}
+
+	// Construct a new view of the operating members according to the accepted
+	// DKG result.
+	operatingMemberIDs := make([]group.MemberIndex, 0)
```
lukasz-zimnoch, line 321: s/operatingMemberIDs/operatingMemberIndexes

pkg/tbtc/node.go
```diff
@@ -205,3 +268,136 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+// slice into another slice that contains addresses of all operators
+// that were neither disqualified nor marked as inactive. This way, the
+// resulting slice has only addresses of properly operating operators
+// who form the resulting group.
+//
+// Example:
+// selectedOperators: [member1, member2, member3, member4, member5]
+// operatingGroupMembersIDs: [5, 1, 3]
+// signingGroupOperators: [member1, member3, member5]
+func (n *node) resolveSigningGroupOperators(
```
lukasz-zimnoch, line 378: s/resolveSigningGroupOperators/resolveFinalSigningGroupOperators. Also, let's remember about updating the comment.

pkg/tbtc/node.go
```diff
@@ -205,3 +268,136 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+	selectedOperators []chain.Address,
+	operatingGroupMembersIDs []group.MemberIndex,
```
lukasz-zimnoch, line 380: s/operatingGroupMembersIDs/operatingMembersIndexes. Also, let's remember about updating the comment.

pkg/tbtc/node.go
```diff
@@ -205,3 +268,136 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+) ([]chain.Address, error) {
+	config := n.chain.GetConfig()
+
+	if len(selectedOperators) != config.GroupSize ||
+		len(operatingGroupMembersIDs) < config.HonestThreshold {
+		return nil, fmt.Errorf("invalid input parameters")
+	}
```
lukasz-zimnoch, line 387: In practice, we will have a `GroupQuorum` parameter (see #3169) so we should check against it instead of the `HonestThreshold`. Let's leave a TODO about changing it.

pkg/tbtc/chain.go
```diff
@@ -26,6 +29,15 @@ type DistributedKeyGenerationChain interface {
...
 	) subscription.EventSubscription
+	OnDKGResultSubmitted(
+		func(event *DKGResultSubmittedEvent),
+	) subscription.EventSubscription
+	SubmitResult(
+		result *dkg.Result,
+		signatures map[group.MemberIndex][]byte,
+		startBlockNumber uint64,
+		participantIndex group.MemberIndex,
+	) error
```
lukasz-zimnoch, line 40: Let's separate functions with blank lines and document them.

pkg/tbtc/chain.go
```diff
@@ -26,6 +29,15 @@ type DistributedKeyGenerationChain interface {
 	OnDKGStarted(
 		func(event *DKGStartedEvent),
 	) subscription.EventSubscription
+	OnDKGResultSubmitted(
+		func(event *DKGResultSubmittedEvent),
+	) subscription.EventSubscription
+	SubmitResult(
```
lukasz-zimnoch, line 35: The `dkg.ResultSubmitter` and `dkg.ResultSigner` should not be directly implemented in the `pkg/chain/ethereum` as we also have application-specific logic that should be implemented in `pkg/tbtc`. Here is what I propose:

1. In `node.go`, create two package-private structs `dkgResultSubmitter` and `dkgResultSigner` that will implement `dkg.ResultSubmitter` and `dkg.ResultSigner` respectively. Both should have references to the`tbtc.Chain`
2. The aforementioned types should do all the app-specific actions like submission eligibility and so on, and then call respective functions of the `tbtc.Chain`. Chain functions should be as small as possible and focus only on things required to interact with the contract. Nothing else. In other words:

    - `dkgResultSubmitter` should call chain's `OnDKGResultSubmitted` and `SubmitDKGResult`
    - `dkgResultSigner` should call functions of the signer returned by chain's `Signing` function and the `CalculateDKGResultHash` function.
3. Add a `TODO` above `node.joinDKGIfEligible` about extracting DKG-specific code into a separate file `pkg/tbtc/dkg.go`. I'll address it once #3169 is merged as the DKG stuff become really big and will clutter the `node.go` file quickly.

lukasz-zimnoch (changes requested):

pkg/tecdsa/dkg/member.go
```diff
@@ -258,10 +258,105 @@ func (fm *finalizingMember) MarkInactiveMembers(
...
+// initializeSubmittingMember performs a transition of a member state to the
+// next phase of the protocol.
+func (sm *signingMember) initializeSubmittingMember() *submittingMember {
+	return &submittingMember{
+		signingMember: sm,
+	}
+}
+
+// newSigningMember creates a new signingMember in the initial state.
+func newSigningMember(
```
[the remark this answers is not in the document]
tomaszslabon: Done in b0732874cd5d5e59bde5da2a5e28f85004f000e8.

pkg/tecdsa/dkg/member.go
```diff
@@ -258,10 +258,105 @@ func (fm *finalizingMember) MarkInactiveMembers(
...
+// If allowed, it submits the result to the chain.
+// A user's turn to publish is determined based on the user's index and block
+// step.
+// If a result is submitted by another member and it's accepted by the chain,
+// the current member finishes the phase immediately, without submitting
+// their own result.
+// It returns the on-chain block height of the moment when the result was
+// successfully submitted on chain by the member. In case of failure or result
+// already submitted by another member it returns `0`.
+func (sm *submittingMember) SubmitDKGResult(
```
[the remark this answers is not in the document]
tomaszslabon: Done in bd4aa1aba4b6fecdcbb2073dc98962255fa98de3.

pkg/tecdsa/dkg/member.go
```diff
@@ -258,10 +258,105 @@ func (fm *finalizingMember) MarkInactiveMembers(
...
+// it determines if the current member is eligible to submit a result.
+// If allowed, it submits the result to the chain.
+// A user's turn to publish is determined based on the user's index and block
+// step.
+// If a result is submitted by another member and it's accepted by the chain,
+// the current member finishes the phase immediately, without submitting
+// their own result.
+// It returns the on-chain block height of the moment when the result was
+// successfully submitted on chain by the member. In case of failure or result
+// already submitted by another member it returns `0`.
```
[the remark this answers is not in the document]
tomaszslabon: Done in 8baa33fbe3130296d47dfa257641e3a8054ecd00.

pkg/tecdsa/dkg/states.go
```diff
@@ -435,3 +448,206 @@ func (fs *finalizationState) MemberIndex() group.MemberIndex {
...
+
+func (rss *resultSubmissionState) DelayBlocks() uint64 {
+	return state.SilentStateDelayBlocks
+}
+
+func (rss *resultSubmissionState) ActiveBlocks() uint64 {
+	// We do not exchange any messages in this phase. We publish result to the
+	// chain but it is an action blocking all group members for the same time
+	// - members exit when the first valid result is accepted by the chain.
+	// How long it takes depends on the block step and group size.
```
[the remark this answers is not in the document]
tomaszslabon: Done in bb050e123ffcf1b0115555063b76d50e3b9ddf0a.

pkg/tecdsa/dkg/result.go
```diff
@@ -1,10 +1,90 @@
...
+		publicKey.Curve,
+		publicKey.X,
+		publicKey.Y,
+	), nil
+}
+
+// MisbehavedMembersBytes returns the IDs of group members that misbehaved during
+// the DKG procedure. The IDs are sorted and returned as bytes.
+func (r *Result) MisbehavedMembersBytes() []byte {
+	// Merge IA and DQ into 'misbehaved' set.
```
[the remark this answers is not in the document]
tomaszslabon: Done in 8e74aa2020bcd399a54494133881a2dfd15b19a8.

pkg/tecdsa/dkg/result.go
```diff
@@ -1,10 +1,90 @@
...
+
+	return elliptic.Marshal(
+		publicKey.Curve,
+		publicKey.X,
+		publicKey.Y,
+	), nil
+}
+
+// MisbehavedMembersBytes returns the IDs of group members that misbehaved during
+// the DKG procedure. The IDs are sorted and returned as bytes.
```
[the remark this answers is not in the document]
tomaszslabon: Done in 8e74aa2020bcd399a54494133881a2dfd15b19a8.

pkg/tecdsa/dkg/result.go
```diff
@@ -1,10 +1,90 @@
...
+	// Convert sorted list of member indexes into bytes.
+	bytes := make([]byte, len(sorted))
+	for i, m := range sorted {
+		bytes[i] = byte(m)
+	}
+
+	return bytes
+}
+
+const hashByteSize = 32
```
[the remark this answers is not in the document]
tomaszslabon: Done in 1eeb72251a8d2fccc5f3b0524d841d18a12fdf3d.

pkg/tecdsa/dkg/result.go
```diff
@@ -1,10 +1,90 @@
...
+
+// ResultHash is a 256-bit hash of DKG Result. The hashing algorithm should
+// be the same as the one used on-chain.
+type ResultHash [hashByteSize]byte
+
+// ResultsVotes is a map of votes for each DKG Result.
+type ResultsVotes map[ResultHash]int
```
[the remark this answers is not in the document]
tomaszslabon: It's not. Removed in 1eeb72251a8d2fccc5f3b0524d841d18a12fdf3d.

pkg/tecdsa/dkg/message.go
```diff
@@ -90,3 +90,27 @@ func (trtm *tssRoundThreeMessage) SenderID() group.MemberIndex {
...
+	senderID group.MemberIndex
+
+	resultHash ResultHash
+	signature  []byte
+	publicKey  []byte
+	sessionID  string
+}
+
+// SenderID returns protocol-level identifier of the message sender.
+func (drhsm *resultSignatureMessage) SenderID() group.MemberIndex {
```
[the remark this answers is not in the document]
tomaszslabon: Done in 3da992131aacf431517132146e586a32b732f1da.

pkg/tecdsa/dkg/dkg.go
```diff
@@ -95,6 +95,74 @@ func (e *Executor) Execute(
...
+		startBlockNumber uint64,
+		participantIndex group.MemberIndex,
+	) error
+}
+
+// Publish signs the DKG result for the given group member, collects signatures
+// from other members and verifies them, and submits the DKG result.
+func Publish(
+	logger log.StandardLogger,
+	seed *big.Int,
```
lukasz-zimnoch, line 124: Let's pass the `sessionID string` directly (see #3169)

pkg/tecdsa/dkg/protocol_test.go
```diff
@@ -1542,3 +1642,39 @@ func newTssPreParams(
...
 	}
 }
+
+type signingData struct {
+	signature  []byte
+	resultHash ResultHash
+	err        error
+}
+
+type testResultSigner struct {
```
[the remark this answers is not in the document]
tomaszslabon: Done in 673aa2f6fd5012f34aa32f7191f911f4a31320b4.

pkg/tecdsa/dkg/protocol_test.go
```diff
@@ -1542,3 +1642,39 @@ func newTssPreParams(
...
+	return nil, nil, ResultHash{}, fmt.Errorf(
+		"could not find singing data for the result",
+	)
+}
+
+func (trs *testResultSigner) VerifySignature(message []byte, signature []byte, publicKey []byte) (bool, error) {
+	return false, nil
+}
+
+func newTestResultSigner(publicKey []byte) *testResultSigner {
```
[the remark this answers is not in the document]
tomaszslabon: Done in 673aa2f6fd5012f34aa32f7191f911f4a31320b4.

pkg/tecdsa/dkg/protocol_test.go
```diff
@@ -1542,3 +1642,39 @@ func newTssPreParams(
 		Q:       newBigInt(q),
 	}
 }
+
+type signingData struct {
```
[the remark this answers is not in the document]
tomaszslabon: Done in 673aa2f6fd5012f34aa32f7191f911f4a31320b4.

pkg/tecdsa/dkg/protocol_test.go
```diff
@@ -1542,3 +1642,39 @@ func newTssPreParams(
...
+	resultHash ResultHash
+	err        error
+}
+
+type testResultSigner struct {
+	publicKey      []byte
+	signingResults map[*Result]*signingData
+}
+
+func (trs *testResultSigner) setSigningResult(result *Result, data *signingData) {
```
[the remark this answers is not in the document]
tomaszslabon: Done in 673aa2f6fd5012f34aa32f7191f911f4a31320b4.

pkg/tecdsa/dkg/protocol_test.go
```diff
@@ -1542,3 +1642,39 @@ func newTssPreParams(
...
+
+type signingData struct {
+	signature  []byte
+	resultHash ResultHash
+	err        error
+}
+
+type testResultSigner struct {
+	publicKey      []byte
+	signingResults map[*Result]*signingData
```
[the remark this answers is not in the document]
tomaszslabon: Done in 673aa2f6fd5012f34aa32f7191f911f4a31320b4.

pkg/tecdsa/dkg/protocol_test.go
```diff
@@ -1140,6 +1140,97 @@ func TestTssFinalize_ResultTimeout(t *testing.T) {
...
+		t.Errorf(
+			"unexpected self DKG result signature\nexpected: %v\nactual:   %v\n",
+			signature,
+			signingMember.selfDKGResultSignature,
+		)
+	}
+
+	if !reflect.DeepEqual(
+		resultHash,
+		signingMember.preferredDKGResultHash,
```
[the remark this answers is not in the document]
tomaszslabon: Done in 54f7751c59f22ab2d866a1cd5476137b05fee36c.

pkg/tecdsa/dkg/protocol_test.go
```diff
@@ -1140,6 +1140,97 @@ func TestTssFinalize_ResultTimeout(t *testing.T) {
...
+			"unexpected signature message \nexpected: %v\nactual:   %v\n",
+			expectedSignatureMessage,
+			actualSignatureMessage,
+		)
+	}
+
+	if !reflect.DeepEqual(
+		signature,
+		signingMember.selfDKGResultSignature,
+	) {
```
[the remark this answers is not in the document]
tomaszslabon: Done in 54f7751c59f22ab2d866a1cd5476137b05fee36c.

pkg/tecdsa/dkg/dkg.go
```diff
@@ -95,6 +95,74 @@ func (e *Executor) Execute(
...
+}
+
+// ResultSubmitter is the interface that provides ability to submit the DKG
+// result to the chain.
+type ResultSubmitter interface {
+	SubmitResult(
+		result *Result,
+		signatures map[group.MemberIndex][]byte,
+		startBlockNumber uint64,
+		participantIndex group.MemberIndex,
```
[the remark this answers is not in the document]
tomaszslabon: Done in 2f324a01ca04e106c2361f4dac775e49f721fbdc.

pkg/tbtc/node.go
tomaszslabon: Done in 4c76a72cb145b310ff8710fa1b778d0382716ac9.

pkg/tbtc/node.go
tomaszslabon: Done in 0bfc371f27dfbdf88d59e76e244a06f03dad8422.

pkg/tbtc/node.go
tomaszslabon: Done in 54916c43296c6911d2229b5ad1af34509c3d77cf.

pkg/tbtc/node.go
tomaszslabon: Done in 54916c43296c6911d2229b5ad1af34509c3d77cf.

pkg/tbtc/node.go
tomaszslabon: Done in 54916c43296c6911d2229b5ad1af34509c3d77cf.

pkg/tbtc/node.go
tomaszslabon: Done in 54916c43296c6911d2229b5ad1af34509c3d77cf.

pkg/tbtc/node.go
tomaszslabon: Done in 54916c43296c6911d2229b5ad1af34509c3d77cf.

pkg/tbtc/chain.go
tomaszslabon: Done in 1c60971644aeb94e2222261931a1d94f2d265d03.

pkg/tecdsa/dkg/dkg.go
```diff
@@ -95,6 +95,74 @@ func (e *Executor) Execute(
 	return finalizationState.result(), endBlockNumber, nil
 }
 
+// ResultSigner is the interface that provides ability to sign the DKG result
+// and verify the results received from other group members.
+type ResultSigner interface {
+	// SignResult signs the provided DKG result. It returns the public key used
+	// during the signing, the resulting signature and the result hash.
+	SignResult(result *Result) ([]byte, []byte, ResultHash, error)
```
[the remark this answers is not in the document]
tomaszslabon: Done in eded771375fb35942e6d0e5c0e08d168b83905f6.

pkg/tecdsa/dkg/dkg.go
tomaszslabon: Done in 62983725edba3fe648e1dda23d6fe01e831799cc.

pkg/tbtc/chain.go
tomaszslabon: Done in 655ee9ba9b2f7e587a16d5c119fd843909339517.

pkg/tbtc/node.go
lukasz-zimnoch: Nice! Leaving it unresolved for reference. I made a note about addressing it in #3041

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
 	"github.com/keep-network/keep-core/pkg/generator"
 	"github.com/keep-network/keep-core/pkg/internal/testutils"
 	"github.com/keep-network/keep-core/pkg/net"
 	"github.com/keep-network/keep-core/pkg/protocol/group"
 	"github.com/keep-network/keep-core/pkg/tecdsa/dkg"
 )
 
+// dkgResultSigner is responsible for signing the DKG result and verification of
+//  signatures generated by other group members.
+type dkgResultSigner struct {
```
lukasz-zimnoch, line 20: The `dkgResultSigner` and `dkgResultSubmitter` look very well! Few comments about them:
- Let's place them at the bottom of the file. The main and most important type of this file is `node` so it would be good to keep it at the top. 
- Let's remove the redundant space from the `dkgResultSigner` docstring.
- Let's add unit tests that will cover both types' methods

pkg/tbtc/chain.go
```diff
@@ -26,6 +29,28 @@ type DistributedKeyGenerationChain interface {
...
+	// over result hash from group participants supporting the result.
+	SubmitDKGResult(
+		memberIndex group.MemberIndex,
+		result *dkg.Result,
+		signatures map[group.MemberIndex][]byte,
+	) error
+
+	// TODO: Check what should be the argument for this function.
+	// IsDKGResultSubmitted checks whether the DKG result was already submitted.
+	IsDKGResultSubmitted(groupPublicKeyBytes []byte) (bool, error)
```
lukasz-zimnoch, line 49: I'm wondering whether we can replace that function with a function that will return the current DKG state. The `WalletRegistry` contract exposes `getWalletCreationState` function. I think we can introduce a `GetDKGState` function to the `DistributedKeyGenerationChain` that will return an enum value corresponding to DKG states and will call contract's `getWalletCreationState` underneath in the future. Then, we can use this function to check whether the result was already submitted in the `tbtc.dkgResultSubmitter`. WDYT?

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+	startBlockNumber uint64,
+	memberIndex group.MemberIndex,
+) error {
+	config := drs.chain.GetConfig()
+
+	// Chain rejects the result if it has less than 25% safety margin.
+	// If there are not enough signatures to preserve the margin, it does not
+	// make sense to submit the result.
+	signatureThreshold := config.HonestThreshold +
+		(config.GroupSize-config.HonestThreshold)/2
```
lukasz-zimnoch, line 95: Is this the case for `WalletRegistry` result submission? I don't see where this is forced on-chain. The result submission is optimistic. For now, we should check whether the number of signatures is above `HonestThreshold` and we should leave a TODO about switching to `GroupQuorum` eventually.

pkg/tecdsa/dkg/member.go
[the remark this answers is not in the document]
tomaszslabon: Done in f8a673b60e90430e3ba96a3e1755a41b7207633f.

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+		subscription.Unsubscribe()
+		close(onSubmittedResultChan)
+		return err
+	}
+
+	groupPublicKeyBytes, err := result.GroupPublicKeyBytes()
+	if err != nil {
+		return returnWithError(
+			fmt.Errorf(
+				"could not extract public key bytes from the result: [%v]",
```
lukasz-zimnoch, line 122: Let's always use `[%w]` for error wrapping. This is the recommended way to wrap errors that allows using `errors.As` and `errors.Is` helpers if needed:
```suggestion
				"could not extract public key bytes from the result: [%w]",
```
Let's apply that rule for all errors in this file.

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+		func(event *DKGResultSubmittedEvent) {
+			onSubmittedResultChan <- event.BlockNumber
+		},
+	)
+
+	returnWithError := func(err error) error {
+		subscription.Unsubscribe()
+		close(onSubmittedResultChan)
+		return err
+	}
```
lukasz-zimnoch, line 116: I think we should get rid of the `returnWithError`. This is copied from the beacon but we can improve things a little bit:
- We should call `defer subscription.Unsubscribe()` just after calling `OnDKGResultSubmitted`
- There is no need to close `onSubmittedResultChan`. We don't need their close signal. Regarding resource management, it will be garbage collected once it become unreachable

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+	if err != nil {
+		return returnWithError(
+			fmt.Errorf(
+				"could not extract public key bytes from the result: [%v]",
+				err,
+			),
+		)
+	}
+
+	alreadySubmitted, err := drs.chain.IsDKGResultSubmitted(groupPublicKeyBytes)
```
lukasz-zimnoch, line 128: Following up on the comment from `pkg/tbtc/chain.go`, we can use the `GetDKGState` function and check whether its result is other than `AWAITING_RESULT`. If so, that means the result was already submitted.

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+			fmt.Errorf(
+				"could not check if the result is already submitted: [%v]",
+				err,
+			),
+		)
+	}
+
+	// Someone who was ahead of us in the queue submitted the result. Giving up.
+	if alreadySubmitted {
+		return returnWithError(nil)
```
lukasz-zimnoch, line 140: Let's log an info about that:
```
logger.Infof("[member:%v] DKG is no longer awaiting the result; aborting DKG result submission", memberIndex)
```

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+
+// SubmitResult submits the DKG result along with submitting signatures to the
+// chain. In the process, it checks if the number of signatures is above
+// the required threshold, whether the result was already submitted and waits
+// until the member is eligible for DKG result submission.
+func (drs *dkgResultSubmitter) SubmitResult(
+	result *dkg.Result,
+	signatures map[group.MemberIndex][]byte,
+	startBlockNumber uint64,
+	memberIndex group.MemberIndex,
```
lukasz-zimnoch, line 87: Let's put `memberIndex` as the first argument. It's important to expose that the result submission is a per-member action.

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+			fmt.Errorf("wait for eligibility failure: [%v]", err),
+		)
+	}
+
+	for {
+		select {
+		case blockNumber := <-eligibleToSubmitWaiter:
+			// Member becomes eligible to submit the result.
+			subscription.Unsubscribe()
+			close(onSubmittedResultChan)
```
lukasz-zimnoch, line 159: This is not needed. If we entered here, there is no possibility to enter into the `<-onSubmittedResultChan` at the same time or in the future because we call the chain and exit from the submitter altogether.

pkg/tbtc/node.go
```diff
@@ -53,6 +258,8 @@ func newNode(
 	}
 }
 
+// TODO: Extract the DKG-specific code into a separate file `pkg/tbtc/dkg.go`
```
lukasz-zimnoch, line 261: Let's put this TODO just after the TODO about unit tests for `node.go`.

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+			// publishing the result.
+			return returnWithError(nil)
+		}
+	}
+}
+
+// waitForSubmissionEligibility waits until the current member is eligible to
+// submit a result to the blockchain. First member is eligible to submit straight
+// away, each following member is eligible after pre-defined block step.
+func (drs *dkgResultSubmitter) waitForSubmissionEligibility(
```
lukasz-zimnoch, line 191: Let's add a TODO about revisiting the eligibility algorithm. The RFC mentions about starting from a random member, not always the first, so we may want to think about it once again.

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+	startBlockNumber uint64,
+	memberIndex group.MemberIndex,
+) (<-chan uint64, error) {
+	// T_init + (member_index - 1) * T_step
```
lukasz-zimnoch, line 195: Let's remove that comment. It doesn't give any value.

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+	}
+
+	// Wait until the current member is eligible to submit the result.
+	eligibleToSubmitWaiter, err := drs.waitForSubmissionEligibility(
+		startBlockNumber,
+		memberIndex,
+	)
+	if err != nil {
+		return returnWithError(
+			fmt.Errorf("wait for eligibility failure: [%v]", err),
```
lukasz-zimnoch, line 150:
```suggestion
			fmt.Errorf("cannot set up eligibility queue: [%w]", err),
```

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+		(config.GroupSize-config.HonestThreshold)/2
+	if len(signatures) < signatureThreshold {
+		return fmt.Errorf(
+			"could not submit result with [%v] signatures for signature threshold [%v]",
+			len(signatures),
+			signatureThreshold,
+		)
+	}
+
+	onSubmittedResultChan := make(chan uint64)
```
lukasz-zimnoch, line 104: s/onSubmittedResultChan/resultSubmittedChan

pkg/tbtc/node.go
```diff
@@ -1,17 +1,222 @@
...
+		)
+	}
+
+	// Someone who was ahead of us in the queue submitted the result. Giving up.
+	if alreadySubmitted {
+		return returnWithError(nil)
+	}
+
+	// Wait until the current member is eligible to submit the result.
+	eligibleToSubmitWaiter, err := drs.waitForSubmissionEligibility(
```
lukasz-zimnoch, line 144: s/eligibleToSubmitWaiter/submitterEligibleChan

pkg/tbtc/node.go
lukasz-zimnoch, line 191: s/waitForSubmissionEligibility/setupEligibilityQueue. The `wait*` prefix indicates that we block here but we don't.

lukasz-zimnoch (changes requested):
Probably the last part of the comments regarding the code. Waiting for missing unit tests to make the last round of review.

pkg/tbtc/node.go
tomaszslabon: Done in 0c3003f1a1afac84613ae2c4a91410f7a8c109a5.

pkg/tbtc/node.go
tomaszslabon: Done in fcd907deeb82eadcda618343fdc5214b2abfb2e8.

pkg/tbtc/node.go
tomaszslabon: Are you sure?
If we don't call `subscription.Unsubscribe()` before calling `chainRelay.SubmitDKGResult(...)`, the handler for this particular member will not be removed from `dkgResultSubmissionHandlers` and a goroutine will be started with this handler when we process the result that we are submitting.

And since the handler writes to the channel:
```
		func(event *event.DKGResultSubmission) {
			onSubmittedResultChan <- event.BlockNumber
		},
```
the goroutine will be waiting forever to write to the channel and we will have a goroutine leak.
So in my opinion we should keep the `subscription.Unsubscribe()` before calling `chainRelay.SubmitDKGResult(...)`. An alternative would be to make the channel buffered:
```
onSubmittedResultChan := make(chan uint64, 1)
```

pkg/tbtc/node.go
tomaszslabon: Added a TODO in 083ec8d0cd8bbf586163024e2073eb64c15c2d43.

pkg/tbtc/node.go
tomaszslabon: Done in 083ec8d0cd8bbf586163024e2073eb64c15c2d43.

pkg/tbtc/node.go
tomaszslabon: Done in 9d52784867c24c8553a9f7c8a73427bf0cf4ecd4.

pkg/tbtc/node.go
tomaszslabon: Done in 8eab2eb3e1ab0a377c70180d4b1a8a5df419d953.

pkg/tbtc/node.go
tomaszslabon: Done in 8eab2eb3e1ab0a377c70180d4b1a8a5df419d953.

pkg/tbtc/node.go
tomaszslabon: Done in 8eab2eb3e1ab0a377c70180d4b1a8a5df419d953.

pkg/tbtc/node.go
tomaszslabon: Done in 77b0472a0831c90bc2f83f105ab4f3b29147efb2.

pkg/tbtc/node.go
tomaszslabon: Done in ab39e88b4d81a90ba06b0c49c4952e9cb5d7fc15.

pkg/tbtc/chain.go
tomaszslabon: Done in 59752ac2bc14d0b6dc8074c2d3a510647652188f.

However, adding `GetDKGState() (DKGState, error)` to the `DistributedKeyGenerationChain` interface forces all the 
implementations of that interface to use the same enums.
If we wanted to use a different chain than ethereum-based one, we are forced to use the same DKG states, which could be a problem.

Returning a bool would be probably more generic, because we don't care about DKG states.

pkg/tbtc/node.go
tomaszslabon: Done in 59752ac2bc14d0b6dc8074c2d3a510647652188f.

pkg/tbtc/chain.go
lukasz-zimnoch: > However, adding GetDKGState() (DKGState, error) to the DistributedKeyGenerationChain interface forces all the implementations of that interface to use the same enums.
If we wanted to use a different chain than ethereum-based one, we are forced to use the same DKG states, which could be a problem.

This is good. The underlying chain has nothing to do with DKG states. This is something specific for tBTC contracts and should be followed by all chains. Also, the client will control DKG challenges and approvals so it must be aware of DKG states.

pkg/tbtc/node.go
lukasz-zimnoch: Leaving it unresolved for future reference. I mentioned about it in #3120

pkg/tbtc/node.go
lukasz-zimnoch: Yeah, that makes sense! Let's just write a comment explaining why we need to unsubscribe here.

pkg/tbtc/node.go
lukasz-zimnoch: Good but let's put it just above `setupEligibilityQueue` function.

pkg/tbtc/node.go
```diff
@@ -205,3 +267,321 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+		select {
+		case blockNumber := <-submitterEligibleChan:
+			// Member becomes eligible to submit the result.
+			subscription.Unsubscribe()
+
+			logger.Infof(
+				"[member:%v] submitting DKG result with public key [0x%x] and "+
+					"[%v] supporting member signatures at block [%v]",
+				memberIndex,
+				result.GroupPublicKeyBytes,
```
lukasz-zimnoch, line 534: We need to call `GroupPublicKeyBytes` not pass the function reference.

pkg/tbtc/node.go
tomaszslabon: Added a comment in acefbbf3226e423dfd3ce169b9b5b465adf538a8.

pkg/tbtc/node.go
tomaszslabon: Done in 4314278a83568c8ade26e485dc60de78c7832a74.

pkg/chain/ethereum/tbtc.go
```diff
@@ -292,10 +299,69 @@ func (tc *TbtcChain) OnDKGStarted(
...
+func (tc *TbtcChain) CalculateDKGResultHash(
+	result *dkg.Result,
+) (dkg.ResultHash, error) {
+	groupPublicKeyBytes, err := result.GroupPublicKeyBytes()
+	if err != nil {
+		return dkg.ResultHash{}, err
+	}
+
+	// Encode DKG result to the format matched with Solidity keccak256(abi.encodePacked(...))
+	hash := crypto.Keccak256(groupPublicKeyBytes, result.MisbehavedMembersIndexes())
```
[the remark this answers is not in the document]
tomaszslabon: Done in 9888f430cb24d64c55c16c6806b6e685f9e1136e.

pkg/tecdsa/dkg/protocol.go
```diff
@@ -404,3 +404,115 @@ func (fm *finalizingMember) tssFinalize(
...
+// VerifyDKGResultSignatures verifies signatures received in messages from other
+// group members.
+// It collects signatures supporting only the same DKG result hash as the one
+// preferred by the current member.
+// Each member is allowed to broadcast only one signature over a preferred DKG
+// result hash.
+// The function assumes that the public key presented in the message is the
+// correct one. This key needs to be compared against the one used by network
+// client earlier, before this function is called.
+func (sm *signingMember) VerifyDKGResultSignatures(
```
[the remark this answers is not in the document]
tomaszslabon: Let's add the unit tests in a separate PR so that this one can be merged.
I left a TODO above this function about the unit tests.

pkg/tbtc/node.go
tomaszslabon: Done in b4c46c4b232dc2ac8c715ecdec51f38b60400f3b, except for the unit tests.

As discussed, I added a TODO above every function in `dkgResultSigner` and `dkgResultSubmitter` to write unit tests.
They can be added in a separate PR.

pkg/tecdsa/dkg/member.go
tomaszslabon: I added a TODO above the `setupEligibilityQueue` function to revisit it later and determine how we should choose which member submits DKG result first.

lukasz-zimnoch (approved):
LGTM! I'm good to merge once we confirm it works as expected.

lukasz-zimnoch (approved):

PR #3123, opened 2022-08-03
Initialization task for ECDSA contracts

nkuba:
This PR adds an initialization task for ECDSA contracts. It exposes the code from the random beacon to be used in ecdsa https://github.com/keep-network/keep-core/pull/3110.

### DIFF
--- a/solidity/ecdsa/tasks/initialize.ts
+++ b/solidity/ecdsa/tasks/initialize.ts
@@ -3,8 +3,11 @@ import {
   TASK_INITIALIZE,
   TASK_AUTHORIZE,
   TASK_REGISTER,
-} from "@keep-network/random-beacon/tasks/initialize"
-import { authorize, register } from "@keep-network/random-beacon/tasks/utils"
+} from "@keep-network/random-beacon/export/tasks/initialize"
+import {
+  authorize,
+  register,
+} from "@keep-network/random-beacon/export/tasks/utils"
 
 import type { HardhatRuntimeEnvironment } from "hardhat/types"
 
--- a/solidity/ecdsa/tasks/initialize.ts
+++ b/solidity/ecdsa/tasks/initialize.ts
@@ -1,0 +1,51 @@
+import { task, types } from "hardhat/config"
+
+import {
+  TASK_INITIALIZE,
+  TASK_AUTHORIZE,
+  TASK_REGISTER,
+} from "@keep-network/random-beacon/tasks/initialize"
+import type { HardhatRuntimeEnvironment } from "hardhat/types"
+
+import { authorize, register } from "@keep-network/random-beacon/tasks/utils"
+
+const TASK_AUTHORIZE_ECDSA = `${TASK_AUTHORIZE}:ecdsa`
+const TASK_REGISTER_ECDSA = `${TASK_REGISTER}:ecdsa`
+
+task(TASK_INITIALIZE, "Initializes staking for an operator").setAction(
+  async (args, hre, runSuper) => {
+    // Run initialization task from @keep-network/random-beacon.
+    await runSuper(args)
+    // Initialize ECDSA.
+    await initializeEcdsa(hre, args)
+  }
+)
+
+task(TASK_AUTHORIZE_ECDSA, "Sets authorization for ECDSA")
+  .addParam("owner", "Stake Owner address", undefined, types.string)
+  .addParam("provider", "Staking Provider", undefined, types.string)
+  .addOptionalParam("authorizer", "Stake Authorizer", undefined, types.string)
+  .addOptionalParam(
+    "authorization",
+    "Authorization amount (default: minimumAuthorization)",
+    undefined,
+    types.int
+  )
+  .setAction(async (args, hre) => {
+    await authorize(hre, "WalletRegistry", args)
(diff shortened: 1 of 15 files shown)


solidity/ecdsa/tasks/initialize.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import type { HardhatRuntimeEnvironment } from "hardhat/types"
+
+const TASK_AUTHORIZE_ECDSA = `${TASK_AUTHORIZE}:ecdsa`
+const TASK_REGISTER_ECDSA = `${TASK_REGISTER}:ecdsa`
+
+task(TASK_INITIALIZE, "Initializes staking for an operator").setAction(
+  async (args, hre, runSuper) => {
+    // Run initialization task from @keep-network/random-beacon.
+    await runSuper(args)
```
dimpar, line 20: If we were about to call task `initialize` from the script for beacon and ecdsa, then `mint` and `approve` will be called twice. I think we need to call `mint` and `approve` once, but `authorize` and `register` per application. We can call `initialize` from the beacon and call `TASK_AUTHORIZE_ECDSA` and `TASK_REGISTER_ECDSA` separately, but I take that the idea behind this PR was grouping the tasks into a single `initialize` task.

solidity/random-beacon/tasks/initialize.ts
```diff
@@ -1,14 +1,23 @@
...
-const TASK_MINT = `${TASK_INITIALIZE}:mint`
-const TASK_STAKE = `${TASK_INITIALIZE}:stake`
-const TASK_AUTHORIZE = `${TASK_INITIALIZE}:authorize`
-const TASK_REGISTER = `${TASK_INITIALIZE}:register`
+export const TASK_INITIALIZE = "initialize"
+export const TASK_MINT = `${TASK_INITIALIZE}:mint`
+export const TASK_STAKE = `${TASK_INITIALIZE}:stake`
+export const TASK_AUTHORIZE = `${TASK_INITIALIZE}:authorize`
+export const TASK_REGISTER = `${TASK_INITIALIZE}:register`
+const TASK_AUTHORIZE_BEACON = `${TASK_AUTHORIZE}:beacon`
```
dimpar, line 19: I think this requires some explanation in the comments. By looking at the beacon code it tells you nothing that half of these tasks are used by `ecdsa` and task like `TASK_AUTHORIZE_BEACON` might appear confusing, ie why not use just `TASK_AUTHORIZE`

dimpar:
Just a concern.. in general, I like the reuse of code, but what bites me a bit is that we are coupling these two projects even more by making `random-beacon` sort of a holder of these tasks. Not sure if there would be a need for some specific `beacon` or `ecdsa` authorization or registration scenarios, but if there are, then we'll need to add `ifs` to these common (as for now) tasks.

solidity/ecdsa/tasks/initialize.ts
nkuba: The `mint` and `approve` tasks are implemented in a such manner that if `initialize` is executed for the second time it will not perform any actions if the current state is fine.

So if we call `initialize` in the random-beacon and then in ecdsa minting and approval will execute transactions just once.

On the other hand if we run `initialize` in the ecdsa only, then it will execute `mint` and `approval` to get to the desired state. In such case, there is no need to call `initialize` in the beacon.

solidity/random-beacon/tasks/initialize.ts
nkuba: d2d851d87f113c9d1796d05787735fb65634dc56

nkuba:
> Just a concern.. in general, I like the reuse of code, but what bites me a bit is that we are coupling these two projects even more by making `random-beacon` sort of a holder of these tasks. 

The minting and staking part of the initialization belongs to `threshold-network/solidity-contracts`, but we define it in the `keep-network/random-beacon` for simplification.
The `keep-network/ecdsa` has the `keep-network/random-beacon` as a dependency defined anyways so we can use the beacon as our base.

We could pull the initialization stuff out to one common place but this adds too much complexity.

> Not sure if there would be a need for some specific `beacon` or `ecdsa` authorization or registration scenarios, but if there are, then we'll need to add `ifs` to these common (as for now) tasks.

If it happens we will handle it. For now, it doesn't look like we can reuse the code.

solidity/ecdsa/tasks/initialize.ts
dimpar: > On the other hand if we run initialize in the ecdsa only, then it will execute mint and approval to get to the desired state. In such case, there is no need to call initialize in the beacon.

Right, and this part doesn't feel right to me. IMO the `initialize` task in `ecdsa` project should handle the initialize tasks in `ecdsa` and probably not trigger the `initialize` task in `beacon`. I see that we have a flag for this `cd ecdsa/solidity && npx hardhat initialize --skip-beacon` but not sure if it should be there in the first place when calling it from the outside. Ecdsa depends on the beacon, but these are 2 applications where some parts are independent.

Anyway, I guess we can leave it for now and I would add a TODO to fix it later, but if you feel it's okay as-is, then up to you.

dimpar (approved):

PR #3126, opened 2022-08-04
Extract contracts artifacts downloading to the main Makefile

nkuba:
In this PR we extract artifacts downloading to the main Makefile, instead of doing it in the bindings generation process.

This is preparation for multi-network clients and supporting local developer machine's contracts artifacts.

Also, the execution experience was improved. To run generation and building for goerli environment you have to run `make goerli` instead of `CONTRACTS_NPM_PACKAGE_TAG=local go generate ./...`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,37 @@
 .PHONY: all
 
-all: generate build cmd-help 
+# environment is used as a tag of the npm packages for contracts artifacts. If
+# not overwritten it defaults to `development`.
+environment = development
+
+all: download_artifacts generate build cmd-help
+
+# List of NPM packages containing contracts needed by the client contracts bindings
+# generation.
+npm_packages := @keep-network/random-beacon \
+	@keep-network/ecdsa \
+	@threshold-network/solidity-contracts \
+	@keep-network/tbtc-v2
+
+# Working directory where contracts artifacts should be fetched to.
+contracts_dir := tmp/contracts
+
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -1,15 +1,56 @@
...
 
-all: generate build cmd-help 
+# List of NPM packages containing contracts needed by the client contracts bindings
+# generation.
+npm_packages := @keep-network/random-beacon \
+	@keep-network/ecdsa \
+	@threshold-network/solidity-contracts \
+	@keep-network/tbtc-v2
+
+# Working directory where contracts artifacts should be fetched to.
```
dimpar, line 24: [...] should be strored.

Makefile
```diff
@@ -1,15 +1,56 @@
...
 build:
+	$(info Building Go code)
 	go build -o keep-client -a . 
 
 cmd-help: build
 	@echo '$$ keep-client start --help' > docs/development/cmd-help
 	./keep-client start --help >> docs/development/cmd-help
 
-# TODO: Consider extracting `download_artifacts` step from go generate command.
+.PHONY: all
```
dimpar, line 56: What do you think about making `development` and `goerli` phony as well?

dimpar:
I've tested the execution, and the `tmp` folder with the artifacts were created successfully.

dimpar (approved):

PR #3130, opened 2022-08-05
Storage layer for `pkg/tbtc`

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3041
Depends on: #3112

Here we implement the storage layer for the tBTC application located in the `pkg/tbtc` package. The implementation is greatly inspired by the beacon storage implementation and by the https://github.com/keep-network/keep-ecdsa code for TSS-related parts. Specific changes are summarized below.

### Rename the `pkg/ecdsa` package to `pkg/tecdsa`

The actual implementation of the storage layer was preceded by the rename of the existing `pkg/ecdsa` package. That package is actually about the threshold ECDSA so, in order to avoid confusion with the built-in `crypto/ecdsa` package, we decided to rename it to `pkg/tecdsa`. This seemingly simple change has serious advantages regarding exported types naming and is an important factor to achieve consistent domain types.

### Basic domain types relevant for the storage layer

In order to have the storage layer implemented, we needed to introduce some basic domain types specific to the tBTC application and tECDSA library. That is, this PR introduces the following types:
- `tbtc.wallet` that represents an identifiable tBTC wallet with the wallet ECDSA public key as the identifier
- `tbtc.signer` that represents a wallet's signer with the signing group member index as the identifier and a tECDSA private key share that makes the signer able to participate in the threshold signing process
- `tecdsa.PrivateKeyShare` which is the tECDSA private key share generated as a result of the DKG process

Those types are in the following relation with each other:
- `tbtc.wallet` has a one-to-many relation with the `tbtc.signer`
- `tbtc.signer` has a one-to-one relation with the `tecdsa.PrivateKeyShare`

### Marshaling of the domain types

In order to persist the domain types using the storage layer, those types must implement the `Marshal` and `Unmarshal` functions allowing to convert them to and from raw bytes. That code is implemented in the `pkg/tbtc/marshaling.go` and `pkg/tecdsa/marshaling.go` files and leverages the generated protobuf types that live in `pkg/tbtc/gen` and `pkg/tecdsa/gen` packages. 

The marshaling code is covered with unit tests. Since `tecdsa.PrivateKeyShare` type is complex, we facilitate the test data setup using pre-generated fixtures placed in the `tecdsatest/testdata` directory and loaded using the `tecdsatest.LoadPrivateKeyShareTestFixtures` function. Those test fixtures are not limited to marshaling-specific tests and can be used for multiple test scenarios.

### Implementation of the wallet storage

The storage layer for `pkg/tbtc` is actually a storage mechanism for wallets. This PR adds the `tbtc.walletStorage` type that does all the heavy lifting. Its implementation is very similar to the group registry implemented in the `pkg/beacon`. The `tbtc.walletStorage` type allows to save and load wallets' signers in a thread-safe way using the underlying persistence represented by the `persistence.Handle` interface. Since the `persistence.Handle` is supposed to represent a slow persistence medium (like disk storage), the `tbtc.walletStorage` also uses an in-memory cache to speed-up reads and limit direct interactions with the underlying persistence. 

### Separate storage layer for each application

We now have two applications (`pkg/beacon` and `pkg/tbtc`) with their own storage layers. Both layers are initialized with a disk-based encrypted persistence during the `start` command (`pkg/cmd/start.go`) and use separate directories for their own data. Such an approach should help prevent data corruption between applications and is more flexible regarding app-specific changes.


pkg/tbtc/storage.go
```diff
@@ -0,0 +1,164 @@
...
+	"crypto/elliptic"
+	"encoding/hex"
+	"fmt"
+	"github.com/keep-network/keep-common/pkg/persistence"
+	"sync"
+)
+
+// walletStorage is the component that persists data of the wallets managed
+// by the given node. All functions of the storage are safe for concurrent use.
+type walletStorage struct {
```
pdyraga, line 14: I think we should separate the in-memory cache from the storage layer. We had this separation in [v1 ECDSA implementation](https://github.com/keep-network/keep-ecdsa/tree/main/pkg/registry) and we have this separation in [v1/v2 Beacon implementation](https://github.com/keep-network/keep-core/tree/main/pkg/beacon/registry). 

Right now the implementation is simple, but we must add snapshots and logic to clean up the storage from wallets that should be archived. Once we do that, without the separation, we will end up with a structure responsible for calling disk read/writes, marshalling/unmarshalling, fetching information from the chain, and maintaining an in-memory cache. This is way too much responsibility in a single structure.

I am fine with having two files or just one Go file, but I really think we should have two structures:
- one handling in-memory cache and in the future calling the chain to clean up the local storage,
- one handling marshaling/unarshaling and interacting with disk IO API.

pkg/tbtc/storage_test.go
```diff
@@ -0,0 +1,139 @@
...
+	})
+	return nil
+}
+
+func (mph *mockPersistenceHandle) Snapshot(
+	data []byte,
+	directory string,
+	name string,
+) error {
+	return nil
```
pdyraga, line 99: Should we `panic` instead? This is not implemented and would be better to panic for now: no code should call this function now and before it is called, it has to be implemented. Same question about `Archive`.

pkg/tecdsa/marshaling.go
```diff
@@ -0,0 +1,156 @@
...
+	"fmt"
+	"github.com/binance-chain/tss-lib/crypto"
+	"github.com/binance-chain/tss-lib/crypto/paillier"
+	"github.com/binance-chain/tss-lib/ecdsa/keygen"
+	"github.com/keep-network/keep-core/pkg/tecdsa/gen/pb"
+	"math/big"
+)
+
+// Marshal converts the PrivateKeyShare to a byte array.
+func (pks *PrivateKeyShare) Marshal() ([]byte, error) {
```
pdyraga, line 13: In `Unmarshal` we assume the curve to be `tecdsa.Curve`. Should we validate the curve when marshaling? Just like noted in one of the previous comments, I think comparing the name should be enough. I see the curve is available in `LocalPartySaveData.ECDSAPub.Curve()`.

pkg/tecdsa/marshaling.go
```diff
@@ -0,0 +1,156 @@
+package tecdsa
```
pdyraga, line 1: For potential fellow reviewers who has not worked on v1: you want to see https://github.com/keep-network/keep-ecdsa/blob/main/pkg/ecdsa/tss/marshaling.go.

pkg/tbtc/gen/pb/wallet.proto
```diff
@@ -0,0 +1,15 @@
...
+option go_package = "pb";
+package tbtc;
+
+message Wallet {
+    bytes publicKey = 1;
+    repeated string signingGroupOperators = 2;
+}
+
+message Signer {
+    bytes wallet = 1;
```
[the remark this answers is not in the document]
lukasz-zimnoch: That was my first try but if we do that, we cannot reuse `wallet.Marshal()`  in `signer.Marshal()`.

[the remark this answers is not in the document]
pdyraga: Maybe I am not following something but do we have to call `s.wallet.Marshal()` when marshaling the signer? We can let the protobuf do this work, right? We never marshal wallet separately, it's always a part of the signer.

pkg/internal/tecdsatest/tecdsatest.go
```diff
@@ -0,0 +1,69 @@
...
+// LoadPrivateKeyShareTestFixtures loads tECDSA private key share test data.
+// Code copied from:
+//   https://github.com/binance-chain/tss-lib/blob/master/ecdsa/keygen/test_utils.go
+// Test data JSON files copied from:
+//   https://github.com/binance-chain/tss-lib/tree/master/test/_ecdsa_fixtures
+func LoadPrivateKeyShareTestFixtures(count int) (
+	[]keygen.LocalPartySaveData,
+	error,
+) {
+	makeTestFixtureFilePath := func(partyIndex int) string {
```
beaushinkle, line 27: Are we able to break out `makeTestFixtureFilePath` to a separate function?

beaushinkle: oh wait, this is all copied code nvm

pkg/tbtc/marshaling.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+	"github.com/keep-network/keep-core/pkg/chain"
+	"github.com/keep-network/keep-core/pkg/protocol/group"
+	"github.com/keep-network/keep-core/pkg/tbtc/gen/pb"
+	"github.com/keep-network/keep-core/pkg/tecdsa"
+)
+
+// Marshal converts the wallet to a byte array.
+func (w *wallet) Marshal() ([]byte, error) {
+	publicKey := elliptic.Marshal(
```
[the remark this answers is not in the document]
beaushinkle: that also means there's a gap in test coverage!

beaushinkle:
`pkg/tbtc/marshaling_test.go` has good tests for the happy path. Do you think it's worth adding unhappy path / chaotic path tests? We could try to (un)marshall some arbitrary public keys, some other curves, and wallets with missing data.

cmd/start.go
```diff
@@ -90,20 +90,21 @@ func start(cmd *cobra.Command) error {
...
 	nodeHeader(netProvider.ConnectionManager().AddrStrings(), clientConfig.LibP2P.Port)
 
-	handle, err := persistence.NewDiskHandle(clientConfig.Storage.DataDir)
+	beaconPersistence, err := initializePersistence(clientConfig, beacon.ProtocolName)
 	if err != nil {
-		return fmt.Errorf("failed while creating a storage disk handler: [%v]", err)
+		return fmt.Errorf("cannot initialize beacon persistence: [%w]", err)
+	}
+
+	tbtcPersistence, err := initializePersistence(clientConfig, tbtc.ProtocolName)
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 90cd898f

pkg/internal/tecdsatest/tecdsatest.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+const (
+	testDataDirFormat                 = "%s/testdata"
+	privateKeyShareTestDataFileFormat = "private_key_share_data_%d.json"
+)
+
+// LoadPrivateKeyShareTestFixtures loads tECDSA private key share test data.
+// Code copied from:
+//   https://github.com/binance-chain/tss-lib/blob/master/ecdsa/keygen/test_utils.go
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 3d181a3f

pkg/internal/tecdsatest/tecdsatest.go
```diff
@@ -0,0 +1,69 @@
...
+// Test data JSON files copied from:
+//   https://github.com/binance-chain/tss-lib/tree/master/test/_ecdsa_fixtures
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 3d181a3f

pkg/tbtc/wallet.go
```diff
@@ -0,0 +1,61 @@
...
+	// element with index 0 has the group.MemberIndex equal to 1 and so on).
+	signingGroupOperators []chain.Address
+}
+
+// signer represents a threshold signer of a tBTC wallet. A signer holds
+// a wallet tECDSA private key share and is able to participate in the
+// signing process.
+type signer struct {
+	// wallet points to the tBTC wallet this signer belongs to.
+	wallet *wallet
```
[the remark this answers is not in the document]
lukasz-zimnoch: Hmm, yeah. A pointer can wrongly indicate that the `wallet` field is optional. See e12613f5

pkg/tbtc/gen/pb/wallet.proto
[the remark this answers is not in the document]
lukasz-zimnoch: > We never marshal wallet separately, it's always a part of the signer.

That was not obvious at the beginning but is clear now. See e3791a7a

pkg/tbtc/marshaling.go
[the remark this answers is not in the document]
lukasz-zimnoch: See d4c0af96

pkg/tbtc/node.go
```diff
@@ -160,12 +163,38 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
 				logger.Infof(
-					"[member:%v] generated group [0x%x]",
+					"[member:%v] generated wallet's signing group [0x%x]",
 					memberIndex,
-					result.GroupPublicKey,
+					elliptic.Marshal(
+						signingGroupPublicKey.Curve,
+						signingGroupPublicKey.X,
+						signingGroupPublicKey.Y,
+					),
```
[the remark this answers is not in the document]
lukasz-zimnoch: Nope, this is completely unrelated to the storage. Just a log that indicates the DKG was done successfully. Will be probably changed once we have the result submission implemented.

pkg/tbtc/node.go
```diff
@@ -160,12 +163,38 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
 
 				// TODO: Submit the result using the chain layer.
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yep, this is clear. Left a TODO about that: 82bd7ade

pkg/tbtc/node.go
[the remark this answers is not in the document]
pdyraga: If we are going to change it, then cool. What I am trying to say is that when logging something we always need to think about how valuable this information is. I believe that in v1 of the beacon we had a group public key logged in a form that has neither reflected directory on disk nor how the key was stored on-chain.

pkg/tbtc/storage.go
lukasz-zimnoch: See 39a977d3

pkg/tbtc/node.go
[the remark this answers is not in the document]
lukasz-zimnoch: Improved that within 39a977d3 by the way.

pkg/tbtc/storage_test.go
lukasz-zimnoch: Done during 39a977d3

pkg/tecdsa/marshaling.go
lukasz-zimnoch: See 659d9b61

pkg/tecdsa/dkg/protocol_test.go
```diff
@@ -987,16 +988,21 @@ func TestTssFinalize(t *testing.T) {
...
+			groupPublicKey.Y,
+		)
 
 		testutils.AssertIntsEqual(
 			t,
 			"length of the group public key",
 			65,
-			len(groupPublicKey),
+			len(groupPublicKeyBytes),
 		)
```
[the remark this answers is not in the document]
lukasz-zimnoch: See f3a2e8a3

lukasz-zimnoch:
> `pkg/tbtc/marshaling_test.go` has good tests for the happy path. Do you think it's worth adding unhappy path / chaotic path tests? We could try to (un)marshall some arbitrary public keys, some other curves, and wallets with missing data.

Honestly, I'm not sure if that makes sense. We do not want to marshal at all if some trash is produced by the TSS lib. We should raise an error instead. Also, we are not fuzzing here since those types are local and not broadcasted through the network like message types.

pkg/tecdsa/marshaling.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/binance-chain/tss-lib/crypto/paillier"
+	"github.com/binance-chain/tss-lib/ecdsa/keygen"
+	"github.com/keep-network/keep-core/pkg/tecdsa/gen/pb"
+	"math/big"
+)
+
+// ErrIncompatiblePublicKey indicates that the given public key is not
+// compatible with the tECDSA, i.e. uses a different elliptic curve.
+// Such a key cannot be marshalled and unmarshalled.
+var ErrIncompatiblePublicKey = fmt.Errorf(
```
pdyraga, line 15: Do we need to export it out of `tecdsa`?

lukasz-zimnoch: My reasoning: `tecdsa.PrivateKeyShare` is a public type so as their `Marshal` function. If that function returns a pre-defined error, it would be good to expose that error too.

pdyraga: I missed that `tecdsa.PrivateKeyShare` is marshalled from `tbtc` package. It all makes sense!

pkg/tbtc/registry.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// loadSigners loads all signers stored using the underlying persistence layer.
+func (ws *walletStorage) loadSigners() map[string][]*signer {
+	signersByWallet := make(map[string][]*signer)
+
+	descriptorsChan, errorsChan := ws.persistence.ReadAll()
+
+	var wg sync.WaitGroup
+	wg.Add(2)
```
pdyraga, line 128: Should we explain the logic behind two goroutines and the wait group?
```suggestion
	// Two goroutines read from descriptors and errors channels and either
	// adds signer to the cache or outputs an error to stderr.
	// The reason for using two goroutines at the same time - one for
	// descriptors and one for errors - is that channels do not have to be
	// buffered and we do not know in what order information is written to
	// channels.
	var wg sync.WaitGroup
	wg.Add(2)
```

pkg/tbtc/registry.go
```diff
@@ -0,0 +1,195 @@
...
+	persistence persistence.Handle
+}
+
+// newWalletStorage creates a new instance of the walletStorage.
+func newWalletStorage(persistence persistence.Handle) *walletStorage {
+	return &walletStorage{persistence}
+}
+
+// saveSigner saves the given signer using the underlying persistence layer
+// of the walletStorage.
```
pdyraga, line 98: Nitpicking but I think it would be worth noting in the documentation this function is not adding signer to the cache and should not be called from the outside of the struct.

pkg/tbtc/registry.go
```diff
@@ -0,0 +1,195 @@
...
+			"could not save membership using the "+
+				"underlying persistence layer: [%w]",
+			err,
+		)
+	}
+
+	return nil
+}
+
+// loadSigners loads all signers stored using the underlying persistence layer.
```
pdyraga, line 121: I would also drop a comment that this function is an internal one and should not be called from outside of the structure.

pdyraga:
The code looks good to me, it's testing time! I will hold on with the merge until @beaushinkle also gives a green light.

pkg/tbtc/registry.go
lukasz-zimnoch: See 696d8753

pkg/tbtc/registry.go
lukasz-zimnoch: See 717040f1

pkg/tbtc/registry.go
lukasz-zimnoch: See 717040f1

beaushinkle:
> We do not want to marshal at all if some trash is produced by the TSS lib. We should raise an error instead.

That's mostly what I'm getting at: right now, we have this:
```
	wallet, err := s.wallet.Marshal()
	if err != nil {
		return nil, fmt.Errorf("cannot marshal wallet: [%w]", err)
	}

	privateKeyShare, err := s.privateKeyShare.Marshal()
	if err != nil {
		return nil, fmt.Errorf("cannot marshal private key share: [%w]", err)
	}
```
but both of those codepaths are untested in `marshaling_test`. We know we should return an error if the underlying wallet/privatekey has trash data, but that isn't asserted (at least in our codebase). Ideally, the wallet marshal test should check to make sure it's properly spitting out errors when handed garbage, the privateKeyShare marshaling test should check to make sure it's properly spitting out errors when handed garbage, and the signer (which calls both), should also check both.

I'm not saying it has to block this PR, or even that we should do it at all! Just trying to get a handle on the level of code coverage we're going for, both for my own code that I write and also what's useful for me to do review wise

beaushinkle:
I'm good to go as well! The context that a lot of this code is directly ported from already-reviewed and battled-tested v1 code helps explain a lot :brain:

I have some open questions above around code coverage, but nothing blocking, so let's ship it 🚀

beaushinkle (approved):

pdyraga:
When testing the storage I noticed a problem with keep client failing to start if the storage directory has no `beacon` and `tbtc` directories created:
```
2022-08-11T14:49:54.291+0200	FATAL	keep-start	cmd/start.go:46	cannot initialize beacon persistence: [cannot create [beacon] disk handle: [cannot read from the storage directory: [open /Users/piotr/dev/keep-storage/3/beacon: no such file or directory]]]
```

```
2022-08-11T14:50:21.489+0200	FATAL	keep-start	cmd/start.go:46	cannot initialize tbtc persistence: [cannot create [tbtc] disk handle: [cannot read from the storage directory: [open /Users/piotr/dev/keep-storage/3/tbtc: no such file or directory]]]
```

I think the client should create `beacon` and `tbtc` directories if not present. The operator provides the location of the storage and expects the client to prepare and organize the storage.

PR #3131, opened 2022-08-05
Remove openzeppelin manifest for fresh installation

dimpar:
It happens that open zeppelin's manifest in ecdsa (.openzeppelin) is not updated properly and the error is thrown during deployment of the `WalletRegistry` as transparent proxy. Removing this folder forces a fresh installation.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -11,7 +11,7 @@ KEEP_CORE_PATH=$PWD
 KEEP_BEACON_SOL_PATH="$KEEP_CORE_PATH/solidity/random-beacon"
 KEEP_ECDSA_SOL_PATH="$KEEP_CORE_PATH/solidity/ecdsa"
 TMP="$KEEP_CORE_PATH/tmp"
-OPENZEPPELIN_MANIFEST=".openzeppelin/*-1101.json"
+OPENZEPPELIN_MANIFEST=".openzeppelin/unknown-*.json"
 
 # Defaults, can be overwritten by env variables/input parameters
 NETWORK_DEFAULT="development"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -11,7 +11,7 @@ KEEP_CORE_PATH=$PWD
 KEEP_BEACON_SOL_PATH="$KEEP_CORE_PATH/solidity/random-beacon"
 KEEP_ECDSA_SOL_PATH="$KEEP_CORE_PATH/solidity/ecdsa"
 TMP="$KEEP_CORE_PATH/tmp"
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -11,6 +11,7 @@ KEEP_CORE_PATH=$PWD
 KEEP_BEACON_SOL_PATH="$KEEP_CORE_PATH/solidity/random-beacon"
 KEEP_ECDSA_SOL_PATH="$KEEP_CORE_PATH/solidity/ecdsa"
 TMP="$KEEP_CORE_PATH/tmp"
+OPENZEPPELIN_MANIFEST=".openzeppelin"
```
nkuba, line 14: Maybe we could remove just the file for the current network?

```suggestion
OPENZEPPELIN_MANIFEST=".openzeppelin/$NETWORK.json"
```

nkuba (changes requested):

scripts/install.sh
dimpar: $NETWORK is "development", but the file that is created on my local is called `unknown-1101.json` so it won't match. I'll see if there's a config param somewhere where we can specify a manifest file name. I'd rather not parse chain id in bash.

dimpar: I think the easiest way of removing that specific file is to take advantage of the chain id which is hardcoded in other config files for the development env. [2ea98e3](https://github.com/keep-network/keep-core/pull/3131/commits/2ea98e3c019ef9689b6a3deb494254a0702f5a78)

scripts/install.sh
```diff
@@ -11,6 +11,7 @@ KEEP_CORE_PATH=$PWD
 KEEP_BEACON_SOL_PATH="$KEEP_CORE_PATH/solidity/random-beacon"
 KEEP_ECDSA_SOL_PATH="$KEEP_CORE_PATH/solidity/ecdsa"
 TMP="$KEEP_CORE_PATH/tmp"
+OPENZEPPELIN_MANIFEST=".openzeppelin/*-1101.json"
```
nkuba, line 14: Maybe we should do `unknown-*.json` https://docs.openzeppelin.com/upgrades-plugins/1.x/network-files
```suggestion
OPENZEPPELIN_MANIFEST=".openzeppelin/unknown-*.json"
```

dimpar: ok, I think we can do it. Basically, I wanted to remove the 'development' manifest, but I think we can fresh install all the local chains. [61bccab](https://github.com/keep-network/keep-core/pull/3131/commits/61bccabd8663dc2a0af3a1a2166bc0bd7741ec78)

nkuba (approved):

PR #3132, opened 2022-08-05
Docker image for contracts hardhat initialization tasks

nkuba:
In this PR we build and publish a docker image on contracts deployment. The image will expose a hardhat handle to run tasks (i.e. initialization scripts) for the contracts deployed for a given test network.

We use [resuable-workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) to add a common workflow that can be used by all scripts for docker image building and publication.


.github/workflows/common-docker.yml
```diff
@@ -0,0 +1,107 @@
+name: Docker
```
michalinacienciala, line 1: I wonder if it wouldn't be better to store the reusable workflows in a separate repository, for example in the https://github.com/keep-network/ci repo or in some new dedicated repo (like `keep-network/reusable-workflows`). We may want to use such workflows outside of `keep-network/keep-core` in the future. I know keeping them in `keep-core` repo doesn't prevent other repos from using them, but keeping them in dedicated place would seem more organized for me.

.github/workflows/common-docker.yml
```diff
@@ -0,0 +1,107 @@
...
+        type: boolean
+      image-name:
+        description: Docker image name
+        required: true
+        type: string
+      image-push:
+        required: true
+        default: false
+        type: boolean
+      image-context:
```
michalinacienciala, line 29: Could we add `description` here? So that it would be clear that context refers to the path to the folder with the files that should be built

.github/workflows/common-docker.yml
```diff
@@ -0,0 +1,107 @@
...
+        run: |
+          echo "IMAGE_NAME=${{ inputs.image-name }}" >> $GITHUB_ENV
+
+      - name: Set Docker image name for GCR
+        if: inputs.image-push == true
+        run: |
+          echo "IMAGE_NAME=${{ env.GCR_REGISTRY_URL }}/${{ env.GOOGLE_PROJECT_ID }}/${{ inputs.image-name }}" >> $GITHUB_ENV
+
+      - name: Build and publish image
+        uses: docker/build-push-action@v2
```
michalinacienciala, line 96: There's a `v3` already available, could we use it?

.github/workflows/common-docker.yml
```diff
@@ -0,0 +1,107 @@
...
+          context: ${{ inputs.image-context }}
+          # GCR image should be named according to following convention:
+          # HOSTNAME/PROJECT-ID/IMAGE:TAG
+          # We don't use TAG yet, will be added at later stages of work on RFC-18.
+          tags: ${{ env.IMAGE_NAME }}
+          labels: |
+            revision=${{ github.sha }}
+          push: ${{ inputs.image-push }}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new
```
michalinacienciala, line 107: In older jobs in which we were using `docker/build-push-action@v2` action we needed to have following step executed afterwards to prevent problems with job size reaching a limit:
```
      - # Temp fix - move cache instead of copying (added below step and
        # modified value of `cache-to`).
        # https://github.com/docker/build-push-action/issues/252
        # https://github.com/moby/buildkit/issues/1896
        # Without the change some jobs were failing with `no space left on device`
        name: Move cache
        run: |
          rm -rf /tmp/.buildx-relay-cache
          mv /tmp/.buildx-relay-cache-new /tmp/.buildx-relay-cache
```
I see that the referenced issues are still open. Maybe it would be safer to add this step here, even if right now the workflows using this reusable workflow don't throw the errors?

.github/workflows/common-docker.yml
```diff
@@ -0,0 +1,107 @@
...
+        required: false
+
+jobs:
+  docker-build-and-publish:
+    name: Build and publish Docker image
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/download-artifact@v3
```
michalinacienciala, line 44: Optional suggestion:
```suggestion
      - name: Download uncompressed artifacts
      - uses: actions/download-artifact@v3
```

.github/workflows/common-docker.yml
```diff
@@ -0,0 +1,107 @@
...
+        if: inputs.download-artifact-tar != true
+        with:
+          name: ${{ inputs.download-artifact-name }}
+          path: ${{ inputs.download-artifact-path }}
+
+      - uses: actions/download-artifact@v3
```
michalinacienciala, line 50: Optional suggestion:
```suggestion
      - name: Download compressed artifacts
      - uses: actions/download-artifact@v3
```

.github/workflows/common-docker.yml
nkuba: We should extract it at some point. But for now, I prefer keeping it here to test and develop it easier. Once we confirm it works as expected we can extract it.

.github/actions/docker-build-push/action.yml
```diff
@@ -0,0 +1,86 @@
...
+    - name: Cache Docker layers
+      uses: actions/cache@v3
+      with:
+        path: /tmp/.buildx-cache
+        key: ${{ runner.os }}-buildx-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-buildx-
+
+    - name: Load environment variables
+      if: env.PUSH == 'true'
```
michalinacienciala, line 38: How does the value of `PUSH` environment variable get set? Same question for other variables like `ENVIRONMENT`, `GCR_JSON_KEY`, `CONTEXT`?

nkuba: It was set automatically with inputs. But you are right, it's not clear, I updated it with f1d3a6a4f002b8a2301ec47eefdd3d31b027deac

michalinacienciala (approved):

PR #3133, opened 2022-08-05
Added extra gas requirement for ECDSA DKG result challenge

pdyraga:
Due to EIP150, 1/64 of the gas is not forwarded to the call, and will be kept to execute the remaining operations in the function after the call inside the try-catch.

To ensure there is no way for the caller to manipulate the gas limit in such a way that the call inside try-catch fails with out-of-gas and the rest of the function is executed with the remaining 1/64 of gas, we require an extra gas amount to be left at the end of the call to `WalletRegistry.challengeDkgResult`.

In practice, with the current version of the code, such a scenario should not be possible anyway but the extra gas requirement provides an extra layer of protection in case of any code changes. In the current version of the code, the cost of `dkg.challegeResult(dkgResult)` is about 1.6M gas units and the cost of everything that happens later is around 144k gas units. It means that the cost of the rest of the logic is about 6% of the cost of the entire function which is more than 1/64.

Another option for solving this problem was trying to eliminate try-catch altogether. This option would be more error-prone for future changes in the code though. We would need to check all functions on the execution stack of the DKG result challenge and ensure they can not revert or if they revert, they revert with a specific error. Any future change to the code would require inspecting the possible cases of reverts again. That's why a general try-catch with an extra gas requirement at the end feels safer long-term.

Note that the functions on the execution stack of the DKG result challenge may revert. One example is the corrupted signature verification covered by `with dkg result submitted with unrecoverable signatures` unit tests. That's why some try-catch is necessary, even if not on the highest level, then more try-catches on the lower levels of the execution stack.

### DIFF
--- a/solidity/ecdsa/contracts/WalletRegistryGovernance.sol
+++ b/solidity/ecdsa/contracts/WalletRegistryGovernance.sol
@@ -977,7 +977,7 @@ contract WalletRegistryGovernance is Ownable {
         newDkgResultChallengePeriodLength = 0;
     }
 
-    /// @notice Begins the DKG result challenge extra gas update process.    .
+    /// @notice Begins the DKG result challenge extra gas update process.
     /// @dev Can be called only by the contract owner.
     /// @param _newDkgResultChallengeExtraGas New DKG result challenge extra gas
     function beginDkgResultChallengeExtraGasUpdate(
--- a/solidity/ecdsa/contracts/WalletRegistryGovernance.sol
+++ b/solidity/ecdsa/contracts/WalletRegistryGovernance.sol
@@ -59,14 +59,14 @@ contract WalletRegistryGovernance is Ownable {
     uint256 public newDkgResultChallengePeriodLength;
     uint256 public dkgResultChallengePeriodLengthChangeInitiated;
 
+    uint256 public newDkgResultChallengeExtraGas;
+    uint256 public dkgResultChallengeExtraGasChangeInitiated;
+
     uint256 public newDkgResultSubmissionTimeout;
     uint256 public dkgResultSubmissionTimeoutChangeInitiated;
 
     uint256 public newSubmitterPrecedencePeriodLength;
     uint256 public dkgSubmitterPrecedencePeriodLengthChangeInitiated;
-
-    uint256 public newDkgResultChallengeExtraGas;
-    uint256 public dkgResultChallengeExtraGasChangeInitiated;
 
     uint256 public newDkgResultSubmissionGas;
     uint256 public dkgResultSubmissionGasChangeInitiated;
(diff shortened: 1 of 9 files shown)


solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -1048,12 +1062,50 @@ contract WalletRegistryGovernance is Ownable {
...
             walletRegistry.dkgParameters().resultSubmissionTimeout,
-            newSubmitterPrecedencePeriodLength
+            newSubmitterPrecedencePeriodLength,
+            walletRegistry.dkgParameters().resultChallengeExtraGas
         );
         dkgSubmitterPrecedencePeriodLengthChangeInitiated = 0;
         newSubmitterPrecedencePeriodLength = 0;
     }
 
+    /// @notice Begins the DKG result challenge extra gas update process.    .
```
tomaszslabon, line 1072: Unnecessary spaces and a dot at the end of the line :)

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
```diff
@@ -1048,12 +1062,50 @@ contract WalletRegistryGovernance is Ownable {
...
+    /// @dev Can be called only by the contract owner.
+    /// @param _newDkgResultChallengeExtraGas New DKG result challenge extra gas
```
tomaszslabon, line 1074: The parameter description should end with a dot.

tomaszslabon:
@pdyraga Shouldn't we add the same gas-left check for the Random Beacon's `challengeDKGResult` function?

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
pdyraga: Yeah. We were really inconsistent about it in `WalletRegistry` 😬 

Fixed in f337b2f7ae864a91261bb8fdcc957d22a2b12f58.

solidity/ecdsa/contracts/WalletRegistryGovernance.sol
pdyraga: f337b2f7ae864a91261bb8fdcc957d22a2b12f58

pdyraga:
> @pdyraga Shouldn't we add the same gas-left check for the Random Beacon's `challengeDKGResult` function?

Correct! I am working on it right now. I wanted to make this change in a separate PR.

tomaszslabon:
@pdyraga Additionally, we have some dummy Wallet Registry contracts used for testing named `WalletRegistryV2`, `WalletRegistryV2MisplacedNewSlot`, `WalletRegistryV2MissingSlot`. Should we aligned them with the real contract?

pdyraga:
> Additionally, we have some dummy Wallet Registry contracts used for testing named `WalletRegistryV2`, `WalletRegistryV2MisplacedNewSlot`, `WalletRegistryV2MissingSlot`. Should we aligned them with the real contract?

Good catch. Done in 8b9f9763846e0a8e892b2536bdfb44c3fe8d059b.

tomaszslabon (approved):

PR #3136, opened 2022-08-05
Use newer version (v3) of the `setup-node` action

michalinacienciala:
New version of the action is now available. With `v2` we're sometimes
seeing problems with parallel execution of jobs. We hope that updating
to `v3` will help.


nkuba:
We could also update the checkout action to v3

nkuba (approved):

PR #3137, opened 2022-08-08
Added extra gas requirement for Beacon DKG result challenge

pdyraga:
See #3133 for ECDSA part.

Due to EIP150, 1/64 of the gas is not forwarded to the call, and will be kept to execute the remaining operations in the function after the call inside the try-catch.

To ensure there is no way for the caller to manipulate the gas limit in such a way that the call inside try-catch fails with out-of-gas and the rest of the function is executed with the remaining 1/64 of gas, we require an extra gas amount to be left at the end of the call to `RandomBeacon.challengeDkgResult`.

In practice, with the current version of the code, such a scenario should not be possible anyway but the extra gas requirement provides an extra layer of protection in case of any code changes. In the current version of the code, the cost of `dkg.challegeResult(dkgResult)` is about 1M gas units and the cost of everything that happens later is around 118k gas units. It means that the cost of the rest of the logic is about 10% of the cost of the entire function which is more than 1/64.

Another option for solving this problem was trying to eliminate try-catch altogether. This option would be more error-prone for future changes in the code though. We would need to check all functions on the execution stack of the DKG result challenge and ensure they can not revert or if they revert, they revert with a specific error. Any future change to the code would require inspecting the possible cases of reverts again. That's why a general try-catch with an extra gas requirement at the end feels safer long-term.

Note that the functions on the execution stack of the DKG result challenge may revert. One example is the corrupted signature verification covered by `with dkg result submitted with unrecoverable signatures` unit tests. That's why some try catch is necessary, even if not on the highest level, then more try-catches on the lower levels of the execution stack.

An unpleasant surprise when working on this PR was the `RandomBeacon` contract size. After adding the extra gas logic, the maximum contract size got exceeded and I had to do two sacrifices to take it back under the limit. The first change is making `genesisSeed` internal and the second one is shortening the revert message in `notifyOperatorInactivity`. With these two changes, the contract size is at 24.576, so exactly at the limit.

### DIFF
--- a/solidity/random-beacon/contracts/RandomBeaconGovernance.sol
+++ b/solidity/random-beacon/contracts/RandomBeaconGovernance.sol
@@ -46,6 +46,9 @@ contract RandomBeaconGovernance is Ownable {
     uint256 public newDkgResultChallengePeriodLength;
     uint256 public dkgResultChallengePeriodLengthChangeInitiated;
 
+    uint256 public newDkgResultChallengeExtraGas;
+    uint256 public dkgResultChallengeExtraGasChangeInitiated;
+
     uint256 public newDkgResultSubmissionTimeout;
     uint256 public dkgResultSubmissionTimeoutChangeInitiated;
 
@@ -146,6 +149,12 @@ contract RandomBeaconGovernance is Ownable {
     event DkgResultChallengePeriodLengthUpdated(
         uint256 dkgResultChallengePeriodLength
     );
+
+    event DkgResultChallengeExtraGasUpdateStarted(
+        uint256 dkgResultChallengeExtraGas,
+        uint256 timestamp
+    );
+    event DkgResultChallengeExtraGasUpdated(uint256 dkgResultChallengeExtraGas);
 
     event DkgResultSubmissionTimeoutUpdateStarted(
         uint256 dkgResultSubmissionTimeout,
@@ -515,6 +524,7 @@ contract RandomBeaconGovernance is Ownable {
             ,
             uint256 groupLifetime,
             uint256 dkgResultChallengePeriodLength,
+            uint256 dkgResultChallengeExtraGas,
             uint256 dkgResultSubmissionTimeout,
             uint256 dkgSubmitterPrecedencePeriodLength
         ) = randomBeacon.groupCreationParameters();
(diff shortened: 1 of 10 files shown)


solidity/random-beacon/contracts/RandomBeaconGovernance.sol
```diff
@@ -621,13 +635,60 @@ contract RandomBeaconGovernance is Ownable {
...
             dkgResultSubmissionTimeout,
             dkgSubmitterPrecedencePeriodLength
         );
         dkgResultChallengePeriodLengthChangeInitiated = 0;
         newDkgResultChallengePeriodLength = 0;
     }
 
+    /// @notice Begins the DKG result challenge extra gas update process.
+    /// @dev Can be called only by the contract owner.
+    /// @param _newDkgResultChallengeExtraGas New DKG result challenge extra gas
```
tomaszslabon, line 648: Missing dot at the end of parameter description.

solidity/random-beacon/test/RandomBeacon.Parameters.test.ts
```diff
@@ -216,6 +218,7 @@ describe("RandomBeacon - Parameters", () => {
             newGroupCreationFrequency,
```
tomaszslabon, line 218: In the `RandomBeacon.Parameters.test.ts` file, in the "updateGroupCreationParameters" describe, in the "when the caller is the governance", we are missing test checking for update of `dkgResultChallengeExtraGas`.
So we need to add 
```
it("should update the DKG result challenge extra gas", async () => {
...
})
```

pdyraga: 0973e854597b3195a60b4303f2378378784e2692

solidity/random-beacon/contracts/RandomBeaconGovernance.sol
pdyraga: The problem with `RandomBeacon`, `WalletRegistry`, and libraries is that they were never updated to use the new convention. Most of the functions use parameter description with no dots, some others use dots (minority). If we want to fix it, then I would prefer to do it in a separate PR.

tomaszslabon (approved):

PR #3141, opened 2022-08-08
Configure Tenderly in CI

nkuba:
Action can be reused by workflows to configure the tenderly token.

TODO:
- [ ] Test if it works


.github/workflows/contracts-ecdsa.yml
```diff
@@ -198,10 +198,10 @@ jobs:
             @keep-network/sortition-pools
 
       - name: Configure tenderly
-        env:
-          TENDERLY_TOKEN: ${{ secrets.TENDERLY_TOKEN }}
-        run: ./config_tenderly.sh
-      
+        uses: ./.github/actions/docker-build-push
```
michalinacienciala, line 201:
```suggestion
        uses: ./.github/actions/tenderly-config
```

lionakhnazarov:
Closing as stale. Tenderly CI configuration here is years old and conflicts with the current CI on main. Re-cut if still needed.

PR #3151, opened 2022-08-10
Adding a warning informing about a min number of required accounts

dimpar:
Hardhat projects use named accounts that have assigned the default
indices to them. The index denotes an index of an account. A minimum number of
all accounts is the highest index of the named account across all the used
projects.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -86,7 +86,7 @@ printf "${LOG_START}Starting installation...${LOG_END}"
 
 printf "Network: $NETWORK\n"
 
-printf "${LOG_WARNING_START}Make sure you have at least ${RECOMMENDED_ACCOUNTS_NUMBER} accounts${LOG_WARNING_END}"
+printf "${LOG_WARNING_START}Make sure you have at least ${RECOMMENDED_ACCOUNTS_NUMBER} ethereum accounts${LOG_WARNING_END}"
 
 cd $KEEP_BEACON_SOL_PATH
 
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -80,6 +86,8 @@ printf "${LOG_START}Starting installation...${LOG_END}"
 
 printf "Network: $NETWORK\n"
 
+printf "${LOG_WARNING_START}Make sure you have at least ${RECOMMENDED_ACCOUNTS_NUMBER} accounts${LOG_WARNING_END}"
```
tomaszslabon, line 89: During unlocking 
Maybe we could be more precise and say:
"Make sure you have at least ${RECOMMENDED_ACCOUNTS_NUMBER} ethereum accounts"
?
We later (during unlocking) call them "ethereum accounts".

dimpar: Sure [abcb8a2](https://github.com/keep-network/keep-core/pull/3151/commits/abcb8a2f2e1d044fb2a35360c5c808909c087cc0)

tomaszslabon (approved):

scripts/install.sh
```diff
@@ -1,17 +1,23 @@
...
 KEEP_CORE_PATH=$PWD
 
 KEEP_BEACON_SOL_PATH="$KEEP_CORE_PATH/solidity/random-beacon"
 KEEP_ECDSA_SOL_PATH="$KEEP_CORE_PATH/solidity/ecdsa"
 TMP="$KEEP_CORE_PATH/tmp"
 OPENZEPPELIN_MANIFEST=".openzeppelin/unknown-*.json"
+# This number should be no less than the highest index assigned to a named account
+# specified in `hardhat.config.ts` configs across all the used projects. Note that
+# account indices start from 0.
+RECOMMENDED_ACCOUNTS_NUMBER=8
```
pdyraga, line 20: Is it recommended or required?

pdyraga (approved):

PR #3158, opened 2022-08-11
Add initcontainer for funding eth address

nkuba:
ETH address of the client has to be funded with ETH before running next
initcontainers.

TODO: 
- [ ] regenerate config
- [ ] add `purse-eth-account-private-key` secret
- [ ] apply to keep-test

Depends on https://github.com/keep-network/keep-core/pull/3156
Refs https://github.com/keep-network/keep-core/pull/3128


lionakhnazarov:
Closing as stale. This funding init-container draft predates current deployment tooling and no longer applies to main.

PR #3160, opened 2022-08-11
Allow configuring tECDSA parameters for the client

pdyraga:
It is now possible to configure three parameters of tECDSA protocol:
- pre-params pool size
- pre-params generation timeout
- pre-params generation concurrency

The parameters can be configured via config file or via command-line flags.

Worth noting, that in tss-lib, concurrency level < 3 translates to a concurrency level equal to 1:

From [v1.3.3 of `tss-lib/ecdsa/keygen/prepare.go`](https://github.com/bnb-chain/tss-lib/blob/v1.3.3/ecdsa/keygen/prepare.go#L41-L43):
```
if concurrency /= 3; concurrency < 1 {
    concurrency = 1
}
```

This is because there are three concurrent generation threads triggered in `prepare.go`.


pkg/tecdsa/tecdsa.go
```diff
@@ -0,0 +1,19 @@
+package tecdsa
+
+import "time"
+
+const (
+	DefaultPreParamsPoolSize              = 50
```
nkuba, line 6: That's a lot. Could we reduce the value? How often do we expect a pre param will be popped from the pool?
Generation is very CPU consuming and generating so many numbers may take a long time and cause a headache on production machines.

pkg/tecdsa/dkg/params_pool.go
```diff
@@ -21,11 +22,20 @@ func newTssPreParamsPool(
 	logger log.StandardLogger,
 	poolSize int,
 	generationTimeout time.Duration,
+	generationConcurrency int,
 ) *tssPreParamsPool {
-	logger.Infof("TSS pre-parameters target pool size is [%v]", poolSize)
+	logger.Infof(
+		"TSS pre-parameters target pool size is [%v], generation timeout is [%v], and concurrency level is [%v]",
```
nkuba, line 28:
```suggestion
		"TSS pre-parameters target pool size is [%d], generation timeout is [%s], and concurrency level is [%d]",
```

pkg/tecdsa/tecdsa.go
```diff
@@ -0,0 +1,19 @@
...
+	DefaultPreParamsGenerationTimeout     = 2 * time.Minute
+	DefaultPreParamsGenerationConcurrency = 1
+)
+
+// Config carries the config for threshold ECDSA protocol.
+type Config struct {
```
lukasz-zimnoch, line 12: Are we sure this type should be declared here? Pre-params pool and its config are something DKG-specific and declaring that type here makes `pkg/tecdsa` package silently dependent on `pkg/tecdsa/dkg` while the proper direction is something opposite (`pkg/tecdsa/dkg` depends on `pkg/tecdsa` by using the `PrivateKeyShare` type). I see that `dkg.NewExecutor` now exposes those params directly so there is probably no point to move `Config` there. Looks like we need it just for convenience when we are parsing config flags/file so maybe we should put `Config` directly into `pkg/tbtc` and expect it as a parameter of `tbtc.Initialize`?

pkg/tecdsa/tecdsa.go
pdyraga: This is an absolute minimum to run two clients locally. The group has 100 members and assuming even distribution (it is usually not even), we need to have 50 members per node.

This PR is just about allowing to configure these values and I used the same value as it was before (see `pkg/tbtc/node.go`). In a separate set of PRs, we'll have to rework the pre-params generation process to generate way more pre-params but only when none of the protocols are currently executing.

pkg/tecdsa/tecdsa.go
pdyraga: I placed the config in `pkg/tecdsa` to not create a dependency from `pkg/tecdsa` to `pkg/tecdsa/dkg`. If we had this config in `pkg/tecdsa/dkg`, then we'd need to accept it as a parameter of `tbtc.Initialize` and we'd need to add an import from `pkg/tecdsa` to `pkg/tecdsa/dkg`.

Putting aside the direction of dependencies, from the protocol perspective, this structure fits best in `pkg/tecdsa/dkg` because it contains only DKG-specific parameters. `pkg/tecdsa` is not that bad because we are still in `tecdsa` land, we may add more parameters to this config, and we'll have to make this code interact with the rest of the protocols executed by the client to pause pre-params generation (separate set of PRs) so `pkg/tecdsa/dkg` is too deep.

I can put this struct in `tbtc` but then `dkg.NewExecutor` needs to accept params directly (not a struct). To me both options are fine. The way I look at it is that these are `tecdsa` parameters so we are not creating a silent dependency from `pkg/tecdsa` to `pkg/tecdsa/dkg`. We are rather having `pkg/tecdsa/dkg` accepting general tECDSA parameters. But I am happy to change it. What do you think?

pkg/tecdsa/dkg/params_pool.go
pdyraga: https://github.com/keep-network/keep-core/pull/3160/commits/b518511a79fb2235511bc53ca9fffc1105d0e44e

pkg/tecdsa/tecdsa.go
lukasz-zimnoch: >  The way I look at it is that these are tecdsa parameters so we are not creating a silent dependency from pkg/tecdsa to pkg/tecdsa/dkg. We are rather having pkg/tecdsa/dkg accepting general tECDSA parameters.

Regarding the "silent dependency" I mean the following: if we have things like `PreParamsPoolSize` right in `pkg/tecdsa` that means this package is aware that there is something like a pre params pool in the DKG implementation. That information leaks from `pkg/tecdsa/dkg` and although it doesn't look as something dangerous now, it can lead to something weird in the future. If this was about something really general for tECDSA like, for example `threshold` or so, everything would be fine. I think the `pkg/tecdsa` package should remain a high-level one declaring only general concepts that are common for DKG and signing.

> I can put this struct in tbtc but then dkg.NewExecutor needs to accept params directly (not a struct).

I think this is fine. The `pkg/tbtc` is the application and it should be responsible for its own config (i.e. should declare it). Since it uses `pkg/tecdsa/dkg` directly, the knowledge about pre-params pool in `pkg/tbtc` seems to be justified.

pdyraga: https://github.com/keep-network/keep-core/pull/3160/commits/6a805056561d1c1c85c62a3c8a3d92da62a63498 sir.

lukasz-zimnoch (approved):
No more comments on my end! Leaving it for @nkuba

nkuba (approved):

PR #3148, opened 2022-08-09
Start a client

dimpar:
This PR uses changes made in https://github.com/keep-network/keep-core/pull/3129 to use the locally deployed artifacts. Now `install.sh` will build the client with the local artifacts. Also, `start.sh` can be used now to start a client.

To start a client we need to:

Manually configure `config.toml` in the `/configs` dir. Providing contract addresses is no longer required for "development" env. Next:
- run `scripts/install.sh`
- run `scripts/initialize.sh --stake-owner <address>`
- run `scripts/start.sh`

Available params for `start.sh`
- `KEEP_ETHEREUM_PASSWORD` env var. If not provided, "password" is used
- flag: `--config-dir` path to the config file. `/configs` is default

Refs https://github.com/keep-network/keep-core/issues/3135

Issue #3135, opened 2022-08-05
Start a client


dimpar:
We need to start a client by executing 3 scripts: `install.sh`, `initialize.sh`, and `start.sh`. This issue should handle all the setups (including config file updates) to automate the starting process as much as possible.

PRs:
- https://github.com/keep-network/keep-core/pull/3148 - initial version
- https://github.com/keep-network/keep-core/pull/3163 - improvement. Reading the operator address from the `ethereum.keyFile`

TODO:
- [ ] update docs https://github.com/keep-network/keep-core/pull/3170

nkuba:
Clients running on a local machine before starting will need to be set with:
- `ethereum.url`
- `ethereum.keyFile`
- `network.peers`
- `network.port`
- `storage.dataDir`

Contracts' addresses will not be needed once https://github.com/keep-network/keep-core/pull/3129 is merged.

For the initial version, we can expect a developer to configure TOML file with those details before starting the client.

In the future, possible improvements that I see are:
- adding interactive prompts in `install.sh` and `initialize.sh` scripts to ask the developer for the required parameters (we've got prompts already in `start.sh`),
- update `initialize.sh` to read the operator address from the `ethereum.keyFile` and use this address for initialization, instead of requiring the user to provide it with `--stake-owner` flag.

dimpar:
> adding interactive prompts in install.sh and initialize.sh scripts to ask the developer for the required parameters (we've got prompts already in start.sh)

Just to clarify, do you have in mind something like:

`install.sh` will prompt for the following configs and can overwrite the existing ones:
- `ethereum.url` - default is the existing `URL` if present
- `storage.dataDir` - default is the existing `dataDir` if present
- `network.peers` - default is empty or existing `peers` if present
- `network.port` - default is empty or existing `port` if present

`initialize.sh` will prompt for the following configs and can overwrite the existing ones:
- `ethereum.keyFile` - default is the existing `keyFile` if present

`start.sh` all the prompts stay as-is

dimpar:
After discussing improvements with @nkuba we decided to add only reading the address from the `ethereum.keyFile`.

Regarding the other params, we think, that prompting for config params in `install.sh` is too much because they are usually set only once and devs can do it by themselves. Also, prompting for each param also slows down the start, cause you have to click enter 4x times for the default values, and the default params will cover most cases.

PR #3163, opened 2022-08-13
Read keyfile and select the address

dimpar:
Refs https://github.com/keep-network/keep-core/issues/3135

Depends on https://github.com/keep-network/keep-core/pull/3148

In this PR we add an option to select the config file during `initialze.sh` script which will prompt to pick the config file and will fetch the stake owner address from the key file.

Now you can simply run `scripts/initialize.sh` without providing `--stake-owner <address>`


scripts/initialize.sh
```diff
@@ -72,9 +76,46 @@ while getopts "n:o:p:d:b:a:s:k:h" opt; do
...
+      read reply
+      if [ "$reply" -ge 1 ] && [ "$reply" -le $config_files_count ]; then
+         CONFIG_FILE_PATH=${config_files["$reply" - 1]}
+         break
+      else
+         printf "\nInvalid choice. Please choose an existing option number.\n"
+      fi
+   done
+
+   sed -n -e '/^keyFile -/p' $CONFIG_FILE_PATH
```
nkuba, line 105: Is this line doing anything?

scripts/initialize.sh
```diff
@@ -72,9 +76,46 @@ while getopts "n:o:p:d:b:a:s:k:h" opt; do
...
+
+   key_file_str=$(grep "^keyFile" $CONFIG_FILE_PATH)
+   # internal field separator, creates array from key_file separated by '"' sign
+   # array[1] is the key file path
+   IFS='"' read -r -a array <<<"$key_file_str"
```
nkuba, line 110: We could try using `awk`:
```suggestion
   key_file_path=$(awk '/keyFile ?=(.*)$/{print $3}' $CONFIG_FILE_PATH)
```

scripts/initialize.sh
```diff
@@ -72,9 +76,46 @@ while getopts "n:o:p:d:b:a:s:k:h" opt; do
...
+
+   # find address:<address> in the key file
+   address_str=$(grep -Eo '"address":.*?[^\\]"' "${array[1]}")
+   # address_array[3] is the ethereum address
+   IFS='"' read -r -a address_array <<<"$address_str"
+   printf "\nStake owner address: ${address_array[3]} \n"
+
+   stake_owner="${address_array[3]}"
```
nkuba, line 118: We should get the address from the key file, as there is no guarantee the file name will contain it. We could use `jq` for that:
```suggestion
   stake_owner=$(jq -jr .address $key_file_path)
```

nkuba (changes requested):

scripts/initialize.sh
dimpar: > We should get the address from the key file, as there is no guarantee the file name will contain it. 

Yes, we get the address from the key file, not from the file name.

scripts/initialize.sh
nkuba: We need to strip the quotes, so:
```suggestion
   key_file_path=$(awk '/keyFile ?=(.*)$/{print $3}' $CONFIG_FILE_PATH | xargs)
```

scripts/initialize.sh
```diff
@@ -72,9 +76,46 @@ while getopts "n:o:p:d:b:a:s:k:h" opt; do
...
+         CONFIG_FILE_PATH=${config_files["$reply" - 1]}
+         break
+      else
+         printf "\nInvalid choice. Please choose an existing option number.\n"
+      fi
+   done
+
+   sed -n -e '/^keyFile -/p' $CONFIG_FILE_PATH
+
+   key_file_str=$(grep "^keyFile" $CONFIG_FILE_PATH)
```
nkuba, line 107: Could we also add some prints to help debugging in case of problems?

```sh
   key_file_path=$(awk '/keyFile ?=(.*)$/{print $3}' $CONFIG_FILE_PATH | xargs)

   printf "Reading an address from the key file: ${key_file_path}\n"

   stake_owner=$(jq -jr .address $key_file_path)

   printf "Discovered the address: ${stake_owner}\n"
```

dimpar: Sure thing [ac7cf84](https://github.com/keep-network/keep-core/pull/3163/commits/ac7cf8493cb9164edaf194a3421b5f922845c111)

scripts/initialize.sh
dimpar: [1265fa5](https://github.com/keep-network/keep-core/pull/3163/commits/1265fa59d426363fee21ebfe1188d5e73bdd43be)

scripts/initialize.sh
dimpar: I like `awk` and `jq` and the only reason I didn't use it because they need to be installed separately and I wanted to use "native" methods. However, since we use them in other scripts I guess there's no harm using it here as well. [1265fa5](https://github.com/keep-network/keep-core/pull/3163/commits/1265fa59d426363fee21ebfe1188d5e73bdd43be)

scripts/initialize.sh
dimpar: not relevant anymore

scripts/initialize.sh
```diff
@@ -72,9 +76,39 @@ while getopts "n:o:p:d:b:a:s:k:h" opt; do
...
+         printf "\nInvalid choice. Please choose an existing option number.\n"
+      fi
+   done
+
+   key_file_path=$(awk '/keyFile ?=(.*)$/{print $3}' $CONFIG_FILE_PATH | xargs)
+
+   printf "Reading an address from the key file: ${key_file_path}\n"
+
+   stake_owner=`cat $key_file_path | jq -jr .address`
+
```
nkuba, line 110: We may need need to ensure `0x` prefix is added to the address, as the key file doesn't have it.

```suggestion

[[ $stake_owner == 0x* ]] || stake_owner="0x$stake_owner"

```

scripts/initialize.sh
```diff
@@ -72,9 +76,39 @@ while getopts "n:o:p:d:b:a:s:k:h" opt; do
...
+      read reply
+      if [ "$reply" -ge 1 ] && [ "$reply" -le $config_files_count ]; then
+         CONFIG_FILE_PATH=${config_files["$reply" - 1]}
+         break
+      else
+         printf "\nInvalid choice. Please choose an existing option number.\n"
+      fi
+   done
+
+   key_file_path=$(awk '/keyFile ?=(.*)$/{print $3}' $CONFIG_FILE_PATH | xargs)
```
nkuba, line 105: Let's support both `KeyFile` and `keyFile` as the TOML configuration file is not strict about properties cases. 

```suggestion
   key_file_path=$(awk '/[k|K]eyFile ?=(.*)$/{print $3}' $CONFIG_FILE_PATH | xargs)
```

nkuba (changes requested):

scripts/initialize.sh
dimpar: It works without `0x`, but I no harm adding [b06f99d](https://github.com/keep-network/keep-core/pull/3163/commits/b06f99d5366d843c555b80a89ebf6506b0a2cd83)

scripts/initialize.sh
dimpar: [1ddc7bd](https://github.com/keep-network/keep-core/pull/3163/commits/1ddc7bd97466960ac83024908d9411fd3326fb60)

nkuba (approved):

Issue #3161, opened 2022-08-12
Better manage computationally-heavy operations


pdyraga:
The client needs to generate pre-parameters for threshold ECDSA key generation. This is a computationally-heavy operation that affects the rest of the client. When testing locally, we observed the Keep client disconnecting from other Keep clients and even disconnecting from Ethereum node. Limiting the concurrency level for TSS pre-param generation improves the situation slightly but it is still not solving the problem.

Keep client, most of the time, does not require a lot of CPU cycles:
- beacon DKG happens every 2 weeks,
- relay entry happens every week,
- ECDSA DKG happens every week,
- sweep happens every 8 hours,
- redemption happens every 8 hours.

(Note that the schedule of these operations is more aggressive for testnet.)

Here is an idea that can hopefully solve this problem:
- [x] We can introduce an abstraction to manage the lifecycle of computationally-heavy operations. Such operations should be paused when the Keep client is executing any of the protocols listed above. 
- [x] The pre-params pool should be persisted on disk to survive client restarts.
- [x] We should require the client to generate a sufficient number of pre-parameters before joining ECDSA sortition pool. 
- [x] We should have a pause of a few blocks between potential ECDSA DKG re-tries to let the miner generate TSS params and re-fill the pool.
- [x] We should significantly increase the target size of pre-params pool.
- [x] We should potentially add a delay between pre-param generation iterations to not starve other processes of the client.

PR #3168, opened 2022-08-15
Marshal pre-parameters

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/3161

Generating tECDSA DKG pre-parameters is computationally expensive. We want to marshal and save on disk generated and not-yet-used pre-parameters so that they survive client restarts and the work does not have to be started from scratch.

This unit of work is just about marshaling pre-parameters.


pkg/tecdsa/gen/pb/preparams.proto
```diff
@@ -0,0 +1,29 @@
...
+message LocalPreParams {
+  // PrivateKey is not present for pre-params generated before DKG and is
+  // filled when DKG finishes work. That being said, this message will be
+  // empty in PreParams message and non-empty in PrivateKeyShare message
+  // (see key.proto).
+  message PrivateKey {
+    bytes publicKey = 1;
+    bytes lambdaN = 2;
+    bytes phiN = 3;
+  }
```
pdyraga, line 19: In 9c62fddb2f6d392bd614d204249d1d06b55fb3ff I promoted `LocalPreParams` to be defined outside of `LocalPartySaveData`. With this change, in 78dc79f16ff66dbd503c2b152b6d775016fcbc0b, I could move `LocalPreParams` message to a separate file. This message is used by `LocalPartySaveData `  and by `PreParams` that will be saved on disk.

I am struggling with `PrivateKey` part: it is empty for `PreParams` (before DKG) and is filled for `LocalPartySaveData`. We can leave it as-is or:
1. Revert 9c62fddb2f6d392bd614d204249d1d06b55fb3ff, keep `LocalPreParams` as internal for `LocalPartySaveData`
2. Introduce separate `LocalPreParams` without `PrivateKey` and use it as an internal message of `PreParams`.

@lukasz-zimnoch thoughts on that?

lukasz-zimnoch: I'd vote for separate protobuf definitions for both cases. My reasoning: in this case, a little duplication seems to be a much smaller problem than introducing the additional complexity related to a field that can be in different states depending on the usage context. Protobuf definitions are meant to describe the shape of data and I think we should not have any details regarding how those data are processed here (i.e. this field should be empty before DKG but filled after). Additionally, the proposed structure can be error-prone, especially regarding the marshaling code. This is why, I think, duplication is better here, especially given the fact those data structures are not supposed to be changed very often (if at all).

pdyraga: Done in https://github.com/keep-network/keep-core/pull/3168/commits/8ff42ee9956273dc3cd54f429dbeb9ce0eba8e91.

pkg/tecdsa/preparams.go
```diff
@@ -0,0 +1,15 @@
+package tecdsa
+
+import "github.com/binance-chain/tss-lib/ecdsa/keygen"
+
+// PreParams represents tECDSA DKG pre-parameters that were not yet consumed
+// by DKG protocol execution.
```
lukasz-zimnoch, line 6: Maybe just:
```suggestion
// PreParams represents a pre-parameters set that is required for tECDSA DKG participation.
```
I think there is no need to mention how those parameters are used in detail. We are on a quite high abstraction level here.

pdyraga: I am not sure about it. I mentioned they are not yet consumed because they do not have `PrivateKey paillierSK`. This representation makes sense only pre-DKG and this sentence is to ensure no one tries to use them or thinks to associate them with DKG execution.

lukasz-zimnoch (approved):

PR #3169, opened 2022-08-15
ECDSA DKG test loop: Retries

lukasz-zimnoch:
Refs: #3041 

#3119

Here we integrate the tECDSA DKG protocol with the retry algorithm developed during #3119. Specific changes are described below.

### The DKG retry loop

Long story short, the tECDSA DKG protocol can fail for multiple reasons. Two very broad categories of errors are inactivities and byzantine faults. Regarding inactivities, we are usually able to detect specific members who were inactive and exclude them during the retry. Worth noting that based on observations made during the v1 lifetime, inactivities are the most common errors that break the DKG protocol execution. Regarding byzantine faults, the tECDSA protocol we use does not support attributability so, in most cases, we are not able to point out specific members who spread invalid data accidentally or on purpose. For that case, we fall back to the random retry mechanism from #3119. This mechanism can be imagined as a black box that takes the list of group operator seats selected by the sortition pool as an input and returns a sublist of operator seats that should participate in the given DKG attempt. That said, we implemented the DKG retry loop as follows:

0. For the first attempt, set all selected members as qualified 
1. Do the DKG attempt if the executing member is qualified for that attempt. Otherwise - proceed to point 4.
2. If there is no error and the result was produced - break. If there is an error - proceed.
3. Unwrap the error. If the error is about inactive members, record their indexes
4. If we are in the first 5 attempts, we have not used random retries yet, and there are inactive members - exclude the inactive members if the group quorum can still be maintained without them. Otherwise - use the random algorithm to exclude members randomly. Members who were not excluded are qualified to participate in the next attempt.
5. Check the stop signal. If triggered - break. Otherwise - proceed to point 1.

Such a construction of the retry loop allows focusing on inactivities from the very beginning. If initial inactivities-based retries don't help, we fall back to the random algorithm and continue with it until success or timeout.

The entire DKG retry loop logic was encapsulated into the `tbtc.dkgRetryLoop` struct. This construction allows to separate the loop's internal state from the actual DKG execution logic and facilitates testing of the retry logic.

### Changes in the DKG executor

In order to work with the test loop, the DKG executor (`pkg/tecdsa/dkg.Executor`) must allow performing a DKG run with a sublist of the original group selection. In order to achieve that, the executor's `Execute` function now takes the `excludedMembers []group.MemberIndex` argument holds indexes of members excluded for the given DKG run. The executing member just considers them disqualified from the very beginning and thus does not accept messages from those members. This information is also used to set up the underlying TSS implementation in a correct way.

Apart from that, the DKG retry loop needs to recognize errors raised due to inactive members. In order to handle this, the `Execute` function can now return the pre-defined `dkg.InactiveMembersError` that holds indexes of members inactive during the given attempt. The DKG retry loop checks for that error using the `errors.As` call.

### The DKG session ID

So far, all TSS-related network messages used the seed-based `sessionID` field that was used to ensure that those messages belong to the right DKG run. Adding retries complicates the situation since there can be multiple messages of the same type that would use the same `sessionID`. From the network's perspective, each retry is a separate session so we now build the `sessionID` as a concatenation of the DKG seed and the DKG attempt index. This way, each retry for the same seed has its own `sessionID` and there is no way to mix messages. Additionally, the support of the `sessionID` field was added to the non-TSS `ephemeralPublicKeyMessage`. 

An alternative to the `sessionID` could be creating a separate broadcast channel for each DKG retry. However, a broadcast channel is a heavyweight concept that is much harder to manage than the `sessionID`. 

### Minor changes

Apart from the aforementioned changes, two smaller ones were introduced as well:
- The `GroupQuorum` parameter was added to the `tbtc.ChainConfig` type. This parameter determines the minimum number of active participants behaving according to the protocol needed to generate a group in TBTC. It is lesser than the `GroupSize` which determines the ideal group size but significantly greater than the `HonestThreshold` parameter which denotes the minimum number of signers required to produce a signature.
- The `chain.Addresses` type was added in order to facilitate dealing with address lists and transforming them to sets. This is especially useful for handling groups formed from operator addresses and allows for reducing a lot of boilerplate code.


lukasz-zimnoch:
Code is ready but not yet tested nor covered with unit tests. I'll undraft once those missing pieces are addressed.

pkg/tbtc/node.go
```diff
@@ -153,18 +161,70 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
 				n.protocolLatch.Lock()
 				defer n.protocolLatch.Unlock()
 
-				result, _, err := n.dkgExecutor.Execute(
+				retryLoop := newDkgRetryLoop(
```
beaushinkle, line 164: `retryLoop` and `newDkgRetryLoop` both evoke a sense of action to me; that we'll be calling `retryLoop` with some arguments because it's *the loop itself*, and `newDkgRetryLoop` is a factory that generates loops.

Looking at what it actually represents, I think it's closer to "loop information" or "loop state", so maybe `retryLoopInfo` or `retryLoopState` something in that vein?

beaushinkle: I take it back!
```
func (drl *dkgRetryLoop) start(
	ctx context.Context,
	dkgAttemptFn dkgAttemptFn,
) (*dkg.Result, error) {
```
It *can* loop!

pkg/tbtc/node.go
```diff
@@ -205,3 +265,203 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+		// was slightly longer than the expected duration determined by the
+		// dkg.ProtocolBlocks function. For example, the attempt may fail at
+		// the end of the protocol but the error is returned after some time
+		// and more blocks than expected are mined in the meantime.
+		// Additionally, we want to strongly extend the delay period
+		// periodically in order to give some additional time for nodes to
+		// recover and re-fill their internal TSS pre-parameters pools.
+		if drl.attemptCounter > 1 {
+			delayBlocks := drl.delayBlocks
+			if drl.attemptCounter%drl.delayBlocksBumpFrequency == 0 {
```
beaushinkle, line 364:
```suggestion
			if drl.attemptCounter % drl.delayBlocksBumpFrequency == 0 {
```

pkg/tbtc/node.go
```diff
@@ -205,3 +265,203 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+		selectedOperators:        selectedOperators,
+		inactiveOperatorsSet:     make(map[chain.Address]bool),
+		chainConfig:              chainConfig,
+		attemptCounter:           0,
+		attemptStartBlock:        initialStartBlock,
+		randomRetryCounter:       0,
+		randomRetrySeed:          randomRetrySeed,
+		delayBlocks:              5,
+		delayBlocksBumpFrequency: 100,
+		delayBlocksBumpFactor:    20,
```
beaushinkle, line 313: This structure is currently a mix of constants like `delayBlocksBumpFactor` and mutable variables like `attemptCounter`.

I would find it easier to reason about if we separated out the constants somewhere else, that way we're calling something like

`delayBlocks *= DkgRetryConstants.DELAY_BLOCKS_BUMP_FACTOR` instead

pkg/tbtc/node.go
```diff
@@ -205,3 +265,203 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+		// block by the number of blocks equal to the protocol duration and
+		// by some additional delay blocks. We need a small fixed delay in
+		// order to mitigate all corner cases where the actual attempt duration
+		// was slightly longer than the expected duration determined by the
+		// dkg.ProtocolBlocks function. For example, the attempt may fail at
+		// the end of the protocol but the error is returned after some time
+		// and more blocks than expected are mined in the meantime.
+		// Additionally, we want to strongly extend the delay period
+		// periodically in order to give some additional time for nodes to
+		// recover and re-fill their internal TSS pre-parameters pools.
```
beaushinkle, line 361: This comment is really chunky, let's add some paragraphs for readability:

```
		// In order to start attempts >1 in the right place, we need to
		// determine how many blocks were taken by previous attempts. We assume
		// the worst case that each attempt failed at the end of the DKG
		// protocol. 
		//
		// That said, we need to increment the previous attempt start
		// block by the number of blocks equal to the protocol duration and
		// by some additional delay blocks. We need a small fixed delay in
		// order to mitigate all corner cases where the actual attempt duration
		// was slightly longer than the expected duration determined by the
		// dkg.ProtocolBlocks function. 
		//
		// For example, the attempt may fail at
		// the end of the protocol but the error is returned after some time
		// and more blocks than expected are mined in the meantime.
		// Additionally, we want to strongly extend the delay period
		// periodically in order to give some additional time for nodes to
		// recover and re-fill their internal TSS pre-parameters pools.
```

pkg/tbtc/node.go
```diff
@@ -205,3 +265,203 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+		for i, operator := range drl.selectedOperators {
+			if !qualifiedOperatorsSet[operator] {
+				memberIndex := group.MemberIndex(i + 1)
+				excludedMembers = append(excludedMembers, memberIndex)
+
+				// If the given member was not qualified for the given attempt,
+				// mark this attempt as skipped in order to skip the execution
+				// and set up the next attempt properly.
+				if memberIndex == drl.memberIndex {
+					attemptSkipped = true
```
beaushinkle, line 386: I think that if we hit this condition, we don't need to continue populating `excludedMembers`. Is it possible to break a `for range`?

pkg/tbtc/node.go
```diff
@@ -205,3 +265,203 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+		len(drl.inactiveOperatorsSet) > 0 {
+		qualifiedOperators := make(chain.Addresses, 0)
+		for _, operator := range drl.selectedOperators {
+			if !drl.inactiveOperatorsSet[operator] {
+				qualifiedOperators = append(qualifiedOperators, operator)
+			}
+		}
+
+		if len(qualifiedOperators) >= drl.chainConfig.GroupQuorum {
+			return qualifiedOperators.Set(), nil
```
beaushinkle, line 447: Why are we building up a slice of operators and then converting it to a set, rather than building up that set directly?

ie:
```
qualifiedOperators := make(map[chain.Address]bool, len(drl.selectedOperators))
operatorCount := 0
for _, operator := range drl.selectedOperators {
	if !drl.inactiveOperatorsSet[operator] {
		qualifiedOperators[operator] = true
		operatorCount++
	}
}

if operatorCount >= drl.chainConfig.GroupQuorum {
	return qualifiedOperators, nil
}
```

beaushinkle: if we're going to immediately turn the list to a set, I can rewrite the retry API to return that directly and reduce complexity; I'm going out of the way to return a list at the moment

pkg/tbtc/node.go
```diff
@@ -205,3 +265,203 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"random operator selection failed: [%w]",
+			err,
+		)
+	}
+
+	drl.randomRetryCounter++
+	return chain.Addresses(qualifiedOperators).Set(), nil
```
beaushinkle, line 466: if we're going to immediately turn the list to a set, I can rewrite the retry API to return that directly and reduce complexity; I'm going out of the way to return a list at the moment

pkg/tbtc/node.go
```diff
@@ -205,3 +265,203 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+		drl.randomRetryCounter == 0 &&
+		len(drl.inactiveOperatorsSet) > 0 {
+		qualifiedOperators := make(chain.Addresses, 0)
+		for _, operator := range drl.selectedOperators {
+			if !drl.inactiveOperatorsSet[operator] {
+				qualifiedOperators = append(qualifiedOperators, operator)
+			}
+		}
+
+		if len(qualifiedOperators) >= drl.chainConfig.GroupQuorum {
```
beaushinkle, line 446: This, in combination with the argument to `EvaluateRetryParticipantsForKeyGeneration` being `drl. selectedOperators`, means if *this* attempt pushes us below quorum, we go into random retries with the full group from sortition, rather than `drl.selectedOperators` right?

I think that's a good way to do it, but probably worth calling that out explicitly in a comment somewhere

pkg/tbtc/node.go
lukasz-zimnoch: This is automatically formatted by `gofmt`

pkg/tbtc/node.go
lukasz-zimnoch: We may want to make them configurable in the near future. Those are not true constants.

pkg/tbtc/chain.go
```diff
@@ -54,8 +54,12 @@ type Chain interface {
 
 // ChainConfig contains the config data needed for the TBTC to operate.
 type ChainConfig struct {
-	// GroupSize is the size of a group in TBTC.
+	// GroupSize is the target size of a group in TBTC.
 	GroupSize int
+	// GroupQuorum is the minimum number of active participants behaving
+	// according to the protocol needed to generate a group in TBTC. This value
+	// is lesser than the GroupSize and bigger than the HonestThreshold.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See aa8d36ce

pkg/tbtc/node.go
```diff
@@ -205,3 +265,203 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
 		logger.Infof("not eligible for DKG with seed [0x%x]", seed)
 	}
 }
+
```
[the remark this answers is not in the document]
lukasz-zimnoch: This is intentional and makes sense when you take a look at how the `tbtc.signer.String()` method is implemented.

pkg/tbtc/node.go
lukasz-zimnoch: Indeed! See 994d37d3

pkg/tbtc/node.go
lukasz-zimnoch: Yep we can break here: 054c6cda

pkg/tbtc/node.go
lukasz-zimnoch: Your version compares the distinct operators count to the `GroupQuorum`. However, we need to compare the number of members' seats. The `qualifiedOperators` slice is built from `drl.selectedOperators` which represents the group selection result which is about members' seats. I know this is a bit confusing as we are juggling between both representations of the group members (seats-oriented and operators-oriented) but I tried to follow a rule that every collection that has the `*Set` prefix is about distinct operators forming the group while everything else is about members seats (which are unfortunately represented using operators addresses as well).

pkg/tbtc/node.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3169#discussion_r953666514

pkg/tbtc/node.go
lukasz-zimnoch: The `drl.selectedOperators` is exactly the sortition (selection) result but the rest is true. I posted a comment about that: 5aceaa60

pkg/tbtc/node.go
beaushinkle: configurable on a by-loop level or by-client level?

pkg/tbtc/node.go
beaushinkle: In mine, I keep track of the number of total seats via incrementing `operatorCount`, and then comparing `operatorCount >= drl.chainConfig.GroupQuorum` at the end, so while we might be overwriting the same `operator` key with true over and over, `operatorCount` keeps increasing. I think it should work!

> I tried to follow a rule that every collection that has the *Set suffix is about distinct operators forming the group while everything else is about members seats (which are unfortunately represented using operators addresses as well)

Yeah - totally understand the seats vs operators separation. I was making the point that this function `qualifiedOperatorsSet()` appears to be returning an operator Set, so we can build that Set up directly. I think it's more efficient computationally, and also easier to tell what the program is doing at a glance: make a set, and build it up vs make a list, build it up, convert it into a set.

pkg/tbtc/node.go
beaushinkle: I read through the referenced comment, but I think this is potentially separate!

As I understand it, right now we have a need to represent the information as both a list of seats and a set of operators. The set of operators can always be inferred from the list of seats, but not the other way around and so the set of operators is *lossy*.

That said, are there other places where we intend to call `EvaluateRetryParticipantsForKeyGeneration`? Because here, we immediately turn the list of seats into a set of operators via `.Set()`. If we *always* plan on doing this (or are okay with some refactoring in the case where we thought we would always do this, but then later need the list), I can make `EvaluateRetryParticipantsForKeyGeneration` return a `Set` directly, and it would simplify a lot of that code, and the calling code as well.

pkg/tbtc/node.go
lukasz-zimnoch: Those params must be the same for all group members so we cannot configure them by-client. However they can be the result of computations done using the same algorithm by all members. Unclear at the moment but I can imagine that we may need to shorten the delay if we are close to the timeout block.

pkg/tbtc/node.go
lukasz-zimnoch: Honestly, I'd leave it as is. The current version is similar to how the sortition pool returns the group selection result. This is the natural and most common way of dealing with group members while the operator set is specific only to this case. The slice also gives some additional information that is not present in the set, like the operator's seat count. Although changing it to a set can seem justified at the moment, keeping it as a slice gives us more flexibility for the future in case we need to refactor this code. At the same time, the computational efficiency difference is rather negligible for such a small number of items (100) so it may be not worth touching it at all.

pkg/tbtc/node.go
lukasz-zimnoch: Oh yeah, I misread how `operatorCount` is constructed. However, I think keeping it as a slice is still ok as we are returning a slice from the retry API so this remains consistent. In the same time, I think that returning a slice form the retry API is also justified: https://github.com/keep-network/keep-core/pull/3169#discussion_r953889322

beaushinkle: Sounds good to me :100:

pkg/tecdsa/dkg/dkg.go
```diff
@@ -42,16 +41,21 @@ func NewExecutor(
 	}
 }
 
-// Execute runs the ECDSA distributed key generation protocol, given a
+// Execute runs the tECDSA distributed key generation protocol, given a
 // broadcast channel to mediate with, a block counter used for time tracking,
 // a member index to use in the group, dishonest threshold, and block height
 // when DKG protocol should start.
+//
+// This function also allows to perform a DKG execution with just a subset of
```
beaushinkle, line 49:
```suggestion
// This function also supports DKG execution with a subset of
```

beaushinkle:
Okay! done with my pass - other than that last grammar nit, I'm good to go here. We have a merge conflict, but once that gets resolved, I'm good to ship

pkg/tecdsa/dkg/dkg.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3169/commits/7ca993e117969cdc1e7b1259e9b786c061e2cb4f

lukasz-zimnoch:
Ready to :shipit: @beaushinkle

beaushinkle (approved):

pkg/beacon/dkg/dkg.go
```diff
@@ -212,12 +212,12 @@ func waitForDkgResultEvent(
...
-//   to the group member with ID N+1.
-// - operatingGroupMembersIDs: Contains group members IDs that were neither
-//   disqualified nor marked as inactive. Slice length is lesser than or equal
-//   to the groupSize.
+//   - selectedOperators: Contains addresses of all selected operators. Slice
+//     length equals to the groupSize. Each element with index N corresponds
+//     to the group member with ID N+1.
+//   - operatingGroupMembersIDs: Contains group members IDs that were neither
+//     disqualified nor marked as inactive. Slice length is lesser than or equal
+//     to the groupSize.
```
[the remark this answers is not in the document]
lukasz-zimnoch: WTF happened here... Looks like `gofmt` was applied on the merge commit. I'll fix that in a separate PR.

PR #3171, opened 2022-08-15
Added a delay between pre-parameters generation

pdyraga:
Refs #3161 

There is a configurable delay between the generation of pre-parameters.
By default, the delay is 10 seconds. If pre-parameters generation takes a long
time and new parameters are generated one after another, all other client
processes are starved, without access to CPU. Adding a delay should help with
releasing some computing power for a moment.

Still testing it, so PR is in a draft. The code is ready for a review though.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -229,6 +229,13 @@ func initTbtcFlags(cmd *cobra.Command, cfg *config.Config) {
 		"tECDSA pre-parameters generation timeout.",
 	)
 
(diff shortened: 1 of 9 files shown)


lukasz-zimnoch (approved):
LGTM! I'm happy to merge once it's undrafted!

pdyraga:
Tested locally and works as advertised. It's in your hands now @lukasz-zimnoch

PR #3172, opened 2022-08-17
ParameterPool abstraction 

pdyraga:
Refs #3161 

To have the generated TSS pre-parameters survive client restarts, we need to add a persistence layer to the pool. The abstraction allows to define and test all functionalities separately, without having to worry about the complexity of the generation of real parameters.

This way, TSS pre-parameter code can be just wiring things up: take the generation function, take the persistence layer, provide them to the introduced abstraction, and voila! The complexity of the pool logic can be separated from the complexity of wiring things up and this was the main goal of the code in this PR. Having the logic and wiring code in one component is too much. `pkg/tecdsa/dkg/preparams.go` will be just about wiring up the building blocks.


pkg/miner/pool.go
```diff
@@ -0,0 +1,97 @@
...
+	"time"
+
+	"github.com/ipfs/go-log"
+)
+
+// Persistence defines the expected interface for storing and loading generated
+// and not-yet-used parameters on persistent storage. Generating parameters is
+// usually a computationally expensive operation and generated parameters should
+// survive client restarts.
+type Persistence[T any] interface {
```
lukasz-zimnoch, line 14: Shouldn't we add some type constraints for `T`? I think we should make sure that this type can be persisted, i.e. implements `Marshal` and `Unmarshal` functions. WDYT?

pkg/tecdsa/dkg/preparams.go
```diff
@@ -0,0 +1,84 @@
...
+// the generated numbers.
+func NewPreParams(data keygen.LocalPreParams) *PreParams {
+	return &PreParams{data}
+}
+
+// tssPreParamsPool is a pool holding TSS pre parameters. It autogenerates
+// entries up to the pool size. When an entry is pulled from the pool it
+// will generate a new entry.
+type tssPreParamsPool struct {
+	*miner.ParameterPool[keygen.LocalPreParams]
```
lukasz-zimnoch, line 27: Shouldn't we build the pool around `dkg.PreParams` instead of `keygen.LocalPreParams`? The reason behind is that the former implements `Marshal` and `Unmarshal` so it can be persisted while the latter cannot. Also, if the storage implementation is implemented somewhere outside, tss-lib internal type will leak outside of the `pkg/tecdsa`.

pdyraga: Indeed! 45d5df162d4bd69546b4b98d729cbaa489739e6b

pkg/miner/pool.go
pdyraga: I was thinking about it but when we would assume how `Persistence` works internally. It is possible some implementation does not want/need to go to bytes to persist.

lukasz-zimnoch (approved):

PR #3174, opened 2022-08-17
Block miner computations until everyone agrees to resume

pdyraga:
Refs #3161 
Depends on #3172 

From now on, Miner has an internal latch counting how many times `Stop()` function has been called. Miner expects `Resume()` function to be called the same number of times. The reasoning behind it is that multiple goroutines can call `Stop()` and we do not want the computations to unblock until the last goroutine is done.

For example:
1. Relay entry signing starts, computations are stopped.
2. ECDSA DKG starts, computations are already stopped.
3. Relay entry signing completes. Relay entry goroutine asks the computations to resume.

With the change in this PR we want to resume the computations only when all goroutines that requested to stop them have _signalled_ that they completed their work and computations can be resumed.

The _signalled_ element is crucial. We really need all of them to tell the miner that the computations can be resumed. We do not know when the operations that are supposed to stop the computations are invoked and how long they are going to take. I was playing with an idea of a timeout but that turned out to be complicated and error-prone.

Since we need to trust goroutines to call `Resume`, it means we assume goroutines are working as expected. I am not totally in love with this assumption but I could not figure out anything else if we do not want the generator to just keep running in the background.

If we assume goroutines are working as expected and always call `Resume` after `Stop`, it also means we assume they are not going to call `Resume` twice because it breaks the assumption of goroutines being correctly implemented.

In the current setup, the lock is acquired only to actually stop or resume goroutines and after the lock is acquired we check the state. `Stop` and `Resume` are not atomic but I think everything should work fine under the assumption goroutines are:
1. Always calling `Stop()` before `Resume()`
2. They never forget to call `Resume()` if they called `Stop()`.

I am open to all suggestions on how to improve this mechanism. I could come up only with this solution if we want to stop the miner. I was also thinking about abandoning this path completely and reworking the pre-params generator code in tss-lib to always work in a single thread, even if generating pre-params will take ages but I don't think it's a better option. 4 years ago, when we were working on the safe prime generator and were optimizing the code pretty heavily, we [decided to go with concurrency](https://github.com/didiercrunch/paillier/blob/master/safe_prime_generator.go#L63) ([benchmarks](https://github.com/didiercrunch/paillier/pull/9#issuecomment-409124089)). What appeals to me about the current approach is that in the real world, we'll not have thousands of goroutines on a single client. Just a few executing DKG or relay entry generation and the call to the miner is quite simple: `miner.Stop(); defer miner.Resume()`.


pkg/miner/miner.go
```diff
@@ -49,11 +50,21 @@ func (m *Miner) Mine(miningFn func()) {
...
 // call the worker function again until the miner's work is resumed.
+// Stop can be called multiple times. Each call increases a count on the
+// internal latch by one. To resume the work, Resume function needs to be as
+// many times as Stop(). This way, several goroutines can stop the computations
+// and they all need to agree on resuming them.
 func (m *Miner) Stop() {
+	m.latch.Add(1)
+
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
```
lukasz-zimnoch, line 61: Is this order of latch and mutex intentional (same in `Resume`)? In the current form, both operations (`Stop` and `Resume`) are not atomic. For example, if we trigger two goroutines, the first calling `Stop` and the second `Resume`, I think the following is possible:
1. G1 calls `Stop` and increments the latch's counter
2. G2 calls `Resume` and overtakes G1 by decrementing the counter and taking the mutex lock then makes some operations and unlocks the mutex.
3. G1 takes the mutex lock and performs its own operations

Although such a scenario is harmless in the current implementation, I think it's quite error-prone and may cause weird problems in the future. Shouldn't we always take the mutex first?

pkg/miner/miner_test.go
```diff
@@ -105,6 +105,70 @@ func TestMineStopResume(t *testing.T) {
 	)
 }
 
+func TestMineStopResumeReentrant(t *testing.T) {
```
lukasz-zimnoch, line 108: Can we add a description similar to other test cases in this file?

pkg/miner/miner.go
```diff
@@ -63,12 +74,21 @@ func (m *Miner) Stop() {
 }
 
 // Resume resumes the work of all worker functions, each in a separate
-// goroutine.
+// goroutine. If Stop() has been called multiple times, Resume() needs to be
+// called the same number of times. Resume() blocks until the work can be
+// resumed.
 func (m *Miner) Resume() {
```
lukasz-zimnoch, line 80: I think `Resume` is not safe for redundant calls. If we have a working miner and call `Resume` by mistake, the latch's counter will become negative and cause a panic. Shouldn't we make redundant `Resume` calls a no-op? It would be also good to document this behavior in the docstring and cover it with a test case.

pkg/miner/miner.go
pdyraga: With the change in this PR we want to resume the computations only when all goroutines that requested to stop them have _signalled_ that they completed their work and computations can be resumed.

The _signalled_ element is crucial. We really need all of them to tell the miner that the computations can be resumed. We do not know when the operations that are supposed to stop the computations are invoked and how long they are going to take. I was playing with an idea of a timeout but that turned out to be complicated and error-prone.

Since we need to trust goroutines to call `Resume`, it means we assume goroutines are working as expected. I am not totally in love with this assumption but I could not figure out anything else if we do not want the generator to just keep running in the background.

If we assume goroutines are working as expected and always call `Resume` after `Stop`, it also means we assume they are not going to call `Resume` twice because it breaks the assumption of goroutines being correctly implemented (this answers the question in another comment).

In the scenario you described, G2 would first need to call `Stop`:
1. G2 calls `Stop` and increments the latch's counter. Let's say it's preempted after the latch.
2. G1 calls `Stop` and increments the latch's counter. It's stopping computations
4. G1 calls `Resume` and blocks on the latch.
5. G2 is still in `Stop`. It acquires the lock and sees that the miner is already stopped so it returns.
6. G2 calls `Resume` and it's resuming computations.
7. G1 is still in `Resume`. It acquires the lock and sees that the miner is already working so it returns.

In the current setup, the lock is acquired only to actually stop or resume goroutines and after the lock is acquired we check the state. `Stop` and `Resume` are not atomic but I think everything should work fine under the assumption goroutines are:
1. Always calling `Stop()` before `Resume()`
2. They never forget to call `Resume()` if they called `Stop()`.

If we tried to make both `Stop()` and `Miner()` atomic and acquire the lock at the beginning, I think we would run into a deadlock:

1. G1 calls `Stop`, acquires the lock, increases the counter on the latch, and stops the computations.
2. G2 calls `Stop`, acquires the lock, and increased the counter on the latch.
3. G2 calls `Resume`, acquires the lock, decreases the counter on the latch, and blocks on wait.
4. G1 calls `Resume` but it can not acquire the lock, so it waits.

I am open to all suggestions on how to improve this mechanism. I could come up only with this solution if we want to stop the miner. I was also thinking about abandoning this path completely and reworking the pre-params generator code in tss-lib to always work in a single thread, even if generating pre-params will take ages but I don't think it's a better option. 4 years ago, when we were working on the safe prime generator and were optimizing the code pretty heavily, we [decided to go with concurrency](https://github.com/didiercrunch/paillier/blob/master/safe_prime_generator.go#L63) ([benchmarks](https://github.com/didiercrunch/paillier/pull/9#issuecomment-409124089)). What appeals to me about the current approach is that in a real world, we'll not have a thousands of goroutines on a single client. Just a few executing DKG or relay entry and the call to the miner is quite simple: `miner.Stop(); defer miner.Resume()`.

pkg/miner/miner.go
pdyraga: See [this](https://github.com/keep-network/keep-core/pull/3174#discussion_r948823835) comment please.

pkg/miner/miner.go
pdyraga: (I have also updated the PR description)

pkg/miner/miner_test.go
pdyraga: https://github.com/keep-network/keep-core/pull/3174/commits/e144844dcbf8dfaee89d23428a0c476907f76fd0

pkg/miner/miner.go
pdyraga: ... And I can also clarify it in the code documentation in case we can't come up with anything better.

lukasz-zimnoch: Ok, that makes sense. Let's just make the assumptions about client goroutines super explicit in the documentation and mention that redundant calls to `Resume` will cause a panic.

lukasz-zimnoch:
Actually I think we have the following alternative to the presented approach:

1. The miner has an internal control loop that decides about resuming and stopping the work based on a stop predicate provided as a constructor parameter (simple boolean function) and checked every X seconds. If the predicate is true, it stops. If false, it resumes.
2. The actual predicate is a boolean function that will do the conjunction of two other boolean function calls. Let's name them `tbtc.node.IsWorking()` and `tbtc.beacon.IsWorking()`
3. Respective `IsWorking` functions return a boolean answering the question: `is the total count of protocol goroutines bigger than 0?`
4. Respective node implementations can expose a helper function `triggerProtocol` that will run goroutines for respective protocols and do the accounting that will be used by `IsWorking`. The `triggerProtocol` function will be invoked for each member of DKG, signing, etc.

This way we don't need to bother about controlling the miner lifecycle by client goroutines. The lifecycle controller will be single-threaded so there will be no concurrency problems as well.

WDYT?

pdyraga:
I think this option is definitely worth considering and experimenting with. Regarding goroutines assumptions, I think we'll have the same as now because it is not enough to say "DKG is happening" - we need to know if the current node participates in that DKG and when the last goroutine finished.

But! This may be simpler in terms of concurrency because we should be able to eliminate the latch. Also, we'll avoid passing miner reference around which I already tried locally and didn't like.

I would do the following: let's review/merge #3177 and #3180 to have the current approach finished in terms of miner and pool (without yet stopping/resuming miner from goroutines) and then I'll start experimenting with the approach you suggested.

pkg/miner/miner.go
pdyraga: https://github.com/keep-network/keep-core/pull/3174/commits/e144844dcbf8dfaee89d23428a0c476907f76fd0

lukasz-zimnoch (approved):

lukasz-zimnoch (approved):

PR #3182, opened 2022-08-18
Hardhat task for requesting new ECDSA wallets

lukasz-zimnoch:
Refs: #3041 

Here we introduce a Hardhat task that allows requesting a new ECDSA wallet. This task should facilitate testing on local Ethereum networks. In order to run it, move to the `solidity/ecdsa` directory and do:
```
npx hardhat request-new-wallet --wallet-owner-address <owner> --network <network-name>
```
where `owner` is the wallet owner's address that can be set using the `initialize-wallet-owner` Hardhat task.

### DIFF
--- a/solidity/ecdsa/tasks/request-new-wallet.ts
+++ b/solidity/ecdsa/tasks/request-new-wallet.ts
@@ -1,0 +1,31 @@
+import { task } from "hardhat/config"
+
+import type { HardhatRuntimeEnvironment } from "hardhat/types"
+
+task("request-new-wallet", "Requests for a new ECDSA wallet")
+  .addParam("walletOwnerAddress", "The Wallet Owner's address")
+  .setAction(async (args, hre) => {
+    const { walletOwnerAddress } = args
+
+    await requestNewWallet(hre, walletOwnerAddress)
+  })
+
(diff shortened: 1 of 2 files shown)


solidity/ecdsa/tasks/request-new-wallet.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (!ethers.utils.isAddress(walletOwnerAddress)) {
+    throw Error(`invalid address: ${walletOwnerAddress}`)
+  }
+
+  const walletOwner = await ethers.getSigner(walletOwnerAddress)
+
+  const walletRegistry = await helpers.contracts.getContract("WalletRegistry")
+
+  const tx = await walletRegistry.connect(walletOwner).requestNewWallet()
```
pdyraga, line 27: I am not sure if this is correct. In the deployment scripts we transfer the wallet owner role to the `Bridge`, so calling the function from the `Bridge` address should not be possible.

Also, I think we no longer need this task after https://github.com/keep-network/keep-core/pull/3253 - we can call beacon genesis and request a new wallet from CLI.

What about closing this PR without merging?

PR #3183, opened 2022-08-18
Pre-param generation: cleanup and wiring this together

pdyraga:
Refs #3161 

The first change is the series of renames and function export changes. To not confuse our operators that the Keep client is "mining" blocks, the `miner` package and struct have been renamed to `generator` and `Scheduler`. The `Scheduler` name better describes the responsibilities of the structure - it is stopping and resuming long-running tasks.

Secondly, we no longer expect protocol goroutines to call `Stop()` and `Resume()`. Instead, there is a global monitoring goroutine started when the `Scheduler` is created that queries the protocols. This change allows removing the quite complicated concept of a `WaitGroup`-based latch.

`ProtocolLatch` increases the internal counter every time protocol execution starts and decreases the counter every time protocol execution completes. The latch implements `Protocol` interface and can be registered in the `Scheduler`.

The protocol code using the latch must guarantee that:
1. `Lock()` is always called before `Unlock()`
2. `Unlock()` is eventually called for every `Lock()`.

Note that the `Unlock()` function may panic if the conditions are not met.

The `Scheduler` and `ProtocolLatch` have been integrated with all protocols executed by the client:
- random beacon DKG,
- ECDSA DKG,
- random beacon relay entry signing.

The `Scheduler` should stop and resume computations according to the state of protocols executed.

The default size of pre-parameters pool has been increased to 3000 elements. Pre-parameters are generated only when no other protocol is running and we should squeeze from that time as much as possible so that we have enough pre-parameters for all potential DKG ECDSA retries. Assuming generating single pre-parameters takes 1 minute, then generating 3000 pre-parameters takes 2 days.


pkg/generator/scheduler.go
```diff
@@ -0,0 +1,151 @@
...
+	state     state
+	workers   []func(context.Context)
+	stops     []context.CancelFunc
+	workMutex sync.Mutex
+
+	protocols      []Protocol
+	protocolsMutex sync.Mutex
+}
+
+func (s *Scheduler) RegisterProtocol(protocol Protocol) {
```
lukasz-zimnoch, line 42: Can we add a docstring for this public function?

pkg/generator/scheduler.go
```diff
@@ -0,0 +1,151 @@
...
+	s.state = working
+
+	for _, worker := range s.workers {
+		s.startWorker(worker)
+	}
+}
+
+// StartWorker takes the provided worker function, creates for it an individual
+// context and starts executing it in the loop until the context is done.
+func (s *Scheduler) startWorker(workerFn func(context.Context)) {
```
lukasz-zimnoch, line 106: Can we add a mention that this function should be executed only under the condition that the `workMutex` is locked?

pkg/generator/latch.go
```diff
@@ -0,0 +1,47 @@
...
+// Scheduler.
+//
+// The protocol code using the latch must guarantee that:
+// 1. `Lock()` is always called before `Unlock()`
+// 2. `Unlock()` is eventually called for every `Lock()`.
+//
+// Note that the Unlock() function may panic if the conditions are not met.
+type ProtocolLatch struct {
+	counter uint64
+	mutex   sync.RWMutex
```
lukasz-zimnoch, line 17: As an alternative, we can get rid of the mutex and use `atomic.AddUint64` for writes and `atomic.LoadUint64` for reads.

pkg/generator/latch.go
```diff
@@ -0,0 +1,47 @@
...
+// starts and decreases the counter every time protocol execution completes.
+// The latch implements Protocol interface and can be registered in the
+// Scheduler.
+//
+// The protocol code using the latch must guarantee that:
+// 1. `Lock()` is always called before `Unlock()`
+// 2. `Unlock()` is eventually called for every `Lock()`.
+//
+// Note that the Unlock() function may panic if the conditions are not met.
+type ProtocolLatch struct {
```
lukasz-zimnoch, line 15: Let's add docstrings for all public functions of this type.

pkg/generator/latch.go
pdyraga: Yeah, and I started with it but `atomic.AddUint64(&counter, ^uint64(0))` to decrement felt less readable than just mutex. But I am fine with both options and can change it to `atomic.AddUint64` if you prefer it.

pkg/generator/scheduler.go
pdyraga: I think `workMutex` has nothing to do here. This function should only be called when the scheduler is working but most importantly, it should only be called by the scheduler. I added this info in 91fc27a77c9976bf147aa942a2d08dd9940c267d but I did not mention the mutex since I do not follow the comment. Can you explain it, please?

pkg/generator/scheduler.go
pdyraga: https://github.com/keep-network/keep-core/commit/91fc27a77c9976bf147aa942a2d08dd9940c267d

pkg/generator/scheduler.go
lukasz-zimnoch: The `startWorker` function is modifying the `s.stops` slice that is protected by the `s.workMutex`. That said, the `startWorker` call should be made only if the `s.workMutex` is locked. Currently, `startWorker` is called by `resume` and `compute` that already lock the  `workMutex` so `startWorker` cannot do it on their own as that would cause a deadlock. My comment was about mentioning the fact that `startWorker` does not lock the mutex explicitly but has an implicit requirement that it must be called on the locked mutex.

pkg/generator/latch.go
pdyraga: https://github.com/keep-network/keep-core/pull/3183/commits/c2a0003b35115557bee64dea9b8dee0d941fc030

pkg/generator/latch.go
lukasz-zimnoch: The current version works for me as well!

pkg/generator/scheduler.go
pdyraga: Ah, right! 03539e534ec6b10ac9ea899ec820464822c528eb

lukasz-zimnoch:
@pdyraga I'm good to merge but we have conflicts here.

PR #3191, opened 2022-08-23
Fix blocked undelegate button

michalsmiarowski:
There is a bug in Keep dashboard which blocked the user to unstake the
delegation from Keep if he unstaked all the tokens from T dashboard.

It was happening because normally user should not be allowed to unstake
when the stake is moved to T, but we were checking if it was moved there
wrongly.  Previosly we examined the "Staked" events in T contract, and
if there were any we assumed that the staked is moved. However, there
can be a situation when the operator stake the keep tokens, and then
unstake them all and we didn't check any "Unstaked" events at all.

To fix this, instead of checking all the events, we will be calling
"stakes" method from TokenStaking contract, which will return all the
tokens staked in T Network (keep, nu and T). Then we will check if keep
tokens are equal to 0 - if it's true, then the T stake is not there
anymore and user can safely unstake his Keeps from Keep dahsboard.

We don't have to check if other tokens are equal to 0, because they have
nothing to do with the stked Keep tokens.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -12,6 +12,7 @@ ca823db2fc25dcf318df84d2b3efc74347aeb552
 e884d097787c1fd3f541088717c946f53503ea79
 e0880e124c2dcefa6a6fbb4e16ef4b8fe2ab861a
 bf037460a86573b61f62f9acec46f1688d44d409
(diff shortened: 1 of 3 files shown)


r-czajkowski (approved):
🚢

PR #3193, opened 2022-08-23
[Backport] Remove specific stakes from t auth data

michalsmiarowski:
Backport of: https://github.com/keep-network/keep-core/pull/3023

Removes stakes that are cancelled (while in initialization period) from
threshold authorization table. Also disable the `Authorize and Stake` button for
the stakes that are in initialization period (and add tooltip informing about
why it is disabled).


r-czajkowski (commented):
👍

r-czajkowski (approved):
:shipit:

PR #3196, opened 2022-08-24
Improvements for the beacon DKG result submitter

lukasz-zimnoch:
Closing the `onSubmittedResultChan` channel is not needed in that case and is the source of the frequently observed test failure:
```
panic: send on closed channel

goroutine 49841 [running]:
github.com/keep-network/keep-core/pkg/beacon/dkg/result.(*SubmittingMember).SubmitDKGResult.func1(0xa27720?)
	/go/src/github.com/keep-network/keep-core/pkg/beacon/dkg/result/submission.go:79 +0x26
github.com/keep-network/keep-core/pkg/chain/local_v1.(*localChain).SubmitDKGResult.func1(0xc00000e300?, 0x0?)
	/go/src/github.com/keep-network/keep-core/pkg/chain/local_v1/local.go:290 +0x23
created by github.com/keep-network/keep-core/pkg/chain/local_v1.(*localChain).SubmitDKGResult
	/go/src/github.com/keep-network/keep-core/pkg/chain/local_v1/local.go:289 +0x41e
```
It is ok to leave Go channels open as they will be garbage collected as soon as they become unreachable.

Regarding the result submission event subscription, it is enough to defer its unsubscription and do it manually only when the given member is about to submit the result by itself. The latter must be done in order to avoid a resource leak that is explained by the comment placed in the code.

### DIFF
--- a/pkg/beacon/dkg/result/submission.go
+++ b/pkg/beacon/dkg/result/submission.go
@@ -84,7 +84,7 @@ func (sm *SubmittingMember) SubmitDKGResult(
 	alreadySubmitted, err := chainRelay.IsGroupRegistered(result.GroupPublicKey)
 	if err != nil {
 		return fmt.Errorf(
-			"could not check if the result is already submitted: [%v]",
+			"could not check if the result is already submitted: [%w]",
 			err,
 		)
 	}
@@ -101,7 +101,7 @@ func (sm *SubmittingMember) SubmitDKGResult(
 		config.ResultPublicationBlockStep,
 	)
 	if err != nil {
-		return fmt.Errorf("wait for eligibility failure: [%v]", err)
+		return fmt.Errorf("wait for eligibility failure: [%w]", err)
 	}
 
 	for {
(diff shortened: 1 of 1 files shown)


pkg/beacon/dkg/result/submission.go
```diff
@@ -79,26 +79,19 @@ func (sm *SubmittingMember) SubmitDKGResult(
...
 
 	alreadySubmitted, err := chainRelay.IsGroupRegistered(result.GroupPublicKey)
 	if err != nil {
-		return returnWithError(
-			fmt.Errorf(
-				"could not check if the result is already submitted: [%v]",
-				err,
-			),
+		return fmt.Errorf(
+			"could not check if the result is already submitted: [%v]",
```
tomaszslabon, line 87: Maybe we could use [%w] instead of [%v] for nested errors in the whole file?

lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3196/commits/be83c97dfc2c654a46247c322026ef0dce0d9013

tomaszslabon (approved):

PR #3200, opened 2022-08-25
Embed predefined bootstrap peers list into the client

nkuba:
Depends on https://github.com/keep-network/keep-common/pull/99

The client requires a list of network peers to connect to. The list had to be configured by a user under the `ethereum.network.peers` property. To simplify running the node we will embed into the client the list of bootstrap peers supporting a given Ethereum network. 

When running the client a user can use one of the flags: `--mainnet`, `--goerli` or `--developer` to specify against which Ethereum network the client should be started. If non of the aforementioned flags is set, the client will assume `--mainnet`

If the list of peers in the config file under `ethereum.network.peers` is empty the client will resolve the default peers that were set in the client build.

The bootstrap peers configuration is expected to be provided in the `config/_peers/` directory in a separate file for each network.

If the client is run with `--developer` flag, the peers are expected to be set in the config file, they won't be resolved automagically.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -8,16 +8,17 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/ipfs/go-log"
+	"github.com/mitchellh/mapstructure"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"golang.org/x/crypto/ssh/terminal"
+
 	commonEthereum "github.com/keep-network/keep-common/pkg/chain/ethereum"
 	"github.com/keep-network/keep-core/pkg/beacon/registry"
 	"github.com/keep-network/keep-core/pkg/diagnostics"
 	"github.com/keep-network/keep-core/pkg/metrics"
 	"github.com/keep-network/keep-core/pkg/net/libp2p"
 	"github.com/keep-network/keep-core/pkg/tbtc"
-	"github.com/mitchellh/mapstructure"
-	"github.com/spf13/pflag"
-	"github.com/spf13/viper"
-	"golang.org/x/crypto/ssh/terminal"
 )
 
 var logger = log.Logger("keep-config")
--- a/config/config.go
+++ b/config/config.go
@@ -62,7 +62,7 @@ func (c *Config) resolveEthereumNetwork(flagSet *pflag.FlagSet) error {
 			return commonEthereum.Unknown, err
 		}
 		if isGoerli {
-			return commonEthereum.Goerli, err
+			return commonEthereum.Goerli, nil
 		}
 
 		isDeveloper, err := flagSet.GetBool(commonEthereum.Developer.String())
@@ -70,7 +70,7 @@ func (c *Config) resolveEthereumNetwork(flagSet *pflag.FlagSet) error {
 			return commonEthereum.Unknown, err
 		}
 		if isDeveloper {
-			return commonEthereum.Developer, err
+			return commonEthereum.Developer, nil
 		}
 
 		return commonEthereum.Mainnet, nil
--- a/config/config.go
+++ b/config/config.go
@@ -56,10 +56,19 @@ func (c *Config) resolveEthereumNetwork(flagSet *pflag.FlagSet) error {
 	var err error
 
 	c.Ethereum.Network, err = func() (commonEthereum.Network, error) {
-		if ok, err := flagSet.GetBool(commonEthereum.Goerli.String()); ok {
+		isGoerli, err := flagSet.GetBool(commonEthereum.Goerli.String())
+		if err != nil {
+			return commonEthereum.Unknown, err
+		}
+		if isGoerli {
 			return commonEthereum.Goerli, err
 		}
-		if ok, err := flagSet.GetBool(commonEthereum.Developer.String()); ok {
+
+		isDeveloper, err := flagSet.GetBool(commonEthereum.Developer.String())
+		if err != nil {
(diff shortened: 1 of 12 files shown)


dimpar:
For future us testing on `Goerli` network:

- build a client: `make goerli`
- run a client:
```
KEEP_ETHEREUM_PASSWORD=$(cat ./<path_to_password_file>) go run . start --ethereum.url wss://<endpoint> --ethereum.keyFile ~/ethereum/data/keystoreGoerli/keyfile.json --storage.dataDir /Users/dp/.keep/keep-core/storage/client-v2 --goerli
```

Resulted in connected client:

```
libp2p/libp2p.go:226	number of connected peers: [2]
```

config/config.go
```diff
@@ -53,6 +51,24 @@ func bindFlags(flagSet *pflag.FlagSet) error {
 	return nil
 }
 
+// Resolve ethereum network based on the set boolean flags.
+func (c *Config) resolveEthereumNetwork(flagSet *pflag.FlagSet) error {
+	var err error
+
+	c.Ethereum.Network, err = func() (commonEthereum.Network, error) {
+		if ok, err := flagSet.GetBool(commonEthereum.Goerli.String()); ok {
+			return commonEthereum.Goerli, err
```
dimpar, line 60: I think if we're inside of this `if` then the `err` is nil. What happens when the error occurs? Do we panic?
Same comment for the `if` below

config/peers.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, err
+	}
+
+	return cleanStrings(strings.Split(string(peers), "\n")), nil
+}
+
+// cleanStrings iterates over entires in a slice and trims spaces from the beginning
+// and the end of a string. It also removes empty entries or entries commented with `#`.
+func cleanStrings(s []string) []string {
+	var r []string
```
dimpar, line 27: r -> peers ?

dimpar:
Some general thoughts about this approach.. What is kind of itchy is that we couple concrete networks into our codebase. Imagine we would do that a couple of months ago when testing on Ropsten, it's gone now. Same with other testing networks. My point is that it might happen to Goerli as well and then we would need to clean this code. Even if we add or remove a network then we need to change code at least in a couple of places (which I guess is not the end of the world, but some additional thing to remember about). I'm not 100% sure what would be the optimal approach here yet, maybe coding against some abstract networks that support default peers and networks that do not. These networks list might live in a config file with their respective chain ids. Or some other solution with a "one stop for networks list" and code against that list.

dimpar:
Don't we need to update `/scripts/start.sh` and pass a `--developer` flag?

config/config.go
nkuba: 36508650cb2a80b8c4b82440a277adc2beadaaf7

config/peers.go
nkuba: 71489c1dec7f809a588e3356a25250becc6ba7c7

nkuba:
> Don't we need to update `/scripts/start.sh` and pass a `--developer` flag?

Good catch! ef5dec08ea364d7422de8246c2db0c8fbfeb33f7

nkuba:
> Some general thoughts about this approach.. What is kind of itchy is that we couple concrete networks into our codebase. Imagine we would do that a couple of months ago when testing on Ropsten, it's gone now. Same with other testing networks. My point is that it might happen to Goerli as well and then we would need to clean this code. Even if we add or remove a network then we need to change code at least in a couple of places (which I guess is not the end of the world, but some additional thing to remember about). I'm not 100% sure what would be the optimal approach here yet, maybe coding against some abstract networks that support default peers and networks that do not. These networks list might live in a config file with their respective chain ids. Or some other solution with a "one stop for networks list" and code against that list.
> 
> Having said that, and if we agree on the above, then I'd start with some smaller steps and merge this PR first. Then we can think about refactoring.

Good point, I'm not sure right now how we could approach it. 

> Some general thoughts about this approach.. What is kind of itchy is that we couple concrete networks into our codebase. Imagine we would do that a couple of months ago when testing on Ropsten, it's gone now. Same with other testing networks. My point is that it might happen to Goerli as well and then we would need to clean this code. Even if we add or remove a network then we need to change code at least in a couple of places (which I guess is not the end of the world, but some additional thing to remember about). I'm not 100% sure what would be the optimal approach here yet, maybe coding against some abstract networks that support default peers and networks that do not. These networks list might live in a config file with their respective chain ids. Or some other solution with a "one stop for networks list" and code against that list.
> 
> Having said that, and if we agree on the above, then I'd start with some smaller steps and merge this PR first. Then we can think about refactoring.

I like this point, but I have no strong idea how we could solve it right away. I created a ticket to track that and handle it in the future https://github.com/keep-network/keep-core/issues/3210.

config/config.go
```diff
@@ -53,6 +52,33 @@ func bindFlags(flagSet *pflag.FlagSet) error {
...
+		if isGoerli {
+			return commonEthereum.Goerli, err
+		}
+
+		isDeveloper, err := flagSet.GetBool(commonEthereum.Developer.String())
+		if err != nil {
+			return commonEthereum.Unknown, err
+		}
+		if isDeveloper {
+			return commonEthereum.Developer, err
```
dimpar, line 73: I think the common pattern would be returning `nil`
```
return commonEthereum.Developer, nil
```
Same for Goerli network

nkuba: 🤦🏻 1b5eaebb4ba5f9117d35765e64c0bd49b822012d

PR #3204, opened 2022-08-25
Validate Chain ID on ethereum connection establishement

nkuba:
We expect the client to be run for a network specified with boolean
flags like `--mainnet` and `--goerli`. Another setting that is required
is an URL to the Ethereum API.
Here we want to ensure that the API URL works with the network that the
client is expected to be run against. We use Chain ID for validation.

The exception is `--developer` network, that we won't validate the Chain
ID as it can vary.

Depends on https://github.com/keep-network/keep-common/pull/100
Depends on https://github.com/keep-network/keep-core/pull/3200


pkg/chain/ethereum/ethereum.go
```diff
@@ -130,6 +130,18 @@ func newBaseChain(
 		)
 	}
 
+	if config.Network != ethereum.Developer &&
+		big.NewInt(config.Network.ChainID()).Cmp(chainID) != 0 {
+		return nil, fmt.Errorf(
```
dimpar, line 135:
```suggestion
		return nil, fmt.Errorf(
			"chain id returned from ethereum api [%s] "+
				"doesn't match the expected chain id [%d] for [%s] network; "+
				"please verify the configured ethereum.url",
			chainID.String(),
			config.Network.ChainID(),
			config.Network,
		)
```

pkg/chain/ethereum/ethereum.go
```diff
@@ -130,6 +130,18 @@ func newBaseChain(
 		)
 	}
 
+	if config.Network != ethereum.Developer &&
```
dimpar, line 133: Don't we need to check against `Unknown` network as well?

nkuba: We don't expect to receive and handle the `Unknown` network anywhere. The way the `resolveEthereumNetwork` works it will always fallback to `Mainnet`.
https://github.com/keep-network/keep-core/blob/6a8e601861ad935a6cc5b1ae381d62919b38a042/config/config.go#L55-L70

pkg/chain/ethereum/ethereum.go
nkuba: 3d8a719dba42908cd1186baee11dbe4bb91d18c0

pkg/chain/ethereum/ethereum.go
dimpar: This functions has changed and we do handle the `Unknown` network https://github.com/keep-network/keep-core/blob/main/config/config.go#L56-L80

nkuba: Nice catch! Yeah, but if we receive the `Unknown` network here clearly something is wrong with the configuration, so the client should fail to start anyway. I don't think we should handle the `Unknown` network case, if the client haven't failed up to this point it should fail now.

dimpar:
Run the client and I didn't notice any errors. lgtm.

dimpar (approved):

PR #3207, opened 2022-08-26
Group size of 100 performance fixes

pdyraga:
During the local testing, we discovered that generating tECDSA key with a group size of 100 is challenging and usually does not complete.

In #3161 we focused on improving TSS pre-param generation. Now, TSS pre-params are generated only when no protocol is executing at the client but it turned out it is not enough to support a group of 100 members. During the profiling, we discovered DLN proof verification in phase 2 of TSS key-generation protocol is the bottleneck.

https://github.com/bnb-chain/tss-lib/pull/203 presents a performance fix to DLN proof verification throttling the number of concurrent goroutines trying to perform the verification at the same time. This change, and increasing the time for the phase should allow the DKG to complete.

This PR updates `bnb-chain/tss-lib` dependency to the feature branch in https://github.com/bnb-chain/tss-lib/pull/203. It is also adding a new configuration option allowing to control key generation concurrency level. By default, DLN proof verification uses all cores available on the machine (`runtime.GOMAXPROCS(0)`). On a 10-core machine, when running 2 clients, it is recommended to limit the concurrency level to 4 for each client. On a 10-core machine, when running 4 clients, it is recommended to limit the concurrency level to 2 for each client.

I would appreciate local testing from every client developer to confirm the values are correct.


docs/resources/client-start-help
```diff
@@ -25,10 +25,11 @@ Flags:
       --metrics.networkMetricsTick duration        Network metrics check tick in seconds. (default 1m0s)
       --metrics.ethereumMetricsTick duration       Ethereum metrics check tick in seconds. (default 10m0s)
       --diagnostics.port int                       Diagnostics HTTP server listening port. (default 8081)
-      --tbtc.preParamsPoolSize int                 tECDSA pre-parameters pool size. (default 50)
+      --tbtc.preParamsPoolSize int                 tECDSA pre-parameters pool size. (default 3000)
       --tbtc.preParamsGenerationTimeout duration   tECDSA pre-parameters generation timeout. (default 2m0s)
       --tbtc.preParamsGenerationDelay duration     tECDSA pre-parameters generation delay. (default 10s)
       --tbtc.preParamsGenerationConcurrency int    tECDSA pre-parameters generation concurrency. (default 1)
+      --tbtc.keyGenerationConcurrency int          tECDSA key generation concurrency. (default 10)
```
lukasz-zimnoch, line 32: I think the default here is the number of cores.

pkg/generator/pool_test.go
```diff
@@ -24,6 +23,7 @@ func TestGetNow(t *testing.T) {
 		if pool.CurrentSize() == 5 {
 			break
 		}
+		runtime.Gosched()
```
lukasz-zimnoch, line 26: Can we put a short explanation of why we need to do that?

pkg/tbtc/tbtc.go
```diff
@@ -26,6 +27,9 @@ const (
 	DefaultPreParamsGenerationConcurrency = 1
 )
 
+// TODO: Can we move it to constants?
```
lukasz-zimnoch, line 30: If that depends on the runtime we can't. Can we remove that TODO in that case?

pdyraga: Yeah, I missed removing this comment. 5b019452b5470e12dfc18ebf11e9e6606489d19b

pkg/generator/pool_test.go
pdyraga: 98122675e077323612fb8120cd3b360e1398d6db

docs/resources/client-start-help
pdyraga: Yeah. I updated it in https://github.com/keep-network/keep-core/pull/3207/commits/003c8ef28ff5c83e2358e44cb1d0bcea7ad768f4. The problem with this file though is that it's auto-generated when running `make`. And this value will go back to 10 if someone regenerates it on a 10-core machine.

I do not want to solve it now and block this work but that's something we'll have to look at in the future cc @nkuba

lukasz-zimnoch (approved):

Issue #3042, opened 2022-06-28
ECDSA signing test loop


pdyraga:
Depends on #3014

For the first 50 blocks after the new group creation, the client should keep signing random data (e.g. new Ethereum block hashes) with the newly created group. This should include the retry algorithm from #3014.

### Roadmap:

- [x] Prepare the ground for signing implementation (#3201, #3209)
- [x] Test loop and signing executor setup (#3219)
- [x] Actual signing protocol implementation that consists of:
    - [x] Ephemeral key exchange and symmetric key generation rounds (#3226)
    - [x] TSS round 1 (#3228)
    - [x] TSS round 2 (#3239)
    - [x] TSS round 3 (#3241)
    - [x] TSS round 4 (#3242)
    - [x] TSS round 5 (#3243)
    - [x] TSS round 6 (#3246)
    - [x] TSS round 7 (#3248)
    - [x] TSS round 8 (#3249)
    - [x] TSS round 9 (#3250)
    - [x] TSS finalization round (#3251)
- [x] Signing process using a group that was formed during a DKG with misbehaved members (#3254)
- [x] Phases duration tweaks that will allow making a successful signing with 100 members on 2 distinct clients (#3259)
- [x] Integration with the retry algorithm (#3260)
- [x] Improve logging by adding the context of the given signing process (#3267)
- [x] PR that will address all non-blocking comments from previous PRs (#3271)

lukasz-zimnoch:
The thing to consider: Reusing symmetric keys generated using the ECDH algorithm upon key generation for securing signing P2P messages. It makes sense at the first sight though we need to think about the security of that.

pdyraga:
Another thing to consider: [renaming `group` package](https://github.com/keep-network/keep-core/pull/3093#discussion_r929976372) to clearly indicating the components are used in DKG.

PR #3209, opened 2022-08-26
tECDSA signing: Cleanup `node.go` 

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-core/pull/3201

Since we want to implement signing as the next feature of `tbtc.node`, we need to make cleanup of the `pkg/tbtc/node.go` file. We do it here by extracting all DKG components to a separate file `pkg/tbtc/dkg.go`. Worth noting that the `tbtc.node.joinDKGIfEligible` function remains unchanged and is still responsible for the orchestration of the aforementioned components. I made some experiments trying to split it as well but that yields an additional level of indirection resulting with a less readable code. 

By the way, this PR addresses two TODOs from #3120 regarding the `GroupQuorum` parameter: https://github.com/keep-network/keep-core/pull/3209/commits/83eb9f02bd2056a208e8e028cb16d7808cccabf6


tomaszslabon (approved):

tomaszslabon:
Tested it manually and it worked as expected.

PR #3212, opened 2022-08-29
Provide a cancel signal for the parameter pool getter

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3161

The `Get` function exposed by the `generator.ParameterPool` type can hang on the `<-pp.pool` instruction in case the parameter pool is empty. Although this behavior is proper if the pool is working, it may cause deadlocks if the pool is stopped by the scheduler who is waiting for resume signals from clients which are in turn waiting for parameters that are supposed to be provided by the empty pool. Such a scenario occurred for tECDSA DKG. Here we aim to fix it by renaming `Get` to `GetNow` function which now uses a "get or fail" logic that provides an immediate result of a `GetNow` call without blocking.

### DIFF
--- a/pkg/generator/pool.go
+++ b/pkg/generator/pool.go
@@ -90,13 +90,21 @@ func NewParameterPool[T any](
 
 // Get returns a new parameter. It is fetched from the pool or generated if the
 // pool is empty.
-func (pp *ParameterPool[T]) Get() (*T, error) {
-	generated := <-pp.pool
-	err := pp.persistence.Delete(generated)
-	if err != nil {
-		return nil, fmt.Errorf("could not delete persisted parameter: [%w]", err)
+func (pp *ParameterPool[T]) Get(ctx context.Context) (*T, error) {
+	select {
+	case generated := <-pp.pool:
+		err := pp.persistence.Delete(generated)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"could not delete persisted parameter: [%w]",
+				err,
+			)
+		}
+
+		return generated, nil
+	case <-ctx.Done():
+		return nil, fmt.Errorf("context is done")
 	}
-	return generated, nil
 }
 
 // CurrentSize returns the current size of the pool - the number of available
--- a/pkg/generator/pool.go
+++ b/pkg/generator/pool.go
@@ -88,9 +88,9 @@ func NewParameterPool[T any](
 	}
 }
 
-// Get returns a new parameter. It is fetched from the pool or generated if the
-// pool is empty.
-func (pp *ParameterPool[T]) Get(ctx context.Context) (*T, error) {
+// GetNow returns a new parameter. It is fetched from the pool or an error is
+// returned when the pool is empty.
+func (pp *ParameterPool[T]) GetNow() (*T, error) {
 	select {
 	case generated := <-pp.pool:
 		err := pp.persistence.Delete(generated)
@@ -102,8 +102,8 @@ func (pp *ParameterPool[T]) Get(ctx context.Context) (*T, error) {
 		}
 
 		return generated, nil
-	case <-ctx.Done():
-		return nil, fmt.Errorf("context is done")
+	default:
+		return nil, fmt.Errorf("pool is empty")
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/generator/pool.go
```diff
@@ -90,13 +90,21 @@ func NewParameterPool[T any](
...
+		if err != nil {
+			return nil, fmt.Errorf(
+				"could not delete persisted parameter: [%w]",
+				err,
+			)
+		}
+
+		return generated, nil
+	case <-ctx.Done():
+		return nil, fmt.Errorf("context is done")
```
pdyraga, line 106: Wouldn't it be clearer to do something like:
```
func (pp *ParameterPool[T]) GetNow() (*T, error) {
	select {
	case generated := <-pp.pool:
		err := pp.persistence.Delete(generated)
		if err != nil {
			return nil, fmt.Errorf(
				"could not delete persisted parameter: [%w]",
				err,
			)
		}

		return generated, nil
	default:
		return nil, fmt.Errorf("pool is empty")
	}
}
```

In `Execute` function, we do not want to wait for the pool even one minute. The scheduler stopped the generation process so there is no chance something will appear in the channel. I think this will also work better once we move getting the parameters to phase 2 of TSS: we will check `pool.CurrentSize()` before `Execute()` call and later, we'll use `GetNow()`.

lukasz-zimnoch: Makes sense, see https://github.com/keep-network/keep-core/pull/3212/commits/ea027cfd85072742f350792d5dfbc1d998c74cac

pkg/generator/pool_test.go
```diff
@@ -11,17 +13,21 @@ import (
...
+func TestGetNow(t *testing.T) {
 	pool, scheduler, _ := newTestPool(5)
 	defer scheduler.stop()
 
-	for i := 0; i < 70; i++ {
-		e, err := pool.Get()
+	for {
+		if pool.CurrentSize() == 5 {
+			break
+		}
```
pdyraga, line 26:
```suggestion
		if pool.CurrentSize() == 5 {
			break
		}
		runtime.Gosched()
```

I would yield the processor on each loop execution to make sure we are not stuck too long in the for and the generating goroutine does not get starved.

pkg/generator/pool_test.go
```diff
@@ -31,6 +37,33 @@ func TestGet(t *testing.T) {
...
+		func(ctx context.Context) *big.Int {
+			<-ctx.Done()
+			return nil
+		},
+	)
+	defer scheduler.stop()
+
+	_, err := pool.GetNow()
+
+	expectedErr := fmt.Errorf("pool is empty")
```
pdyraga, line 55: Let's define `ErrPoolEmpty` in `pool.go` and use `testutils.AssertErrorsSame` here. The test should be cleaner but more importantly, we should be able to inspect the reason of a failure of getting parameters from the pool, if needed.

pkg/tecdsa/dkg/dkg.go
```diff
@@ -64,7 +64,7 @@ func (e *Executor) Execute(
 
 	registerUnmarshallers(channel)
 
-	preParams, err := e.tssPreParamsPool.Get()
+	preParams, err := e.tssPreParamsPool.GetNow()
 	if err != nil {
 		return nil, 0, fmt.Errorf("failed fetching pre-params: [%v]", err)
```
pdyraga, line 69: We could also change it to `%w` so that we can unwrap `ErrPoolEmpty` and differentiate if it's an empty pool or e.g. persistence problem if needed.

pdyraga (approved):
I left some small comments I do not consider blocking. Let me address them in #3207.

pkg/generator/pool_test.go
pdyraga: https://github.com/keep-network/keep-core/pull/3207/commits/f44ebc38ff7291cb47df1e3b081337065c840d11

pkg/tecdsa/dkg/dkg.go
pdyraga: https://github.com/keep-network/keep-core/pull/3207/commits/28eb1c7814703a137bed34e428f752b188dd2c39

PR #3211, opened 2022-08-29
Iniitial documentation for V2 client

nkuba:
This PR contains initial documentation for the V2 client. It is based on the existing documentation for the V1.

Please see the documentation preview under https://docs.keep.network/docs-v2/index.html

PR #3214, opened 2022-08-29
Add comment to PR-s modifying docs that links to the published preview

michalinacienciala:
We're adding a step to the `docs.yml` workflow that publishes the comment under
the PR modifying the documents, after preview of the dosc gets published to the
GCP bucket. The comment contains a link to the `index.html` page containing the
table of contents.
Note: we need to have `./docs/index.adoc` file with right config inside to have
it working in the way I described it.

Ref: https://github.com/keep-network/keep-core/pull/3211

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -72,5 +72,5 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: 'Preview uploaded to https://docs.keep.network/${{ github.head_ref }}/index.html.'
+              body: 'Documentation preview uploaded to https://docs.keep.network/${{ github.head_ref }}/index.html.'
             })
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -62,3 +62,15 @@ jobs:
...
+      - name: Post preview URL to PR
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@v5
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Preview uploaded to https://docs.keep.network/${{ github.head_ref }}/index.html.'
```
nkuba, line 75:
```suggestion
              body: 'Documentation preview uploaded to https://docs.keep.network/${{ github.head_ref }}/index.html.'
```

michalinacienciala: [8c0965a](https://github.com/keep-network/keep-core/pull/3214/commits/8c0965a789f340145327c27a3faf7418bdadcc57)

nkuba (approved):

PR #3215, opened 2022-08-29
Updates to the CLI commands and flags

nkuba:
With the current implementation, the `ethereum` commands fails with: `Error: unknown flag: --ethereum.url`.

We add `TraverseChildren` flag to the main command to let each of the subcommands define their flags.

The change mentioned above was not enough to make it work as the way we were reading the config with `config.ReadEthereumConfig` required flags to be provided, but the flags were defined on the parent `ethereum` command, which wasn't accessible in the context of the contract functions commands.

In https://github.com/keep-network/keep-common/pull/101 we modify the way we chain handle is initialized in the generated ethereum commands, and in this PR we align with that.

We defined a wrapper for the `cobra.Command` that also holds the Ethereum config that is used by the generated commands bindings.

Depends on https://github.com/keep-network/keep-common/pull/101

We also took a chance to modify how we initialize the `RootCmd` and move all command-related stuff to the `cmd` package.

The `RootCommand` is initialized with the persistent (global) flags that can be configured on any command in the command chain, so each of the examples will work:
```
keep-core --config configs/config.1.toml --developer start
keep-core start --config configs/config.1.toml --developer
keep-core --config configs/config.1.toml start --developer
```

### DIFF
--- a/pkg/chain/ethereum/beacon/gen/cmd/cmd.go
+++ b/pkg/chain/ethereum/beacon/gen/cmd/cmd.go
@@ -2,15 +2,19 @@ package cmd
 
 import (
 	"github.com/spf13/cobra"
+
+	"github.com/keep-network/keep-core/pkg/chain/ethereum"
 )
 
-// Command is the exported list of generated commands that can be
+// ModuleCommand is the exported list of generated commands that can be
 // installed on a CLI app. Generated contract command files set up init
 // functions that add the contract's command and subcommands to this global
 // variable, and any top-level command that wishes to include these commands can
 // reference this variable and expect it to contain all generated contract
 // commands.
-var Command = &cobra.Command{
-	Use:   "beacon",
-	Short: `Provides access to Keep Random Beacon contracts.`,
+var ModuleCommand = ethereum.Command{
+	Command: cobra.Command{
+		Use:   "beacon",
+		Short: `Provides access to Keep Random Beacon contracts.`,
+	},
 }
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1,7 +1,13 @@
 package cmd
 
 import (
+	"fmt"
+	"os"
+	"path"
+
 	"github.com/ipfs/go-log"
+	"github.com/spf13/cobra"
+
 	"github.com/keep-network/keep-core/config"
 )
 
@@ -13,3 +19,28 @@ var configFilePath string
 var clientConfig = &config.Config{}
 
 var logger = log.Logger("keep-cmd")
+
+// RootCmd contains the definition of the root command-line command.
+var RootCmd = &cobra.Command{
+	Use:              path.Base(os.Args[0]),
+	Short:            "CLI for The Keep Network",
+	Long:             "Command line interface (CLI) for running a Keep provider",
+	TraverseChildren: true,
+}
+
+func init() {
+	initGlobalFlags(RootCmd, &configFilePath)
+
+	RootCmd.AddCommand(
+		StartCommand,
+		PingCommand,
+		EthereumCommand,
+	)
+}
+
+// Initialize initializes the root command and returns it.
+func Initialize(version, revision string) *cobra.Command {
+	RootCmd.Version = fmt.Sprintf("%s (revision %s)", version, revision)
+
+	return RootCmd
+}
--- a/cmd/ethereum.go
+++ b/cmd/ethereum.go
@@ -1,12 +1,14 @@
 package cmd
 
 import (
+	"github.com/spf13/cobra"
+
 	"github.com/keep-network/keep-core/config"
+
 	beaconcmd "github.com/keep-network/keep-core/pkg/chain/ethereum/beacon/gen/cmd"
 	ecdsacmd "github.com/keep-network/keep-core/pkg/chain/ethereum/ecdsa/gen/cmd"
 	tbtccmd "github.com/keep-network/keep-core/pkg/chain/ethereum/tbtc/gen/cmd"
 	thresholdcmd "github.com/keep-network/keep-core/pkg/chain/ethereum/threshold/gen/cmd"
-	"github.com/spf13/cobra"
 )
 
 // EthereumCommand contains the definition of the ethereum command-line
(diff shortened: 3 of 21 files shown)


dimpar:
When running `/scripts/start.sh` I get the following error:
```
2022-08-30T12:06:47.592+0200	FATAL	keep-main	keep-core/main.go:51	unknown flag: --config
```

nkuba:
> When running `/scripts/start.sh` or manually `./keep-client --config /Users/dp/go/src/github.com/keep-network/keep-core/configs/config.toml` shows the following error:
> 
> ```
> 2022-08-30T12:06:47.592+0200	FATAL	keep-main	keep-core/main.go:51	unknown flag: --config
> ```

Have you tried to rebuild the client? Try running:
```
go run . start --config /Users/dp/go/src/github.com/keep-network/keep-core/configs/config.toml
```

dimpar:
> ./keep-client --config /Users/dp/go/src/github.com/keep-network/keep-core/configs/config.toml



> > When running `/scripts/start.sh` or manually `./keep-client --config /Users/dp/go/src/github.com/keep-network/keep-core/configs/config.toml` shows the following error:
> > ```
> > 2022-08-30T12:06:47.592+0200	FATAL	keep-main	keep-core/main.go:51	unknown flag: --config
> > ```
> 
> Have you tried to rebuild the client? Try running:
> 
> ```
> go run . start --config /Users/dp/go/src/github.com/keep-network/keep-core/configs/config.toml
> ```

I've rebuilt the client and tried again. Same error. However, `go run . start --config /Users/dp/go/src/github.com/keep-network/keep-core/configs/config.toml` works

pkg/chain/ethereum/beacon/gen/cmd/cmd.go
```diff
@@ -2,15 +2,19 @@ package cmd
...
 // variable, and any top-level command that wishes to include these commands can
 // reference this variable and expect it to contain all generated contract
 // commands.
-var Command = &cobra.Command{
-	Use:   "beacon",
-	Short: `Provides access to Keep Random Beacon contracts.`,
+var ModuleCommand = ethereum.Command{
+	Command: cobra.Command{
+		Use:   "beacon",
+		Short: `Provides access to Keep Random Beacon contracts.`,
```
dimpar, line 18: I'm not 100% sure, but didn't we stop using "Keep" in favor of "T"? Keep Network -> Threshold Network? Same comment for other places where we use "Keep"

nkuba:
You need to use `start` command:
```
./keep-client start --config /Users/dp/go/src/github.com/keep-network/keep-core/configs/config.toml
```

dimpar:
> You need to use `start` command:
> 
> ```
> ./keep-client start --config /Users/dp/go/src/github.com/keep-network/keep-core/configs/config.toml
> ```

ah.. yea, this works. Can you pls update `start.sh` as well?

dimpar:
Tested locally with the deployed contracts. I was able to run a client `./keep-client start --config /Users/dp/go/src/github.com/keep-network/keep-core/configs/config.toml --developer`

Also tested a couple of ethereum configs to make sure we can retrieve the necessary information like so: 
```
./keep-client ethereum --ethereum.url ws://localhost:8546 --ethereum.keyFile /Users/dp/ethereum/data/keystore/UTC--2019-08-27T11-40-44.016554000Z--407190f04d838ec47dad4c0e0d2a9c49d7737479  beacon random-beacon staking
```

```
The ethereum command allows interacting with Keep's Ethereum
contracts directly. Each subcommand corresponds to one contract, and has subcommands
corresponding to each method on that contract, which respectively each take parameters
based on the contract method's parameters.

See the subcommand help for additional details.

Usage:
  keep-client ethereum [command]

Available Commands:
  beacon      Provides access to Keep Random Beacon contracts.
  ecdsa       Provides access to Keep ECDSA contracts.
  tbtc        Provides access to Keep TBTCv2 contracts.
  threshold   Provides access to Threshold Network contracts.

Flags:
  -c, --config string                           Path to the configuration file. Supported formats: TOML, YAML, JSON.
      --developer                               Developer network
      --goerli                                  Görli network
      --mainnet                                 Mainnet network
      --ethereum.url string                     WS connection URL for Ethereum client.
      --ethereum.keyFile string                 The local filesystem path to Keep operator account keyfile.
      --ethereum.miningCheckInterval duration   The time interval in seconds in which transaction mining status is checked. If the transaction is not mined within this time, the gas price is increased and transaction is resubmitted. (default 1m0s)
      --ethereum.maxGasFeeCap wei               The maximum gas fee the client is willing to pay for the transaction to be mined. If reached, no resubmission attempts are performed. (default 500 gwei)
      --ethereum.requestPerSecondLimit int      Request per second limit for all types of Ethereum client requests. (default 150)
      --ethereum.concurrencyLimit int           The maximum number of concurrent requests which can be executed against Ethereum client. (default 30)
      --ethereum.balanceAlertThreshold wei      The minimum balance of operator account below which client starts reporting errors in logs. (default 500000000 gwei)
  -h, --help                                    help for ethereum
```

pkg/chain/ethereum/beacon/gen/cmd/cmd.go
nkuba: I'm not sure we fully transitioned. I'd keep it as is and if we decide to rename we would do this in a bulk PR.

dimpar:
Can we split "Flags" into "Flags" and "Global Flags" just as we have in the subcommands? Also, the order is a bit strange here, ex --help is between --goerli and --mainnet.

```
./keep-client -h
Command line interface (CLI) for running a Keep provider

Usage:
  keep-client [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  ethereum    Provides access to Keep network Ethereum contracts.
  help        Help about any command
  ping        bidirectional send between two peers to test the network
  start       Starts the Keep Client

Flags:
  -c, --config string   Path to the configuration file. Supported formats: TOML, YAML, JSON.
      --developer       Developer network
      --goerli          Görli network
  -h, --help            help for keep-client
      --mainnet         Mainnet network
  -v, --version         version for keep-client

Use "keep-client [command] --help" for more information about a command.
```

dimpar:
no more comments

nkuba:
The help is autogenerated by the cobra. So not much we can do without modifying the cobra templates. Would you mind creating an issue listing all the improvements you would like to see? We could try to implement them in a separate PR.

> Also, the order is a bit strange here, ex --help is between --goerli and --mainnet.

Unfortunately, global flags ordering is not working properly, they are sorted in the alphabetical order and we cannot change that (see https://github.com/spf13/cobra/issues/1033)

dimpar:
> @dimpar The help is autogenerated by the cobra. So not much we can do without modifying the cobra templates. Would you mind creating an issue listing all the improvements you would like to see? We could try to implement them in a separate PR.
> 
> > Also, the order is a bit strange here, ex --help is between --goerli and --mainnet.
> 
> Unfortunately, global flags ordering is not working properly, they are sorted in the alphabetical order and we cannot change that (see [spf13/cobra#1033](https://github.com/spf13/cobra/issues/1033))

Ok, understand, we can leave it for now. Here's an issue for improvements https://github.com/keep-network/keep-core/issues/3224

dimpar (approved):

PR #3221, opened 2022-08-30
Make Docker build use `make build` command

nkuba:
We use the `make build` command to build the docker image.
The `make build` command is a common command for building the client.

TODO:
- [x] Pass `VERSION` and `REVISION` args to the docker build command in the CI.

Depends on https://github.com/keep-network/keep-core/pull/3220


dimpar (approved):
Tested building a dockerfile image and then running `docker run keepcore --version`. A new version was present. Leaving a final call to you @michalinacienciala

michalinacienciala (approved):

PR #3222, opened 2022-08-30
Build client binary for multiple platforms

nkuba:
We add a build command to build  the client binary for multiple
platforms.

To build the binaries run: `./scripts/build.sh`, and the files will be stored in `out/bin` directory.

Each binary will be archived to `.tar.gz` file and md5 and sha256 checksums will be calculated for the tarball.

Currently, we build binaries for the following platforms:
- linux/amd64
- darwin/amd64

The binaries are build as part of the CI process, the binaries will be packed and md5 and sha256 checksums calculated for the packages. They will be uploaded to a GCP bucket: https://console.cloud.google.com/storage/browser/keep-test-client-binaries

TODO:
- [x] publish files to a GCP bucket

Depends on https://github.com/keep-network/keep-core/pull/3220


Makefile
```diff
@@ -88,17 +90,62 @@ generate: gen_proto
...
-go_build_cmd = go build -ldflags "-X github.com/keep-network/keep-core/build.Version=$(version) -X github.com/keep-network/keep-core/build.Revision=$(revision)" -a -o $(1) .
+	$(call go_build_cmd,out/bin,$(file_name),$(os),$(arch))
+endef
 
 build:
 	$(info Building Go code)
-	$(call go_build_cmd,keep-client)
+	$(call go_build_cmd,.,$(app_name))
+
+platforms := linux/386 \
```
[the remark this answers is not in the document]
nkuba: [42d2dc2](https://github.com/keep-network/keep-core/pull/3222/commits/42d2dc213622199e4df092fa54d76dcba26b5675)

.dockerignore
```diff
@@ -2,9 +2,10 @@
 .*
 
 # Top-level directories unrelated to the build.
```
michalinacienciala, line 4: We could also add `.github/` here, right? Also, what about `infrastructure/`, `scripts/` - could we add them here as well?

.github/workflows/client.yml
```diff
@@ -121,10 +124,15 @@ jobs:
         env:
           IMAGE_NAME: "keep-client"
         with:
+          target: runtime-docker
           # GCR image should be named according to following convention:
           # HOSTNAME/PROJECT-ID/IMAGE:TAG
           # We don't use TAG yet, will be added at later stages of work on RFC-18.
```
michalinacienciala, line 130: This comment can be removed now.

.github/workflows/client.yml
```diff
@@ -134,6 +142,37 @@ jobs:
...
+        with:
+          target: output-bins
+          outputs: type=local,dest=./out/bin/
+          build-args: |
+            ENVIRONMENT=${{ github.event.inputs.environment }}
+            VERSION=${{ env.version }}
+            REVISION=${{ env.revision }}
+          push: false
+
+      - name: Archive Client Binaries
```
michalinacienciala, line 156: Just so I would understand - is this step here just so that we would be able to manually download the binaries from the details of the run if we want to check them out?

.dockerignore
pdyraga: We already have them.

`.github` should be covered by 

https://github.com/keep-network/keep-core/blob/a2ae4270fe2446fa1fabf44398326fbbc88dc4d6/.dockerignore#L1-L2

`infrastructure/` and `scripts/` are here:
https://github.com/keep-network/keep-core/blob/a2ae4270fe2446fa1fabf44398326fbbc88dc4d6/.dockerignore#L6-L7

michalinacienciala: Gosh, I'm blind 🙈

.github/workflows/client.yml
pdyraga: We want to publish binaries when releasing a new version, not just the Docker image. The binaries built in the previous step are compressed and we compute hashes for them. Then, they are published to a bucket. We do not want to build binaries locally, we want the CI to do that for us.

pdyraga (approved):
Tested locally on MBP. md5, sha256 hashes are matching, darwin binary works as expected.

Leaving it in @michalinacienciala's hands.

.github/workflows/client.yml
michalinacienciala: Yeah, I get that, but this particular step (using `actions/upload-artifact@v3` action) is not part of the building of binaries or publishing them to a bucket (unless I'm missing something?). It just archives all the files mentioned in the `path` param to one file called `binaries` and makes the file accessible for download in the GH page with the details of the workflow run. The file could be also used to pass it to some other job in the workflow via the `actions/download-artifact` action, but we don't use it like that in this workflow.
So what I wanted to confirm is that this action is used just to create an archive with the binary and the hashes that can be downloaded via GH UI. Or is there some other purpose that I'm missing?

nkuba: > Just so I would understand - is this step here just so that we would be able to manually download the binaries from the details of the run if we want to check them out?

Yes, this is just for us in case we want to verify the CI process. The actual publication happens in the GCP upload step.

.github/workflows/client.yml
nkuba: aba24071481296aeb4aa3e3547eb905d82f31c95

michalinacienciala (approved):

PR #3219, opened 2022-08-30
tECDSA signing: Setup of the test loop and signing executor

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3042
Depends on: https://github.com/keep-network/keep-core/pull/3209

Here we are setting up the tECDSA signing test loop and the tECDSA signing protocol executor, without executing the actual signing logic yet. Specific changes are summarized below.

### Test loop setup

The signing loop was set up using a temporary chain mock that generates a new `tbtc.SignatureRequestedEvent` periodically targeting the recently created wallet. The signature request is about signing the hash of the block the event was generated at. The `tbtc` application node listens for the aforementioned events and decides about joining the signing process if the node controls at least one of the requested wallet's signers. 

Worth mentioning that this trigger of the signing process is temporary, and ultimately, the signing process will not be initiated explicitly by the chain. Instead of that, the `tbtc` node will observe domain events emitted by the `Bridge` contract` and react to them by constructing and signing specific BTC transactions or other data required by the given use case.

### Signing protocol executor setup

Once the `tbtc` node decides to join the given signing execution, it must set up the tECDSA signing protocol executor properly. This includes setting up the right broadcast channel and determining all the input parameters required by the tECDSA signing executor. This PR addresses those requirements within the `joinSigningIfEligible` function.

Additionally, we are also exposing the executor function itself.  The function is called `pkg/tecdsa/signing.Execute` and this PR focuses on defining its public API. The actual logic is just a stub at the moment and will be addressed in further PRs.

### Next steps

This PR is just a part of the entire work regarding the tECDSA signing test loop. Please refer to the description of https://github.com/keep-network/keep-core/issues/3042 for the full roadmap and the current status of the work.

PR #3226, opened 2022-08-31
tECDSA signing: Ephemeral key exchange and symmetric keys generation

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-core/pull/3219

Here we introduce the two first phases of the tECDSA signing protocol which are about exchanging ephemeral keys and generating symmetric keys. Symmetric keys are meant to secure P2P communication in later phases of the protocol. Basically, the code was ported from the sibling `pkg/tecdsa/dkg` package that also uses those two phases at the beginning.

### Next steps

This PR is just a part of the entire work regarding the tECDSA signing test loop. Please refer to the description of https://github.com/keep-network/keep-core/issues/3042 for the full roadmap and the current status of the work. At this point, the next step will be introducing the first round of TSS signing.


pkg/tecdsa/signing/marshaling.go
```diff
@@ -0,0 +1,91 @@
...
+	ephemeralPublicKeys, err := marshalPublicKeyMap(epkm.ephemeralPublicKeys)
+	if err != nil {
+		return nil, err
+	}
+
+	return (&pb.EphemeralPublicKeyMessage{
+		SenderID:            uint32(epkm.senderID),
+		EphemeralPublicKeys: ephemeralPublicKeys,
+		SessionID:           epkm.sessionID,
+	}).Marshal()
```
pdyraga, line 22: I've just realized the ephemeral public key message in the random beacon has a `receiverID` field that is unused. Can we take the opportunity and clean it up?

pkg/tecdsa/signing/protocol_test.go
```diff
@@ -0,0 +1,302 @@
...
+		if ok {
+			t.Errorf(
+				"[member:%v] found ephemeral key pair generated to self",
+				member.id,
+			)
+		}
+
+		// Assert key pairs are non-nil.
+		for otherMemberID, keyPair := range member.ephemeralKeyPairs {
+			if keyPair.PrivateKey == nil {
```
pdyraga, line 67: Should we also assert that `keyPair` is not nil?

pkg/tecdsa/signing/protocol_test.go
```diff
@@ -0,0 +1,302 @@
...
+	// Assert that each message is formed correctly.
+	for memberID, message := range messages {
+		// We should always be the sender of our own messages.
+		testutils.AssertIntsEqual(
+			t,
+			"message sender",
+			int(memberID),
+			int(message.senderID),
+		)
+
```
pdyraga, line 94: Should we assert `groupSize - 1` public keys are generated in each message?

pdyraga (approved):

pkg/tecdsa/signing/marshaling.go
lukasz-zimnoch: Addressed in https://github.com/keep-network/keep-core/pull/3228/commits/a4e17c9bec28d2bc117652e7a869132d384dc48c

pkg/tecdsa/signing/protocol_test.go
lukasz-zimnoch: Addressed in https://github.com/keep-network/keep-core/pull/3228/commits/c0fca3b97e2908eb5d6956a825aa141836d8951a

PR #3216, opened 2022-08-30
Add Private Testnet bundles generator

nkuba:
We need to generate a bundles for the parties that are joining our
Private Testnet.

The generated bundles contains a preinitialized Ethereum Account details. The account
receives a stake delegation to the `provider` account with authorized `beacon` and
`tbtc` application. This is a production-like experience for the Staking Providers,
where they will receive stakes from their customers.

We have a scripts to create a bundle and initialize the provider
account.

Please see README.md for more details.

Depends on https://github.com/keep-network/keep-core/pull/3211

PR #3227, opened 2022-08-31
Documentation fixes

nkuba:
This PR is a follow-up containing fixing to the documentation published in: https://github.com/keep-network/keep-core/pull/3211 and https://github.com/keep-network/keep-core/pull/3216


docs-v1/run-random-beacon.adoc
```diff
@@ -280,15 +280,15 @@ This is a sample run command for illustration purposes only.
...
+export KEEP_ETHEREUM_PASSWORD=$(cat .secrets/eth-account-password.txt)
 export KEEP_CLIENT_CONFIG_DIR=$(pwd)/config
 export KEEP_CLIENT_PERSISTENCE_DIR=$(pwd)/persistence
 
 docker run -d \
 --entrypoint /usr/local/bin/keep-client
 --volume $KEEP_CLIENT_PERSISTENCE_DIR:/mnt/keep-client/persistence \
 --volume $KEEP_CLIENT_CONFIG_DIR:/mnt/keep-client/config \
---env KEEP_ETHEREUM_PASSWORD=$KEEP_CLIENT_ETHEREUM_PASSWORD \
+--env KEEP_ETHEREUM_PASSWORD=$KEEP_ETHEREUM_PASSWORD \
```
pdyraga, line 291: This doc is for v1 and I believe `KEEP_CLIENT_ETHEREUM_PASSWORD` was the right choice there. Should we change it?

docs/registration.adoc
```diff
@@ -60,17 +60,17 @@ Command templates for the supported applications:
...
-
 export KEEP_CLIENT_ETHEREUM_WS_URL="<Ethereum API WS URL>"
+export OPERATOR_ADDRESS="<Operator Account Address>"
+
+export KEEP_CLIENT_CONFIG_DIR="<Path to the directory containing the Staking Provider Key File>"
 
-export STAKING_PROVIDER_KEY_FILE_PASSWORD="<Staking Provider Account Key File Password>"
 export STAKING_PROVIDER_KEY_FILE_NAME="<Staking Provider Account Key File Name>"
-export OPERATOR_ADDRESS="<Operator Account Address>"
+export STAKING_PROVIDER_KEY_FILE_PASSWORD="<Staking Provider Account Key File Password>"
```
pdyraga, line 69: Two late comments here:
1. I believe `export` is not needed. We use environment variables only to make this example easy to follow and explain the values.
    ```
    $ echo $ETHEREUM_WS_URL

    $ ETHEREUM_WS_URL="ws://my.host/ws"
    $ echo $ETHEREUM_WS_URL
    ws://my.host/ws
2. The only environment variable accepted by Keep Client is `KEEP_ETHEREUM_PASSWORD`. Everything else is in the config. I think we should remove all `KEEP_CLIENT_*` prefixes from these environment variables to not suggest these environment variables are read by the client. I was highly confused about them for a moment so I expect all readers will be even more confused.

To sum up, I would make it look as follows:
```
ETHEREUM_WS_URL="<Ethereum API WS URL>"
OPERATOR_ADDRESS="<Operator Account Address>"
CONFIG_DIR="<Path to the directory containing the Staking Provider Key>"
STAKING_PROVIDER_KEY_FILE_NAME="<Staking Provider Account Key File Name>"
STAKING_PROVIDER_KEY_FILE_PASSWORD="<Staking Provider Account Key File Password>"

docker run ...
```

infrastructure/eth-networks/private-testnet/bundles/bundle-guide.adoc
```diff
@@ -60,6 +60,11 @@ submitted from the Staking Provider Account, please refer to
 link:https://docs.keep.network/registration.html#register-operator[Register Operator]
 documentation.
 
+TIP: When starting the client, remember about running the `keep-client start` command with the `--goerli` flag.
```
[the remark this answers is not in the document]
nkuba: The flag has to be added also for docker.

michalinacienciala:
FYI, I've noticed that we have some out of date information in the https://github.com/keep-network/keep-core/blob/main/README.adoc (for example we link to the no longer existing `Run Random Beacon` doc). We can update it now or later in a separate PR.

docs-v1/run-random-beacon.adoc
nkuba: 52a93cfa893c83ce2b2cb845a7901edad6fe8bc1

docs/run-keep-node.adoc
```diff
@@ -134,8 +137,21 @@ Default port mappings.
...
+An Announced Address is a layered addressing information (`multiaddress`/`multiaddr`)
+announced to the Keep Network that is used by peers to connect with your node, e.g.:
+`/dns4/bootstrap-0.test.keep.network/tcp/3919` or `/ip4/104.154.61.116/tcp/3919`.
+
+You should set the `network.AnnouncedAddresses` (flag: `--network.announcedAddresses`)
+configuration property to the addresses under which your node is reachable for the
+public.
+
+To read more about `multiaddress` see the
+link:https://docs.libp2p.io/reference/glossary/#multiaddr[libp2p docummentation].
```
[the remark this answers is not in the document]
nkuba: e29e65e4675b4c64d42dd745a5a69636b60084ce

infrastructure/eth-networks/private-testnet/bundles/bundle-guide.adoc
```diff
@@ -60,6 +60,11 @@ submitted from the Staking Provider Account, please refer to
...
+
+IMPORTANT: The Operator Account has to be funded with Goerli ETH (GöETH) so the
+client can submit transactions to the Ethereum chain.
```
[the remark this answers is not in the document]
nkuba: 04ba7064aea416eab43d352f86e83efb980363fe

docs/registration.adoc
nkuba: e5da9de9d42f6a1b647d68f3a529fc561e607c0a

I left `KEEP_` prefix for config and storage variables, as we want to be clear and not confuse them in case other applications are being set up on a machine.

nkuba:
> FYI, I've noticed that we have some out of date information in the https://github.com/keep-network/keep-core/blob/main/README.adoc (for example we link to the no longer existing `Run Random Beacon` doc). We can update it now or later in a separate PR.

Nice catch! Let's handle it in another PR. For now I updated just the link you mentioned c42d2a4d9f619e507b75269c4e00a7d530dcf8e4

docs/registration.adoc
pdyraga: I would also remove `KEEP_` prefix for config and storage variables. We do not export them so they are set only for the current terminal session. And it's just a simplification for docs, no one even has to set them, they can pass values to flags directly. 

If we leave `KEEP_` for some of them, this will IMO cause confusion as well. Operator Account is also only for Keep, and Staking Provider Account is also only for Keep. Regarding the conflicts, I imagine `ETHERUM_WS_URL` could be conflicting as well. 

We also still have at least one occurrence of `KEEP_CLIENT_ETHEREUM_WS_URL` in docs.

nkuba: As you wish Lord Commander 11d1cb3b9b087e32c0582f05de88a1c4eae70ecf

pdyraga (approved):

PR #3228, opened 2022-08-31
tECDSA signing: TSS round 1

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-core/pull/3226

Here we are implementing the TSS round 1 for the tECDSA signing protocol and doing some code refactoring in order to reduce the duplication level between tECDSA protocols. Specific changes are described below.

### Extract common TSS tooling to a separate `pkg/tecdsa/common` package

During the work on tECDSA signing, we noticed that some TSS tooling used in tECDSA DKG can be reused. We decided to extract them to a separate package `pkg/tecdsa/common` and reuse them from within both protocol packages (`pkg/tecdsa/dkg` and `pkg/tecdsa/signing`). That is, the following functions were moved:
- `dkg.generateTssPartiesIDs` -> `common.GenerateTssPartiesIDs`
- `dkg.newTssPartyIDFromMemberID` -> `common.newTssPartyIDFromMemberID`
- `dkg.memberIDToTssPartyIDKey` -> `common.memberIDToTssPartyIDKey`
- `dkg.tssPartyIDToMemberID` -> `common.tssPartyIDToMemberID`
- `dkg.resolveSortedTssPartyID` -> `common.ResolveSortedTssPartyID`

Apart from that, we realized that the outgoing messages aggregation logic used in TSS round two of DKG can be encapsulated and extracted as a common tool as well. This way, we introduced `common.AggregateTssMessages` function that is now used in TSS round two of DKG and will be used twice during the signing protocol. Extraction of this logic to a separate function also allowed us to cover this mechanism with thorough unit tests.

### TSS round 1 for tECDSA signing

The main part of this PR is about introducing the TSS round 1 for the tECDSA signing. Specifically, we integrated it with the state machine by adding the required state, message, and member definitions just like we need to do for all protocols based on the state machine pattern. Apart from that, we implemented the actual TSS round 1 logic that consists of starting the underlying TSS party, gathering outgoing messages from the TSS library, aggregating them into a single message, and broadcasting that aggregate message over the network channel.

### Next steps

This PR is just a part of the entire work regarding the tECDSA signing test loop. Please refer to the description of https://github.com/keep-network/keep-core/issues/3042 for the full roadmap and the current status of the work. At this point, the next step will be introducing the second round of TSS signing.


pkg/tecdsa/common/common.go
```diff
@@ -0,0 +1,156 @@
...
+		groupPartiesIDs[i] = newPartyID
+	}
+
+	return partyID, groupPartiesIDs
+}
+
+// NewTssPartyIDFromMemberID creates a new instance of a TSS party ID using
+// the given member ID. Such a created party ID has an unset index since it
+// does not yet belong to a sorted parties IDs set.
+func NewTssPartyIDFromMemberID(memberID group.MemberIndex) *tss.PartyID {
```
pdyraga, line 39: Should we export this function? It's used only from `common` package. Same question to the functions below.

pkg/tecdsa/common/common_test.go
```diff
@@ -0,0 +1,454 @@
...
+			tssMessages: []tss.Message {
+				newMockTssMessage([]byte{0x02}, group.MemberIndex(2)),
+				newMockTssMessage([]byte{0xAA}), // broadcast message
+				newMockTssMessage([]byte{0x01}, group.MemberIndex(1)),
+				newMockTssMessage([]byte{0x03}, group.MemberIndex(3)),
+			},
+			symmetricKeys: map[group.MemberIndex]ephemeral.SymmetricKey{
+				1: &mockSymmetricKey{[]byte{0x01}},
+				2: &mockSymmetricKey{[]byte{0x02}},
+				3: &mockSymmetricKey{[]byte{0x03}},
```
pdyraga, line 197: We could use `0x0A`, `0x0B`, `0x0C` as keys so that we could differentiate the message payload from the key used by the `mockSymmetricKey`:

```
1: {0x01, 0xFF, 0x0A},
2: {0x02, 0xFF, 0x0B},
3: {0x03, 0xFF, 0x0C},
```

vs
```
1: {0x01, 0xFF, 0x01},
2: {0x02, 0xFF, 0x02},
3: {0x03, 0xFF, 0x03},
```

pkg/tecdsa/signing/protocol_test.go
```diff
@@ -234,22 +256,210 @@ func TestGenerateSymmetricKeys_InvalidEphemeralPublicKeyMessage(t *testing.T) {
...
+			memberID,
+		),
+		groupSize-1,
+		len(message.peersPayload),
+	)
+
+	// Each P2P payload should be encrypted using the proper symmetric key.
+	for receiverID, encryptedPayload := range message.peersPayload {
+		symmetricKey := memberSymmetricKeys[receiverID]
+		if _, err := symmetricKey.Decrypt(encryptedPayload); err != nil {
```
pdyraga, line 416: Are we sure it's producing an error and not some mojibake?

lukasz-zimnoch: This is what happens under the hood: https://github.com/keep-network/keep-common/blob/b87c2a9c66090c172dbd4ab4444ef80f206e603e/pkg/encryption/box.go#L48. It returns an error in case of a broken ciphertext. See also https://github.com/keep-network/keep-core/blob/84b8fd8d3e582bd6f0985e113c3bea67bf2935bc/pkg/crypto/ephemeral/symmetric_key_test.go#L69

pkg/tecdsa/common/common.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3228/commits/91a321da613ec37d61a1c236d195020737be0757

pkg/tecdsa/common/common.go
```diff
@@ -0,0 +1,156 @@
...
+	tssParameters *tss.Parameters,
+	memberID group.MemberIndex,
+) *tss.PartyID {
+	sortedPartiesIDs := tssParameters.Parties().IDs()
+	partyIDKey := MemberIDToTssPartyIDKey(memberID)
+	return sortedPartiesIDs.FindByKey(partyIDKey)
+}
+
+// AggregateTssMessages takes a list of TSS messages and build an aggregate
+// consisting of unencrypted broadcast part and encrypted P2P parts intended
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3228/commits/d6f952c79a0ad375afb3a7269a7813a116031f98

pkg/tecdsa/common/common_test.go
lukasz-zimnoch: See d708c42f

pdyraga:
I gave it a test spin with two nodes running locally. A new group got generated and after some time it executed three signing rounds ending with:

> 2022-09-02T16:40:33.133+0200	INFO	keep-tbtc	tbtc/node.go:488	[member:28] generated signature [R: 0xb9f740322c692fa6ca40ebc37e0ec0f33da2573e5851f576460429d12abf059d, S: 0xb9f740322c692fa6ca40ebc37e0ec0f33da2573e5851f576460429d12abf059e, RecoveryID: 1] for message [84114731277345082633178839575533979599871792452423473780111089115273380169116]

PR #3231, opened 2022-08-31
Fix random beacon registration hardhat task

nkuba:
The argument that is expected by the `register` function is an operator
not owner. We fix it.

### DIFF
--- a/solidity/random-beacon/tasks/initialize.ts
+++ b/solidity/random-beacon/tasks/initialize.ts
@@ -126,5 +126,5 @@ task(
   .addParam("provider", "Staking Provider", undefined, types.string)
   .addParam("operator", "Staking Operator", undefined, types.string)
   .setAction(async (args, hre) => {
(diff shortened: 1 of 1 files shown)


solidity/random-beacon/tasks/initialize.ts
```diff
@@ -126,5 +126,5 @@ task(
   .addParam("provider", "Staking Provider", undefined, types.string)
   .addParam("operator", "Staking Operator", undefined, types.string)
   .setAction(async (args, hre) => {
-    await register(hre, "RandomBeacon", args.provider, args.owner)
+    await register(hre, "RandomBeacon", args.provider, args.operator)
```
pdyraga, line 129: Was about to ask but I see that for ecdsa this is correct:

https://github.com/keep-network/keep-core/blob/9436184155607f06f49f253d4e48baef167b4169/solidity/ecdsa/tasks/initialize.ts#L72

pdyraga (approved):

PR #3232, opened 2022-08-31
Add Boar Network bootstrap node to Goerli peers

nkuba:
We add a bootstrap node run by the Boar Network to the list of Goerli
peers.

To test it run `keep-client start` command with `--goerli` flag and verify if the client connects to the three bootstrap nodes.


pdyraga:
Confirmed it's publicly available:
```
❯ telnet bst-a01.test.keep.boar.network 4001
Trying 20.81.168.158...
Connected to bst-a01.test.keep.boar.network.
Escape character is '^]'.
/multistream/1.0.0
```

pdyraga (approved):

PR #3234, opened 2022-08-31
Extend diagnostics to show the client version

dimpar:
Added a client version when calling `/diagnostics` endpoint. This change will help operators keep an eye on the client updates so they can run the latest code. It is especially important during the chaosnet phase where the functionalities of the client will be tested.

Two nodes are running. Calling `curl localhost:8081/diagnostics` and the result is the following:
```
{"client_info":{"chain_address":"0x407190f04d838Ec47dad4C0e0D2a9c49d7737479","network_id":"16Uiu2HAm4WrMWq9Gc7WsymsqgrwbaD1KeW2eH8eY63DTZHXbRsuE","version":"v1.3.1-5899-g4cd974782 (revision 4cd974782)"},"connected_peers":[{"chain_address":"0xBB123EeB3654437b2bcAAd97568c8782988B2CD9","network_id":"16Uiu2HAmUbdAkmW7VnikaVfFejmVWXmXLyWqkFuWqBL6QTnGSJGQ"}]}
```


nkuba (approved):

cmd/start.go
```diff
@@ -126,7 +127,7 @@ func start(cmd *cobra.Command) error {
 	}
 
 	initializeMetrics(ctx, clientConfig, netProvider, blockCounter)
-	initializeDiagnostics(ctx, clientConfig, netProvider, signing)
+	initializeDiagnostics(clientConfig, netProvider, signing, rootCmd.Version)
```
nkuba, line 130: So what is currently returned in the diagnostics call is e.g. `"version":"v1.3.1-5900-g7600556b1 (revision 7600556b1)`.
I'm not sure reading the value from the root command's Version attribute is the best way.
For the command we may build the version output in a different way, we could add some more information (like contracts addresses, etc.), and structure the printed string differently.

For diagnostics, we may want to have a different pattern, e.g. instead of having one property: `"version":"v1.3.1-5900-g7600556b1 (revision 7600556b1)`.
We may want to structure them to validate them easier automatically in the monitoring solutions or for rewards allocation:

```json
"client_info":
    {
        "chain_address": "0xaabeB99Ece50a1dbeE06217e31fc45Bf18D37eEC",
        "network_id": "16Uiu2HAkwHzFwbHjoWRT7qPfmD7MzobSYSe8579ud1Rhjidssna8",
        "version": "v1.3.1-5900-g7600556b1",
        "revision": "7600556b1"
    },
```

We may need more flexibility.

We expose the `main.version` variable and do:
```
initializeDiagnostics(clientConfig, netProvider, signing, main.Version)
```

nkuba (changes requested):

cmd/start.go
dimpar: Ok, I've split the version string to `version` and `revision` [719fbbf](https://github.com/keep-network/keep-core/pull/3234/commits/719fbbfaf17197d64afa5f0aaaa353381230e3c2)
Now
```
"client_info":
    {
        "chain_address": "0xaabeB99Ece50a1dbeE06217e31fc45Bf18D37eEC",
        "network_id": "16Uiu2HAkwHzFwbHjoWRT7qPfmD7MzobSYSe8579ud1Rhjidssna8",
        "version": "v1.3.1-5900-g7600556b1",
        "revision": "7600556b1"
    }
```

pdyraga: Revision `7600556b1` is the commit hash, right? And what is `5900` part?

dimpar: > Revision 7600556b1 is the commit hash, right?
Right

> And what is 5900 part?
It is the number of commits on top of the `v1.3.1`. Note that `v1.3.1` was created on 2020/10/20 so we have 5900 commits since then.

nkuba: @pdyraga note that this is how `git describe` resolves the version. If we run it on a HEAD that was a tagged commit it will resolve to that tag, and not add the number of commits.

pdyraga: Is this piece of information something we want to have in the version then?

I am missing some context in how we resolve the version but my understanding of @nkuba's comment is that this will *not* be added to the client build and it's here only because @dimpar built the client locally on `main` branch and there was no tag on that commit. Is this correct?

Instead of this piece of information, I think we should include the build tag (stable, dev) so our version should be: - For mainnet: `v1.3.1-stable-7600556b1`
For testnet dev builds: `v1.3.1-dev-7600556b1`(during the work on v1.3.1, before v1.3.1 is tagged)

nkuba: This is out of the scope of this PR. We set the version when building the client, and we can set it to anything we want. Let's discuss it in another thread.

pdyraga: OK, if there is no problem in setting the version number during the build, I am fine!

cmd/start.go
```diff
@@ -215,10 +216,10 @@ func initializeMetrics(
 }
 
 func initializeDiagnostics(
```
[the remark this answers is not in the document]
pdyraga: Yes, we should. It's not blocking this PR and I think it's good we cleared up the unused context (no false assumption it's used for something) but yeah - we need to fix it one day.

pdyraga (changes requested):

cmd/start.go
nkuba: Option 2

dimpar: I misunderstood the original Kuba's comment and didn't get the 'main.Version' part.

Sure, option 2 sounds good. We have to provide the full path to the `build.Version` & `build.Revision` in the Makefile, otherwise it won't work. Here's the change [868f332](https://github.com/keep-network/keep-core/pull/3234/commits/868f332c2571d4ddc4f5cd263ca48f50d6d6bda7)

build/build.go
```diff
@@ -0,0 +1,16 @@
...
+var (
+	Version  = "unknown" // Client software version. Set with `-ldflags -X` during the build.
+	Revision = "unknown" // Client software revision. Set with `-ldflags -X` during the build.
+	logger   = log.Logger("keep-build")
+)
+
+func LogVersion() {
+	logger.Info("version=", Version)
+	logger.Info("revision=", Revision)
+}
```
pdyraga, line 16: I don't think we should log the version via logger on every binary execution. Version is available with `--version` command and the fact we log version and revision via logger when executing every possible command does not look good IMO:
```
❯ ./keep-client --version
2022-09-11T16:36:05.007+0200    INFO    keep-build      build/build.go:14       version=v1.3.1-5904-gafea16360
2022-09-11T16:36:05.008+0200    INFO    keep-build      build/build.go:15       revision=afea16360
keep-client version v1.3.1-5904-gafea16360 (revision afea16360)
```

```
❯ ./keep-client --help
2022-09-11T16:34:57.331+0200    INFO    keep-build      build/build.go:14       version=v1.3.1-5904-gafea16360
2022-09-11T16:34:57.331+0200    INFO    keep-build      build/build.go:15       revision=afea16360
Command line interface (CLI) for running a Keep provider

Usage:
  keep-client [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  ethereum    Provides access to Keep network Ethereum contracts.
  help        Help about any command
  ping        bidirectional send between two peers to test the network
  start       Starts the Keep Client

Flags:
  -c, --config string   Path to the configuration file. Supported formats: TOML, YAML, JSON.
      --developer       Developer network
      --goerli          Görli network
  -h, --help            help for keep-client
      --mainnet         Mainnet network
  -v, --version         version for keep-client

Use "keep-client [command] --help" for more information about a command.
```

tl;dr; I'd remove `LogVersion` and the call to it from `main.go`.

.gitignore
```diff
@@ -1,6 +1,6 @@
 node_modules/
 artifacts/
-build/
+**/gen/**/build/
```
pdyraga, line 3: I don't think we should ignore this file. I would say this is fine if this was just about a text file with a value but we have some Go code there. I think it's better if we don't add it to `.gitignore` even if it shows in the local diff occasionally.

pdyraga (commented):
Two small comments and I am go merge.

Tested and works as advertised:
```
$ curl localhost:9001/diagnostics
{"client_info":{"chain_address":"0x146748a2b46B99ee1470b587bc9812Ea59b79597","network_id":"16Uiu2HAm2GxAmGNcQ6Aq8GMEYsYCnuZqbWLqCKD5fGWciwvUMtoj","revision":"afea16360","version":"v1.3.1-5904-gafea16360"},"connected_peers":[]}
```

.gitignore
dimpar: Sorry, I'm not following. I've changed `build/` to `**/gen/**/build/` because the newly added package `build` that handles versioning now (`build.go`) was ignored and CI was failing. `build/` was ignoring every `build` folder. We have a bunch of files under `**/gen/**/build/` that probably should still be ignored.

pdyraga: Ah, my fault!

pdyraga:
@dimpar Can you solve merge conflicts, please? 

@nkuba This PR looks good to me but it will also require your approval given the change request from the past.

build/build.go
dimpar: Sure [7ab8b8d](https://github.com/keep-network/keep-core/pull/3234/commits/7ab8b8d1e59184c7c906acebde0f810c66967a6c)

pdyraga (approved):

nkuba (approved):

.gitignore
dimpar: Looks like generated abis. Here's a path where you can find those files `keep-core/pkg/chain/ethereum/beacon/gen/_contracts/package/build/contracts`

PR #3240, opened 2022-09-05
Pre params pool size in diagnostics

dimpar:
This PR extends the diagnostics endpoint to expose information about the number of pre-parameters in the pool.

Usage:
`localhost:8081/diagnostics` result:

```
{
   "client_info":{
      "chain_address":"0x407190f04d838Ec47dad4C0e0D2a9c49d7737479",
      "network_id":"16Uiu2HAm4WrMWq9Gc7WsymsqgrwbaD1KeW2eH8eY63DTZHXbRsuE",
      "version":"v1.3.1-5900-g7600556b1,
      "revision": 7600556b1,
   },
    "tbtc": {
       "preParamsPoolSize":12,
     }
  }
}
```

Depends on https://github.com/keep-network/keep-core/pull/3234


pkg/tbtc/node.go
```diff
@@ -342,4 +342,8 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
 	} else {
 		logger.Infof("not eligible for DKG with seed [0x%x]", seed)
 	}
+}
+
+func (n *node) DkgExecutor()(*dkg.Executor) {
+	return n.dkgExecutor
```
pdyraga, line 348: I do not think this is terrible but I don't like it either. DKG executor is something that is very internal to `tbtc` package and I do not think it should leak out of there. On the other hand, there is certainly a challenge of how to expose some diagnostic information without leaking internals from a package. I do not think the current DKG pre-params pool size will be the last one.

Instead of exporting the executor, I would implement a diagnostics API allowing any package to register some endpoint under its namespace.

Do not stick to name/API, it's just a rough idea:

```
scheduler := generator.StartScheduler()

initializeMetrics(ctx, clientConfig, netProvider, blockCounter)
diagnosticsEndpoint = diagnostics.Initialize(ctx, clientConfig, netProvider, signing)

tbtc.Initialize(..., diagnosticsEndpoint.ForNamespace("tbtc"))
```

```
type Diagnostics interface {
  ForNamespace(namespace string) DiagnosticsNamespace
}

type DiagnosticsNamespace interface {
  Register(key string, value interface{})
}
```

And then, we would have:
```
{
   "client_info":{
      "chain_address":"0x407190f04d838Ec47dad4C0e0D2a9c49d7737479",
      "network_id":"16Uiu2HAm4WrMWq9Gc7WsymsqgrwbaD1KeW2eH8eY63DTZHXbRsuE",
      "version":"v1.3.1-5900-g7600556b1,
      "revision": 7600556b1,
   },
  {
    "tbtc": {
       "preParamsPoolSize":12,
     }
  }
}
```

dimpar: Ok, I've changed the approach, pls take a look at [157fd0b](https://github.com/keep-network/keep-core/pull/3240/commits/157fd0b37fe2144fe929300403510c1165a3eb6a)

pdyraga:
@dimpar Can you please rebase this branch now when #3234 is merged?

cmd/start.go
```diff
@@ -115,21 +116,24 @@ func start(cmd *cobra.Command) error {
 		return fmt.Errorf("error initializing beacon: [%v]", err)
 	}
 
+	initializeMetrics(ctx, clientConfig, netProvider, blockCounter)
+	registry := initializeDiagnostics(clientConfig)
+	diagnostics.RegisterConnectedPeersSource(registry, netProvider, signing)
+	diagnostics.RegisterClientInfoSource(registry, netProvider, signing, build.Version, build.Revision)
```
pdyraga, line 122: Instead of passing `registry` as the first argument, can we call these functions on the registry pointer? This would allow us to avoid the weird construction with passing `commonDiagnostics` package. 

We would introduce `Registry` structure inside `diagnostics` package that would wrap `Registry` from `keep-common` and that would expose `RegisterConnectedPeersSource`, `RegisterClientInfoSource`, `RegisterApplicationSource`.

dimpar: yea, I like it [3987347](https://github.com/keep-network/keep-core/pull/3240/commits/3987347511864a4001e8d8323da1c5097aabcd5d)

pkg/tbtc/tbtc.go
```diff
@@ -53,10 +54,18 @@ func Initialize(
...
 	scheduler *generator.Scheduler,
 	config Config,
+	registry *diagnostics.Registry,
 ) error {
 	node := newNode(chain, netProvider, persistence, scheduler, config)
 	deduplicator := newDeduplicator()
 
+	assembleTbtcDiagnostics := func() map[string]interface{} {
+		return map[string]interface{}{
+			"preParamsPoolSize": node.dkgExecutor.PreParamsPool().CurrentSize(),
```
pdyraga, line 64: I am not sure if we should expose `PreParamsPool` on the DKG executor. The pre-params pool feels like something really internal to DKG and if we have the executor abstraction, it should hide the complexity.

This comment also applies to #3273 which depends on this PR. 

Although it's a bit more of boilerplate code, I would prefer to do it in the following way:

```
func assembleDiagnostics() map[string]interface{}  {
  dkgDiagnostics := node.dkgExecutor.Diagnostics()

  diagnostics := make(map[string]interface{})
  for k, v := range dkgDiagnostics {
    diagnostics[k] = v
  }

  return diagnostics
}
```

And then here, call just `registry.RegistryApplicationSource("tbtc", assembleDiagnostics)`.

We may find ourselves in a situation when we'd want to add diagnostics for signing as well. Every sub-package of `tbtc` would be responsible for assembling its own internal diagnostics.

I do not want to block this PR on it, but maybe that's something we can address in #3273 @lukasz-zimnoch ?

pkg/diagnostics/diagnostics.go
```diff
@@ -117,3 +125,17 @@ func RegisterClientInfoSource(
 		return string(bytes)
 	})
 }
+
+// RegisterApplicationSource registers the diagnostics source providing
+// information about the application.
+func (r *Registry) RegisterApplicationSource(application string, fetchApplicationDiagnostics func() map[string]interface{}) {
```
pdyraga, line 131: Would be good to wrap this code around 80-character margin.

pdyraga (approved):

pkg/tbtc/tbtc.go
lukasz-zimnoch: See my response at https://github.com/keep-network/keep-core/pull/3273#discussion_r970831193.

However, regarding diagnostics specifically, I have mixed feelings about "Every sub-package of tbtc would be responsible for assembling its own internal diagnostics.". The concept of diagnostics is something _above_ the application and something the application should not be aware of. I think each application component (like DKG executor, tbtc node, etc) should just expose some basic information about their current state and that is. The layer above should gather this and expose it as diagnostics or metrics.

pdyraga: The reasoning about the application component gathering the metrics is solid but I still think that in the current setup we have an abstraction leak. Pre-params pool is something that is used by the executor and is created in `dkg` package. Only inside `dkg` package we should call `GetNow()` on the pool.

With `node.dkgExecutor.PreParamsPool()` exposes from the executor, the pool is leaking and `GetNow()` can be called from `tecdsa`.

If we do not want each sub-package to fill diagnostics - which after second thought I do agree could be overkill - I would expose `CurrentPreParamsPoolSize()` from the executor instead of returning the entire pool.

lukasz-zimnoch: Yep, this is what I proposed in https://github.com/keep-network/keep-core/pull/3273#discussion_r970831193 as well.

pdyraga: I now see @lukasz-zimnoch proposed the same [here](https://github.com/keep-network/keep-core/pull/3273#discussion_r970831193) so I think we should do this change in #3273.

PR #3241, opened 2022-09-05
tECDSA signing: TSS round 3

lukasz-zimnoch:
Depends on: #3239

Here we implement the TSS round three for the tECDSA signing protocol. This round expects aggregate messages produced in TSS round two as input. Each input message consists of encrypted parts intended for specific participants. The TSS round three produces a simple message as output. The output message has only the public broadcast part.

Apart from that, we made some small fixes in TSS round three for tECDSA DKG and we made the `MarkInactiveMembers` functions package-private for both `pkg/tecdsa/dkg` and `pkg/tecdsa/signing` packages.

### Next steps

This PR is just a part of the entire work regarding the tECDSA signing test loop. Please refer to the description of https://github.com/keep-network/keep-core/issues/3042 for the full roadmap and the current status of the work. At this point, the next step will be introducing the fourth round of TSS signing.


pkg/tecdsa/signing/gen/pb/message.proto
```diff
@@ -20,4 +20,10 @@ message TSSRoundTwoMessage {
     uint32 senderID = 1;
     map<uint32, bytes> peersPayload = 2;
     string sessionID = 3;
+}
+
+message TSSRoundThreeMessage {
+    uint32 senderID = 1;
+    bytes payload = 2;
```
pdyraga, line 27: `broadcastPayload`?

pdyraga: Hmm... I think we were using `payload` and `broadcastPayload` pretty inconsistently unless this is not the same type of a payload:
- signing round one: `broadcastPayload`
- signing round three: `payload`
- key generation round one: `payload`
- key generation round two: `broadcastPayload`
- key generation round three: `payload`

pkg/tecdsa/signing/message.go
```diff
@@ -71,3 +71,23 @@ func (trtm *tssRoundTwoMessage) SenderID() group.MemberIndex {
 func (trtm *tssRoundTwoMessage) Type() string {
 	return messageTypePrefix + "tss_round_two_message"
 }
+
+// tssRoundThreeMessage is a message payload that carries the sender's
+// TSS round three components.
+type tssRoundThreeMessage struct {
+	senderID group.MemberIndex
+
+	payload   []byte
```
pdyraga, line 80: Same question about `payload` vs `broadcastPayload`. I see `pkg/tecdsa/dkg/message.go` is also inconsistent about it.

pdyraga (approved):
Left some non-blocking questions, let's address them in the last PR in the signing series or even a separate one.

pkg/tecdsa/signing/gen/pb/message.proto
```diff
@@ -20,4 +20,10 @@ message TSSRoundTwoMessage {
...
+    string sessionID = 3;
 }
```
tomaszslabon, line 29: It seems the file is not formatted with `go fmt` (e.g. there is no newline a the end of the file).
I would consider formatting them.

pkg/tecdsa/signing/gen/pb/message.proto
lukasz-zimnoch: Basically, I use `payload` if there is only one type of payload in the given message and add the `peers*` and `broadcast*` prefixes if there are more types. I can change that to `broadcastPayload` even in the case of a single type.

pdyraga: Yeah, I think we should use `broadcastPayload` everywhere. This is easier to reason about.

lukasz-zimnoch: Addressed by #3271

pkg/tecdsa/signing/message.go
lukasz-zimnoch: Addressed by #3271

PR #3242, opened 2022-09-05
tECDSA signing: TSS round 4

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3042
Depends on: https://github.com/keep-network/keep-core/pull/3241

Here we implement the TSS round four for the tECDSA signing protocol. This round expects simple broadcast messages produced in round three as input and produces a simple broadcast message as output.

### Next steps

This PR is just a part of the entire work regarding the tECDSA signing test loop. Please refer to the description of https://github.com/keep-network/keep-core/issues/3042 for the full roadmap and the current status of the work. At this point, the next step will be introducing the fifth round of TSS signing.

PR #3243, opened 2022-09-05
tECDSA signing: TSS round 5

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-core/pull/3242

Here we implement the TSS round five for the tECDSA signing protocol. This round expects simple broadcast messages produced in round four as input and produces a simple broadcast message as output.

### Next steps

This PR is just a part of the entire work regarding the tECDSA signing test loop. Please refer to the description of https://github.com/keep-network/keep-core/issues/3042 for the full roadmap and the current status of the work. At this point, the next step will be introducing the sixth round of TSS signing.

### DIFF
--- a/pkg/tecdsa/signing/gen/pb/message.proto
+++ b/pkg/tecdsa/signing/gen/pb/message.proto
@@ -33,3 +33,9 @@ message TSSRoundFourMessage {
     bytes payload = 2;
     string sessionID = 3;
 }
+
+message TSSRoundFiveMessage {
+    uint32 senderID = 1;
+    bytes payload = 2;
+    string sessionID = 3;
+}
--- a/pkg/tecdsa/signing/marshaling.go
+++ b/pkg/tecdsa/signing/marshaling.go
@@ -188,6 +188,34 @@ func (trfm *tssRoundFourMessage) Unmarshal(bytes []byte) error {
 	return nil
 }
 
+// Marshal converts this tssRoundFiveMessage to a byte array suitable for
+// network communication.
+func (trfm *tssRoundFiveMessage) Marshal() ([]byte, error) {
+	return (&pb.TSSRoundFiveMessage{
+		SenderID:  uint32(trfm.senderID),
+		Payload:   trfm.payload,
+		SessionID: trfm.sessionID,
+	}).Marshal()
+}
+
+// Unmarshal converts a byte array produced by Marshal to an tssRoundFiveMessage.
+func (trfm *tssRoundFiveMessage) Unmarshal(bytes []byte) error {
+	pbMsg := pb.TSSRoundFiveMessage{}
+	if err := pbMsg.Unmarshal(bytes); err != nil {
+		return err
+	}
+
+	if err := validateMemberIndex(pbMsg.SenderID); err != nil {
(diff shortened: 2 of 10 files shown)


pkg/tecdsa/signing/gen/pb/message.proto
```diff
@@ -32,4 +32,10 @@ message TSSRoundFourMessage {
     uint32 senderID = 1;
     bytes payload = 2;
     string sessionID = 3;
+}
+
+message TSSRoundFiveMessage {
+    uint32 senderID = 1;
+    bytes payload = 2;
```
pdyraga, line 39: `payload` vs `broadcastPayload` as noted [here](https://github.com/keep-network/keep-core/pull/3241#discussion_r964523282).

pkg/tecdsa/signing/marshaling.go
```diff
@@ -188,6 +188,34 @@ func (trfm *tssRoundFourMessage) Unmarshal(bytes []byte) error {
...
+// network communication.
+func (trfm *tssRoundFiveMessage) Marshal() ([]byte, error) {
+	return (&pb.TSSRoundFiveMessage{
+		SenderID:  uint32(trfm.senderID),
+		Payload:   trfm.payload,
+		SessionID: trfm.sessionID,
+	}).Marshal()
+}
+
+// Unmarshal converts a byte array produced by Marshal to an tssRoundFiveMessage.
```
pdyraga, line 201:
```suggestion
// Unmarshal converts a byte array produced by Marshal to a tssRoundFiveMessage.
```

pkg/tecdsa/signing/message.go
```diff
@@ -111,3 +111,23 @@ func (trfm *tssRoundFourMessage) SenderID() group.MemberIndex {
...
+	senderID group.MemberIndex
+
+	payload   []byte
+	sessionID string
+}
+
+// SenderID returns protocol-level identifier of the message sender.
+func (trfm *tssRoundFiveMessage) SenderID() group.MemberIndex {
+	return trfm.senderID
+}
```
pdyraga, line 127: I just realized - why do we declare `SenderID()` at all? `SenderID()` is used only from the same package.

I think we used to have some sender-ID-based filtering code in the past, but I think it's gone now.

pdyraga (approved):
Left some non-blocking questions, let's address them in the last PR in the signing series or even a separate one.

pkg/tecdsa/signing/gen/pb/message.proto
lukasz-zimnoch: Addressed by https://github.com/keep-network/keep-core/pull/3271

pkg/tecdsa/signing/marshaling.go
lukasz-zimnoch: Addressed by https://github.com/keep-network/keep-core/pull/3271

pkg/tecdsa/signing/message.go
lukasz-zimnoch: Yeah but now we use it from within `deduplicateBySender` function so we need that getter.

PR #3247, opened 2022-09-06
Store generated TSS PreParams on disk

nkuba:
Depends on https://github.com/keep-network/keep-core/pull/3265, https://github.com/keep-network/keep-common/pull/104

TSS protocol requires PreParams to be generated upfront, we generate them and store them in a pool of size 3000. It can take days to fill up the pool. Currently, the pool is kept in memory which is pretty inconvenient as a restart of the client erases the generated entries.

This PR implements persistent storage to keep the generated preparams on disk. Once preparams are used they are removed from the storage.

Sample storage structure:
```
.keep/storage/client-1
├── keystore
│   ├── beacon
│   │   ├── archive
│   │   ├── current
│   │   └── snapshot
│   └── tbtc
│       ├── archive
│       ├── current
│       └── snapshot
└── work
    └── tbtc
        └── preparams
            ├── pp_1663245782615_a4cf865
            ├── pp_1663245798022_3161a36
            └── pp_1663245930587_7f757d8
```

This is a draft until https://github.com/keep-network/keep-core/pull/3265 is merged.


pkg/tbtc/registry_test.go
```diff
@@ -223,6 +224,10 @@ func (mph *mockPersistenceHandle) Archive(directory string) error {
 	panic("not implemented")
 }
 
+func (mph *mockPersistenceHandle) Delete(directory string, name string) error {
```
lukasz-zimnoch, line 227: Do we need it? The `registry` uses `persistence.ProtectedHandle` that doesn't expose the `Delete` function.

pkg/tecdsa/dkg/preparams_storage.go
```diff
@@ -0,0 +1,158 @@
+package dkg
+
+import (
+	"encoding/hex"
+	"fmt"
+	"sort"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/crypto"
```
lukasz-zimnoch, line 9: We don't need to hash with keccak256 specifically so I think we can get rid of the `go-ethereum` dependency and use `crypto/sha256` from stdlib. We already use it in our codebase: https://github.com/keep-network/keep-core/blob/e302e7152efd418b90f45cf0977f16fa08b5d7d9/pkg/tbtc/dkg.go#L52

pkg/tecdsa/dkg/preparams_storage.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	fileName := fmt.Sprintf(
+		"pp_%d_%s",
+		// Use timestamp in the filename so that when the data are read from the
+		// disk the First In, First Out algorithm applies.
+		pp.creationTimestamp.UnixMilli(),
+		// Add part of the hash for an ultra unlikely scenario that two saves
+		// happen in the exactly the same millisecond.
+		hex.EncodeToString(crypto.Keccak256(ppBytes))[:7],
```
lukasz-zimnoch, line 60: Just to confirm, we want to take the first 7 bytes of the hash right?

pkg/tecdsa/dkg/preparams_storage.go
```diff
@@ -0,0 +1,158 @@
+package dkg
```
lukasz-zimnoch, line 1: I think we can put this code directly to `preparams.go` file because:
- `pkg/tecdsa/dkg` contains multiple files and this code is strictly connected with pre-params so it would be good to have it close to each other
- `preparams.go` is very small currently so this doesn't harm readability

pkg/tecdsa/dkg/preparams_storage.go
nkuba: Yes, just a part of the hash.

pkg/beacon/registry/groups_test.go
```diff
@@ -269,6 +271,10 @@ func (phm *persistenceHandleMock) Archive(directory string) error {
 	return nil
 }
 
+func (phm *persistenceHandleMock) Delete(directory string, name string) error {
```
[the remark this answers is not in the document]
nkuba: fa6ee0abbc04f580fb10431403ceb5058b87a869

pkg/tbtc/registry_test.go
nkuba: fa6ee0abbc04f580fb10431403ceb5058b87a869

pkg/tecdsa/dkg/preparams_storage.go
nkuba: dd9d2233192032cfae5cdc76c2d8375895aef3d4

pkg/tecdsa/dkg/preparams_storage.go
nkuba: efd70e811556099f3fb52da0a546bb55ec727b3e

lukasz-zimnoch (approved):

PR #3252, opened 2022-09-07
Made client-format check fail on wrongly formatted go code

tomaszslabon:
Added a check that causes the client-format check fail when the go code is wrongly formatted.
In order to avoid errors, this command has to be run before pushing:
```
gofmt -w . 
```
If the code is not formatted with this command, the client-format job will fail.


.github/workflows/client.yml
```diff
@@ -188,4 +188,4 @@ jobs:
         with:
           go-version-file: "go.mod"
       - name: gofmt
-        run: gofmt -d -e .
+        run: if [ "$(gofmt -l . | wc -l)" -gt 0 ]; then exit 1; fi
```
nkuba, line 191: It works, but there is no context outputted in case of an error, see example execution: https://github.com/keep-network/keep-core/runs/8265007828?check_suite_focus=true

image

I'd expect to see a list of the files that are failing the check.

nkuba (changes requested):

.github/workflows/client.yml
tomaszslabon: Yes. I thought it would be even better to be more precise and show the diff of what went wrong instead of just list.
I added it In https://github.com/keep-network/keep-core/pull/3252/commits/f24958dd9c3162a9898346f3097897fc1b1bfdf4.

nkuba: Awesome! https://github.com/keep-network/keep-core/runs/8302348675?check_suite_focus=true

nkuba (approved):
I tested it in https://github.com/keep-network/keep-core/runs/8302348675?check_suite_focus=true
Works like a charm!

PR #3251, opened 2022-09-07
tECDSA signing: TSS finalization

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3042
Depends on: https://github.com/keep-network/keep-core/pull/3250

Here we implement the TSS finalization for the tECDSA signing protocol. This round expects simple broadcast messages produced in round nine as input and produces a TSS result representing the signature.

### Next steps

This PR is just a part of the entire work regarding the tECDSA signing test loop. Please refer to the description of https://github.com/keep-network/keep-core/issues/3042 for the full roadmap and the current status of the work. At this point, the next step will be introducing the retry algorithm.

PR #3254, opened 2022-09-08
tECDSA signing: Consistent party key between TSS runs

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-core/pull/3251

So far, we were doing a simple mapping between the `group.MemberIndex` and the TSS party key for both tECDSA DKG and signing. Unfortunately, this turned out to be wrong. The TSS party key generated based on the `group.MemberIndex` has an impact on the private key share produced as a DKG result which means the same TSS party key must be used by the share's holder during the signing. However, our DKG implementation contains a path where the final `group.MemberIndex` of a member can differ from the original `group.MemberIndex` used during protocol execution. That typically takes place when the DKG process completes with misbehaved members that must be excluded from the final signing group. As result, the `group.MemberIndex` <-> TSS party key mapping logic does not work for the aforementioned case and makes signature generation impossible if the signing group was generated with misbehaved members.

To make it work, we need to make sure that the key used to identify the party during a TSS execution is preserved between runs. We achieve that by introducing different mapping logic for DKG and signing.

For DKG, the TSS party key is now constructed by taking the `group.MemberIndex` and adding it to the DKG seed. The seed component allows avoiding confusion between the `group.MemberIndex` and party key and also ties the TSS keys to a given DKG instance. Here is how it works for an example DKG run:

- Seed: 200
- Member indexes: [1, 2, 3, 4, 5]
- TSS keys: [201, 202, 203, 204, 205]

If member 3 misbehaves and gets excluded from the final signing group, the member indexes and TSS keys arrays will look as follows:

- Member indexes: [1, 2, 4, 5]
- TSS keys: [201, 202, 204, 205]

The final member indexes must be shifted accordingly but TSS keys must remain the same since they are connected with specific private key shares held by members:

- Member indexes: [1, 2, 3, 4]
- TSS keys: [201, 202, 204, 205]

For signing, member 1 must use key `201`, member 2 must use key `202`, and so on. Fortunately, each key share contains the full TSS keys list used during DKG as a reference. That information is held in the `Ks` field, and now we are using it to map the `group.MemberIndex` to a proper TSS party key during the signing.

### Next steps

This PR is just a part of the entire work regarding the tECDSA signing test loop. Please refer to the description of https://github.com/keep-network/keep-core/issues/3042 for the full roadmap and the current status of the work. At this point, the next step will be tweaking the protocol phases duration.


pdyraga:
I am not sure I fully understand the conversion requirements but I think that if we agree to eliminate adding the seed when constructing the `tss.PartyID` for key generation, we can eliminate the converter and have just three functions in `common` package.

I looked at [`tss-lib`](https://github.com/bnb-chain/tss-lib) `README` and it does not mention the party ID has to be a large number. Is there something in the protocol that is not mentioned there?

I tried to put everything together to fully understand the process.

## Key generation

1. DKG is kicked off. We receive an array of 100 operators selected to the signing group by the sortition pool.
2. Each operator has its member index assigned, based on the position in the array. `selected[0]` has member index = 1, `selected[1]` has member index = 2,…, `selected[99]` has member index = 100.
3. When the DKG attempt fails, some members are excluded from the next attempt. The exclusion algorithm provides indexes of members excluded from the next run.
4. During the execution of a retry of a DKG protocol, indexes are not getting shifted. For example, if the original selection was `[0xA, 0xB, 0xC, 0xD, 0xE]` and operator 0xC has been excluded from the attempt, the member indexes stay as `1 (0xA), 2 (0xB), 4 (0xD), 5 (0xE)`. This allows filtering out messages from excluded members using `pkg/protocol/group/membership_validator.go` that is created based on the sortition result, before the retry loop.
5. Once DKG completes successfully, indexes of members excluded from the execution must be set in `EcdsaDkg.Result.misbehavedMembersIndices` and `EcdsaDkg.Result.membersHash` should have the misbehaving members filtered out. When the DKG result gets approved, `EcdsaDkg.dkgResult.membersHash` is stored on-chain.
6. When establishing the transaction submission order when publishing the result to the chain, we take into account the original selection result and we do not skip misbehaving members. It means that the submission may take a bit longer but is more predictable and less complex.
7. Before saving information to disk, misbehaving members are excluded from the result and the member indexes are shifted in `node.go`’s `finalSigningGroup` function. It make sense because first, we do not want to have unnecessary delays when determining the order for publishing a signature, and second, if we need to round-robin the responsibility of proposing the sweep batch, we also do not want unnecessary delays.

We could have a simple mapping logic: `MemberIndex(1) <-> tss.PartyID(1), MemberIndex(2) <-> tss.PartyID(2), …`

And then, the mapping logic would be simple:
```
MemberIndexToTssPartyID(memberIndex group.MemberIndex) *tss.PartyID {
  key := big.NewInt(memberIndex) 
  return tss.NewPartyID(key.Test(10), fmt.Sprintf(“member-%v”, memberIndex), key)
}

TssPartyIDToMemberIndex(partyID *tss.PartyID) group.MemberIndex {
  return group.MemberIndex(partyID.KeyInt())
}
```
There is a catch though! `tssParty.UpdateFromBytes(…)` function requires the `tss.PartyID` passed as a parameter to have an index set and the index is set by the library when `tss.PartyIDs` are sorted into `tss.SortedPartyIDs`. The explanation of it is that the `tss-lib` protocol has its own party indexes inside (similar to our `group.MemberIndex`) and it needs to sort `tss.PartyID`s based on the arbitrary Key value. Note that from the perspective of tss-lib there is no guarantee the Keys used to construct `tss.PartyID`s were all increasing by 1. They could be large, random numbers as well. It’s only specific to our code that these numbers are {1,2,…100}.

Then, we need one more function in `common`:
```
MemberIndexToSortedTssPartyID(memberIndex group.MemerIndex, tssParameters *tss.Parameters) *tss.PartyID {
  sortedPartiesIDs := tssParameters.Parties().IDs()
  key := big.NewInt(memberIndex)
  return sortedPartiesIDs.FindByKey(key)
}
```

That function would be used instead of `ResolveSortedTssPartyID`.


## Signing

1. When DKG result is stored on disk, every signing group member stores `tecdsa.PrivateKeyShare`. This structure includes `privateKeyShare.Data().Ks` which is `[]*big.Int` of party ID keys. The member indexes reconstructed from this array will correspond to the ones published to the chain, that is, with misbehaving members filtered out.
2. Group member info saved to disk has the indexes shifted as noted in point 7 of Key generation section, so if we agree for a simple mapping logic where `MemberIndex(1) <-> tss.PartyID(1)`, then we do not need a separate converter for signing and I believe we can use the same logic as for the key generation.

pdyraga:
My reasoning in the previous comment was not valid. I did not take into account that when we shift indexes in the key generation algorithm at point 7, these indexes no longer match the ones in `Ks`. 

1. The successful DKG run DOES NOT have the member indexes shifted. So if the successful DKG run was with `[MemberIndex(1), MemberIndex(3), MemberIndex(5)]`, the TSS Party IDs will be `[PartyID(1), PartyID(3), PartyID(5)]`.
2. We DO SHIFT member indexes before submitting the result to the chain and saving the result on-disk. Hence, the indexes saved on disk will be `[MemberIndex(1), MemberIndex(2), MemberIndex(3)]`.
3. They DO NOT match  `[PartyID(1), PartyID(3), PartyID(5)]` saved in `Ks`.

pkg/tecdsa/dkg/member.go
```diff
@@ -333,3 +339,35 @@ func (sm *signingMember) initializeSubmittingMember() *submittingMember {
...
+	memberIndex group.MemberIndex,
+) *big.Int {
+	return new(big.Int).Add(ic.seed, big.NewInt(int64(memberIndex)))
+}
+
+func (ic *identityConverter) TssPartyIDToMemberIndex(
+	partyID *tss.PartyID,
+) group.MemberIndex {
+	return group.MemberIndex(
+		new(big.Int).Sub(partyID.KeyInt(), ic.seed).Int64(),
```
pdyraga, line 371: We allow it to go <0 because we'll fail in `AggregateTssMessages` on `cannot get symmetric key with member ...` and no harm can be done this way, correct?

pdyraga: Actually, I think we should return `MemberIndex(0)` in this case and make it clear in the docs.

pdyraga (approved):

pkg/tecdsa/signing/member_test.go
```diff
@@ -111,4 +113,76 @@ func TestShouldAcceptMessage(t *testing.T) {
...
+}
+
+func TestIdentityConverter_MemberIndexToTssPartyID(t *testing.T) {
+	converter := &identityConverter{keys: []*big.Int{
+		big.NewInt(301),
+		big.NewInt(303),
+		big.NewInt(304),
+		big.NewInt(305),
+	}}
+	memberID := group.MemberIndex(2)
```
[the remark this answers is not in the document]
pdyraga: Done in https://github.com/keep-network/keep-core/pull/3262.

pkg/tecdsa/dkg/member_test.go
```diff
@@ -105,4 +109,61 @@ func TestShouldAcceptMessage(t *testing.T) {
 			)
 		})
 	}
+}
+
+func TestIdentityConverter_MemberIndexToTssPartyID(t *testing.T) {
+	converter := &identityConverter{seed: big.NewInt(300)}
+	memberID := group.MemberIndex(2)
```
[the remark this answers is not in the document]
pdyraga: Done in https://github.com/keep-network/keep-core/pull/3262

pkg/tecdsa/dkg/member.go
pdyraga: Will be done in https://github.com/keep-network/keep-core/pull/3262.

pkg/tecdsa/signing/member.go
```diff
@@ -420,3 +426,37 @@ func (fm *finalizingMember) markInactiveMembers(
...
+}
+
+func (ic *identityConverter) TssPartyIDToMemberIndex(
+	partyID *tss.PartyID,
+) group.MemberIndex {
+	index := slices.IndexFunc(ic.keys, func(key *big.Int) bool {
+		return key.Cmp(partyID.KeyInt()) == 0
+	})
+
+	return group.MemberIndex(index + 1)
```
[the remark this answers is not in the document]
pdyraga: Will be done in https://github.com/keep-network/keep-core/pull/3262.

pkg/tecdsa/dkg/member.go
lukasz-zimnoch: Yep, this is one reason. Another one is the fact that should never happen if conversions are done through the converter.

PR #3255, opened 2022-09-08
Query bootstrap node

dimpar:
Depends on https://github.com/keep-network/keep-core/pull/3240

> The script should query the bootstrap node for the list of connected peers and for each peer, extract information about the client version and the number of pre-parameters in the pool. Aggregate the result and publish it to a bucket on GCP

This is the 1 out of 2 PRs that extract peer information. This change adds a new method called `ConnectedPeersAddrInfo()` that retrieves peer IP/DNS addresses. These addresses are later used to call peers to get their diagnostics & metrics data like `chain address, version, and pre params pool size`

The bash scripts part will be handled in a different [PR](https://github.com/keep-network/keep-core/pull/3290).

Example JSON that would be stored on GCP https://storage.googleapis.com/diagnostics_test/peers_1663587803.json
```
{
   "peers":[
      {
         "chain_address":"0x5232E393602c498D216d680C4cAbb0Cd179d1d16",
         "version":"v1.3.1-5904-g06081c6af (revision 06081c6af)",
         "preParamPoolSize":279
      },
      {
         "chain_address":"0xcAB2a402bAc470686d14956FB310D51BbEF9fA31",
         "version":"v1.3.1-5904-g06081c6af (revision 06081c6af)",
         "preParamPoolSize":103
      }
   ]
}
```


scripts/diagnostics.sh
```diff
@@ -0,0 +1,88 @@
...
+      echo "${peerAddress}:${port}/${ENDPOINT} is not a valid endpoint"
+    fi
+  done
+done
+
+peersJsonArray=$(jq -s <<< "${peers[@]}")
+peersJson=$(jq --null-input -r --argjson peersJsonArray "${peersJsonArray}" '{peers: $peersJsonArray}')
+
+printf "${LOG_START}Saving diagnostics to a file...${LOG_END}"
+echo $peersJson > ${DIAGNOSTICS_DIR}/peers.json
```
nkuba, line 86: Maybe we could append a timestamp to the file name? We don't want to override files that were created and uploaded before in previous rewards allocation periods.

scripts/diagnostics.sh
```diff
@@ -0,0 +1,88 @@
...
+mkdir ${DIAGNOSTICS_DIR}
+
+peerAddresses=$(curl ${ADDRESS}:${PORT}/${ENDPOINT} | jq -r '.connected_peers[].address')
+
+peers=()
+# Iterate over the connected peers using their IP addresses
+for peerAddress in ${peerAddresses[@]}; do
+  # Iterate over the eligible peer ports. First port that returns no error breaks
+  # the loop and assign data to peer[] array.
+  for port in ${PEER_PORTS[@]}; do
```
nkuba, line 71: So `PEER_PORTS` is defined as an array, could we make it configurable as a range? The reason for that is we could have 99 ports to check, so we don't want to list them all but simply set the range to `9701-9799`.

scripts/upload_peers.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# Run script
+printf "${LOG_START}Authenticating with gcloud...${LOG_END}"
+gcloud auth activate-service-account --key-file $key_file_path
+
+printf "${LOG_START}Setting cloud context to the ${PROJECT_NAME}...${LOG_END}"
+gcloud config set project $SOURCE_PROJECT_ID
+
+printf "${LOG_START}Uploading peers info to a ${DESTINATION_BUCKET_NAME} bucket...${LOG_END}"
+gcloud storage cp ${PEERS_LOCATION} gs://$DESTINATION_BUCKET_NAME/
```
nkuba, line 59: I like using `gcloud` but I'm afraid it might cause some problems on the bootstrap VM, as this is an additional tool that has to be installed and maintained. 

Could we try a simple way, e.g. with `curl`? 

Resources:
- https://cloud.google.com/storage/docs/uploading-objects#rest-upload-objects
- https://stackoverflow.com/questions/42418493/how-do-i-upload-to-google-cloud-storage-with-curl-and-an-api-key

scripts/diagnostics.sh
dimpar: Sure, done in here [2f09704](https://github.com/keep-network/keep-core/pull/3255/commits/2f09704753355c105421d8d1bc0cb4a77c5d55df). What is the plan for older files? What I mean is after some time we probably have a lot of files in the bucket. Do we want to clean files that are ex 3 months old? There should be probably a separate script for that. If yes, I'll create an issue for that. @pdyraga Thoughts?

pdyraga: I wouldn't care about it. It's only 360 files per year, it's not that crazy. We can even archive them manually.

pdyraga: Also, note that this reward mechanism is just temporary. Eventually, we'll switch to sortition pool rewards.

scripts/upload_peers.sh
dimpar: This was also my concern but since we already used `gcloud` in our scripts, I figured we can use it here as well. However, I agree that it might be an additional tool that a bootstrap node has to keep an eye on, so here's a change [d154f99](https://github.com/keep-network/keep-core/pull/3255/commits/d154f996b3090424477429f5f1d10fe2b64a417c)

scripts/diagnostics.sh
dimpar: Sounds good, here's a change [ce70300](https://github.com/keep-network/keep-core/pull/3255/commits/ce70300207b2130f48a66a8c75baa17c274c506a)

pkg/diagnostics/diagnostics.go
```diff
@@ -49,10 +56,27 @@ func (r *Registry) RegisterConnectedPeersSource(
...
+				// Peer may contain public and local addresses. We need to fetch a
+				// public address only. Range of local addresses starting from 127.* are
+				// skipped.
+				if strings.Contains(peerAddrInfo.String(), eligibleDNS4) ||
+					(strings.Contains(peerAddrInfo.String(), eligibleIP4) &&
+						!strings.Contains(peerAddrInfo.String(), "/127.")) {
+					// address is formatted as follows /<protocol_code>/<ip_address>/<connection_protocol>/<port>
+					// Address is  the 1st index
+					peerAddr = strings.Split(strings.Trim(peerAddrInfo.String(), "/"), "/")[1]
+				}
```
pdyraga, line 77: Is it possible for a peer to bind to more than one address and one of them not being a loopback? 

I think I can bind to a public, externally available address as well as to an internal address that is not a loopback address (e.g. 10.0.0.0/8  or 172.16.0.0/12). Is it possible and if so, how is this handled?  I *think* this code may return 10.0.0.0/8 if it is first on the list.

pdyraga:
Did you test this script with our testnet? Can you provide an example of a call for this scenario?

dimpar:
> Did you test this script with our testnet? Can you provide an example of a call for this scenario?

I've run my local node instance pointing to Goerli network, but Goerli nodes don't have all the latest changes so the API call will return `null` for the version and pre-params size. This PR also adds `address` to the JSON result. To fully test it on Goerli, this PR should be merged first and then we can run `diagnostics.sh` and `upload_peers.sh`.

I've tested it on my local box too.

pkg/diagnostics/diagnostics.go
dimpar: Good point, the current version filters out only ip4 IPs that start with `127.*`. It looks like the reserved "local" range of IPs are the following and we should add them as well:
- `10.*`
- `172.*`
- `192.*`
Thoughts?

pdyraga: It would not be precise because each private address space has a different network mask: 10/8, 172.16/12, 192.168/16 so while 10/8 can be expressed as `10.*`, 192.168/16 can not be expressed as `192.*`. See [here](https://www.rfc-editor.org/rfc/rfc1918) please, section "3. Private Address Space".

I could also imagine the node can be configured with a private hostname, not accessible from the internet.

Let's take a step back. Why do we do this filtering in diagnostics at all? What is returned by `connectionManager.ConnectedPeersAddrInfo()` locally and what is returned by `connectionManager.ConnectedPeersAddrInfo()` on testnet?

dimpar: `connectionManager.ConnectedPeersAddrInfo()` on Goerli returns an array of peers with addresses. As of writing this comment, we have 7 peers
```
[
   {
      "16Uiu2HAmCcfVpHwfBKNFbQuhvGuFXHVLQ65gB4sJm7HyrcZuLttH":[
         /dns4/bootstrap-0.test.keep.network/tcp/3919
      ]
   }{
      "16Uiu2HAm3eJtyFKAttzJ85NLMromHuRg4yyum3CREMf6CHBBV6KY":[
         /dns4/bootstrap-1.test.keep.network/tcp/3919
      ]
   }{
      "16Uiu2HAmMosdpAuRSw1ahNhqFq8e3Y4d4c5WZkjW1FGQi5WJwWZ7":[
         /dns4/bst-a01.test.keep.boar.network/tcp/4001
      ]
   }{
      "16Uiu2HAmQaxEutkqzahSE19b9DhSXGCdb1BAfkGrt26obq1n6N9k":[
         /ip4/3.137.162.218/tcp/3919
      ]
   }{
      "16Uiu2HAky2Y4Tyq5vTA1CxikcDes6o5EH11i2qcg5dBV9W3Lks5c":[
         /ip4/34.141.9.57/tcp/3919
      ]
   }{
      "16Uiu2HAm77eSvRq5ioD4J8VFPkq3bJHBEHkssCuiFkgAoABwjo2S":[
         /dns4/keep-validator-0.eks-ap-northeast-2-secure.staging.staked.cloud/tcp/3919
      ]
   }{
      "16Uiu2HAm1buBFu8sHmgToezz6994uqQCZZ42azxrJP5GTxWmcBPA":[
         /ip4/138.68.240.252/tcp/3919
      ]
   }
]
```

pdyraga: OK, and then why do we need to filter them out? Isn't it that local addresses are returned only for local deployments which is totally fine?

dimpar: Looking at this response we do not. If we can assume that there would be no local addresses in the response on the test / mainnet then we can remove this filter. I was leaning towards "it won't hurt" adding an extra "if" in case I don't know something about the libp2p lib.

pkg/diagnostics/diagnostics.go
```diff
@@ -49,10 +56,23 @@ func (r *Registry) RegisterConnectedPeersSource(
 	r.Registry.RegisterSource("connected_peers", func() string {
 		connectionManager := netProvider.ConnectionManager()
 		connectedPeers := connectionManager.ConnectedPeers()
+		connectedPeersAddrInfo := connectionManager.ConnectedPeersAddrInfo()
 
 		peersList := make([]map[string]interface{}, len(connectedPeers))
 		for i := 0; i < len(connectedPeers); i++ {
 			peer := connectedPeers[i]
+			peersAddrInfo := connectedPeersAddrInfo[i].Addrs
```
pdyraga, line 64: Here we assume `len(connectedPeers) == len(peersAddrInfo)` and that the order of peers is the same.

I am pretty sure this is not always true. If one or more peers connect or disconnect after `connectionManager.ConnectedPeers()` and before `connectionManager.ConnectedPeersAddrInfo()`, the length of the slices will be different.

Also, there is nothing in libp2p `network.Dialer` interface documentation for `Peers() []peer.ID` saying that the returned peers are ordered. It could happen that elements of slices returned by `connectionManager.ConnectedPeersAddrInfo()` and `connectionManager.ConnectedPeers()` does not match.

I think we do not need to call `connectionManager.ConnectedPeers()` in this endpoint at all. Isn't `connectionManager.ConnectedPeersAddrInfo()` enough?

pkg/diagnostics/diagnostics.go
```diff
@@ -49,10 +56,23 @@ func (r *Registry) RegisterConnectedPeersSource(
...
+
+			var peerAddr string
+			for _, peerAddrInfo := range peersAddrInfo {
+				if strings.Contains(peerAddrInfo.String(), eligibleDNS4) ||
+					(strings.Contains(peerAddrInfo.String(), eligibleIP4)) {
```
pdyraga, line 69: Why do we filter out everything that is not ip4 or dns4? I think it's possible that we can get someone connected via ip6. 

I think we should return all addresses known by the peer store, and if we want to do some filtering, we should do it in the code consuming the output of this endpoint.

pkg/net/local/local.go
```diff
@@ -87,7 +89,8 @@ func (lp *localProvider) ConnectionManager() net.ConnectionManager {
 type localConnectionManager struct {
 	mutex sync.Mutex
 
-	peers map[string]*operator.PublicKey
+	peers         map[string]*operator.PublicKey
+	peersAddrInfo map[string]peer.AddrInfo
```
pdyraga, line 93: It will always be empty, right? 

Then, `ConnectedPeersAddrInfo` will always return an empty slice even though `ConnectedPeers` will not be empty. 

Instead of adding a new mapping, I'd use `peers` in `ConnectedPeersAddrInfo` and returned some custom address, like `/local/$peerID`.

pdyraga (changes requested):

pkg/diagnostics/diagnostics.go
dimpar: All the protocol codes are here: https://github.com/multiformats/multiaddr#protocols I don't think we should return all of them. After talking to @nkuba we decided to focus on ip4 and dns4. If you think we should also allow for ip6 then I can add it as well.

.gitignore
```diff
@@ -54,3 +54,4 @@ yarn-error.log
 
 # Temporary directory
 tmp/
+diagnostics/
```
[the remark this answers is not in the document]
dimpar: ah, this should be here https://github.com/keep-network/keep-core/pull/3290 in the PR that is a follow-up of this one. Will change.

pkg/diagnostics/diagnostics.go
pdyraga: How are these protocol codes assembled? Aren't they resolved by the peerstore based on how the client can be reached?

dimpar: I've timeboxed my search, and this is the closest what I found to answer your question
https://github.com/libp2p/specs/blob/master/addressing/README.md#transport-multiaddrs

My understanding is that libp2p supports some of the `multiaddr` protocols described in this table https://github.com/multiformats/multiaddr/blob/master/protocols.csv The docs says 
> Most libp2p transports use the IP protocol as a foundational layer, and as a result, most transport multiaddrs will begin with a component that represents an IPv4 or IPv6 address.
Also reading further, it seems that libp2p will try to resolve protocols to IPv4 or IPv6. I'm not sure if there are any edge cases when libp2p can't/won't resolve to ip4/ip6 but this lib is not trivial and it's hard to answer. 

Having said that, in most cases I think we should be fine without any filtering just like you suggested, but I'm not 100% confident.

pkg/net/local/local.go
dimpar: Is this what you mean? [697609a](https://github.com/keep-network/keep-core/pull/3255/commits/697609a47913d9a2653081bb73c677d38c1d034c)

pkg/diagnostics/diagnostics.go
dimpar: You're right [cd25e92](https://github.com/keep-network/keep-core/pull/3255/commits/cd25e929b19592e13aba7e81c8c63017f5c003fd)

pkg/net/net.go
```diff
@@ -69,6 +70,7 @@ type Provider interface {
 // from any given connected peer.
 type ConnectionManager interface {
 	ConnectedPeers() []string
+	ConnectedPeersAddrInfo() []peer.AddrInfo
```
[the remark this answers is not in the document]
dimpar: Adding a custom type in `libp2p.go` creates a cyclic dependency. I've simplified it to just `map[string][]string` [cd25e92](https://github.com/keep-network/keep-core/pull/3255/commits/cd25e929b19592e13aba7e81c8c63017f5c003fd)

.gitignore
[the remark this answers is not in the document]
dimpar: [7434329](https://github.com/keep-network/keep-core/pull/3255/commits/7434329e6fa85d40cddb54798332138396f4a01b)

pkg/diagnostics/diagnostics.go
```diff
@@ -12,6 +13,12 @@ import (
 
 var logger = log.Logger("keep-diagnostics")
 
+// All available protocols https://github.com/multiformats/multiaddr#protocols
+// Eligible protocols are used to fetch the address to call the /diagnostics
+// endpoint
+var eligibleIP4 = "ip4"
+var eligibleDNS4 = "dns4"
```
nkuba, line 20: Are those used anywhere?

nkuba:
Overall it works great. But I noticed that sometimes happens that the `addresses` property is returned as `null`.
```
{
    "client_info":
    {
        "chain_address": "0xaabeB99Ece50a1dbeE06217e31fc45Bf18D37eEC",
        "network_id": "16Uiu2HAkwHzFwbHjoWRT7qPfmD7MzobSYSe8579ud1Rhjidssna8",
        "revision": "unknown",
        "version": "unknown"
    },
    "connected_peers":
    [
        {
            "addresses":
            [
                "138.68.240.252"
            ],
            "chain_address": "0xB8d1511aFab121f07cfDC7D1086371b1Fa758Fa2",
            "network_id": "16Uiu2HAm1buBFu8sHmgToezz6994uqQCZZ42azxrJP5GTxWmcBPA"
        },
        {
            "addresses":
            [
                "bootstrap-1.test.keep.network",
                "keep-client-1.default.svc.cluster.local"
            ],
            "chain_address": "0xcAB2a402bAc470686d14956FB310D51BbEF9fA31",
            "network_id": "16Uiu2HAm3eJtyFKAttzJ85NLMromHuRg4yyum3CREMf6CHBBV6KY"
        },
        {
            "addresses":
            [
                "bootstrap-0.test.keep.network",
                "keep-client-0.default.svc.cluster.local"
            ],
            "chain_address": "0x0eC14BC7cCA82c942Cf276F6BbD0413216dDB2bE",
            "network_id": "16Uiu2HAmCcfVpHwfBKNFbQuhvGuFXHVLQ65gB4sJm7HyrcZuLttH"
        },
        {
            "addresses":
            [
                "keep-validator-0.eks-ap-northeast-2-secure.staging.staked.cloud"
            ],
            "chain_address": "0xCa8754f7060A0648824f274e3a4d897FA497139D",
            "network_id": "16Uiu2HAm77eSvRq5ioD4J8VFPkq3bJHBEHkssCuiFkgAoABwjo2S"
        },
        {
            "addresses":
            [
                "34.141.9.57"
            ],
            "chain_address": "0x1F91c3e91745aB09cc8e619ceD8Fa79B602e3649",
            "network_id": "16Uiu2HAky2Y4Tyq5vTA1CxikcDes6o5EH11i2qcg5dBV9W3Lks5c"
        },
        {
            "addresses":
            [
                "3.137.162.218"
            ],
            "chain_address": "0x37847B906612885d7172639947Cecf76036458bA",
            "network_id": "16Uiu2HAmQaxEutkqzahSE19b9DhSXGCdb1BAfkGrt26obq1n6N9k"
        },
        {
            "addresses": null,
            "chain_address": "0x5232E393602c498D216d680C4cAbb0Cd179d1d16",
            "network_id": "16Uiu2HAmMosdpAuRSw1ahNhqFq8e3Y4d4c5WZkjW1FGQi5WJwWZ7"
        }
    ]
}
```

pkg/diagnostics/diagnostics.go
```diff
@@ -48,11 +55,19 @@ func (r *Registry) RegisterConnectedPeersSource(
...
+
+		var peersList []map[string]interface{}
+		for peer, multiaddrs := range connectedPeersAddrInfo {
+			var addresses []string
+			var peerAddr string
+			for _, address := range multiaddrs {
+				// Multiaddr is formatted as follows /<protocol_code>/<address>/<connection_protocol>/<port>
+				// Address is the 1st index
+				peerAddr = strings.Split(strings.Trim(address, "/"), "/")[1]
+				addresses = append(addresses, peerAddr)
```
pdyraga, line 68: Wouldn't it be better to return the entire multiaddr and let the consuming code extract address, if needed?

I think knowing the port address and protocol code could be beneficial for debugging problems.

nkuba: +1 ^^

pdyraga: I *guess* the parsing could also be the reason we sometimes see null addresses [here](https://github.com/keep-network/keep-core/pull/3255#issuecomment-1255898552). Not all cases may be addressed.

dimpar: Sure, we can remove all the filtering from here. It is harder to parse strings in the bash scripts, but we won't touch the client if we'll need some other info from the multiaddrs. [414caae](https://github.com/keep-network/keep-core/pull/3255/commits/414caae76270fe8e549a361de8accf4d00fe9ba4)

Not sure what's the reason for `null`, I can't recreate it
```
{
   "client_info":{
      "chain_address":"0x6357132BA57Ef90b71ffEfdaf0ABDdEA91C07008",
      "network_id":"16Uiu2HAmCrd9nvuum995QoPwdfvTn2siCQtyBkLZnrcYHoLuKnuS",
      "revision":"unknown",
      "version":"unknown"
   },
   "connected_peers":[
      {
         "chain_address":"0xCa8754f7060A0648824f274e3a4d897FA497139D",
         "multiaddrs":[
            "/dns4/keep-validator-0.eks-ap-northeast-2-secure.staging.staked.cloud/tcp/3919"
         ],
         "network_id":"16Uiu2HAm77eSvRq5ioD4J8VFPkq3bJHBEHkssCuiFkgAoABwjo2S"
      },
      {
         "chain_address":"0x1F91c3e91745aB09cc8e619ceD8Fa79B602e3649",
         "multiaddrs":[
            "/ip4/34.141.9.57/tcp/3919"
         ],
         "network_id":"16Uiu2HAky2Y4Tyq5vTA1CxikcDes6o5EH11i2qcg5dBV9W3Lks5c"
      },
      {
         "chain_address":"0x5232E393602c498D216d680C4cAbb0Cd179d1d16",
         "multiaddrs":[
            "/dns4/bst-a01.test.keep.boar.network/tcp/4001"
         ],
         "network_id":"16Uiu2HAmMosdpAuRSw1ahNhqFq8e3Y4d4c5WZkjW1FGQi5WJwWZ7"
      },
      {
         "chain_address":"0xcAB2a402bAc470686d14956FB310D51BbEF9fA31",
         "multiaddrs":[
            "/dns4/bootstrap-1.test.keep.network/tcp/3919",
            "/dns4/keep-client-1.default.svc.cluster.local/tcp/3919"
         ],
         "network_id":"16Uiu2HAm3eJtyFKAttzJ85NLMromHuRg4yyum3CREMf6CHBBV6KY"
      },
      {
         "chain_address":"0x0eC14BC7cCA82c942Cf276F6BbD0413216dDB2bE",
         "multiaddrs":[
            "/dns4/bootstrap-0.test.keep.network/tcp/3919",
            "/dns4/keep-client-0.default.svc.cluster.local/tcp/3919"
         ],
         "network_id":"16Uiu2HAmCcfVpHwfBKNFbQuhvGuFXHVLQ65gB4sJm7HyrcZuLttH"
      },
      {
         "chain_address":"0xB8d1511aFab121f07cfDC7D1086371b1Fa758Fa2",
         "multiaddrs":[
            "/ip4/138.68.240.252/tcp/3919"
         ],
         "network_id":"16Uiu2HAm1buBFu8sHmgToezz6994uqQCZZ42azxrJP5GTxWmcBPA"
      }
   ]
}
```

pkg/diagnostics/diagnostics.go
dimpar: Nah, I forgot to remove them after removing addresses filtering [414caae](https://github.com/keep-network/keep-core/pull/3255/commits/414caae76270fe8e549a361de8accf4d00fe9ba4)

pkg/diagnostics/diagnostics.go
pdyraga: Here we have confirmation in the code:

> libp2p makes a distinction between a peer's identity and its location. A peer's identity is stable, verifiable, and valid for the entire lifetime of the peer (whatever that may be for a given application). Peer identities are derived from public keys as described in the [peer id spec](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md).
> 
> On a particular network, at a specific point in time, a peer may have one or more locations, which can be represented using addresses. For example, I may be reachable via the global IPv4 address of 7.7.7.7 on TCP port 1234.
> 
> In a system that only supported TCP/IP or UDP over IP, we could easily write our addresses with the familiar <ip>:<port> notation and store them as tuples of address and port. However, libp2p was designed to be transport agnostic, which means that we can't assume that we'll even be using an IP-backed network at all.

(from https://github.com/libp2p/specs/blob/master/addressing/README.md#overview)

Now, our `libp2p.go`:

https://github.com/keep-network/keep-core/blob/723bc65b9a46914a333d6f8a317ada4cf8b8d099/pkg/net/libp2p/libp2p.go#L400-L405

From [libp2p `options.go`](https://github.com/libp2p/go-libp2p/blob/b54bacef92dd04d034cc2608364499d015c80f60/options.go#L44-L50):

```
// ListenAddrs configures libp2p to listen on the given addresses.
func ListenAddrs(addrs ...ma.Multiaddr) Option {
	return func(cfg *Config) error {
		cfg.ListenAddrs = append(cfg.ListenAddrs, addrs...)
		return nil
	}
}
```

And then, in [`libp2p.go` documentation for `func New(opts ...Option) (host.Host, error)`](https://github.com/libp2p/go-libp2p/blob/b54bacef92dd04d034cc2608364499d015c80f60/libp2p.go#L34-L38):
```
// - If no transport and listen addresses are provided, the node listens to
// the multiaddresses "/ip4/0.0.0.0/tcp/0" and "/ip6/::/tcp/0";
//
// - If no transport options are provided, the node uses TCP, websocket and QUIC
// transport protocols;
```

dimpar: Makes sense. Also, worth noting that this part 
```
// - If no transport options are provided, the node uses TCP, websocket and QUIC
// transport protocols;
```
leads to [TCP](https://github.com/libp2p/specs/blob/master/addressing/README.md#tcp), [WebSockets](https://github.com/libp2p/specs/blob/master/addressing/README.md#websockets) and [QUIC](https://github.com/libp2p/specs/blob/master/addressing/README.md#quic) which all of them resolve to `<ip-multiaddr>` which is described in [IP and Name resolution](https://github.com/libp2p/specs/blob/master/addressing/README.md#ip-and-name-resolution)

nkuba (approved):

pkg/diagnostics/diagnostics.go
```diff
@@ -48,11 +48,10 @@ func (r *Registry) RegisterConnectedPeersSource(
...
+		connectedPeersAddrInfo := connectionManager.ConnectedPeersAddrInfo()
 
-		peersList := make([]map[string]interface{}, len(connectedPeers))
-		for i := 0; i < len(connectedPeers); i++ {
-			peer := connectedPeers[i]
+		var peersList []map[string]interface{}
+		for peer, multiaddrs := range connectedPeersAddrInfo {
 			peerPublicKey, err := connectionManager.GetPeerPublicKey(peer)
 			if err != nil {
 				logger.Error("error on getting peer public key: [%v]", err)
```
pdyraga, line 57:
```suggestion
				logger.Errorf("error on getting peer public key: [%v]", err)
```

pdyraga (approved):

PR #3260, opened 2022-09-09
tECDSA signing: Retries

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-core/pull/3259

Here we introduce the retry mechanism for the tECDSA signing protocol. The overall construction is very similar to the tECDSA DKG retries though is simpler as the signing relies only on the random retry algorithm. Unlike the DKG retry loop, the signing retry loop does not care about inactive members as signing is performed with a much smaller subset of members (at least 51 members), chosen from the initial set (at least 90 members). In that case, the success chance is quite high even when retries are done randomly. This results in a much simpler code that should do the work until we employ a signing algorithm that supports full attributability.

### Next steps

This PR is just a part of the entire work regarding the tECDSA signing test loop. Please refer to the description of https://github.com/keep-network/keep-core/issues/3042 for the full roadmap and the current status of the work. At this point, the next step will be introducing logging improvements for signing.


pkg/tbtc/signing.go
```diff
@@ -0,0 +1,224 @@
...
+	signingGroupMemberIndex group.MemberIndex
+	signingGroupOperators   chain.Addresses
+
+	chainConfig *ChainConfig
+
+	attemptCounter    uint
+	attemptStartBlock uint64
+	attemptSeed       int64
+
+	randomRetryCounter uint
```
pdyraga, line 29: Why do we need `attemptCounter` and `randomRetryCounter`, can't we just use `attemptCounter` everywhere? I see `randomRetryCounter` is used in `qualifiedOperatorsSet` but I think we could use `attemptCounter` there as well.

pkg/tbtc/signing.go
```diff
@@ -0,0 +1,224 @@
...
+	randomRetrySeed    int64
```
pdyraga, line 30: I see `randomRetrySeed` is used in `qualifiedOperatorsSet` and `attemptSeed` is used for an additional trim. Can't we use `attemptSeed` everywhere? I see no security reasons for using two separate seeds and I must confess `attemptCounter` vs `randomRetryCounter` and `attemptSeed` vs `randomRetrySeed` was (and still is) the biggest point of confusion when reviewing this code.

pdyraga (commented):
I wasn't looking at tests yet but here is the first set of questions.

pkg/tbtc/signing.go
lukasz-zimnoch: The main reason is that `attemptCounter` counts from `1` and `randomRetryCounter` from `0` to match the retry algorithm expectations. However, looks like I misread the docs as this is true for keygen retry and basically doesn't matter for the signing retry logic. I'll change to use `attemptCounter` everywhere.

pkg/tecdsa/retry/retry.go
```diff
@@ -54,7 +54,10 @@ func EvaluateRetryParticipantsForSigning(
 		)
 	}
 	operatorToSeatCount := calculateSeatCount(groupMembers)
-	rand.Seed(seed + int64(retryCount))
```
pdyraga, line 57: I am reading the commit description explaining this change:

> Multiple concurrent members seeding the global RNG and using it to shuffle
operators list simultaneously cause non-deterministic results. Each call
to `EvaluateRetryParticipantsForSigning` must use a local fresh RNG instance
to work properly.

Was the source of the problem that the `Seed` was called multiple times and every new call, even though it's with the same value, leads to other results of `rand.Shuffle`?

pkg/tbtc/signing.go
lukasz-zimnoch: Basically, same response as in https://github.com/keep-network/keep-core/pull/3260#discussion_r971691910, and the separate seed was just for clear separation of those two random operations. I will change to use `attemptSeed` everywhere.

pkg/tecdsa/retry/retry.go
lukasz-zimnoch: The problematic scenario was:
- member 1 seeds the global RNG with seed A1 for attempt 1
- member 2 seeds the global RNG with seed A1 for attempt 1 and does the shuffle before member 1. It gets the list L of excluded members for attempt 1
- list L excludes member 2 for attempt 1 so member 2 prepares for attempt 2 by seeding RNG with seed A2
- member 1 does shuffling for attempt 1 but the RNG was already seeded with A2

Of course, this is the simplified version of the problem and the real course was much more complicated for >30 members doing that logic simultaneously

pkg/tbtc/signing_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/keep-network/keep-core/pkg/tecdsa/signing"
+	"math/big"
+	"reflect"
+	"testing"
+)
+
+func TestSigningRetryLoop(t *testing.T) {
+	chainConfig := &ChainConfig{
+		GroupSize:       10,
+		GroupQuorum:     8,
```
pdyraga, line 18: Do we need to set the quorum for signing retry tests? I think it is only important for DKG.

pkg/tbtc/signing.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3260/commits/aabe50e8f6a50c092282be5c31ccff0de6e0d61e

pkg/tbtc/signing.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3260/commits/aabe50e8f6a50c092282be5c31ccff0de6e0d61e

pkg/tbtc/signing_test.go
lukasz-zimnoch: We don't: 62f829d7

pkg/tbtc/signing.go
```diff
@@ -0,0 +1,224 @@
...
+	signingGroupMemberIndex group.MemberIndex,
+	signingGroupOperators chain.Addresses,
+	chainConfig *ChainConfig,
+) *signingRetryLoop {
+	// Pre-compute the 8-byte seed that is needed for the random
+	// retry algorithm. Since the original message passed as parameter
+	// can have a variable length, it is safer to take the first 8 bytes
+	// of sha256(seed) as the randomRetrySeed.
+	messageSha256 := sha256.Sum256(message.Bytes())
+	randomRetrySeed := int64(binary.BigEndian.Uint64(messageSha256[:8]))
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/commit/aabe50e8f6a50c092282be5c31ccff0de6e0d61e

pkg/tbtc/signing.go
```diff
@@ -0,0 +1,224 @@
...
+		includedMembers := make([]group.MemberIndex, 0)
+		excludedMembers := make([]group.MemberIndex, 0)
+		for i, operator := range srl.signingGroupOperators {
+			memberIndex := group.MemberIndex(i + 1)
+
+			if !qualifiedOperatorsSet[operator] {
+				excludedMembers = append(excludedMembers, memberIndex)
+			} else {
+				includedMembers = append(includedMembers, memberIndex)
+			}
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3260/commits/71a5bba81d9a6fe12ad1d639af1c3a5f8b6847f0

pkg/tecdsa/retry/retry.go
pdyraga: To rephrase and confirm my understanding is correct: previously, `EvaluateRetryParticipantsForSigning` used a global-state `rand`, so it was effectively stateful. At the same time, `EvaluateRetryParticipantsForSigning` could be used by multiple goroutines concurrently. Now, every call to `EvaluateRetryParticipantsForSigning` is stateless, because we initiate a new instance of `rng` with `rand.New`

Is this understanding correct?

I don't think it's blocking this PR but I think it would be great to add a unit test to `retry_test.go` with multiple goroutines that would cover this scenario.

pkg/tbtc/node.go
```diff
@@ -432,67 +433,81 @@ func (n *node) joinSigningIfEligible(
...
+
+							return nil, err
+						}
+
+						return result, nil
+					},
 				)
 				if err != nil {
 					logger.Errorf(
 						"[member:%v] signing of message [%v] "+
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 6b786cf8

pkg/tecdsa/retry/retry.go
lukasz-zimnoch: Yep, that's correct!

pkg/tbtc/signing.go
```diff
@@ -39,16 +36,12 @@ func newSigningRetryLoop(
...
 	chainConfig *ChainConfig,
 ) *signingRetryLoop {
-	// Pre-compute the 8-byte seed that is needed for the random
-	// retry algorithm. Since the original message passed as parameter
-	// can have a variable length, it is safer to take the first 8 bytes
-	// of sha256(seed) as the randomRetrySeed.
+	// Compute the 8-byte seed needed for the random retry algorithm. We take
+	// the first 8 bytes of the hash of the signed message. This allows us to
+	// not care in this piece of the code about the length of the message and
+	// how this message is proposed.
```
pdyraga, line 42: Can we do similar comment update in `pkg/tbtc/dkg.go`?

lukasz-zimnoch: See 916f3173

pkg/tbtc/signing.go
```diff
@@ -57,8 +50,6 @@ func newSigningRetryLoop(
 		attemptCounter:          0,
 		attemptStartBlock:       initialStartBlock,
 		attemptSeed:             attemptSeed,
-		randomRetryCounter:      0,
-		randomRetrySeed:         randomRetrySeed,
```
pdyraga: I'd love to do the same change in `pkg/tbtc/dkg.go`. We still have there `attemptCounter` and `randomRetryCounter`. This or separate PR?

lukasz-zimnoch: DKG is a separate story. We must have distinct counters as 1 attempt does not equal 1 retry algorithm run. We start with eliminating inactive members in the 5 first attempts and then use the retry algorithm.

pdyraga: Makes sense! Can we just add this explanation here:
https://github.com/keep-network/keep-core/blob/7b5a913e8f25c3ededf79c1bf2753a1b6af61a21/pkg/tbtc/dkg.go#L26-L30

lukasz-zimnoch: See 6c40e767

pdyraga (approved):
The code looks good and the tests are passing. I will test this functionality manually, together with the changes from #3273, #3267, #3274, and #3271.

PR #3267, opened 2022-09-12
tECDSA signing: Context logging

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-core/pull/3260

A single keep client can be part of multiple wallets that handle signing requests simultaneously (plus their retries). This can lead to messy logs where specific entries cannot be easily attributed to the specific protocol executions. As result, debugging can be really hard. Here we propose an improvement that is about introducing context logging for tECDSA signing and DKG. Instead of using the global `pkg/tbtc`'s logger that is passed down the way to all components, each signing (or DKG) instance now uses a dedicated logger that holds the execution context. This allows tying execution-specific information (like wallet, signed message, attempt index, etc.) to log messages. That way, we make debugging easier and allow for more sophisticated log processing using automated tools.


pkg/generator/pool.go
```diff
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/ipfs/go-log"
+	"github.com/ipfs/go-log/v2"
```
pdyraga, line 8: Shouldn't we replace it in *all* files?

pkg/tbtc/node.go
```diff
@@ -5,6 +5,7 @@ import (
 	"crypto/ecdsa"
 	"encoding/hex"
 	"fmt"
+	"go.uber.org/zap"
```
pdyraga: Where does this dependency come from in our codebase?

lukasz-zimnoch: It's the actual logger implementation used by `github.com/ipfs/go-log/v2`

pkg/generator/pool.go
lukasz-zimnoch: Not necessarily, `v2` is backward compatible so we can use it in places where needed.

PR #3268, opened 2022-09-13
Gitignore cleanup

nkuba:
In this PR we cleanup `.gitignore` file. Please see the commit's descriptions for details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,15 +1,14 @@
-node_modules/
 artifacts/
 **/gen/**/build/
 vendor/
 _local/
-*.swp
-*.swo
-*.DS_Store
 ,*
 *.out
 *.abi
 *.account
+
+# MacOS
+*.DS_Store
 
 # Executables
 debug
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,34 +1,25 @@
-node_modules/
-artifacts/
-**/gen/**/build/
-vendor/
-_local/
-*.swp
-*.swo
+# Hidden files
+.*
```
dimpar, line 2: This is a bit confusing. In the commit msg it says "all hidden files and directories", but we have some directories that we don't want to ignore like `.github, .gitmodules, .gitignore` and probably a couple of others. I'm not sure if we want to ignore all the hidden files & directories. Besides we double ignore dirs like: `.vscode` or `.idea` if we ignore all hidden.

nkuba: Good point! Normally I'd just ignore all dot-prefixed files and exclude the ones we don't want to be ignored:
```
.*
!.gitignore
!.dockerignore
```

But taking into account the number of dot-prefixed files we want to commit it will be easier to list the ones we want to ignore.

nkuba: e50512f594158c611537327df812df08cef86e94

dimpar (approved):

PR #3271, opened 2022-09-13
tECDSA signing: Address all outstanding review comments

lukasz-zimnoch:
Closes: https://github.com/keep-network/keep-core/issues/3042
Depends on: https://github.com/keep-network/keep-core/pull/3267

Here we address all outstanding non-blocking comments left during the review of all PRs listed in #3042. Fixes that were done here are only about renames and documentation improvements.


pkg/tecdsa/dkg/gen/pb/message.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
```
pdyraga, line 1: Let's merge `main` back to this branch once #3273 lands and regenerate this file to make sure we use the right proto version for that. In #3257 we updated the protobuf and now `make` should be used for generating files.

Same comment to all other `*.pb.go` files in this PR.

pdyraga (commented):
One comment about generated files. Other than that, it looks good.

pkg/tecdsa/dkg/gen/pb/message.pb.go
lukasz-zimnoch: This was generated using the new way presented in https://github.com/keep-network/keep-core/pull/3257

pdyraga: I confirmed it locally by merging `main` to this branch and re-generating. No differences.

pdyraga (approved):

PR #3273, opened 2022-09-13
TSS pre-params: Do not start tECDSA DKG if TSS pre-params pool is too small

lukasz-zimnoch:
Depends on: #3240

Here we add a check of the TSS pre-params pool size upon tECDSA DKG start. So far, the DKG started regardless of the TSS pre-params pool size which resulted in pre-params waste and DKG execution error in case the pool was too small. This change allows to graceful abort the DKG execution and preserve existing pre-params by the given node if their pool turns out to be too small to perform a successful DKG run.


pkg/tbtc/node.go
```diff
@@ -112,17 +112,29 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
 		}
 	}
 
-	// Create temporary broadcast channel name for DKG using the
-	// group selection seed with the protocol name as prefix.
-	channelName := fmt.Sprintf("%s-%s", ProtocolName, seed.Text(16))
+	if membersCount := len(indexes); membersCount > 0 {
+		if poolSize := n.dkgExecutor.PreParamsPool().CurrentSize(); membersCount > poolSize {
```
pdyraga, line 116: I am not sure if we should reach to DKG executor's pre-params pool from `tbtc` package. We have the executor abstraction to hide the complexity and internals. I would rather expose a function on the executor `IsReady` that would return bool and if not ready, logged the reason.

Please also see [this](https://github.com/keep-network/keep-core/pull/3240#discussion_r970048842) comment about diagnostics.

lukasz-zimnoch: I think the problem with the `IsReady` function is that the meaning of "ready" depends on the context. In the context of this PR, ready means: "executor has enough pre-params to cover all node members selected for DKG". On the other hand, in the context of https://github.com/keep-network/keep-core/pull/3274, ready means: "executor has a full pre-params pools". 

Also, note that the `tbtc` package is aware of the pre-params as it manages the configuration passed to the DKG executor so they are actually part of the execution abstraction. What I'd change is to remove `PreParamsPool()` from the executor API and replace it with `PreParamsCount()` to hide the details about the pre-params pool itself. 

As an alternative, we can try to introduce executor functions like `CurrentCapacity` that will indicate how many DKG instances (so members) can be triggered on the executor and `IsMaxCapacity` that will allow determining whether the pool is full. Personally, I think this will obfuscate the executor API a lot and we'll do better if we just return the current pre-params count.

pdyraga: > What I'd change is to remove PreParamsPool() from the executor API and replace it with PreParamsCount() to hide the details about the pre-params pool itself.

Agreed, I came up with the same idea after reading your response to [the discussion](https://github.com/keep-network/keep-core/pull/3240#discussion_r9700488420) about diagnostics.

Can we do this change here and also incorporate [this](https://github.com/keep-network/keep-core/pull/3240#discussion_r970049368) comment?

I would also eliminate `assembleTbtcDiagnostics` and passed the function assembling diagnostics as a parameter directly, without an intermediate variable.

lukasz-zimnoch: Here it is: 7e7f751c

pkg/tbtc/node.go
```diff
@@ -110,16 +110,27 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
 
-	// Create temporary broadcast channel name for DKG using the
-	// group selection seed with the protocol name as prefix.
-	channelName := fmt.Sprintf("%s-%s", ProtocolName, seed.Text(16))
+	if membersCount := len(indexes); membersCount > 0 {
+		if preParamsCount := n.dkgExecutor.PreParamsCount(); membersCount > preParamsCount {
+			dkgLogger.Infof(
+				"cannot join DKG as pre-parameters pool is too small; "+
+					"[%v] pre-parameters is required but only [%v] is "+
+					"available",
```
pdyraga, line 118:
```suggestion
				"cannot join DKG as pre-parameters pool size is too small; "+
					"[%v] pre-parameters are required but only [%v] are"+
					"available",
```

pdyraga (commented):
One log message change proposed and I am good to go.

pkg/tbtc/node.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3273/commits/f8a2c5c17885b1ee0641abce89c423b4bf089902

pdyraga (approved):

PR #3274, opened 2022-09-13
TSS pre-params: Do not join the sortition pool if TSS pre-params pool is too small

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-core/pull/3240

Here we enforce the sufficient size of the DKG pre-params pool before joining the sortition pool. Following this courtesy rule will make the network healthier as we increase the chance that only operable nodes join the sortition pool and
are selected for protocol executions.


pdyraga:
Really elegant solution 👌🏻

pdyraga (approved):

PR #3276, opened 2022-09-14
Etherscan verification fixes

michalinacienciala:
In the `Deploy contracts` step of the workflow we deploy contracts and verify them on Etherscan. We must configure `ETHERSCAN_API_KEY` variable, which is used to authenticate in Etherscan.
As part of this PR we also add Etherscan verification for `WalletRegistryGovernance` contract.


michalinacienciala:
Verified in https://github.com/keep-network/keep-core/actions/runs/3051207280

nkuba (approved):

PR #3149, opened 2022-08-10
Prepare monitoring solutions in Keep Test project

electropolis:
Refs #3012 
Refs #3279

 - added deployment for prometheus
 - added deployment for grafana
 - added storageclass to insert persistent storage using cloud provider
 - prepared PersistentVolumeClaim both for prometheus and grafana
 - prepared basic service using NodePort to use with Ingress
 - preaparing README file with instructions

PR #3277, opened 2022-09-14
3264 set version latest

electropolis:


nkuba:
Can you rebase this branch and remove the unrelated commits that are duplicated from https://github.com/keep-network/keep-core/pull/3149?

Also, could you please add a PR description?

infrastructure/kube/keep-test/geth-node/eth-goerli-node.yaml
```diff
@@ -44,7 +44,7 @@ spec:
     spec:
       containers:
         - name: geth-goerli
-          image: ethereum/client-go:v1.10.20
+          image: ethereum/client-go:latest
```
nkuba, line 47: I'm not sure about using the `latest` tag here. For example, Geth 1.10.22 had a bug that was fixed later in Geth 1.10.23. We should be aware of the version we use and ensure the version is stable before upgrading it.

nkuba (changes requested):

Issue #3264, opened 2022-09-12
update geth to the newest version


electropolis:
As a: member of Development team
I want to: Update geth version to the [v1.10.24](https://hub.docker.com/layers/ethereum/client-go/v1.10.24/images/sha256-af9a494538bb278325adda15776b63face4f4773f9170867e3c7101e6fe5d89a?context=explore) on K8S in GCP keep-test project
So I: will have the newest version of `geth` on test environment 

Acceptance Criteria:

1. Set in k8s manifest `v1.20.24` for docker image
2. Run and check if everything works

electropolis:
```
  Normal  Pulling    20s   kubelet            Pulling image "ethereum/client-go:latest"
  Normal  Pulled     18s   kubelet            Successfully pulled image "ethereum/client-go:latest" in 1.774689954s
  Normal  Created    18s   kubelet            Created container geth-goerli
  Normal  Started    18s   kubelet            Started container geth-goerli
```

```
/ # geth version
Geth
Version: 1.11.0-unstable
Git Commit: 818ff32ff5bf4674c5ee234f1b7a261148cf91b1
Git Commit Date: 20220913
Architecture: amd64
Go Version: go1.18.6
Operating System: linux
GOPATH=
GOROOT=go
```

pdyraga:
I am closing this issue as well as PR #3280. Upgrading the version is not enough and the entire new setup for Prysm is needed. For now, we will stick to external providers on testnet.

PR #3280, opened 2022-09-14
Update geth to the newest version

electropolis:
Refs #3264

  - changed image version from `1.10.20` to `1.10.24`

### DIFF
--- a/infrastructure/kube/keep-test/geth-node/eth-goerli-node.yaml
+++ b/infrastructure/kube/keep-test/geth-node/eth-goerli-node.yaml
@@ -44,7 +44,7 @@ spec:
     spec:
       containers:
         - name: geth-goerli
-          image: ethereum/client-go:v1.10.20
+          image: ethereum/client-go:v1.10.24
           ports:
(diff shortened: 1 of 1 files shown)


infrastructure/kube/keep-test/geth-node/eth-goerli-node.yaml
```diff
@@ -44,7 +44,7 @@ spec:
     spec:
       containers:
         - name: geth-goerli
-          image: ethereum/client-go:v1.10.20
+          image: ethereum/client-go:v1.10.24
```
pdyraga, line 47: @nkuba Do we need to run Prysm for testnet?

nkuba: Yes we do need to run it, see https://github.com/keep-network/keep-core/issues/3264#issuecomment-1246591600

pdyraga:
Closing without merging. Upgrading the version is not enough and the entire new setup for Prysm is needed. For now, we will stick to external providers on testnet.

PR #3282, opened 2022-09-16
Document deployments and contracts addresses

michalinacienciala:
We need one source of truth when it comes to listing deployment info and addresses of the contracts on different environments. We create a separate doc that would hold that information and would be linked to from other documents referencing its content. Thanks to this, after new deployment we'll need to update the deployment info only in one place.

Ref: https://github.com/threshold-network/solidity-contracts/pull/124


docs/deployments.adoc
```diff
@@ -0,0 +1,185 @@
...
+
+|WalletRegistry
+|link:https://goerli.etherscan.io/address/0x82BE0F8C8d43fAC584B03f4b782370E202A34527#writeProxyContract[0x82BE0F8C8d43fAC584B03f4b782370E202A34527]
+
+|Bridge
+|link:https://goerli.etherscan.io/address/0x52E861c35240d722e9EF7B605dC72C71ff8fc185#code[0x52E861c35240d722e9EF7B605dC72C71ff8fc185]
+
+|===
+
+== Ropsten Testnet
```
nkuba, line 134: I think we can get rid of ropsten.

docs/deployments.adoc
```diff
@@ -0,0 +1,185 @@
+:toc: left
+:toclevels: 3
+:sectanchors: true
+:sectids: true
+:source-highlighter: rouge
+:icons: font
+
+= Deployments and contracts addresses
```
nkuba, line 8: Maybe we shouldn't document all these here but in docs.threshold.network? https://docs.threshold.network/extras/contract-addresses

michalinacienciala: Yup, I moved the information to https://docs.threshold.network/extras/contract-addresses in https://github.com/threshold-network/threshold/commit/49395d6e3fe79e15647d78beeafaa3b0429971fe. I'll update the `keep-core` docs to link to that document.

michalinacienciala: Updated, see d4248f76b01ec41955ce5aa1427f33c16afc935f.

docs/deployments.adoc
michalinacienciala: Agree - I've included only Goerli testnet in the Threshold docs.

nkuba (approved):

docs/registration.adoc
```diff
@@ -127,4 +127,5 @@ page for the recent Mainnet addresses of the contracts.
 // TODO: Add section with the Keep Network contracts addresses to the Threshold Network docs.
```
nkuba, line 127: 1. In the line above we could link to the mainnet subpage directly.
2. We can remove this TODO.

nkuba (approved):

docs/registration.adoc
michalinacienciala: Fixed in https://github.com/keep-network/keep-core/pull/3381.

PR #3287, opened 2022-09-19
Validate PreParams recovered from a file

nkuba:
The tss-lib [panics if provided PreParams fail the validation](https://github.com/bnb-chain/tss-lib/blob/cbfa6cf63f18f471429eaab0a5f51cf72b7e9df8/ecdsa/keygen/local_party.go#L71-L73). To avoid the panic we use the same validation function to check if the given recovered PreParams set is valid, if not we won't add them to the PreParams pool.

The LocalPreParams struct from tss-lib defines a PaillierSK property.
This property is generated for PreParams and is required for DKG
execution. We need to marshal and unmarshal it to not lose the data.


pkg/tecdsa/dkg/marshaling.go
```diff
@@ -237,6 +238,13 @@ func (rsm *resultSignatureMessage) Unmarshal(bytes []byte) error {
 // Marshal converts the PreParams to a byte array.
 func (pp *PreParams) Marshal() ([]byte, error) {
 	localPreParams := &pb.PreParams_LocalPreParams{
+		PaillierSK: &pb.PreParams_PrivateKey{
+			PublicKey: &pb.PreParams_PublicKey{
+				N: pp.data.PaillierSK.N.Bytes(),
+			},
+			LambdaN: pp.data.PaillierSK.LambdaN.Bytes(),
+			PhiN:    pp.data.PaillierSK.PhiN.Bytes(),
+		},
```
pdyraga, line 247: We were not marshaling `PaillierSK` on purpose. If the given pre-params were not used for the key generation they should have no private key set. To me, it looks like a bug in tss-lib. The pre-params supplied to `keygen. NewLocalParty` should NOT have `PaillierSK ` set and now, the library expects the opposite.

What if we do the following:
1. Fill an issue to `bnb-chain/tss-lib`.
2. Set the `PaillierSK` to an empty `&pb.PreParams_PrivateKey{}` and post a comment in the code with a link to the issue and the explanation.

I think it should satisfy [this](https://github.com/bnb-chain/tss-lib/blob/cbfa6cf63f18f471429eaab0a5f51cf72b7e9df8/ecdsa/keygen/save_data.go#L63) check without us having to propagate the bug here.

nkuba: Mmmmkay!

pdyraga (approved):

PR #3290, opened 2022-09-20
Adding diagnostics scripts

dimpar:
> The script should query the bootstrap node for the list of connected peers and for each peer, extract information about the client version and the number of pre-parameters in the pool. Aggregate the result and publish it to a bucket on GCP

This PR handles the bash scripts part. It aggregates the results and publishes them to a GCP bucket.

Depends on https://github.com/keep-network/keep-core/pull/3255

There are two scripts that need to be run to publish peer data to the GCP bucket.
- `scripts/diagnostics.sh` - collects and aggregates the peers data and saves it locally under `diagnostics/peers_<timestamp>.json` file
- `scripts/upload_peers.sh --oauth2-token <path-to-oauth2-token> - uploads the `diagnostics/peers_<timestamp>.json` to the GCP bucket. The oauth2-token must be obtained from GCP and stored locally.

---
### Bucket creation
Cloud Storage -> Buckets -> Create. Follow the steps to create a bucket.

### Create a service account
- IAM & Admin -> Service Accounts -> Create Service Account -> follow the steps

### Making a bucket public
Go to the buckets list -> select a bucket -> edit access
Screenshot 2022-09-16 at 13 42 33

### Add a service account to a bucket
- Cloud Storage -> Buckets -> select a bucket -> Permissions tab -> Add a service account with "Storage Object Creator" role.

### List GCP accounts and set a service account
- `gcloud auth list`
- `gcloud config set account <SERVICE_ACCOUNT>`

### Get oauth2 token
- `gcloud auth print-access-token`
- or https://developers.google.com/oauthplayground/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,4 @@ yarn-error.log
 
 # Temporary directory
 tmp/
+diagnostics/
--- a/scripts/diagnostics.sh
+++ b/scripts/diagnostics.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+set -eou pipefail
+
+LOG_START='\n\e[1;36m'  # new line + bold + color
+LOG_END='\n\e[0m'       # new line + reset color
+DONE_START='\n\e[1;32m' # new line + bold + green
+DONE_END='\n\n\e[0m'    # new line + reset
+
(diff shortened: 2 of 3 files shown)


dimpar:
Closing if favor of https://github.com/keep-network/keep-core/pull/3290

PR #3291, opened 2022-09-20
Key generation and signing retry loop fixes

pdyraga:
# Wait in the retry loop for an attempt start block

When some member is in a signing group but is excluded from the current signing or key-generation attempt, they enter the state machine's `Execute` method and wait there on the block counter for an appropriate block height. The problem is that `m.channel.Recv(ctx, handler)` is called BEFORE `m.blockCounter.WaitForBlockHeight(startBlockHeight)`, so the `recvBuffer` gets populated while the member is blocked and the `recvBuffer` gets filled with no consumer reading from it. This leads to a bunch of justified `message handler is too slow; dropping message` warnings in the logs.

In this PR, we modify the retry loop for key generation and signing. The member is not supposed to enter the state machine before the attempt start block calculated by the retry loop. This way, no broadcast channel receiver is registered before it is really needed.

See https://github.com/keep-network/keep-core/commit/e9031bd11bb5bbcc2b8c2322193d75507395a272 for this change.

# Stop Pill workaround for no consensus integration

`StopPill` is a temporary workaround for a missing chain integration. When a group member is not selected for the current attempt of key generation or signing and there is no other member from the same client selected for the protocol execution, the member does not know what was the result of the protocol execution and if it completed or not. In other words, `defer cancelLoopCtx()` does not get executed and the member will stay hung on the block waiter, waiting for their turn. The `StopPill` is sent via broadcast channel on a successful protocol execution and tells all members waiting for their turn in the retry loop to stop because the result was produced.

See https://github.com/keep-network/keep-core/commit/9196573b32a737ef40164e2e972a0be833284e99 for this change.


go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/ipfs/go-datastore v0.5.1
```
pdyraga, line 22: For the changes in this file please see the commit message of https://github.com/keep-network/keep-core/pull/3291/commits/12a469082076cb9ae84441099e8bf370669211c8

pkg/tbtc/dkg.go
```diff
@@ -139,6 +136,23 @@ func (drl *dkgRetryLoop) start(
 			}
 		}
 
+		// Wait for the right moment to execute the attemptFn, as calculated
+		// in drl.attemptStartBlock.
+		err := waitForDkgAttemptFn(ctx, drl.attemptStartBlock)
+		if err != nil {
+			return nil, 0, fmt.Errorf(
+				"failed waiting on block [%v] for attempt [%v]: [%v]",
```
nkuba, line 144:
```suggestion
				"failed waiting for block [%v] for attempt [%v]: [%v]",
```

pkg/tbtc/dkg.go
```diff
@@ -139,6 +136,23 @@ func (drl *dkgRetryLoop) start(
...
+				drl.attemptStartBlock,
+				drl.attemptCounter+1,
```
nkuba, line 146: Why do we increment it? Shouldn't it match the value we pass to the `dkgAttemptFn` in line 162? It's already incremented in the `for` loop beginning.

pkg/tbtc/signing.go
```diff
@@ -64,11 +65,16 @@ type signingAttemptParams struct {
 // signingAttemptFn represents a function performing a signing attempt.
 type signingAttemptFn func(*signingAttemptParams) (*signing.Result, error)
 
+// waitForSigningAttemptFn represents a function blocking the attempt execution
+// until the given block height.
+type waitForSigningAttemptFn func(context.Context, uint64) error
```
nkuba, line 70: Same naming concerns as in my previous comment for `waitForDkgAttemptFn`.

pkg/tbtc/signing.go
```diff
@@ -168,6 +166,23 @@ func (srl *signingRetryLoop) start(
...
 		)
 
+		// Wait for the right moment to execute the signingAttemptFn, as
+		// calculated in srl.attemptStartBlock.
+		err := waitForSigningAttemptFn(ctx, srl.attemptStartBlock)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"failed waiting on block [%v] for attempt [%v]: [%v]",
+				srl.attemptStartBlock,
+				srl.attemptCounter+1,
```
nkuba, line 176: Why increment?

pkg/tbtc/dkg.go
```diff
@@ -75,11 +75,16 @@ type dkgAttemptParams struct {
...
+// waitForAttemptFn represents a function blocking the attempt execution until
+// the given block height.
+type waitForDkgAttemptFn func(context.Context, uint64) error
+
 // start begins the DKG retry loop using the given DKG attempt function.
 // The retry loop terminates when the DKG result is produced or the ctx
 // parameter is done, whatever comes first.
 func (drl *dkgRetryLoop) start(
 	ctx context.Context,
+	waitForDkgAttemptFn waitForDkgAttemptFn,
```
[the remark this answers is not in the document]
pdyraga: Yeah, good idea: d34a5741576431e7f45ad6a8b50a25f88f380cd3

pkg/tbtc/signing.go
pdyraga: d34a5741576431e7f45ad6a8b50a25f88f380cd3

pkg/tbtc/dkg.go
pdyraga: 14847eaa8932588bb9c6669f5427591e18618471

pkg/tbtc/dkg.go
pdyraga: Yes, this is wrong. Fixed in 749fd2a3cd0293fb605054a8d5c9010ba4948545.

pkg/tbtc/signing.go
pdyraga: 749fd2a3cd0293fb605054a8d5c9010ba4948545

nkuba (approved):

PR #3298, opened 2022-09-22
Set network properties for bootstrap nodes

nkuba:
We need to configure bootstrap nodes with publicly exposed announced addresses and make the service use a static IP predefined in GCP DNS config.

### DIFF
--- a/infrastructure/kube/keep-test/keep-client/keep-clients.yaml
+++ b/infrastructure/kube/keep-test/keep-client/keep-clients.yaml
@@ -86,7 +86,7 @@ spec:
         - --network.port
         - "3919"
         - --network.announcedAddresses
-        - /dns4/keep-client-0.default.svc.cluster.local/tcp/3919
+        - /dns4/keep-client-0.default.svc.cluster.local/tcp/3919,/dns4/bootstrap-0.test.keep.network/tcp/3919
         - --metrics.port
         - "9601"
         - --diagnostics.port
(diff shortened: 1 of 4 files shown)


infrastructure/kube/keep-test/keep-client/keep-clients.yaml
```diff
@@ -86,7 +89,7 @@ spec:
         - --network.port
         - "3919"
         - --network.announcedAddresses
-        - /dns4/keep-client-0.default.svc.cluster.local/tcp/3919
+        - /dns4/keep-client-0.default.svc.cluster.local/tcp/3919,/dns4/bootstrap-0.test.keep.network/tcp/3919
```
pdyraga, line 92: I take we announce both for test purposes?

pdyraga (approved):

infrastructure/kube/keep-test/keep-client/keep-clients.yaml
nkuba: Yes, the first one is for internal keep-test nodes, and the second one is for external participants.

pdyraga: Right, but the internal keep-test nodes could also connect using the second one, right? And we expose both to see if external and internal clients can handle the situation and choose the right address?

nkuba: Yes.
Also as you can see in the `data.yaml` we define `publicAnnouncedAddress` only for two nodes (the bootstrap nodes). Other nodes are using only the internal announced address.

I have plans to mix the cases even more, but will do that in the future.

PR #3301, opened 2022-09-23
Hardhat task to add an operator to beta operators set for the random beacon

pdyraga:
See https://github.com/keep-network/keep-core/pull/3297
See https://github.com/keep-network/sortition-pools/pull/188

After adding chaosnet feature to the sortition pool, we need to add an operator to the beta operators set during the operator initialization.

It is done here: hardhat task doing it is provided and it is attached to the initialize task.


tomaszslabon (approved):

dimpar:
lgtm, no comments

PR #3302, opened 2022-09-23
Hardhat task to add an operator to beta operators set for ecdsa

pdyraga:
~~Depends on https://github.com/keep-network/keep-core/pull/3301; Dependency update in `package.json` will be needed.~~

See https://github.com/keep-network/keep-core/pull/3297
See https://github.com/keep-network/sortition-pools/pull/188

After adding chaosnet feature to the sortition pool, we need to add an operator to the beta operators set during the operator initialization.

PR #3303, opened 2022-09-23
Chaosnet support in the client

pdyraga:
See https://github.com/keep-network/keep-core/pull/3301
See https://github.com/keep-network/keep-core/pull/3302

The code from these two PRs ☝🏻  is needed for testing this code. 

See https://github.com/keep-network/sortition-pools/pull/188

After adding chaosnet support to the sortition pool, we need to modify the client to check the chaosnet state before attempting to join the sortition pool.

`BetaOperatorPolicy` is a `JoinPolicy` implementation checking chaosnet and beta operator status. If chaosnet has been deactivated, the operator is allowed to join the pool. If chaosnet is active and the operator is beta operator, the operator is allowed to join the pool. If chaosnet is active and the operator is not beta operator, the operator is not allowed to join the pool.

The `BetaOperatorPolicy` is applied both to the random beacon and tbtc.


pkg/sortition/policy.go
```diff
@@ -0,0 +1,85 @@
+package sortition
+
+import "github.com/ipfs/go-log"
+
+// JoinPolicy determines how the client is supposed to join to the sortition
+// pool. The policy can encapsulate special conditions that the client want
```
tomaszslabon, line 6:
```suggestion
// pool. The policy can encapsulate special conditions that the client wants
```

pkg/sortition/policy.go
```diff
@@ -0,0 +1,85 @@
+package sortition
+
+import "github.com/ipfs/go-log"
+
+// JoinPolicy determines how the client is supposed to join to the sortition
```
tomaszslabon, line 5:
```suggestion
// JoinPolicy determines how the client is supposed to join the sortition
```

tomaszslabon (changes requested):

pkg/sortition/policy.go
pdyraga: 4acc73b7b69199645a3b03d046761e064d204fc8

pkg/sortition/policy.go
pdyraga: 4acc73b7b69199645a3b03d046761e064d204fc8

tomaszslabon:
It seems to be working as expected.
Run some basic tests: chaosnet was active, but operator that was not a beta operator could not join the sortition pool.

tomaszslabon (approved):

PR #3310, opened 2022-09-26
Merge diagnostics and metrics into a single clientinfo registry

pdyraga:
~~Depends on  #105~~

Previously, we were exposing diagnostics and metrics on separate ports but the reality was that we had two HTTP servers exposing exactly the same content: the diagnostics port could be queried for both /diagnostics and /metrics and the metrics port could be queried for both /metrics and /diagnostics. This change simplifies the setup into a single clientinfo package and registry exposing metrics and diagnostics.

From the operator's perspective the only change is the config file and the fact one less port is used by the client.


nkuba (approved):

pdyraga:
Reviewed changes in https://github.com/keep-network/keep-core/pull/3310/commits/a39e952699ce164a20b6daa7191b35081d36d098. They look great, thanks @nkuba !

PR #3319, opened 2022-09-28
Solidity 0.8.17

pdyraga:
See https://github.com/threshold-network/solidity-contracts/pull/127
See https://github.com/keep-network/sortition-pools/pull/189
See https://github.com/keep-network/keep-core/pull/3318

Fix Solidity version on 0.8.17 before the deployment.

### DIFF
--- a/solidity/ecdsa/hardhat.config.ts
+++ b/solidity/ecdsa/hardhat.config.ts
@@ -47,6 +47,15 @@ const config: HardhatUserConfig = {
   solidity: {
     compilers: [
       {
+        version: "0.8.17",
+        settings: {
+          optimizer: {
+            enabled: true,
+            runs: 200,
+          },
+        },
+      },
+      {
         version: "0.8.9",
         settings: {
           optimizer: {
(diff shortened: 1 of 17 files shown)


solidity/ecdsa/hardhat.config.ts
```diff
@@ -46,6 +46,15 @@ export const testConfig = {
...
     compilers: [
+      {
+        version: "0.8.17",
+        settings: {
+          optimizer: {
+            enabled: true,
+            runs: 200,
+          },
+        },
+      },
```
lukasz-zimnoch, line 57: Shouldn't we remove the config for `0.8.9`?

pdyraga: We can't. We still need 0.8.9 compiler to compile contracts from `@threshold-network/solidity-contracts` so instead of replacing 0.8.9 with 0.8.17 we have them both. Hardhat should pick the right compiler: https://hardhat.org/hardhat-runner/docs/advanced/multiple-solidity-versions#multiple-solidity-versions

tomaszslabon (approved):

lukasz-zimnoch (approved):

PR #3326, opened 2022-09-28
Build linux binaries on debian based Docker image

nkuba:
We were using alpine as a base for the binaries building which was causing some inconveniences when running the binary on ubuntu as it required `musl` library installation.
Alpine is using `musl` instead of `glibc`, so the binary build on alpine won't work right away on ubuntu.
In this PR we use a debian-based (bullseye) golang docker image to build the binaries.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,9 +96,13 @@ CMD []
 #
 # Build Binaries
 #
-FROM build-sources AS build-bins
+FROM golang:1.18.3-bullseye AS build-bins
+
+ENV APP_DIR=/go/src/github.com/keep-network/keep-core
 
 WORKDIR $APP_DIR
+
+COPY --from=build-sources $APP_DIR $APP_DIR
 
 ARG ENVIRONMENT
 


nkuba:
Another idea to solve the problem is building static binaries for Linux, just like geth does it: https://github.com/ethereum/go-ethereum/blob/master/build/ci.go#L265-L276

nkuba:
I downloaded the binary from the CI job execution https://github.com/keep-network/keep-core/actions/runs/3147365691 and was able to run it on Ubuntu 20.04.1

```
$ ldd keep-client
	linux-vdso.so.1 (0x00007ffc02eb4000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f2f54d33000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2f54b41000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f2f54d5f000)
```

pdyraga (approved):

Issue #3218, opened 2022-08-30
Bootstrap node should filter out itself from the bootstrap peers list


nkuba:
A node will use a list of embedded peers as bootstrap nodes for LibP2P connection initialization as per https://github.com/keep-network/keep-core/pull/3200.
If a bootstrap node tries to connect to itself it will log a warning:

> could not establish connection with bootstrap peer [16Uiu2HAmCcfVpHwfBKNFbQuhvGuFXHVLQ65gB4sJm7HyrcZuLttH]: [dial to self attempted]

We should filter out the node itself from the list of the bootstrap peers when establishing a connection.

PR #3342, opened 2022-10-04
Made keep-client exclude own address from bootstrap peer list

tomaszslabon:
Closes https://github.com/keep-network/keep-core/issues/3218.

This PR adds a check for the keep-client's own address on the bootstrap peer list.
If own address is found, it is ignored and the bootstrapping continues.
This change prevents a constantly appearing warning in keep-client's logs.

### DIFF
--- a/pkg/net/libp2p/libp2p.go
+++ b/pkg/net/libp2p/libp2p.go
@@ -479,8 +479,8 @@ func (p *provider) bootstrap(
 	ownID := p.identity.id
 	filteredPeerInfos := make([]peer.AddrInfo, 0)
 
-	// If the client's own address is mistakenly added to the list of bootstrap
-	// peers, filter it out. This will prevent self-dialing.
+	// If the client's own address is on the list of bootstrap peers, filter it
+	// out to prevent self-dialing.
 	for _, peerInfo := range peerInfos {
 		if peerInfo.ID != ownID {
 			filteredPeerInfos = append(filteredPeerInfos, peerInfo)
--- a/pkg/net/libp2p/libp2p.go
+++ b/pkg/net/libp2p/libp2p.go
@@ -476,7 +476,21 @@ func (p *provider) bootstrap(
 		return err
 	}
 
-	bootstrapConfig := bootstrapConfigWithPeers(peerInfos)
+	ownID := p.identity.id
+	filteredPeerInfos := make([]peer.AddrInfo, 0)
+
+	for _, peerInfo := range peerInfos {
+		if peerInfo.ID == ownID {
+			logger.Infof(
+				"Found own address [%v] on bootstrap peer list. Ignoring it.",
+				peerInfo.ID,
+			)
+		} else {
+			filteredPeerInfos = append(filteredPeerInfos, peerInfo)
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/net/libp2p/libp2p.go
```diff
@@ -476,7 +476,21 @@ func (p *provider) bootstrap(
...
-	bootstrapConfig := bootstrapConfigWithPeers(peerInfos)
+	ownID := p.identity.id
+	filteredPeerInfos := make([]peer.AddrInfo, 0)
+
+	for _, peerInfo := range peerInfos {
+		if peerInfo.ID == ownID {
+			logger.Infof(
+				"Found own address [%v] on bootstrap peer list. Ignoring it.",
+				peerInfo.ID,
+			)
```
lukasz-zimnoch, line 487: Does it make sense to log that information at all? IMO, it's enough to just `continue` in that case.

tomaszslabon: I wasn't sure whether to add the log or not.
I added it, because the operator may be surprised that we silently remove one item from their configuration file.
If we leave the log, they will know what happened.

tomaszslabon: After thinking about it more, I changed my mind. 
Let's not overwhelm the operator with unnecessary logs.
Removed the log and added a comment in 27d52e5125bd04b52697d25e464c9c5e112191cd.

pkg/net/libp2p/libp2p.go
```diff
@@ -476,7 +476,18 @@ func (p *provider) bootstrap(
 		return err
 	}
 
-	bootstrapConfig := bootstrapConfigWithPeers(peerInfos)
+	ownID := p.identity.id
+	filteredPeerInfos := make([]peer.AddrInfo, 0)
+
+	// If the client's own address is mistakenly added to the list of bootstrap
+	// peers, filter it out. This will prevent self-dialing.
```
pdyraga, line 483: I think it's not a mistake. The list of bootstrap peers is hardcoded in the client.

tomaszslabon: Ok, let's not call it a mistake.
I changed the comment a little in 7dd1bc2d3292e11fbf6294cd557009f342829571.

BTW, I haven't seen the hardcoded peer addresses yet.
I thought you can set this list either via configuration parameter or via ".../_peers/..." file (I think we use the latter in the testnet network).

pdyraga:
👀

pkg/net/libp2p/libp2p.go
pdyraga: > BTW, I haven't seen the hardcoded peer addresses yet.
I thought you can set this list either via configuration parameter or via ".../_peers/..." file (I think we use the latter in the testnet network).

They are hardcoded for [testnet](https://github.com/keep-network/keep-core/blob/f0b08217226bdefd96455c162fffdb8ef9b9618e/config/_peers/goerli) and [mainnet](https://github.com/keep-network/keep-core/blob/f0b08217226bdefd96455c162fffdb8ef9b9618e/config/_peers/mainnet).

pdyraga:
Tested locally with two peers pointing to self and each other. Works as advertised.

pdyraga (approved):

Issue #3345, opened 2022-10-04
Mechanism for computing chaosnet rewards


pdyraga:
What we are expecting from chaosnet participants:

0. Random Beacon and tBTC applications authorized.
1. The client is running on a machine meeting hardware requirements.
2. The client has a high uptime.
3. The client is up to date.
4. The client is connected to the network.
5. The client is exposing metrics.

Technically, we can translate it to something like:

0. Check authorizations on the smart contract side.
1. At least 100 pre-parameters in the pool.
2. At least 96% uptime: it means the node can be down for a maximum of 24h total over the course of a month.
3. We allow for a one/two-week delay in updates to the most recent version.
4. Bootstrap nodes see the client connected and Prometheus can discover the node.
5. Metrics are exposed publicly.

Once the requirements are checked, the reward calculation mechanism will have to be updated. See https://github.com/threshold-network/merkle-distribution

PR #3346, opened 2022-10-04
Beta operators rewards

dimpar:
Refs #3345

This PR calculates rewards for beta tBTCv2 operators. The output rewards JSON file will be used as input in https://github.com/threshold-network/merkle-distribution to generate a merkle distribution tree.

There are 5 requirements for an operator to receive rewards in a given rewards interval:
- Random Beacon authorized
- tBTC authorized
- Total node(s) uptime 96% across all the instances after a peer joins the network. See examples
- Pre-generated params avg of 500 across all the instances. See examples
- Latest client build version. A 2-week update delay is allowed after the latest release.

E.g. 1
```
<operator_address>:
  "instances":
        "10.102.1.33:9701": {
          "upTimePercent": 50,
          "preParams": 400,
          "version": "v2.0.0-m1-9-g4049dc015"
        }
        "10.102.2.20:9601": {
          "upTimePercent": 46,
          "preParams": 600,
          "version": "v2.0.0-m1-9-g4049dc015"
        }
}
"upTimePercent" - total: 96% meets a 96% req across all the instances
"preParams" - total avg. 500 meets a 500 avg across all the instances
```

E.g. 2
```
<operator_address>:
  "instances":
        "10.102.1.33:9701": {
          "upTimePercent": 50,
          "preParams": 399,
          "version": "v2.0.0-m1-9-g4049dc015"
        }
        "10.102.2.20:9601": {
          "upTimePercent": 45,
          "preParams": 600,
          "version": "v2.0.0-m1-9-g4049dc015"
        }
}
"upTimePercent" - total: 95% does not meet a 96% req across all the instances
"preParams" - total avg. 499.5 does not meet a 500 avg across all the instances
```

`rewards.ts` will:
- call beacon and wallet registry contracts for authorization data. It calculates "weighted" authorization based on increase and decreases events during the interval. E.g.
```
- rewards interval Sep 1 - Sep 30
- beacon and tBTC authorization of 100k happened on Sep 1
- beacon and tBTC authorization increase event happened on Sep 10. It was increased to 150k

uptimeCoeficient: off-chain client's uptime
beaconWeightedAuthorization = 100k * 10/30 + 150k * 20/30
tbtcWeightedAuthorization = 100k * 10/30 + 150k * 20/30
weightedAuthorization = uptimeCoeficient * min(beaconWeightedAuthorization, tbtcWeightedAuthorization)
```
- query [Prometheus service](https://monitoring.test.threshold.network/prometheus/graph) for bootstrap peers' info
- aggregate and validate rewards requirements for each operator
- save rewards data for eligible operators

<details>
  <summary>Example operator details to calculate final rewards</summary>

```
   "0x5cD847903Bb7F29de77Eecc135628cA5b104A355": {
      "applications": {
        "beaconAuthorization": "800000000000000000000000",
        "tbtcAuthorization": "800000000000000000000000"
      },
      "instances": {
        "10.102.1.19:9601": {
          "upTimePercent": 74.52969075397857,
          "avgPreParams": 922,
          "version": "v2.0.0-m1-9-g4049dc015"
        },
        "10.102.1.77:9601": {
          "upTimePercent": 24.81376522052997,
          "avgPreParams": 971,
          "version": "v2.0.0-m1-9-g4049dc015"
        }
      },
      "requirements": {
        "isBeaconAuthorized": true,
        "isTbtcAuthorized": true,
        "isUptimeSatisfied": true,
        "isPreParamsSatisfied": true,
        "isVersionSatisfied": true
      }
    }
```

</details>

If all of the requirements are satisfied, then the rewards can be calculated with the following formula: 
```
amount = uptimeCoeficient * min(beaconWeightedAuthorization, tbtcWeightedAuthorization) * APR/12
```
APR = 15%

Example output format that is used to generate a merkle distribution tree. A parent key is a staking provider address, not operator. 

```
  {
    "0x0eC14BC7cCA82c942Cf276F6BbD0413216dDB2bE": {
      "beneficiary": "0x0eC14BC7cCA82c942Cf276F6BbD0413216dDB2bE",
      "amount": "9900000000000000000000"
    }
  },
```

### How to run
```
./rewards.sh --rewards-start-date Y-m-d --rewards-end-date Y-m-d --etherscan-token "<token>"
```

For more details that include optional flags please run `./rewards.sh --help`


beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,229 @@
...
+    const queryUptime = `${prometheusAPI}/query`;
+    // Offset is set in case the end time interval is not aligned with execution
+    // of this script. It "goes" back in time relevant to the current time.
+    const offset = Math.floor(Date.now() / 1000) - endRewardsTimestamp;
+    // Sum of all uptimes since the endpoint became available in a given
+    // timeframe. "up" metric won't take into account when a node wasn't available,
+    // hence we need to multiply the result by the scrape interval
+    // (set in the config file) and divide by the uptime search range.
+    const paramsUptime = {
+      query: `sum_over_time(up{instance='${peer.metric.instance}', job='${prometheous_job}'}[${uptimeSearchRange}s] offset ${offset}s) * ${scrapeInterval} / ${uptimeSearchRange}`,
```
beaushinkle, line 142: Can we multiline this like:
```
query: `sum_over_time(up{instance='${peer.metric.instance}', job='${prometheous_job}'}` +
  `[${uptimeSearchRange}s] offset ${offset}s) * ${scrapeInterval} / ${uptimeSearchRange}`,
```

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,229 @@
...
+  const queryStep = 600; // 10min in sec
+  const prometheous_job = options.job;
+  const prometheusAPI = options.api
+  const startRewardsTimestamp = parseInt(options.start);
+  const endRewardsTimestamp = parseInt(options.end);
+  const scrapeInterval = parseInt(options.interval);
+  const rewardsInterval = endRewardsTimestamp - startRewardsTimestamp;
+  const requiredUptime = 96; // percent
+  const preParamsAvgInterval = 24 // hours
+  const preParamsResolution = 24 // hours
```
beaushinkle, line 52: there's a small number of lines that don't end in semicolons; i don't know if the linter is built to catch/care about that kind of thing, or even if we care, but i figure if you went to the trouble of putting semicolons in a bunch of places you'd appreciate them being everywhere :D

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,229 @@
...
+  const minPreParams = 500 // min requirement for pre params daily avg
+  const factors = {
+    isBeaconAuthorized: "isBeaconAuthorized",
+    isTbtcAuthorized: "isTbtcAuthorized",
+    upTime: "upTime",
+    version: "verion",
```
beaushinkle, line 58:
```suggestion
    version: "version",
```

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,229 @@
...
+  const prometheusAPI = options.api
+  const startRewardsTimestamp = parseInt(options.start);
+  const endRewardsTimestamp = parseInt(options.end);
+  const scrapeInterval = parseInt(options.interval);
+  const rewardsInterval = endRewardsTimestamp - startRewardsTimestamp;
+  const requiredUptime = 96; // percent
+  const preParamsAvgInterval = 24 // hours
+  const preParamsResolution = 24 // hours
+  const minPreParams = 500 // min requirement for pre params daily avg
+  const factors = {
```
beaushinkle, line 54: As far as I can tell, this `factors` object is never used as a collection, rather it's used to group some related constants. As in, later we might call: `peerData.set(factors.isBeaconAuthorized, 0);` or `peerData.set(factors.upTime, upFactor);`, and so having `factors.upTime` defined saves us from typing out `peerData.set("upTime", upFactor)`, which would be a magic string.

What do you think about taking this one step further, defining a file like `reward-constants.ts` and exporting a bunch of ALL_CAPS constants so their purpose is *super clear*?

Then, this file would be calling `peerData.set(UP_TIME, upFactor)`, and `if (Number(peerPreParams[1]) < MIN_PRE_PARAMS) {`

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,229 @@
...
+    isBeaconAuthorized: "isBeaconAuthorized",
+    isTbtcAuthorized: "isTbtcAuthorized",
+    upTime: "upTime",
+    version: "verion",
+    preParams: "preParams",
+  };
+  const upTimeRewardsCoefficient = "upTimeRewardsCoefficient"
+  const peersDataFile = options.output
```
beaushinkle, line 62: I would group together all of the options parsing into tight block

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,229 @@
...
+      peerData.set(factors.isTbtcAuthorized, 1);
+    }
+
+    /// Up time requirement
+
+    // First registered 'up' metric in a given timeframe <start:end>. We start
+    // evaluating uptime from this point.
+    const firstRegisteredUptime = peer.values[0][0];
+    const uptimeSearchRange = endRewardsTimestamp - firstRegisteredUptime;
+    const queryUptime = `${prometheusAPI}/query`;
```
beaushinkle, line 133: multiple variables are bound to this same string: `${prometheusAPI}/query`

I think we can rename it something generic and only bind it once

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,229 @@
...
+  const jsonObject = await convertToJSON(peersData);
+  // Save to file
+  fs.writeFileSync(
+    peersDataFile,
+    JSON.stringify(jsonObject, null, 2)
+  )
+}
+
+async function convertToJSON(map: Map<string, Map<string, any>>) {
+  var json: {[k:string]: any} = {};
```
beaushinkle, line 204:
```suggestion
  let json: {[k:string]: any} = {};
```
doesn't matter here, but i think it's good practice in general to be in the habit of using `let` instead of `var`. `var` has a really strange scope: https://stackoverflow.com/questions/762011/what-is-the-difference-between-let-and-var

beaushinkle:
summarizing the broad approach to make sure I understand:

1) we call `rewards.sh` from arguments somewhere (from a service, manually, etc). That parses arguments, and eventually...
2) we call `beta-operator-rewards/rewards-requirements.ts` with the data. We parse all the arguments passed from `rewards.sh`, and interpolate prometheus queries.
3) we run `queryPrometheus(url: string, params: any)` on those queries (uptime, pre-param count, build version)
4) we fetch the other two (authorized random beacon, authorized tbtc) from querying the relevant contracts directly via `ethers` and having the ABIs
5) we save all of those results into an object named `peerData` and return it

Seems like a direct, lightweight approach, and I don't see why we'd need to engineer it any more than this! I think so long as we make sure the file itself is well structured, it won't be a problem maintaining it going forward either 👍

pdyraga:
I did not look at the code, leaving this part to @beaushinkle but I have two comments regarding the structure of the JSON output:
1. The key is the IP address. I think this is wrong because the IP address may change (e.g. cluster relocation). I think we should use the operator address as a key instead. Alternatively, consider using staking provider address as a key. This needs to be agreed upon with the Nu team.
2. We do have the `upTimeRewardsCoefficient` but this is not enough. Someone who has 1M T staked and only 100k  T authorized for random beacon and tBTC should not receive rewards for 1M T. Instead of `upTimeRewardsCoefficient` I would consider exposing unsigned integer `rewardWeight` that would be calculated as `rewardWeight = uptime_coefficient * AVG(beacon_authorization, ecdsa_auhtorization)`.

dimpar:
> The key is the IP address. I think this is wrong because the IP address may change (e.g. cluster relocation). I think we should use the operator address as a key instead. Alternatively, consider using staking provider address as a key. This needs to be agreed upon with the Nu team.

This can be easily changed and that was my original approach actually, but I noticed that we have cases where the same address is on two different IPs like here:
```
"10.102.4.11:9601": {
    "address": "0x3FF855895EF4aC833c32Ab6A0d6C7fBfA137E26E",
    "isBeaconAuthorized": 1,
    "isTbtcAuthorized": 1,
    "upTime": 0,
    "upTimeRewardsCoefficient": 0,
    "preParams": 1
  },
  "10.102.4.20:9601": {
    "address": "0x3FF855895EF4aC833c32Ab6A0d6C7fBfA137E26E",
    "isBeaconAuthorized": 1,
    "isTbtcAuthorized": 1,
    "upTime": 1,
    "upTimeRewardsCoefficient": 0.1388888888888889,
    "preParams": 0
  },
(...)
```

How do we deal with that? What if one peer goes down and the other stays up?

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,229 @@
...
+    .version("0.0.1")
+    .requiredOption(
+      "-s, --start <timestamp>",
+      "starting time for rewards calculation"
+    )
+    .requiredOption(
+      "-e, --end <timestamp>",
+      "ending time for rewards calculation"
+    )
+    .requiredOption("-i, --interval <timestamp>", "scrape interval") // IMPORTANT, it has to be the same as in prometheous config
```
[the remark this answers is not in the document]
dimpar: Fixed, as part of [982a1f0](https://github.com/keep-network/keep-core/pull/3346/commits/982a1f01cbd3036fa4c035d697edb323b75493c6)

beta-operator-rewards/rewards-requirements.ts
dimpar: This is my linter adding `;`. I didn't add our custom keep-core linter because I wanted to keep things minimal for now. I don't have a strong preference on adding `;` or not, but to keep things consistent I run linter again which added `;`. As part of [982a1f0](https://github.com/keep-network/keep-core/pull/3346/commits/982a1f01cbd3036fa4c035d697edb323b75493c6)

beta-operator-rewards/rewards-requirements.ts
dimpar: Fixed as part of [982a1f0](https://github.com/keep-network/keep-core/pull/3346/commits/982a1f01cbd3036fa4c035d697edb323b75493c6)

beta-operator-rewards/rewards-requirements.ts
dimpar: Sorry, I'm not sure I'm following, can you pls give an example?

beta-operator-rewards/rewards-requirements.ts
dimpar: Sure, as part of [982a1f0](https://github.com/keep-network/keep-core/pull/3346/commits/982a1f01cbd3036fa4c035d697edb323b75493c6)

beta-operator-rewards/rewards-requirements.ts
dimpar: lol.. I think I was experimenting with smth and just straight-up forgot to change it. Yes, I changed it back to `let` as part of [982a1f0](https://github.com/keep-network/keep-core/pull/3346/commits/982a1f01cbd3036fa4c035d697edb323b75493c6)

beta-operator-rewards/rewards-requirements.ts
beaushinkle: sure!

```typescript
  // Parse the program options
  const options = program.opts();
  const prometheus_job = options.job;
  const prometheusAPI = options.api;
  const clientVersions = options.versions;
  const startRewardsTimestamp = parseInt(options.start);
  const endRewardsTimestamp = parseInt(options.end);
  const scrapeInterval = parseInt(options.interval);
  const peersDataFile = options.output;
  // End program option parsing
```

beaushinkle: Addressed in 734c3b4e9cd57d5a03e29b90e1037a3a4ec4ae33

beaushinkle: Addressed in 734c3b4e9cd57d5a03e29b90e1037a3a4ec4ae33

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,478 @@
...
+}
+
+// Calculates the weighted authorization for rewards interval based on events.
+// The general idea of weighted rewards calculation describes the following example.
+// Please note that this example operates on dates for simplicity purposes,
+// however the actual calculation is based on block numbers.
+// Ex.
+// events:         ^     ^      *    *   ^
+// timeline:  -|---|-----|------|----|---|--------|-->
+//         Sep 0   3     8      14  18   22       30
```
beaushinkle, line 360: this graph is awesome

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,478 @@
...
+) {
+  if (events.length == 0) {
+    return BigNumber.from("0");
+  }
+
+  let authorization = BigNumber.from("0");
+  const deltaRewardsBlock = endRewardsBlock - startRewardsBlock;
+  // ascending order
+  events.sort(function (a, b) {
+    return a.blockNumber < b.blockNumber
```
beaushinkle, line 385: I think this can be simplified to `events.sort((a,b) => a.blockNumber - b.blockNumber)`

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,478 @@
...
+      ? -1
+      : a.blockNumber == b.blockNumber
+      ? 0
+      : 1;
+  });
+
+  let tmpBlock = startRewardsBlock; // prev tmp block
+  let event;
+  for (let i = 0; i < events.length; i++) {
+    event = events[i];
```
beaushinkle, line 395: I've never actually checked to see what the javascript virtual machine does here, though now I want to know!

Is 
```js
let foo;
for (...) {
  foo = bar()
}
```
more efficient than
```
for (...) {
  let foo = bar()
}
```

It would be in really old languages since we have to allocate `foo` in each loop, but I don't know if the compiler makes the optimization for us in JS or not. I found https://stackoverflow.com/questions/50807989/es6-javascript-const-inside-or-let-outside-loop which is inconclusive.

For the sake of style! I vote for:

```js
for (let i = 0; i < events.length; i++) {
  const event = events[i];
}
```
sine we're not changing the value of event in the loop or using it outside the loop

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,478 @@
...
+    const coefficientStr = (
+      ((event.blockNumber - tmpBlock) / deltaRewardsBlock) *
+      PRECISION
+    ).toString();
+    const coefficient = parseInt(coefficientStr);
```
beaushinkle, line 400: can we bind this directly? I don't see us using `coefficientStr` anywhere else:

```js
    const coefficient = ((event.blockNumber - tmpBlock) / deltaRewardsBlock) *
      PRECISION
```

Same with below

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,478 @@
...
+  });
+
+  return await authorizationForRewardsInterval(
+    applicationEvents,
+    startRewardsBlock,
+    endRewardsBlock
+  );
+}
+
+async function filterEventsByApplications(events: any[]) {
```
beaushinkle, line 323: I don't think this function needs to be `async`

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,478 @@
...
+  const filteredEvents = new Array();
+  for (let j = 0; j < events.length; j++) {
```
beaushinkle, line 325: I think we can condense this with the built in `filter` function:

```js
return events.filter(event => {
  return event.args.application == RandomBeaconAddress ||
    event.args.application == WalletRegistryAddress
})
```

beta-operator-rewards/rewards-requirements.ts
dimpar: Good point! [7dc0e10](https://github.com/keep-network/keep-core/pull/3346/commits/7dc0e104023322be686fe08e91ed0d39ec282e57) and [a6acabe](https://github.com/keep-network/keep-core/pull/3346/commits/a6acabe161a73f65ca15b263a551620c4488e12b)

beta-operator-rewards/rewards-requirements.ts
dimpar: I've actually moved `event` outside the loop to reference it while calculating the last sub-interval authorization `event.arg.amountTo`, but I agree, it can be written nicer. I've just referenced the last event as `events[events.length - 1].args.toAmount.` to calculate the `authorization`. As part of [4505ecf](https://github.com/keep-network/keep-core/pull/3346/commits/4505ecf04eda7051cc0ddea2e336107d3c292fd2)

beta-operator-rewards/rewards-requirements.ts
dimpar: The idea was to make a floating number like `42.0` to an integer `42` because BigNumber doesn't support floating numbers. I used now `Math.floor` instead of the previous approach which I agree was over-engineered. As part of [4505ecf](https://github.com/keep-network/keep-core/pull/3346/commits/4505ecf04eda7051cc0ddea2e336107d3c292fd2)

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,478 @@
...
+//         0 -> Sep 1 at 00:00:00
+// event authorizations:
+//  Sep 0 - 3 from 100k to 110k
+//  Sep 3 - 8 from 110k to 135k
+//  Sep 8 - 14 from 135k to 120k
+//  Sep 14 - 18 from 120k to 100k
+//  Sep 18 - 30 constant 100k (last sub-interval)
+// authorization = (3-0)/30*100 + (8-3)/30*110 + (14-8)/30*135
+//               + (18-14)/30*120 + (30-18)/30*100
+async function authorizationForRewardsInterval(
```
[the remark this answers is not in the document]
dimpar: Removed as part of [84e3d1e](https://github.com/keep-network/keep-core/pull/3346/commits/84e3d1ec956a618e8dba70d40f5f305b2a03db5d)

beta-operator-rewards/rewards-requirements.ts
dimpar: Removed as part of [84e3d1e](https://github.com/keep-network/keep-core/pull/3346/commits/84e3d1ec956a618e8dba70d40f5f305b2a03db5d)

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,478 @@
...
+
+  return filteredEvents;
+}
+
+async function filterEventsByStakingProvider(
+  events: any[],
+  stakingProvider: string
+) {
+  const filteredEvents = new Array();
+  for (let j = 0; j < events.length; j++) {
```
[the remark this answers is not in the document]
dimpar: Agree! [1e4b765](https://github.com/keep-network/keep-core/pull/3346/commits/1e4b765b1cd76d693ca127134e68948300e4a0e7)

beta-operator-rewards/rewards-requirements.ts
dimpar: Agree! [1e4b765](https://github.com/keep-network/keep-core/pull/3346/commits/1e4b765b1cd76d693ca127134e68948300e4a0e7)

beta-operator-rewards/rewards-requirements.ts
dimpar: image

beta-operator-rewards/rewards-requirements.ts
```diff
@@ -0,0 +1,529 @@
...
+            [${rewardsInterval}s:${UPTIME_QUERY_RESOLUTION}s] offset ${offset}s`,
+  };
+
+  const instancesByOperator = (
+    await queryPrometheus(prometheusAPIQuery, paramsOperatorUptime)
+  ).data.result;
+
+  // First registered 'up' metric in a given interval <start:end> for a given
+  // operator. Start evaluating uptime from this point.
+  const firstRegisteredUptime = instancesByOperator[0].values[0][0];
```
beaushinkle, line 340: is it always the case that the first instance is also has the first registered uptime?

dimpar: Good point! I checked again on Prometheus, and indeed it looks like the returned instances are not sorted for a given operator. Instances for a given operator can be in a random order (or at least not in ascending order). The first time I looked at the Prometheus graph which had sorted data and I mistakenly made an assumption that returned instances are sorted by their timestamps. However, timestamps within a single instance are sorted from my observation. Added reduce function [f3322a5](https://github.com/keep-network/keep-core/pull/3346/commits/f3322a565c435851cb0a7ec01481034d07b3f7c6)

beta-operator-rewards/rewards.ts
```diff
@@ -0,0 +1,693 @@
...
+    //                       v1 or v2                v2
+    //           |-------------------------------\|-------|
+    // Timeline -|-------------*------------------*-------|->
+    //         Sep1            v2             v2+delay   Sep30
+    // Where:
+    // v2 was released Sep10
+    // delay = 14days
+    // v2 + delay = Sep10 + 14days = Sep24
+    // Between Sep1 - Sep24 a client is allowed to run v1 or v2
+    // Between Sep24 - Sep30 a client is allowed to run only v2
```
beaushinkle, line 294: Can you move the version validation logic out to its own function? I think it's complex enough that it deserves its own call

beaushinkle:
Here's a little audit on the prometheus queries:


```
1) present_over_time(up{chain_address="0x0954efEFeb970D317a51736201B4EB2De75FF5DE", job="keep-discovered-nodes"} [2591999s] offset 74124s)
2) up{chain_address="0x0954efEFeb970D317a51736201B4EB2De75FF5DE", job="keep-discovered-nodes"} [2591999s:120s] offset 74124s
3) sum_over_time(up{chain_address="0x0954efEFeb970D317a51736201B4EB2De75FF5DE", job="keep-discovered-nodes"} [364679s:120s] offset 74124s) * 120 / 364679
4) avg_over_time(tbtc_pre_params_count{chain_address="0x0954efEFeb970D317a51736201B4EB2De75FF5DE", job="keep-discovered-nodes"} [2591999s:120s] offset 74124s)
5) client_info{chain_address="0x0954efEFeb970D317a51736201B4EB2De75FF5DE", job="keep-discovered-nodes"}[2591999s:120s] offset 74124s
```

I switched out the `chain_address` to point at `0x0954efEFeb970D317a51736201B4EB2De75FF5DE` because it's an operator with a lot going on and it's on testnet.

Running those above queries in the UI: https://monitoring.test.threshold.network/prometheus/graph

We get

1)
image

2) 
image

3) 
image

4)
image

5) 
image

Here's how to read these:

`present_over_time(up{chain_address="0x0954efEFeb970D317a51736201B4EB2De75FF5DE", job="keep-discovered-nodes"} [2591999s] offset 74124s)`

We can read these as an inner query and outer query. The inner query is `up{chain_address="0x0954efEFeb970D317a51736201B4EB2De75FF5DE", job="keep-discovered-nodes"} [2591999s] offset 74124s`, which has a couple things going on. The base is `up{chain_address="0x0954efEFeb970D317a51736201B4EB2De75FF5DE", job="keep-discovered-nodes"}`, which means "return the `up` data where the chain address is 0x0954efEFeb970D317a51736201B4EB2De75FF5DE and the job is keep-discovered-nodes"

Then, we specify a time range: `2591999s`, which is ~30 days. So that part says "run this `up` query on the data for the last 30 days".

Finally, we supply an offset of 74124s which is 20.6 hours. This shifts our range backwards, and makes the query "run this `up` query on 30 days worth of data starting 20.6 hours ago".

Then, we aggregate that data via the `present_over_time` function which return a value of 1 if there is any data present in the time interval.

All of the other functions follow a similar format with one addition: the resolution param. 

```
sum_over_time(up{chain_address="0x0954efEFeb970D317a51736201B4EB2De75FF5DE", job="keep-discovered-nodes"} [364679s:120s] offset 74124s) * 120 / 364679
```

In the above query, `364679s:120s` says "retrieve 364679s worth of data, and grab a data point every 120s". 

Since these data points are either 1s or 0s, we can sum up how many data points we found, multiply that by 120 (the number of seconds in the interval), and divide by the total seconds in the range to get the total uptime percentage.

Everything appears to be doing it's job :100:

---

In a follow up PR, I would investigate whether we need all of the data we're fetching from query 2 (`paramsOperatorUptime`) and query 5 (`buildVersionInstancesParams`). I think for `paramsOperatorUptime`, we're only using that to figure out what the first registered uptime is, and there has to be a better way than making prometheus ship us all of the data points. The version instance params I think only need the *latest* timestamp.

beta-operator-rewards/rewards.ts
```diff
@@ -0,0 +1,725 @@
...
+    let { uptimeCoefficient, isUptimeSatisfied } = await checkUptime(
+      operatorAddress,
+      rewardsInterval,
+      instancesData
+    );
+    // BigNumbers cannot operate on floats. Coefficient needs to be multiplied by 100
+    uptimeCoefficient = Math.floor(uptimeCoefficient * HUNDRED);
+    requirements.set(IS_UP_TIME_SATISFIED, isUptimeSatisfied);
+
+    /// Pre-params requiremnt
```
beaushinkle, line 294:
```suggestion
    /// Pre-params requirement
```

beta-operator-rewards/rewards.ts
```diff
@@ -0,0 +1,725 @@
...
+    ((endRewardsBlock - tmpBlock) / deltaRewardsBlock) * PRECISION
+  );
+  authorization = authorization.add(
+    intervalEvents[intervalEvents.length - 1].args.toAmount.mul(coefficient)
+  );
+
+  return authorization.div(PRECISION);
+}
+
+// Get the authorization from the first event that occured after the rewards
```
beaushinkle, line 512:
```suggestion
// Get the authorization from the first event that occurred after the rewards
```

beaushinkle:
Okay! left some misc review but it's all non-blocking. The code works as-is and is ready for use, so I'm going to ship it! Totally down to handle other concerns in follow-up PRs

beaushinkle (approved):

Issue #3340, opened 2022-10-03
Excluded members who failed finalization from the signing group


pdyraga:
Originally noticed on testnet by @michalinacienciala and investigated by @lukasz-zimnoch:

The problematic case is when members fail at the last phase of DKG (finalization round) which is not followed by any communication round allowing for detection of such failures. In that scenario, members who failed the last round do not produce the key material but, at the same time, are not excluded by other participants as misbehaved, so they can be selected for signing normally. As a result, signature attempts including those members fail and the wallet needs more attempts in total to exclude those members using the random algorithm and come up with a successful signature. I opened a PR that should soften that problem and make it less frequent https://github.com/keep-network/keep-core/pull/3334. Thanks to this change, we now guarantee that the total number of signing-ready members is at least equal to the `GroupQuorum` which should strongly increase the success chance of particular signing attempts and result in a lower number of total trials (compared to at least `HonestThreshold` signing-ready members). Note that the problem may still occur if the number of successful participants is greater than the `GroupQuorum` yet less than the total number of members participating in the last DKG attempt. The ultimate solution should probably involve an additional communication round after the ECDSA DKG finalization round.

PR #3347, opened 2022-10-05
ECDSA DKG: Add confirmation state at the end of the protocol

lukasz-zimnoch:
Closes: #3340

So far, members' failures in the last state (finalization) of the ECDSA DKG could remain undetected by other participants as no explicit confirmation step was done after the result was produced. The result signature process that occurs after the DKG finalization is able to detect such failures to some extent, only when the signatures count is lower than the group quorum parameter. Undetected DKG finalization failures lead to a spurious view regarding the wallet's state because some members considered as properly operating are actually inactive. Such inactive members are normally selected for the signing work despite being unable to participate in the protocol. That leads to multiple signing failures that must be compensated by the random retry algorithm. This extends the signature time abnormally.

To mitigate that problem, we are extending the DKG finalization state to produce and exchange messages informing about successful finalization. We are also introducing a new confirmation state that occurs after the finalization and makes sure that all DKG participants selected for the current DKG attempt sent a finalization message and thus were able to produce the result.


pkg/tecdsa/dkg/protocol_test.go
```diff
@@ -1134,7 +1161,7 @@ func TestTssFinalize_ResultTimeout(t *testing.T) {
 		cancelCtx()
 		member.tssResultChan = make(<-chan keygen.LocalPartySaveData)
 
-		err := member.tssFinalize(ctx, receivedMessages)
+		_, err := member.tssFinalize(ctx, receivedMessages)
 
 		expectedErr := fmt.Errorf(
 			"TSS result was not generated on time",
```
pdyraga, line 1167: Should we add a test for the confirmation state ensuring that it fails when not all finalization messages were received?

lukasz-zimnoch: This is a basic inactivity check that we don't usually cover with protocol unit tests.

pdyraga (approved):

PR #3352, opened 2022-10-10
Handle inactive members in the signing retry loop

lukasz-zimnoch:
Here we extend the signing retry loop and make it able to explicitly filter out inactive members in the initial attempts, just like in the DKG retry loop.

The signing retry loop tries to exclude specific inactive members only if the following conditions are met:
- the attempt number is lesser than or equal to `5`
- there are known inactive members (determined using the `signing.InactiveMembersError` error)
- the random retry algorithm was not used so far
- the honest threshold for the given attempt can still be maintained 

Otherwise, the loop falls back to the random retry algorithm.

Such a logic of the signing retry loop should increase the success chance in the initial signing attempts as many failures are caused by inactive members that can be easily identified and excluded from the signing process.

### DIFF
--- a/pkg/tbtc/signing.go
+++ b/pkg/tbtc/signing.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"math/big"
 	"math/rand"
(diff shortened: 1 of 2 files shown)


lukasz-zimnoch:
During testing, we discovered that the changeset presented here cannot work as expected. Unlike the DKG retry loop, the signing retry loop works with a subset of the original signing group from the very first attempt. This means that members excluded for the first attempt follow the random algorithm for the second attempt by default, as they don't know anything about the course of the first attempt. At the same time, members selected to participate in the first attempt may encounter some inactive members and try to exclude them explicitly for the second attempt. Two different ways of computing members subset for the second attempt result in a near-fifty-fifty brain-split of the wallet members and cause signing failures in the consecutive attempts. We need to develop another approach to handle that problem effectively.

PR #3359, opened 2022-10-13
Prevent pre-params generation during DKG

tomaszslabon:
This PR adds a fix for the problem of parameter generating goroutines not being stopped on procedure start.
The problem occurs when the parameter pool is full - the parameter generation goroutine generates one more parameter and it hangs on writing it to the pool. Stopping parameter generation does not stop the goroutine  since it hangs, leaving a memory leak.
Each round of parameter generation (resumption <=> stopping) generate such a hanging goroutine, and they are only stopped by TBTC DKG removing some parameters from the pool.

The simplest solution was to check if the pool is full and if it is so sleep instead of parameter generation.

The solution was tested locally.

### DIFF
--- a/pkg/generator/pool.go
+++ b/pkg/generator/pool.go
@@ -74,30 +74,32 @@ func NewParameterPool[T any](
 	logger.Infof("loaded [%d] parameters from persistence", len(pool))
 
 	scheduler.compute(func(ctx context.Context) {
-		start := time.Now()
+		if len(pool) < poolSize {
+			start := time.Now()
 
-		generated := generateFn(ctx)
+			generated := generateFn(ctx)
 
-		// The generateFn returns nil when the context is done. We should not
-		// add nil element to the pool.
-		if generated == nil {
-			return
-		}
+			// The generateFn returns nil when the context is done. We should not
+			// add nil element to the pool.
+			if generated == nil {
+				return
+			}
 
-		persisted, err := persistence.Save(generated)
-		if err != nil {
-			logger.Errorf(
-				"failed to persist generated parameter: [%w]",
-				err,
+			persisted, err := persistence.Save(generated)
+			if err != nil {
+				logger.Errorf(
+					"failed to persist generated parameter: [%w]",
+					err,
+				)
+			}
+			pool <- persisted
+
+			logger.Infof(
+				"generated new parameters, took: [%s] current pool size: [%d]",
+				time.Since(start),
+				len(pool),
 			)
 		}
-		pool <- persisted
-
-		logger.Infof(
-			"generated new parameters, took: [%s] current pool size: [%d]",
-			time.Since(start),
-			len(pool),
-		)
 
 		// Wait some time after delivering the result regardless if the delivery
 		// took some time or not. We want to ensure all other processes of the
--- a/pkg/generator/pool.go
+++ b/pkg/generator/pool.go
@@ -74,8 +74,8 @@ func NewParameterPool[T any](
 	logger.Infof("loaded [%d] parameters from persistence", len(pool))
 
 	scheduler.compute(func(ctx context.Context) {
-		// Only generate new parameters if the pool is not full. Otherwise
-		// go to sleep.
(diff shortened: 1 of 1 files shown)


pkg/generator/pool.go
```diff
@@ -74,34 +74,44 @@ func NewParameterPool[T any](
...
+			}
+			pool <- persisted
+
+			logger.Infof(
+				"generated new parameters, took: [%s] current pool size: [%d]",
+				time.Since(start),
+				len(pool),
 			)
 		}
-		pool <- persisted
```
pdyraga, line 94: There is something I do not understand about the fix.

This is how I believe it works on `main`:

Let's say that the `poolSize` is 1000 and there are already 1000 parameters in the pool. The pool code will generate 1001th parameter but it will block on `pool <- persisted`. The DKG starts, and one or more parameters are retrieved from the pool. `pool <- persisted` is unblocked, `"generated new parameters, took: [%s] current pool size: [%d]"` is logged. Given that the context is canceled, the generator will not call this function again, and no more parameters will be generated during the DKG:

https://github.com/keep-network/keep-core/blob/e6f2a9d78a79e13c012fe02c55407e1887515a32/pkg/generator/scheduler.go#L114-L128

The PR description says:

> This PR adds a fix for the problem of parameters being generated after stopping generation.

Which I do not think is what is really happening. We are not generating any more parameters, we are just writing one already generated to the pool, and we stop. 

> Stopping parameter generation does not stop the goroutine since it hangs, leaving a memory leak.

It is blocked on writing to the buffered channel that is full but I don't think it creates a memory leak. That goroutine will complete and exit properly as soon as there is a space in the buffer (when the DKG starts).

If that's the case - and is entirely possible I missed something - why we just swap these two instructions?
https://github.com/keep-network/keep-core/blob/4c933d3bc4c2b4e52246221c9805906def7037a0/pkg/generator/pool.go#L97-L103

I think it's a bit cleaner to block on the channel write when the pool is full rather than calling the function in a loop and sleeping. Even if it means that we generated one more parameter than needed.

tomaszslabon: @pdyraga 
The problematic scenario:
1) The pool is full (1000 parameters) and we generate one additional parameter
2) Some procedure (e.g. beacon DKG) happens - since the param generation goroutine hangs on writing to the pool it's not stopped (we have 1 goroutine hanging)
3) We resume the param generating creating another goroutine - it generates one parametr (1001st) and hangs (we have 2 goroutines hanging)
4) Some procedure (e.g. relay entry) happens - since the param generation goroutines hang on writing to the pool they are not stopped
5) We resume the param generating creating another goroutine - it generates one parametr (1001st) and hangs (we have 3 goroutines hanging)
6) Finally tbt DKG happens, takes some parameters from the pool allowing the 3 hanging goroutines to stop.

At some point we had 3 hanging parameter generating goroutines. If the tbtc DKG happens rarely compared to other procedures - the number of hanging goroutines will grow and keep-client's memory consumption will keep growing with time.

tomaszslabon:
```
Which I do not think is what is really happening. We are not generating any more parameters, we are just writing one already generated to the pool, and we stop.
```
Yes, strictly speaking you are right. The additional parameter is generated before stopping param generation.

pdyraga: I have been experimenting with the code and figured out we can use `select` instead of the wait in the pool, please see https://github.com/keep-network/keep-core/pull/3363. I think I am fine with both approaches but the `select` feels cleaner to me. In the approach proposed here, we have a loop that keeps sleeping, when the pool is full and the operator is never selected to ECDSA key generation. With the approach from #3363 we use native Go `select`. Please give it some thought and let me know what you think.

tomaszslabon: I agree, `select` seems to be more Go-like. Closing this PR.

tomaszslabon:
Closing in favour of https://github.com/keep-network/keep-core/pull/3363.

PR #3360, opened 2022-10-14
Vet and lint the Go code in the CI

michalinacienciala:
We've been already performing some static analysis of the Go code in the CI, by running `gofmt` tool (which checks if the code complies to the formatting standards). But this is not the only tool that can be used to improve the code quality - we can also use `go vet` (which highlights the places in code which may work different than intended) and we can lint code using `Staticcheck` tool (which checks if the code complies to the coding conventions). In this commit we add two jobs to the `client.yml` workflow, which will be used to run those tools. The jobs wil be executed in PRs that may affect the client code and also after those PRs get merged to main.

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -229,3 +229,29 @@ jobs:
             gofmt -d -e .
             exit 1
           fi
+
+  client-vet:
+    needs: client-detect-changes
+    if: |
+      github.event_name == 'push'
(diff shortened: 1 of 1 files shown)


michalinacienciala:
@nkuba, ready for review. I've looked at your commit and it looks good, you can take a look at the rest.

nkuba (approved):

PR #3361, opened 2022-10-14
tECDSA signing: Announcement phase

lukasz-zimnoch:
Here introduce a new message type `tbtc.signingAnnouncementMessage` and leverage it to implement a readiness announcement phase before each signing attempt of the given item. This phase is meant to determine a set of active
members from which the random algorithm can safely select a 51-members subset supposed to perform the signing attempt with a high success chance.

This change should improve the overall signing success rate by alleviating the negative impact of inactivites. Moreover, the mechanism presented here plays nicely with the design described in the tBTC's RFC-7 (https://github.com/keep-network/tbtc-v2/pull/414) and can be used to synchronize members that must sign the right thing proposed by the coordinator. Last but not least, thanks to announcements, operators whose clients were restarted will have a chance to catch up with the ongoing signing processes and increase their success chance.


pkg/tbtc/signing.go
```diff
@@ -233,3 +324,172 @@ func (srl *signingRetryLoop) qualifiedOperatorsSet() (
...
+	qualifiedOperatorsSet map[chain.Address]bool,
+	readyMembersIndexes []group.MemberIndex,
+) []group.MemberIndex {
+	includedMembersIndexes := make([]group.MemberIndex, 0)
+	excludedMembersIndexes := make([]group.MemberIndex, 0)
+	for i, operator := range srl.signingGroupOperators {
+		memberIndex := group.MemberIndex(i + 1)
+
+		if qualifiedOperatorsSet[operator] &&
+			slices.Contains(readyMembersIndexes, memberIndex) {
```
pdyraga, line 340: I am confused. In `qualifiedOperatorsSet` we pass only `readySigningGroupOperators` to `retry.EvaluateRetryParticipantsForSigning`. From what I understand, it should not return anyone who is not on that list. Why do we check here again then?

pkg/tbtc/signing_test.go
```diff
@@ -202,6 +342,252 @@ func TestSigningRetryLoop(t *testing.T) {
...
+					test.expectedResults,
+					results,
+				)
+			}
 		})
 	}
 }
+
+type mockSigningAnnouncer struct {
+	outgoingAnnouncements map[uint64]struct {
```
pdyraga, line 571: Maybe `performedAnnouncements` / `announcementsExecuted` / `announcementParams`?

pkg/tbtc/signing.go
```diff
@@ -213,15 +290,29 @@ func (srl *signingRetryLoop) start(
...
+// in the given signing attempt. The set of qualified operators is taken
+// from the set of active operators who announced readiness through
+// their controlled signing group members.
+func (srl *signingRetryLoop) qualifiedOperatorsSet(
+	readyMembersIndexes []group.MemberIndex,
+) (map[chain.Address]bool, error) {
+	// The retry algorithm expects that we count retries from 0. Since
+	// the first invocation of the algorithm will be for `attemptCounter == 1`
+	// we need to subtract one while determining the number of the given retry.
+	retryCount := srl.attemptCounter - 1
```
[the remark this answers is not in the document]
lukasz-zimnoch: Nope, DKG has a different logic where the counters are separated. Here it makes sense because if we are on attempt `1` we did `0` retries. If we are on attempt `2` we did `1` retry and so on.

pkg/tbtc/signing_test.go
```diff
@@ -202,6 +342,252 @@ func TestSigningRetryLoop(t *testing.T) {
...
+						t.Errorf(
+							"unexpected outgoing announcement's message "+
+								"for attempt [%v]\n"+
+								"expected: [%+v]\n"+
+								"actual:   [%+v]",
+							attemptNumber,
+							message,
+							outgoingAnnouncement.message,
+						)
+					}
```
pdyraga, line 386: Could we use `testutils.AsertIntsEqual`?

pkg/tbtc/signing_test.go
```diff
@@ -202,6 +342,252 @@ func TestSigningRetryLoop(t *testing.T) {
...
+				results[r.memberIndex] = r.readyMembersIndexes
+			}
+
+			close(errorsChan)
+			errors := make(map[group.MemberIndex]error)
+			for e := range errorsChan {
+				errors[e.memberIndex] = e.err
+			}
+
+			if !reflect.DeepEqual(test.expectedErrors, errors) {
```
pdyraga, line 547: We could also define the error as a `var` and use `testutils.AssertErrorsSame` here. I think this approach is cleaner and the linter will not complain.

pkg/tbtc/signing.go
```diff
@@ -233,3 +324,172 @@ func (srl *signingRetryLoop) qualifiedOperatorsSet() (
...
+			return excludedMembersIndexes[i] < excludedMembersIndexes[j]
+		})
+	}
+
+	return excludedMembersIndexes
+}
+
+// broadcastSigningAnnouncer is an implementation of the signingAnnouncer that
+// performs the readiness announcement over the provided broadcast channel.
+type broadcastSigningAnnouncer struct {
```
pdyraga, line 388: Worth considering extracting it to a separate file. Not necessarily in this PR but once we add the logic for operators joining the retries after the client restart.

pkg/tbtc/signing.go
lukasz-zimnoch: Yeah, this is confusing but that's the only option to make it work as expected for cases where some members of the same operator are ready and others are not. 

Suppose we have operator `O` controlling members `1`, `3`, and `5` where `1` did not announce their readiness. The `retry.EvaluateRetryParticipantsForSigning` does not care about the exact mapping between the operator and the exact members it controls. It only operates on the seat count which is `2` for operator `O` as we filtered out ready members using the `readyMembersIndexes` argument at the beginning of `qualifiedOperatorsSet`. Once we return from `qualifiedOperatorsSet` we only know a set of qualified operators (selected by the random algorithm) whose total seat count is enough to maintain the threshold. To properly take ready members (`3` and `5`) for operator `O` in the part you mentioned, we need to check against `readyMembersIndexes` again to properly take the two ready members controlled by operator `O` which are `3` and `5`. Otherwise, we would end up with `1` and `3` which is wrong as `1` did not announce their readiness.

pdyraga:
What kind of scenarios were tested and what should be tested before the merge?

The following scenarios come to my mind:
- All members announce their readiness. Success on the first attempt is expected.
- Not all but more than t+1 members announce their readiness. Success on the first attempt is expected.
- Less than t+1 members announce readiness on the first attempt. Then all of them announce readiness on the second attempt. Success on the second attempt is expected.
- All members announce their readiness. One of the members stops cooperating during the signing and goes offline. In the second attempt that member does not announce readiness but we still have more than t+1 active members. Success on the second attempt is expected.

pkg/tbtc/signing.go
lukasz-zimnoch: We can think about it but honestly, I think it's a good place for this type. The `signing.go` file contains a set of signing-related components that are closely related to each other. This is a quite Go-like way to organize things. We probably don't want a Java-like style where each significant type has its own file.

pkg/tbtc/signing_test.go
lukasz-zimnoch: What's wrong with `outgoingAnnouncements`? Those are announcements that left the announcer which is similar to the real implementation of the announcer.

pkg/tbtc/signing_test.go
```diff
@@ -202,6 +342,252 @@ func TestSigningRetryLoop(t *testing.T) {
...
 	}
 }
+
+type mockSigningAnnouncer struct {
+	outgoingAnnouncements map[uint64]struct {
+		signingGroupMemberIndex group.MemberIndex
+		message                 *big.Int
+	}
+
+	incomingAnnouncementsFn func(
```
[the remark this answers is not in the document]
lukasz-zimnoch: Similarly to https://github.com/keep-network/keep-core/pull/3361#discussion_r998102312, I think `announceFn` will be confusing as we invoke that in `announce` function anyway. The `incomingAnnouncementsFn` name translates to "function that returns incoming announcements" which clearly indicates its purpose.

lukasz-zimnoch:
> What kind of scenarios were tested and what should be tested before the merge?
> 
> The following scenarios come to my mind:
> 
>     * All members announce their readiness. Success on the first attempt is expected.
> 
>     * Not all but more than t+1 members announce their readiness. Success on the first attempt is expected.
> 
>     * Less than t+1 members announce readiness on the first attempt. Then all of them announce readiness on the second attempt. Success on the second attempt is expected.
> 
>     * All members announce their readiness. One of the members stops cooperating during the signing and goes offline. In the second attempt that member does not announce readiness but we still have more than t+1 active members. Success on the second attempt is expected.

So far, I tested this PR in the following way:

**Environment**: 3 local clients each controlling ~1/3 of the group signers so always 2 clients are needed to sign.

**Scenario 1**: All clients up. All members announced their readiness. Successful signing on the first attempt.

**Scenario 2**: Two clients up. All members but the ones controlled by the offline client announced their readiness. Successful signing on the first attempt.

**Scenario 3**: All clients up. All members announced their readiness on the first attempt. During the first attempt, one of the clients goes offline so the attempt fails. On the second attempt, all members but the ones controlled by the offline client announced their readiness. Successful signing on the second attempt.

pkg/tbtc/signing.go
```diff
@@ -233,3 +324,172 @@ func (srl *signingRetryLoop) qualifiedOperatorsSet() (
...
+// announce is an implementation of the signingAnnouncer.announce method.
+//
+// This implementation fulfills the method specification and additionally:
+// - blocks until the ctx passed as argument is done
+// - returns an error if the total number of ready members is lesser than the
+//   honest threshold parameter
+func (bsa *broadcastSigningAnnouncer) announce(
+	ctx context.Context,
+	signingGroupMemberIndex group.MemberIndex,
+	message *big.Int,
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 9cbe7401

pkg/tbtc/signing.go
pdyraga: Instead of calling `srl.qualifiedOperatorsSet(readyMembersIndexes)` and `srl.excludedMembersIndexes` from the `start` function, can we introduce another function that would wrap `qualifiedOperatorsSet` and `excludedMembersIndexes` and would have the explanation from your comment?

Also, I wonder if we should check `slices.Contains(readyMembersIndexes, memberIndex)` as a part of this loop, or maybe have a separate one, after `qualifiedOperatorsSet` that would do it. It's more work but I think it would be easier to understand. 

We could also add another function to `retry` that would call `EvaluateRetryParticipantsForSigning` internally and did care about the seats.

pkg/tbtc/signing_test.go
pdyraga: I see now what you meant by `outgoingAnnouncements` and `incomingAnnouncementsFn`. I thought the first one is just capturing the parameters for the `announce` function calls so that we can later assert if this function was called with the right parameters. And for the second, I assumed this is just what we want to return, not what is incoming from the outside to the given announcer instance.

Can we add some code comments explaining these two fields?

pkg/tbtc/signing_test.go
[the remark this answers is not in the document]
pdyraga: Resolving with [this](https://github.com/keep-network/keep-core/pull/3361#discussion_r998136258) comment.

pkg/tbtc/signing.go
lukasz-zimnoch: > Instead of calling srl.qualifiedOperatorsSet(readyMembersIndexes) and srl.excludedMembersIndexes from the start function, can we introduce another function that would wrap qualifiedOperatorsSet and excludedMembersIndexes and would have the explanation from your comment?

See 26b343a7.

> Also, I wonder if we should check slices.Contains(readyMembersIndexes, memberIndex) as a part of this loop, or maybe have a separate one, after qualifiedOperatorsSet that would do it. It's more work but I think it would be easier to understand.

I think that should be clear now given the change from 26b343a7

> We could also add another function to retry that would call EvaluateRetryParticipantsForSigning internally and do care about the seats.

I'd rather avoid changing anything in the `retry` package unless absolutely needed. The current interface is generic enough and it does what it is supposed to do. Also, we have consistent functions for both signing and DKG. I think all the nuances should be handled in the client packages.

pkg/tbtc/signing.go
```diff
@@ -107,104 +171,117 @@ func (srl *signingRetryLoop) start(
...
+					"[member:%v] failed waiting for announcement end "+
+						"block [%v] for attempt [%v]: [%v]",
+					srl.signingGroupMemberIndex,
+					announcementEndBlock,
+					srl.attemptCounter,
+					err,
 				)
 			}
-		}
+		}()
```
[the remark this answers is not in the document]
pdyraga: In this model, we do not have to do `ctx.Err() != nil` before the announcement, right? If so, I think it works for me.

pkg/tbtc/signing_test.go
lukasz-zimnoch: No longer relevant due to ab218401

pkg/tbtc/signing_test.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3361/commits/771b8707781a69eccf1dd5adaba3a712f75c3b78

lukasz-zimnoch:
All comments addressed @pdyraga

pkg/tbtc/signing_test.go
```diff
@@ -202,6 +374,230 @@ func TestSigningRetryLoop(t *testing.T) {
...
+					results,
+				)
+			}
+		})
+	}
+}
+
+type mockSigningAnnouncer struct {
+	// outgoingAnnouncements holds all announcements that are supposed to
+	// be sent by the announcer.
```
pdyraga, line 578:
```suggestion
	// outgoingAnnouncements holds all announcements that are 
	// sent by the announcer.
```

I am not sure if this is "supposed to" because it's set in `announce` when the announce happens. So I think this is "what is sent by the announcer" and we later assert it against what it was supposed to send.

pkg/tbtc/signing_test.go
```diff
@@ -202,6 +374,230 @@ func TestSigningRetryLoop(t *testing.T) {
...
+	outgoingAnnouncements map[uint64]struct {
+		signingGroupMemberIndex group.MemberIndex
+		message                 *big.Int
+	}
+
+	// incomingAnnouncementsFn returns all announcements that are supposed to
+	// be received by the announcer for the given attempt.
```
pdyraga, line 585:
```suggestion
	// incomingAnnouncementsFn returns all announcements that are
	// received by the announcer for the given attempt.
```

Same reasoning as for `outgoingAnnouncements`.

pdyraga:
Testing it now.

pkg/tbtc/signing_test.go
lukasz-zimnoch: https://github.com/keep-network/keep-core/pull/3361/commits/9d61a51cd434b23fdb57c54df08f068592fdec7a

pkg/tbtc/signing_test.go
lukasz-zimnoch: https://github.com/keep-network/keep-core/pull/3361/commits/9d61a51cd434b23fdb57c54df08f068592fdec7a

pdyraga (approved):
Tested locally with four clients the following scenarios:
1. All four clients responded to the announcement, signing completed.
2. One client did not respond to the announcement and got excluded from signing. Signing completed in the first try.
3. One client stopped responding in phase 2 of signing, signing failed, the retry was scheduled, that client did not respond to the announcement of the second attempt and got excluded from signing. Signing completed in the second try.

PR #3362, opened 2022-10-17
Fast state machine

pdyraga:
Package `faststate` contains a generic state machine implementation that is meant to be used with interactive protocols which do not require a strict synchronization mechanism between protocol members. The synchronization is based on an signal from each participant that they are ready to proceed to the next step when, for example, they received all the necessary information from all the other participants.

This approach allows for faster execution of protocols but has strict requirements regarding the implementation of states.

# Requirement 1: Context lifetime and retransmissions

The context passed to `faststate.NewMachine` must be active as long as the result is not published to the chain or until a fixed time for the protocol execution has not passed.

The context is used for the retransmission of messages and all protocol participants must have a chance to receive messages from other participants. Consider the following example: there are two participants of the protocol: A and B, and they are executing the final state of the protocol. If the context was canceled right after the completion of work by the participant, without a confirmation that the result was published on-chain, we could run into a situation when A received a message from B and exited immediately, without giving B a chance to receive a message:
```
A: |-S------R-|
B:        |-S----------(...)
```
`|-` denotes when the last state starts and `-|` denotes when it ends. `S` denotes when the given participant sent its message and `R` denotes when the given participant received the message from the other. Since B initiated the last state later than A (it could execute some time-consuming computations in the last state), A already sent its message and now B needs to wait for the retransmissions. B sends its message immediately upon the initiation and this message is received by A. Since A exits and cancels the context immediately after receiving a message from B, it is no longer retransmitting its message and B hangs forever.

There are two solutions. One is that both A and B observe the chain and they keep the context active until the result is published. Another is that each member waits for some time before completing the protocol and canceling the context for retransmissions or that the entire protocol has a fixed maximum execution time until the timeout.

# Requirement 2: Store all received messages

Since the state machine does not require strict synchronization between participants, it is not guaranteed at which moment of the execution the rest of the group is. If the current member is at the first state of the execution, and all other members advanced to further states, the current member should accept and store messages from further states "for the future" instead of rejecting them. This can be achieved using `faststate.BaseState` structure.

### DIFF
--- a/pkg/protocol/faststate/faststate.go
+++ b/pkg/protocol/faststate/faststate.go
@@ -110,17 +110,12 @@ func (bs *BaseState) receive(msg net.Message) {
 	defer bs.messagesMutex.Unlock()
 
 	messageType := msg.Type()
-	received, ok := bs.messages[messageType]
-
-	if ok {
-		bs.messages[messageType] = append(received, msg)
-	} else {
-		bs.messages[messageType] = []net.Message{msg}
-	}
+	bs.messages[messageType] = append(bs.messages[messageType], msg)
 }
 
 // GetAllReceivedMessages returns all messages of the given Type() received
-// so far.
+// so far. If no messages of the given Type() were received, nil slice is
+// returned.
 func (bs *BaseState) GetAllReceivedMessages(messageType string) []net.Message {
 	bs.messagesMutex.RLock()
 	defer bs.messagesMutex.RUnlock()
--- a/pkg/protocol/faststate/faststate.go
+++ b/pkg/protocol/faststate/faststate.go
@@ -104,8 +104,14 @@ func NewBaseState() *BaseState {
 	}
 }
 
-// receive stores the received message by its Type()
-func (bs *BaseState) receive(msg net.Message) {
+// ReceiveToHistory stores the received message by its Type() in the internal
+// history of received messages.
+//
+// This function is NOT performing any validation of the received net.Message,
+// especially if the public key matches member index or if the given operator is
+// allowed to publish messages in the broadcast channel. This function should
+// be wrapped with a set of validations.
+func (bs *BaseState) ReceiveToHistory(msg net.Message) {
 	bs.messagesMutex.Lock()
 	defer bs.messagesMutex.Unlock()
 
--- a/pkg/protocol/faststate/faststate.go
+++ b/pkg/protocol/faststate/faststate.go
@@ -1,0 +1,129 @@
+// Package faststate contains a generic state machine implementation that is
+// meant to be used with interactive protocols which do not require a strict
+// synchronization mechanism between protocol members. The synchronization is
+// based on an signal from each participant that they are ready to proceed to
+// the next step when, for example, they received all the necessary information
+// from all the other participants.
+//
+// This approach allows for faster execution of protocols but has strict
+// requirements regarding the implementation of states.
+//
+// Requirement 1: Context lifetime and retransmissions
+//
+// The context passed to `faststate.NewMachine` must be active as long as the
+// result is not published to the chain or until a fixed time for the protocol
+// execution has not passed.
+//
+// The context is used for the retransmission of messages and all protocol
+// participants must have a chance to receive messages from other participants.
+//
+// Consider the following example: there are two participants of the protocol:
+// A and B, and they are executing the final state of the protocol. If the
+// context was canceled right after the completion of work by the participant,
+// without a confirmation that the result was published on-chain, we could run
+// into a situation when A received a message from B and exited immediately,
+// without giving B a chance to receive a message:
+//
+// A: |-S------R-|
+// B:        |-S----------(...)
+//
+// |- denotes when the last state starts and -| denotes when it ends. S denotes
+// when the given participant sent its message and R denotes when the given
+// participant received the message from the other. Since B initiated the last
+// state later than A (it could execute some time-consuming computations in the
+// last state), A already sent its message and now B needs to wait for the
+// retransmissions. B sends its message immediately upon the initiation and this
+// message is received by A. Since A exits and cancels the context immediately
+// after receiving a message from B, it is no longer retransmitting its message
+// and B hangs forever.
+//
+// There are two solutions. One is that both A and B observe the chain and they
+// keep the context active until the result is published. Another is that each
+// member waits for some time before completing the protocol and canceling the
+// context for retransmissions or that the entire protocol has a fixed maximum
+// execution time until the timeout.
+//
+// Requirement 2: Store all received messages
+//
+// Since the state machine does not require strict synchronization between
+// participants, it is not guaranteed at which moment of the execution the rest
+// of the group is. If the current member is at the first state of the
+// execution, and all other members advanced to further states, the current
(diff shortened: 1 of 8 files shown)


pkg/protocol/faststate/machine_test.go
```diff
@@ -0,0 +1,232 @@
...
+	"math/rand"
+	"strconv"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/keep-network/keep-core/pkg/internal/testutils"
+	"github.com/keep-network/keep-core/pkg/net"
+	netLocal "github.com/keep-network/keep-core/pkg/net/local"
```
lukasz-zimnoch, line 15: According to Go's convention about package names/aliases (https://go.dev/blog/package-names) we should use lowercase here:
```suggestion
	netlocal "github.com/keep-network/keep-core/pkg/net/local"
```

pkg/protocol/faststate/machine_test.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	channel.SetUnmarshaler(func() net.TaggedUnmarshaler {
+		return &round2Message{}
+	})
+	channel.SetUnmarshaler(func() net.TaggedUnmarshaler {
+		return &round3Message{}
+	})
+
+	ctx, cancelFn := context.WithCancel(context.Background())
```
lukasz-zimnoch, line 39: Can we rename it to:
```suggestion
	ctx, cancelCtx := context.WithCancel(context.Background())
```
? `cancelFn` does not reflect that this function cancels a context.

pkg/protocol/faststate/faststate.go
```diff
@@ -0,0 +1,129 @@
+// Package faststate contains a generic state machine implementation that is
```
lukasz-zimnoch, line 1: I'm wondering about alternative ways of structuring the code of both state machines. Packages `state` and `faststate` can be really mysterious for someone not familiar with this code and the differences are not obvious at the first sight.

What comes into my mind is the fact that we can call the `state.Machine` a "synchronous" machine as states are synchronized using the block counter. On the other hand, the new `faststate.Machine` is "asynchronous" as states decide about the transition using their own internal predicates. That said we can refactor the current structure twofold:

**Option 1**: We can rename the package `state` to `syncstate` and `faststate` to `asyncstate`. This way we will end up with the following types:
- `syncstate.State`
- `syncstate.Machine`
- `asyncstate.State`
- `asyncstate.Machine`

**Option 2**: We can move everything into a single package `state` and rename specific types. This way we would have:
- `state.SyncState`
- `state.SyncMachine`
- `state.AsyncState`
- `state.AsyncMachine`

Although **O1** can be tempting as we would have two separate and small packages with well-defined types, I think we should lean towards **O2**. The main reasons are:
- All this code is about state and both machines (and associated types) have a lot of common things so it makes sense to keep them in a single `state` package. We can probably extract some common parts to reduce duplication (both `State` interfaces share a lot of methods).
- This option is closer to how we typically group things into packages. We deal with some different flavors of the same component. For example, we have the `pkg/tecdsa/retry` package that exposes two retry algorithms, for two separate use cases.
- The size of the resulting package will still be on a sane level

pkg/protocol/faststate/faststate.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if ok {
+		bs.messages[messageType] = append(received, msg)
+	} else {
+		bs.messages[messageType] = []net.Message{msg}
+	}
+}
+
+// GetAllReceivedMessages returns all messages of the given Type() received
+// so far.
```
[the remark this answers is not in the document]
pdyraga: I was also thinking about returning `([]net.Message, bool)` from this function as:
```
received, ok := bs.messages[messageType]
return received, ok
```

This feels even better but is quite unwieldy when used in `CanTransition()`. Right now we can do just:
```
func (ts *testState2) CanTransition() bool {
	return len(ts.GetAllReceivedMessages(round2MessageType)) == partyCount
}
```

With the `([]net.Message, bool)` option we would have to:
```
func (ts *testState2) CanTransition() bool {
	if received, ok := ts.GetAllReceivedMessages(round2MessageType); !ok || len(received) != partyCount {
		return false
	}

	return true
}
```

I would just flag that `nil` is returned but I would probably not return an empty slice given that Go is fine with `len` of a `nil` slice.

pkg/protocol/faststate/faststate.go
```diff
@@ -0,0 +1,129 @@
...
+	defer bs.messagesMutex.Unlock()
+
+	messageType := msg.Type()
+	received, ok := bs.messages[messageType]
+
+	if ok {
+		bs.messages[messageType] = append(received, msg)
+	} else {
+		bs.messages[messageType] = []net.Message{msg}
+	}
```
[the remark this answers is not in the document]
pdyraga: 9d61b8337b689803d451acb6f85be619754d1c53

pkg/protocol/faststate/faststate.go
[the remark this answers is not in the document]
pdyraga: 9d61b8337b689803d451acb6f85be619754d1c53

pkg/protocol/faststate/machine.go
```diff
@@ -0,0 +1,168 @@
...
+			onStateDone, err = stateTransition(
+				m.ctx,
+				m.logger,
+				currentState,
+			)
+			if err != nil {
+				return nil, err
+			}
+
+			continue
```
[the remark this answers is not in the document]
pdyraga: It is. d4fde9c8d9f48ac66766101330b6e00cb587ac50

pkg/protocol/faststate/machine_test.go
pdyraga: 14777d2f580d8259c457317cb7e4bd1c2b19e363

pkg/protocol/faststate/faststate.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func NewBaseState() *BaseState {
+	return &BaseState{
+		messages: make(map[string][]net.Message),
+	}
+}
+
+// receive stores the received message by its Type()
+func (bs *BaseState) receive(msg net.Message) {
```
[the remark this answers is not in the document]
pdyraga: I renamed it to `ReceiveToHistory`. I think we do not want to name it `Receive` - we want to force the state implementations to implement `Receive` and execute the set of validations for `net.Message`. That was the reasoning behind the lowercase version but... of course, I missed that it has to be exported. 😉 

See cb6a6c1aed17f8b53ba439748bdc2dc5ff3abc0c please.

pkg/protocol/faststate/machine_test.go
pdyraga: d695c91cc64f4db2831e26e8ba8f3a87712d9dd7

pkg/protocol/faststate/faststate.go
lukasz-zimnoch: Sure, let's do it separately!

pdyraga: Done in https://github.com/keep-network/keep-core/pull/3365.

PR #3363, opened 2022-10-20
Do not wait on writing to pool if the context got canceled

pdyraga:
This is an alternative approach to the fix presented in #3359. Please see [this](https://github.com/keep-network/keep-core/pull/3359#discussion_r997928620) dicussion.

If the context for generating the pre-parameters has been canceled, we should not leave the generator goroutine dangling on writing to the channel.

Here is the problematic scenario detected by @tomaszslabon:

1. The pool is full (1000 parameters) and we generate one additional parameter
2. Some procedure (e.g DKG) happens - since the param generation goroutine hangs on writing to the pool it's not stopped (we have 1 goroutine hanging)
3. We resume the param generation creating another goroutine - it generates one parameter (1001st) and hangs (we have 2 goroutines hanging)

This scenario is happening again and again and we end up with a bunch of blocked generator goroutines.

### DIFF
--- a/pkg/generator/pool.go
+++ b/pkg/generator/pool.go
@@ -91,7 +91,17 @@ func NewParameterPool[T any](
 				err,
 			)
 		}
-		pool <- persisted
+
+		select {
+		case pool <- persisted:
+		case <-ctx.Done():
+			// Do not wait on writing to the full channel if the context for
+			// the generateFn was cancelled. This way we can avoid multiple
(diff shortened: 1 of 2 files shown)


tomaszslabon:
@pdyraga 
I tested this solution locally and it worked as expected. 
This solution seems to be be more Go-like compared to https://github.com/keep-network/keep-core/pull/3359.

tomaszslabon (approved):

PR #3374, opened 2022-10-24
Tbtc Maintainer: Added the maintainer command

tomaszslabon:
This PR adds a command for launching the maintainer for Bitcoin difficulty.
The relay maintainer can be launched using the command:
```
KEEP_ETHEREUM_PASSWORD="password" ./keep-client maintainer  --bitcoinDifficulty  --config /Users/tslabon/projekt/keep-core/configs/maintainer.toml --developer
```

In the future we can add other maintainer tasks and launch the maintainer, e.g.:
```
KEEP_ETHEREUM_PASSWORD="password" ./keep-client maintainer  --bitcoinDifficulty --spv  --config /Users/tslabon/projekt/keep-core/configs/maintainer.toml --developer
```

Maintainer uses 3 groups of configuration flags:
- `ethereum` (url, keyFile, etc.)
- `electrum` (url) 
- `maintainer` (bitcoinDifficulty, spv, etc.)

All the options can be provided on the command line or using a configuration file or a mix of them.
Here is an example configuration toml file:
```
[ethereum]
  URL = "ws://127.0.0.1:8546"
  KeyFile = "/path/to/ethereum/data/keystore/UTC--2022-10-03T10-45-23.667111000Z--7ca4c3c109851221b6560c7d4f3917db6d056977"

[bitcoin.electrum]
  URL = "https://blockstream.info/api/"

[tbtc.maintainer]
 BitcoinDifficulty = true
```

Tested locally.


lukasz-zimnoch:
I'll open a PR with the bitcoin groundwork soon. Please hold off with the merge of this PR until so.

lukasz-zimnoch:
The PR with the Bitcoin toys is ready: https://github.com/keep-network/keep-core/pull/3384

cmd/relay.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/keep-network/keep-core/pkg/bitcoin"
+	"github.com/keep-network/keep-core/pkg/maintainer"
+	"github.com/spf13/cobra"
+)
+
+var maintainerConfig = &maintainer.Config{}
+
+// RelayCommand contains the definition of the relay command-line subcommand.
+var RelayCommand = &cobra.Command{
```
lukasz-zimnoch, line 15: I think we must think about this command a little bit wider. The package `pkg/maintainer` will hold not only the relay maintainer but also the SPV maintainer code, wallet lifecycle maintainer code, and so on. That said, the `./keep-client relay` command is probably not a good fit here and we have basically two options to run that stuff:

Option 1: A bulk run of all maintainers with a single command. In this option, we would have a `./keep-client maintainer` command that will trigger all the maintainers in a single shot. This is more convenient but less flexible.

Option 2: A granular approach where each maintainer is triggered separately. In this option, we would have `./keep-client maintainer relay`, `./keep-client maintainer spv`, and so on. This is the exact opposite of the first option in terms of pros and cons.

I think we need to agree on a solution here @pdyraga @nkuba

pkg/maintainer/relay.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+
+	"github.com/keep-network/keep-common/pkg/chain/ethereum"
+	"github.com/keep-network/keep-core/pkg/bitcoin"
+	"github.com/keep-network/keep-core/pkg/tbtc"
+	"github.com/mitchellh/mapstructure"
+	"github.com/spf13/viper"
+)
+
+type Config struct {
```
lukasz-zimnoch, line 13: I'd move the config-related stuff to a separate file `pkg/maintainer/config.go` and cover the functions with unit tests.

pkg/maintainer/relay.go
```diff
@@ -0,0 +1,85 @@
...
+func NewRelay(btcChain bitcoin.Chain, tbtcChain tbtc.Chain) *Relay {
+	return &Relay{
+		btcChain:  btcChain,
+		tbtcChain: tbtcChain,
+	}
+}
+
+type Relay struct {
+	btcChain  bitcoin.Chain
+	tbtcChain tbtc.Chain
```
lukasz-zimnoch, line 84: I think this is wrong. Ideally, the `pkg/maintainer` should be independent of `pkg/tbtc`. That means we should declare our own `maintainer.RelayChain` interface with expectations regarding the chain required by the relay maintainer and implement it within `pkg/chain/ethereum`.

lukasz-zimnoch (changes requested):

pkg/maintainer/relay.go
tomaszslabon: Done in 1a02c7248bce4395be3830bc8b5bc4646f63f12a.

cmd/relay.go
nkuba: I like the idea of introducing the `maintainer` command. To decide between Option 1 and Option 2 we need to consider how much of the config and command's flags/args will be common for all the types of maintiners.

Maybe we could use something like:
`./keep-client maintainer --relay`
`./keep-client maintainer --spv`
`./keep-client maintainer --relay --spv`
`./keep-client maintainer` <- this one will run all the maintainers

pdyraga: I like @nkuba's proposition.

Also, I think we should mark `maintainer` feature as experimental in help. This will be in-development feature and we'll release new versions in the meantime.

tomaszslabon: Done. I changed the way the maintainer tasks are started.
Now it's started like that:
`KEEP_ETHEREUM_PASSWORD="password" ./keep-client maintainer  --relay  --config /path/to/config/maintainer.toml`

The `KEEP_ETHEREUM_PASSWORD` environment variable is needed because we use ethereum configuration to start the maintainer.

In the future more maintainers could be started by providing more flags:
`KEEP_ETHEREUM_PASSWORD="password" ./keep-client maintainer  --relay --spv  --config /path/to/config/maintainer.toml`

pkg/maintainer/relay.go
tomaszslabon: I moved the config to `pkg/maintainer/config.go`, but I didn't add any unit tests, since I simplified the config and now it's a simple structure without any methods.

pkg/bitcoin/chain.go
```diff
@@ -1,5 +1,12 @@
 package bitcoin
 
+// Config contains the config data needed to connect to the Bitcoin chain.
+type Config struct {
+	URL      string
+	Username string
+	Password string
+}
```
lukasz-zimnoch, line 8: This is the wrong place for that. This config is implementation-specific and should be exposed by `bitcoin.Client` implementations.

tomaszslabon: I wanted to put this `Config` struct, so that we have a clean `bitcoin.Config` field in the general client configuration that any possible command of `keep-client` could reuse:
```
// Config is the top level config structure.
 type Config struct {
 	Ethereum   commonEthereum.Config
 	Bitcoin    bitcoin.Config
 	LibP2P     libp2p.Config `mapstructure:"network"`
 	Storage    storage.Config
 	ClientInfo clientinfo.Config
 	Maintainer maintainer.Config
 	Tbtc       tbtc.Config
 }
 ```
 
 My reasoning was that `URL`, `username` and`password` fields inside bitcoin.Config` are pretty generic and will fit any bitcoin implementation.
 
Should I move this config into the electrum package, so that we have 
```Electrum   electrum.Config```
inside the client config?

Comment (lukasz-zimnoch) [reply pkg/bitcoin/chain.go:8]: The `pkg/bitcoin` should only contain types from the Bitcoin protocol specification (see package documentation). There is no such type as `Config` in the Bitcoin spec. This is a pure implementation-specific concept and it should live in the implementation package. Also, different implementations could have different configs, one of them can use a username and password but another one can use an authentication token. This is another argument that we should keep the config on the implementation level. Last but not least, note that the `Ethereum commonEthereum.Config` is also implementation-specific, we don't declare it on the `pkg/chain` level. So as `LibP2P libp2p.Config` which is libp2p-specific and is not declared on `pkg/net` level. That being said, I think we should declare `Electrum electrum.Config` just as you proposed.

File: cmd/relay.go (hunk 1 above)
Comment (lukasz-zimnoch) [reply cmd/relay.go]: Can we consider this resolved? I recall @tomaszslabon mentioned something about problems with the presented approach. What's the status here?

File: pkg/maintainer/relaychain.go [hunk 5]
```diff
@@ -0,0 +1,13 @@
+package maintainer
+
+import (
+	"github.com/keep-network/keep-core/pkg/bitcoin"
+)
+
+// RelayChain is an interface that provides the ability to communicate with the
+// relay on-chain contract.
+type RelayChain interface {
```
lukasz-zimnoch, line 9: What's the point of putting `maintainer.RelayChain` into a separate file? This won't be a huge interface so I'd keep it at the point of use (`relay.go`) to avoid jumping between files. Also, this is a Go-like way of organizing things, we don't need to create a separate file for each public type.

pkg/maintainer/maintainer.go
```diff
@@ -0,0 +1,30 @@
...
+	// btcChain, err := bitcoin.Connect(ctx, &maintainerConfig.Bitcoin)
+	// if err != nil {
+	// 	return fmt.Errorf("could not connect BTC chain: [%v]", err)
+	// }
+
+	// TODO: Add connection to the tbtc chain:
+	// relayChain, err := connectRelayChain(config)
+	// if err != nil {
+	// 	return fmt.Errorf("could not connect relay chain: [%v]", err)
+	// }
```
lukasz-zimnoch, line 18: I think we should use dependency injection here and make `Initialize` accept `bitcoin.Chain` and `RelayChain` as parameters, just like we do in `beacon.Initialize` and `tbtc.Initialize`. Ideally, package `pkg/maintainer` should depend on abstractions of both chains and should be not responsible for instantiating concrete implementations.

pkg/maintainer/relay.go
```diff
@@ -0,0 +1,54 @@
...
+	go relay.startControlLoop(ctx)
+
+	return relay
+}
+
+// Relay is the part of maintainer responsible for maintaining the state of
+// the relay on-chain contract.
+type Relay struct {
+	btcChain   bitcoin.Chain
+	relayChain RelayChain
```
lukasz-zimnoch, line 32: I think we can make it simpler:
```suggestion
	chain RelayChain
```

pkg/bitcoin/chain.go
tomaszslabon: Replaced `Bitcoin` configuration with `Electrum` in 2eb93eaefef4c21cbbcb13e47a1f1c465fe81576.
Once https://github.com/keep-network/keep-core/pull/3395 is merged to main, we will replace `Electrum` with `Electrs`. For now, let's stick to `Electrum`, as that's the name we have in the base branch of my PR.

pkg/maintainer/relaychain.go
tomaszslabon: Moved it back to `relay.go` in 7e975d796e0dd9de239fc80b0489538061d221ec.
I think the interface will have about 6 methods (there will be more methods than just `Retarget()`, e.g.  `ProofLength()`, `CurrentEpoch()`, `Ready()`, etc. ).

cmd/relay.go
tomaszslabon: @lukasz-zimnoch 
My concern was whether launching multiple maintainers with one command is really beneficial.
I left a TODO to consider whether or not we should do that.

But I don't think this issue is a blocker for this PR, as we have just one maintainer now.
If we decide we don't want multiple maintainers run within one process, we will have disallow providing keep-client with multiple maintainer flags.

lukasz-zimnoch: Ok, works for me! Thanks for clarification!

pkg/bitcoin/chain.go
lukasz-zimnoch: > For now, let's stick to Electrum, as that's the name we have in the base branch of my PR.

We can change it even now. The `pkg/bitcoin/electrum` was just supposed to demonstrate the intended package structure. Maybe we can coordinate with @nkuba and change the config accordingly in this PR to have that covered?

tomaszslabon: I added the same Config as @nkuba in #3395 in 9152aed1f0c75575d09f60d66bb71ffe35a558d7.
Updated all the code changed by this PR accordingly.
@nkuba should now have conflicts with this PR.

lukasz-zimnoch (dismissed):
LGTM! Letting the config-related things in @nkuba hands!

cmd/maintainer.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/keep-network/keep-core/config"
+	"github.com/keep-network/keep-core/pkg/maintainer"
+	"github.com/spf13/cobra"
+)
+
+// MaintainerCommand contains the definition of the maintainer command-line
+// subcommand.
+var MaintainerCommand = &cobra.Command{
+	Use:   "maintainer",
+	Short: `Starts keep maintainers`,
```
pdyraga, line 16:
```suggestion
	Short: `Starts tBTC maintainer`,
```

cmd/maintainer.go
```diff
@@ -0,0 +1,64 @@
...
+	Long:  `The maintainer command starts keep maintainers`,
```
pdyraga, line 17:
```suggestion
	Long:  `The maintainer command starts tBTC maintainer`,
```

pkg/maintainer/relay.go
```diff
@@ -0,0 +1,62 @@
...
+// Relay is the part of maintainer responsible for maintaining the state of
+// the relay on-chain contract.
+type Relay struct {
+	btcChain bitcoin.Chain
+	chain    RelayChain
+}
+
+// startControlLoop launches the loop responsible for controlling the relay.
+func (r *Relay) startControlLoop(ctx context.Context) {
+	logger.Info("starting headers relay")
```
pdyraga, line 45:
```suggestion
	logger.Info("starting Bitcoin difficulty maintainer")
```

pkg/maintainer/relay.go
```diff
@@ -0,0 +1,62 @@
...
+
+	defer func() {
+		logger.Info("stopping headers relay")
```
pdyraga, line 48:
```suggestion
		logger.Info("stopping Bitcoin difficulty maintainer")
```

test/config.toml
```diff
@@ -10,6 +10,11 @@ ConcurrencyLimit = 56
 MaxGasFeeCap = "148 Gwei"
 BalanceAlertThreshold = "2.3 ether"
 
+[Electrs]
```
pdyraga, line 13: Given the recent discussion about supporting ElectrumX RPC API, I am not sure if this should be named `[Electrs]` or `[Electrum]`. @nkuba thoughts?

test/config.toml
```diff
@@ -27,6 +32,9 @@ Port = 3498
 NetworkMetricsTick = "43s"
 EthereumMetricsTick = "1m27s"
 
+[maintainer]
+Relay = true
```
pdyraga, line 36: I would start calling it `BitcoinDifficulty` instead of the `Relay`. I think the `Relay` is pretty cryptic. We've been also using "threshold relay" as an alternative name for the random beacon.

pkg/maintainer/maintainer.go
```diff
@@ -0,0 +1,27 @@
...
+	chain RelayChain,
+) {
+	if config.Relay {
+		go newRelay(ctx, btcChain, chain)
+	}
+
+	// TODO: Check if launching multiple maintainers is really beneficial.
+	//       Panic on one maintainer goroutine will crush the whole program.
+	//       Also, if we misconfigure one maintainer and it cannot launch,
+	//       should we cancel all the maintainers that already launched properly?
```
pdyraga, line 22: That's the same problem as with the client handling multiple wallets 🤷🏻 

I think we should allow launching all types of maintainers at once. Even if the service crashed, the service should get restarted.

cmd/maintainer.go
tomaszslabon: Done in 9fe493876ecf200cc755932defeb2498e0026d16.

cmd/maintainer.go
tomaszslabon: Done in 9fe493876ecf200cc755932defeb2498e0026d16.

pkg/maintainer/relay.go
tomaszslabon: Done in 9fe493876ecf200cc755932defeb2498e0026d16.

pkg/maintainer/relay.go
tomaszslabon: Done in 9fe493876ecf200cc755932defeb2498e0026d16.

pkg/maintainer/maintainer.go
tomaszslabon: OK.
I updated the TODO in 43106ec2d46347892753e797b45acc83a0f9eb1d.

test/config.toml
tomaszslabon: @pdyraga 
Should I also change the option for starting from `./keep-client maintainer --relay` to `./keep-client maintainer --bitcoinDifficulty`?

test/config.toml
tomaszslabon: @nkuba @pdyraga 
Alternatively, we could keep the name as `electrs` and treat it as kind of a placeholder and update it once [Kuba's PR](https://github.com/keep-network/keep-core/pull/3406) is merged and we know what config is used there and which fields are required to be configured.

nkuba: We could introduce a parent `bitcoin` section, so:
```
[bitcoin.electrum]
URL=.....
```

If we decide to keep multiple APIs supported we could add them in te future as in the example below:

```
[bitcoin]
[electrum]
URL=.....
[electrs]
URL=....
```

pdyraga: I think `[bitcoin]` section makes a lot of sense, especially since we have `[ethereum]`. I just wonder if we couldn't keep it simple like:
```
[bitcoin]
  ElectrumURL = ""
```

ElectrumURL does not mean it has to be ElectrumX or Electrs. That's just an Electrum API-compatible node.

pdyraga: OK, I've just realized there are more options for electrum than just URL. Then I think `[bitcoin.electrum]` is the best one for now.

test/config.toml
```diff
@@ -27,6 +32,9 @@ Port = 3498
 NetworkMetricsTick = "43s"
 EthereumMetricsTick = "1m27s"
 
+[maintainer]
```
pdyraga, line 35: What do you think about renaming it to `[tbtc.maintainer]`? The reasoning is that it's not a generic maintainer but a maintainer specific for tBTC. That would play nicely with the idea of `[bitcoin.electrum]` I think.

cmd/cmd.go
```diff
@@ -35,6 +35,7 @@ func init() {
 		StartCommand,
 		PingCommand,
 		EthereumCommand,
+		MaintainerCommand,
```
pdyraga, line 38:
```suggestion
		TbtcMaintainerCommand,
```

cmd/flags.go
```diff
@@ -43,6 +46,8 @@ func initFlags(
 			initClientInfoFlags(cmd, cfg)
 		case config.Tbtc:
 			initTbtcFlags(cmd, cfg)
+		case config.Maintainer:
```
pdyraga, line 49:
```suggestion
		case config.TbtcMaintainer:
```

cmd/flags.go
```diff
@@ -35,6 +36,8 @@ func initFlags(
 		switch category {
 		case config.Ethereum:
 			initEthereumFlags(cmd, cfg)
+		case config.Electrs:
```
pdyraga, line 39: If we do in the config `[bitcoin.electrum]`, then this one should probably become:

```suggestion
		case config.BictoinElectrum:
```

cmd/flags.go
```diff
@@ -150,6 +155,30 @@ func initEthereumFlags(cmd *cobra.Command, cfg *config.Config) {
 	)
 }
 
+// Initialize flags for Electrs configuration.
+func initElectrsFlags(cmd *cobra.Command, cfg *config.Config) {
```
pdyraga, line 159: Similarly to `config.BitcoinElectrum`, if we do `[bitcoin.electrum]` in the config, then this one should probably be:
```suggestion
func initBitcoinElectrumFlags(cmd *cobra.Command, cfg *config.Config) {
```

cmd/flags.go
tomaszslabon: Done in b0285cf27f51896294dff419b3dec8cf416e7b49.

cmd/maintainer.go
```diff
@@ -0,0 +1,64 @@
+package cmd
```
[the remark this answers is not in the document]
tomaszslabon: Done in 3fca39275fcfb17c63d9fdf8553f8d369d79a629.

test/config.toml
tomaszslabon: Done in https://github.com/keep-network/keep-core/blob/b26f5297258e43a7f57047a0157e5f3b111adab5/pkg/bitcoin/config/config.go#L8.

Unfortunately, there is some complication: in order to have `bitcoin.electrum` configuration, I had to define [this Config struct](https://github.com/keep-network/keep-core/blob/b26f5297258e43a7f57047a0157e5f3b111adab5/pkg/bitcoin/config/config.go#L8) somwhere.

I couldn't do this in the `bitcoin` package because of circular dependency problem (`electrum` imports `bitcoin` and `bitcoin` would import `electrum`). It would also mean that the specific bitcoin implementation (electrum) leaks into its parent package.

Therefore I decided to create a separate package `pkg/bitcoin/config` that would contain the configs for all the possible bitcoin implementations (electrum, electrs, etc).

I don't like this idea too much. Are you ok with this or should I do sth else?

cmd/flags.go
tomaszslabon: Done in b26f5297258e43a7f57047a0157e5f3b111adab5.

cmd/flags.go
tomaszslabon: Done in b26f5297258e43a7f57047a0157e5f3b111adab5.

test/config.toml
lukasz-zimnoch: Why not just in `pkg/config`? Package `pkg/bitcoin` does not have to contain it.

tomaszslabon: @lukasz-zimnoch 
Yes, I also thought about it.
Done in 96b0e94e3ce9f1faca845bc3e2a3ee5a26a100b5.

BTW, should it be `keep-core/pkg/config` or simply `keep-core/config`? I chose the first option, but I had to create a new directory for `pkg/config`.

pkg/maintainer/bitcoindifficulty.go
```diff
@@ -0,0 +1,63 @@
+package maintainer
```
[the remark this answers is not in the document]
tomaszslabon: Done in 29ad4a69229b533c79cf9f19bc80f2e872bbf31a.

test/config.toml
tomaszslabon: Done in ed67c93fbe4f57b8aaa434aca80e0100f535f31d.

test/config.toml
pdyraga: `pkg/bitcoin` defines Bitcoin chain abstractions and `pkg/bitcoin/electrum` uses these abstractions. I think this dependency makes sense and we should leave it as-is.

On the other hand, all other configuration structures lie in respective packages now: `libp2p`, `storage`, `clientinfo`, `ethereum`... `pkg/config` with `BitcoinConfig` is the only exception which I am not a huge fan of. 

Wouldn't it be better to move `BitcoinConfig` to `pkg/bitcoin` and embed electrum configuration in this file? It is still not ideal but better in my opinion:

```
type BitcoinConfig struct {
   Electrum ElectrumConfig 
}

type ElectrumConfig struct {
	URL string
}
```

Alternatively, introduce `pkg/bitcoin/config` package. Given we have just one configuration option for electrum now, I'd prefer the first solution as the simplest one.

pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -0,0 +1,63 @@
...
+
+var logger = log.Logger("keep-maintainer-bitcoin-difficulty")
+
+// BitcoinDifficultyChain is an interface that provides the ability to
+// communicate with the Bitcoin difficulty on-chain contract.
+type BitcoinDifficultyChain interface {
+	// Retarget adds a new epoch to the Bitcoin difficulty relay by providing
+	// a proof of the difficulty before and after the retarget.
+	Retarget(headers []bitcoin.BlockHeader) error
+}
```
pdyraga, line 19: We could also extract it to `chain.go` and name it just `Chain`. This would be consistent with what we do in `pkg/sortition` and I believe that was our desired organization of packages interacting with chain for v2 client.

pdyraga:
I tried to test it starting the maintainer with `KEEP_ETHEREUM_PASSWORD="password" ./keep-client --config configs/config.1.toml maintainer` but nothing is printed to the console and it looks like if the process hanged. What am I doing wrong?

tomaszslabon:
> I tried to test it starting the maintainer with `KEEP_ETHEREUM_PASSWORD="password" ./keep-client --config configs/config.1.toml maintainer` but nothing is printed to the console and it looks like if the process hanged. What am I doing wrong?

You need to provide the `--bitcoinDifficulty` flag: 
```
KEEP_ETHEREUM_PASSWORD="password" ./keep-client maintainer  --bitcoinDifficulty  --config /Users/tslabon/projekt/keep-core/configs/maintainer.toml 
```

test/config.toml
pdyraga: I looked at `pkg/bitcoin` one more time and you are right @lukasz-zimnoch. This package is just about Bitcoin types, as the referenced documentation explains. From this perspective, we are not configuring anything in `pkg/bitcoin` but we are configuring Electrum.

From the config file's perspective though, I think it makes sense to have the `bitcoin` section, especially that we have `ethereum` section with an URL there and we do not care if it is Geth, Nethermind, Erigon, or others.

What about having`[bitcoin.electrum]` in the config file, just like we do now but having the config structure in `pkg/bitcoin/electrum`?

pdyraga:
I was under the impression that we start all maintainers by default?

test/config.toml
tomaszslabon: > What about having[bitcoin.electrum] in the config file, just like we do now but having the config structure in pkg/bitcoin/electrum?

But what if we ever want to add another typy of Bitcoin implementation (e.g. `electrs`) and use them both, so that we have:
```
[bitcoin]
  [electrum]
    URL = ...
  [electrs]
    URL = ... 
```

We would be defining the Bitcoin structure with `electrum` and `electrs` in electrum?

tomaszslabon:
> I was under the impression that we start all maintainers by default?
I [left a TODO](https://github.com/keep-network/keep-core/blob/29ad4a69229b533c79cf9f19bc80f2e872bbf31a/pkg/tbtc/maintainer/maintainer.go#L21) for that once we have more than one maintainer.

test/config.toml
lukasz-zimnoch: Yeah, this is the best option IMO. Having `bitcoin` section is a pure config thing and it should be declared there.

cmd/flags.go
```diff
@@ -150,6 +152,17 @@ func initEthereumFlags(cmd *cobra.Command, cfg *config.Config) {
...
 
+// Initialize flags for Bitcoin electrum configuration.
+func initBitcoinElectrumFlags(cmd *cobra.Command, cfg *config.Config) {
+	cmd.Flags().StringVar(
+		&cfg.Bitcoin.Electrum.URL,
+		"bitcoin.electrum.url",
+		"",
+		"URL for Bitcoin electrum client connection.",
+	)
+	// TODO: Initialize other config options from `electrum.Config`
```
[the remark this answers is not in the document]
tomaszslabon: There are already 3 additional options in that [config](https://github.com/keep-network/keep-core/blob/d69e971fc734031111699f95eccc3bbbafe4345f/pkg/bitcoin/electrum/config.go#L16).

cmd/flags_test.go
```diff
@@ -309,6 +326,11 @@ func TestFlags_Mixed(t *testing.T) {
 			readValueFunc: func(c *config.Config) interface{} { return c.Ethereum.Account.KeyFile },
 			expectedValue: "./keyfile-path/from/flag",
 		},
+		// Properties provided in the config file and overwritten by the flags.
```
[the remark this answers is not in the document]
tomaszslabon: This test uses [config_flags.toml](https://github.com/keep-network/keep-core/blob/29ad4a69229b533c79cf9f19bc80f2e872bbf31a/test/config_flags.toml#L10) and the value for `bitcoin.electrum.url` is different there.

cmd/tbtc_maintainer.go
```diff
@@ -0,0 +1,64 @@
...
+		clientConfig,
+		config.Ethereum,
+		config.BitcoinElectrum,
+		config.Tbtc,
+	)
+}
+
+// TbtcMaintainer initializes maintainer tasks specified by flags passed to the
+// maintainer command.
+func TbtcMaintainer(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
tomaszslabon: Done in 7685dacdbce95115282ba34cc661eab347ffcea2.

config/category.go
```diff
@@ -5,16 +5,30 @@ type Category int
...
+var StartCmdCategories = []Category{
+	General,
+	Ethereum,
+	Network,
+	Storage,
+	ClientInfo,
+	Tbtc,
+	Developer,
+}
+
```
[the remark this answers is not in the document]
tomaszslabon: Done in f41b9f9ab7152ffe20f2bb779ec6bf61d0b0a6ec.

test/config.toml
tomaszslabon: Moved the `BitcoinConfig` struct to the electrum package in 232b141cac2dc47b6aff6b3e24c924a85aeb157a.
Hope this is what you wanted.

pkg/tbtc/maintainer/bitcoin_difficulty.go
tomaszslabon: Done in 13487d5617e34cb6ae4a17b0001baa452fd57229.
I moved the `BitcoinDifficultyChain` into `chain.go`. 
But I didn't name that structure `Chain`, because we will have more than one maintainer interface (e.g.`SpvChain`).
Is that ok?

pkg/bitcoin/electrum/config.go
```diff
@@ -0,0 +1,13 @@
+package electrum
+
+// BitcoinConfig defines the configuration for Bitcoin.
+type BitcoinConfig struct {
```
pdyraga, line 4: What about naming it just `Config`? We would have `electrum.Config` living in `pkg/bitcoin/electrum` referenced from `[bitcoin.electrum]` section. I think it all makes sense.

test/config.toml
pdyraga: Yep!

cmd/flags.go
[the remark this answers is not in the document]
pdyraga: I see... another PR with potentially conflicting changes. Cool, let's leave it this way!

pkg/bitcoin/electrum/config.go
tomaszslabon: There is already such a [structure](https://github.com/keep-network/keep-core/blob/c87ccffd4c948c25b629641e022c91451826225a/pkg/bitcoin/electrum/config.go#L10)
 in the electrum package and it contains config for electrum (it was added by Kuba and I copied it). It's the structure used by BitcoinConfig.

pdyraga: Right, I think we should have just one config (cc @nkuba) structure named `Config`:
```
package electrum

// Config defines the configuration for the Electrum Bitcoin client.
type Config struct {
  // URL to the Electrum server in format: `hostname:port`.
  URL string
}
```

Based on [this](https://github.com/keep-network/keep-core/pull/3374#discussion_r1024157900) conversation, we agreed that we are *not* configuring Bitcoin but we are configuring Electrum Bitcoin client. The configuration is in `[bitcon.electrum]` in the config and the structure is in `pkg/bitcoin/electrum/config.go`.

tomaszslabon: If we want to use a nested structure in the config file:
```
[bitcoin.electrum]
URL = "url.to.electrum:1111"
```
then the corresponding structure in Golang should also be nested:
```
type BitcoinConfig struct {
 Electrum Config
}

type Config struct {
 	URL string
}
```.

If we don't use a nested structure, but just a simple non-nested `Config` structure, then the `Viper` package that we use for mapping config file to the Golang structure will be unable to read the configuration properly.

pdyraga (approved):

pkg/bitcoin/electrum/config.go
nkuba: I proposed an update in https://github.com/keep-network/keep-core/pull/3406/commits/40ff528ac1797d6ce3cf2a449ef745e10480b538

Issue #3366, opened 2022-10-24
Faster and more resilient ECDSA DKG and signing


pdyraga:
Currently, DKG and signing use a fixed duration of blocks for each protocol phase. The durations are adjusted based on the hardware requirements and the number of operations that must be executed in the given phase.

This is problematic for at least two reasons:
1. Sometimes the fixed duration is not enough to finish the protocol - this is problematic on dev machines locally when the entire group sits on the same machine.
2. For mainnet deployment, when group members are distributed between multiple machines, the fixed protocol duration is usually too slow. 

After a long conversation with @lukasz-zimnoch we proposed to try an alternative approach: an async state machine that waits for the slowest participant or for a bit longer timeout. This approach should be applied both for DKG and signing.

PR #3379, opened 2022-10-25
tECDSA signing: Integration with the asynchronous state machine

lukasz-zimnoch:
Refs: #3366

Here we integrate the tECDSA signing protocol with the asynchronous state machine introduced by https://github.com/keep-network/keep-core/pull/3362. Changes presented in this pull request involve changes in the `pkg/tecdsa/signing` states implementations in order to conform to the asynchronous state interface. As result, the async state machine becomes the default runner used within `signing.Execute` function. Thanks to the changes presented in this PR, the time of a single signing attempt can be greatly reduced as the protocol is no longer fixed and can adjust to the slowest participant.


pkg/tbtc/node.go
```diff
@@ -525,23 +519,43 @@ func (n *node) joinSigningIfEligible(
...
 
+						// Set up the attempt timeout signal.
+						attemptCtx, cancelAttemptCtx := context.WithCancel(
+							loopCtx,
+						)
+						go func() {
+							defer cancelAttemptCtx()
+
+							err := n.waitForBlockHeight(
+								loopCtx,
```
pdyraga, line 530: Should we use the `loopCtx` or `attemptCtx` here?

pkg/tecdsa/signing/states.go
```diff
@@ -2,126 +2,73 @@ package signing
...
 	member  *ephemeralKeyPairGeneratingMember
-
-	phaseMessages []*ephemeralPublicKeyMessage
 }
 
-func (ekpgs *ephemeralKeyPairGenerationState) DelayBlocks() uint64 {
-	return ephemeralKeyPairStateDelayBlocks
-}
+func (ekpgs *ephemeralKeyPairGenerationState) Initiate(ctx context.Context) error {
+	ekpgs.action.run(func() error {
```
pdyraga, line 24: Why do we need to wrap the call to `generateEphemeralKeyPair()` with `ekpgs.action.run`? The state machine should guarantee `Initiate` is called only one time.

pkg/tecdsa/signing/states.go
```diff
@@ -997,3 +793,115 @@ func (fs *finalizationState) MemberIndex() group.MemberIndex {
...
+	channel net.BroadcastChannel,
+	message net.TaggedMarshaler,
+) error {
+	// The channel.Send function sets up retransmissions and expects a
+	// unique context to properly differentiate the sending routine among
+	// others. The context received as argument of the Initiate function
+	// cannot be used directly because it is global for the entire
+	// state machine. Because of that, channel.Send must be called with
+	// a child context.
+	sendCtx := context.WithValue(ctx, "message", message.Type())
```
pdyraga, line 860: Why do we use `WithValue`? Is it to trick the `ticker.go` to distinguish contexts and not treat contexts for two different states as equal?

pkg/tecdsa/signing/states.go
lukasz-zimnoch: This is not about the number of calls to `Initiate`. The reasoning is as follows:
- we need to wait until all ephemeral keys are generated and the state message is sent before we proceed to the next state
- currently, the state machine implementation guarantees that `Initiate` is called before `CanTransition` but this is an implementation detail that can change in the future so we cannot consider it as a strong guarantee. Because of that, we should check whether the state completed its computations and sent its message before proceeding in the state implementation as well. We don't want to check state internals (i.e. the number of generated keys) as this is not the right abstraction level for that. Additionally, that check does not guarantee that `channel.Send` was called
- `Initiate` should return as fast as possible as the message buffer can become filled meanwhile. Running an async action matches that requirement
- Almost all states of signing perform async computations so using this pattern for `ephemeralKeyPairGenerationState` is justified as we have consistency across all states

pkg/tecdsa/signing/states.go
lukasz-zimnoch: We don't have a simple way to create a context copy in Go apart wrapping it with `context.WithValue`. All other functions like `WithCancel` or `WithDeadline` produce a cancel function that must be managed somehow. We can't just ignore it as, in theory, we don't have a guarantee that the parent context is always canceled (that will cause context leaks). On the other hand, we cannot setup a proper call to the cancel function at this level. This is why the `WithValue` trick is a good option here, especially if we treat this solution as a temporary one until we fix the problem directly in `ticker.go`.

pkg/tecdsa/signing/states.go
pdyraga: It seems to me that this is a problem with a state machine and not something the state implementation should worry about.

What do you think about:
- Making `AsyncState.Initiate` asynchronous in the state machine.
- Making `AsyncState.Initiate` return a channel when the computations are done.
- Probably getting rid of `stateAction` structure.

I can try experimenting with this approach, if it works fine, we can open another PR that will be merged immediately after this one.

lukasz-zimnoch: I think that will make the state machine more complicated as we need to examine the state of computations triggered by `Initiate` in `CanTransition`. There is also a risk that we'll specialize the machine too much and it will not fit other possible use cases well enough. Of course, we can always drive some experiments and see.

pdyraga: Agreed, this will move more logic to the state machine but right now this logic is outside of the state machine and complicates state implementation. We do not necessarily need to modify the machine, we could also move some of this logic to the base async state. Anyway, something to experiment with.

Some loose thoughts here:
1. `stateAction` ensures the given action can be triggered only one time. This is something the Initiate interface should be clear about and ensure. The state implementation should just trust this assumption.
 
2. In the async state machine we gather all messages received, also those from the future. Blocking `Initiate` makes sense for the sync state machine, where we may want to initiate a state before we are ready to accept messages. For the async state machine, we want to accept messages all the time and blocking Initiate is not really needed.

3. On the abstract level, the state machine should not be asking the state if it `CanTransition` if `Initiate` of the state has not been completed yet, no matter if `Initiate` is blocking or not. This is happening when `Initiate` is blocking, but requires some more logic from the state if `Initiate` logic is async.

pkg/tbtc/signing.go
```diff
@@ -19,6 +19,10 @@ import (
 	"sort"
 )
 
+// signingAttemptMaxBlockDuration determines the maximum block duration of a
+// single signing attempt.
+const signingAttemptMaxBlockDuration = 100
```
pdyraga, line 24: Just flagging that this value requires more thinking. 100 blocks assuming 12sec for a single block is 20 minutes.

With 2 malicious members in a signing group, we need 5 retries of the protocol in the worst case (`P = (98 choose 51) / (100 choose 51) = 0.23757575757`), which takes 1h 40min.

With 3 malicious members in a signing group, we need 10 retries of the protocol in the worst case (`P = (97 choose 51) / (100 choose 51) = 0.11393939393`), which takes 3h 20 min.

Say we want to sweep no more than 10 deposits at once, this will take almost 17 hours for 2 malicious members case and twice that much for 3 malicious members case. Not blocking this PR, rather something for https://github.com/keep-network/tbtc-v2/pull/414/

pdyraga: ...plus the blocks for announcement and delay!

pdyraga:
The code here so far looks good but I wasn't testing yet. Waiting for two more changes planned by @lukasz-zimnoch:
- delay sending the stop pill,
- exit the state machine's `Execute` when the context is done.

The comments left are for future PRs, I would like to do some experimentation with the state machine based on the comments.

lukasz-zimnoch:
> The code here so far looks good but I wasn't testing yet. Waiting for two more changes planned by @lukasz-zimnoch:
> 
> * delay sending the stop pill,
> * exit the state machine's `Execute` when the context is done.
> 
> The comments left are for future PRs, I would like to do some experimentation with the state machine based on the comments.

Done in https://github.com/keep-network/keep-core/pull/3379/commits/16007ddb8cb3645302a19f1ae3d52fba44ac78ee and https://github.com/keep-network/keep-core/pull/3379/commits/b2b033ece135ebfba5977b8935c2642340c3349b

pkg/tecdsa/signing/states.go
```diff
@@ -997,3 +793,115 @@ func (fs *finalizationState) MemberIndex() group.MemberIndex {
...
+
+	for _, item := range list {
+		if _, exists := senders[item.SenderID()]; !exists {
+			senders[item.SenderID()] = true
+			result = append(result, item)
+		}
+	}
+
+	return result
+}
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, I was thinking about it as well and had the same feelings.

pkg/tbtc/node.go
```diff
@@ -569,13 +569,22 @@ func (n *node) joinSigningIfEligible(
...
-							)
-						}
+						// Schedule the stop pill to be sent a fixed amount of
+						// time after the result is returned. Do not do it
+						// immediately as other members can be very close
+						// to produce the result as well. This mechanism should
+						// be more sophisticated but since it is temporary, we
+						// can live with it for now.
+						go func() {
+							time.Sleep(1 * time.Minute)
```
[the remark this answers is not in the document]
lukasz-zimnoch: This is not a hard requirement. It will work even if this time is longer.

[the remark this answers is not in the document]
pdyraga: But then operators who were excluded from the current attempt may kick off the execution of the next attempt, correct?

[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, but once the stop pill arrives they will break. Even if the pill will not arrive they will just fail eventually.

lukasz-zimnoch:
Tested the following scenarios using a local environment with 3 clients.

Scenario 1: All clients running, signature requested, and produced without errors after 12 blocks.

Scenario 2: All clients running, signature requested but one client goes down during execution. Other clients wait until the 100 blocks timeout gets hit and start the second attempt which completes with success after 15 blocks.

Scenario 3: All clients running, signature requested but resource (CPU) shortage occurs during the processing (DKG round two happens at the same time). Execution slows down but finally, the signature is produced after 20 blocks. 

Scenario 4: All clients running, signature requested but some members start with a delay (temporary hack in the code). Signature produced after 18 blocks.

pdyraga:
Scenarios tested: 

Scenario 1: All clients running, signature requested, and produced successfully.

Scenario 2: All clients running, signature requested but one client goes down during execution. Other clients wait until the 100 blocks timeout gets hit and start the second attempt which completes successfully. 

There are some suggestions left in this PR but I'll try to address them separately. This PR is a huge improvement and the suggestions require some experimentation.

pdyraga (approved):

pkg/tecdsa/signing/states.go
pdyraga: Addressed in #3383.

pkg/tecdsa/signing/states.go
pdyraga: Addressed in https://github.com/keep-network/keep-core/pull/3385.

pkg/tecdsa/signing/states.go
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed in https://github.com/keep-network/keep-core/pull/3386

PR #3385, opened 2022-10-26
Asynchronous state machine: do not block on state initiation

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/3366
See https://github.com/keep-network/keep-core/pull/3379#discussion_r1004285362

In the async state machine we gather all messages received, also those from the "future" states. Blocking `Initiate` makes sense for the sync state machine, where we may want to initiate a state before we are ready to accept messages. For the async state machine, we want to accept messages all the time and blocking `Initiate` is not needed. Having `Initiate` running asynchronously will allow us to `Recv` messages to states and do not overflow the buffer in the state machine in case the computations take longer than expected. It should also allow for a faster execution of states because the messages are received also at the time of Initiating the state so we can potentially transition right after the computations are done.

The state machine should not be asking the state if it `CanTransition` if `Initiate` of the state has not been completed yet. This was guaranteed with a blocking `Initiate` and has been maintained for the non-blocking `Initiate`. I've made it clear in the documentation of the AsyncState.

This change has been done in 428e0132dfb088d212519c42540a0d62cae4bfd5. The next commit, 4a8a8c574dc5bc32971e15bebbd60be4390263c5, is a proof how we could simplify state code with that change. I recommend reviewing the two commits separately.


pdyraga:
Scenarios tested:
1. All 4 clients announcing readiness to sign. Signature calculated, took about 3 minutes.
2. 1/4 clients does not announce the readiness to sign. Signature calculated by the rest of the group. Took about 4 minutes but I had beacon DKG happening at the same time.

I did not test retries but I do not think this change affects them in any way. I will test retries in one of the next PRs when adding session ID to the stop pill.

lukasz-zimnoch (commented):
LGTM! Just one comment!

pkg/protocol/state/async_machine.go
```diff
@@ -103,7 +95,11 @@ func (am *AsyncMachine) Execute() (AsyncState, error) {
 				)
 			}
 
-		case <-onStateDone:
+		case err := <-onStateDone:
+			if err != nil {
+				return nil, fmt.Errorf("failed to initiate new state [%w]", err)
```
lukasz-zimnoch, line 100:
```suggestion
				return nil, fmt.Errorf(
					"failed to initiate state [%T]: [%w]",
					currentState,
					err,
				)
```

lukasz-zimnoch (approved):

PR #3392, opened 2022-11-01
tECDSA DKG: Integration with the asynchronous state machine

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3366
Depends on: #3396

Here we integrate the tECDSA DKG protocol with the async state machine introduced by https://github.com/keep-network/keep-core/pull/3362. The presented changes are analogous to the recent changes introduced to the tECDSA signing protocol by #3379. Long story short, we no longer use a fixed block duration for protocol's states but move to a global timeout for the entire protocol attempt and switch states as soon as possible. As result, the time of a single DKG attempt can be reduced as the protocol is no longer fixed and can adjust to the slowest participant.

The aforementioned change in the DKG protocol also involves a similar change in the DKG result publication step. The publication step is now based on the asynchronous state machine as well. This has two major consequences:
- Firstly, so far, the participants were waiting for at least a group quorum of DKG result signatures for a fixed block duration and then moved to the result submission state. Now, all members wait for the actual group size of DKG result signatures and move to the result submission. This way, we are introducing an additional health check of group participants at the result publication step thus we maximize the final number of operating group members. This is based on the assumption that if DKG went to the publication step, all participants had to complete the computation step successfully, so they should be still operating for publication. 
- Secondly, the result submission queue is now time-based. Since the publication step relies on the async state machine, we can no longer use the block counter to set the submission eligibility queue. Each member preserves a time delay, according to its member index, before attempting to submit the DKG result

Last but not least, we were forced to improve the message retransmission mechanism. As the DKG's TSS round two is computationally expensive, it takes ~100 blocks to complete their computations. The current retransmission strategy retransmits every message on every new block. On TSS round two, each member already retransmits messages from the ephemeral key generation state and from TSS round one. If the TSS round two state takes ~100 blocks, the amount of messages flying around the network is really huge and often caused network congestion. Also, the TSS round two state always takes almost all available CPU so, the situation became even worse. This resulted in a poor DKG success rate. To improve the situation and lower the number of messages passed through the network, we introduced a retransmission backoff strategy (#3396). All messages are now retransmitted at a decreasing rate so older messages are retransmitted less and less frequently. This allowed us to achieve a good success rate of the DKG protocol.

Changes presented in this PR were tested on a local environment with three clients. Performed five DKGs and all were completed with success after ~18 minutes. That means we achieved an improvement as the DKG based on the sync machine always took ~26 minutes (fixed duration of 125 computation blocks and 5 publication blocks with an average block time of 12 seconds)


pkg/tecdsa/dkg/states.go
```diff
@@ -416,167 +301,93 @@ func (trts *tssRoundThreeState) MemberIndex() group.MemberIndex {
...
-	}, nil
-}
-
-func (fs *finalizationState) MemberIndex() group.MemberIndex {
-	return fs.member.id
-}
-
-// confirmationState is the state during which members confirm a successful
-// course of the distributed key generation process.
-type confirmationState struct {
```
pdyraga, line 501: Checking my understanding of why we can remove `confirmationState`.

From #3340:

> The problematic case is when members fail at the last phase of DKG (finalization round) which is not followed by any communication round allowing for detection of such failures. In that scenario, members who failed the last round do not produce the key material but, at the same time, are not excluded by other participants as misbehaved, so they can be selected for signing normally. As a result, signature attempts including those members fail and the wallet needs more attempts in total to exclude those members using the random algorithm and come up with a successful signature.

Some members could fail to execute `fs.member.tssFinalize` in time and the group did not know about it. In #3347 we introduced `tssFinalizationMessage` and one more phase after `finalizationState` to confirm enough group members sent `tssFinalizationMessage`.

That phase after `finalizationState` is no longer needed because `CanTransition()` will block until `fs.member.group.OperatingMemberIDs()` `tssFinalizationMessage`s are received. This can be translated to: everyone who participated in DKG completed computations in `fs.member.tssFinalize`.

Does it make sense?

pkg/tecdsa/dkg/states.go
```diff
@@ -416,167 +301,93 @@ func (trts *tssRoundThreeState) MemberIndex() group.MemberIndex {
...
-	return silentStateActiveBlocks
-}
-
-func (cs *confirmationState) Initiate(ctx context.Context) error {
-	cs.member.markInactiveMembers(cs.previousPhaseMessages)
-
-	if len(cs.member.group.InactiveMemberIDs()) > 0 {
-		return newInactiveMembersError(cs.member.group.InactiveMemberIDs())
-	}
+func (fs *finalizationState) CanTransition() bool {
```
pdyraga, line 343: How do we ensure all members had a chance of receiving messages in this state? 

https://github.com/keep-network/keep-core/blob/254aa487f60fdc1662d10a77fd55c4a6ecb4d4d4/pkg/protocol/state/state.go#L92-L125

pkg/tecdsa/dkg/protocol_test.go
```diff
@@ -1313,144 +1313,6 @@ func TestVerifyDKGResultSignatures(t *testing.T) {
 				3:                         []byte("sign 3"),
 			},
 		},
-		"multiple messages from other member with different signatures for " +
```
pdyraga, line 1316: My understanding of the reasoning behind removing these unit tests is that all these cases do not matter because only the first message is accepted. On the other hand, there is no guarantee the deduplication code is called from `pkg/tescdsa/dkg/states.go` so maybe it would make sense to leave one or two of these tests that would cover that the deduplication works properly? Right now, looking at just the unit tests, it is unclear what happens if the same member sends two messages with correct or incorrect signature.

pkg/tecdsa/dkg/states.go
lukasz-zimnoch: Yep, exactly!

pkg/tecdsa/dkg/states.go
lukasz-zimnoch: Basically, if a given member collected all the required messages to exit that state, we can say for sure that:
- the given member sent their message (their `Initiate` function completed for sure)
- other members are listening for that message (if other members sent their messages, their `Initiate` function is completed for sure and they are listening for incoming messages)

With the current construction of the async machine and the DKG protocol, there is no option that other members are behind the finalization state when the given member exits. I think that should be enough to fulfill the requirement "all members had a chance of receiving messages in this state". Am I missing something?

pkg/tecdsa/dkg/protocol_test.go
lukasz-zimnoch: We need to remove those tests because we made a change in messages validation. So far, each `protocol.go` function was responsible for duplication of its own state messages, hence the test cases. Now, we put the deduplication on the higher abstraction level and this is done in `states.go` so `protocol.go` functions work with an assumption that the messages received as input are correctly deduplicated, hence we need to remove those cases.

pkg/tecdsa/dkg/states.go
pdyraga: I am thinking about the following scenario:
1. We have two members that exited `tssRoundThreeState` more or less at the same time.
2. There was some delay with the second member and they entered `finalizationState` after the first member completed `Initiate` and sent their message.
3. Now the first member waits for a message from the second member and the second member waits for a retransmission of a message already sent by the first member.
4. The second member completes `Initiate`, and they sent their message.
5. The first member has all the messages and is exiting the state machine, closing the context, before the retransmission happens.
6. The second member hangs.

```
M1:   | Initiate() + send ↓                     | context.Done()         R <- here we would have a retransmission         
M2:                         | Initiate + send ↑ ..................
```

pkg/tecdsa/dkg/protocol_test.go
pdyraga: I see. I think we should implement unit tests, similar to the ones in `pkg/beacon/dkg/result/states_test.go` where we instantiate member and state manually, to confirm the state's `Initiate` uses `receivedMessages` to filter out duplicates. Alternatively, at least cover the `receivedMessages` with a unit test.

pkg/tecdsa/dkg/states.go
lukasz-zimnoch: Yeah, I got it. If we ask "did M2 have a chance to receive M1's message?" the answer is "yes, it had". M2 listened to M1's message from the very beginning and M1 sent it for sure and there is a chance it retransmitted it at least once. The real question is "did M2 have enough time to receive M1's message?". Well, this is not so simple and I'm not sure if we can guarantee that on the machine level. We can add a delay for M1, of course. But, what is the right duration of the delay? We can also extend the broadcast channel's `Send` by adding an option to guarantee N retransmissions. Not sure if this is the right direction though.

pkg/tecdsa/dkg/protocol_test.go
lukasz-zimnoch: Added unit tests for `receivedMessages`, just like we have for signing: https://github.com/keep-network/keep-core/pull/3392/commits/846f158d4c91ab2947bdb831d42ba35ce7405cf8

pkg/tecdsa/dkg/states.go
pdyraga: After revisiting it one more time, I think we are actually good. The context for an attempt is active until the timeout for the attempt, even if the attempt finished earlier. It means that for a good timeout, we should have enough retransmissions for everyone:

https://github.com/keep-network/keep-core/blob/254aa487f60fdc1662d10a77fd55c4a6ecb4d4d4/pkg/tbtc/node.go#L534-L553

Does it make sense?

lukasz-zimnoch: Ah yeah! That's right!

pkg/tbtc/node.go
```diff
@@ -304,12 +336,22 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
+				// Set up the publication stop signal that should allow to
+				// perform all the result-signing-related actions and
+				// handle the worst case when the result is submitted by the
+				// last group member.
+				publicationCtx, cancelPublicationCtx := context.WithTimeout(
+					context.Background(),
+					dkgResultSigningDuration+
+						(time.Duration(chainConfig.GroupSize)*dkgResultSubmissionDelayStep),
+				)
+				defer cancelPublicationCtx()
```
pdyraga, line 348: Don't take it for granted, I am still testing/experimenting but I think this construction of a context does not work well in all cases. I have been running tests locally with 4 clients. 3 clients managed to submit the result successfully. One client got stuck on `resultSigningState` and timed out after 20 minutes.

Looking at the logs, I think everyone completed work before client 4 started it, and given the original message did not get retransmitted, the client got stuck:

From the client 3:
``` 
2022-11-04T14:52:29.002+0100	INFO	keep-tbtc	state/async_machine.go:179	[member:12,state:*dkg.resultSigningState] transitioned to new state	{"seed": "0x63bcb8b9e18ce096aacd89b73bad7f4c7da5307507c5a329e31948d6c27bed58"}
2022-11-04T14:52:29.095+0100	INFO	keep-tbtc	state/async_machine.go:179	[member:97,state:*dkg.resultSigningState] transitioned to new state	{"seed": "0x63bcb8b9e18ce096aacd89b73bad7f4c7da5307507c5a329e31948d6c27bed58"}
2022-11-04T14:52:29.103+0100	INFO	keep-tbtc	state/async_machine.go:179	[member:12,state:*dkg.signaturesVerificationState] transitioned to new state{"seed": "0x63bcb8b9e18ce096aacd89b73bad7f4c7da5307507c5a329e31948d6c27bed58"}
(...)
2022-11-04T14:52:40.009+0100	INFO	keep-tbtc	state/async_machine.go:148	[member:96,state:*dkg.signaturesVerificationState] transitioning to a new state	{"seed": "0x63bcb8b9e18ce096aacd89b73bad7f4c7da5307507c5a329e31948d6c27bed58"}
(...)
2022-11-04T14:52:40.114+0100	INFO	keep-tbtc	state/async_machine.go:148	[member:96,state:*dkg.resultSubmissionState] transitioning to a new state	{"seed": "0x63bcb8b9e18ce096aacd89b73bad7f4c7da5307507c5a329e31948d6c27bed58"}
(...)
2022-11-04T14:52:40.331+0100	INFO	keep-tbtc	state/async_machine.go:117	[member:19,state:*dkg.resultSubmissionState] reached final state	{"seed": "0x63bcb8b9e18ce096aacd89b73bad7f4c7da5307507c5a329e31948d6c27bed58"}
```

From the client 4:
```
2022-11-04T14:52:29.102+0100	INFO	keep-tbtc	state/async_machine.go:148	[member:78,state:*dkg.resultSigningState] transitioning to a new state	{"seed": "0x63bcb8b9e18ce096aacd89b73bad7f4c7da5307507c5a329e31948d6c27bed58"}
2022-11-04T14:52:29.102+0100	INFO	keep-tbtc	state/async_machine.go:179	[member:78,state:*dkg.resultSigningState] transitioned to new state	{"seed": "0x63bcb8b9e18ce096aacd89b73bad7f4c7da5307507c5a329e31948d6c27bed58"}
(...)
2022-11-04T15:14:08.959+0100	WARN	keep-tbtc	tbtc/node.go:369	[member:29] DKG result publication process failed [context deadline exceeded]	{"seed": "0x63bcb8b9e18ce096aacd89b73bad7f4c7da5307507c5a329e31948d6c27bed58"}
```

I think it is the scenario that we are discussing [here](https://github.com/keep-network/keep-core/pull/3392#discussion_r1013854529) and we could solve it the same way - not canceling the context until the timeout. Something like this:

```
publicationCtx, cancelPublicationCtx := context.WithCancel(context.Background())
go func() {
  time.Sleep(
    dkgResultSigningDuration +
      time.Duration(chainConfig.GroupSize)*dkgResultSubmissionDelayStep,
    )
  cancelPublicationCtx()
}()
```

pdyraga: After testing, this suggestion looks like a good one. I had multiple scenarios where due to spending more time in the result signing phase, some clients were still waiting for messages while other clients completed the execution by submitting the result:

From client 1/4:
```
2022-11-04T18:13:09.032+0100	INFO	keep-tbtc	state/async_machine.go:179	[member:5,state:*dkg.resultSubmissionState] transitioned to new state	{"seed": "0xf0a34daccf046f255f09723ee923de695f7c4236981a56cc5287125041c25d17"}
```

From client 4/4:
```
2022-11-04T18:13:28.144+0100	INFO	keep-tbtc	dkg/states.go:448	[member:43, resultSigningState] CanTransition() has [99] messages received	{"seed": "0xf0a34daccf046f255f09723ee923de695f7c4236981a56cc5287125041c25d17"}
2022-11-04T18:13:28.144+0100	INFO	keep-tbtc	state/async_machine.go:148	[member:34,state:*dkg.signaturesVerificationState] transitioning to a new state	{"seed": "0xf0a34daccf046f255f09723ee923de695f7c4236981a56cc5287125041c25d17"}
```

lukasz-zimnoch: Here it is https://github.com/keep-network/keep-core/pull/3392/commits/0d176312d6450d92544b7dd19ec24979d5b1851a

pkg/tbtc/node.go
```diff
@@ -340,12 +340,16 @@ func (n *node) joinDKGIfEligible(seed *big.Int, startBlockNumber uint64) {
...
 					context.Background(),
-					dkgResultSigningDuration+
-						(time.Duration(chainConfig.GroupSize)*dkgResultSubmissionDelayStep),
+					publicationTimeout,
 				)
-				defer cancelPublicationCtx()
+				go func() {
+					defer cancelPublicationCtx()
+					time.Sleep(publicationTimeout)
+				}()
```
pdyraga, line 352: Do we need this goroutine given that we are constructing `context.WithTimeout`?

lukasz-zimnoch: This addresses https://github.com/keep-network/keep-core/pull/3392#discussion_r1014211031

pdyraga: Right, what I mean is that we create context `WithTimeout` which closes the `Done` channel after `publicationTimeout` even without this goroutine. From the docs:

> The returned context's Done channel is closed when the deadline expires, when the returned cancel function is called, or when the parent context's Done channel is closed, whichever happens first.

In other words, this goroutine does the same `context.WithTimeout` already does internally.

In the future, we will cancel the publication context earlier, when the result is published to the chain. So I think we should add a TODO here to avoid confusion. Something like:

 ```
// TODO: Call cancelPublicationCtx() when the result is available
// and published and remove this goroutine.
// This goroutine is duplicating context.WithTimeout work right now.
go func() {
  defer cancelPublicationCtx()
  time.Sleep(publicationTimeout)
}()
```

Alternatively, remove this goroutine and silence the linter warning explaining that the cancel function needs to be called in the future.

Another option would be to, switch this construction to `context.WithCancel` and have this goroutine.

lukasz-zimnoch: Ah got it! Here it is https://github.com/keep-network/keep-core/pull/3392/commits/45ed1f3300373eb6b6c763af406de4edfff5fd51

pdyraga (approved):

pdyraga:
🚀 🚀 🚀

PR #3393, opened 2022-11-02
Add `staticcheck` to Go pre-commit hooks

michalinacienciala:
[The Go's `lint` tool got deprecated](https://github.com/golang/lint) some time ago. It can be replaced by the `staticcheck` tool, which offers similar functionalities.

In the `keep-network/keep-core` project we've previously added a GH Action job that runs the `staticcheck` tool as part of the CI. We want now to also add it to the Go pre-commit hooks, so that developers could be informed about the problems in their code before they push it to the remote repository. In a separate PR we've adding a new custom hook to the `keep-network/pre-commit-golang` repository. In this PR we're changing the config to use that new custom hook. We're also removing deprecated `lint` hook from the config.
Apart from modifyng the `.pre-commit-config.yaml` we're also updating the `macos-setup.sh` script - so that it would be in line with the modified config.

To adjust the pre-commit hooks on local machine to the new configuration users should:
1. Install the `staticcheck` tool:
```
go install honnef.co/go/tools/cmd/staticcheck@latest
```
2. Apply the pre-commit hooks config:
```
pre-commit install --install-hooks
```

Ref: https://github.com/keep-network/pre-commit-golang/pull/5

TODO:

- [x] Update the `rev` value after https://github.com/keep-network/pre-commit-golang/pull/5 gets merged
- [x] Set the `client-lint` and `client-vet` as required for PR merge (@nkuba)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
  - repo: https://github.com/keep-network/pre-commit-golang.git
-   rev: 979f199 # update after merging https://github.com/keep-network/pre-commit-golang/pull/5 
+   rev: 466ade9296d8445d520ec7ab246a30cfa8a07b71
    hooks:
     - id: go-imports
     - id: go-vet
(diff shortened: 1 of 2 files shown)


tomaszslabon (dismissed):

michalinacienciala:
@tomaszslabon, I've updated `rev` value to point to the commit merging https://github.com/keep-network/pre-commit-golang/pull/5, ready for final review.

tomaszslabon (approved):

PR #3394, opened 2022-11-02
Bitcoin difficulty maintainer: Logic

tomaszslabon:
This PR adds the actual logic of Bitcoin difficulty maintainer.
The logic summary:
- once the Bitcoin difficulty maintainer is started, it checks whether it is eligible to proceed with proving Bitcoin epochs (it checks whether the genesis has been performed and whether it is authorised to submit headers).
- if it is authorised, it keeps proving epochs by providing block headers on each side of the retarget (equal number of headers from the old epoch and new epoch).
- after an epoch is proven, the maintainer waits until the epoch proof can be processed by the on-chain contract.
- if all the epochs has been proven (i.e. we have reached the tip of the Bitcoin chain), the maintainer goes to sleep and checks periodically whether there is an epoch to be proven.
- if an error (e.g. no connection with clients, on-chain processing error, etc.) happens, the maintainer restarts its control loops and continues with trying proving epochs.


pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -2,54 +2,271 @@ package maintainer
...
+	// 522144 <- new difficulty target (first block of the new epoch)
+	// 522145 <- new difficulty target
+	// 522146 <- new difficulty target
+	firstBlockHeaderHeight := uint(newEpochHeight - proofLength)
+	lastBlockHeaderHeight := uint(newEpochHeight + proofLength - 1)
+
+	// The required range of block headers can be pulled from the Bitcoin
+	// blockchain only if the blockchain height is equal to or greater than
+	// the end of the range.
+	if currentBlockNumber >= lastBlockHeaderHeight {
```
tomaszslabon, line 210: One thing to note here is that after each Bitcoin epoch we will have to wait for a while before we can prove it.
For example, if `proofLength` is 20 blocks, we will have to wait 20 Bitcoin blocks (about 3 hours and 20 minutes) before the  new epoch is proven in the `LightRelay`.

If the wallet makes some transaction that requires a proof and the transaction is made shortly after the Bitcoin difficulty epoch is changed, we will not be able to prove it in the Bridge for 20 blocks. 

 This window lasting `proofLength` blocks will happen every ~2 weeks.
 Are we ok with that?

pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -2,54 +2,271 @@ package maintainer
...
+					"the Bitcoin difficulty chain [%v]",
+				firstBlockHeaderHeight,
+				lastBlockHeaderHeight,
+				err,
+			)
+		}
+
+		logger.Infof(
+			"Successfully submitted block headers [%v:%v] to the Bitcoin "+
+				"difficulty chain. The current proven epoch is %v.",
```
nkuba, line 234:
```suggestion
			"Successfully submitted block headers [%d:%d] to the Bitcoin "+
				"difficulty chain. The current proven epoch is %d.",
```

pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -2,54 +2,271 @@ package maintainer
...
+		if err != nil {
+			return fmt.Errorf(
+				"failed to get block headers from Bitcoin chain [%v]",
+				err,
+			)
+		}
+
+		if err := bdm.chain.Retarget(headers); err != nil {
+			return fmt.Errorf(
+				"failed to submit block headers from range [%v, %v] to "+
```
nkuba, line 224: To be consistent with an info message logged a couple of lines below 
```suggestion
				"failed to submit block headers [%d:%d] to "+
```

pkg/tbtc/maintainer/bitcoin_difficulty.go
nkuba, line 234: To be consistent with the codebase can we update all the logged messages to start lowercase?
```suggestion
			"successfully submitted block headers [%d:%d] to the Bitcoin "+
				"difficulty chain; the current proven epoch is %d",
```

@pdyraga do we have any guidelines for log and error messages written down?

pkg/tbtc/maintainer/chain.go
```diff
@@ -2,12 +2,38 @@ package maintainer
...
+	// AuthorizationRequired checks whether the relay requires the address
+	// submitting a retarget to be authorised in advance by governance.
+	AuthorizationRequired() (bool, error)
+
+	// IsAuthorized checks whether the given address has been authorised by
+	// governance to submit a retarget.
+	IsAuthorized(address chain.Address) (bool, error)
+
+	// Signing returns the signing associated with the chain.
+	Signing() chain.Signing
```
nkuba, line 26: I'm not sure we should define this function in `BitcoinDifficultyChain` interface. It is already exposed in the `tbtc.Chain` interface:
 https://github.com/keep-network/keep-core/blob/e49fef51fdb56e470a49577cca339d05c9ab0f15/pkg/tbtc/chain.go#L119

pkg/tbtc/maintainer/chain.go
```diff
@@ -2,12 +2,38 @@ package maintainer
```
nkuba, line 2: Shouldn't the `maintainer` package be moved out of the `tbtc` package to `pkg/maintainer` like we have `pkg/sortition`?

If we decide to leave it as is, shouldn't the `BitcoinDifficultyChain` interface be linked to the `tbtc.Chain` interface just like other contracts?

cc @pdyraga and @lukasz-zimnoch

pkg/tbtc/maintainer/chain.go
```diff
@@ -2,12 +2,38 @@ package maintainer
...
 // communicate with the Bitcoin difficulty on-chain contract.
 type BitcoinDifficultyChain interface {
-	// Retarget adds a new epoch to the Bitcoin difficulty relay by providing
-	// a proof of the difficulty before and after the retarget.
-	Retarget(headers []bitcoin.BlockHeader) error
+	// Ready checks whether the relay is active (i.e. genesis has been performed).
+	// Note that if the relay is used by querying the current and previous epoch
+	// difficulty, at least one retarget needs to be provided after genesis;
+	// otherwise the prevEpochDifficulty will be uninitialised and zero.
+	Ready() (bool, error)
```
nkuba, line 15:
```suggestion
	IsReady() (bool, error)
```

pkg/tbtc/maintainer/chain.go
```diff
@@ -2,12 +2,38 @@ package maintainer
...
+
+	// AuthorizationRequired checks whether the relay requires the address
+	// submitting a retarget to be authorised in advance by governance.
+	AuthorizationRequired() (bool, error)
```
nkuba, line 19:
```suggestion
	IsAuthorizationRequired() (bool, error)
```

pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -2,54 +2,271 @@ package maintainer
...
+		)
 	}
+
+	for {
+		if err := bdm.proveNextEpoch(); err != nil {
+			return fmt.Errorf("cannot prove Bitcoin blockchain epoch [%v]", err)
+		}
+
+		select {
+		case <-time.After(bdm.epochProvenBackOffTime):
```
nkuba, line 98: What is the reasoning behind using the backoff here? As I understand, we can have a couple of situations:
1. `bdm.proveNextEpoch()` returns an error, right now we return the error, so `epochProvenBackOffTime` won't be used.
2.  `bdm.proveNextEpoch()` succeeds but the relay contract is much behind the bitcoin chain, so a next epoch could be submitted to the relay right away.
3. `bdm.proveNextEpoch()` succeeds for the most recent epoch, and in this case, it's true that we should delay the next loop iteration.

pkg/tbtc/maintainer/chain.go
nkuba: NVM, this is a function named after a `LightRelay` contract's method.

pkg/tbtc/maintainer/chain.go
```diff
@@ -2,12 +2,38 @@ package maintainer
 
 import (
 	"github.com/keep-network/keep-core/pkg/bitcoin"
+	"github.com/keep-network/keep-core/pkg/chain"
 )
 
 // BitcoinDifficultyChain is an interface that provides the ability to
 // communicate with the Bitcoin difficulty on-chain contract.
 type BitcoinDifficultyChain interface {
```
nkuba, line 10: Shouldn't this interface be named after the contract as we do with other contracts?
```suggestion
type LightRelayChain interface {
```

pkg/bitcoin/local.go
```diff
@@ -0,0 +1,73 @@
+package bitcoin
+
+import "fmt"
+
+// LocalChain represents a local Bitcoin chain.
+type LocalChain struct {
```
lukasz-zimnoch, line 6: I think that should be placed inside `tbtc/maintainer` package. We rejected the idea of the single local chain implementation for `pkg/chain`'s interfaces in favor of per-application local chains (specific local implementations for `pkg/beacon`, `pkg/tbtc` and so on). I think we should take the same approach for `bitcoin.Chain` implementations. The per-application local chain approach is much more flexible and can be heavily customized in an app-specific way without creating a single monster implementation trying to satisfy all of the use cases. Of course, the per-app approach incurs some code duplication but this will be rather small so it's still worth it.

pkg/tbtc/maintainer/chain.go
lukasz-zimnoch: Yep, we should definitely have `pkg/maintainer` from the very beginning. Moreover, `pkg/maintainer` should be standalone, just like `pkg/sortition`, and should not depend on `pkg/tbtc` in any way. That means it should declare it's own `maintainer.Chain` interface that will expose their chain's expectations.

pkg/tbtc/maintainer/chain.go
lukasz-zimnoch: If we do https://github.com/keep-network/keep-core/pull/3394/files#r1038165482, I think it should. The `BitcoinDifficultyChain/LightRelayChain` must not depend on the `tbtc.Chain` and should declare its own expectations. That will lead to the situation where both `tbtc` and `maintainer` packages will declare their own `Chain` interfaces both focused on the tBTC sub-domains specific for those packages. For example,
- `tbtc.Chain` will declare `Bridge` functions essential for operations performed by the wallets itself, like getting revealed deposits and so on
- `maintainer.Chain` (apart from light relay functions) will declare `Bridge` functions responsible for submitting SPV proofs and so on

Does it make sense

pkg/tbtc/maintainer/bitcoin_difficulty.go
tomaszslabon: We have to wait here, because we need some time before the `LightRelay` smart contract is updated (the transaction has to be mined).
If we didn't wait here, the contract's `CurrentEpoch()` would return the same number and we would attempt to submit the same headers as before.

pkg/tbtc/maintainer/chain.go
tomaszslabon: @nkuba 
There are different ideas how to name this interface.
@pdyraga suggested naming it `BitcoinDifficultyChain` and avoid using `Relay` in the name to avoid confusion with other parts of our system.

pkg/tbtc/maintainer/chain.go
tomaszslabon: On the other hand all the maintainers will perform work that pertains to Tbtc.
@pdyraga [suggested](https://github.com/keep-network/keep-core/pull/3374#discussion_r1025290903) that the `maintainer` functionalities should be inside `/pkg/tbtc/maintainer` and not `/pkg/maintainer`.

pkg/tbtc/maintainer/chain.go
```diff
@@ -2,12 +2,38 @@ package maintainer
...
+	// Note that if the relay is used by querying the current and previous epoch
+	// difficulty, at least one retarget needs to be provided after genesis;
+	// otherwise the prevEpochDifficulty will be uninitialised and zero.
+	Ready() (bool, error)
+
+	// AuthorizationRequired checks whether the relay requires the address
+	// submitting a retarget to be authorised in advance by governance.
+	AuthorizationRequired() (bool, error)
+
+	// IsAuthorized checks whether the given address has been authorised by
```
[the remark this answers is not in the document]
tomaszslabon: Done in 8afe104a5c44adcbbedd8f93be0f064f96f422ef.

pkg/tbtc/maintainer/chain.go
```diff
@@ -2,12 +2,38 @@ package maintainer
...
 // BitcoinDifficultyChain is an interface that provides the ability to
 // communicate with the Bitcoin difficulty on-chain contract.
 type BitcoinDifficultyChain interface {
-	// Retarget adds a new epoch to the Bitcoin difficulty relay by providing
-	// a proof of the difficulty before and after the retarget.
-	Retarget(headers []bitcoin.BlockHeader) error
+	// Ready checks whether the relay is active (i.e. genesis has been performed).
+	// Note that if the relay is used by querying the current and previous epoch
+	// difficulty, at least one retarget needs to be provided after genesis;
+	// otherwise the prevEpochDifficulty will be uninitialised and zero.
```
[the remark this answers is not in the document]
tomaszslabon: Done in 8afe104a5c44adcbbedd8f93be0f064f96f422ef.

pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -2,54 +2,271 @@ package maintainer
...
+			"The Bitcoin difficulty chain is up-to-date with the " +
+				"Bitcoin blockchain",
+		)
+	}
+
+	return nil
+}
+
+// getBlockHeaders returns block headers from the given range.
+func (bdm *BitcoinDifficultyMaintainer) getBlockHeaders(
```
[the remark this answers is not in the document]
tomaszslabon: We could replace `GetBlockHeader(height)` with `GetBlockHeaders(startHeight, count)`.
But I would stick with returning the `bitcoin.BlockHeader` structure and not serialized data, because the Bitcoin chain interface could possibly be re-used by many different applications and therefore we can't make assumption they will all need serialized data. Returning an unserialized representation of Bitcoin header (or an array of headers)  seems the natural way to do this.

pkg/bitcoin/local.go
tomaszslabon: Done in 2e2559f549d7b50daed63993c9a017edb8841713.

pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -2,54 +2,271 @@ package maintainer
...
+	// error logs in case of a permanent error in the Bitcoin difficulty
+	// maintainer.
+	defaultRestartBackoffTime = 120 * time.Second
+
+	// Default value for backoff time which should be applied after each attempt
+	// to prove a single Bitcoin epoch.
+	defaultEpochProvenBackOffTime = 60 * time.Second
+
+	// The number of blocks in a Bitcoin epoch.
+	bitcoinEpochLength = 2016
```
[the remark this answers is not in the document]
tomaszslabon: Done in 65c293286286b9c46feaf313d7044407941fedad.

pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -2,54 +2,271 @@ package maintainer
...
+
+	// The new epoch to be proven in the Bitcoin difficulty chain.
+	newEpoch := currentEpoch + 1
+
+	// Height of the first block of the new epoch.
+	newEpochHeight := newEpoch * bitcoinEpochLength
+
+	// The range of block headers to be pull from the Bitcoin chain should
+	// start `proofLength` blocks before the first block of the new epoch
+	// and end `proofLength`-1 after it.
```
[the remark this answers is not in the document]
tomaszslabon: Done in 524d5af294ad4edcbc97c76e134d6739056c606b.

pkg/tbtc/maintainer/bitcoin_difficulty.go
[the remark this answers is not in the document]
lukasz-zimnoch: Yep, I agree with Tomasz. Returning `bitcoin.BlockHeader` slice seems to be reasonable as it is more flexible. The `BlockHeader` structure exposes the `Serialize` method so use cases that need serialization can be satisfied easily.

pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -2,54 +2,271 @@ package maintainer
...
+			err,
+		)
+	}
+
+	// The new epoch to be proven in the Bitcoin difficulty chain.
+	newEpoch := currentEpoch + 1
+
+	// Height of the first block of the new epoch.
+	newEpochHeight := newEpoch * bitcoinEpochLength
+
```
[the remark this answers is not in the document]
tomaszslabon: I think this assertion is unnecessary, because the smart contract guarantees  `0` < `proofLength` < `2016` and we know `newEpochHeight` will always be `2016` or higher, because it's obtained like that:
```
newEpoch := currentEpoch + 1
newEpochHeight := newEpoch * bitcoinDifficultyEpochLength
```
and `currentEpoch` is always 0 or more.

pkg/tbtc/maintainer/chain.go
lukasz-zimnoch: Yeah, that's right but that depends on how we look at our package structure. Currently, we have two packages representing the standalone applications that can be run as part of the client:
- `pkg/beacon` 
- `pkg/tbtc` 

We also have auxiliary packages that can be used by applications to do their work:
- `pkg/bitcoin`
- `pkg/chain`
- `pkg/net`
- `pkg/sortition`
- `pkg/tecdsa`

In the aforementioned structure, the `pkg/tbtc` package actually represents the code of the tbtc node that performs actions specific to a wallet member. This package does not represent the entire tbtc domain so it should not be considered a bag for everything related to tbtc. 

The typical usage of sub-packages in Go is to place implementation adapters. Examples:
- `pkg/net` contains the network domain and `pkg/net/libp2p` its LibP2P-based implementation
- `pkg/bitcoin` contains a part of the Bitcoin domain and `pkg/bitcoin/electrum` holds an Electrum implementation of the Bitcoin client
- `pkg/tecdsa` contains the tECDSA domain objects, the `pkg/tecdsa/dkg` defines how to generate a key, and the `pkg/tecdsa/signing` implements the signing algorithm
- standard library's `net/http` contains the HTTP-related domain and `net/http/httptrace` is an implementation of tracing

In that context, the tbtc maintainer is not an implementation of the tbtc wallet node. It's a standalone being within the same tbtc domain but with its own bounded context (focused around SPV actions). This is why I think it should be placed into its own `pkg/maintainer` package. Additionally, if we need to add another kind of maintainer, not-related to tbtc, the `pkg/maintainer` will be a good match for it. For example, we could argue if the Bitcoin difficulty maintainer is actually something specific to tbtc. Of course it is used as part of tbtc but this is a broader concept that can be reused by other applications.

An open question is whether `pkg/tbtc` should remain as is or be changed to something specific, reflecting that this code is actually about the tbtc wallet node. However, I have no idea about a good proposition here.

pkg/tbtc/maintainer/local_bitcoin_chain_test.go
```diff
@@ -0,0 +1,88 @@
+package maintainer
```
lukasz-zimnoch, line 1: Let's rename it to `bitcoin_chain_test.go`. The `local` prefix is redundant as this is a test file that suggests we are not dealing with a real implementation.

pkg/tbtc/maintainer/local_bitcoin_difficulty_chain_test.go
```diff
@@ -0,0 +1,130 @@
+package maintainer
```
lukasz-zimnoch, line 1: Maybe just `chain_test.go`? This will be shorter and will indicate that this is a test implementation of interfaces placed in `chain.go` (we have the same pattern in `pkg/tbtc`).

pkg/tbtc/maintainer/chain.go
lukasz-zimnoch: Yep, I'd also vote to leave it as is. We don't need to have 1:1 mappings between contract names and chain interfaces. Contract names are just implementation details that can change over time. Also +1 for not using the "relay" word as it's a constant source of confusion.

pkg/tbtc/maintainer/chain.go
```diff
@@ -2,12 +2,38 @@ package maintainer
...
+	// Signing returns the signing associated with the chain.
+	Signing() chain.Signing
+
+	// Retarget adds a new epoch to the relay by providing a proof
+	// of the difficulty before and after the retarget.
+	Retarget(headers []*bitcoin.BlockHeader) error
+
+	// CurrentEpoch returns the number of the latest epoch whose difficulty is
+	// proven to the relay.
+	CurrentEpoch() (uint64, error)
```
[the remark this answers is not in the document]
tomaszslabon: Done in 6d46551d5cd3aa2cc29e87acda2fea91fc8aa6c6.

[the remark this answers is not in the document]
tomaszslabon: Added explanation in a5d7fe168fb184b9d788e07cb06c29286f866e2b.

pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -2,54 +2,271 @@ package maintainer
...
+			firstBlockHeaderHeight,
+			lastBlockHeaderHeight,
+			newEpoch,
+		)
+	} else {
+		logger.Infof(
+			"The Bitcoin difficulty chain is up-to-date with the " +
+				"Bitcoin blockchain",
+		)
+	}
```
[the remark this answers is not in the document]
tomaszslabon: Good idea. Done in 0dec4c569080bcfbdac5c5ea9f9fee1fafd8049e.

pkg/tbtc/maintainer/bitcoin_difficulty.go
tomaszslabon: Done in ef7f7d86186a2d8e2b7b5cb5bd84d27527abecd9.

pkg/tbtc/maintainer/local_bitcoin_chain_test.go
tomaszslabon: Done in 70604b7878805b61aa7b60325d603d656c3799c8.

pkg/tbtc/maintainer/local_bitcoin_difficulty_chain_test.go
tomaszslabon: Done in 70604b7878805b61aa7b60325d603d656c3799c8.

pkg/tbtc/maintainer/chain.go
pdyraga: I was thinking of `pkg/tbtc/maintainer` as an extension of `tbtc` node: some additional responsibilities added to a regular node and all these responsibilities being specific to `tbtc`. We could argue SPV maintainer is not tbtc-related but wallet lifecycle maintainer is definitely tbtc-related.

I think `pkg/sortition` is a different kind of an animal - this is a utility code used by `pkg/tbtc` and `pkg/beacon` nodes.

If you guys think it makes more sense to have `pkg/maintainer` and this will save us a headache, let's go this way.

lukasz-zimnoch: > If you guys think it makes more sense to have `pkg/maintainer` and this will save us a headache, let's go this way

I think so. This structure will allow us to have independent packages that live their own lives. This will reflect that those two packages operate in different contexts of the tbtc domain (node and maintainer). This structure is also more flexible if we need to add non-tbtc maintainers in the future. Having `pkg/tbtc/maintainer` will tempt us to introduce dependencies on the `pkg/tbtc` package to reuse some code from it. However, as mentioned, the `pkg/tbtc` code is tailored for the tbtc node domain context and we can have a hard way to satisfy the maintainer context as well. That would lead to increased complexity and worse code.

tomaszslabon: Moved the maintainer package to `pkg/maintainer` in 1674a6d736b31644dee86192ee653b442f89dd27.

pkg/tbtc/maintainer/bitcoin_difficulty.go
```diff
@@ -2,54 +2,271 @@ package maintainer
...
 		}
+	}
+}
 
-		time.Sleep(1 * time.Second)
+// proveEpochs proves Bitcoin blockchain epochs in the Bitcoin difficulty chain.
+func (bdm *BitcoinDifficultyMaintainer) proveEpochs(ctx context.Context) error {
+	if err := bdm.verifySubmissionEligibility(); err != nil {
+		return fmt.Errorf(
+			"cannot proceed with proving Bitcoin blockchain epochs [%v]",
```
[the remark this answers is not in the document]
tomaszslabon: Done for all errors in 49bb6d67a6b9f7240e2d82f5a634f97b059be3dd.

pkg/tbtc/maintainer/bitcoin_difficulty.go
tomaszslabon: Done in ef7f7d86186a2d8e2b7b5cb5bd84d27527abecd9.

pkg/tbtc/maintainer/bitcoin_difficulty.go
tomaszslabon: Waiting for `proofLength` blocks after the new epoch's start seems to be the simplest idea.

pkg/tbtc/maintainer/bitcoin_difficulty.go
tomaszslabon: I modified it a little bit in 42f53097e9843d1b5f3ab61fc7e34e70b1c01ae7.
Now we only use `epochProvenBackOffTime` after Bitcoin difficulty maintainer couldn't prove next epoch and we should wait.
I also renamed it to `idleBackOffTime`.

pkg/maintainer/maintainer.go
```diff
@@ -17,7 +17,13 @@ func Initialize(
 	launchAll := !config.BitcoinDifficulty
 
 	if config.BitcoinDifficulty || launchAll {
-		newBitcoinDifficultyMaintainer(ctx, btcChain, chain)
+		initializeBitcoinDifficultyMaintainer(
+			ctx,
+			btcChain,
+			chain,
```
lukasz-zimnoch, line 23: Non-blocking for this PR but we should think about the way of passing multiple host chain interfaces to `Initialize` function. `Initialize` function runs all or specific maintainers. Each maintainer will probably have its own host chain interface, e.g: the difficulty maintainer has `BitcoinDifficultyChain`, the wallet maintainer will have `WalletChain`, and so on.

We can approach that problem twofold:

**Option 1** 
We can introduce the following interface in the `chain.go` file
```
package maintainer

type Chain interface {
    BitcoinDifficultyChain
    WalletChain
}
```
Then, we can pass it to the `Initialize` function.

**Option 2**
We pass specific maintainer chain interfaces to `Initialize` function:
```
func Initialize(
	ctx context.Context,
	config Config,
	btcChain bitcoin.Chain,
	difficultyChain BitcoinDifficultyChain,
        walletChain WalletChain,
)
```

IMO, option 1 seems better at the moment. However, we don't have to decide right now and think about it during the development of the wallet maintainer (https://github.com/keep-network/keep-core/pull/3446).

pkg/maintainer/bitcoin_difficulty.go
```diff
@@ -0,0 +1,347 @@
...
+	"github.com/ipfs/go-log"
+	"github.com/keep-network/keep-core/pkg/bitcoin"
+)
+
+const (
+	// Default value for back-off time which should be applied when the Bitcoin
+	// difficulty maintainer is restarted. It helps to avoid being flooded with
+	// error logs in case of a permanent error in the Bitcoin difficulty
+	// maintainer.
+	defaultRestartBackoffTime = 120 * time.Second
```
lukasz-zimnoch, line 17: This is specific to the BTC diff maintainer so maybe:
```suggestion
	bitcoinDifficultyDefaultRestartBackoffTime = 120 * time.Second
```

pkg/maintainer/bitcoin_difficulty.go
```diff
@@ -0,0 +1,347 @@
...
+
+	// Default value for back-off time which should be applied when there are no
+	// more Bitcoin epochs to be proven because the difficulty maintainer is
+	// up-to-date with the Bitcoin blockchain or there are not enough blocks yet
+	// to prove the new epoch.
+	defaultIdleBackOffTime = 60 * time.Second
```
lukasz-zimnoch, line 23:
```suggestion
	bitcoinDifficultyDefaultIdleBackOffTime = 60 * time.Second
```

pkg/maintainer/bitcoin_difficulty.go
```diff
@@ -0,0 +1,347 @@
...
+
+	// The number of blocks in a Bitcoin difficulty epoch.
+	bitcoinDifficultyEpochLength = 2016
+)
+
+var logger = log.Logger("keep-maintainer-bitcoin-difficulty")
```
lukasz-zimnoch, line 29: This log name is to long and won't we displayed properly. Maybe just `maintainer-btcdiff`?

pkg/maintainer/bitcoin_difficulty.go
```diff
@@ -0,0 +1,347 @@
...
+		idleBackOffTime:    idleBackOffTime,
+		restartBackOffTime: restartBackOffTime,
+	}
+
+	go bitcoinDifficultyMaintainer.startControlLoop(ctx)
+}
+
+// BitcoinDifficultyMaintainer is the part of maintainer responsible for
+// maintaining the state of the Bitcoin difficulty on-chain contract.
+type BitcoinDifficultyMaintainer struct {
```
lukasz-zimnoch, line 60: No need to expose it outside of the package:
```suggestion
type bitcoinDifficultyMaintainer struct {
```

pkg/maintainer/bitcoin_difficulty.go
```diff
@@ -0,0 +1,347 @@
...
+	}()
+
+	for {
+		err := bdm.proveEpochs(ctx)
+		if err != nil {
+			logger.Errorf(
+				"restarting Bitcoin difficulty maintainer due to error while "+
+					"proving Bitcoin blockchain epochs: [%v]",
+				err,
+			)
```
lukasz-zimnoch, line 84: Let's make it more concise: `error while proving Bitcoin epochs: [%v]; restarting maintainer`

pkg/maintainer/bitcoin_difficulty.go
```diff
@@ -0,0 +1,347 @@
...
+			return
+		}
+	}
+}
+
+// proveEpochs proves Bitcoin blockchain epochs in the Bitcoin difficulty chain.
+func (bdm *BitcoinDifficultyMaintainer) proveEpochs(ctx context.Context) error {
+	if err := bdm.verifySubmissionEligibility(); err != nil {
+		return fmt.Errorf(
+			"cannot proceed with proving Bitcoin blockchain epochs: [%w]",
```
lukasz-zimnoch, line 99:
```suggestion
			"cannot verify submission eligibility: [%w]",
```

pkg/maintainer/bitcoin_difficulty.go
```diff
@@ -0,0 +1,347 @@
...
+			return false, fmt.Errorf(
+				"error while waiting for current Bitcoin difficulty epoch "+
+					"update: [%w]",
+				err,
+			)
+		}
+
+		logger.Infof(
+			"successfully submitted block headers [%d:%d] to the Bitcoin "+
+				"difficulty chain. The current proven epoch is %d.",
```
lukasz-zimnoch, line 265:
```suggestion
				"difficulty chain; the current proven epoch is [%d]",
```

pkg/maintainer/maintainer.go
tomaszslabon: I agree, option 1 seems to be better. I'll implement it in #3446.

pkg/maintainer/bitcoin_difficulty.go
tomaszslabon: Done in f360163494a01b05f91c07f8ba4cafafb1e9951d.

pkg/maintainer/bitcoin_difficulty.go
tomaszslabon: Done in f360163494a01b05f91c07f8ba4cafafb1e9951d.

pkg/maintainer/bitcoin_difficulty.go
tomaszslabon: Done in 11fa7e5b52750fec4a700f8a863e8d0fdd83dfbe.

pkg/maintainer/bitcoin_difficulty.go
tomaszslabon: Done in b75c0745044ba7a0859d19c38f2b01ccdc1858df.

pkg/maintainer/bitcoin_difficulty.go
tomaszslabon: Done in 1019eaa16ab83309fbd0f34124f250d91106dd7d.

pkg/maintainer/bitcoin_difficulty.go
tomaszslabon: Done in 5fa25eb824348e461e52ca1279373fe17a376f12.

pkg/maintainer/bitcoin_difficulty.go
tomaszslabon: Done in 5fa25eb824348e461e52ca1279373fe17a376f12.

pkg/maintainer/maintainer.go
lukasz-zimnoch: Leaving it open for future reference.

pkg/tbtc/maintainer/bitcoin_difficulty.go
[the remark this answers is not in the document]
tomaszslabon: Let's handle this in a follow-up PR.
Left a TODO in 176e7b40674d17eb216fe4cb73dd08220de9df87.

lukasz-zimnoch (approved):

pkg/tbtc/maintainer/chain.go
lukasz-zimnoch: Leaving it unresolved for future reference.

PR #3384, opened 2022-10-26
Introduce the `pkg/bitcoin` package

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3349

Here we introduce the new `pkg/bitcoin` package that defines some basic Bitcoin types and interfaces required to work with the Bitcoin chain. This package is meant to reflect the part of the Bitcoin protocol to an extent required by the client applications. The `pkg/bitcoin` package is supposed to hold components specific to the Bitcoin domain and must remain free of application-specific items. Third-party helper libraries (like `btcd`) are allowed for use within this package though no external type should leak outside. Preserving those rules will allow us to achieve a clean package structure with proper dependencies. The exact types introduced by this pull request are:
- `bitcoin.ByteOrder`: An enum that captures all the caveats related to the byte order used across the Bitcoin ecosystem
- `bitcoin.CompactSizeUint`: A documentation type that holds the details about the CompactSize Unsigned Integer
- `bitcoin.Hash`: A type that represents a 32-byte Bitcoin hash widely used as identifier in the Bitcoin protocol
- `bitcoin.Transaction` & related: A type that represents a Bitcoin transaction and related types to reflect the internals
- `bitcoin.BlockHeader`: A representation of the Bitcoin block header that will be used for SPV purposes
- `bitcoin.Chain`: A minimal interface required to interact with the Bitcoin chain
- `electrum.Client`: An example stub implementation of the `bitcoin.Chain` supposed to establish a package structure for Bitcoin interfaces implementations

Issue #3369, opened 2022-10-24
Electrum server integration


pdyraga:
The Keep client needs to be able to communicate with the Bitcoin chain.

We probably want to add support for two servers that have compatible API: ElectrumX and Blockstream's Electrs. This will give an advantage of failover and a publicly available API for small stakers. The required API will be specified during the work on #3349.

nkuba:
Closed by https://github.com/keep-network/keep-core/pull/3406

PR #3395, opened 2022-11-02
Electrs integration for Bitcoin chain

nkuba:
This PR implements Bitcoin package integration with Electrs API.

TODO:
- [ ] command flags for config properties
- [ ] tests
- [ ] PR description

Refs https://github.com/keep-network/keep-core/pull/3384
Refs https://github.com/keep-network/keep-core/issues/3369

### DIFF
--- a/pkg/bitcoin/chain.go
+++ b/pkg/bitcoin/chain.go
@@ -20,12 +20,12 @@ type Chain interface {
 	// rejected eventually.
 	BroadcastTransaction(transaction *Transaction) error
 
-	// GetLatestBlockNumber gets the height of the latest block (tip). If the
+	// GetLatestBlockHeight gets the height of the latest block (tip). If the
 	// latest block was not determined, this function returns an error.
-	GetLatestBlockNumber() (uint, error)
+	GetLatestBlockHeight() (uint, error)
 
-	// GetBlockHeader gets the block header for the given block number. If the
-	// block with the given number was not found on the chain, this function
+	// GetBlockHeader gets the block header for the given block height. If the
+	// block with the given height was not found on the chain, this function
 	// returns an error.
-	GetBlockHeader(blockNumber uint) (*BlockHeader, error)
+	GetBlockHeader(blockHeight uint) (*BlockHeader, error)
 }
--- a/pkg/bitcoin/chain.go
+++ b/pkg/bitcoin/chain.go
@@ -20,9 +20,9 @@ type Chain interface {
 	// rejected eventually.
 	BroadcastTransaction(transaction *Transaction) error
 
-	// GetCurrentBlockNumber gets the number of the current block. If the
-	// current block was not determined, this function returns an error.
-	GetCurrentBlockNumber() (uint, error)
+	// GetLatestBlockNumber gets the height of the latest block (tip). If the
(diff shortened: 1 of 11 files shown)


pkg/bitcoin/chain.go
```diff
@@ -20,12 +20,12 @@ type Chain interface {
 	// rejected eventually.
 	BroadcastTransaction(transaction *Transaction) error
 
-	// GetCurrentBlockNumber gets the number of the current block. If the
-	// current block was not determined, this function returns an error.
-	GetCurrentBlockNumber() (uint, error)
+	// GetLatestBlockHeight gets the height of the latest block (tip). If the
+	// latest block was not determined, this function returns an error.
+	GetLatestBlockHeight() (uint, error)
```
nkuba, line 25: @lukasz-zimnoch I hope you're fine with this rename.

The function expects the "chain tip", so the most recent mined block. I propose we use "latest", as "current" may be confusing and suggest that the block is being currently mined.

Also, I replaced "block number" with "block height", as it seems to be widely used in the Bitcoin ecosystem.

pkg/bitcoin/electrs/block.go
```diff
@@ -0,0 +1,47 @@
+package electrs
+
+import (
+	"fmt"
+
+	"github.com/keep-network/keep-core/pkg/bitcoin"
+)
+
+type blockHeader struct {
```
tomaszslabon, line 9: Let's not forget about descriptions of all the functions and types.

pkg/bitcoin/electrs/block.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (b *blockHeader) convert() (bitcoin.BlockHeader, error) {
+	result := bitcoin.BlockHeader{
+		Version: b.Version,
+		Time:    b.Timestamp,
+		Bits:    b.Bits,
+		Nonce:   b.Nonce,
+	}
+
+	previousBlockHeaderHash, err := bitcoin.NewHashFromString(b.PreviousBlockHash, bitcoin.ReversedByteOrder)
```
tomaszslabon, line 26: Could we split this lengthy line?

pkg/bitcoin/electrs/electrs.go
```diff
@@ -0,0 +1,204 @@
...
+	retryTimeout time.Duration
+}
+
+type httpClient interface {
+	Get(url string) (*http.Response, error)
+	Post(url string, contentType string, body io.Reader) (*http.Response, error)
+}
+
+// Connect initializes handle with provided Config.
+func Connect(config Config) (bitcoin.Chain, error) {
```
tomaszslabon, line 45: `Connect` won't return an error even if the URL provided in the config is wrong.
We will only learn about the wrong URL when we try to use the Bitcoin chain for the first time.
I would consider adding a connection test before we leave `Connect`, we did that such test when connecting to btc client in [the old relay maintainer](https://github.com/keep-network/tbtc/blob/d18ef9aec7656f0ec7d317ece3e3d5c7aca92cda/relay/pkg/btc/remote.go#L45)

nkuba:
Closing in favour of https://github.com/keep-network/keep-core/pull/3406

PR #3396, opened 2022-11-03
Introduce broadcast channel retransmission strategies

lukasz-zimnoch:
Refs: #3366

So far, the broadcast channel used a simple retransmission strategy that triggered a new retransmission on every new tick. Such an aggressive approach turned out to be not suitable for all use cases. Specifically, this approach tends to cause network congestion if a large amount of messages is retransmitted for a long time. To make the retransmission mechanism more flexible, here we introduce the way to customize it through strategies. We are also adding an implementation of a backoff strategy that decreases its retransmission frequency over time in an exponential way.

### DIFF
--- a/pkg/net/net.go
+++ b/pkg/net/net.go
@@ -2,9 +2,21 @@ package net
 
 import (
 	"context"
-
 	"github.com/keep-network/keep-core/pkg/internal/pb"
 	"github.com/keep-network/keep-core/pkg/operator"
+)
+
+// RetransmissionStrategy represents a specific retransmission strategy.
+type RetransmissionStrategy int
+
+const (
+	// StandardRetransmissionStrategy is the default retransmission strategy
+	// that retransmit the message with a constant frequency.
+	StandardRetransmissionStrategy RetransmissionStrategy = iota
+	// BackoffRetransmissionStrategy is a retransmission strategy that
+	// retransmit the message with an exponentially increasing delay between
+	// subsequent retransmissions.
+	BackoffRetransmissionStrategy
 )
 
 // TransportIdentifier represents a protocol-level identifier. It is an opaque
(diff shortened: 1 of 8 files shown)


pkg/net/net.go
```diff
@@ -93,10 +105,18 @@ type TaggedUnmarshaler interface {
...
-	Send(ctx context.Context, m TaggedMarshaler) error
+	// retransmitted by the channel for the lifetime of the provided context
+	// according to the default StandardRetransmissionStrategy. Retransmission
+	// strategy can be set through the `strategy` vararg. If the vararg is
+	// given more than one value, the first value is used as the valid
+	// strategy.
+	Send(
+		ctx context.Context,
+		message TaggedMarshaler,
+		strategy ...RetransmissionStrategy,
```
pdyraga, line 118: I would call this field `retransmissionStrategy`, just `strategy` does not tell what it is about without looking at the type. Is it pub-sub strategy, routing strategy, or what? This comment applies to all files with `strategy ...RetransmissionStrategy` in the function signature.

```suggestion
		retransmissionStrategy ...RetransmissionStrategy,
```

pkg/net/net.go
```diff
@@ -2,11 +2,23 @@ package net
...
 
+// RetransmissionStrategy represents a specific retransmission strategy.
+type RetransmissionStrategy int
+
+const (
+	// StandardRetransmissionStrategy is the default retransmission strategy
+	// that retransmit the message with a constant frequency.
+	StandardRetransmissionStrategy RetransmissionStrategy = iota
+	// BackoffRetransmissionStrategy is a retransmission strategy that
+	// retransmit the message with an exponentially increasing delay between
```
pdyraga, line 17:
```suggestion
	// retransmits the message with an exponentially increasing delay between
```

pdyraga (approved):
Left two small, non-blocking comments. Let's discuss them and potentially address as a part of the work on #3392.

pkg/net/net.go
lukasz-zimnoch: Addressed in https://github.com/keep-network/keep-core/pull/3392/commits/a1a9ec4f552bc11bb7ee10c1436d90076b42fa05

pkg/net/net.go
lukasz-zimnoch: Addressed in https://github.com/keep-network/keep-core/pull/3392/commits/a1a9ec4f552bc11bb7ee10c1436d90076b42fa05

PR #3397, opened 2022-11-03
Parallel signing

pdyraga:
# Background
Sweeping requires producing multiple signatures. One per every input of the sweep transaction. The signing of sweep inputs will start with the announcement phase where each signing group member will announce their readiness to participate in the signing of sweep transaction inputs. 51 signers from the ones who announced their readiness will be selected for the protocol. They will unlock each deposit UTXO with a signature. Signatures will be computed in parallel and no signature should be completed without all others. This approach simplifies defeating fraud proofs as this is an almost all-or-nothing situation for a sweep transaction signing:

```
|                  | ------------ signature 1 ------------> |              |
| ... announce ... | ------------ signature 2 ------------> | protocol end | timeout
|                  | ------------ signature 3 ------------> |              |
```

# Changes
In this PR, the signing protocol is being prepared for sweep implementation. Instead of triggering multiple state machines, we will execute one progressing over signing phases, one by one. Each phase will be updating multiple TSS lib participants, one per every message being signed.

All TSS messages exchanged during signing have been updated to follow a composite pattern. One composite message contains per-signed-message TSS messages from the given signing group member. This approach allows to validate incoming messages only one time, for all parallel signing protocols being executed. This is an important property from the perspective of broadcast channel performance, especially important during computationally expensive phases.


lukasz-zimnoch:
We decided to close this pull request without merging. The parallel signing approach presented in this pull request turned out to be really unstable. We were not able to have a successful signing execution for a batch of 10 messages. The problem with this approach is basically about very long periods of computations with a lot of CPU spikes that can starve all other processes, for example, networking. This makes the clients unresponsive and affects the execution of all protocols which is unacceptable for production use. Moreover, the code changes required by this approach have a negative impact on the overall code quality and significantly increase the complexity level. That being said, we are planning to go with an alternative approach for batch signing.

PR #3399, opened 2022-11-08
Cleanup of tBTCv2 rewards calculation code

dimpar:
The rewards calculation for tBTCv2 was moved to
https://github.com/threshold-network/merkle-distribution repo and become a part of all Threshold rewards calculation, ie PRE bonus, PRE ongoing, and tBTCv2. All the reviews and new PRs around Threshold rewards should be opened there, so we do not duplicate code.


beaushinkle:
:fire: :fire: :fire: :fire: :fire:

beaushinkle (approved):

Issue #3367, opened 2022-10-24
Stub Random Beacon for Chaosnet


pdyraga:
For the initial Chaosnet phases, Random Beacon is not really needed and we can focus on other work. We can stub the random beacon contract and have the stub answer relay requests by hitting the `WalletRegistry`'s callback with an arbitrary number to trigger DKG. We can tie that random beacon contract to the `WalletRegistry` temporarily using `upgradeRandomBeacon` function until we have the real-world random beacon client implementation ready.

PR #3402, opened 2022-11-10
Adding RandomBeaconChaosnet contract

dimpar:
Refs https://github.com/keep-network/keep-core/issues/3367

This is a stub contract that will replace the original RandomBeacon contract until we have off-chain clients ready. It will be used only for the initial choasnet phases. The request of a relay entry will set the callback contract and also execute a callback function with an arbitrary number instead of a "random beacon".

Wallet registry consumer test is here: https://github.com/keep-network/keep-core/pull/3403


solidity/random-beacon/contracts/RandomBeaconChaosnet.sol
```diff
@@ -0,0 +1,71 @@
...
+        external
+    {
+        require(
+            authorizedRequesters[msg.sender],
+            "Requester must be authorized"
+        );
+
+        callback.setCallbackContract(callbackContract);
+
+        callback.executeCallback(entry, _callbackGasLimit);
```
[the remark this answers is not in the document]
tomaszslabon: In 882a3984f42c8d374b41bf6775dec8d73415ed23 I added a `seed` number that is incremented on every `requestRelayEntry` call. This `seed` is hashed using `keccak-256` converted to `uint256` and used as the relay entry.

Is that ok?

solidity/random-beacon/contracts/RandomBeaconChaosnet.sol
```diff
@@ -0,0 +1,80 @@
...
+
+        uint256 relayEntry = uint256(keccak256(abi.encodePacked(seed)));
+
+        // Update the seed number so that a different relay entry is produced
+        // every time this function is called. Using the same seed would result
+        // in the same group of wallet operators being selected in
+        // `WalletRegistry`.
+        seed++;
+
+        callback.executeCallback(relayEntry, _callbackGasLimit);
```
lukasz-zimnoch, line 68: I think we don't need to increment the seed but just have an `entry` field and update it with the hashed value:
```
entry = uint256(keccak256(abi.encodePacked(entry)));
callback.executeCallback(entry, _callbackGasLimit);
```
This way the callback is provided with a different value on each `requestRelayEntry` call

solidity/random-beacon/contracts/RandomBeaconChaosnet.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3402#discussion_r1056218405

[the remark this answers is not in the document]
tomaszslabon: Done in bfc06da919be7fce0c1640a05727739c353a2967.

solidity/random-beacon/contracts/RandomBeaconChaosnet.sol
tomaszslabon: Done in bfc06da919be7fce0c1640a05727739c353a2967.

lukasz-zimnoch (approved):

solidity/random-beacon/deploy/09_deploy_random_beacon_chaosnet.ts
```diff
@@ -0,0 +1,35 @@
...
+    await hre.tenderly.verify({
+      name: "RandomBeaconChaosnet",
+      address: RandomBeaconChaosnet.address,
+    })
+  }
+}
+
+export default func
+
+func.tags = ["RandomBeaconChaosnet"]
```
pdyraga, line 35: This step is deploying `RandomBeaconChaosnet` but it is not switching `WalletRegistry` to use it.

If we deployed contracts with the scripts as-is to testnet, would it be possible to use `RandomBeaconChaosnet` at all?

`RandomBeacon` and `WalletRegistry` are already deployed to mainnet so this has to be done manually by Threshold Council but we should have it done automatically for testnet deployment.

solidity/random-beacon/contracts/RandomBeaconChaosnet.sol
```diff
@@ -0,0 +1,76 @@
...
+            "Requester must be authorized"
+        );
+
+        callback.setCallbackContract(callbackContract);
+
+        // Update the entry so that a different group of wallet operators is
+        // selected in `WalletRegistry` on each request.
+        entry = uint256(keccak256(abi.encodePacked(entry)));
+
+        callback.executeCallback(entry, _callbackGasLimit);
```
pdyraga, line 64: Since we are responding immediately to the request, does it make sense to use the `Callback` library? The purpose of the `Callback` library was not to fail to submit the relay entry - and slash operators for relay entry timeout - if the callback reverts. This is not the case here because the relay entry is known at the moment `requestRelayEntry` is called.

There are two complications regarding the use of `Callback` library:
- We need to set the gas limit. Is the current gas limit enough for now? What if the pricing of opcodes changes in the future? We do not have the mechanism to update the gas limit.
- There is a gas cost hidden behind `callback.setCallbackContract` and `callback.executeCallback`.

I think we should call `IRandomBeaconConsumer` right away, without a gas limit, and without a try-catch.

solidity/random-beacon/contracts/RandomBeaconChaosnet.sol
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /// @notice Authorizes a requester of the relay entry.
+    function setRequesterAuthorization(address requester, bool isAuthorized)
+        external
+        onlyOwner
```
pdyraga, line 70: How many requesters do we expect? I think there is only one - `WalletRegistry` (proxy) contract.

solidity/random-beacon/deploy/09_deploy_random_beacon_chaosnet.ts
```diff
@@ -0,0 +1,35 @@
...
+  const { getNamedAccounts, deployments, helpers } = hre
+  const { deployer } = await getNamedAccounts()
+
+  const deployOptions: DeployOptions = {
+    from: deployer,
+    log: true,
+    waitConfirmations: 1,
+  }
+
+  const RandomBeaconChaosnet = await deployments.deploy(
```
nkuba, line 14: We could integrate `RandomBeaconChaosnet` contract deployment with the `RandomBeacon` contract deployment https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/deploy/04_deploy_random_beacon.ts

If we run the deployment right now we will deploy both `RandomBeacon` `RandomBeaconChaosnet`. We should determine a contract to be deployed based on the network we're running the script against. See example here: https://github.com/keep-network/tbtc-v2/blob/6ff3d7b1c7d0a421507037acf3a421c4261dfcdd/solidity/deploy/00_resolve_relay.ts#L21-L26

solidity/random-beacon/contracts/RandomBeaconChaosnet.sol
tomaszslabon: I addressed this issue in the [follow-up PR](https://github.com/keep-network/keep-core/pull/3459).

solidity/random-beacon/contracts/RandomBeaconChaosnet.sol
tomaszslabon: I addressed this issue in the https://github.com/keep-network/keep-core/pull/3459.

solidity/random-beacon/deploy/09_deploy_random_beacon_chaosnet.ts
tomaszslabon: We decided to introduce a flag (`chaosnet`) that decides whether we should deploy something or not.
See: https://github.com/keep-network/keep-core/pull/3470 and https://github.com/keep-network/keep-core/pull/3460.

solidity/random-beacon/deploy/09_deploy_random_beacon_chaosnet.ts
tomaszslabon: We decided to handle that in the follow-up PR: https://github.com/keep-network/keep-core/pull/3460.

PR #3405, opened 2022-11-14
Bitcoin difficulty maintainer: Added LightRelay handle

tomaszslabon:
This PR adds functionalities for accessing the `LightRelay` smart contract:
- added `BitcoinDifficultyChain` which allows for calling `LightRelay` contract functions
- additionally, code generation for `LightRelay` had to be added.


pdyraga:
@tomaszslabon Can you please rebase this PR now when #3378 is merged?

tomaszslabon:
> @tomaszslabon Can you please rebase this PR now when #3378 is merged?

Yes. I already did.

pkg/chain/ethereum/tbtc.go
```diff
@@ -103,11 +108,38 @@ func newTbtcChain(
 		)
 	}
 
+	lightRelayAddress, err := config.ContractAddress(LightRelayContractName)
```
tomaszslabon, line 111: This `LightRelay` contract address extraction requires that we have `LightRelayAddress` defined in our configs when starting `keep-client`.

pdyraga: I do not think this is an acceptable solution. It is possible to start the client on mainnet without a config file, [just by providing the required parameters in CLI](https://docs.threshold.network/staking-and-running-a-node/running-a-node/self-managed/tbtc-v2-node-setup/binary-installation#configure-tbtc-v2-client) and even if there is a config file, `LightRelay` is not there.

Since we are not pulling `tbtc-v2` contract addresses from mainnet NPM package (see #3408), all clients running on mainnet would fail to start after the upgrade, no matter if they use config file or just CLI parameters.

One option would be to move the difficulty-related functions to a separate chain structure but I do not think it is worth the complications, especially that this functionality is tightly-coupled with tBTC. 

A quick fix could be to:
1. Do not fail here if the address is not configured, just don't create the light relay contract.
2. For every light relay function, check if the contract was instantiated. If not, report an error.

In the code reading the config, we'd need to check whether the bitcoin difficulty maintainer is enabled and if it is enabled, then whether the ethereum address for the `LightRelay` contract is configured.

pdyraga: I think we could validate it [here](https://github.com/keep-network/keep-core/blob/fc2b0d36d5b09fd770c0c4769acc31839aa02ce1/config/config.go#L154-L187). Currently, we do not do a cross-section validations, but I think doing so is perfectly acceptable.

tomaszslabon: One thing to note is that we could set the `LightRelay` address via command line option:
`--developer.lightRelayAddress 0x0000000000000000000000000000000000000000`

pdyraga: It is true and I think this would be acceptable for development environments but this is a breaking change for production configs and something that would be temporary - we will resolve the addresses from the package at some point.

pkg/tbtc/chain.go
```diff
@@ -125,6 +152,7 @@ type Chain interface {
 	GroupSelectionChain
 	DistributedKeyGenerationChain
 	BridgeChain
+	RelayChain
```
pdyraga, line 155: I would consider organizing it as:
```
type MaintainerChain interface {
  BitcoinDifficultyMaintainer() (chain.BitcoinDifficultyMaintainer, error)
}

type BitcoinDifficultyMaintainer interface {
  Ready (bool, error)
  AuthorizationRequired() (bool, error)
  // (...)
}
```

This way, we can add more types of maintainers to the `MaintainerChain` later. SPV maintainer, group lifecycles maintainer etc.

pkg/chain/ethereum/tbtc.go
tomaszslabon: I moved the Bitcoin difficulty-related functionalities to a separate chain in 631768e43339d9a2481810fbcfb4614326d142c2.

Therefore there is no longer need to set any value for `LightRelayAddress` when starting the keep-client for keep nodes running.

pkg/tbtc/chain.go
tomaszslabon: I added the `MaintainerChain` to the `tbtc.Chain` interface and therefore forced the implementation of `TbtcChain` to have the `BitcoinDifficultyMaintainer()` function.

Then I realise it did not make much sense: we would create the `BitcoinDifficultyMaintainer` and pass it inside the `TbtcChain`, just to return it when we need to use it.
`TbtcChain` would sometimes store that `BitcoinDifficultyMaintainer` (when running `keep-client` as a maintainer) and sometimes it wouldn't (when we use `keep-client` for running keep nodes).

Therefore I treated the `BitcoinDifficultyMaintainer` a separate chain, just like we treat the `TbtcChain` or `BeaconChain`. They all are just handles to the smart contracts.

I put the interface in the `pkg/bitcoin_difficulty/chain.go`, so that it's symmetrical to location of Tbtc interface (`pkg/tbtc/chain.go`).

I put the implementation in the `pkg/chain/ethereum/bitcoin_difficulty`, so that it's symmetrical to the location of Tbtc implementation (`pkg/ethereum/tbtc.go`).

Also I named it `bitcoin_difficulty` for short.

pkg/bitcoin_difficulty/chain.go
```diff
@@ -0,0 +1,34 @@
+package bitcoin_difficulty
```
pdyraga, line 1: From https://go.dev/blog/package-names:

> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns (...)
>
> The style of names typical of another language might not be idiomatic in a Go program. Here are two examples of names that might be good style in other languages but do not fit well in Go:
> - computeServiceClient
> - priority_queue

That being said, what do you think about naming this package `btcdiff`? (`btcdifficulty` feels too long/complicated in my opinion).

pkg/bitcoin_difficulty/chain.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"github.com/keep-network/keep-core/pkg/bitcoin"
+	"github.com/keep-network/keep-core/pkg/chain"
+)
+
+// Chain defines the interface that pertains specifically to the process of
+// updating Bitcoin difficulty.
+type Chain interface {
+	// Ready checks whether the relay is active (i.e. genesis has been
+	// performed).
```
pdyraga, line 12: Please keep in mind [this limitation](https://github.com/keep-network/tbtc-v2/blob/83310bdc9ed934e286bc9ea5091cc16979950134/solidity/contracts/relay/LightRelay.sol#L130-L132): 

```
    /// @dev If the relay is used by querying the current and previous epoch
    /// difficulty, at least one retarget needs to be provided after genesis;
    /// otherwise the prevEpochDifficulty will be uninitialised and zero.
```

We should make it clear in the docs for `Ready`. `Ready` returning `true` does not mean `getPrevEpochDifficulty` will give a meaningful result.

pkg/bitcoin_difficulty/chain.go
```diff
@@ -0,0 +1,34 @@
...
+	Ready() (bool, error)
+
+	// AuthorizationRequired checks whether the relay requires the address
+	// submitting a retarget to be authorised in advance by governance.
+	AuthorizationRequired() (bool, error)
+
+	// IsAuthorized checks whether the given address has been authorised to
+	// submit a retarget by governance.
```
pdyraga, line 20:
```suggestion
	// IsAuthorized checks whether the given address has been authorised by
	// governance to submit a retarget.
```

pkg/chain/ethereum/bitcoin_difficulty.go
```diff
@@ -0,0 +1,102 @@
...
+			baseChain.key,
+			baseChain.client,
+			baseChain.nonceManager,
+			baseChain.miningWaiter,
+			baseChain.blockCounter,
+			baseChain.transactionMutex,
+		)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"failed to attach to EcdsaSortitionPool contract: [%v]",
```
pdyraga, line 53:
```suggestion
			"failed to attach to LightRelay contract: [%v]",
```

pkg/chain/ethereum/bitcoin_difficulty.go
```diff
@@ -0,0 +1,102 @@
...
+			err,
+		)
+	}
+
+	return &BitcoinDifficultyChain{
+		lightRelay: lightRelay,
+	}, nil
+}
+
+// Ready checks whether the relay is active (i.e. genesis has been performed).
```
pdyraga, line 63: Same comment about the `getPrevEpochDifficulty` limitation clarification in docs as for the interface.

pkg/chain/ethereum/tbtc.go
pdyraga: Nice, that makes sense!

pkg/bitcoin_difficulty/chain.go
tomaszslabon: Makes sense. Done in 3ec575c469ee0d1dc37503cd59cae39660cbfa11.

pkg/bitcoin_difficulty/chain.go
tomaszslabon: Done in 37426409216d716c85a610028748547a963172d7.

pkg/chain/ethereum/bitcoin_difficulty.go
tomaszslabon: Done in c3bac43be0723d973ae2d095979f3e105945fb35.

pkg/bitcoin_difficulty/chain.go
dimpar: can we also change `pkg/bitcoin_difficulty` -> `pkg/btcdiff`?

pkg/chain/ethereum/bitcoin_difficulty.go
```diff
@@ -0,0 +1,107 @@
...
+// BitcoinDifficultyChain represents a Bitcoin difficulty-specific chain handle.
+type BitcoinDifficultyChain struct {
+	*baseChain
+
+	lightRelay *contract.LightRelay
+}
+
+// NewBitcoinDifficultyChain construct a new instance of the Bitcoin difficulty
+// - specific Ethereum chain handle.
+func NewBitcoinDifficultyChain(
```
dimpar, line 28: Does it have to be exported?

pkg/bitcoin_difficulty/chain.go
```diff
@@ -0,0 +1,38 @@
+package btcdiff
+
+import (
+	"github.com/keep-network/keep-core/pkg/bitcoin"
+	"github.com/keep-network/keep-core/pkg/chain"
+)
+
+// Chain defines the interface that pertains specifically to the process of
+// updating Bitcoin difficulty.
+type Chain interface {
```
dimpar, line 10: How/where can we check what should be included "specifically to the process of updating Bitcoin difficulty"?

dimpar:
no more comments

pkg/bitcoin_difficulty/chain.go
tomaszslabon: Sure, changed package name to `btcdiff` in 3ff897098673c429bc9696f78cfcbde217d814cb.

pkg/bitcoin_difficulty/chain.go
tomaszslabon: Bitcoin difficulty updating means updating the [LightRelay](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/relay/LightRelay.sol#L83) contract.

The most important function there is `Retarget` which does the actual updating, but there are also some getters that we must use before we can update the `LightRelay`.
Here is [the PR which adds the updating](https://github.com/keep-network/keep-core/pull/3394).

pkg/chain/ethereum/bitcoin_difficulty.go
tomaszslabon: It needs to be exported, because it will be used outside this package.
I realised this function can already be called and added the call to it in 6d2f2574c59e0e72ab16c4497a590ef980487f36.

pkg/btcdiff/chain.go
```diff
@@ -0,0 +1,41 @@
+package btcdiff
```
lukasz-zimnoch, line 1: What's the point of the `pkg/btcdiff` package? Can't we just have the `BitcoinDifficultyChain` interface right in the `pkg/maintainer`?

tomaszslabon: The idea is that we treat the `BitcoinDifficultyChain` as a separate chain, just like `BeaconChain` and `TbtcChain`.
Since the two have their interfaces defined in `pkg/beacon` and `pkg/tbtc` it seemed reasonable to define one for `BitcoinDifficultyChain` (`pkg/btcdiff`) as well. 

But since it is indeed not used anywhere, I deleted it the whole `btcdiff` directory in 6bd38d7bdc5e20bd5adda825fd08df472f14d4c3.

pdyraga:
Note that based on [this discussion](https://github.com/keep-network/keep-core/pull/3394#discussion_r1038137159) some components added in this PR should be located in `pkg/maintainer` and not `pkg/tbtc/maintainer`. I am merging it as-is it do not create merge conflicts later, but please take it into account when working on #3394.

PR #3410, opened 2022-11-17
Fixing client docker image

dimpar:
The current version of Dockerfile is split into a couple of layers. The
top layers are responsible for Go code generators. However, part of this
generation includes the extraction of addresses from three contracts:
TokenStaking, RandomBeacon, and WalletRegistry. Addresses are written in
the files that the client will later use during the startup.
But the following `COPY` command that copies the root dir to a Docker working dir
overwrites these 3 files with empty content. To mitigate the issue, we
can exclude these 3 files from the Docker context and they won't
overwrite addresses as part of `COPY ./ $APP_DIR/` command.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -35,3 +35,7 @@ token-tracker/
 # Legacy V1 contracts bindings.
 # We won't generate new bindings in the docker build process, but use the existing ones.
 !pkg/chain/gen
+
+pkg/chain/ethereum/threshold/gen/_address/TokenStaking
+pkg/chain/ethereum/beacon/gen/_address/RandomBeacon
+pkg/chain/ethereum/ecdsa/gen/_address/WalletRegistry
--- a/.dockerignore
+++ b/.dockerignore
@@ -36,6 +36,7 @@ token-tracker/
 # We won't generate new bindings in the docker build process, but use the existing ones.
 !pkg/chain/gen
 
-pkg/chain/ethereum/threshold/gen/_address/TokenStaking
-pkg/chain/ethereum/beacon/gen/_address/RandomBeacon
-pkg/chain/ethereum/ecdsa/gen/_address/WalletRegistry
+# Contract addresses.
+# We do not want to overwrite extracted contract addresses with empty content while
+# coping local root directory to docker image context.
+**/gen/_address/*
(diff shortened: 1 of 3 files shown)


.dockerignore
```diff
@@ -35,3 +35,7 @@ token-tracker/
 # Legacy V1 contracts bindings.
 # We won't generate new bindings in the docker build process, but use the existing ones.
 !pkg/chain/gen
+
+pkg/chain/ethereum/threshold/gen/_address/TokenStaking
+pkg/chain/ethereum/beacon/gen/_address/RandomBeacon
+pkg/chain/ethereum/ecdsa/gen/_address/WalletRegistry
```
nkuba, line 41: Could the rule be more general?

E.g.
```suggestion
**/gen/_address
```
Would this work fine?

.dockerignore
```diff
@@ -35,3 +35,7 @@ token-tracker/
 # Legacy V1 contracts bindings.
 # We won't generate new bindings in the docker build process, but use the existing ones.
 !pkg/chain/gen
+
```
nkuba, line 38: Can we add a comment to clarify this section?

```suggestion

# Contracts addresses.
```

.dockerignore
dimpar: sure, `**/gen/_address` won't work, but `**/gen/_address/*` will. [2e24cca](https://github.com/keep-network/keep-core/pull/3410/commits/2e24cca713dc3d508669376918c4bda602859bf0)

.dockerignore
dimpar: [2e24cca](https://github.com/keep-network/keep-core/pull/3410/commits/2e24cca713dc3d508669376918c4bda602859bf0)

michalinacienciala (approved):
I tested the changes in https://github.com/keep-network/keep-core/actions/runs/3497020932, which published the image https://console.cloud.google.com/gcr/images/keep-test-f3e0/global/keep-client-wip@sha256:1f675124639a6acccfc22bd0ac52c6d90700124d278ebb1c8ad05c989006d859. The image can be successfully used to run the client, no errors about missing contracts addresses are reported.

nkuba (approved):

Issue #3349, opened 2022-10-06
Go code for assembling Bitcoin sweep transaction


lukasz-zimnoch:
Implement Go code for assembling Bitcoin sweep transactions. Expose interface for Bitcoin client.

PR #3421, opened 2022-11-28
Deposit sweep: Assemble Bitcoin transactions

lukasz-zimnoch:
Closes: #3349

Here we introduce the Go code for assembling deposit sweep transactions. Specific changes can be divided into three categories:
- Generic Bitcoin-level transaction assembling code in the `pkg/bitcoin` package
- Domain-specific sweep assembling code in the `pkg/tbtc` package
- Sweep testing machinery in the `pkg/internal/tbtctest` package

Here is a description of concrete changes in each category.

### Generic transaction assembling code in the `pkg/bitcoin` package

This pull request introduces the `bitcoin.TransactionBuilder` component that encapsulates the Bitcoin transaction assembling logic. It is designed to handle different types of Bitcoin transactions that will be crafted as part of the tBTC application so it does not contain anything specific to sweep transactions. In order to not reinvent the wheel, the `TransactionBuilder` component makes heavy use of the `github.com/btcsuite/btcd` library to do the heavy lifting. According to the `pkg/bitcoin` philosophy, the `github.com/btcsuite/btcd` library is used internally and no lib-specific types leak outside of the `pkg/bitcoin` thus do not pollute the client code. The `TransactionBuilder` component exposes the following methods reflecting the process of assembling a full-fledged Bitcoin transaction:
- `AddPublicKeyHashInput` that constructs an unsigned input pointing to a UTXO locked using the P2(W)PKH script.
- `AddScriptHashInput` that creates an unsigned input pointing to a UTXO locked using the P2(W)SH script. 
- `AddOutput` that adds an output of the given value locked using an arbitrary locking script
- `ComputeSignatureHashes` that computes a slice of sighashes corresponding to the transaction's input vector. The outcome of this function can serve as an input for the tECDSA serial signing protocol
- `AddSignatures` that adds the signatures slice produced outside to the transaction's input vector and returns a signed `bitcoin.Transaction` instance that can be spread over the Bitcoin network. This method can accept the output of the tECDSA serial signing protocol

Additionally, this pull request introduces some tooling required to build the basic Bitcoin scripts that can be used to construct outputs added through the builder's `AddOutput` method. Like in the `TransactionBuilder` case, this code also uses the `github.com/btcsuite/btcd` library internally to do its work in a simple way. This tooling consists of several methods placed in the `script.go` file and exposed publicly outside of the `pkg/bitcoin` package:
- `PayToWitnessPublicKeyHash` that constructs a P2WPKH script based on the provided 20-byte public key hash
- `PayToPublicKeyHash` that creates a P2PKH script based on the provided 20-byte public key hash
- `PublicKeyHash` that converts an ECDSA public key to a 20-byte public key hash that can be used as an input of the two above functions 
- `PayToWitnessScriptHash` that creates a P2WSH script based on the provided 32-byte redeem script hash
- `WitnessScriptHash` that converts an arbitrary redeem script to a 32-byte redeem script hash that can be used as an input of the above `PayToWitnessScriptHash` function
- `PayToScriptHash` that creates a P2SH script based on the provided 20-byte redeem script hash
- `ScriptHash` that converts an arbitrary redeem script to a 20-byte redeem script hash that can be used as an input of the above `PayToScriptHash` function

Although some of the above functions are unused right now, they were added to establish a pattern as the `pkg/bitcoin` is a kind of library. Also, there is a high chance that unused methods will be used by production or test code in the near future.

The last thing introduced by this PR regarding the `pkg/bitcoin` package is the logic for `Transaction` serialization and computation of the traditional and witness hashes. Those things are essential to work with Bitcoin transactions, especially for broadcasting them over the network.

### Domain-specific sweep assembling code in the `pkg/tbtc` package

As the core transaction assembling code is handled within the `pkg/bitcoin`, the domain-specific code living in the `pkg/tbtc` package just needs to pull the right levers. This is achieved by introducing the `assembleDepositSweepTransaction` function that crafts the right unsigned sweep transaction according to the provided input data. It's worth noting that this function returns an instance of the `bitcoin.TransactionBuilder` whose state is ready to perform the signing process using the `ComputeSignatureHashes` and `AddSignatures` calls. The `assembleDepositSweepTransaction` is placed in the new `deposit_sweep.go` file. The `deposit` prefix is needed to differentiate the deposit sweep code from the moved funds sweep mechanism that will be presumably added in the future.

Moreover, this pull request introduces the `deposit` domain object. This component is supposed to represent a revealed deposit and matches the data emitted by the on-chain events. Additionally, the `deposit` structure exposes a `script` method that assembles the deposit's Bitcoin redeem script using a pre-defined format stored in the `depositScriptFormat` constant. This redeem script is required to unlock deposits during deposit sweep. Worth noting that the redeem script must have a fixed length and this is the main reason why `deposit`'s fields are defined as fixed-size arrays.

### Sweep testing machinery in the `pkg/internal/tbtctest` package

The sweep assembling code should preferably be tested using real-world data. This guarantees that the produced transactions can be properly handled by Bitcoin nodes. Here we achieve that requirement by adding some testing machinery in the `pkg/internal/tbtctest` package. That package contains some JSON testing scenarios that consist of the input data and the expected output. The package also exposes the functions that can load and unmarshal the test data into a form that can be easily used by actual tests. 

The data used to construct the deposit sweep scenarios were produced during the development of the `tbtc-v2.ts` library (https://github.com/keep-network/tbtc-v2/tree/main/typescript) and represent real testnet transactions that can be browsed using popular blockchain explorers. For reference, please see: https://github.com/keep-network/tbtc-v2/blob/main/typescript/test/data/deposit-sweep.ts

The mechanism described in this paragraph is used to feed the `TestAssembleDepositSweepTransaction` test function that covers the aforementioned `pkg/tbtc.assembleDepositSweepTransaction` function. It is designed in a way that will allow using it for testing future tBTC client features like redemption or moving funds transaction assembling.

### Important note about the `tecdsa.Curve`

During the development of the unit tests for the code presented in this pull request, we noticed a flaw in the `tecdsa.Curve` that is widely used by the tBTC client code. Specifically, the `elliptic.UnmarshalCompressed` function does not work as expected for this curve and always produces nil point coordinates. This is because the `elliptic.UnmarshalCompressed` function always executes the `y² = x³ - 3x + b` equation to compute the `y` coordinate while the actual equation of the `tecdsa.Curve` (secp256k1 underneath) is `y² = x³ + 7`, so the `a` parameter is `0` not `-3`. That means the following code is problematic:
```
// Construct some public key coordinates.
privateKey := ...
x, y := tecdsa.Curve.ScalarBaseMult(privateKey)

// Assert the coordinates are on the curve. The isOnCurve variable is true.
isOnCurve := tecdsa.Curve.IsOnCurve(x, y)

// Marshal the key to the compressed form.
bytes := elliptic.MarshalCompressed(tecdsa.Curve, x, y)

// Try to unmarshal the compressed key. The x1 and y1 variables are nil
// because the elliptic.UnmarshalCompressed performs the wrong equation
// for y so the produced points are not on the curve.
x1, y1 := elliptic.UnmarshalCompressed(tecdsa.Curve, bytes)
```

This problem does not impact the current codebase as we don't use the `elliptic.MarshalCompressed -> elliptic.UnmarshalCompressed` roundtrip. However, we flagged that problem in the `tecdsa.Curve` documentation.


pkg/tbtc/deposit_sweep.go
```diff
@@ -0,0 +1,78 @@
...
+// The resulting bitcoin.TransactionBuilder instance holds all the data
+// necessary to sign the transaction and obtain a bitcoin.Transaction instance
+// ready to be spread across the Bitcoin network.
+func assembleDepositSweepTransaction(
+	bitcoinChain bitcoin.Chain,
+	walletPublicKey *ecdsa.PublicKey,
+	walletMainUtxo *bitcoin.UnspentTransactionOutput,
+	deposits []*deposit,
+	fee int64,
+) (*bitcoin.TransactionBuilder, error) {
```
tomaszslabon, line 27: In the Bridge smart contract we allow the deposit sweep transaction to have a witness (P2WPKH) or non witness (P2PKH) output.
Perhaps we could add a parameter `witness` to this function which could allow choosing the output type?
We did that in our [typescript implementation](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/typescript/src/deposit-sweep.ts#L119).

pkg/internal/tbtctest/tbtctest.go
```diff
@@ -0,0 +1,132 @@
...
+//
+// - deposit_sweep_scenario_2.json: Bitcoin deposit sweep transaction in which
+//   one input (a P2SH deposit) was swept into a P2WPKH main UTXO.
+//   For reference see:
+//   https://live.blockcypher.com/btc-testnet/tx/f5b9ad4e8cd5317925319ebc64dc923092bef3b56429c6b1bc2261bbdc73f351
+//
+// - deposit_sweep_scenario_3.json: Bitcoin deposit sweep transaction in which
+//   one input (a P2WSH deposit) was swept into a P2WPKH main UTXO.
+//   For reference see:
+//   https://live.blockcypher.com/btc-testnet/tx/9efc9d555233e12e06378a35a7b988d54f7043b5c3156adc79c7af0a0fd6f1a0
```
tomaszslabon, line 22: If we decide to add the possibility to choose the deposit sweep transaction output type ([see comment](https://github.com/keep-network/keep-core/pull/3421#discussion_r1042249662)), we could add another scenario which results in a non-witness main UTXO output. We tested that in our [typescript implementation](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/typescript/test/deposit-sweep.test.ts#L287).

pkg/bitcoin/transaction_builder.go
```diff
@@ -0,0 +1,393 @@
...
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"math/big"
+)
+
+// TransactionBuilder is a component that is responsible for the whole
+// transaction creation process. It assembles an unsigned transaction,
+// prepares it for signing, and applies the given signatures in order to
+// produce a full-fledged signed transaction that can be spread across
```
tomaszslabon, line 16: Maybe:
```suggestion
// produce a full-fledged signed transaction that can be broadcast across
```
?

pkg/tbtc/deposit_sweep.go
lukasz-zimnoch: Yep, I know. The non-witness support in the contracts and typescript was just in case and allowed us to have some flexibility. However, the default choice should be witness as those transactions are usually cheaper. Here we reflect that in the contract code. If for some reason (rather unlikely), we need to support non-witness, we can add it easily in the future.

pkg/internal/tbtctest/tbtctest.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3421#discussion_r1044538034

pkg/bitcoin/chain_test.go
```diff
@@ -0,0 +1,59 @@
...
+	transactions map[Hash]*Transaction
+}
+
+func newLocalChain() *localChain {
+	return &localChain{
+		transactions: make(map[Hash]*Transaction),
+	}
+}
+
+func (lc *localChain) GetTransaction(
```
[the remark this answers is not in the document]
lukasz-zimnoch: Those functions are documented in the `bitcoin.Chain` interface. Not sure if we need a copy here, especially for the test local chain.

tomaszslabon (approved):

PR #3423, opened 2022-11-29
Increase announcer's receiver buffer to 512

### LINKED ISSUE #3419: Message handler is too slow; dropping message during tECDSA signing
#3418 fixed the problem with dangling message handlers no longer processing messages and having their buffers full. During the tests of #3404 on commit 9b28e73aff86bd52e1feb9e976e00b9149d26778, so after the fix was reverse-merged, I noticed similar behavior on one of the clients.

The second client (`happy-path-2.txt`) at `2022-11-26T00:22:25.120+0100` started reporting a massive number of `message handler is too slow; dropping message` warnings. This problem was not present on other nodes at that time. The number of warnings suggests it could be - again - some problem with a dangling receiver no longer processing messages.

The logs from all 4 clients running were uploaded to google drive given the size of the files:
https://drive.google.com/drive/folders/11MU2L6gDbUHgQ_b3X_m9cFjZLE3_mjxS?usp=sharing

pdyraga:
Closes #3419
Depends on #3422

During the tests of the code orchestrating tECDSA signing in, we noticed the announcer's buffer sometimes gets full and the messages are dropped with the famous "message handler is too slow" warning. It turns out that the problem lies in the size of the buffer used by the announcer. Before a message in which the announcer is not interested is dropped, it is buffered in this channel. This also includes retransmissions from the previous signing protocols because the retransmission cache filter is scoped to the given `Recv` handler:
```
func WithRetransmissionSupport(delegate func(m net.Message)) func(m net.Message) {
  mutex := &sync.Mutex{}
  cache := make(map[string]bool)
  // (...)
}
```

The announcer's buffer size has been increased to 512 elements, just like the buffer of the asynchronous state machine, and the problem seems to be gone based on local machine tests.

### DIFF
--- a/pkg/protocol/announcer/announcer.go
+++ b/pkg/protocol/announcer/announcer.go
@@ -14,6 +14,14 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// announceReceiveBuffer is a buffer for messages received from the broadcast
+// channel needed when the announcer's consumer is temporarily too slow to
+// handle them. Keep in mind that although we expect only 51 announce messages,
+// it may happen that the announcer receives retransmissions of messages from
+// the previous signing protocol and before they are filtered out as not
+// interesting for the announcer, they are buffered in the channel.
+const announceReceiveBuffer = 512
+
 // announcementMessage represents a message that is used to announce
 // member's participation in the given session of the protocol.
 type announcementMessage struct {
@@ -100,7 +108,8 @@ func (a *Announcer) Announce(
 	[]group.MemberIndex,
 	error,
 ) {
-	messagesChan := make(chan net.Message, a.groupSize)
+	messagesChan := make(chan net.Message, announceReceiveBuffer)
+
 	a.broadcastChannel.Recv(ctx, func(message net.Message) {
 		messagesChan <- message
 	})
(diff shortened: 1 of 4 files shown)


pdyraga:
I was testing this change along with the one from #3422 locally, using the same configuration as for the tests of #3404 where the problems were noticed. I've observed no single "message handler is too slow" warning. 7x10 signatures were produced in the first attempt.

pkg/protocol/announcer/announcer.go
```diff
@@ -100,7 +108,8 @@ func (a *Announcer) Announce(
 	[]group.MemberIndex,
 	error,
 ) {
-	messagesChan := make(chan net.Message, a.groupSize)
+	messagesChan := make(chan net.Message, announceReceiveBuffer)
```
lukasz-zimnoch, line 111: We can remove the `groupSize` field from the `Announcer` struct as it is no longer used.

pdyraga: Oops, I completely overlooked it. See 255aae9d8dbf2997f89e35bcb008f6d683532521, please.

lukasz-zimnoch (approved):

PR #3426, opened 2022-12-01
Simplified DKG result processing logic

pdyraga:
Depends on #3425

We should not look at the misbehaved members from DKG result because this value should be equal to the one returned from the DKG retry loop. The protocol requires all members selected to the given attempt to cooperate to the very end. This condition allows to simplify the code in the following way:

1. Result is produced. It means that everyone who was supposed to participate did participate to the very end.
2. Everyone agreed to the result. They did participate to the very end. We can shift the indexes based on the information we have and register the signer. No snapshot is needed, it's a regular registration.
3. We publish the result to the chain. We sign our version with shifted indexes and everyone else should have the same version.

It means:
- There is no need to `decideSigningMemberFate` and this function was removed.
- We now compute the final signing group with indexes shifted and members excluded earlier, right after the signing protocol returned the result and before we registered the signer.

I have also extracted the code for registering a signer and submitting ECDSA DKG result to separate functions. This allows testing the signer registration path ensuring final indexes and final group members are properly computed. Unit test added.

Issue #3368, opened 2022-10-24
Full ECDSA DKG integration


pdyraga:
The following functions should be covered:
- `WalletRegistry.submitDkgResult` 
- `WalletRegistry.approveDkgResult`
- `WalletRegistry.challengeDkgResult`

PR #3427, opened 2022-12-02
On-chain integration for TBTC DKG

pdyraga:
Refs: https://github.com/keep-network/keep-core/issues/3368
Depends on https://github.com/keep-network/keep-core/pull/3426

This is a work-in-progress PR with an on-chain integration for all `TbtcChain` functions.

Also, it disables mock random beacon chain functions.


pkg/chain/operator_id.go
```diff
@@ -0,0 +1,15 @@
+package chain
+
+import "math"
+
+// OperatorID is a unique identifier of an operator assigned by the Sortition
+// Pool when the operator enters the pool for the first time. ID is never
+// changing for the given operator address.
+type OperatorID = uint32
```
lukasz-zimnoch, line 8: Is there any particular reason to use a type alias here? The default is rather type definition (`type OperatorID uint32`). See https://yourbasic.org/golang/type-alias/

pkg/chain/ethereum/tbtc.go
```diff
@@ -318,49 +348,323 @@ func (tc *TbtcChain) SelectGroup(seed *big.Int) (chain.Addresses, error) {
...
+		return [32]byte{}, err
+	}
+
+	return crypto.Keccak256Hash(bytes), nil
+}
+
+// convertSignaturesToChainFormat converts signatures map to two slices. First
+// slice contains indices of members from the map, second slice is a slice of
+// concatenated signatures. Signatures and member indices are returned in the
+// matching order. It requires each signature to be exactly 65-byte long.
```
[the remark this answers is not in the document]
pdyraga: Yes, it is important!

https://github.com/keep-network/keep-core/blob/089f9b549647feda54d4f86a86480dc4514736b6/solidity/ecdsa/contracts/EcdsaDkgValidator.sol#L168-L180

pkg/chain/ethereum/tbtc.go
```diff
@@ -318,49 +348,323 @@ func (tc *TbtcChain) SelectGroup(seed *big.Int) (chain.Addresses, error) {
...
+		OnEvent(onEvent)
 }
 
-// TODO: Implement a real SubmitDKGResult action. The current implementation
-// just creates and pipes the DKG submission event to the handlers
-// registered in the dkgResultSubmissionHandlers map.
 func (tc *TbtcChain) SubmitDKGResult(
 	memberIndex group.MemberIndex,
-	result *dkg.Result,
+	tecdsaDkgResult *dkg.Result,
```
pdyraga, line 523: I am super confused about this parameter and `convertDkgResultToChainFormat`. 

In `SubmitDKGResult` we take `dkg.Result` and convert it to `ecdsaabi.EcdsaDkgResult` to pass it to `walletRegistry.SubmitDkgResult`. At the same time, we have another function, `convertDkgResultToChainFormat` that accepts `*tbtc.DKGChainResult` and converts it to `ecdsaabi.EcdsaDkgResult`. That being said, I am confused about the relation of `tbtc.DKGChainResult` and `dkgResult`. If we expect all these functions to work with `tbtc.DKGChainResult`, why don't we expect the same from `SubmitDKGResult`?

```

	// IsDKGResultValid checks whether the submitted DKG result is valid from
	// the on-chain contract standpoint.
	IsDKGResultValid(dkgResult *DKGChainResult) (bool, error)

	// ChallengeDKGResult challenges the submitted DKG result.
	ChallengeDKGResult(dkgResult *DKGChainResult) error

	// ApproveDKGResult approves the submitted DKG result.
	ApproveDKGResult(dkgResult *DKGChainResult) error
```

pdyraga: Maybe we should rename `tbtc.DKGChainResult` to `tbtc.SubmittedDKGChainResult`?

pkg/chain/ethereum/tbtc.go
```diff
@@ -370,214 +674,195 @@ func (tc *TbtcChain) GetDKGState() (tbtc.DKGState, error) {
...
+		{Type: uint256Type},
+		{Type: bytesType},
+		{Type: uint8SliceType},
+		{Type: uint256Type},
+	}.Pack(
+		chainID,
+		groupPublicKey,
+		misbehavedMembersIndexes,
+		startBlock,
+	)
```
pdyraga, line 734: I am looking at `EcdsaDkg.submitResult` and the hash is computed as:
https://github.com/keep-network/keep-core/blob/089f9b549647feda54d4f86a86480dc4514736b6/solidity/ecdsa/contracts/libraries/EcdsaDkg.sol#L268

and there are more fields in `Result` structure compared to the ones included here:
https://github.com/keep-network/keep-core/blob/089f9b549647feda54d4f86a86480dc4514736b6/solidity/ecdsa/contracts/libraries/EcdsaDkg.sol#L86-L118

Did we miss something?

pkg/chain/ethereum/tbtc.go
```diff
@@ -370,214 +674,195 @@ func (tc *TbtcChain) GetDKGState() (tbtc.DKGState, error) {
...
 
-	return subscription.NewEventSubscription(func() {
-		mwr.dkgResultSubmissionHandlersMutex.Lock()
-		defer mwr.dkgResultSubmissionHandlersMutex.Unlock()
+// parseDkgResultValidationOutcome parses the DKG validation outcome and returns
+// a boolean indicating whether the result is valid or not. The outcome parameter
+// must be a pointer to a struct containing a boolean flag as the first field.
+//
+// TODO: Find a better way to get the validity flag. This would require
+//       changes in the contracts binding generator.
```
pdyraga, line 760: Can you please open an issue in `keep-common` and link it from here?

mhluongo:
> Worth noting that this PR does not handle DKG retries that should be made upon a DKG result challenge. This scenario will be handled separately.

Is this worth opening a new issue so we don't lose track of it?

> Temporary wallet heartbeats

Ditto, we should keep track of when we consider heartbeats "final".

pkg/tbtc/chain_test.go
```diff
@@ -14,32 +16,40 @@ import (
...
+	dkgMutex       sync.Mutex
+	dkgState       DKGState
+	dkgResult      *DKGChainResult
+	dkgResultValid bool
 
 	blockCounter       chain.BlockCounter
 	chainConfig        *ChainConfig
 	operatorPrivateKey *operator.PrivateKey
+
+	operatorsIDs map[chain.Address]chain.OperatorID
```
pdyraga, line 42: Do we need this map? I think the only place where it is used is `GetOperatorID` and we could return the hardcoded `localChainOperatorID` there. Plus, save the address in `localChain` and compare in `GetOperatorID` if the one from the parameter is as expected.

My point here is that the `localChain` implementation allows for only one address and one operator ID while this mapping suggests otherwise. I was even wondering why we don't have a mutex protecting it.

pkg/tbtc/chain_test.go
```diff
@@ -129,74 +143,291 @@ func (lc *localChain) OnDKGResultSubmitted(
...
 
-// SubmitDKGResult submits the DKG result to the chain, along with signatures
-// over result hash from group participants supporting the result.
+func (lc *localChain) OnDKGResultChallenged(
+	handler func(event *DKGResultChallengedEvent),
+) subscription.EventSubscription {
+	lc.dkgResultChallengeHandlersMutex.Lock()
+	defer lc.dkgResultChallengeHandlersMutex.Unlock()
+
+	handlerID := local_v1.GenerateHandlerID()
```
pdyraga, line 152: What do you think about copy-pasting `GenerateHandlerID` to `pkg/tbtc/chain_test.go` so that we are revering `local_v1` only when absolutely necessary? This should make future refactoring easier.

pkg/tbtc/chain_test.go
```diff
@@ -129,74 +143,291 @@ func (lc *localChain) OnDKGResultSubmitted(
...
-	return AwaitingResult, nil
+	lc.dkgMutex.Lock()
+	defer lc.dkgMutex.Unlock()
+
+	return lc.dkgState, nil
 }
 
-// CalculateDKGResultHash calculates 256-bit hash of DKG result using SHA3-256
-// hashing algorithm.
 func (lc *localChain) CalculateDKGResultHash(
```
pdyraga, line 277: I had similar confusion when looking at this function vs `computeTestDkgResultHash` at the bottom of the file as when writing [this](https://github.com/keep-network/keep-core/pull/3427#discussion_r1052246527) comment. 

In the contract, we use the same `Result` struct everywhere: to submit the result, to challenge the result, and to approve the result. The hash is computed based on the `Result` struct.

In `tbtc.go`, when we submit the result we use `*dkg.Result` plus some additional parameters, and when we approve or challenge the result we use `*DKGChainResult`. The hash is computed based on the `*dkg.Result` struct (which I am not sure is entirely correct - see [the other comment](https://github.com/keep-network/keep-core/pull/3427#discussion_r1052261813) please).

In the test chain implementation, we follow the same approach as `tbtc.go`. The hash is computed based on the `*dkg.Result` struct, `*DKGChainResult` is used to approve and challenge a result. Internally, the result is stored as `*DKGChainResult`

Depending on the result of the discussion from two other comments referenced, I think we should clarify this code as well. 

I think we can:
1. Provide a function that will transform `*dkg.Result` to `DKGChainResult` and compute hash from `DKGChainResult`. Also, use `DKGChainResult` to `SubmitDKGResult`. In local chain implementation, we would have only one function for computing a hash.
2. Rename `DKGChainResult` to `DKGChainSubmittedResult` to clearly indicate it's an on-chain implementation of a result that is already submitted and can only be challenged or approved, improve the documentation to make it clear. But then... what to do about hashes, why to compute hashes in two different ways if it's not what the chain expects 🤔 .

pkg/chain/ethereum/tbtc.go
pdyraga: Please also see [this](https://github.com/keep-network/keep-core/pull/3427#discussion_r1053172506) comment.

pkg/chain/ethereum/tbtc.go
pdyraga: Please also see [this](https://github.com/keep-network/keep-core/pull/3427#discussion_r1053172506) comment.

pkg/chain/operator_id.go
lukasz-zimnoch: I think it's better to leave it as is. Using an alias requires less explicit conversions in the code.

pkg/chain/ethereum/tbtc.go
```diff
@@ -23,6 +28,8 @@ import (
 
 // Definitions of contract names.
 const (
+	// TODO: The WalletRegistry address is taken from the Bridge contract.
+	//       Remove the possibility of passing it through the config.
```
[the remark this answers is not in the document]
lukasz-zimnoch: Opened an issue: https://github.com/keep-network/keep-core/issues/3448

pkg/tbtc/deduplicator.go
```diff
@@ -55,3 +62,25 @@ func (d *deduplicator) notifyDKGStarted(
...
+	// yet and the client should proceed with the execution.
+	if !d.dkgResultHashCache.Has(cacheKey) {
+		d.dkgResultHashCache.Add(cacheKey)
+		return true
+	}
+
+	// Otherwise, the DKG result is a duplicate and the client should not
+	// proceed with the execution.
+	return false
+}
```
pdyraga, line 86: I think we'll have a security hole if we do it this way. I am looking at `EcdsaDkg.sol` and nothing in `submitResult` is blocking me from submitting the same result after it's been challenged. Consider this scenario:
1. Mallory is submitting malicious result.
2. All nodes challenge it, Mallory gets slashed.
3. Mallory is submitting the same result one more time.
4. None of the nodes care about it because they reject the event as a duplicate.

We could be checking if the event comes from the same block as the previous one for the given seed. If yes, reject as a duplicate. But I think this will still be vulnerable in this scenario:
1. Mallory is submitting malicious result at block N.
2. The node receives an event and challenges the result at block N+1.
3. There is soft work and the chain gets reorganized so that block N+1 is rejected but block N stays.

I would consider not looking at the events at all for challenges and instead, periodically check the state of the chain. If we are in the challenge period, check the result.

pkg/chain/ethereum/tbtc.go
```diff
@@ -293,16 +320,19 @@ func (tc *TbtcChain) IsBetaOperator() (bool, error) {
...
 }
 
-// SelectGroup returns the group members for the group generated by
-// the given seed. This function can return an error if the beacon chain's
-// state does not allow for group selection at the moment.
-func (tc *TbtcChain) SelectGroup(seed *big.Int) (chain.Addresses, error) {
-	groupSize := big.NewInt(int64(tc.GetConfig().GroupSize))
-	seedBytes := [32]byte{}
-	seed.FillBytes(seedBytes[:])
+func (tc *TbtcChain) GetOperatorID(
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 0a8eef3d

pkg/chain/ethereum/tbtc.go
```diff
@@ -293,16 +320,19 @@ func (tc *TbtcChain) IsBetaOperator() (bool, error) {
...
+	)
+}
 
-	// TODO: Replace with a call to the WalletRegistry.selectGroup function.
-	operatorsIDs, err := tc.sortitionPool.SelectGroup(groupSize, seedBytes)
+// SelectGroup returns the group members selected for the current group
+// selection. The function returns an error if the chain's state does not allow
+// for group selection at the moment.
+func (tc *TbtcChain) SelectGroup() (*tbtc.GroupSelectionResult, error) {
+	operatorsIDs, err := tc.walletRegistry.SelectGroup()
```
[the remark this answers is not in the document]
lukasz-zimnoch: We can do it just in case. See 18429dc6

pkg/chain/ethereum/tbtc.go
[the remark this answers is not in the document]
lukasz-zimnoch: Yep, forgot to mention it in the function docs. See 436b114c

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: Well, we actually have two sorts of hashes. 

The first sort is used when members sign the result to show their support and mutually verify their signatures before attempting to submit the result. This hash is also used during the result validation to make sure group members signed the right thing. See https://github.com/keep-network/keep-core/blob/089f9b549647feda54d4f86a86480dc4514736b6/solidity/ecdsa/contracts/EcdsaDkgValidator.sol#L221 This is the hash the client needs to compute by itself.

The second type of hash is the hash you mentioned here. This one is used internally by the contract to store the result in a gas-efficient way. The client does not need to compute it though that hash is populated in DKG-related events that are emitted after submission.

I should make it clear in the code and use better naming. I'll try to improve that.

pkg/tbtc/dkg.go
```diff
@@ -480,47 +484,180 @@ func (de *dkgExecutor) registerSigner(
...
+		dkgResult,
+	)
+}
+
+// executeDkgValidation performs the submitted DKG result validation process.
+// If the result is not valid, this function submits an on-chain result
+// challenge. If the result is valid and the given node was involved in the DKG,
+// this function schedules an on-chain approve that is submitted once the
+// challenge period elapses.
+func (de *dkgExecutor) executeDkgValidation(
```
pdyraga, line 517: I'd consider extracting this function to `dkg_approve.go` or `dkg_validate.go`, for example.

pkg/tbtc/dkg.go
```diff
@@ -480,47 +484,180 @@ func (de *dkgExecutor) registerSigner(
...
+				)
+				return
+			}
+
+			dkgLogger.Infof("[member:%v] approving DKG result", memberIndex)
+		}(currentMemberIndex)
+	}
 }
 
 // finalSigningGroup takes three parameters:
```
pdyraga, line 663: Triple-checking this part. It this all correct?

1. Signing group members are selected on-chain. Some of them may be inactive or malicious and excluded from the given attempt during the announcement phase or by the retry loop.
2. During the execution of the off-chain key generation protocol they always have the same member indexes as their positions in the signing group selected on-chain.
3. The member index is used in the cryptography key generation protocol.
4. Once we are done with the protocol, we shift some member indexes before saving them on disk so that no empty places are left after inactive/misbehaving members that would lead to delayed submission of work to the chain.
5. We can do it because the member index no longer matters for the signing protocol. The original member index was saved in TSS data specific to the given key share.

I also wonder if it makes sense to improve `finalSigningGroup` documentation by including all these points.

pkg/tbtc/dkg_loop.go
```diff
@@ -15,6 +15,32 @@ import (
...
+	// announcement phase that is performed at the beginning of each DKG
+	// attempt.
+	dkgAttemptAnnouncementActiveBlocks = 5
+	// dkgAttemptProtocolBlocks determines the maximum block duration of the
+	// actual protocol computations.
+	dkgAttemptMaximumProtocolBlocks = 150
+	// dkgAttemptCoolDownBlocks determines the duration of the cool down
+	// period that is preserved between subsequent DKG attempts.
+	dkgAttemptCoolDownBlocks = 5
+)
```
pdyraga, line 33: `dkgRetryLoop struct` still has `announcementDelayBlocks` and `announcementActiveBlocks`. Should we remove them?

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: See 07d4b585

pkg/tbtc/dkg.go
```diff
@@ -296,14 +312,14 @@ func (de *dkgExecutor) generateSigningGroup(
 
 					// Set up the attempt timeout signal.
```
pdyraga, line 313: Given that we moved all attempt-related timeouts to `dkg_loop.go`, would it be cleaner if we passed the timeout block as a parameter in `dkgAttemtParams`? We do this for signing:

https://github.com/keep-network/keep-core/blob/71e0ac8fb1cd6a63994d897216a2d85764fe85a7/pkg/tbtc/signing_loop.go#L282

Also, not something we should solve in this PR, but I think it's worth opening an issue if you agree: instead of passing timeout block we should pass a context. The context should be cancelled inside `dkg_loop.go` / `signing_loop.go`.

Right now, the retry loop assumes the attempt should start at the given block:
https://github.com/keep-network/keep-core/blob/71e0ac8fb1cd6a63994d897216a2d85764fe85a7/pkg/tbtc/signing_loop.go#L182-L185

but it can't guarantee the previous attempt ended before. It has to trust `signingAttemptFn`. The same situation is for DKG.

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: I reworked it. See fdbd05f8 (also 07d4b585 and cb192599 to capture the big picture)

pkg/tbtc/chain_test.go
lukasz-zimnoch: Please see https://github.com/keep-network/keep-core/pull/3427#discussion_r1053347458

pkg/tbtc/dkg.go
```diff
@@ -19,23 +19,26 @@ import (
...
+	// dkgResultSubmissionDelayStep determines the delay step in blocks that
+	// is used to calculate the submission delay period that should be respected
+	// by the given member to avoid all members submitting the same DKG result
+	// at the same time.
+	dkgResultSubmissionDelayStepBlocks = 5
+	// dkgResultApprovalDelayStepBlocks determines the delay step in blocks
+	// that is used to calculate the approval delay period that should be
+	// respected by the given member to avoid all members approving the same
+	// DKG result at the same time.
+	dkgResultApprovalDelayStepBlocks = 5
```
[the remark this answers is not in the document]
lukasz-zimnoch: This timeout has nothing to do with result approvals. Once the result is submitted and waits for approval, that timeout is not relevant. Looks like we can just increase the `dkgResultApprovalDelayStepBlocks` constant to 10 or 15 blocks. Am I missing something?

pkg/tbtc/dkg.go
```diff
@@ -138,47 +147,40 @@ func (de *dkgExecutor) executeDkgIfEligible(
...
 	}
 
-	dkgLogger.Infof("selected group members for DKG = %s", selectedSigningGroupOperators)
+	dkgLogger.Infof(
+		"selected group members for DKG = %s",
+		groupSelectionResult.OperatorsAddresses,
+	)
 
-	if len(selectedSigningGroupOperators) > de.chain.GetConfig().GroupSize {
+	if len(groupSelectionResult.OperatorsAddresses) > de.chain.GetConfig().GroupSize {
```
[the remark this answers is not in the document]
lukasz-zimnoch: > The requirements about the group size or threshold do not come from the chain

Is that true? Those are constants defined in `EcdsaDkg` and `EcdsaDkgValidator`. If we want to use different group sizes, we need to update the contracts and the client must adjust.

pkg/chain/ethereum/tbtc.go
```diff
@@ -590,13 +924,38 @@ func (mwr *mockWalletRegistry) OnSignatureRequested(
...
+func (tc *TbtcChain) activeWalletPublicKey() ([]byte, bool, error) {
+	walletPublicKeyHash, err := tc.bridge.ActiveWalletPubKeyHash()
+	if err != nil {
+		return nil, false, fmt.Errorf(
+			"cannot get active wallet public key hash: [%v]",
+			err,
+		)
+	}
+
+	if walletPublicKeyHash == [20]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0} {
```
tomaszslabon, line 936: I think this could be simplified to:
```suggestion
	if walletPublicKeyHash == [20]byte{} {
```

pkg/tbtc/dkg.go
```diff
@@ -244,6 +253,30 @@ func (de *dkgExecutor) generateSigningGroup(
...
+							"key [0x%x] and result hash [0x%x] submitted "+
+							"at block [%v] by member [%v]",
+						memberIndex,
+						event.Result.GroupPublicKey,
+						event.ResultHash,
+						event.BlockNumber,
+						event.Result.SubmitterMemberIndex,
+					)
+				})
+			defer subscription.Unsubscribe()
```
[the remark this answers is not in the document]
lukasz-zimnoch: I think our approach should be: If the result got challenged, we should start a new DKG with the malicious submitter member excluded. This is simpler and safer. We can't rely on a result produced with the participation of a malicious member. The contract does not enforce how we act once a result is challenged, it just returns to `AWAITING_RESULT` phase which also includes result computation. If we agree with that, I think the current shape of `generateSigningGroup` is proper. If a result is challenged, we just start a new DKG using the original start block shifted by `resultSubmissionStartBlockOffset`.

pkg/tbtc/dkg.go
```diff
@@ -296,14 +312,14 @@ func (de *dkgExecutor) generateSigningGroup(
...
+						ctx,
 					)
 					go func() {
 						defer cancelAttemptCtx()
 
 						err := de.waitForBlockFn(
-							loopCtx,
-							attempt.startBlock+dkgAttemptMaxBlockDuration,
+							ctx,
+							attempt.startBlock+dkgAttemptMaximumProtocolBlocks,
```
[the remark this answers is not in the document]
lukasz-zimnoch: If we want to adjust something, we need to adjust the on-chain `resultSubmissionTimeout`. The client `dkgAttemptMaximumProtocolBlocks` was chosen empirically to cover the worst case of protocol computations.

pkg/tbtc/dkg.go
lukasz-zimnoch: > Once we are done with the protocol, we shift some member indexes before saving them on disk **so that no empty places are left after inactive/misbehaving members that would lead to delayed submission of work to the chain.**

The bold part is not true. Indeed, we shift but just to have proper indexes for signing. For result submission, we still need to operate on original indexes as the contract does some validations of group membership based on them.

pkg/tbtc/tbtc.go
```diff
@@ -125,9 +136,39 @@ func Initialize(
...
+				event.Seed,
+				event.BlockNumber,
+				event.Result,
+				event.ResultHash,
+			)
+		}()
+	})
+
+	// TODO: Repeat DKG in case of the result challenge and archive the
+	//       signers stored for the invalid result.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3427#discussion_r1053377999. Regarding:

> The risk with the key generation happening more than once is that someone can try to submit the result for another DKG that happens in the future.

I don't quite get it. Can you write a problematic scenario?

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: Opened https://github.com/keep-network/keep-common/issues/111

pkg/tbtc/chain.go
```diff
@@ -67,42 +116,51 @@ type DKGStartedEvent struct {
...
-	//       have a test signing loop so we are simulating that signatures are
-	//       requested by the Bridge explicitly.
-	OnSignatureRequested(
-		func(event *SignatureRequestedEvent),
+	// OnHeartbeatRequested registers a callback that is invoked when an
+	// on-chain notification of the Bridge heartbeat request is seen.
+	OnHeartbeatRequested(
+		func(event *HeartbeatRequestedEvent),
 	) subscription.EventSubscription
 }
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yep, this is just the next iteration. I opened an issue in order to flag that we need a final heartbeat mechanism to be implemented in the client: https://github.com/keep-network/keep-core/issues/3451

pkg/tbtc/chain.go
```diff
@@ -67,42 +116,51 @@ type DKGStartedEvent struct {
...
+// emitted after a submitted DKG result lands on the chain.
 type DKGResultSubmittedEvent struct {
-	MemberIndex         uint32
-	GroupPublicKeyBytes []byte
-	Misbehaved          []uint8
+	Seed        *big.Int
+	ResultHash  [32]byte
+	Result      *DKGChainResult
+	BlockNumber uint64
+}
```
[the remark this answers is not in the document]
lukasz-zimnoch: I'd rather prefer to keep `ResultHash` and `Result` close to each other.

pkg/tbtc/chain_test.go
lukasz-zimnoch: See d78aa84f

pkg/tbtc/dkg.go
```diff
@@ -50,6 +53,8 @@ type dkgExecutor struct {
 // newDkgExecutor creates a new instance of dkgExecutor struct. There should
 // be only one instance of dkgExecutor.
 func newDkgExecutor(
+	operatorIDFn func() (chain.OperatorID, error),
```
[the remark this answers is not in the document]
lukasz-zimnoch: Because the `chain.OperatorID` is `0` if the client has not joined the sortition pool yet. This is noted in a comment within `newNode` function.

pkg/tbtc/dkg.go
lukasz-zimnoch: It's part of the `dkgExecutor` that lives in `dkg.go` entirely. Not sure if scattering that component across multiple files will give us any benefits.

pkg/tbtc/dkg.go
lukasz-zimnoch: See d659537d. Regarding:

> instead of passing timeout block we should pass a context

I don't think it's worth our time now. I think so because:
- the current shape is good enough (leave well alone!) and this mechanism is quite complex. We may not be aware of all implications of this change right now
- the current approach has clear and balanced boundaries: the loop just determines the start and timeout blocks and does not care what `signingAttemptFn` does. After all, `signingAttemptFn` triggers the actual action and has the full knowledge to clean up it properly. Using a context prepared within the loop is less flexible and will push much more responsibilities on the loop side.
- the current approach is easier to test as we can easily check whether the blocks proposed by the loop makes sense for the given attempt

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: Yep! See d05c5c08

pkg/tbtc/deduplicator.go
lukasz-zimnoch: I'm afraid we need to use events anyway. A periodic check of the DKG state is not an option as you need to have a plain-text result that was submitted. The contract stores just the hash. I think we have to shorten the cache time to a reasonable value, that will exclude duplicated events on the one hand, and prevent the described scenario on the other hand. WDYT?

pdyraga: We can use `PastDkgResultSubmittedEvents` exposed by the `WalletRegistry` contract Go bindings to obtain the last event if we decide for the periodic checks option.

I am not sure if there is a *reasonable value* for the cache time in this case. It would depend too much on the current situation: what is the block time, did we have a chain reorg, when the malicious result got re-submitted.

I think we have two options:
1. Do the check periodically, use `PastDkgResultSubmittedEvents`
2. Accept we may receive duplicates for `DkgResultSubmittedEvent` and always do the validation.

I think the first option is cleaner and safer: all nodes, all the time, watch the situation. If they detect a problem, they all try to act. This is like a watchman process in the background that never sleeps.

lukasz-zimnoch: I think `PastDkgResultSubmittedEvents` will also provide us with duplicates. This is often the case when the client needs to reconnect with their ETH node. That said, it seems this will not improve our situation. My personal preference is to do the following:
- Always validate without checking duplicates and challenge if the result is invalid
- If the result is valid, check if the event is not duplicated and if not, schedule the result approval 

What do you think?

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
pdyraga: There is `dkgResultSubmissionDelayStepBlocks` one line earlier that is related to `dkg.setResultSubmissionTimeout(100 * 20)` in the contract.

I think updating both `dkgResultSubmissionDelayStepBlocks` and `dkgResultApprovalDelayStepBlocks` to 10/15 blocks makes sense but we'll have to update the on-chain timeout as well. Let's keep discussing the on-chain parameter update in [this](https://github.com/keep-network/keep-core/pull/3427#discussion_r1053381266) thread.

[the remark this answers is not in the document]
lukasz-zimnoch: Ok, I'll update both to `10` blocks.

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
pdyraga: These are the numbers of DKG retries we need to execute for non-attributable protocol:
- 1 malicious member: 100 retries worst case, 50 retries average case. (P = (99 choose 99) / (100 choose 99) = 0.01)
- 2 malicious member: 4950 retries worst case, 2475 retries average case.
- 3 malicious member: 161700 retries worst case, 80850 retries average case.

This is quite a lot and I think we should agree that if there is even one malicious member in the group, it is better not to create this group at all because it will require multiple retries during the signing:
- 1 malicious member, 99 honest members, P = (99 choose 51) / (100 choose 51) = 0.49
- 2 malicious members, 98 honest members, P = (98 choose 51) / (100 choose 51) = 0.23757575757
- 3 malicious members, 97 honest members,  P = (97 choose 51) / (100 choose 51) = 0.11393939393

So let's assume we do not allow for any malicious behavior during DKG but we take into account inactivities after the announcement phase. If we allow for 5 retries of the protocol we need `5 x (1 + 5 + 150 + 5) + 15x100 = 2305` blocks, which is about 7.5h assuming 15 blocks delay for a result publication. What do you think?

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
pdyraga: I think we could go to 15, see [this](https://github.com/keep-network/keep-core/pull/3427#discussion_r1055472132) thread.

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
pdyraga: What I mean is that we have a bunch of assumptions about the group size all over `pkg/tbtc`. For example, the `dkgAttemptMaximumProtocolBlocks` was chosen based on the tests of the 100-members group.

The chain has its own expectations about the group size but the client also has them, all across the board. In the current form, with the current constants, `pkg/tbtc` will work for groups of 100 but will not work for groups of 1000 on M1 dev machine. It will work for groups of 10 but retries may be highly inefficient. 

In other words, I think the group size of 100 is now also an expectation of `pkg/tbtc`.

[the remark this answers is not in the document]
lukasz-zimnoch: Ok, got it now. Thanks for the clarification! I'll change it.

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
pdyraga: The result can be submitted by anyone who is in the sortition pool, it does not have to be a member of the selected signing group. If we start another DKG then we'll have two valid results and no guarantee which one will be submitted/accepted. I am pretty sure our assumption when implementing the contract was that there is only one successful DKG attempt, and the single valid result will get re-submitted after the challenge if someone malicious submits their version first. No matter if the malicious player is from the group or not from the group.

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
lukasz-zimnoch: That makes sense!

pkg/tbtc/tbtc.go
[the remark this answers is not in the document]
pdyraga: I was not super precise here: if we allow for another DKG to happen, then we end up with two valid DKG results and both of them can be submitted and pass the on-chain validation. We can continue this discussion in [this](https://github.com/keep-network/keep-core/pull/3427#discussion_r1055513333) thread as it's basically about the same concerns. I'll leave this comment open so that we can remember about potentially updating/removing the TODO.

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
lukasz-zimnoch: Ok, that makes sense. However, I'd leave it as is for now and approach it during the work on #3450. For now, it's not clear how the challenge reaction will be exactly implemented so it's hard to predict everything. Right now we have a clear path: try to execute and publish DKG until the timeout or result submission, dot. No need to bother with timeout shifts if we don't yet re-submit the result upon challenge.

pkg/tbtc/dkg.go
pdyraga: By *delayed submission of work to the chain*  I did not mean the DKG result submission because for this one, as you say, we need to use the original member index from the group selection, but everything that will potentially happen later: inactivity claim or one or two Bridge functions only group member can call.

Does it make sense now?

Also, regarding signing, how do we use the member index for signing? My understanding is that we could not shift the indexes as the index used for cryptography protocol in tss-lib is stored in `*tecdsa.PrivateKeyShare` and this is the original one from the group selection. The member index shifting is just to not have gaps in logs when it comes to signing. For example if we have a group of 95 members, we want the indexes to be from 1-95.

Is this correct?

pkg/tbtc/tbtc.go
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, that was cleared out in the referenced discussion. I removed that TODO (9ea6adf4) since we have this work captured by https://github.com/keep-network/keep-core/issues/3450

pkg/tbtc/dkg.go
lukasz-zimnoch: > By delayed submission of work to the chain I did not mean the DKG result submission because for this one, as you say, we need to use the original member index from the group selection, but everything that will potentially happen later: inactivity claim or one or two Bridge functions only group member can call.
>
> Does it make sense now?

Yep, perfectly makes sense.

> Also, regarding signing, how do we use the member index for signing? My understanding is that we could not shift the indexes as the index used for cryptography protocol in tss-lib is stored in *tecdsa.PrivateKeyShare and this is the original one from the group selection. The member index shifting is just to not have gaps in logs when it comes to signing. For example if we have a group of 95 members, we want the indexes to be from 1-95.
>
> Is this correct?

If we end up with 95 members as a result of DKG, the underlying indexes in the key shares are in the same range, so 1-95. Shifting, however, is not only about log gaps but mostly about how the signing algorithm selects the members for the given attempt. It needs to pick up 51 members out of X (actual group size). If we don't shift, X is 95, and we have excluded 5 members with indexes <= 95, the algorithm may select those excluded members and will never select proper members with indexes > 95. This is the main reason

pdyraga: > If we end up with 95 members as a result of DKG, the underlying indexes in the key shares are in the same range, so 1-95. 

Can you please point me to the place in the code where this is happening? I am struggling to find it.

lukasz-zimnoch: Wait, I was wrong saying this: "If we end up with 95 members as a result of DKG, the underlying indexes in the key shares are in the same range, so 1-95". Those are original indexes from group selection as you said, see [here](https://github.com/keep-network/keep-core/blob/b8dcba4a37e68febff082bd2c6282b5fac1f7d04/pkg/tecdsa/dkg/member.go#L132). However, the rest I said above is still true.

lukasz-zimnoch: All conversions between our member index and the underlying index from the key share (TSS id) are then handled using `identityConverter` that has two implementations, one for DKG and one for signing.

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
pdyraga: Sure, we can leave it as-is for now. Can we just add a TODO in the code? Something as simple as:
```
// TODO: This subscription has to be updated once we implement
// re-submitting DKG result to the chain after a challenge.
// See https://github.com/keep-network/keep-core/issues/3450
```

pkg/tbtc/dkg.go
pdyraga: Ah, ok - I see now `dkg_submit.go` hosts `dkgResultSigner` and `dkg_loop.go` hosts `dkgRetryLoop`. That makes sense!

pkg/tbtc/dkg.go
pdyraga: I see we have all of it explained perfectly in the identify converter docs:

https://github.com/keep-network/keep-core/blob/b8dcba4a37e68febff082bd2c6282b5fac1f7d04/pkg/tecdsa/common/common.go#L14-L60

Can we please just drop one sentence to the end of `finalSigningGroup` docs to avoid having this discussion next time? Something like:
```
// Please see docs of identityConverter from pkg/tecdsa/common/ for more
// information about shifting indexes.
```

pkg/tbtc/dkg_test.go
```diff
@@ -160,6 +165,211 @@ func TestRegisterSigner(t *testing.T) {
...
+				if slices.Contains(test.controlledMembersIndexes, memberIndex) {
+					operatorsIDs[memberIndex-1] = operatorID
+					operatorsAddresses[memberIndex-1] = operatorAddress
+				}
+			}
+
+			groupSelectionResult := &GroupSelectionResult{
+				OperatorsIDs:       operatorsIDs,
+				OperatorsAddresses: operatorsAddresses,
+			}
```
pdyraga, line 267: I am lost about `controlledMembersIndexes` construction and this test as a whole.

First, I think we have inconsistency in test data preparation: `signatures` will have `groupSize` entries but `GroupSelectionResult` will only have `len(controlledMembersIndexes)` entries.

Second, the submitter member index in `executeDkgValidation` is obtained from `operatorID, err := de.operatorIDFn()` and then looking at `*GroupSelectionResult`. Why do we care about a collection of `controlledMembersIndexes`?

pkg/tbtc/dkg_submit_test.go
```diff
@@ -283,6 +379,13 @@ func TestSubmitResult_MemberDoesNotSubmitsResult(t *testing.T) {
...
+	// Set up a global listener that will cancel the common context upon result
+	// submission. That mimics the real-world scenario.
+	localChain.OnDKGResultSubmitted(
+		func(event *DKGResultSubmittedEvent) {
+			cancelCtx()
+		})
+
 	secondMemberSubmissionChannel := make(chan error)
 
 	// Attempt to submit result for the second member on a separate goroutine.
```
pdyraga, line 391: Not something we need to worry about now but... why does it have to be a separate goroutine? Can't we do it synchronously? We `time.Sleep` one line later to have this code executed first.

pkg/tbtc/dkg_submit.go
```diff
@@ -114,66 +127,77 @@ func (drs *dkgResultSubmitter) SubmitResult(
...
+
+	if ctx.Err() != nil {
+		// The context was cancelled by the upstream. Regardless of the cause,
+		// that means the DKG is no longer awaiting the result, and we can
+		// safely return.
+		drs.dkgLogger.Infof(
+			"[member:%v] DKG is no longer awaiting the result; "+
+				"aborting DKG result on-chain submission",
+			memberIndex,
+		)
```
pdyraga, line 174: Is this always true? I think we do not know why the context got canceled, this could also be something else than the timeout 🤔 Isn't this assumption too much here?

Shouldn't we, for example:
- log here that we are aborting result submission because the parent context got cancelled
- when cancelling the parent context log that it's because of a timeout
?

pdyraga (commented):
I went through all files but was doing no tests. Have no more comments but those pasted so far.

pkg/tbtc/tbtc.go
```diff
@@ -125,9 +136,36 @@ func Initialize(
 		}()
```
pdyraga, line 136: There is one more thing that occurred to me: at line 112 in this file (not modified in this PR) we have:
```
_ = chain.OnDKGStarted(func(event *DKGStartedEvent) {
  go func() {
    if ok := deduplicator.notifyDKGStarted(
      event.Seed,
    ); !ok {

    // (...)
    
    node.joinDKGIfEligible(
      event.Seed,
      event.BlockNumber,
    )
  }()
})
```

I am pretty sure this will be problematic on mainnet as-is because of the fact `event.BlockNumber` is a part of the signature:

https://github.com/keep-network/keep-core/blob/b8dcba4a37e68febff082bd2c6282b5fac1f7d04/solidity/ecdsa/contracts/libraries/EcdsaDkg.sol#L103-L106

We will have small chain reorganizations on mainnet resulting in `DKGStartedEvent`s emitted one after another, with different `startBlock`. They will be deduplicated by `deduplicator.notifyDKGStarted` but the `startBlock` matters later for the on-chain result submission.

What I would suggest doing here is to wait for 20 blocks (this feels like safe enough finalization for Ethereum) then take the last, `GetPastDkgStartedEvent`, and trigger DKG based on the values from that event.

pkg/tbtc/dkg_submit.go
lukasz-zimnoch: In this case, the cancellation of parent context means the result was submitted or a timeout was hit. In both cases, the DKG no longer awaits the result. Hence, in my opinion, that log makes sense and is more verbose than a mysterious "context got cancelled" that often causes confusion, even among keep-core developers.

pkg/tbtc/tbtc.go
lukasz-zimnoch: Yeah, good point. Just a couple of questions:
- Are we sure `GetPastDkgStartedEvent` will return the same event for all? I can imagine a case when not all members receive the same due to ETH node re-connections or slight state differences
- What about deduplication? Do we want to keep the current logic? Do we want to compare the original event with the one fetched via `GetPastDkgStartedEvent`?

pkg/tbtc/dkg_test.go
lukasz-zimnoch: The `controlledMembersIndexes` is for simulating a case when the given member is not controlled by the tested node so their action can be omitted. For example, we want to check the approval after the submitter's precedence period. We use member 1 as the submitter but we do not make it "controlled" (their index is not in `controlledMembersIndexes`). This way, member 1 does not approve so the result can be approved by member 2 after the precedence period.

pkg/tbtc/tbtc.go
pdyraga: > Are we sure GetPastDkgStartedEvent will return the same event for all? I can imagine a case when not all members receive the same due to ETH node re-connections or slight state differences

If we wait for 20 blocks then I am pretty sure the answer is yes. This function, under the hood, is just fetching filtered logs from the contract. It's a simple call to Ethereum node, it does not maintain a state. If the smart contract's state has enough confirmations than all nodes should receive the same result.

> What about deduplication? Do we want to keep the current logic? Do we want to compare the original event with the one fetched via GetPastDkgStartedEvent?

I think we need to deduplicate to not start multiple monitoring loops for 20 confirmations. I don't think we need to compare events, I think we should just accept the fact the event can change and use the one fetched after the confirmations. 

In my mind, this works as: 
- I know DKG is probably going to happen. So I start just one (!) monitoring.
- I know the information I have right now may change, I need to wait 20 blocks to see the final version.
- After 20 blocks, I confirm if DKG is in progress by checking the event. I take that event as a final source of truth.

lukasz-zimnoch: Ok, I'll rework that.

pkg/tbtc/dkg_test.go
pdyraga: I am sorry but I still do not follow. In this test (`TestDkgExecutor_ExecuteDkgValidation`) we test the `dkgExecutor.executeDkgValidation` function. `dkgExecutor` is created with only one `operatorIDFn`, so it is controlling exactly one operator - the one fetched from the local chain: `operatorID, err := localChain.GetOperatorID(operatorAddress)`. `test.controlledMembersIndexes` are used only in one place: to construct `operatorsIDs` and `operatorsAddresses` that are later used to construct `GroupSelectionResult`:
```
			for memberIndex := uint8(1); memberIndex <= groupSize; memberIndex++ {
				signatures[memberIndex] = []byte{memberIndex}

				if slices.Contains(test.controlledMembersIndexes, memberIndex) {
					operatorsIDs[memberIndex-1] = operatorID
					operatorsAddresses[memberIndex-1] = operatorAddress
				}
			}

			groupSelectionResult := &GroupSelectionResult{
				OperatorsIDs:       operatorsIDs,
				OperatorsAddresses: operatorsAddresses,
			}
```

With this code, `GroupSelectionResult` contain only operators from the  `test.controlledMembersIndexes` while it should contain `groupSize` of operators - all selected to the given group.

pkg/chain/ethereum/tbtc.go
```diff
@@ -318,266 +352,566 @@ func (tc *TbtcChain) SelectGroup(seed *big.Int) (chain.Addresses, error) {
...
+
+	return crypto.Keccak256Hash(bytes), nil
 }
 
-func (mwr *mockWalletRegistry) OnDKGStarted(
-	handler func(event *tbtc.DKGStartedEvent),
-) subscription.EventSubscription {
-	ctx, cancelCtx := context.WithCancel(context.Background())
-	blocksChan := mwr.blockCounter.WatchBlocks(ctx)
+// convertSignaturesToChainFormat converts signatures map to two slices. First
```
tomaszslabon, line 619:
```suggestion
// convertSignaturesToChainFormat converts signatures map to two slices. The first
```

pkg/tbtc/deduplicator.go
lukasz-zimnoch: Option 2 turned out to be much more complicated than it seemed:
- We don't have a simple way to determine an efficient block span and the current seed to pre-filter fetched events. That may cause problems with some ETH nodes that limit the number of returned past events. Determining proper filter values requires a separate mechanism that greatly increases the complexity of the entire solution
- We need to have a mechanism that makes sure we schedule the approval for the given result only once
- A lot of calls to the ETH node must be made
- We run the check every 30 minutes while DKG is much more sparse

I finally came up with option 1 with some modifications to the deduplication logic (see b70ccb8b). The deduplicator now considers the DKG seed and the submission block for the given result. This way, we can still validate the same malicious result submitted multiple times, even for different DKGs. This idea comes from https://github.com/keep-network/keep-core/pull/3427#discussion_r1053192176 and, as mentioned, is still vulnerable to chain reorgs. However, this problem should be mitigated by a challenge confirmation mechanism that was introduced according to the original option 1 description. That said, the solution presented in b70ccb8b should address all problems mentioned in this discussion. Even though the client loses the challenge confirmation state upon restart, that should not be a problem as we have redundancy coming from other clients working in the network that should also take care of the challenge. In my opinion, the presented solution is also an optimal one regarding code quality. The original option 1 leads to a weird structure when the result deduplication logic is mixed with the DKG executor code.

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
lukasz-zimnoch: f9c0d1ba

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
lukasz-zimnoch: See 349bdfa8

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
lukasz-zimnoch: See f341e4e0

pkg/tbtc/dkg.go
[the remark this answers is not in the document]
lukasz-zimnoch: Created an issue to capture this work https://github.com/keep-network/keep-core/issues/3455

pkg/tbtc/dkg.go
lukasz-zimnoch: See de2d1826

pkg/chain/ethereum/tbtc.go
```diff
@@ -318,266 +352,566 @@ func (tc *TbtcChain) SelectGroup(seed *big.Int) (chain.Addresses, error) {
...
+
+	return calculateDKGResultSignatureHash(
+		tc.chainID,
+		unprefixedGroupPublicKeyBytes,
+		misbehavedMembersIndexes,
+		big.NewInt(int64(startBlock)),
+	)
+}
+
+// computeDkgResultHash computes the keccak256 hash for the given DKG result
```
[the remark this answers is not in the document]
lukasz-zimnoch: See a630a63a

pkg/chain/ethereum/tbtc.go
```diff
@@ -318,266 +352,566 @@ func (tc *TbtcChain) SelectGroup(seed *big.Int) (chain.Addresses, error) {
...
+						continue
+					}
+
+					if !ok {
+						logger.Infof("there is no active wallet for heartbeat at the moment")
+						continue
 					}
 
-					mwr.activeWalletMutex.RUnlock()
+					if len(walletPublicKey) > 0 {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Good catch! See be0aac81

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: See 55d01d65

pkg/tbtc/dkg_submit_test.go
lukasz-zimnoch: We need to make member 2 waiting for its submission turn and member 1 must actually submit during member 2 wait time. We can't do that without the goroutine.

pkg/tbtc/dkg_test.go
lukasz-zimnoch: See 9536de6a

pkg/tbtc/tbtc.go
lukasz-zimnoch: I spent some time on it and I'm afraid this is not so simple. If we wait for 20 confirmation blocks, we start the actual off-chain protocol with a delay that breaks a lot of things. This is because we use the original block number from the event as the starting point but the actual block is more than 20 blocks in the future. This PR is already a ball of mud so I think we should think about it and address that problem separately. I created an issue capturing it https://github.com/keep-network/keep-core/issues/3456

tomaszslabon (approved):
@pdyraga @lukasz-zimnoch 
I tested a network of three clients and one bootstrap. The DKG was submitted successfully:
```
2023-01-02T18:47:12.447+0100    INFO    keep-tbtc       tbtc/dkg.go:279 [member:63] DKG result with group public key [0xc0162fd817b7cbb9e4130bec8d69b7830c2610502e5c8e845bca0d1ab37fe6ac9d3e4972586aaa5de424d52505bbba347466a3078a9cb9fe493839a2f2a34d1b] and result hash [0x0ff2b9f501d2d81e608beb1645e1de1abed567dc24895fbf34b0e03982622613] submitted at block [1704] by member [1] {"seed": "0xbed8cada2f7e8cbfaae4b68557faab80ca3f6442532c6540ed181160fdb68dc5"}
```

Note that I used a simple `RandomBeaconChaosnet` contract that calls `WalletRegistry`'s callback without any gas limit.

PR #3440, opened 2022-12-07
Split the signing done check into a listen and a wait phases

pdyraga:
Consider the following scenario: three members in a signing group: Bob, Eve, and Alice. They all participate in signing and are expected to confirm they are done.

Bob finishes signing at block 1000.
Eve finishes signing at block 1001.
Alice finishes signing at block 1013.

Bob starts signaling he is done at block 1000.
Eve starts signaling she is done at block 1001.
Alice starts signaling she is done at block 1013.

This all looks good except that in the previous version of the done check, Eve was not accepting Done messages from Bob before block 1001 and Alice was not accepting Done messages from Bob and Eve before block 1013. Eve and Alice had to count on retransmissions.

Since we use a backoff strategy for retransmissions (`R _ R _ _ R _ _ _ _ R _ _ _ _ _ _ _ _ R...`), they could be lucky to receive a message early or they could not be lucky and wait for a longer period of time.

We observed this on testnet as a problem with synchronization of the announcement phase. Even though Bob and Eve received a signing Done message from Alice and they moved to the next message in the batch, Alice was still waiting for a Done message from Bob and Eve and was not joining the signing of the next message on time.

This is a graphical representation explaining the situation:
```
B: D _ D _ _ D _ _ _ _ D _ x _ _ _ _ _ _ D
E:   D _ D _ _ D _ _ _ _ D x _ _ _ _ _ _ _ D
A:                         D
```

Everyone is sending (retransmitting) their Done message at D. Bob and Eve receive Done message from Alice at x and they continue to the next message. Alice waits for the next 8 blocks to receive the Done message retransmission from Eve and 7 blocks to receive the Done message retransmission from Bob.

This problem has been solved by splitting `listen` and `exchange` functions into three pieces. The first piece is `listen` which is supposed to be called before the signing starts. The second piece is `signalDone` which is supposed to be called by all members participating in the given signing attempt when they are done. Finally, all members should call `waitUntilAllDone`. This approach is similar to the one used in the asynchronous state machine where we accept all types of messages all the time, even though we might not be interested in some of them at the given moment in time.

I have also added two unit tests to improve the coverage of the done check:
- cover the path when not everyone sends the done signal,
- cover the path when one signature does not match other signatures.

I tested this change locally with 4 clients over the night. They produced 8x10 signatures with no problems.

### DIFF
--- a/pkg/tbtc/signing_loop.go
+++ b/pkg/tbtc/signing_loop.go
@@ -56,24 +56,24 @@ type signingAnnouncer interface {
 // signingDoneCheckStrategy is a strategy that determines the way of signaling
 // a successful signature calculation across all signing group members.
 type signingDoneCheckStrategy interface {
-	exchange(
-		ctx context.Context,
-		memberIndex group.MemberIndex,
-		message *big.Int,
-		attemptNumber uint64,
-		attemptTimeoutBlock uint64,
-		attemptMembersIndexes []group.MemberIndex,
-		result *signing.Result,
-		endBlock uint64,
-	) (uint64, error)
-
 	listen(
 		ctx context.Context,
 		message *big.Int,
 		attemptNumber uint64,
 		attemptTimeoutBlock uint64,
 		attemptMembersIndexes []group.MemberIndex,
-	) (*signing.Result, uint64, error)
+	)
+
+	signalDone(
+		ctx context.Context,
+		memberIndex group.MemberIndex,
+		message *big.Int,
+		attemptNumber uint64,
+		result *signing.Result,
+		endBlock uint64,
+	) error
+
+	waitUntilAllDone(ctx context.Context) (*signing.Result, uint64, error)
 }
 
 // signingRetryLoop is a struct that encapsulates the signing retry logic.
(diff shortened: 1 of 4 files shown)


pkg/tbtc/signing_loop.go
```diff
@@ -316,64 +324,50 @@ func (srl *signingRetryLoop) start(
 				srl.attemptCounter,
 			)
 
-			latestEndBlock, err := srl.doneCheck.exchange(
+			err = srl.doneCheck.signalDone(
```
lukasz-zimnoch, line 327: Just making sure about the intention. Currently:
1. A participating member just broadcasts the produced signature through the done message and actually abandons it for a while
2. Then, it relies on the incoming done messages, including the message from itself, in order to extract the final signature that should be the same in all received done messages. 

If this is what we want, I think we are good. I was just wondering about the possible security implications of the fact that the given member does not compare incoming signatures against the one produced on its own. However, I don't see any problems at the first sight.

pdyraga: Yes, this is the intention. At first, I refactored the `signingDoneCheck` to be a bit more stateful and capture the signature, but the current approach looks cleaner in the code: everyone has to send, and all signatures must match. It's also important to note we accept only one message from the given sender.

lukasz-zimnoch (approved):

PR #3446, opened 2022-12-16
Wallet maintainer

tomaszslabon:
Depends on https://github.com/keep-network/keep-core/pull/3427.

This PR adds the wallet maintainer.


pkg/tbtc/maintainer/chain.go
```diff
@@ -11,3 +9,36 @@ type BitcoinDifficultyChain interface {
...
+
+	GetWalletInfo(walletPubKeyHash [20]byte) (
+		publicKeyBytes []byte,
+		mainUtxoHash [32]byte,
+		createdAt uint32,
+		err error,
+	)
+}
+
+// TODO: Reuse the tbtc.DKGState. It cannot be used now because of a import cycle.
```
lukasz-zimnoch, line 36: According to our [discussion](https://github.com/keep-network/keep-core/pull/3394#discussion_r1040943076), I don't think that's a good idea. That's clear we will have some small area of duplication between `pkg/maintainer` and `pkg/tbtc` but this is the cost of not having a direct dependency between them. That makes the whole package structure more flexible for future changes and modifications. Also, the real cost of those duplications is really low as they exist on the interface level (`maintainer.Chain` and `tbtc.Chain`) and the implementation (`pkg/chain/ethereum`) will be normally reused across both `maintainer` and `tbtc` packages.

lionakhnazarov:
Closing as stale. This long-open "Wallet maintainer" draft predates the current tBTC maintainer architecture and no longer applies to main.

PR #3454, opened 2022-12-22
Change SemVer range for `@tenderly/hardhat-tenderly` dependency

michalinacienciala:
We can't use `@tenderly/hardhat-tenderly` in versions `1.4.x` and higher on Node.js `v14`, as those `1.4.x` versions introduce dependency to `tslog` package, which is incompatible with `v14` of Node.js (works on `v16` or higher). We use Node `v14` in some of the projects having the `@keep-network/random-beacon` and `@keep-network/ecdsa` as their dependency.

There are also some problems with deployment scripts (possibly fixable) on `@tenderly/hardhat-tenderly` in versions `1.2.x`, so we're limiting the allowed versions to `<1.2.0`.

Refs:
https://github.com/threshold-network/solidity-contracts/pull/134
https://github.com/keep-network/tbtc-v2/pull/440
https://github.com/keep-network/coverage-pools/pull/222

### DIFF
--- a/solidity/random-beacon/package.json
+++ b/solidity/random-beacon/package.json
@@ -46,7 +46,7 @@
     "@nomiclabs/hardhat-etherscan": "^3.1.0",
     "@nomiclabs/hardhat-waffle": "^2.0.1",
     "@openzeppelin/hardhat-upgrades": "^1.20.0",
-    "@tenderly/hardhat-tenderly": ">=1.0.13 <1.2.0",
+    "@tenderly/hardhat-tenderly": "1.0.12",
     "@thesis-co/eslint-config": "github:thesis/eslint-config#v0.2.0",
     "@typechain/ethers-v5": "^9.0.0",
     "@typechain/hardhat": "^4.0.0",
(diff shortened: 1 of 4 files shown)


solidity/random-beacon/package.json
```diff
@@ -46,7 +46,7 @@
     "@nomiclabs/hardhat-etherscan": "^3.1.0",
     "@nomiclabs/hardhat-waffle": "^2.0.1",
     "@openzeppelin/hardhat-upgrades": "^1.20.0",
-    "@tenderly/hardhat-tenderly": "^1.0.12",
+    "@tenderly/hardhat-tenderly": "1.0.12",
```
r-czajkowski, line 49: Shouldn't we set range `>=1.0.13 <1.2.0` as in `ecdsa` instead of a fixed value?

r-czajkowski: nvm here is an explanation f8f74ca

r-czajkowski (approved):

PR #3459, opened 2023-01-03
RandomBeaconChaosnet: Simplify the contract

tomaszslabon:
Refs https://github.com/keep-network/keep-core/issues/3367

This PR simplifies the RandomBeaconChaosnet smart contract. The callback function of the  passed random beacon consumer is excuted directly, without any gas limit.


solidity/random-beacon/contracts/RandomBeaconChaosnet.sol
```diff
@@ -15,28 +15,19 @@
...
 ///         the real-world random beacon client implementation is ready.
-/// @dev Used for testing purposes only.
 contract RandomBeaconChaosnet is IRandomBeacon, Ownable {
-    using Callback for Callback.Data;
-
-    /// @notice Relay entry callback gas limit.
-    uint256 public constant _callbackGasLimit = 64_000;
-
     /// @notice Authorized addresses that can request a relay entry.
     mapping(address => bool) public authorizedRequesters;
```
pdyraga, line 25: Instead of mapping, I think we should pass `address authorizedRequester` in the constructor. This will be the address of `Bridge` contract proxy. We can make this field `immutable` which will lower the gas cost. We will then require `msg.sender` to be the requester.

tomaszslabon: Unfortunately, it seems `RandomBeaconChaosnet` is deployed before `WalletRegistry`. Therefore we don't know the address of `WalletRegistry` when we are deploying `RandomBeaconChaosnet`.

tomaszslabon: @pdyraga Should I replace the mapping with a single address, though?
This can be done.

pdyraga: You are right. Let's leave it as-is, it's a minor thing.

pdyraga (approved):

PR #3460, opened 2023-01-03
RandomBeaconChaosnet: update ECDSA deployment scripts

tomaszslabon:
Refs #3367

This PR: 
- introduces `useRandomBeaconChaosnet ` tag
- adds script for resolving `RandomBeaconChaosnet`
- adds script for upgrading `RandomBeaconChaosnset` in `WalletRegistry`
- adds script for authorizing `WalletRegistry` in `RandomBeaconChaosnet`
- adds script for transferring `RandomBeaconChaosnet` to governance (Threshold Council) 

The scripts are run only for networks tagged with `useRandomBeaconChaosnet`.


solidity/ecdsa/deploy/00_resolve_random_beacon.ts
```diff
@@ -10,12 +10,19 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
 
   const RandomBeacon = await deployments.getOrNull("RandomBeacon")
 
+  const isRandomBeaconNeeded = function () {
+    if (hre.network.tags.chaosnet) {
```
pdyraga, line 14: Where is this tag set and in what circumstances?

solidity/ecdsa/deploy/03_deploy_wallet_registry.ts
```diff
@@ -74,7 +74,7 @@ export default func
 func.tags = ["WalletRegistry"]
 func.dependencies = [
   "ReimbursementPool",
-  "RandomBeacon",
+  "RandomBeaconChaosnet",
```
pdyraga, line 77: Trying to confirm my understanding: this is important only for Goerli and local Ethereum instance (client dev, not Hardhat) because for unit tests (Hardhat), the random beacon is replaced with a smock instance here:
https://github.com/keep-network/keep-core/blob/a0817e53bc6ce1c85013ee254d24b88cc78b881a/solidity/ecdsa/test/utils/randomBeacon.ts#L14-L16

In other words, for unit tests, it doesn't matter if we have `RandomBeacon` or `RandomBeaconChaosnet` because it is always a mock.

What about system tests @nkuba @dimpar ?

solidity/ecdsa/deploy/00_resolve_random_beacon.ts
```diff
@@ -10,12 +10,19 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+  }
+
   if (RandomBeacon && helpers.address.isValid(RandomBeacon.address)) {
     log(`using existing RandomBeacon at ${RandomBeacon.address}`)
-  } else if (
-    !hre.network.tags.allowStubs ||
-    (hre.network.config as HardhatNetworkConfig)?.forking?.enabled
-  ) {
+  } else if (isRandomBeaconNeeded()) {
     throw new Error("deployed RandomBeacon contract not found")
```
pdyraga, line 26: This file has `func.tags = ["RandomBeacon"]` and no one uses this tag so we could actually... delete it?

solidity/ecdsa/deploy/12_upgrade_random_beacon.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  const RandomBeacon = await deployments.get("RandomBeacon")
+
+  await execute(
+    "WalletRegistryGovernance",
+    { from: governance, log: true, waitConfirmations: 1 },
+    "upgradeRandomBeacon",
+    RandomBeacon.address
+  )
+}
```
pdyraga, line 17: Why do we do it?

solidity/ecdsa/deploy/00_resolve_random_beacon.ts
tomaszslabon: I thought it would be passed during deployment.
But I can add it permanently to [mainnet configuration in hardhat.config.ts](https://github.com/keep-network/keep-core/blob/a0817e53bc6ce1c85013ee254d24b88cc78b881a/solidity/ecdsa/hardhat.config.ts#L129).

solidity/ecdsa/deploy/12_upgrade_random_beacon.ts
tomaszslabon: I did it because some unit tests fail if we don't use the real `RandomBeacon`.

tomaszslabon: It makes sense, although it also means that the real `RandomBeacon` will be needed during the chaosnet deployment - because it will have to be set in the `WalletRegistry`'s constructor.

pdyraga: I think it makes sense because right now it gets pretty complicated and I must confess I was confused looking at deployment scripts and trying to grasp the dependency graph.

The original deployment was with a real `RandomBeacon`. The final version of the deployment will be with a real `RandomBeacon`. All the `solidity/ecdsa` unit tests were either implemented for a real `RandomBeacon` or for the mocked instance.

If we replace the dependency for chaosnet deployments to `RandomBeaconChaosnet` in one final step, then we are doing the minimal number of modifications needed and we can easily unplug it later. It's not the most efficient approach in terms of deployment time but I think we can sacrifice it for less complexity. It also mimicks what we are going to do on mainnet.

To sum up, I think we need:
- The resolve script `00_resolve_random_beacon.ts`, always triggered.
- The resolve script `00_resolve_random_beacon_chaosnet.ts`, triggered only for chaosnet
- The upgrade script, `99_upgrade_random_beacon_chaosnet.ts`, triggered only for chaosnet

Additionally, we will tag as chaosnet the following networks in `hardhat.config.ts`:
- `goerli`
- `development`

Does it make sense?

I would also explain in a comment in `99_upgrade_random_beacon_chaosnet.ts` our reasoning - why we decided for an upgrade for chaosnet.

tomaszslabon: Done in b7035c52994d1179c98bcb46d84a4fb7d5da8a8d.

However, I have a few thoughts:
1) Shouldn't `mainnet` network be marked as chaosnet since the next deployment for mainnet should include `RandomBeaconChaosnet`?

2) Should we introduce the `chaosnet` tag to random-beacon deployments? The [script that deploys RandomBeaconChasonet](https://github.com/keep-network/keep-core/blob/a0817e53bc6ce1c85013ee254d24b88cc78b881a/solidity/random-beacon/deploy/09_deploy_random_beacon_chaosnet.ts#L14) could be run only for chasonet deployments.

solidity/ecdsa/deploy/00_resolve_random_beacon.ts
tomaszslabon: In the end `chasonet` was added to `development` and `goerli` networks.

solidity/ecdsa/deploy/03_deploy_wallet_registry.ts
tomaszslabon: Some unit tests must use the real `RandomBeacon`, but since we decided to use `RandomBeaconChaosnet` only for `goerli` and `development`, the tests will be ok.

solidity/ecdsa/deploy/00_resolve_random_beacon.ts
tomaszslabon: After recent changes the tag is used again [here](https://github.com/keep-network/keep-core/blob/c58aa96ec1da823d5345e588fa7f6202e6c24ad6/solidity/ecdsa/deploy/03_deploy_wallet_registry.ts#L77).

solidity/ecdsa/deploy/00_resolve_random_beacon_chaosnet.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+}
+
+export default func
+
+func.tags = ["RandomBeaconChaosnet"]
+
+// Only execute for chaosnet deployments.
+func.skip = async (hre: HardhatRuntimeEnvironment): Promise<boolean> =>
+  !hre.network.tags.chaosnet
```
pdyraga, line 30: I think "chaosnet" tag will be a common point of confusion. Chaosnet is our rollout program (https://blog.threshold.network/tbtc-v2-hits-its-first-launch-milestone/) and the network we currently have deployed on mainnet is the Chaosnet. 

See that in `hardhat.config.ts`, in this PR, we tag `development` and `goerli` networks as Chaosnet and it is not the same as our `mainnet` Chaosnet program.

Here is what we are trying to do:
We have `RandomBeaconChaosnet` to deploy and to use it from `WalletRegistry` (proxy) contract.
- We do not want to start a revolution right now and rework a bunch of unit tests in `ecdsa/solidity` that work with a real `RandomBeacon`. So for `hardhat` network, we should use the real `RandomBeacon`. 
- I do not know about system tests (cc @dimpar) but I expect they use a real `RandomBeacon` as well given the `RandomBeaconChaosnet` was introduced quite recently.
- For Goerli, Mainnet, and local development (local geth) networks we want to use `RandomBeaconChaosnet`.

If we want to use the tag, then I would name it `randomBeaconChaosnet` to avoid confusion and tag three networks: `development`, `goerli`, and `mainnet`.

If we do not want to use the tag, I would specify here the networks explicitly. But keep in mind we'll have to modify `00_resolve_random_beacon.ts` as well, so maybe the tag works better?

pdyraga: > If we want to use the tag, then I would name it `randomBeaconChaosnet` to avoid confusion and tag three networks: `development`, `goerli`, and `mainnet`.

Or maybe even `useRandomBeaconChaosnet` to be super-explicit about what is happening.

solidity/ecdsa/deploy/90_upgrade_random_beacon_chaosnet.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  const RandomBeaconChaosnet = await deployments.get("RandomBeaconChaosnet")
+
+  // Upgrade the random beacon smart contract in `WalletRegistry` to
+  // `RandomBeaconChaosnet`. This is a temporary solution to enable usage of
+  // `WalletRegistry` before the random beacon functionalities in the client
+  // are ready.
+  await execute(
+    "WalletRegistryGovernance",
+    { from: governance, log: true, waitConfirmations: 1 },
```
pdyraga, line 17: I do not think this will work for Goerli. The governance address is an external address, owned by the dev team. I think we should move this step earlier, before `WalletRegistryTransferGovernance`.

solidity/ecdsa/deploy/91_authorize_in_random_beacon_chaosnet.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export default func
+
+func.tags = ["WalletRegistryAuthorizeInBeaconChaosnet"]
+func.dependencies = ["UpgradeRandomBeaconChaosnet", "WalletRegistry"]
+
+// Only execute for chaosnet deployments.
+func.skip = async (hre: HardhatRuntimeEnvironment): Promise<boolean> =>
+  !hre.network.tags.chaosnet
```
pdyraga: Once we do it, let's transfer the ownership of `RandomBeaconChaosnet` to the `governance` (Threshold Council).

solidity/ecdsa/deploy/06_transfer_ownership_random_beacon_chaosnet.ts
```diff
@@ -0,0 +1,21 @@
+import type { HardhatRuntimeEnvironment } from "hardhat/types"
+import type { DeployFunction } from "hardhat-deploy/types"
+
+const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
+  const { getNamedAccounts, helpers } = hre
+  const { deployer, governance } = await getNamedAccounts()
+
+  await helpers.ownable.transferOwnership(
+    "RandomBeaconChaosnet",
+    governance,
```
tomaszslabon, line 10: @pdyraga 
Is this the correct address we are transferring to?

solidity/ecdsa/deploy/91_authorize_in_random_beacon_chaosnet.ts
tomaszslabon: Done in a7859770d0342956f7049a3a468fe74c181c3695.

solidity/ecdsa/deploy/90_upgrade_random_beacon_chaosnet.ts
tomaszslabon: In a7859770d0342956f7049a3a468fe74c181c3695 I moved upgrading `RandomBeaconChaosnet` just after `WalletRegistry` deployment.

solidity/ecdsa/deploy/06_transfer_ownership_random_beacon_chaosnet.ts
```diff
@@ -0,0 +1,21 @@
...
+  )
+}
+
+export default func
+
+func.tags = ["RandomBeaconChaosnetTransferOwnership"]
+func.dependencies = ["RandomBeaconChaosnet"]
+
+func.skip = async (hre: HardhatRuntimeEnvironment): Promise<boolean> =>
+  !hre.network.tags.useRandomBeaconChaosnet
```
pdyraga, line 21: We will have to execute this step and the previous one on mainnet during the deployment. Can we add `useRandomBeaconChaosnet` tag for the `mainnet` network in `hardhat.config.ts`?

tomaszslabon: Done in 3f949159d96a4aa599e63931a95c1917fc1e6823.

pdyraga (approved):

PR #3465, opened 2023-01-06
Adjust deployment scripts for testnet

lukasz-zimnoch:
Refs: https://github.com/keep-network/tbtc-v2/issues/456

Here we make some adjustments to increase the flexibility of testnet deployments. The main goal is to have a short test feedback cycle that will allow to check multiple scenarios using different combinations of contract's governance parameters. In order to achieve that, we must be able to change governable parameters quickly. This change sets a 60-second governance delay for testnet and passes the governance over testnet contracts to an address controlled by the dev team.

### DIFF
--- a/solidity/ecdsa/deploy/06_deploy_wallet_registry_governance.ts
+++ b/solidity/ecdsa/deploy/06_deploy_wallet_registry_governance.ts
@@ -7,8 +7,8 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
 
   const WalletRegistry = await deployments.get("WalletRegistry")
 
-  // 1 week for mainnet, 1 second for other test and dev networks.
-  const GOVERNANCE_DELAY = hre.network.name === "mainnet" ? 604800 : 1
+  // 60 seconds for Goerli. 1 week otherwise.
+  const GOVERNANCE_DELAY = hre.network.name === "goerli" ? 60 : 604800
 
   const WalletRegistryGovernance = await deployments.deploy(
     "WalletRegistryGovernance",
--- a/solidity/ecdsa/deploy/06_deploy_wallet_registry_governance.ts
+++ b/solidity/ecdsa/deploy/06_deploy_wallet_registry_governance.ts
@@ -7,7 +7,8 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
 
   const WalletRegistry = await deployments.get("WalletRegistry")
 
-  const GOVERNANCE_DELAY = 604800 // 1 week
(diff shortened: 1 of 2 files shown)


pdyraga (approved):

solidity/ecdsa/deploy/06_deploy_wallet_registry_governance.ts
```diff
@@ -7,7 +7,8 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
 
   const WalletRegistry = await deployments.get("WalletRegistry")
 
-  const GOVERNANCE_DELAY = 604800 // 1 week
+  // 1 week for mainnet, 1 second for other test and dev networks.
+  const GOVERNANCE_DELAY = hre.network.name === "mainnet" ? 604800 : 1
```
pdyraga, line 11: Ah, wait - I think we should have another value here for Hardhat network. Maybe `60` sec only for Goerli?

pdyraga (changes requested):

solidity/ecdsa/deploy/06_deploy_wallet_registry_governance.ts
lukasz-zimnoch: Why? Aren't we interested in effectively getting rid of the governance delay for all non-mainnet networks?

lukasz-zimnoch: Ah yeah, tests. Will change

lukasz-zimnoch: See b838911

pdyraga (approved):

solidity/ecdsa/deploy/06_deploy_wallet_registry_governance.ts
```diff
@@ -7,7 +7,8 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
 
   const WalletRegistry = await deployments.get("WalletRegistry")
 
-  const GOVERNANCE_DELAY = 604800 // 1 week
+  // 60 seconds for Goerli. 1 week otherwise.
+  const GOVERNANCE_DELAY = hre.network.name === "goerli" ? 60 : 604800
```
pdyraga, line 11: Not in this PR but I propose we do the same for the beacon for consistency:

https://github.com/keep-network/keep-core/blob/ff77e6eefdd9123da3f72388b1f470a444dd7346/solidity/random-beacon/deploy/07_deploy_random_beacon_governance.ts#L10

Issue #3456, opened 2022-12-28
DKG trigger and chain reorgs


lukasz-zimnoch:
Currently, the tECDSA DKG is triggered using a simple event listener acting on `DKGStarted` events. However, small chain reorgs may cause different `startBlock` to be received by DKG participants. That can lead to invalid signatures over the produced DKG result as the `startBlock` is one of the signature's components. We need to improve that. See [discussion](https://github.com/keep-network/keep-core/pull/3427#discussion_r1056132574)

pdyraga:
An early idea on how we could approach it...

Things that may happen:
- The block is mined and `DKGStarted` event is emitted. Given the chain reorg, the block is dropped and DKG start transaction is not mined again. In this scenario, the client should not start the DKG.
- The block with `DKGStarted` event is emitted. The event is emitted twice with the same block number given the specifics of Ethereum gateway deployment. In this scenario, the client should ignore the second event.
- The block with `DKGStarted` event is emitted. The event is emitted twice with the same block number given the [background event monitoring loop](https://github.com/keep-network/keep-common/pull/63). In this scenario, the client should ignore the second event.
- The block with `DKGStarted` event is mined. Given the chain reorg, the block is dropped. A moment later, another block with `DKGStarted` event is mined. Two events were emitted in this scenario but the client should take into account only the second one.

Here is how can we approach it:
- `DKGStarted` event is emitted at block `N`. We start the monitoring goroutine and wait for 20 blocks.
- After 20 blocks, we check the DKG state. If it is not in progress, we leave.
- If DKG is in progress we take the last `DKGStarted` event emitted. 
- We start the DKG based on the block number from the latest `DKGStarted` event.

In this approach, it may happen that the latest `DKGStarted` event has only 15 confirmations, for example. I think this is fine.

If we want to be more restrictive, we could:
- `DKGStarted` event is emitted at block `N`. We start the monitoring goroutine and wait for 20 blocks.
- After 20 blocks, we check the DKG state. If it is not in progress, we leave.
- If DKG is in progress we take the last `DKGStarted` event emitted. 
- We wait until the event has 20 confirmations.
- Optional: We periodically double-check if DKG is in progress to not wait forever if 20 confirmations never happen because the block with the event was dropped.
- We start the DKG based on the block number from the latest `DKGStarted` event.

Issue #3462, opened 2023-01-04
DKG improvements to the challenge mechanism


pdyraga:

lionakhnazarov:
Closing as stale. Targets the legacy keep-core v1 (random beacon / KEEP staking) architecture, superseded by the Threshold v2 stack. Not actionable against current main.

PR #3468, opened 2023-01-10
Confirm DKG start before triggering off-chain protocol

lukasz-zimnoch:
Closes: #3456 

So far, the tECDSA DKG was triggered using a simple event listener acting on `DKGStarted` events. However, small chain reorgs may cause different `startBlock` to be received by DKG participants. That can lead to invalid signatures over the produced DKG result as the `startBlock` is one of the signature's components. Here we improve that by adding a confirmation mechanism that delays the start of the off-chain protocol until the event and DKG on-chain state is confirmed. This way clients are sure to use a finalized event that will no longer be the subject of chain reorgs.


pkg/tbtc/dkg.go
```diff
@@ -224,13 +231,15 @@ func (de *dkgExecutor) setupBroadcastChannel(
...
-// publishes the result to the chain.
+// publishes the result to the chain. The execution can be delayed by an
+// arbitrary number of blocks using the delayBlocks argument.
 func (de *dkgExecutor) generateSigningGroup(
 	dkgLogger *zap.SugaredLogger,
 	seed *big.Int,
 	memberIndexes []uint8,
 	groupSelectionResult *GroupSelectionResult,
 	startBlock uint64,
+	delayBlocks uint64,
```
pdyraga, line 242: Does it make sense to pass `startBlock` and `delayBlocks` separately? 

The only reason for that is to log `zap.Uint64("attemptStartBlock", attempt.startBlock)` but I think we could treat the `attemptStartBlock` as the event block with the confirmation delay added. In fact, this is when the attempt starts.

It would also allow us to avoid this indirection: In `pkg/tbtc/tbtc.go` we react on the event and wait for `dkgStartedConfirmationBlocks`. If everything checks, we dispatch the action to `node.go` making the `dkgStartedConfirmationBlocks` an abstract `delayBlocks`. But then, we go back to `pkg/tbtc` and we no longer use `dkgStartedConfirmationBlocks` but an abstract `delayBlocks`. It is like jumping out of the context for a while just to lose some details.

Instead, we could use `confirmationBlock` in `pkg/tbtc/tbtc.go` when calling `node.joinDKGIfEligible` as a `startBlock` (instead of `lastEvent.BlockNumber`).

pkg/tbtc/tbtc.go
```diff
@@ -146,16 +146,69 @@ func Initialize(
 				return
```
pdyraga, line 146: One thing I realized when looking at the line earlier:
```
logger.Warnf(
					"DKG started event with seed [0x%x] and "+
						"starting block [%v] has been already processed",
					event.Seed,
					event.BlockNumber,
				)
```

I am not sure if this should be a `Warning`. The background event polling loop (see https://github.com/keep-network/keep-common/pull/63) will fetch this event multiple time, this is an expected behavior. I would consider changing the log level to `Info`.

pdyraga: Also, the "starting block" part is confusing. I've got confused for a couple of minutes after reading this log message thinking we take a combo of seed+block number to duplicate events which would lead to duplicate keygen member processes being triggered.

Can we drop the `starting block [%v]` part?

This is a similar comment to [this one](https://github.com/keep-network/keep-core/pull/3427#discussion_r1061383154) made during the review of #3427.

pkg/tbtc/tbtc.go
```diff
@@ -146,16 +146,69 @@ func Initialize(
...
+					&DKGStartedEventFilter{
+						StartBlock: event.BlockNumber - dkgStartedConfirmationBlocks,
+					},
+				)
+				if err != nil {
+					logger.Errorf("failed to get past DKG started events: [%v]", err)
+					return
+				}
+
+				lastEvent := pastEvents[len(pastEvents)-1]
```
pdyraga, line 184: It should not happen but... can we validate this slice is not empty?

pkg/tbtc/tbtc.go
```diff
@@ -146,16 +146,69 @@ func Initialize(
...
+
+			dkgState, err := chain.GetDKGState()
+			if err != nil {
+				logger.Errorf("failed to check DKG state: [%v]", err)
+				return
+			}
+
+			if dkgState == AwaitingResult {
+				// Fetch all past DKG started events starting from one
+				// confirmation period before the original event's block.
```
pdyraga, line 173: And we do it because if there was a reorg, the event we received could be moved to a block with a lower number than the one we received, correct? Can we clarify it in this comment?

pkg/tbtc/tbtc.go
```diff
@@ -146,16 +146,69 @@ func Initialize(
...
+					dkgStartedConfirmationBlocks,
+				)
+			} else {
+				logger.Infof(
+					"DKG started event with seed [0x%x] and starting "+
+						"block [%v] was not confirmed",
+					event.Seed,
+					event.BlockNumber,
+				)
+			}
```
pdyraga, line 211: Not blocking this PR but the entire content of `OnDKGStarted` is getting quite complicated. We could consider extracting it to a separate function in `tbtc.go` to unit-test it. Maybe a separate PR?

pdyraga (commented):
I can't test this code given #3460 is not yet finished and I am struggling to set up the environment. If you performed tests locally @lukasz-zimnoch I will trust your verification and merge the PR once the code comments are addressed. We will deploy it to testnet for another verification as well.

pkg/tbtc/dkg.go
lukasz-zimnoch: Can you clarify this part:
```
But then, we go back to pkg/tbtc and we no longer use dkgStartedConfirmationBlocks but an abstract delayBlocks. It is like jumping out of the context for a while just to lose some details.
```

Also, the whole story is more complicated. Actually, we need to do two things at the same time:
1. Keep the original start block which must be the same as stored on the contract side. This is a must to produce valid signatures.
2. Delay the off-chain protocol start by the confirmation period

A single start block argument cannot satisfy both requirements. Using `confirmationBlock` based on a uncertain event (the one we confirm) cannot do the trick as the real start block held by the contract may be different. We need to base on the last event we fetch after the confirmation period. In the same time, we need to delay the off-chain protocol execution by the confirmation period length to not fail on the first attempt.

pkg/tbtc/dkg.go
```diff
@@ -224,13 +231,15 @@ func (de *dkgExecutor) setupBroadcastChannel(
 
 // generateSigningGroup executes off-chain protocol for each member controlled
 // by the current operator and upon successful execution of the protocol
-// publishes the result to the chain.
+// publishes the result to the chain. The execution can be delayed by an
+// arbitrary number of blocks using the delayBlocks argument.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3468/commits/f2c9ff81685d629a0357e462527111bddc644950

pkg/tbtc/node.go
```diff
@@ -143,9 +143,14 @@ func (n *node) operatorID() (chain.OperatorID, error) {
 // distributed key generation if this node's operator proves to be eligible for
 // the group generated by that seed. This is an interactive on-chain process,
 // and joinDKGIfEligible can block for an extended period of time while it
-// completes the on-chain operation.
-func (n *node) joinDKGIfEligible(seed *big.Int, startBlock uint64) {
-	n.dkgExecutor.executeDkgIfEligible(seed, startBlock)
+// completes the on-chain operation. The execution can be delayed by an
+// arbitrary number of blocks using the delayBlocks argument.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3468/commits/f2c9ff81685d629a0357e462527111bddc644950

pkg/tbtc/dkg.go
```diff
@@ -104,10 +108,12 @@ func (de *dkgExecutor) preParamsCount() int {
 // executeDkgIfEligible is the main function of dkgExecutor. It performs the
 // full execution of ECDSA Distributed Key Generation: determining members
 // selected to the signing group, executing off-chain protocol, and publishing
-// the result to the chain.
+// the result to the chain. The execution can be delayed by an arbitrary number
+// of blocks using the delayBlocks argument.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3468/commits/f2c9ff81685d629a0357e462527111bddc644950

pkg/tbtc/tbtc.go
lukasz-zimnoch: See 02888efc

pkg/tbtc/tbtc.go
lukasz-zimnoch: See 90955c80

pkg/tbtc/tbtc.go
lukasz-zimnoch: See e89b0059

pkg/tbtc/dkg.go
pdyraga: I see it now! I missed the usage of `startBlock` in `generateSigningGroup` when calling `de.publishDkgResult`. 

It all makes sense. Can we add this last sentence to `generateSigningGroup` documentation to make it clear:
```
// (...)
// confirming blocks - before executing the off-chain action. Note that the
// startBlock represents the block at which DKG started on-chain. This is
// important for the result submission.
```

pkg/tbtc/tbtc.go
lukasz-zimnoch: Sure! I can do it during the work on https://github.com/keep-network/keep-core/issues/3462

pkg/tbtc/dkg.go
lukasz-zimnoch: See [890f7e4](https://github.com/keep-network/keep-core/pull/3468/commits/890f7e4c1871792d37ef3115bd6558cdcf15bbe7)

lukasz-zimnoch:
> I can't test this code given #3460 is not yet finished and I am struggling to set up the environment. If you performed tests locally @lukasz-zimnoch I will trust your verification and merge the PR once the code comments are addressed. We will deploy it to testnet for another verification as well.

I have tested the basic scenario: request a new wallet to trigger DKG, wait until the confirmation period elapses, and check if nodes executed the protocol as expected. The corner cases involving reorgs and event duplicates must be checked on testnet

pdyraga (approved):

Issue #3461, opened 2023-01-04
DKG result validation before submission


pdyraga:
We could validate the result with a call to `EcdsaDkgValidator` contract before submitting it. This way, we would eliminate the risk of slashing the operator in case of a bug in the client. The result assembled by the client *should* be valid but double-checking does not hurt.

PR #3469, opened 2023-01-10
Validate submitted DKG result

lukasz-zimnoch:
Closes: #3461

Here we add an on-chain validation of the submitted DKG result. This way, honest clients may abort the submission of an invalid result and avoid slashing.

### DIFF
--- a/pkg/tbtc/dkg_submit_test.go
+++ b/pkg/tbtc/dkg_submit_test.go
@@ -294,6 +294,10 @@ func TestSubmitResult_MemberSubmitsResult(t *testing.T) {
 	ctx, cancelCtx := context.WithCancel(context.Background())
 	defer cancelCtx()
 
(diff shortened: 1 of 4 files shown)


pkg/tbtc/dkg_submit_test.go
```diff
@@ -294,6 +294,10 @@ func TestSubmitResult_MemberSubmitsResult(t *testing.T) {
 	ctx, cancelCtx := context.WithCancel(context.Background())
 	defer cancelCtx()
 
+	if err = localChain.setDKGResultValidity(true); err != nil {
+		t.Fatal(err)
+	}
```
pdyraga, line 299: Another - and maybe more intuitive? - option would be to treat the result as valid by default. We could set `dkgResultValid` in `Connect` / `ConnectWithKey`. Not a big deal though.

pdyraga (approved):

PR #3470, opened 2023-01-10
RandomBeaconChaosnet: update random beacon deployment scripts

tomaszslabon:
#depends on #3460.

This PR:
- adds the chaosnet tag
- updates the deployment script for RandomBeaconChaosnet - the script will only be executed for chaos net deployments


solidity/random-beacon/deploy/09_deploy_random_beacon_chaosnet.ts
```diff
@@ -33,3 +33,7 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
 export default func
```
tomaszslabon, line 34: @pdyraga
Just wanted to point out that we don't transfer ownership of `RandomBeaconChasonet` to any governance.
After the deployment the contract will still be owned by the deployer.
This is different compared to the real `RandomBeacon` which is owned by a governance.

solidity/random-beacon/deploy/09_deploy_random_beacon_chaosnet.ts
```diff
@@ -33,3 +33,7 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
...
 
 func.tags = ["RandomBeaconChaosnet"]
+
+// Only execute for chaosnet deployments.
```
pdyraga, line 37: I think "chaosnet" tag will be a common point of confusion. Chaosnet is our rollout program (https://blog.threshold.network/tbtc-v2-hits-its-first-launch-milestone/) and the network we currently have deployed on mainnet is the Chaosnet.

In this PR, in `hardhat.config.ts` we flip it: everything but `mainnet` network is a Chaosnet.

What is the reason for introducing the `chaosnet` tag? I think we want to deploy `RandomBeaconChaosnet` for all networks.

solidity/random-beacon/deploy/10_transfer_governance_random_beacon_chaosnet.ts
```diff
@@ -0,0 +1,22 @@
+import type { HardhatRuntimeEnvironment } from "hardhat/types"
+import type { DeployFunction } from "hardhat-deploy/types"
+
+const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
+  const { getNamedAccounts, helpers } = hre
+  const { deployer, chaosnetOwner } = await getNamedAccounts()
+
+  await helpers.ownable.transferOwnership(
+    "ReimbursementPool",
```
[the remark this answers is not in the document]
tomaszslabon: I thought we will prepare yet another PR where we will update the ecdsa scripts so that the `chaosnetOwner` does the  authorising.

tomaszslabon:
> Question in terms of the order of PRs: my understanding is that this PR should be merged _before_ #3460 given we need to be able to resolve the `RandomBeaconChaosnet` in `00_resolve_random_beacon_chaosnet.ts`. Is this correct?

I don't think so. Here we are just adding a tag. We can merge them both.

solidity/random-beacon/deploy/09_deploy_random_beacon_chaosnet.ts
tomaszslabon: In the ecdsa deployment script we use the `chaosnet` tag as well.
Wouldn't it be even more confusing if in `random-beacon` we always execute `RandomBeaconChaosnet`-related scripts while in ecdsa we do that conditionally (when `chaosnet` tag is set)?

solidity/random-beacon/deploy/10_transfer_governance_random_beacon_chaosnet.ts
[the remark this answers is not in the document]
pdyraga: The `chaosnetOwner` role should be used only for Sortition Pools to [add beta operators and deactivate chaosnet phase one day](https://github.com/keep-network/sortition-pools/blob/5d462f7382f7a2811026aa75f80b5f4edca806a7/contracts/Chaosnet.sol#L35-L38). The ownership of the `RandomBeaconChaosnet` should be transferred to the `governance` role. 

On Goerli, the `governance` role should be transferred to the dev team (see #3465 - it is doing it only for `ecdsa` but we should also do it for the `beacon`). The dev team account is not the deployer account. If we transfer the ownership here, we will not be able to authorise the wallet registry later.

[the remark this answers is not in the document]
tomaszslabon: So if we don't need the `chaosnet` and we don't do the ownership transfer then this PR can be closed.

tomaszslabon:
Closing the PR: the chaosnet tag is not needed and ownership transfer will be done in ecdsa deployment scripts.

solidity/random-beacon/deploy/09_deploy_random_beacon_chaosnet.ts
pdyraga: Let's take a step back because I think we are confused about what we are trying to achieve. 

We have `RandomBeaconChaosnet` to deploy and to use it from `WalletRegistry` (proxy) contract.
- We do not want to start a revolution right now and rework a bunch of unit tests in `ecdsa/solidity` that work with a real `RandomBeacon`. So for `hardhat` network, we should use the real `RandomBeacon`. 
- I do not know about system tests (cc @dimpar) but I expect they use a real `RandomBeacon` as well given the `RandomBeaconChaosnet` was introduced quite recently.
- For Goerli, Mainnet, and local development (local geth) networks we want to use `RandomBeaconChaosnet`.

I do not think it hurts if we deploy `RandomBeaconChaosnet` for Hardhat network, it's just a few milliseconds. 

Regarding `solidity/ecdsa` contracts, I would make a rule that for all networks but `hardhat` we use the `RandomBeaconChaosnet`. Let's just check what to do about system tests.

Instead, we want to deploy `RandomBeaconChaosnet` and wire it up with `WalletRegistry`

PR #3474, opened 2023-01-13
Add governance and chaosnetOwner addresses for goerli

tomaszslabon:
This PR adds the missing addresses for `governance` and `chaosnetOwner` in random-beacon hardhat configuration.
The addresses are the same as in ecdsa configuration.

### DIFF
--- a/solidity/ecdsa/hardhat.config.ts
+++ b/solidity/ecdsa/hardhat.config.ts
@@ -146,17 +146,17 @@ const config: HardhatUserConfig = {
     },
     governance: {
       default: 2,
-      goerli: "0xCac19049825F370dB0836cB0d8E4D024F78eb2eB", // Dev team
+      goerli: 0,
       mainnet: "0x9f6e831c8f8939dc0c830c6e492e7cef4f9c2f5f", // Threshold Council
     },
     chaosnetOwner: {
       default: 3,
-      goerli: "0xCac19049825F370dB0836cB0d8E4D024F78eb2eB", // Dev team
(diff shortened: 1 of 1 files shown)


solidity/random-beacon/hardhat.config.ts
```diff
@@ -134,12 +134,12 @@ const config: HardhatUserConfig = {
...
     governance: {
       default: 2,
-      goerli: 0,
+      goerli: "0xCac19049825F370dB0836cB0d8E4D024F78eb2eB", // Dev team
       mainnet: "0x9f6e831c8f8939dc0c830c6e492e7cef4f9c2f5f", // Threshold Council
     },
     chaosnetOwner: {
       default: 3,
-      goerli: 0,
+      goerli: "0xCac19049825F370dB0836cB0d8E4D024F78eb2eB", // Dev team
```
pdyraga, line 142: Is this the address set for `ecdsa` as well? How do we add operators as beta operators when deploying code to the test cluster? Is it happening automatically? If so, this will fail.

tomaszslabon: Yes, this is the same address as in ecdsa.

michalinacienciala (dismissed):

solidity/random-beacon/hardhat.config.ts
tomaszslabon: After discussing it, we decided to set all the addresses for Goerli to `0`.

michalinacienciala (approved):

PR #3476, opened 2023-01-17
Add gas limit margin for approve DKG result calls

lukasz-zimnoch:
The original estimate made by the `go-ethereum` library for the `WalletRegistry.approveDkgResult` turned out to be too low. The root cause of the problem is probably related with the reimbursement pool. To overcome that problem, we add a 20% margin to the original gas limit estimate.

### DIFF
--- a/pkg/chain/ethereum/tbtc.go
+++ b/pkg/chain/ethereum/tbtc.go
@@ -7,6 +7,7 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"fmt"
+	"github.com/keep-network/keep-common/pkg/chain/ethereum/ethutil"
 	"math/big"
 	"reflect"
 	"sort"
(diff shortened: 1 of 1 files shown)


pkg/chain/ethereum/tbtc.go
```diff
@@ -859,8 +860,22 @@ func (tc *TbtcChain) ChallengeDKGResult(dkgResult *tbtc.DKGChainResult) error {
...
-		convertDkgResultToAbiType(dkgResult),
+	result := convertDkgResultToAbiType(dkgResult)
+
+	gasEstimate, err := tc.walletRegistry.ApproveDkgResultGasEstimate(result)
+	if err != nil {
+		return err
+	}
+
+	// The original estimate for this contract call turned out to be too low.
+	// Here we add a 20% margin to overcome out of gas problems.
```
tomaszslabon, line 871:
```suggestion
	// Here we add a 20% margin to overcome the gas problems.
```

lukasz-zimnoch: See c54f17f

tomaszslabon (approved):

dimpar (approved):

PR #3494, opened 2023-02-03
Updating SECURITY.txt to the common Threshold version

pdyraga:
Depends on https://github.com/threshold-network/solidity-contracts/pull/137

This version is the same as one published in `threshold-network/solidity-contracts `repository. It may need further updates when Immunefi Bug Bounty program is established. If you have any comments to the Threshold version of `SECURITY.txt` please comment at https://github.com/threshold-network/solidity-contracts/pull/137 and I'll backport changes here.

### DIFF
--- a/SECURITY.adoc
+++ b/SECURITY.adoc
@@ -36,7 +36,7 @@ The following attacks are excluded from the Bug Bounty program:
 The following activities are prohibited by this bug bounty program:
 
 - Any testing with mainnet or public testnet contracts; all testing should be done on private testnets.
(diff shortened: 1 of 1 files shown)


nkuba (approved):

SECURITY.adoc
```diff
@@ -1,151 +1,46 @@
-= Security Policy
+# Security Policy
```
nkuba, line 1: We should keep using `=` for section levels in adoc.

PR #3510, opened 2023-03-03
Document contracts

michalinacienciala:
We're modifying the index page of our documentation to link to the pages documenting contracts. So far we've got only TBTC-v2 contracts documentation in the works, but soon we'll document contracts from other projects.


michalinacienciala:
Closed in favor of https://github.com/keep-network/keep-core/pull/3534

Issue #3512, opened 2023-03-06
Sweeping prototype


pdyraga:
The first implementation of the sweeping mechanism should skip the risky coordination part and be triggered manually. All security guarantees should still hold.

Selected addresses will be able to trigger sweeping from a smart contract specifying how many inputs should be swept. The inputs will be swept from the oldest to the newest. The current group coordinator will receive an event from a contract, select inputs to be swept, and send a sweeping command to the group's broadcast channel. Every signing group member will validate the command and if everything checks, begin sweeping.

This plan requires:
- [ ] Revisiting [RFC-7](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-7.adoc)
- [ ] Implementing and deploying a temporary sweep coordination contract.
- [ ] Implementing Go client code for sweeping.
- [ ] Implementing a temporary command for submitting SPV proofs. We would like to avoid the time cost of implementing a full SPV maintainer, if possible. If it makes sense (to be decided during the implementation), we could expose a command in `tbtc-v2.ts` library to submit an SPV proof manually, given the action is triggered manually anyway.
- [ ] Performing extensive tests on Goerli and Bitcoin testnet.

PR #3520, opened 2023-04-11
Client-side deposit sweep: Observe incoming proposals

lukasz-zimnoch:

Here is the first part of the client-side deposit sweep saga. In this PR, we are focusing on the integration with the on-chain `WalletCoordinator` contract introduced by https://github.com/keep-network/tbtc-v2/pull/575 and we are introducing an ability to observe incoming deposit sweep proposals. Here is a brief description of particular changes.

### `WalletCoordinator` contract integration code

We are adding all the boilerplate code required to talk with the `WalletCoordinator` contract. This work includes:
- Changes in the configuration mechanism allowing to point a specific deployment of the `WalletCoordinator` contract
- Contract bindings allowing interaction with the contract
- Some Makefile tricks to resolve name clashes in generated bindings due to a bug in the currently used version of abigen
- Integration with the `tbtc` package through the `tbtc.WalletCoordinatorChain` interface and `chain/ethereum` implementation

### Observe incoming proposals

This part uses the code described in the previous section to:
- Handle incoming deposit sweep proposals submitted to the `WalletCoordinator` contract
- De-duplicate incoming proposals
- Confirm the finality of the incoming proposals
- Temporarily log about this fact 

### Next steps

After receiving a confirmed proposal, clients should validate it. This work will be covered by a next PR: https://github.com/keep-network/keep-core/pull/3521


Dockerfile
```diff
@@ -17,6 +17,7 @@ RUN apk add --update --no-cache \
 	nodejs \
 	npm \
 	bash \
+    perl \
```
pdyraga, line 20: image

?

pkg/tbtc/chain.go
```diff
@@ -196,6 +198,59 @@ type HeartbeatRequestedEvent struct {
...
+	ProposalSubmitter chain.Address
+	BlockNumber       uint64
+}
+
+// DepositSweepProposalSubmittedEventFilter is a component allowing to
+// filter DepositSweepProposalSubmittedEvent.
+type DepositSweepProposalSubmittedEventFilter struct {
+	StartBlock          uint64
+	EndBlock            *uint64
+	ProposalSubmitter   []chain.Address
```
pdyraga, line 250: Why this is an array and not a single value?

lukasz-zimnoch: Contract binding allows passing an array as a filter to get proposals submitted by a given list of submitters. Here we are just externalizing this ability.

lukasz-zimnoch: Tested it and works as expected. Here is the explanation:

1. Suppose we use a slice `ProposalSubmitter` with two items, "PS1" and "PS2".
2. We call `PastDepositSweepProposalSubmittedEvents` method. Under the hood, we reach `FilterLogs` (https://github.com/ethereum/go-ethereum/blob/23bee16208718975f9b9e012949b8d4ee1223108/accounts/abi/bind/base.go#L407).
3. Within `FilterLogs`, a topic set is constructed (https://github.com/ethereum/go-ethereum/blob/23bee16208718975f9b9e012949b8d4ee1223108/accounts/abi/bind/base.go#L413). Basically, this is the event ID, followed with our `ProposalSubmitter` slice. This gives the `{{ID}, {PS1, PS2}}` pattern. Here is a playground demostrating it: https://go.dev/play/p/l1zW2gNAI0J
4. According to the section you mentioned (https://github.com/ethereum/go-ethereum/blob/23bee16208718975f9b9e012949b8d4ee1223108/interfaces.go#L172), the `{{ID}, {PS1, PS2}}` pattern translates to: "matches topic (ID) in first position AND (PS1 OR PS2) in second position". In simple words this is: "matches the DepositSweepProposalSubmittedEvent AND (PS1 OR PS2) as submitters"

pkg/tbtc/deduplicator.go
```diff
@@ -90,3 +99,37 @@ func (d *deduplicator) notifyDKGResultSubmitted(
...
+	for _, depositKey := range newProposal.DepositsKeys {
+		buffer.Write(depositKey.FundingTxHash[:])
+		fundingOutputIndex := make([]byte, 4)
+		binary.BigEndian.PutUint32(fundingOutputIndex, depositKey.FundingOutputIndex)
+		buffer.Write(fundingOutputIndex)
+	}
+	buffer.Write(newProposal.SweepTxFee.Bytes())
+
+	bufferSha256 := sha256.Sum256(buffer.Bytes())
+	cacheKey := hex.EncodeToString(bufferSha256[:])
```
pdyraga, line 123: It makes sense that the deduplicator is doing only what it is supposed to be doing: deduplicating an event, without going deeper and not trying to take the responsibility of wallet locking.

My only comment to this code is that the hashing logic should live next to `type DepositSweepProposal struct`. I think we should have some sort of `HashCode()` function (like in JVM world) defined next to the struct and use this function here. This way, if we add one more field to the struct, it will be easier to notice that the hashing must be updated. Also, there is a chance we'll reuse the `HashCode()` in other places.

I think all the tests in `deduplicator_test.go` should stay as-is, I'd just add some tests for `HashCode()`.

lukasz-zimnoch: I was thinking about it but that depends on how we look at the "hash code" function. I'm rather looking at it as an internal deduplication-specific thing, hence why it is placed here. My reasoning: the deduplicator must decide what is and what is not a duplicate, hence it must build the cache key accordingly. Having a general "hash code" function exposed by `DepositSweepProposal` may be dangerous as their connection with the deduplicator will be not obvious and we can mess up things by adding/removing fields.

pkg/chain/ethereum/tbtc/gen/Makefile
```diff
@@ -1,6 +1,27 @@
...
+# for us because both Bridge and WalletCoordinator contracts use BitcoinTx.Info
+# struct which is then re-declared in the same package once abigen does its work.
+# An ultimate solution would be upgrading go-ethereum (thus abigen too) to v1.11 but
+# that version contains some breaking changes that make the upgrade non-trivial.
+# As a short-term workaround, we use some Makefile shenanigans to slightly rename
+# the conflicting BitcoinTxInfo struct in the WalletCoordinator output file.
+# We use perl for that purpose as sed is not cross-platform and works a bit
+# differently on GNU and BSD.
+#
+# TODO: Remove once go-ethereum is upgraded to v1.11.
```
pdyraga, line 18: Do we have an issue on GitHub for that work? If not, can you please add it and link it from here?

pkg/tbtc/deduplicator.go
lukasz-zimnoch: See 47f9f482

pkg/chain/ethereum/tbtc.go
```diff
@@ -996,3 +1031,148 @@ func (tc *TbtcChain) activeWalletPublicKey() ([]byte, bool, error) {
...
+
+	convertedEvents := make([]*tbtc.DepositSweepProposalSubmittedEvent, 0)
+	for _, event := range events {
+		// If the wallet PKH filter is set, omit all events that target
+		// different wallets.
+		if walletPublicKeyHash != [20]byte{} {
+			if event.Proposal.WalletPubKeyHash != walletPublicKeyHash {
+				continue
+			}
+		}
```
pdyraga, line 1094: Is it needed at all? The filter - if set - should not return events with wallets with a different key.

It's the same situation as with the proposal submitter filter.

lukasz-zimnoch: The main difference is that `walletPublicKeyHash` is not indexed as this is part of the proposal struct. We need to do that filtering manually and this is what we do here.

pkg/tbtc/tbtc.go
```diff
@@ -312,6 +312,109 @@ func Initialize(
...
+				lastEvent := pastEvents[len(pastEvents)-1]
+
+				// TODO: Validate the proposal, assemble, sign and
+				//       broadcast the sweep tx.
+				logger.Infof(
+					"received deposit sweep proposal [%+v]",
+					lastEvent.Proposal,
+				)
+			} else {
+				logger.Infof(
```
pdyraga, line 408: Could be worth logging the reason: what were the cause and expiration. I don't think we should have it in the logs long-term but for a few weeks/months after we launch sweeps, this could be helpful.

pkg/tbtc/tbtc.go
```diff
@@ -312,6 +312,109 @@ func Initialize(
...
+				if len(pastEvents) == 0 {
+					logger.Errorf(
+						"no past deposit sweep proposal events for "+
+							"wallet PKH [0x%x]",
+						walletPublicKeyHash,
+					)
+					return
+				}
+
+				lastEvent := pastEvents[len(pastEvents)-1]
```
pdyraga, line 399: I am not sure if this is correct in all cases:
- If there the same event was moved +- one block after the reorg, the deduplicator will handle it.
- If another event was emitted on a fork, the client will receive two events and they will pass the deduplicator. Two goroutines will be triggered. The goroutine from the earliest event should give up to let the "more recent" goroutine do its confirmation handling.

I believe the logic - as currently implemented - allows the "older" goroutine to proceed even though the most recent event on the chain is not sufficiently confirmed yet.

pkg/chain/ethereum/tbtc.go
pdyraga: Ah, I see now. `walletPublicKeyHash` is not passed to `PastDepositSweepProposalSubmittedEvents`. Makes sense.

pkg/tbtc/tbtc.go
pdyraga: We discussed it with @lukasz-zimnoch synchronously and came up with a conclusion that we are going to remove the last event check and leave the rest of the logic as-is. 

The reasoning behind `WalletCoordinator` is that it is just a coordination point based on the chain consensus. If enough clients received the event, they should follow it and execute a signature. All input parameters for that signature are validated, so even if there was a reorg and another event will land on the canonical chain later, the first signature will still be valid and approved by Bitcoin. The only reason we want to wait for 20 blocks after the coordination even is received is to make sure the right *type* of action is executed given different *types* of actions may have different lock times. We do not want to run into a situation when the majority of clients execute sweep with N blocks wallet lock time and the chain has M < N blocks wallet lock time because the canonical chain - as a result of a reorg - is supposed to execute e.g. redemption.

lukasz-zimnoch: Left that explanation in the code as well: a322ad60

pkg/tbtc/tbtc.go
lukasz-zimnoch: See 796ad378

pdyraga (approved):

PR #3521, opened 2023-04-13
Client-side deposit sweep: Validate proposals

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-core/pull/3520

Here we continue the work started by https://github.com/keep-network/keep-core/pull/3520 by adding the validation of the received deposit sweep proposals. The mechanism requires a lot of changes described below.

### Wallet dispatcher and action abstraction

Existing `dkgExecutor` and `signingExecutor` represent low-level operations of the `node`. For more high-level wallet actions like deposit sweeps (as well as redemptions and moving funds) we need something else. Hence, we are introducing the `walletAction` interface to represent that kind of job. This is a simple interface that encapsulates all the necessary steps within the `execute` method. The `walletAction` is supposed to be implemented by concrete structs representing well-defined wallet actions. The `node` is responsible for orchestrating specific `walletAction` implementations and passing them to the `walletDispatcher`. The `walletDispatcher` is a component that provides the execution layer for `walletActions`. It also makes sure that only one action can be taken by the given wallet at the same time. 

### Chain functions required to perform proposal validation

The `WalletCoordinator` contract exposes the `validateDepositSweepProposal` view function that allows making the majority of required validation using a single call. Here we enhance our code by adding the support for that function to the `pkg/chain/ethereum` package and exposing it through the `tbtc.Chain` interface.

The aforementioned validation function requires some additional data to be fetched by the node. Hence, we are enhancing `tbtc.Chain` with the `GetDepositRequest` and `PastDepositRevealedEvents` methods. Both functions allow fetching deposit extra data required to perform the on-chain validation using `validateDepositSweepProposal`.

Last but not least, we are introducing the `GetBlockNumberByTimestamp` method that allows estimating the block number for the given timestamp. This function is necessary to make efficient calls using `PastDepositRevealedEvents` and do not violate limits enforced by popular Ethereum clients during the `eth_getLogs` calls.

### Bitcoin functions required to perform proposal validation

The `WalletCoordinator.validateDepositSweepProposal` function expects the deposits' funding transactions to be passed in a specific format. So far, our Go `bitcoin` package was not able to fulfill that requirement. Here we improve that by adding new `Serialize*` methods to the `bitcoin.Transaction` type. Thanks to them, a `bitcoin.Transaction` instance obtained using `bitcoin.Client` can be easily converted to the format required by `validateDepositSweepProposal` on-chain function.

### Enhance the wallet registry 

Validation of sweep proposal requires an ability to identify registered wallets using their 20-byte public key hashes. Since the hash is a one-way function and a direct conversion is not possible, we are leveraging the wallet registry to compute 20-byte PKHs upfront and do reverse matching once needed. Thanks to that, the node can identify whether it should start handling an incoming proposal or ignore it altogether.

### Deposit sweep wallet action

Finally, we are introducing the `depositSweepAction` component that encapsulates all actions required to perform a proper sweep based on the received proposal. This component implements the aforementioned `walletAction` and leverages the new `walletDispatcher` mechanism. Right now, the `depositSweepAction` does the following:
- Gathers the prerequisites required to perform on-chain proposal validation. To do so, it iterates over proposal's deposits. For each deposit, it checks that the funding transaction is sufficiently confirmed and it fetches the deposit extra data from past `DepositRevealed` on-chain event matching the particular deposit
- Performs the on-chain proposal validation

In the future, this action will assemble the deposit sweep transaction, trigger its signing process, and broadcast it over the Bitcoin network.

### Next steps

- Implement assembling, signing, and broadcasting of the Bitcoin sweep transaction
- Add missing unit tests. This PR does not provide exhaustive coverage of all components (especially `depositSweepAction`) as the work is still in progress. Once the missing pieces are implemented, proper unit tests should be added where applicable.


pkg/bitcoin/transaction.go
```diff
@@ -89,6 +93,60 @@ func (t *Transaction) Serialize(
...
 }
 
+// SerializeVersion serializes the transaction version to a little-endian
+// 4-byte array.
+func (t *Transaction) SerializeVersion() [4]byte {
+	result := [4]byte{}
+	binary.LittleEndian.PutUint32(result[:], uint32(t.Version))
+	return result
+}
+
```
pdyraga, line 103: It would really help to have [the tables](https://github.com/keep-network/tbtc-v2/blob/031249d0001ab065e80a48e331da1155a636b9e8/solidity/contracts/bridge/BitcoinTx.sol#L27-L91) on the top of `type Transaction struct` and then for each `Serialize*` function denote which parts we are serializing.

For example, in `SerializeInputs` the output has `[ tx_in_count | tx_in]`.

pkg/bitcoin/transaction_test.go
```diff
@@ -72,6 +72,72 @@ func TestTransaction_SerializeRoundtrip(t *testing.T) {
...
+}
+
+func TestTransaction_SerializeInputs(t *testing.T) {
+	transaction := transactionFixture(t)
+
+	serializedInputs := transaction.SerializeInputs()
+	expectedSerializedInputs := hexToSlice(
+		t,
+		"036896f9abcac13ce6bd2b80d125bedf997ff6330e999f2f60"+
+			"5ea15ea542f2eaf80000000000ffffffffed0ae94da996c6f3b89dfe967675d"+
```
pdyraga, line 94: Why is the transaction hash reversed when serialized? I am looking at `transactionFixture` and the outpoint hash of `f8eaf242a55ea15e602f9f990e33f67f99dfbe25d1802bbde63cc1caabf99668`. It is present in the serialized version but the byte order is changed.

pkg/chain/ethereum/tbtc.go
```diff
@@ -996,6 +997,125 @@ func (tc *TbtcChain) OnHeartbeatRequested(
...
+		return nil, err
+	}
+
+	convertedEvents := make([]*tbtc.DepositRevealedEvent, 0)
+	for _, event := range events {
+		var vault *chain.Address
+		if event.Vault != [20]byte{} {
+			v := chain.Address(event.Vault.Hex())
+			vault = &v
+		}
```
pdyraga, line 1035: Wouldn't it be easier to call `event.Vault.Hex()` and let it be zero-bytes? 

The client will have to compare the address against `TBTCVault` contract address anyway.

Same question to the conversion in `GetDepositRequest`.

pkg/chain/ethereum/tbtc.go
```diff
@@ -1176,3 +1322,44 @@ func parseWalletAction(value uint8) (tbtc.WalletAction, error) {
...
+
+	valid, err := tc.walletCoordinator.ValidateDepositSweepProposal(
+		convertDepositSweepProposalToAbiType(proposal),
+		dei,
+	)
+	if err != nil {
+		return fmt.Errorf("validation failed: [%v]", err)
+	}
+
+	// Should never happen but just in case.
```
pdyraga, line 1359: Let's explain for future-us why this should never happen (`ValidateDepositSweepProposal` reverts or returns true).

pkg/tbtc/chain.go
```diff
@@ -198,6 +215,56 @@ type HeartbeatRequestedEvent struct {
...
+// DepositRevealedEventFilter is a component allowing to filter DepositRevealedEvent.
+type DepositRevealedEventFilter struct {
+	StartBlock          uint64
+	EndBlock            *uint64
+	Depositor           []chain.Address
+	WalletPublicKeyHash [][20]byte
+}
+
+// DepositChainRequest represents a deposit request stored on-chain.
+type DepositChainRequest struct {
```
pdyraga, line 259: What do you think about renaming it to `RevealedDeposit`?

I know it is `Deposit.DepositRequest` in Solidity but from the time perspective, I am not sure why we named it that way.

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: > Wouldn't it be easier to call event.Vault.Hex() and let it be zero-bytes?

Actually, this is a matter of how we define an optional value. I think it's more intuitive to have a pointer that explicitly indicates the value can be `nil` than a `chain.Address` which can be empty or zeroed. Remember that `chain.Address` is a chain-agnostic type so 20 zeroed bytes can be a valid value for Ethereum but this may be not the case for other chains. If we use a pointer, the client does not need to know what value represents an empty address for the given chain. It just checks whether the value is `nil` or not.

> The client will have to compare the address against TBTCVault contract address anyway.

I don't see any place where the client will need to compare this with the `TBTCVault` at the moment. Do you have something particular in mind?

pkg/bitcoin/transaction_test.go
lukasz-zimnoch: The serialized hash must be in the `bitcoin.InternalByteOrder`. The `transactionFixture` takes a hash in the human-friendly `bitcoin.ReversedByteOrder` and uses `hexToHash` to construct a proper `bitcoin.Hash` instance. We pass hashes in `bitcoin.ReversedByteOrder` there as it is easier to check them in block explorers out of the box without any conversion.

pkg/tbtc/chain.go
lukasz-zimnoch: I wanted to strongly emphasize that this is a chain-related struct and that it represents deposit data recorded on-chain. We already have a `tbtc.Deposit` struct which represents a deposit in this package. Adding `tbtc.RevealedDeposit` will cause confusion. 

I think the reasoning behind this naming was:
- A "deposit" is something actually made on the BTC chain. On the client side, it is represented by `tbtc.Deposit` and it contains fields required to reconstruct the script hash and data about the funding tx.
- a "depositRequest" is kind of "hey Bridge, I made this deposit on BTC chain, please handle it on your side".  On the client side it is represented by `tbtc.DepositChainRequest` and contains bridge-specific info like: when it was revealed, when it was swept, etc.

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: Sure, see e68ed993

pkg/tbtc/deposit_sweep.go
```diff
@@ -4,15 +4,227 @@ import (
...
+				depositDisplayIndex,
+				err,
+			)
+		}
+
+		// Worth mentioning this should be treated as an estimation, giving
+		// the caveats mentioned in the docstring of GetBlockNumberByTimestamp.
+		revealBlock, err := dsa.chain.GetBlockNumberByTimestamp(
+			uint64(depositRequest.RevealedAt.Unix()),
+		)
```
pdyraga, line 133: The logic in `GetBlockNumberByTimestamp` is quite complicated and I am not sure if we really need it.

Both depositor address and wallet public key hash are indexed fields and we can easily filter by them.
How many deposits to the given wallet a single depositor address can make at maximum? 1000? 5000? 10000? 

If we are afraid we can exceed the limits of popular Ethereum node providers when pulling events (Infura, Alchemy, Quicknode), we should implement partial event pulls like in https://github.com/keep-network/tbtc-v2/pull/604 in the smart contract bindings if protection about this kind of situations is not already there.

pkg/tbtc/deposit_sweep.go
```diff
@@ -4,15 +4,227 @@ import (
...
+	return &depositSweepAction{
+		chain:           chain,
+		btcChain:        btcChain,
+		sweepingWallet:  sweepingWallet,
+		signingExecutor: signingExecutor,
+		proposal:        proposal,
+	}
+}
+
+// TODO: Cover this function with unit tests once everything is completed.
```
pdyraga, line 47: I would consider splitting `execute()` into several smaller functions and testing them separately. Doing one big test for this whole flow will be... complicated 😬 Especially since we have multiple corner cases.

pkg/tbtc/deposit_sweep.go
lukasz-zimnoch: The general logic of how `eth_getLogs` work under the hood is (see https://github.com/ethereum/go-ethereum/blob/master/eth/filters/filter.go#L97):
- Receive a block range. If not provided use [1, latest]
- Iterate over block headers in the range and check the bloom filter
- If there are logs matching the filter criteria, return them

In other words, we must specify a good range to make the call efficient. The indexed field is just used to limit the number of returned data. Both the range and amount of returned data are important in the context of node provider limitations.

Partial event pulling is not a solution in this context. We still need a starting block range to fetch that must be established. Otherwise, we will partially pull events from the range [1, latest].  Partial pulls help when we have a limited range that is too wide or the amount of returned logs are too big. However, we still need to establish a range first and this is why we have `GetBlockNumberByTimestamp`. Also, in the case of a deposit sweep, time is crucial. We must gather the necessary data as quickly as possible and I think the presented solution is all we can do here.

pkg/tbtc/deposit_sweep.go
lukasz-zimnoch: Yep definitely. However, I plan to do so once I have this function completed so expect that in the next PR.

pkg/tbtc/deposit_sweep.go
```diff
@@ -4,15 +4,227 @@ import (
 	"crypto/ecdsa"
 	"fmt"
 	"github.com/keep-network/keep-core/pkg/bitcoin"
+	"github.com/keep-network/keep-core/pkg/chain"
+	"go.uber.org/zap"
 )
 
 const (
 	// depositSweepProposalConfirmationBlocks determines the block length of the
 	// confirmation period that is preserved after a deposit sweep proposal
```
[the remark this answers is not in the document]
lukasz-zimnoch: Package `tbtc` is chain-agnostic. We can't say `Ethereum` here 😁

pkg/tbtc/deposit_sweep.go
```diff
@@ -4,15 +4,227 @@ import (
...
 	// submission.
 	depositSweepProposalConfirmationBlocks = 20
+	// depositSweepRequiredFundingTxConfirmations determines the minimum
+	// number of confirmations that are needed for a deposit funding transaction
```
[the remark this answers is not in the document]
lukasz-zimnoch: See bae3661c

pkg/tbtc/deposit_sweep.go
```diff
@@ -4,15 +4,227 @@ import (
...
+		[]struct {
+			*Deposit
+			FundingTx *bitcoin.Transaction
+		},
+		len(dsa.proposal.DepositsKeys),
+	)
+
+	depositsCount := len(dsa.proposal.DepositsKeys)
+
+	actionLogger.Infof("gathering prerequisties for proposal validation")
```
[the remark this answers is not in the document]
lukasz-zimnoch: See bae3661c

pkg/tbtc/node.go
```diff
@@ -54,11 +56,14 @@ type node struct {
 	// signingExecutors is the cache holding signing executors for specific wallets.
 	// The cache key is the uncompressed public key (with 04 prefix) of the wallet.
 	signingExecutors map[string]*signingExecutor
+
+	walletDispatcher *walletDispatcher
 }
```
pdyraga, line 61: Let's change the order of fields in `node struct` and add some docs. Figuring out which component is responsible for what is getting complicated now.

```
// node represents the current state of an ECDSA node.
type node struct {
	groupParameters *GroupParameters

	chain          Chain
	btcChain       bitcoin.Chain
	netProvider    net.Provider
	walletRegistry *walletRegistry

	// walletDispatcher ensures only one action is executed by a wallet at
	// a time. All possible activities of a created wallet must be represented
	// by appropriate actions dispatched through this component.
	walletDispatcher *walletDispatcher

	// protocolLatch makes sure no expensive number generator operations are
	// running when signing or generating a wallet key are executed. The
	// protocolLatch is used by dkgExecutor and signingExecutor.
	protocolLatch *generator.ProtocolLatch

	// dkgExecutor encapsulates the logic of distributed key generation.
	//
	// dkgExecutor MUST NOT be used outside of this struct.
	dkgExecutor *dkgExecutor

	signingExecutorsMutex sync.Mutex
	// signingExecutors is the cache holding signing executors for specific wallets.
	// The cache key is the uncompressed public key (with 04 prefix) of the wallet.
	// signingExecutor encapsulates the generic logic of signing messages.
	//
	// signingExecutors MUST NOT be used outside of this struct. Please use
	// wallet actions and walletDispatcher to execute an action on an existing
	// wallet.
	signingExecutors map[string]*signingExecutor
}
```

pkg/tbtc/node.go
```diff
@@ -54,11 +56,14 @@ type node struct {
 	// signingExecutors is the cache holding signing executors for specific wallets.
 	// The cache key is the uncompressed public key (with 04 prefix) of the wallet.
 	signingExecutors map[string]*signingExecutor
```
pdyraga, line 58: Having `walletDispatcher` we must ensure `signingExecutors` are not accessed directly outside of actions.

We already have one problem - `OnHeartbeatRequested` bypasses the `walletDispatcher` and calls a `signingExecutor` directly. As a result, we may have heartbeat and sweeping happening at the same time. This will only worsen when we realize multiple heartbeats can happen close to each other, given that a single machine can host multiple wallets.

Let's move the logic that is in `OnHeartbeatRequested` to a heartbeat wallet action. If the wallet is busy sweeping at the time the heartbeat should be executed, we should skip the heartbeat. Sweeping works as a heartbeat.

lukasz-zimnoch: I was thinking about it but there is a problem with heartbeats being a wallet action. Some time ago we agreed that an action is something complex that causes the change of wallet main UTXO while executors represent a lower-level action like signing. Hence, we have the `WalletAction` enum in the `WalletCoordinator` contract and its client-side representation `tbtc.WalletActionType`. Each `walletAction` has the `actionType` method that ties it to a specific `tbtc.WalletActionType` value. Now, heartbeats are just bare signing of random data and do not fulfill the definition of a wallet action. They are not managed by the `WalletCoordinator` and have no representation in the `tbtc.WalletActionType`. Having that in mind, things are more complex than they seem. 

I'm aware of the conflicts you mentioned but I'm not sure if making heartbeats a proper wallet action (despite of the aforementioned arguments against doing it) would help. If a heartbeat action is dispatched before a sweep it will block the sweep anyway and vice versa. I'm wondering whether we shouldn't use a kind of priority system. Something like "a sweep is more important than heartbeat" and "a redemption is more important than a sweep". And allow breaking the currently executed signing in favor of a higher-priority one. Not saying this is a good solution, just exploring possible ideas. WDYT?

pkg/tbtc/node.go
```diff
@@ -262,6 +269,74 @@ func (n *node) getSigningExecutor(
...
+		"node controls signers of wallet PKH [0x%x]; "+
+			"plain-text uncompressed public key of that wallet is [0x%x]; "+
+			"starting orchestration of the deposit sweep action",
+		proposal.WalletPublicKeyHash,
+		walletPublicKeyBytes,
+	)
+
+	// No need to check the boolean flag returned by getSigningExecutor.
+	// We know the node controls some wallet signers as we just got the wallet
+	// from the registry using their public key hash.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 9b1ce327

pkg/tbtc/node.go
lukasz-zimnoch: Holding off with this one until we come to an agreement in https://github.com/keep-network/keep-core/pull/3521#discussion_r1173652203

pkg/tbtc/node.go
pdyraga: I think an action shouldn't necessarily change the UTXO but it should be _something the wallet is doing_. Heartbeat falls into this category. I think heartbeat could be a `WalletAction` in the `WalletCoordinator`.

It is not ideal that we'd coordinate heartbeats from the chain but it's also not ideal for sweeps, moving funds, and redemptions. Moreover, currently, only the most recent wallet executes heartbeats. We have to change it but it's also hard to expect all wallets to execute heartbeat at the same time. At least until we do not change the signing algorithm into something way less computationally-complex. How we'd enforce some order of heartbeats off-chain? That's the same problem as with coordinating sweeps - we'd have to reinvent the consensus mechanism. 

I would consider adding `Heartbeat` value to `WalletCoordinator`'s `WalletAction` and then having a separate `actionType` for it. We do not have to block this PR on it, but we should do it right after and leave a huge TODO block next to `chain.OnHeartbeatRequested(func(event *HeartbeatRequestedEvent)` in `tbtc.go`.

Does it make sense?

lukasz-zimnoch: Sure, we can change the action's definition in this way and pull the heartbeats under it. My concern was only about the fact that we need to treat all actions equally which means adding heartbeats to the `WalletCoordinator` as well. If we agree regarding this part that's good! If so, an additional question arises: how do we want to trigger the heartbeat? Just like sweeps using a function of `WalletCoordinator`?

pkg/tbtc/deposit_sweep.go
pdyraga: Let's take a step back to make sure we are on the same page.

Those are the limits of popular Ethereum node providers I could find:

> Infura has a cap on requests of 10,000 events per query. 

https://blog.infura.io/post/ethereum-rpcs-methods#using-eth-get-logs

> You can make eth_getLogs requests with up to a 2K block range and 150MB limit on the response size OR you can request any block range with a cap of 10K logs in the response.

https://docs.alchemy.com/docs/deep-dive-into-eth_getlogs#making-a-request-to-eth_getlogs

> eth_getLogs and eth_newFilter are limited to a 10,000 blocks range

https://support.quicknode.com/hc/en-us/articles/10258449939473-Is-there-a-block-range-limit-for-querying-logs-and-events-

The first problem we have are API limits. Assuming a 12s average block time, Ethereum will produce 7200 blocks/24h. Even if we use `GetBlockNumberByTimestamp`, sweeping deposits older than two days with Quicknode will not be possible unless the underlying Ethereum communication layer does something smart.

For Alchemy and Infura, realistically, it doesn't look like a problem because it's rather unrealistic a single depositor can make 10k deposits to the same wallet.

Nevertheless, I propose we confirm where we are by executing a test and/or checking the code. If there is such a limitation, we should add a GitHub issue and prioritize it accordingly. 

The second problem we have is how efficient `GetBlockNumberByTimestamp` actually is compared to e.g. checking always the last N blocks.

I did some experiments and it takes:
- 5 `blockByNumber` calls to evaluate block number for yesterday timestamp `1681997871`
- 7 `blockByNumber` calls to evaluate block number for week ago timestamp `1681479471`
- 8 `blockByNumber` calls to evaluate block number for 6 months ago timestamp `1666263335`
- The function executes `blockByNumber` calls for 15 seconds until it times out for 12 months ago timestamp `1650538535`.

The last experiment is concerning because I do not know if that's a bug in `GetBlockNumberByTimestamp` or if it just takes that long because it was 12 months ago. Quite surprised though given it takes 8 blocks for a timestamp 6 months ago.

For this conversation, let's assume the depositor is not doing more than 10,000 deposits to the given wallet. 
Let's also assume the wallet is sweeping deposits not older than a week. This is how it should work once we activate sweeping and deal with pre-sweeping wallets. 

If we hardcode N to 100,000 blocks (~ a week) then it should take:
- One Infura call
- One Alchemy call
- Ten Quicknode calls

This is less than executing additional 7 blocks to find the block timestamp. 

It could be worth checking how long it takes to execute such a call on Infura or Alchemy but I doubt it will take more than 7 additional calls to `blockByNumber`.

The situation we are currently in is exceptional because we'll have to sweep deposits from 6 months ago. But I think it's still all doable with one call to Infura and one call to Alchemy setting the start block to the time when Bridge contract was deployed. 

My tl;dr; is:
- I am concerned it looks that  `GetBlockNumberByTimestamp` does not always work/does not always work efficient enough.
- We should see how long it takes to scan 6-12 months of blocks on Ethereum node and then take the decision if 
`GetBlockNumberByTimestamp` is an optimization or not.
- Quicknode needs special support no matter if we have `GetBlockNumberByTimestamp` or not.

pkg/tbtc/node.go
pdyraga: > If so, an additional question arises: how do we want to trigger the heartbeat? Just like sweeps using a function of WalletCoordinator?

I think yes, there is no point in complicating it and making special paths. 

The more we talk about it and the more problems we find I am keener to have the coordination on some L2 and a bot calling it. The cost will be lower than the dev time spent on the broadcast channel consensus.

pkg/tbtc/deposit_sweep.go
lukasz-zimnoch: > The first problem we have are API limits. Assuming a 12s average block time, Ethereum will produce 7200 blocks/24h. Even if we use GetBlockNumberByTimestamp, sweeping deposits older than two days with Quicknode will not be possible unless the underlying Ethereum communication layer does something smart.

Not sure if I understand correctly but Quicknode has a general limitation "no more than 10k requested blocks" not necessarily from `[now-10k, now]` range. In this case `GetBlockNumberByTimestamp` allows us to limit the requested range to the minimum.

> For Alchemy and Infura, realistically, it doesn't look like a problem because it's rather unrealistic a single depositor can make 10k deposits to the same wallet.

This is not the problem about the number of returned data as this, as you noticed, is limited using the depositor and wallet indexed fields. The problem is about the requested block range. The wider the block range the more time the node needs to fetch the logs. Even if the given provider doesn't have specific limitations on the requested range, the time of a single request can be problematic. There may also be some timeouts on the provider side that may bite us. 

Also, we should also consider self-hosted Ethereum nodes that can be easily nuked by a wide block range `eth_getLogs` call if such a node does not have appropriate protections. See Alechemy's note about `eth_getLogs` vulnerabilities: https://docs.alchemy.com/docs/deep-dive-into-eth_getlogs#making-a-request-to-eth_getlogs

Regarding the performance of `GetBlockNumberByTimestamp`: its efficiency is not linearly correlated with how the requested timestamp is far from now. The clue is the actual average block time between the predicted and current block. Since the Ethereum merge in September 2022, the average block time is slightly about 12 seconds. The `GetBlockNumberByTimestamp` assumes a bit higher average block time (13 seconds) which allows it to make fewer jumps to the requested block and not overshoot it. Here is why the number of calls for 6 months ago timestamp is quite low. This timestamp is after September 2022 and the actual average block time has a stable value above 12 seconds. On the other hand, the problematic 1-year-ago timestamp from your exeriment is before September 2022 and the actual average block time between that timestamp and now is much more variable and much bigger than 12 seconds. That means `GetBlockNumberByTimestamp` will jump by far more blocks to the past than necessary. In that case, the function will start going forward block by block to compensate for the error and will need much more `blockByNumber` requests. The timeout you observed comes from `blockByNumber` and is set to 10 seconds initially. This timeout may be too short for a node that must get many blocks from the far past. Here is what you observed in the last experiment. The conclusion is that this is not a bug directly in `GetBlockNumberByTimestamp` but we should rather think about using a longer timeout.

pkg/tbtc/node.go
lukasz-zimnoch: Seec 08769b5f

pkg/tbtc/deposit_sweep.go
pdyraga: > > The first problem we have are API limits. Assuming a 12s average block time, Ethereum will produce 7200 blocks/24h. Even if we use GetBlockNumberByTimestamp, sweeping deposits older than two days with Quicknode will not be possible unless the underlying Ethereum communication layer does something smart.
> 
> Not sure if I understand correctly but Quicknode has a general limitation "no more than 10k requested blocks" not necessarily from [now-10k, now] range. In this case GetBlockNumberByTimestamp allows us to limit the requested range to the minimum.

Yes, but I think this is still not enough to support Quicknode. Say that `GetBlockNumberByTimestamp` returned block `N - 20000` where `N` is the current block. If we do a request with the `[N - 20000, N]` range, it will fail for Quicknode. We need two separate requests: `[N-20000, N - 10000]` and `[N-10000, N]`. I do not think it is something we need to worry about in this PR but we should address it soon after. It could also be addressed by making it clear in the docs that Quicknode is currently not supported and opening an issue for implementing a similar mechanism as in https://github.com/keep-network/tbtc-v2/pull/604 for Go. But let's first confirm this is really the case and nothing in our Go bindings or `go-ethereum` makes it work. Just a simple call to Quicknode for past events with more than 10,000 blocks in range should do the trick.

Regarding `GetBlockNumberByTimestamp` in general, what I am trying to say is that I am not sure if this is more performant than just selecting the range of the last month's blocks. For every deposit being swept, we will have to execute 7-8 additional calls. Maybe that takes more time than just one call with `[N - 30 * 7200, N]` blocks. We can not do it now given we will be sweeping deposits from the last few months though.

So tl;dr; of my proposition would be:
1. Leave this as-is.
2. Once this PR is merged, please check if this works with Quicknode. If not, let's add a note to the docs that Quicknode is not supported.

Separately, I'll open an issue to investigate the performance of `GetBlockNumberByTimestamp` compared to just making a call with a wider block range.

How does it sound?

pkg/bitcoin/transaction_test.go
pdyraga: Can we add your answer to my question as a comment to `transactionFixture`?

pkg/tbtc/deposit_sweep.go
lukasz-zimnoch: > Yes, but I think this is still not enough to support Quicknode. Say that GetBlockNumberByTimestamp returned block N - 20000 where N is the current block. If we do a request with the [N - 20000, N] range, it will fail for Quicknode. We need two separate requests: [N-20000, N - 10000] and [N-10000, N]. 

Thanks to `GetBlockNumberByTimestamp` we don't have to use such a wide range. We are actually using just a small margin (10 blocks) on both sides of the estimated reveal block to handle possible corner cases. See https://github.com/keep-network/keep-core/blob/f49238d8ca102eefc7b9b9bb6aaab0e0d14622d4/pkg/tbtc/deposit_sweep.go#L157

That means Quicknode will work as we use only 20-block ranges which are far from the 10k blocks limit.

Regarding your proposition, it sounds good! We should explore the performance of `GetBlockNumberByTimestamp` later and maybe introduce some kind of event indexer that would cache chain events and will be used as the first-choice source of data and only falling back to the current mechanism with `GetBlockNumberByTimestamp` in case of a cache miss.

pkg/chain/ethereum/tbtc.go
pdyraga: > Actually, this is a matter of how we define an optional value. I think it's more intuitive to have a pointer that explicitly indicates the value can be nil than a chain.Address which can be empty or zeroed. Remember that chain.Address is a chain-agnostic type so 20 zeroed bytes can be a valid value for Ethereum but this may be not the case for other chains. If we use a pointer, the client does not need to know what value represents an empty address for the given chain. It just checks whether the value is nil or not.

Good reasoning. Can we add a code comment to `DepositRevealedEvent` and `DepositChainRequest` in `pkg/tbtc/chain.go` and make it an API requirement that the `Vault` should be `nil` if it wasn't set on the chain?

> I don't see any place where the client will need to compare this with the TBTCVault at the moment. Do you have something particular in mind?

You are right. The `WalletCoordinator` contract checks if all deposits [target the same vault](https://github.com/keep-network/tbtc-v2/blob/031249d0001ab065e80a48e331da1155a636b9e8/solidity/contracts/bridge/WalletCoordinator.sol#L532-L540) and this is enough.

pkg/tbtc/chain.go
pdyraga: That makes sense. Can we just extend the documentation of `DepositChainRequest` a bit:

```
// DepositChainRequest represents a deposit request stored on-chain.
// This is a deposit revealed to the Bridge and recorded on-chain. There is no
// guarantee this deposit actually happened on the Bitcoin side.
type DepositChainRequest struct {
```

?

pkg/bitcoin/transaction_test.go
lukasz-zimnoch: Sure 94d72d77

pkg/tbtc/deposit_sweep.go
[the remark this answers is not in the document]
pdyraga: Good call. So maybe *the host chain*? 

```
	// depositSweepProposalConfirmationBlocks determines the block length of the
	// confirmation period on the host chain that is preserved after a deposit
	// sweep proposal submission.
	depositSweepProposalConfirmationBlocks = 20
```

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: See 0bfa45aa

pkg/tbtc/chain.go
lukasz-zimnoch: See 80b6a3e9

pkg/bitcoin/transaction.go
pdyraga: I was looking at the Solidity documentation for`BitcoinTx.Info` to understand why are we taking 4 bytes or why are we skipping 4 bytes in other functions. The references to `developer.bitcoin.org` are helpful but IMO they are not as helpful as having the structure described in the code comments as we have in `BitcoinTx.Info`. It is not a blocker, I can leave this conversation open to marinate for a while to see if we can find a way to port some of those docs to Go.

pkg/tbtc/deposit_sweep.go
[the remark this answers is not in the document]
lukasz-zimnoch: Much better! See 7a41a8f0

pkg/chain/ethereum/ethereum.go
```diff
@@ -318,6 +319,144 @@ func (bc *baseChain) OperatorKeyPair() (
 	return privateKey, publicKey, nil
 }
 
+// GetBlockNumberByTimestamp gets the block number for the given timestamp.
+// In the best case, the block with the exact same timestamp is returned.
+// If the aforementioned is not possible, it tries to return the closest
+// possible block. However, there may be cases where another close block
+// is returned instead of the closest one. This may happen if the actual
```
pdyraga, line 326: >  However, there may be cases where another close block is returned instead of the closest one.

Can you give an example? Should the two final loops `for block.Time() < timestamp` and `if block.Time() > timestamp` fix it?

pkg/bitcoin/transaction.go
lukasz-zimnoch: Sure! What I want to avoid is putting too many docs as they add additional maintenance work and can obfuscate the picture even more. The Go `bitcoin` package differs from the Solidity code we were working with in the past. In Solidity, we used to deal with pure bytes quite often so a huge amount of docs was the only way to explain what happens. In Go `bitcoin` package, the code itself is more verbose. All Bitcoin primitives and their contents are represented by dedicated Go types which have their own documentation. In that context, having a huge block of docs above `bitcoin.Transaction` just feels a bit redundant.

pkg/chain/ethereum/ethereum_integration_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/keep-network/keep-common/pkg/chain/ethereum/ethutil"
+	"github.com/keep-network/keep-core/pkg/internal/testutils"
+	"reflect"
+	"testing"
+	"time"
+)
+
+// TODO: Include integration test in the CI.
+// To run the tests execute `go test -v -tags=integration ./...`
```
pdyraga, line 17: I do not feel comfortable merging it without having the code for `GetBlockNumberByTimestamp` covered in CI and it may take a moment to implement integration tests action right. :( I think we could quite easily rework it into a unit test though.

One option is to implement `_getBlockNumberByTimestamp` that will accept two functions as parameters: one to get the current block and another to get the block by number. Then, `GetBlockNumberByTimestamp` would call it. This approach would let us implement even more tests - we can generate any array of blocks we want.

Another option is to instantiate `baseChain` manually passing a mock implementation of `client` and `blockCounter` but I feel it is more work and boilerplate code.

pkg/chain/ethereum/ethereum.go
```diff
@@ -318,6 +319,144 @@ func (bc *baseChain) OperatorKeyPair() (
...
+		return 0, fmt.Errorf("requested timestamp is in the future")
+	}
+
+	// Corner case shortcut.
+	if block.Time() == timestamp {
+		return block.NumberU64(), nil
+	}
+
+	// The Ethereum average block time (https://etherscan.io/chart/blocktime)
+	// is slightly over 12 most of the time. If we assume it to be 12 here,
```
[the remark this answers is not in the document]
lukasz-zimnoch: Done as part of 9e52fa1f

pkg/chain/ethereum/ethereum.go
lukasz-zimnoch: Yeah, this is no longer the case. Fixed this as part of 9e52fa1f

pkg/chain/ethereum/ethereum_integration_test.go
lukasz-zimnoch: Yeah, I was wondering about integration vs unit tests for that. I chose the former because:
- it makes more sense to test this function with a real ETH client, especially performance-wise
- `blockByNumber` is not straightforward to mock as their return type `*types.Block` holds the relevant data as private fields. It is possible to overcome this by adding additional interfaces but that seemed non-elegant to me and hence I decided to stick with integration tests

pkg/tbtc/deposit_sweep.go
```diff
@@ -4,15 +4,228 @@ import (
...
+		if err != nil {
+			return fmt.Errorf(
+				"cannot get on-chain request data for deposit [%v]: [%v]",
+				depositDisplayIndex,
+				err,
+			)
+		}
+
+		// Worth mentioning this should be treated as an estimation, giving
+		// the caveats mentioned in the docstring of GetBlockNumberByTimestamp.
```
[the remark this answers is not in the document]
lukasz-zimnoch: Removed 9571928e

pdyraga (approved):

pkg/tbtc/node.go
pdyraga: Happened in https://github.com/keep-network/tbtc-v2/pull/607 and https://github.com/keep-network/tbtc-v2/pull/609.

PR #3526, opened 2023-04-17
Regenerating code: smart contract bindings and client's start help

tomaszslabon:
Refs https://github.com/keep-network/keep-core/issues/3532

This PR adds missing generated code.
The code was generated using the `make all` command.

Note: return to this PR later when LightRelayMaintainerProxy is deployed and generate the bindings for mainnet.


pkg/chain/ethereum/ecdsa/gen/abi/WalletRegistry.go
```diff
@@ -1355,6 +1400,27 @@ func (_WalletRegistry *WalletRegistryTransactorSession) CloseWallet(walletID [32
 	return _WalletRegistry.Contract.CloseWallet(&_WalletRegistry.TransactOpts, walletID)
 }
 
+// ForceAddWallet is a paid mutator transaction binding the contract method 0x2717ee0d.
+//
+// Solidity: function forceAddWallet(bytes groupPubKey, bytes32 membersIdsHash) returns()
+func (_WalletRegistry *WalletRegistryTransactor) ForceAddWallet(opts *bind.TransactOpts, groupPubKey []byte, membersIdsHash [32]byte) (*types.Transaction, error) {
+	return _WalletRegistry.contract.Transact(opts, "forceAddWallet", groupPubKey, membersIdsHash)
+}
```
pdyraga, line 1408: This looks like a function from dApp-only contracts and it should not be in the development package.

It should definitely not be there for mainnet package.

pdyraga (changes requested):

pkg/chain/ethereum/ecdsa/gen/abi/WalletRegistry.go
pdyraga: The problem was described here: https://github.com/keep-network/keep-core/issues/3531

Until this is solved, as a workaround, please try generating contract bindings based on the mainnet package.

tomaszslabon: We decided to return to this PR later and generate the bindings based on the `mainnet` package.
At present, we cannot generate the bindings based on the `mainnet` package because we haven't deployed LightRelayMaintainerProxy yet and it causes an error.

PR #3534, opened 2023-04-19
Automatically generate contracts documentation

michalinacienciala:
We are creating GH Actions workflows which automatically generates the contracts documentation in Markdown, based on the functions and the NatSpec-format comments in the Solidity files stored in the `contracts` folders in `keep-network/keep-core/solidity/random-beacon` and `keep-network/keep-core/solidity/ecdsa`. For certain workflow triggers, the generated documentation will be pushed to the `threshold-network/threshold` repository, under `./docs/app-development/random-beacon/random-beacon-api` / `./docs/app-development/ecdsa/ecdsa-api`. As the `.docs` folder is synched with Threshold docs GitBook space, the pushed docs will be displayed in HTML under `docs.threshold.network` domain.

There are two main jobs in the workflows:

* `contracts-docs-publish-preview`
* `contracts-docs-publish`

Both call a reusable workflow `reusable-solidity-docs.yml` which resides in the `keep-network/ci` repository under `.github/workflows`. The jobs differ in parameters with witch the reusable action is called. The common part of the jobs is the beginning stage, where the Solidity files are being prepared for Markdown generation. During that stage the jobs will remove the spaces chars between the NatSpec comments markers (`///`) and the text of the comment. This is done to ensure proper rendering of the lists in the output docs and is a default action made by `reusable-solidity-docs.yml`.
Once files are ready, the jobs use Docgen tool to generate the Markdown docs. The tool needs to be installed in the project and configured in the `hardhat.config.ts`. The configuration that we use specifies that the docs should be generated to the `geerated-docs` subfolder, to one single `index.md` file. The `.sol` files in the `./solidity/contracts/test` folder will be ignored during generation (those are test/stub contracts which are not used on Mainnet). A custom template will be used during docs generations. The template was created based on the default
https://github.com/OpenZeppelin/solidity-docgen/blob/master/src/themes/markdown/common.hbs template, but it removes the cursive from the `@dev` type comments (because the cursive didn't work well with formatting of the lists). Once the Docgen is run and the `index.md` file is generated, the jobs will add a Table of Contents to the file, to improve the navigation. The TOC will be added by the `markdown-toc` tool and will be generated using `--maxdepth 2` param, which results in listing all the contract names, but not the functions. Once the TOC is added, we'll start to see the difference in the behavior of the jobs.

The `contracts-docs-publish-preview` job is triggered for:
* `pull_request` events, if the PR modifies either files in `./solidity/<project>/contracts` or the workflow files itself,
* `push` events, when the push is made to a branch which name starts with `releases/mainnet/solidity/` and project's files are modified,
* `workflow_dispatch` events. The job publishes the generated artifacts as artifacts of the GH Actions workflow run. For PRs, link to the run with the artifacts will be posted as a PR comment.

The `contracts-docs-publish` job is triggered for:
* published releases, which names start with `refs/tags/solidity/`. The job pushes the generated file to the
`./docs/app-development/random-beacon/random-beacon-api` / `./docs/app-development/ecdsa/ecdsa-api`folder in the `threshold-network/threshold` repo, to the `main` branch. The `.docs` folder is synched with the GitBook Threshold documentation, so thanks to the job, the changes in the contracts should be reflected on the docs.threshold.network. In order for the job to work, we need to configure a user which will create the commit The user needs to have right to commit to the `threshold-network/threshold` repo. We also need to configure following secrets:
- [x] `THRESHOLD_DOCS_GITHUB_TOKEN`
- [x] `THRESHOLD_DOCS_GPG_PRIVATE_KEY_BASE64`
- [x] `THRESHOLD_DOCS_GPG_PASSPHRASE ` We're using a GPG key to be able to do verified commits.

Refs:
https://github.com/keep-network/ci/pull/43
https://github.com/keep-network/tbtc-v2/pull/584
https://github.com/threshold-network/solidity-contracts/pull/138


michalinacienciala:
The docs got published to this PR: https://github.com/threshold-network/threshold/pull/32/files.

.github/workflows/contracts-ecdsa-docs.yml
```diff
@@ -0,0 +1,86 @@
...
+    if: github.event_name == 'release' && startsWith(github.ref, 'refs/tags/solidity/')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Wait 4 minutes
+        run: sleep 240
+
+  # This job will be triggered for releases which name starts with
+  # `refs/tags/solidity/`. It will generate contracts documentation in
+  # Markdown and sync it with a specific path of
+  # `threshold-network/threshold` repository. If changes will be detected,
```
pdyraga, line 67: Let's keep in mind to update this comment as a part of the `threshold-network/threshold` repo rename work.

michalinacienciala: Already taken care of in https://github.com/keep-network/keep-core/pull/3653 👍

PR #3535, opened 2023-04-20
Bitcoin difficulty maintainer logic update

tomaszslabon:
This PR updates the logic of the Bitcoin difficulty maintainer, so that relay retargetting can be performed either via  `LightRelayMaintainerProxy` or directly via `LightRelay`. The decision if based on the value of the  `disableBitcoinDifficultyProxy ` flag. 

By default, this flag is set to false and `LightRelayMaintainerProxy` is used. If set to true, retargets are done directly via `LightRelay`.

### DIFF
--- a/pkg/chain/ethereum/bitcoin_difficulty.go
+++ b/pkg/chain/ethereum/bitcoin_difficulty.go
@@ -9,6 +9,7 @@ import (
 	"github.com/keep-network/keep-core/pkg/bitcoin"
 	"github.com/keep-network/keep-core/pkg/chain"
 	"github.com/keep-network/keep-core/pkg/chain/ethereum/tbtc/gen/contract"
+	"github.com/keep-network/keep-core/pkg/maintainer"
 )
 
 // Definitions of contract names.
@@ -28,10 +29,13 @@ type BitcoinDifficultyChain struct {
 // NewBitcoinDifficultyChain construct a new instance of the Bitcoin difficulty
 // - specific Ethereum chain handle.
 func NewBitcoinDifficultyChain(
-	config ethereum.Config,
+	ethereumConfig ethereum.Config,
+	maintainerConfig maintainer.Config,
 	baseChain *baseChain,
 ) (*BitcoinDifficultyChain, error) {
-	lightRelayAddress, err := config.ContractAddress(LightRelayContractName)
+	lightRelayAddress, err := ethereumConfig.ContractAddress(
+		LightRelayContractName,
+	)
 	if err != nil {
 		return nil, fmt.Errorf(
 			"failed to attach to LightRelay contract: [%v]",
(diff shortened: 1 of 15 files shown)


pkg/maintainer/bitcoin_difficulty.go
```diff
@@ -138,19 +138,6 @@ func (bdm *bitcoinDifficultyMaintainer) verifySubmissionEligibility() error {
...
-				"block headers: [%w]",
-			err,
-		)
-	}
-
-	if !authorizationRequired {
-		return nil
-	}
-
 	maintainerAddress := bdm.chain.Signing().Address()
```
tomaszslabon, line 141: This function was removed as the BTC difficulty bot submits transactions to `LightRelayMaintainerProxy` and therefore only needs to be authorized in `LightRelayMaintainerProxy`.

However, we could in theory perform some additional checks to verify the whole setup:
- retrieve the `LightRelay` from `LightRelayMaintainerProxy` and see if `LightRelayMaintainerProxy`  is authorised in it,
- retrieve the `ReimbursementPool` from `LightRelayMaintainerProxy` and see if `LightRelayMaintainerProxy`  is authorised in it.

That would give us certainty that the whole maintainer setup is correct. If we are lacking some authorisation, we will only learn about it when attempting `retarget` which happens every 2 weeks.

pkg/chain/ethereum/bitcoin_difficulty.go
```diff
@@ -85,6 +85,21 @@ func NewBitcoinDifficultyChain(
...
+				"contract: [%v]",
+			err,
+		)
+	}
+
+	// Verify the LightRelay set in LightRelayMaintainerProxy is the same
+	// instance as the one set in the client via configuration.
+	if lightRelayAddress != retrievedLightRelayAddress {
+		return nil, fmt.Errorf("mismatch between LightRelay addresses")
+	}
```
pdyraga, line 101: 👌🏻

pkg/maintainer/bitcoin_difficulty.go
pdyraga: Is there a way to support both the proxied and the non-proxied mode?

If we were to support only the proxied mode, I think we should only pass the address of `LightRelayProxy` to the client and add functions to the contract allowing to expose all the `LightRelay` functions "hidden" by the proxy. This would be simpler: only one contract binding, only one address to provide.

I was not complaining about two addresses and bindings in https://github.com/keep-network/keep-core/pull/3525 because I was thinking we will support both proxied and non-proxied modes. In other words, there is some gas overhead with going through `LightRelayMaintainerProxy` so if the DAO decides to run a bot or to refund the bot runner by simple ETH transfer, the bot would work against `LightRelay` directly.

One option could be to add a flag to disable retargeting via proxy. We'd retarget via proxy by default. We would have two variations of some functions in `pkg/chain/ethereum/bitcoin_difficulty.go`: `Retarget` and `RetargetWithRefund`. Also `IsAuthorized` and `IsAuthorizedForRefund`.

What do you think?

pdyraga (approved):

PR #3538, opened 2023-04-25
Client-side deposit sweep: Bitcoin transaction

lukasz-zimnoch:

This is the continuation of the work presented in https://github.com/keep-network/keep-core/pull/3521. The goal of this PR is to wrap up the client-side deposit sweep support implementation by adding the ability to assemble, sign and broadcast the deposit sweep Bitcoin transaction. This PR consists of multiple parts described below.

### Assemble the deposit sweep Bitcoin transaction

This part leverages the `assembleDepositSweepTransaction` function to build an unsigned deposit sweep transaction that is ready to be signed using the tECDSA protocol. This process requires some prerequisites:
- List of deposits that will be part of the sweep transaction
- The sweeping wallet public key
- The sweep transaction fee
- The plain-text wallet main UTXO that matches the on-chain hash stored in the `Bridge` contract

All of them but the last can be determined using the existing code. To fulfill the last one, this changeset introduces the `DetermineWalletMainUtxo` function that is responsible for obtaining the current plain-text wallet main UTXO. Under the hood, it uses the recent wallet's Bitcoin transaction history to find the UTXO matching the main UTXO hash stored in the `Bridge`. This approach is more straightforward than possible alternatives. An example alternative would be using the past events emitted by `Bridge` transactions that change the main UTXO hash. However, this approach would be more complex and less efficient comparing to the presented one.

### Sign the deposit sweep transaction

Once the `assembleDepositSweepTransaction` return the unsigned transaction, the deposit sweep action can proceed to tECDSA signing using the provided `signingExecutor` instance. Running the signing process within the deposit sweep action has some caveats:
- The signing start block
- The signing timeout

Regarding the first one, the signing process uses a start block to synchronize all participants and provide smooth execution. The deposit sweep action knows the block of the sweep proposal which may be used as the signing start block. However, the action performs some pre-signing steps (deposit data gathering and proposal validation) that take a significant amount of time. Using the block of the sweep proposal as a signing coordination point would cause failures of the first signing attempts as nodes would still do the pre-signing steps and wouldn't be able to join the process. That means the action logic must delay signing by a constant factor, depending on the number of proposal's deposits, in order to start signing at the right moment. This is achieved using the `depositSweepSigningDelayBlocks` constant.

Regarding the signing timeout, the sweep action logic must make sure that the signing terminates far before the entire action's timeout. Without that, signing may complete late and not leave enough time to broadcast the sweep transaction properly. Such a case is dangerous as it may cause signature leaks and make the wallet subject to fraud accusations. This problem is addressed using the `depositSweepSigningTimeoutSafetyMargin` constant.

### Broadcast and monitor the deposit sweep transaction

Once the sweep transaction is signed, the deposit sweep action begins the broadcasting step. In detail, this process looks like this:
1. The deposit sweep action calls the `Broadcast` function of the `bitcoin.Chain`. The outcome is logged. An error returned from that function may indicate the transaction was already broadcasted by another wallet operator so it should not be considered as an action failure.
2. Regardless of the broadcast outcome, the action waits for a fixed delay and checks whether the transaction is known on the Bitcoin chain. It is enough that the transaction is in the mempool as this is when the wallet can consider its job as done and pass the responsibility back to the coordinator and maintainer.
3. If the transaction is known on the Bitcoin chain, the action completes with success. Otherwise, it goes back to point 1 and tries to broadcast again. This loop is repeated until the transaction lands in the Bitcoin mempool or the broadcast timeout is hit, whichever happens first.


pkg/tbtc/tbtc.go
```diff
@@ -339,6 +339,7 @@ func Initialize(
 
 				node.handleDepositSweepProposal(
 					event.Proposal,
+					expiresAt,
```
pdyraga, line 342: Really nice approach to determine it from the `WalletCoordinator`'s lock and then subtract `dsa.signingTimeoutSafetyMargin`! 🙌🏻

pdyraga (approved):
Outstanding work 💫

pdyraga:
Hey @lukasz-zimnoch, can you please merge `main` to this branch? The build will start passing then. We need the changes from #3540.

PR #3541, opened 2023-05-01
Heartbea(s)t III: Trigger heartbeats from WalletCoordinator contract

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/3512

See https://github.com/keep-network/tbtc-v2/pull/609
See https://github.com/keep-network/keep-core/pull/3539
See https://github.com/keep-network/keep-core/pull/3540

To make sure that older wallets are still accessible for redemption, we need to perform heartbeats.

With the existing, pretty computationally-expensive tECDSA signing algorithm, we would prefer not to execute heartbeats at the same time as other wallet actions (sweeping, redemptions, moving funds). Moreover, we would prefer not to execute heartbeats from multiple wallets at the same time.

Such coordination is complicated off-chain for the same reason as the coordination of sweeps and redemptions. We could - as we used to - execute heartbeats at fixed hours but it does not help with coordination between heartbeats and sweeps. Also, it does not solve the problem of multiple heartbeats simultaneously.

For this reason, the heartbeat has been made a first-class wallet action coordinated in the on-chain contract just like the rest of the wallet actions.

This changeset removes the fixed schedule for heartbeats and uses the `WalletCoordinator` contract as a trigger.

## Testing

I tested this change by requesting heartbeats from two wallets I had in my local dev setup. Heartbeats were requested when all nodes were active and when the majority of nodes were offline. In all cases, the heartbeat eventually succeeded.

Heartbeat can be requested with:
```
❯ KEEP_ETHEREUM_PASSWORD="password" ./keep-client ethereum tbtc wallet-coordinator request-heartbeat \ 
0x3805eed0bb0792eff8815addedb36add2c7257e5 \
0xFFFFFFFFFFFFFFFF0000000000000001 \
--config configs/config.deployer.toml --submit
```

Keep in mind deployment scripts do not register coordinator address and it has to be done from the deployer account manually:
```
 ❯ KEEP_ETHEREUM_PASSWORD="password" ./keep-client ethereum tbtc wallet-coordinator add-coordinator \
0x524f2E0176350d950fA630D9A5a59A0a190DAf48 \
--config configs/config.deployer.toml --submit
```


pkg/tbtc/deposit_sweep.go
```diff
@@ -114,18 +118,7 @@ func newDepositSweepAction(
...
-	}
-
-	actionLogger := logger.With(
-		zap.String("wallet", fmt.Sprintf("0x%x", walletPublicKeyBytes)),
-		zap.String("action", dsa.actionType().String()),
-		zap.Uint64("startBlock", dsa.proposalProcessingStartBlock),
-	)
-
-	validateProposalLogger := actionLogger.With(
+	validateProposalLogger := logger.With(
```
lukasz-zimnoch, line 121: You refer to the global logger here, not the one held in the action. Also, you won't be able to do `.With` on the `log.StandardLogger` type.

pdyraga: Good catch. I fixed it in https://github.com/keep-network/keep-core/pull/3541/commits/cdfcdd71e978489d1d6d02de96c76e017b36e901. I was looking for an interface but it seems like there is none in zap and we have to use `*zap.SugaredLogger` for now. I did it only for the deposit sweep action because the heartbeat action is not creating any new logger.

pkg/tbtc/deposit_sweep.go
```diff
@@ -146,7 +139,7 @@ func (dsa *depositSweepAction) execute() error {
 		)
 	}
 
-	createTxLogger := actionLogger.With(
+	createTxLogger := logger.With(
```
lukasz-zimnoch, line 142:
```suggestion
	createTxLogger := dsa.logger.With(
```

pkg/tbtc/deposit_sweep.go
```diff
@@ -159,7 +152,7 @@ func (dsa *depositSweepAction) execute() error {
 		return fmt.Errorf("create transaction step failed: [%v]", err)
 	}
 
-	broadcastTxLogger := actionLogger.With(
+	broadcastTxLogger := logger.With(
```
lukasz-zimnoch, line 155:
```suggestion
	broadcastTxLogger := dsa.logger.With(
```

pkg/tbtc/node.go
```diff
@@ -285,6 +286,93 @@ func (n *node) getSigningExecutor(
...
+
+	logger.Infof(
+		"node controls signers of wallet PKH [0x%x]; "+
+			"plain-text uncompressed public key of that wallet is [0x%x]; "+
+			"starting orchestration of the heartbeat action",
+		walletPublicKeyHash,
+		walletPublicKeyBytes,
+	)
+
+	// The  request processing started after a confirmation period represented
```
lukasz-zimnoch, line 346:
```suggestion
	// The request processing started after a confirmation period represented
```

pdyraga: 289d1e520c2e63a164115f232861dc64f06b0e53

pkg/tbtc/deposit_sweep.go
pdyraga: 289d1e520c2e63a164115f232861dc64f06b0e53

lukasz-zimnoch (approved):

PR #3545, opened 2023-05-02
CLI command to list tBTC deposits

nkuba:
We added a CLI command to list the tBTC deposits. The command queries the chain for revealed deposits and outputs their details to stdout.

To list all deposits run:
```
keep-client coordinator deposits
```

Following flags can be used with the command:
- `--wallet <address>` - filter deposits only for a given wallet
- `--hide-swept` - doesn't show deposits that were already swept
- `--sort-amount` - sort output by the deposit amount
- `--head <number>` - show `<number>` first deposits
- `--tail <number>` - show `<number>` last deposits

The command requires `ethereum` and `bitcoin` node details to be provided in the config.

### Funding Transaction

The command outputs the Bitcoin funding transaction details in a specific format:
`<transaction hash>:<output index>` (e.g. `a3d1781b59d5e8680772a8bb7f897c4ff0459d3465d7fa678f80a4f0ec900574:1`). The same format will be required by another CLI command to propose deposits sweep.

## Example

Sample execution of the command for Goerli network:
```
✗ keep-client --config ./configs/forked/config.toml --goerli \
  coordinator \
  deposits \
  --wallet 0x03b74d6893ad46dfdd01b9e0e3b3385f4fce2d1e \
  --sort-amount \
  --head 5
```
**Output:**
```
 index                                     wallet value (BTC)                                                        deposit key                                                funding transaction confirmations swept
     0 0x03b74d6893ad46dfdd01b9e0e3b3385f4fce2d1e     0.01000 0x06cdc15cbf3f024d05d164b9fcabfe3a6af7180b20ed2b685cf3a2241426667f c0a91a81b401aae3bc639ef24bc686e1b59b83472a9b571510c34c03cc60414a:0             0 false
     1 0x03b74d6893ad46dfdd01b9e0e3b3385f4fce2d1e     0.01000 0xc0612aed21365628908474f634a60124eb4ab3440f133602aaedb06310a6275d 63cb18fdbe4ceb8b0747dbfcb11989605f78cffbe4379b9047124118cfc20aef:0         14212 false
     2 0x03b74d6893ad46dfdd01b9e0e3b3385f4fce2d1e     0.01000 0xc77584f754d98abef75223e7dadfc02484c87052670de6038514a9b951e94156 46a9b50dd44f11364d9924bd7995830cd7cd1a772910f22ed7ab507137653c6d:0         14199 false
     3 0x03b74d6893ad46dfdd01b9e0e3b3385f4fce2d1e     0.01000 0x89835e76759b3e0f2ca5057c0738d6f792083be4aaf8e064d9733720f9973e0c d07564117859bf7b3ee02bd6659d846d1ea5c4b7c204295ab1270948be615047:0         14052 false
     4 0x03b74d6893ad46dfdd01b9e0e3b3385f4fce2d1e     0.01000 0x78cd69a78750f31d49a4693281a97448581361c7e93a9a70ea3a4931199b6c87 8356227b87fb191d61cff2455cd3f8c33ae208c69a7698b5baf1a331db6d4d8f:1         13981 false
```

### DIFF
--- a/cmd/wallet.go
+++ b/cmd/wallet.go
@@ -15,6 +15,8 @@ var (
 	walletFlagName       = "wallet"
 	hideSweptFlagName    = "hide-swept"
 	sortByAmountFlagName = "sort-amount"
+	headFlagName         = "head"
+	tailFlagName         = "tail"
 )
 
 // WalletCommand contains the definition of tBTC wallets tools.
@@ -57,6 +59,16 @@ var depositsCommand = cobra.Command{
 			return fmt.Errorf("failed to find show swept flag: %v", err)
 		}
 
+		head, err := cmd.Flags().GetInt(headFlagName)
+		if err != nil {
+			return fmt.Errorf("failed to find head flag: %v", err)
+		}
+
+		tail, err := cmd.Flags().GetInt(tailFlagName)
+		if err != nil {
+			return fmt.Errorf("failed to find tail flag: %v", err)
+		}
+
 		_, tbtcChain, _, _, _, err := ethereum.Connect(ctx, clientConfig.Ethereum)
 		if err != nil {
 			return fmt.Errorf(
@@ -76,6 +88,8 @@ var depositsCommand = cobra.Command{
 			wallet,
 			hideSwept,
 			sortByAmount,
+			head,
+			tail,
 		)
 	},
 }
@@ -106,6 +120,18 @@ func init() {
 		sortByAmountFlagName,
 		false,
 		"sort by deposit amount",
+	)
+
+	depositsCommand.Flags().Int(
+		headFlagName,
+		0,
+		"get head of deposits",
+	)
+
+	depositsCommand.Flags().Int(
+		tailFlagName,
+		0,
+		"get tail of deposits",
 	)
 
 	WalletCommand.AddCommand(&depositsCommand)
--- a/cmd/wallet.go
+++ b/cmd/wallet.go
@@ -1,0 +1,100 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	walletcmd "github.com/keep-network/keep-core/cmd/wallet"
+	"github.com/keep-network/keep-core/config"
+	"github.com/keep-network/keep-core/pkg/chain/ethereum"
+)
+
+var (
+	walletFlagName       = "wallet"
+	hideSweptFlagName    = "hide-swept"
+	sortByAmountFlagName = "sort-amount"
+)
+
+// WalletCommand contains the definition of tBTC wallets tools.
+var WalletCommand = &cobra.Command{
+	Use:              "wallet",
+	Short:            "tBTC wallets tools",
+	Long:             "The tool exposes commands for interactions with tBTC wallets.",
+	TraverseChildren: true,
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		if err := clientConfig.ReadConfig(
+			configFilePath,
+			cmd.Flags(),
+			config.General, config.Ethereum, config.BitcoinElectrum,
+		); err != nil {
+			logger.Fatalf("error reading config: %v", err)
+		}
+	},
+}
+
+var depositsCommand = cobra.Command{
+	Use:              "deposits",
+	Short:            "get deposits",
+	Long:             "Gets tBTC deposits details from the chain and prints them.",
+	TraverseChildren: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		ctx := cmd.Context()
+
+		wallet, err := cmd.Flags().GetString(walletFlagName)
+		if err != nil {
+			return fmt.Errorf("failed to find wallet flag: %v", err)
+		}
+
+		hideSwept, err := cmd.Flags().GetBool(hideSweptFlagName)
+		if err != nil {
+			return fmt.Errorf("failed to find show swept flag: %v", err)
+		}
+
+		sortByAmount, err := cmd.Flags().GetBool(sortByAmountFlagName)
+		if err != nil {
+			return fmt.Errorf("failed to find show swept flag: %v", err)
+		}
+
+		_, tbtcChain, _, _, _, err := ethereum.Connect(ctx, clientConfig.Ethereum)
+		if err != nil {
+			return fmt.Errorf(
+				"could not connect to Bitcoin difficulty chain: [%v]",
+				err,
+			)
+		}
+
+		return walletcmd.ListDeposits(tbtcChain, wallet, hideSwept, sortByAmount)
+	},
+}
+
+func init() {
+	initFlags(
+		WalletCommand,
+		&configFilePath,
+		clientConfig,
+		config.General, config.Ethereum, config.BitcoinElectrum,
+	)
+
+	// Wallet Command
+	WalletCommand.PersistentFlags().String(
+		walletFlagName,
+		"",
+		"wallet public key hash",
+	)
+
+	// Deposits Subcommand
+	depositsCommand.Flags().Bool(
+		hideSweptFlagName,
+		false,
+		"hide swept deposits",
+	)
+
+	depositsCommand.Flags().Bool(
+		sortByAmountFlagName,
+		false,
+		"sort by deposit amount",
+	)
+
+	WalletCommand.AddCommand(&depositsCommand)
+}
--- a/cmd/wallet.go
+++ b/cmd/wallet.go
@@ -5,10 +5,10 @@ import (
 
 	"github.com/spf13/cobra"
 
-	walletcmd "github.com/keep-network/keep-core/cmd/wallet"
 	"github.com/keep-network/keep-core/config"
 	"github.com/keep-network/keep-core/pkg/bitcoin/electrum"
 	"github.com/keep-network/keep-core/pkg/chain/ethereum"
+	"github.com/keep-network/keep-core/pkg/coordinator"
 )
 
 var (
@@ -82,7 +82,7 @@ var depositsCommand = cobra.Command{
 			return fmt.Errorf("could not connect to Electrum chain: [%v]", err)
 		}
 
-		return walletcmd.ListDeposits(
+		return coordinator.ListDeposits(
 			tbtcChain,
 			btcChain,
 			wallet,
--- a/cmd/wallet.go
+++ b/cmd/wallet.go
@@ -7,6 +7,7 @@ import (
 
 	walletcmd "github.com/keep-network/keep-core/cmd/wallet"
 	"github.com/keep-network/keep-core/config"
+	"github.com/keep-network/keep-core/pkg/bitcoin/electrum"
 	"github.com/keep-network/keep-core/pkg/chain/ethereum"
 )
 
@@ -59,12 +60,23 @@ var depositsCommand = cobra.Command{
 		_, tbtcChain, _, _, _, err := ethereum.Connect(ctx, clientConfig.Ethereum)
 		if err != nil {
 			return fmt.Errorf(
-				"could not connect to Bitcoin difficulty chain: [%v]",
+				"could not connect to Ethereum chain: [%v]",
 				err,
 			)
 		}
 
-		return walletcmd.ListDeposits(tbtcChain, wallet, hideSwept, sortByAmount)
+		btcChain, err := electrum.Connect(ctx, clientConfig.Bitcoin.Electrum)
+		if err != nil {
+			return fmt.Errorf("could not connect to Electrum chain: [%v]", err)
+		}
+
+		return walletcmd.ListDeposits(
+			tbtcChain,
+			btcChain,
+			wallet,
+			hideSwept,
+			sortByAmount,
+		)
 	},
 }
 
--- a/cmd/wallet/deposits.go
+++ b/cmd/wallet/deposits.go
@@ -25,17 +25,22 @@ type depositEntry struct {
 	fundingTransactionHash        bitcoin.Hash
 	fundingTransactionOutputIndex uint32
 	amountBtc                     float64
-	// TODO: Add Bitcoin confirmations number
+	confirmations                 uint
 }
 
 // ListDeposits gets deposits from the chain.
 func ListDeposits(
 	tbtcChain tbtc.Chain,
+	btcChain bitcoin.Chain,
 	walletPublicKeyHashString string,
 	hideSwept bool,
 	sortByAmount bool,
 ) error {
-	deposits, err := getDeposits(tbtcChain, walletPublicKeyHashString)
+	deposits, err := getDeposits(
+		tbtcChain,
+		btcChain,
+		walletPublicKeyHashString,
+	)
 	if err != nil {
 		return fmt.Errorf(
 			"failed to get deposits: [%w]",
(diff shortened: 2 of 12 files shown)


cmd/wallet/deposits.go
```diff
@@ -0,0 +1,189 @@
...
+	if len(walletStr) > 0 {
+		walletPublicKeyHash, err := hexToWalletPublicKeyHash(walletStr)
+		if err != nil {
+			return []depositEntry{}, fmt.Errorf("failed to extract wallet public key hash: %v", err)
+		}
+
+		filter.WalletPublicKeyHash = [][20]byte{walletPublicKeyHash}
+	}
+
+	depositRevealedEvent, err := tbtcChain.PastDepositRevealedEvents(filter)
```
pdyraga, line 97: Did you manage to test it on mainnet with the oldest wallet? Did you try calling it against geth server which is not Infura/Alchemy? If not, can you please perform such a test? We had a [discussion](https://github.com/keep-network/keep-core/pull/3521/files#r1173592721) with @lukasz-zimnoch on how long it takes to retrieve old events without passing a block range. I expect Alchemy and Infura may have some fancier cache/indexing but does it work fast enough with a standard geth? Knowing this would help us determine how fancy we need to be with calls like that and this simple CLI command is a perfect opportunity.

cmd/wallet/deposits.go
```diff
@@ -0,0 +1,189 @@
...
+) string {
+	fundingOutputIndexBytes := make([]byte, 4)
+	binary.BigEndian.PutUint32(fundingOutputIndexBytes, fundingOutputIndex)
+
+	depositKey := crypto.Keccak256Hash(
+		append(fundingTxHash[:], fundingOutputIndexBytes...),
+	)
+
+	return depositKey.Hex()
+}
```
pdyraga, line 185: There is a `buildDepositKey` in `pkg/chain/ethereum/tbtc.go` and we are passing a handle to `tbtc.Chain` to `ListDeposits`. The code for building a deposit key is chain-specific so I would consider exposing a function in `BridgeChain interface` of `pkg/tbtc/chain.go` to `BuildDepositKey`.

cmd/wallet/deposits.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	if err := w.Flush(); err != nil {
+		return fmt.Errorf("failed to flush the writer: %v", err)
+	}
+
+	return nil
+}
+
+func hexToWalletPublicKeyHash(str string) ([20]byte, error) {
```
pdyraga, line 161: Feels like something we could place in `pkg/bitcoin` and import from here. Thoughts on the best location for this logic @lukasz-zimnoch?

lukasz-zimnoch: Not sure if `pkg/bitcoin` is the right place as it should only contain components specific to the Bitcoin protocol. This function is not such a component. It's rather a helper specific to `cmd` package and IMO, should remain as is.

pdyraga: I looked at this function one more time, it's just `string` -> `[20]byte` conversion with no hashing magic. Let's leave it as is. It was late in the evening and I was wrong.

cmd/wallet/deposits.go
```diff
@@ -0,0 +1,189 @@
...
+	walletHex, err := hexutil.Decode(str)
```
lukasz-zimnoch, line 162: I don't think we should use `github.com/ethereum/go-ethereum` package here as it makes the Ethereum-specific implementation details leak into a code that should be chain-agnostic. The only place this package can be used safely is `pkg/chain/ethereum`. Hence, I propose using `hex.DecodeString` instead. Just remember about stripping the `0x` in case it exists at the beginning.

cmd/wallet/deposits.go
lukasz-zimnoch: Agreed. This will allow us to get rid of the `github.com/ethereum/go-ethereum` import altogether.

lukasz-zimnoch (changes requested):

cmd/wallet.go
```diff
@@ -0,0 +1,100 @@
...
+			cmd.Flags(),
+			config.General, config.Ethereum, config.BitcoinElectrum,
+		); err != nil {
+			logger.Fatalf("error reading config: %v", err)
+		}
+	},
+}
+
+var depositsCommand = cobra.Command{
+	Use:              "deposits",
```
lukasz-zimnoch, line 37: Can we rename this command to `list-wallet-deposits`? This will make it more specific.

cmd/wallet/deposits.go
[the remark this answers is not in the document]
pdyraga: I was thinking about it as well and if this is a quick change, I'd go for it. 

> one day

It's not that far in time. This need to happen for redemptions.

cmd/wallet/deposits.go
```diff
@@ -0,0 +1,189 @@
...
+
+		depositRequest, err := tbtcChain.GetDepositRequest(event.FundingTxHash, event.FundingOutputIndex)
+		if err != nil {
+			return result, fmt.Errorf(
+				"failed to get deposit request: [%w]",
+				err,
+			)
+		}
+
+		result[i] = depositEntry{
```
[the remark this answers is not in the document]
nkuba: > If there is at least one deposit in the proposal that has no required confirmations, the entire proposal will be rejected by wallet members.

This is why we add proposal validation before submitting it in https://github.com/keep-network/keep-core/pull/3549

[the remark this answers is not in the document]
pdyraga: This makes sense and validating before submission makes sense as well. However, I'd like to know from the very beginning if I should even look at the given deposit and consider it for sweeping. This will be also important once we automate this job.

cmd/wallet/deposits.go
```diff
@@ -0,0 +1,189 @@
...
+
+	fundingTransactionHash        bitcoin.Hash
+	fundingTransactionOutputIndex uint32
+	amountBtc                     float64
+}
+
+// ListDeposits gets deposits from the chain.
+func ListDeposits(
+	tbtcChain tbtc.Chain,
+	walletStr string,
```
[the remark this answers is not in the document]
nkuba: 82b381cab4843ae4b8f2137cc768b228a4a17130

cmd/wallet/deposits.go
```diff
@@ -0,0 +1,189 @@
...
+
+	"github.com/keep-network/keep-core/pkg/bitcoin"
+	"github.com/keep-network/keep-core/pkg/tbtc"
+)
+
+type depositEntry struct {
+	wallet [20]byte
+
+	depositKey      string
+	revealedAtBlock uint64
```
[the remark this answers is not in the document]
nkuba: 82b381cab4843ae4b8f2137cc768b228a4a17130

cmd/wallet/deposits.go
```diff
@@ -0,0 +1,189 @@
...
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/crypto"
+
+	"github.com/keep-network/keep-core/pkg/bitcoin"
+	"github.com/keep-network/keep-core/pkg/tbtc"
+)
+
+type depositEntry struct {
+	wallet [20]byte
```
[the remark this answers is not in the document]
nkuba: 82b381cab4843ae4b8f2137cc768b228a4a17130

cmd/wallet/deposits.go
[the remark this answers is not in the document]
nkuba: Yes, sure, I understood that. It will be added in a follow-up.

cmd/wallet/deposits.go
nkuba: I did it in an initial version but then decided to just copy the `buildDepositKey` function as I thought that we will soon get rid of the CLI in favor of automated proposals and won't need the exported `BuildDepositKey` in the interface.

cmd/wallet.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/spf13/cobra"
+
+	walletcmd "github.com/keep-network/keep-core/cmd/wallet"
+	"github.com/keep-network/keep-core/config"
+	"github.com/keep-network/keep-core/pkg/chain/ethereum"
+)
+
+var (
+	walletFlagName       = "wallet"
```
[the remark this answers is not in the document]
nkuba: Will we have another identifier for the wallet? In the flag's description we are specifc what the value is.
```
--wallet string   wallet public key hash
```

cmd/wallet/deposits.go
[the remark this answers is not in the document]
nkuba: `cmd/wallet` is a temporary package just for now. I was thinking that on the **one day** we move the code to the `pkg/tbtc`, next to the `deposit_sweep.go`

cmd/wallet/deposits.go
nkuba: The reason why I used `hexutil` is that it handles prefixes. I'll refactor it to not relay on the `github.com/ethereum/go-ethereum`.

cmd/wallet/deposits.go
```diff
@@ -0,0 +1,189 @@
...
+				"failed to get deposit request: [%w]",
+				err,
+			)
+		}
+
+		result[i] = depositEntry{
+			wallet:                        event.WalletPublicKeyHash,
+			depositKey:                    depositKey,
+			revealedAtBlock:               event.BlockNumber,
+			isSwept:                       depositRequest.SweptAt.After(depositRequest.RevealedAt),
```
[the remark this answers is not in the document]
nkuba: It's a little more complicated as `SweptAt` is `time.Time`, so `depositRequest.SweptAt.Unix() != 0`

fc1cf3a95a8e4d26711143d8e9c736dc13b0224f

cmd/wallet.go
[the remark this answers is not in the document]
nkuba: I wanted to keep the command as short as possible as when we add all the flags and arguments it will get really long.

cmd/wallet.go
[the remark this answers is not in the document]
lukasz-zimnoch: Not directly here but in general, a wallet can be identified by:
- it's uncompressed public key
- it's compressed public key
- it's ECDSA wallet ID
- it's wallet public key hash

This is why I proposed to make it specific 😀

cmd/wallet/deposits.go
nkuba: ### Get All Deposits (no filter):

**Mainnet:**
Alchemy: 0.205702 seconds
Infura: 0.523039 seconds
Geth: 265.665695 seconds

### Get Wallet's Deposits (with wallet filter):

**Mainnet:**
Alchemy: 0.186728 seconds
Infura: 0.271587 seconds
Geth: 114.994290 seconds

cmd/wallet/deposits.go
lukasz-zimnoch: IMO, we should do it anyway. Even if we automate it, this code will live in a chain-agnostic package (either `pkg/tbtc` or a dedicated one like `pkg/coordinator`). That means we would still need this function to be exposed by the chain interface.

cmd/wallet.go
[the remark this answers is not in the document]
lukasz-zimnoch: In my opinion, shorter doesn't always mean better. In this particular case, `wallet` is one of the most widespread names in the system. We use it in multiple contexts, so having a way to make it more specific, is always a good idea. That said, I think `wallet-coordinator` is quite a harmless way to say that this CLI is about wallet coordination and nothing more.

cmd/wallet/deposits.go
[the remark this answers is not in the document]
lukasz-zimnoch: Actually, we are not sure `pkg/tbtc` will be the right place for this code at the moment. We still need to determine the right way to automate the proposals. Having wallets self-coordinating is one option but definitely not something carved in stone yet. That said, I think having that code within `pkg/coordinator` (or smth similar) from the very beginning makes sense as this is a sane compromise between having it in `cmd/wallet` (purely a temporary place) and having it in `pkg/tbtc` (not sure if we move it there at all).

cmd/wallet.go
lukasz-zimnoch: I meant something similar to what I mentioned in https://github.com/keep-network/keep-core/pull/3545#discussion_r1187239035. In this case, `deposits` are also ubiquitous, and making it at least `list-deposits` would make the command clear at first sight, even without referring to its description.

cmd/wallet/deposits.go
nkuba: 6e4e90ad3ba3b0bd864ca0ccf991ee7392295450

cmd/wallet/deposits.go
nkuba: 3f4e2b3f01d8adf52ea32633485e043c240692ce

cmd/wallet.go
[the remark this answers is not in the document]
nkuba: a717cc2cdd6a7b9981ec9eeb4226f4aba8bfa76a

cmd/wallet.go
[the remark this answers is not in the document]
nkuba: I would stick with `--wallet` as for the user the public key hash seems to be a default wallet identifier used across tBTC.

cmd/wallet.go
lukasz-zimnoch: Eventually, the wallet coordinator will manage redemptions, moving funds, and sweeps of moved funds. Most likely, the CLI will have to handle them as well. This is why I'm flagging the need to establish a clear and consistent API of the CLI from the beginning. I'm afraid that we won't be able to keep them short in the long term so it seems that it's better to use a predictable pattern from the beginning. For example, sweeping refers to both deposits and moved funds so we can't just have a single 'sweep' command.

pkg/internal/hexutils/hexutils.go
```diff
@@ -0,0 +1,35 @@
+package hexutils
+
+import (
+	"encoding/hex"
+	"fmt"
+)
+
+// Decode decodes a hex string with 0x prefix.
+func Decode(input string) ([]byte, error) {
```
pdyraga, line 9: Non-blocking of course but just noting we could move this code to `keep-common` at some point and use it [here](https://github.com/keep-network/keep-common/pull/116#discussion_r1181538174).

cmd/wallet.go
pdyraga: I would also vote for `list-deposits`. We will have more actions in the coordinator soon and `list-deposits` is in my opinion short enough and unique to not confuse it with anything else.

The actions we'll have to implement in the coordinator:
- list deposits (this PR)
- sweep (#3549)
- increase sweep fee
- list redemptions (will also have to be automated)
- redeem (will also have to be automated)
- increase redemption fee

... then all the actions related to moving funds. I think it would also be nice to allow submitting heartbeat requests via the coordinator CLI.

pdyraga (approved):
Tested locally and works as expected. There are some ongoing discussions but I find them non-blocking, so let's address them in #3549.

pkg/internal/hexutils/hexutils.go
lukasz-zimnoch: Side note: we can use strings.TrimPrefix to do prefix removal.

lukasz-zimnoch (approved):

pkg/internal/hexutils/hexutils.go
nkuba: We would need to trim `0x` and `0X`, so I'm not sure which approach would be better.

pkg/internal/hexutils/hexutils_test.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+func TestDecode(t *testing.T) {
+	for _, test := range decodeBytesTests {
+		t.Run(test.input, func(t *testing.T) {
+			dec, err := Decode(test.input)
+			if !reflect.DeepEqual(err, test.wantErr) {
+				t.Fatalf("unexpected error\nexpected: %v\nactual:   %v", test.wantErr, err)
+			}
+			if !bytes.Equal(test.want, dec) {
```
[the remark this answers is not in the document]
nkuba: d4f1c1ffe875311c4786b01a4c83a32432fff5ea

cmd/wallet.go
nkuba: Discussion continued in https://github.com/keep-network/keep-core/pull/3549#discussion_r1185811670

PR #3550, opened 2023-05-04
Leverage deposits reveal blocks passed within the sweep proposal

lukasz-zimnoch:

This changeset simplifies the deposit sweep action by leveraging the deposits reveal blocks that are now part of the deposit sweep proposal and are set by the coordinator. This way, the action does not have to determine the reveal block on their own. That reduces the execution time and the number of points of failure.

Moreover, the action does no longer need to fetch the deposit request from the `Bridge` (the `GetDepositRequest` call) because the `PastDepositRevealedEvents` now targets just one block, and it is enough to filter by wallet PKH. Additional filtering using the depositor address is redundant.


pkg/tbtc/deposit_sweep.go
```diff
@@ -185,6 +184,10 @@ func (dsa *depositSweepAction) validateProposal(
 
 	validateProposalLogger.Infof("gathering prerequisites for proposal validation")
 
+	if len(dsa.proposal.DepositsKeys) != len(dsa.proposal.DepositsRevealBlocks) {
+		return nil, fmt.Errorf("proposal's reveal blocks list has a wrong length")
```
pdyraga, line 188: Not a big one but we could print the expected and actual length in the error.

pkg/tbtc/deposit_sweep.go
```diff
@@ -259,21 +241,12 @@ func (dsa *depositSweepAction) validateProposal(
...
-
+		// We have the revealBlock passed by the coordinator within the proposal
+		// so, we can use it to make a narrow call. Moreover, we use the
+		// wallet PKH as additional filter to limit the size of returned data.
 		events, err := dsa.chain.PastDepositRevealedEvents(&DepositRevealedEventFilter{
-			StartBlock:          startBlock,
-			EndBlock:            &endBlock,
-			Depositor:           []chain.Address{depositRequest.Depositor},
+			StartBlock:          revealBlock,
+			EndBlock:            &revealBlock,
```
pdyraga, line 249: Good call on making it narrow. We could in theory allow for the `revealBlockMargin` but... why. The coordinator should pass the right data.

pdyraga (approved):

PR #3551, opened 2023-05-04
Fix deposit script function

lukasz-zimnoch:
The `hex.DecodeString` expects an unprefixed hex string while the `d.Depositor` contains the `0x` prefix by default. To make it work, we need to trim that prefix.


pdyraga (approved):

pdyraga:
Let's do not block on it for now but would be great to cover it with a unit test in a separate PR.

PR #3561, opened 2023-05-11
Increase the gas limit for retarget with refund

tomaszslabon:
This PR increases the gas limit for calls to `LightRelayMaintainerProxy`'s `retarget` function.
The gas limit used is the gas estimate increased by 20%. 
The transactions would revert if the default limit was used.

### DIFF
--- a/pkg/chain/ethereum/bitcoin_difficulty.go
+++ b/pkg/chain/ethereum/bitcoin_difficulty.go
@@ -187,11 +187,25 @@ func (bdc *BitcoinDifficultyChain) RetargetWithRefund(headers []*bitcoin.BlockHe
 		serializedHeaders = append(serializedHeaders, serializedHeader[:]...)
 	}
 
+	gasEstimate, err := bdc.lightRelayMaintainerProxy.RetargetGasEstimate(
+		serializedHeaders,
+	)
+	if err != nil {
+		return fmt.Errorf(
+			"failed to estimate gas for retarget with refund: [%w]",
+			err,
+		)
+	}
+
+	// Add 20% to the gas estimate as the transaction tends to fail with the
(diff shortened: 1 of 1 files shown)


pkg/chain/ethereum/bitcoin_difficulty.go
```diff
@@ -187,7 +188,12 @@ func (bdc *BitcoinDifficultyChain) RetargetWithRefund(headers []*bitcoin.BlockHe
 	}
 
 	// Update Bitcoin difficulty via LightRelayMaintainerProxy.
-	_, err := bdc.lightRelayMaintainerProxy.Retarget(serializedHeaders)
+	_, err := bdc.lightRelayMaintainerProxy.Retarget(
```
tomaszslabon, line 191: I only increased the limit for LightRelayMaintainerProxy's retarget function.
I left the LightRelay's retarget function unchanged as it's possible the function could succeed with the default gas limit.

pkg/chain/ethereum/bitcoin_difficulty.go
```diff
@@ -187,7 +188,12 @@ func (bdc *BitcoinDifficultyChain) RetargetWithRefund(headers []*bitcoin.BlockHe
...
+		serializedHeaders,
+		ethutil.TransactionOptions{
+			GasLimit: 310000,
```
pdyraga, line 194: Instead of hardcoding, I'd take the estimation and add some % on top. Hardcoding the gas limit may stop working if EVM opcode call costs change.

pdyraga (approved):

pdyraga:
@tomaszslabon The code looks good and I am good to merge but please update the PR description to reflect the recent changes first.

PR #3568, opened 2023-05-16
Reimburse the coordinator for the deposit sweep proposal

### LINKED ISSUE #3566: Use reimbursable version of the function when proposing deposit sweep
From @lukasz-zimnoch:

> The coordinator CLI uses the non-reimbursable version of submitDepositSweepProposal so the coordinator doesn't receive the reimbursement by default. I think we need to use the reimbursable version by default and use a flag allowing to call the non-reimbursable function as well.

https://github.com/keep-network/keep-core/blob/0a6a7ae63e1afbd1488b835e7fb276adf2843f00/pkg/coordinator/sweep.go#L72

pdyraga:
`SubmitDepositSweepProposalWithReimbursement` should be called instead of `SubmitDepositSweepProposal` to reimburse the gas cost to the coordinator.

Closes #3566

### DIFF
--- a/pkg/tbtc/chain.go
+++ b/pkg/tbtc/chain.go
@@ -333,8 +333,8 @@ type WalletCoordinatorChain interface {
 		},
 	) error
 
-	// SubmitDepositSweepProposal submits a deposit sweep proposal to the chain.
-	// It reimburses the gas cost to the caller.
+	// SubmitDepositSweepProposalWithReimbursement submits a deposit sweep
+	// proposal to the chain. It reimburses the gas cost to the caller.
 	SubmitDepositSweepProposalWithReimbursement(
 		proposal *DepositSweepProposal,
 	) error


pkg/tbtc/chain.go
```diff
@@ -334,7 +334,10 @@ type WalletCoordinatorChain interface {
 	) error
 
 	// SubmitDepositSweepProposal submits a deposit sweep proposal to the chain.
```
nkuba, line 336:
```suggestion
	// SubmitDepositSweepProposalWithReimbursement submits a deposit sweep proposal to the chain.
```

nkuba (changes requested):

pkg/tbtc/chain.go
pdyraga: be2a6dc59eccbf62e8edf881e5bda0c48215ed5d

PR #3569, opened 2023-05-16
Don't include stub contracts in NPM packages

### LINKED ISSUE #3531: Remove *Stub contracts from @development NPM package
# The problem

We have the [`WalletRegistryStub` contract](https://github.com/keep-network/keep-core/blob/main/solidity/ecdsa/contracts/test/WalletRegistryStub.sol). This contract lands in the `@development` NPM package and is used by the client to generate bindings when running `make all`. The additional functionalities of the stub, `forceAddWalle`t and `getDkgData`, are used only in unit tests of `solidity/ecdsa`.

# The context

We have two deployment scripts defined:
https://github.com/keep-network/keep-core/blob/0484ddbc1c19715afb897a914d0f0f633421ba78/solidity/ecdsa/package.json#L31-L32

`deploy:test` is used in tests and deploys the `WalletRegistryStub` with additional functions. `deploy `is used in regular deployments. For example: https://github.com/keep-network/keep-core/blob/0484ddbc1c19715afb897a914d0f0f633421ba78/.github/workflows/contracts-ecdsa.yml#L328

For some reason, we decided to use `deploy:test` for the development NPM package publication:
https://github.com/keep-network/keep-core/blob/0484ddbc1c19715afb897a914d0f0f633421ba78/.github/workflows/npm-ecdsa.yml#L40-L43

That's why `WalletRegistryStub` gets into the `@development` package.

# The solution

Having `@development` as a `deploy` from `main` (with no stubs) makes more sense. `deploy:test` should only be used for tests.

michalinacienciala:
If `hardhat deploy` is run with `TEST_USE_STUBS_ECDSA` environment variable set to `true`, the deployment will include deployment of stub contracts, which are needed for testing purposes (execution of unit tests). So far we've been running `hardhat deploy TEST_USE_STUBS_ECDSA=true` together with `USE_EXTERNAL_DEPLOY=true` in `ecdsa` in the `deploy:test` script. The `USE_EXTERNAL_DEPLOY` variable specifies how the contracts of the dependencies should be deployed - whether to deploy them from scratch (`true`) or reuse the already deployed artifacts (`false`).
We were using the `deploy:test` script in `ecdsa` in two places - when running `contracts-deployment-dry-run` job (test of deployment on `hardhat` local network) and in `npm-compile-publish-contracts` job (publishing of `development`-tagged NPM package). This second use proved to be problematic, as the `development`-tagged package is used to generate the client bindings when running `make all`, resulting in the presence of unneeded (and unwanted) functionalities (like `forceAddWallet` and `getDkgData` from `WalletRegistryStub`). The first use is not causing such problems, but it should still be modified, as it's better to run the dry-run of deploy on the real contracts and not on the stubs.
What we decided to do is to replace
```
"deploy:test": "USE_EXTERNAL_DEPLOY=true TEST_USE_STUBS_ECDSA=true hardhat deploy",
```
with
```
"deploy:local": "USE_EXTERNAL_DEPLOY=true hardhat deploy",
```
and use the new `deploy:local` script in both jobs that previously used `deploy:test`.
This way we'll no longer include stub functionalities in NPM packages (and hence in the clinet bindings) and we'll execute dry-run of the unmodified contracts. Scripts in `random-beacon` also got updated to follow the new naming.

Refs:
https://github.com/keep-network/keep-core/issues/3531
https://github.com/keep-network/tbtc-v2/pull/623

### DIFF
--- a/.github/workflows/contracts-ecdsa.yml
+++ b/.github/workflows/contracts-ecdsa.yml
@@ -175,7 +175,7 @@ jobs:
         run: yarn install --frozen-lockfile
 
       - name: Deploy contracts
-        run: yarn deploy:test
+        run: yarn deploy:local
 
       - name: Build Docker Image
         uses: ./.github/actions/docker-build-push
--- a/.github/workflows/contracts-random-beacon.yml
+++ b/.github/workflows/contracts-random-beacon.yml
@@ -173,7 +173,7 @@ jobs:
         run: yarn install --network-concurrency 1 --frozen-lockfile
 
       - name: Deploy contracts
-        run: yarn deploy:test
+        run: yarn deploy:local
 
       - name: Build Docker Image
         uses: ./.github/actions/docker-build-push
--- a/.github/workflows/npm-ecdsa.yml
+++ b/.github/workflows/npm-ecdsa.yml
@@ -38,9 +38,9 @@ jobs:
             @threshold-network/solidity-contracts
 
       # Deploy contracts to a local network to generate deployment artifacts that
-      # are required by dashboard and client compilation.
+      # are required by client compilation.
       - name: Deploy contracts
-        run: yarn deploy:test --network hardhat --write true
+        run: yarn deploy:local --network hardhat --write true
 
       - name: Bump up package version
         id: npm-version-bump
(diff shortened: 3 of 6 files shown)


lionakhnazarov:
Closing — targets solidity-v1 NPM packaging, which is being retired (#3947) and extracted (#3998). No longer applicable to current main.

Issue #3565, opened 2023-05-16
Select Electrum server when the client starts


pdyraga:
Starting from `v2.0.0-m3`, the client needs the connection to the Electrum server. [TIP-045: Bootstrap Node Proposal v2](https://forum.threshold.network/t/tip-045-bootstrap-node-proposal-v2/492) should allow us to use 3 Electrum servers from staking providers. So far, only Boar provided Electrum server, so we have three servers available to choose from: 
- Boar's Electrum,
- Keep's Electrum,
- Blockstream's public Electrum.

Ideally, this list would be embedded in the client config just like the list of mainnet bootstrap servers. The client would select from this list randomly but would also allow overwriting the Electrum server [in the configuration](https://github.com/keep-network/keep-core/blob/0a6a7ae63e1afbd1488b835e7fb276adf2843f00/configs/config.toml.SAMPLE#L49-L51).

We should apply the change to the client code and improve docs.threshold.network to explain Electrum configuration and how the client learns about Bitcoin state.

PR #3576, opened 2023-05-18
Embed Electrum configs for Bitcoin mainnet and testnet

lukasz-zimnoch:
Closes: https://github.com/keep-network/keep-core/issues/3565

Here we add some default Electrum configs that can be used for Bitcoin mainnet and testnet networks in case the Electrum `URL` is not explicitly provided in the `bitcoin.electrum` section of the configuration file. In such a case, one of the default embedded configs is picked up randomly.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"github.com/keep-network/keep-core/pkg/bitcoin"
 	"os"
 	"strings"
 	"syscall"
@@ -47,6 +48,7 @@ type Config struct {
 
 // BitcoinConfig defines the configuration for Bitcoin.
 type BitcoinConfig struct {
+	bitcoin.Network
 	// Electrum defines the configuration for the Electrum client.
 	Electrum electrum.Config
 }
@@ -60,28 +62,32 @@ func bindFlags(flagSet *pflag.FlagSet) error {
 	return nil
 }
 
-// Resolve ethereum network based on the set boolean flags.
-func (c *Config) resolveEthereumNetwork(flagSet *pflag.FlagSet) error {
+// Resolve Ethereum and Bitcoin networks based on the set boolean flags.
+func (c *Config) resolveChainsNetworks(flagSet *pflag.FlagSet) error {
 	var err error
 
-	c.Ethereum.Network, err = func() (commonEthereum.Network, error) {
+	c.Ethereum.Network, c.Bitcoin.Network, err = func() (
+		commonEthereum.Network,
+		bitcoin.Network,
+		error,
+	) {
 		isGoerli, err := flagSet.GetBool(commonEthereum.Goerli.String())
 		if err != nil {
-			return commonEthereum.Unknown, err
+			return commonEthereum.Unknown, bitcoin.Unknown, err
 		}
 		if isGoerli {
-			return commonEthereum.Goerli, nil
+			return commonEthereum.Goerli, bitcoin.Testnet, nil
 		}
 
 		isDeveloper, err := flagSet.GetBool(commonEthereum.Developer.String())
 		if err != nil {
-			return commonEthereum.Unknown, err
+			return commonEthereum.Unknown, bitcoin.Unknown, err
 		}
 		if isDeveloper {
-			return commonEthereum.Developer, nil
+			return commonEthereum.Developer, bitcoin.Regtest, nil
 		}
 
-		return commonEthereum.Mainnet, nil
+		return commonEthereum.Mainnet, bitcoin.Mainnet, nil
 	}()
 
 	return err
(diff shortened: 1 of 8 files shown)


config/electrum.go
```diff
@@ -0,0 +1,98 @@
...
+	neturl "net/url"
+	"strings"
+)
+
+//go:embed _electrum_urls/*
+var electrumURLs embed.FS
+
+// readElectrumConfigs reads Electrum configs from an embedded file for the
+// given Bitcoin network.
+func readElectrumConfigs(network bitcoin.Network) ([]electrum.Config, error) {
```
pdyraga, line 18: `*Configs` or `*Config`? Even if we had multiple entries in the embedded file, we'd still be reading "a configuration" for Electrum connection.

config/config.go
```diff
@@ -60,28 +62,32 @@ func bindFlags(flagSet *pflag.FlagSet) error {
...
-			return commonEthereum.Unknown, err
+			return commonEthereum.Unknown, bitcoin.Unknown, err
 		}
 		if isDeveloper {
-			return commonEthereum.Developer, nil
+			return commonEthereum.Developer, bitcoin.Regtest, nil
 		}
 
-		return commonEthereum.Mainnet, nil
+		return commonEthereum.Mainnet, bitcoin.Mainnet, nil
```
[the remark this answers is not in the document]
lukasz-zimnoch: > We should rename --goerli flag to --testnet

We could but certainly not this PR because this is a broad change that impacts client deployment. On the other hand, I think that it makes sense to have explicit Ethereum networks instead of a general `--testnet` flag. If we want to add, for example, Sepolia, we will add another flag in parallel so we can have two Ethereum testnets independently.

> The fact --goerli dictates Bitcoin network used is confusing.

Why? It makes sense to use Bitcoin testnet with Ethereum testnets, Bitcoin mainnet with Ethereum mainnet etc. The only exception that makes sense is using Bitcoin testnet with Ethereum local chain, but then we can always explicitly set the `bitcoin.ethereum` section.

config/electrum.go
lukasz-zimnoch: I think the current version makes sense as this function returns a list of possible configurations. Then `resolveElectrum` picks up one randomly.

config/config.go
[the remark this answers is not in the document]
pdyraga: As someone not familiar with this PR - say another client developer or user - I would be confused about `--goerli` flag.  This flag clearly connects the client to Goerli Ethereum testnet, but how do I select Bitcoin testnet? 

Regarding `--sepolia` and `--goerli`, the argument makes sense but in reality, we will support only one testnet network. 

This is not the current state, but ultimately, I'd like to have one client binary where I select `--testnet` for Sepolia (will replace Goerli) + Bitcoin testnet, and `--developer` for local Ethereum and regtest. Mainnet Ethereum + mainnet Bitcoin would be the default one. Renaming `--goerli` to `--testnet` is the first step. The next step is figuring out the build process (out of scope currently) so that we build one client binary instead of testnet-specific and mainnet-specific versions.

config/electrum.go
pdyraga: I see. I got confused about what is actually returned. Why don't we rename `readElectrumConfigs` to `readElectrumUrls` and change its returned type? See that we only return host + protocol, and later in `resolveElectrum` we ignore all other settings that would potentially come in the config returned by `readElectrumConfigs`:

https://github.com/keep-network/keep-core/blob/51127b93140c795409116b42948f5fe0dad6f181/config/electrum.go#L92-L95

Side note (cc @nkuba - maybe we can fix it as a part of the work on the `wss://` support?), our config for `[bitcoin.Electrum]` is not consistent with the config for `[ethereum]`.

Compare:

https://github.com/keep-network/keep-core/blob/51127b93140c795409116b42948f5fe0dad6f181/configs/config.toml.SAMPLE#L3-L4

https://github.com/keep-network/keep-core/blob/51127b93140c795409116b42948f5fe0dad6f181/configs/config.toml.SAMPLE#L49-L56

Scheme (`wss://`,`ssl://`, `tcp://`) should [be a part of the URL](https://www.rfc-editor.org/rfc/rfc1738#section-5) just like we have it for `[ethereum]`. If we want to resolve the protocol later, this should be an auxiliary function on the `Config` type parsing the protocol from the URL and not a separate field:

https://github.com/keep-network/keep-core/blob/51127b93140c795409116b42948f5fe0dad6f181/pkg/bitcoin/electrum/config.go#L27-L28

If we do this change, then we can return just an URL from `readElectrumUrls` without bothering about tuples or introducing other types.

config/config.go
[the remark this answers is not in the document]
lukasz-zimnoch: Ok makes sense. Added a TODO c31590d5

config/electrum.go
lukasz-zimnoch: > Why don't we rename readElectrumConfigs to readElectrumUrls and change its returned type? See that we only return host + protocol, and later in resolveElectrum we ignore all other settings that would potentially come in the config returned by readElectrumConfigs:

That was my first approach but it was better to return `[]electrum.Config` than an array of tuples (note that we need to parse `electrum.Protocol`). However, if we plan to merge `electrum.URL` with `electrum.Protocol` in the config, here is the intermediary solution: cf266a69

config/config.go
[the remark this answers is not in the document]
pdyraga: Captured this work here: https://github.com/keep-network/keep-core/issues/3579

config/electrum.go
pdyraga: Captured it in https://github.com/keep-network/keep-core/issues/3580. I will keep this thread open for reference.

pdyraga (approved):

Issue #3562, opened 2023-05-15
Suggest Bitcoin fee for the sweeping coordinator


pdyraga:
Add a CLI command - or make it a part of the list deposits command - to print the suggested Bitcoin network fee the coordinator can use to propose a deposit sweep. Make sure this amount is not higher than the maximum one allowed by the Bridge.

Calculating a precise value is hard: we'd need to assemble the sweep transaction and then multiply knowing the sat per vByte value. We can do simpler though: we know what is the number of inputs and outputs. We can estimate the number of bytes and provide the value based on that.

As a Coordinator, I want to be able to copy-paste this value to propose a sweep. Please take into account what is the expected input for the sweep proposition and what is the expected format.

PR #3578, opened 2023-05-19
Added GetTransactionMerkle to Electrum client

tomaszslabon:
This PR adds `GetTransactionMerkleProof` function to the Electrum client implementation.
This function is needed to assemble SPV proof.

PR #3581, opened 2023-05-22
CLI for estimating deposit sweep Bitcoin fee

lukasz-zimnoch:
Closes: https://github.com/keep-network/keep-core/issues/3562

Here we introduce the CLI command allowing us to estimate fees for deposit sweep Bitcoin transactions. 

To estimate fees for all possible deposit sweep transactions (from 1 deposit up to the maximum sweep size allowed by the `WalletCoordinator` contract), invoke:
```
keep-client coordinator estimate-deposits-sweep-fee
```
Following flags can be used with the command:

- `--deposits-count <number>` - get the fee estimate only for deposit sweep transaction containing the specific count of deposits inputs

The command requires `ethereum` and `bitcoin` node details to be provided in the config.

### Examples
Sample execution of the command for Goerli network:
```
 ./keep-client --config ./configs/config.toml \
    --goerli \
    coordinator \
    estimate-deposits-sweep-fee 
```
Output:
```
 deposits count fee (satoshis)
              1            201
              2            292
              3            384
              4            475
              5            566
              6            657
              7            749
              8            840
              9            931
             10           1022
             11           1114
             12           1205
             13           1296
             14           1387
             15           1479
             16           1570
             17           1661
             18           1752
             19           1844
             20           1935

```

Sample execution of the command with `--deposits-count` flag for Goerli network:
```
 ./keep-client --config ./configs/config.toml \
    --goerli \
    coordinator \
    estimate-deposits-sweep-fee \
    --deposits-count 5
```
Output:
```
 deposits count fee (satoshis)
              5            566
```

### DIFF
--- a/cmd/coordinator.go
+++ b/cmd/coordinator.go
@@ -12,13 +12,14 @@ import (
 )
 
 var (
-	walletFlagName       = "wallet"
-	hideSweptFlagName    = "hide-swept"
-	sortByAmountFlagName = "sort-amount"
-	headFlagName         = "head"
-	tailFlagName         = "tail"
-	feeFlagName          = "fee"
-	dryRunFlagName       = "dry-run"
+	walletFlagName        = "wallet"
+	hideSweptFlagName     = "hide-swept"
+	sortByAmountFlagName  = "sort-amount"
+	headFlagName          = "head"
+	tailFlagName          = "tail"
+	feeFlagName           = "fee"
+	dryRunFlagName        = "dry-run"
+	depositsCountFlagName = "deposits-count"
 )
 
 // CoordinatorCommand contains the definition of tBTC Wallet Coordinator tools.
@@ -53,12 +54,12 @@ var listDepositsCommand = cobra.Command{
 
 		hideSwept, err := cmd.Flags().GetBool(hideSweptFlagName)
 		if err != nil {
-			return fmt.Errorf("failed to find show swept flag: %v", err)
+			return fmt.Errorf("failed to find hide swept flag: %v", err)
 		}
 
 		sortByAmount, err := cmd.Flags().GetBool(sortByAmountFlagName)
 		if err != nil {
-			return fmt.Errorf("failed to find show swept flag: %v", err)
+			return fmt.Errorf("failed to find sort by amount flag: %v", err)
 		}
 
 		head, err := cmd.Flags().GetInt(headFlagName)
@@ -132,7 +133,7 @@ var proposeDepositsSweepCommand = cobra.Command{
 
 		dryRun, err := cmd.Flags().GetBool(dryRunFlagName)
 		if err != nil {
-			return fmt.Errorf("failed to find fee flag: %v", err)
+			return fmt.Errorf("failed to find dry run flag: %v", err)
 		}
 
 		_, tbtcChain, _, _, _, err := ethereum.Connect(cmd.Context(), clientConfig.Ethereum)
@@ -158,6 +159,51 @@ Expects --wallet and --fee flags along with deposits to sweep provided
 as arguments.
 
 ` + coordinator.DepositsFormatDescription
+
+var estimateDepositsSweepFeeCommand = cobra.Command{
+	Use:              "estimate-deposits-sweep-fee",
+	Short:            "estimates deposits sweep fee",
+	Long:             estimateDepositsSweepFeeCommandDescription,
+	TraverseChildren: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		ctx := cmd.Context()
+
+		depositsCount, err := cmd.Flags().GetInt(depositsCountFlagName)
+		if err != nil {
+			return fmt.Errorf("failed to find deposits count flag: %v", err)
+		}
+
+		_, tbtcChain, _, _, _, err := ethereum.Connect(ctx, clientConfig.Ethereum)
+		if err != nil {
+			return fmt.Errorf(
+				"could not connect to Ethereum chain: [%v]",
+				err,
+			)
+		}
+
+		btcChain, err := electrum.Connect(ctx, clientConfig.Bitcoin.Electrum)
+		if err != nil {
+			return fmt.Errorf("could not connect to Electrum chain: [%v]", err)
+		}
+
+		return coordinator.EstimateDepositsSweepFee(
+			tbtcChain,
+			btcChain,
+			depositsCount,
+		)
+	},
+}
+
+var estimateDepositsSweepFeeCommandDescription = "Estimates the satoshi fee for the entire Bitcoin deposits sweep " +
+	"transaction, based on the number of input deposits. By default, " +
+	"provides estimations for transactions containing a various number" +
+	"of input deposits, from 1 up to the maximum count allowed by the " +
+	"WalletCoordinator contract. The --deposits-count flag can be " +
+	"used to obtain a fee estimation for a Bitcoin sweep transaction" +
+	"containing a specific count of input deposits. All estimations" +
+	"assume the wallet main UTXO is used as one of the transaction's input" +
+	"so the estimation may be overpriced for the very first sweep " +
+	"transaction of each wallet."
 
 func init() {
 	initFlags(
@@ -224,4 +270,14 @@ func init() {
 	)
 
 	CoordinatorCommand.AddCommand(&proposeDepositsSweepCommand)
+
+	// Estimate Deposits Sweep Fee Subcommand.
+
+	estimateDepositsSweepFeeCommand.Flags().Int(
+		depositsCountFlagName,
+		0,
+		"get estimation for a specific count of input deposits",
+	)
+
+	CoordinatorCommand.AddCommand(&estimateDepositsSweepFeeCommand)
 }
--- a/pkg/bitcoin/chain.go
+++ b/pkg/bitcoin/chain.go
@@ -46,4 +46,8 @@ type Chain interface {
 	// that pays the given public key hash using either a P2PKH or P2WPKH script.
 	// The returned transactions are in an indefinite order.
 	GetMempoolForPublicKeyHash(publicKeyHash [20]byte) ([]*Transaction, error)
+
+	// EstimateSatPerVByteFee returns the estimated sat/vbyte fee for a
+	// transaction to be confirmed within the given number blocks.
+	EstimateSatPerVByteFee(blocks uint32) (int64, error)
 }
--- a/pkg/bitcoin/electrum/electrum.go
+++ b/pkg/bitcoin/electrum/electrum.go
@@ -566,6 +566,38 @@ func (c *Connection) getScriptMempool(
 	return convertedItems, nil
 }
 
+// EstimateSatPerVByteFee returns the estimated sat/vbyte fee for a
+// transaction to be confirmed within the given number blocks.
+func (c *Connection) EstimateSatPerVByteFee(blocks uint32) (int64, error) {
+	// According to Electrum protocol docs, the returned fee is BTC/KB.
+	btcPerKbFee, err := requestWithRetry(
+		c,
+		func(
+			ctx context.Context,
+			client *electrum.Client,
+		) (float32, error) {
+			return client.GetFee(ctx, blocks)
+		})
+	if err != nil {
+		return 0, fmt.Errorf("failed to get fee: [%v]", err)
+	}
+
+	// According to Electrum protocol docs, if the daemon does not have
+	// enough information to make an estimate, the integer -1 is returned.
+	if btcPerKbFee < 0 {
+		return 0, fmt.Errorf(
+			"daemon does not have enough information to make an estimate",
+		)
+	}
+
+	// To convert from BTC/KB to sat/vbyte, we need to multiply by 1e8/1e3.
+	satPerVByte := (1e8 / 1e3) * float64(btcPerKbFee)
+	// Make sure the minimum returned sat/vbyte fee is always 1.
+	satPerVByte = math.Max(satPerVByte, 1)
+
+	return int64(math.Round(satPerVByte)), nil
+}
+
 func (c *Connection) electrumConnect() error {
 	var client *electrum.Client
 	var err error
--- a/pkg/bitcoin/estimator.go
+++ b/pkg/bitcoin/estimator.go
@@ -1,6 +1,7 @@
 package bitcoin
 
 import (
+	"fmt"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/mempool"
 	"github.com/btcsuite/btcd/txscript"
@@ -215,4 +216,43 @@ func (tse *TransactionSizeEstimator) VirtualSize() (int64, error) {
 	}
 
 	return mempool.GetTxVirtualSize(btcutil.NewTx(tse.internal.MsgTx)), nil
+}
+
+// TransactionFeeEstimator is a component allowing to estimate the total fee
+// for the given transaction virtual size.
+type TransactionFeeEstimator struct {
+	chain Chain
+}
+
+func NewTransactionFeeEstimator(chain Chain) *TransactionFeeEstimator {
+	return &TransactionFeeEstimator{chain: chain}
+}
+
+// EstimateFee estimates the total fee for the given transaction virtual size,
+// assuming the transaction is supposed to be confirmed within the next Bitcoin
+// block. The desired confirmation period can be adjusted using the optional
+// `blocks` argument.
+func (tfe *TransactionFeeEstimator) EstimateFee(
+	transactionVirtualSize int64,
+	blocks ...uint32,
+) (int64, error) {
+	resolvedBlocks := uint32(1)
+
+	if len(blocks) == 1 {
+		resolvedBlocks = blocks[0]
+	}
+
+	satPerVByteFee, err := tfe.chain.EstimateSatPerVByteFee(resolvedBlocks)
+	if err != nil {
+		return 0, fmt.Errorf("cannot get estimated sat/vbyte fee: [%v]", err)
+	}
+
+	fee := satPerVByteFee * transactionVirtualSize
+
+	// Just in case check.
+	if fee <= 0 {
+		return 0, fmt.Errorf("estimated fee is less than or equal zero")
+	}
+
+	return fee, nil
 }
--- a/pkg/bitcoin/estimator.go
+++ b/pkg/bitcoin/estimator.go
@@ -1,0 +1,218 @@
+package bitcoin
+
+import (
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/mempool"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+)
+
+// About half of all signatures generated with a random nonce are 72-byte, about
+// half are one byte smaller, and a small percentage are smaller than that.
+// For fee estimation purposes, we take the greatest possible value.
+var signaturePlaceholder = make([]byte, 72)
+
+// Compressed public keys have always 33 bytes.
+var publicKeyPlaceholder = make([]byte, 33)
+
+// TransactionSizeEstimator is a component allowing to estimate the size
+// of a Bitcoin transaction of the provided shape, without constructing it.
+type TransactionSizeEstimator struct {
+	internal *internalTransaction
+	err      error
+}
+
+func NewTransactionSizeEstimator() *TransactionSizeEstimator {
+	return &TransactionSizeEstimator{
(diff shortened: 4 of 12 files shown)


cmd/coordinator.go
```diff
@@ -12,13 +12,14 @@ import (
...
-	feeFlagName          = "fee"
-	dryRunFlagName       = "dry-run"
+	walletFlagName        = "wallet"
+	hideSweptFlagName     = "hide-swept"
+	sortByAmountFlagName  = "sort-amount"
+	headFlagName          = "head"
+	tailFlagName          = "tail"
+	feeFlagName           = "fee"
+	dryRunFlagName        = "dry-run"
+	depositsCountFlagName = "deposits-count"
```
pdyraga, line 22: What do you think about annotating the flags to which command they belong? It's getting quite crowded here.

```
var (
	// listDepositsCommand:
	// proposeDepositsSweepCommand:
	walletFlagName = "wallet"

	// listDepositsCommand:
	hideSweptFlagName    = "hide-swept"
	sortByAmountFlagName = "sort-amount"
	headFlagName         = "head"
	tailFlagName         = "tail"

	// proposeDepositsSweepCommand:
	feeFlagName    = "fee"
	dryRunFlagName = "dry-run"

	// estimateDepositsSweepFeeCommand:
	depositsCountFlagName = "deposits-count"
)
```

Non-blocking of course.

pkg/bitcoin/chain.go
```diff
@@ -46,4 +46,8 @@ type Chain interface {
 	// that pays the given public key hash using either a P2PKH or P2WPKH script.
 	// The returned transactions are in an indefinite order.
 	GetMempoolForPublicKeyHash(publicKeyHash [20]byte) ([]*Transaction, error)
+
+	// EstimateSatPerVByteFee returns the estimated sat/vbyte fee for a
+	// transaction to be confirmed within the given number blocks.
```
pdyraga, line 51:
```suggestion
	// transaction to be confirmed within the given number of blocks.
```

pkg/bitcoin/electrum/electrum.go
```diff
@@ -566,6 +566,38 @@ func (c *Connection) getScriptMempool(
 	return convertedItems, nil
 }
 
+// EstimateSatPerVByteFee returns the estimated sat/vbyte fee for a
+// transaction to be confirmed within the given number blocks.
```
pdyraga, line 570:
```suggestion
// transaction to be confirmed within the given number of blocks.
```

pkg/bitcoin/estimator.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+}
+
+// AddPublicKeyHashInputs adds the provided count of P2WPKH (isWitness is true)
+// or P2PKH (isWitness is false) inputs to the estimation. If the estimator
+// already errored out during previous actions, this method does nothing.
+func (tse *TransactionSizeEstimator) AddPublicKeyHashInputs(
+	count int,
+	isWitness bool,
+) *TransactionSizeEstimator {
```
pdyraga, line 39: Builder! 🏗️

pdyraga (commented):
First pass.

pkg/coordinator/sweep.go
```diff
@@ -127,3 +132,129 @@ func ValidateDepositString(depositString string) error {
...
+func estimateDepositsSweepFee(
+	btcChain bitcoin.Chain,
+	depositsCount int,
+	perDepositMaxFee uint64,
+) (int64, error) {
+	transactionSize, err := bitcoin.NewTransactionSizeEstimator().
+		// 1 P2WPKH main UTXO input.
+		AddPublicKeyHashInputs(1, true).
+		// depositsCount P2WSH deposit inputs.
+		AddScriptHashInputs(depositsCount, depositScriptByteSize, true).
```
pdyraga, line 196: We also allow for P2SH input, correct? 

P2SH is longer than P2WSH. What it the % margin we can be off with the estimation? Worth making it very clear in `EstimateDepositsSweepFee` docs.

pkg/bitcoin/electrum/electrum.go
```diff
@@ -566,6 +566,38 @@ func (c *Connection) getScriptMempool(
...
+func (c *Connection) EstimateSatPerVByteFee(blocks uint32) (int64, error) {
+	// According to Electrum protocol docs, the returned fee is BTC/KB.
+	btcPerKbFee, err := requestWithRetry(
+		c,
+		func(
+			ctx context.Context,
+			client *electrum.Client,
+		) (float32, error) {
+			return client.GetFee(ctx, blocks)
```
pdyraga, line 579: Electrum documentation [says](https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain-estimatefee) that `blockchain.estimatefee` is deprecated since version 1.4.2. Do you know what is the reason and what should be used instead (cc @nkuba)?

pkg/coordinator/sweep.go
```diff
@@ -127,3 +132,129 @@ func ValidateDepositString(depositString string) error {
...
+	writer := tabwriter.NewWriter(
+		os.Stdout,
+		2,
+		4,
+		1,
+		' ',
+		tabwriter.AlignRight,
+	)
+
+	_, err := fmt.Fprintf(writer, "deposits count\tfee (satoshis)\t\n")
```
pdyraga, line 229: We could make it clear in the table that this is the fee for the entire sweep transaction. It is pretty unclear to someone not familiar with this code they can just copy&paste the value to propose the deposit sweep.

This is what is currently printed:
```
 deposits count fee (satoshis)
             20           1935
```
IMO it is unclear if it is per-deposit or per deposits count fee.

This is the documentation of the deposit sweep proposal command:
```
--fee int         fee for the entire bitcoin transaction (satoshi)
```


```suggestion
	_, err := fmt.Fprintf(writer, "deposits count\ttotal fee (satoshis)\t\n")
```

pdyraga (commented):
The next month, we'll have to automate it by auto-estimating the fee when submitting the proposal for sweep and redemption. This may require assembling the transaction in the maintainer bot which will allow us for a more precise estimation, given we do not know if we are dealing with P2SH or P2WSH inputs.

The current approach is better for the beginning because we can confirm the estimation against 3rd party tools and get comfortable with the mechanism before we automate it.

cmd/coordinator.go
lukasz-zimnoch: See 49176d14

pkg/bitcoin/electrum/electrum.go
lukasz-zimnoch: See 0d7f237b

pkg/bitcoin/estimator.go
```diff
@@ -0,0 +1,258 @@
...
+	"fmt"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/mempool"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+)
+
+// About half of all signatures generated with a random nonce are 72-byte, about
+// half are one byte smaller, and a small percentage are smaller than that.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://transactionfee.info/charts/bitcoin-script-ecdsa-length/

pkg/bitcoin/estimator_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/keep-network/keep-core/pkg/internal/testutils"
+	"reflect"
+	"testing"
+)
+
+func TestTransactionSizeEstimator_VirtualSize(t *testing.T) {
+	var tests = map[string]struct {
+		estimator           *TransactionSizeEstimator
+		expectedVirtualSize int
+		expectedError       error
```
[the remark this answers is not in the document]
lukasz-zimnoch: The `TransactionSizeEstimator.VirtualSize` method is forced to return an error as the estimator uses some functions from `bitcoin/script.go` that may return an error in some cases. In the current implementation, the estimator uses those functions in a way that will never produce an error, but still, we need to reflect that fact in the API and in the unit tests as this may change in the future. The `expectedError` in those tests is always `nil` so I think it should be here to emphasize this fact.

pkg/bitcoin/electrum/electrum.go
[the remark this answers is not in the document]
lukasz-zimnoch: The `client.GetFee` is part of the `go-electrum` library. IIRC, the docstring of this function points to the protocol docs anyway.

pkg/bitcoin/electrum/electrum.go
```diff
@@ -566,6 +566,38 @@ func (c *Connection) getScriptMempool(
...
+	if btcPerKbFee < 0 {
+		return 0, fmt.Errorf(
+			"daemon does not have enough information to make an estimate",
+		)
+	}
+
+	// To convert from BTC/KB to sat/vbyte, we need to multiply by 1e8/1e3.
+	satPerVByte := (1e8 / 1e3) * float64(btcPerKbFee)
+	// Make sure the minimum returned sat/vbyte fee is always 1.
+	satPerVByte = math.Max(satPerVByte, 1)
```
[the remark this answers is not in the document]
lukasz-zimnoch: This is dynamic and I don't think there is a range that makes sense. I checked the 3y fee history on testnet and it is very broad (source https://mempool.space/testnet/graphs/mining/block-fee-rates#3y):
image

I don't think that checking if the value is in the range `[1, 30000] ` tests this conversion in any way. On the other hand, if we use a narrower range, we may observe random failures of the integration test. In my opinion, the only thing that makes sense is asserting the value is `>=1` and that's it.

pkg/coordinator/sweep.go
lukasz-zimnoch: The dapp produces P2WSH deposits addresses and I don't think we have a single P2SH made manually atm. Even if we have one P2SH sometimes, that should not have a big impact if we sweep ~5 deposits at the same time. One P2SH input is `~150 vB` bigger than a P2WSH.

pkg/coordinator/sweep.go
lukasz-zimnoch: Good point, see 3239f5d7

pkg/bitcoin/electrum/electrum.go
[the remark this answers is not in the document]
pdyraga: What about this:
1. Let's add an integration test proving this mechanism works on all Electrum implementations ([this comment](https://github.com/keep-network/keep-core/pull/3581#discussion_r1201592389)).
2. Let's add a unit test that covers this logic: https://github.com/keep-network/keep-core/blob/43043cbbd06fa7387d118d2aa08361d452502b70/pkg/bitcoin/electrum/electrum.go#L585-L598
  Since the `Connection struct` refers to `client *electrum.Client` we can't easily inject the mock there to test the entire interaction. We can, however, introduce a package-private `convertBtcKBToSatVByte` function (rename if you can think of any better name) that will contain this logic and be tested from `electrum_test.go`. Not a perfect solution but better than nothing, especially since this logic is very important.  `convertBtcKBToSatVByte` would be called from `EstimateSatPerVByteFee` on the result returned by the Electrum server.

pkg/bitcoin/electrum/electrum.go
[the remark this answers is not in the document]
pdyraga: Correct, I was hoping for a PR to our fork: https://github.com/keep-network/go-electrum

Eventually, I think we should [drop the replacement rule](https://github.com/keep-network/keep-core/blob/5bb7d524e4124fcb1cfc42ee7dab75915bba6228/go.mod#L11) and switch to our fork. I don't think the referenced library is actively maintained.

The docstring refers to the protocol docs but this information is quite important and I'd love to have it in the Go documentation as well, especially since it mentions [the fee is per kilobytes](https://github.com/keep-network/go-electrum/blob/567f2c1de4bc694a7b864b04944f86fc2018f08d/electrum/misc.go#L14-L17). I do not consider it a blocker but a really-nice-to-have followup.

pdyraga: This was a good decision but can we just drop a TODO above the `client.GetFee` call to warn about the deprecation?

pkg/coordinator/sweep.go
```diff
@@ -127,3 +132,129 @@ func ValidateDepositString(depositString string) error {
 	}
 	return nil
 }
+
+func EstimateDepositsSweepFee(
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 7eda5686

pkg/coordinator/sweep.go
```diff
@@ -127,3 +132,129 @@ func ValidateDepositString(depositString string) error {
...
+
+	// Compute the maximum possible fee for the entire sweep transaction.
+	maxFee := uint64(depositsCount) * perDepositMaxFee
+	// Make sure the proposed fee does not exceed the maximum possible fee.
+	fee = int64(math.Min(float64(fee), float64(maxFee)))
+
+	return fee, nil
+}
+
+func printDepositsSweepFeeTable(fees map[int]int64) error {
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 3e37119c

pkg/coordinator/sweep.go
pdyraga: The estimated transaction size for 10 inputs is 1022 vBytes. Currently, the estimator suggests `satPerVByte = 1` so the fee for the transaction is 1022 sats. If half of those inputs are P2SH, it means we need to add `2 x 150` sat to the fee. `300` sat is not a lot in USD but it is almost 30% more compared to the original fee.

The estimated transaction size for 5 inputs is 566 vBytes. If 2 inputs are P2SH, it means we need to add `2 x 150` sat to the fee. This is more than 50%  compared to the original fee.

Unless I missed something, the difference is quite significant. 

I don't consider it to be a blocker but I think we should:
- Short term (this PR): Warn about it in `EstimateDepositsSweepFee` docs and in the `estimateDepositsSweepFeeCommandDescription`.
- Mid-term (consider next week): in `getDeposits` function of the coordinator, check the type of a deposit having `event.FundingTxHash`, `event.FundingOutputIndex` to say if it is P2SH or P2WSH.
- Long-term (when automating): estimate the cost based on the actual transaction.

pdyraga: I see the short-term proposition was already handled in 7eda5686d1e2b32d4403f628dc240a654be3f7ca when I was writing this comment. Let me know please what you think about mid-term and long-term goals and I can open separate issues for them.

lukasz-zimnoch: What really matters is the actual `sat/vbyte` fee of the transaction. Let's use an example:

At the moment, according to https://mempool.space/pl, we need to:
- Pay `~52 sat/vbyte` to include the transaction in the next block
- Pay `~47 sat/vbyte` to include the transaction in the second block
- Pay `~41 sat/vbyte` to include the transaction in the third block

If we assume we have a 10 P2WSH deposit transaction of size `1022 vBytes`, we will propose `1022*52 = 53144 sat` as a total fee to include it in the next block. If it turns out that half of the actual transaction are P2SH deposits, the transaction size will be around `1770 bytes`. We used `53144 sat` as the total fee which gives `53144/1770 = 30 sat/vbyte` as the actual fee. That means the transaction will be included in 4-5 blocks. This still works and won't cause any serious problems despite the transaction will be included a bit later. 

If we want to be super cautious, the mid-term proposition makes perfect sense here. We can print the type of the deposit transaction while listing deposits and, if there are some P2SH ones and we want to propose them to the sweep, we just manually propose a higher fee than the one printed by the CLI.

Regarding the long-term proposition, I don't think we can build an actual transaction in another place than the wallet's sweep action code. In order to obtain a super-accurate estimation, we just automatically check the deposit types (just like in the mid-term proposition), check whether there is a wallet main UTXO (now we always assume it is present), and then use the current `TransactionSizeEstimator`. This will give the same result (+/- 1 byte for signature padding) as building an actual transaction which is really hard in an environment other than wallet action.

pkg/bitcoin/electrum/electrum.go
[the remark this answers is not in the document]
lukasz-zimnoch: See c61ec3b7

pkg/bitcoin/electrum/electrum.go
```diff
@@ -566,6 +566,38 @@ func (c *Connection) getScriptMempool(
 	return convertedItems, nil
 }
 
+// EstimateSatPerVByteFee returns the estimated sat/vbyte fee for a
+// transaction to be confirmed within the given number blocks.
+func (c *Connection) EstimateSatPerVByteFee(blocks uint32) (int64, error) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: See c61ec3b7

pkg/coordinator/sweep.go
pdyraga: > At the moment, according to https://mempool.space/pl, we need to:
> 
> * Pay `~52 sat/vbyte` to include the transaction in the next block
> * Pay `~47 sat/vbyte` to include the transaction in the second block
> * Pay `~41 sat/vbyte` to include the transaction in the third block

This is potentially alarming because when I run `./keep-client coordinator estimate-deposits-sweep-fee` against `electrum.blockstream.info:60001` I am always getting 1 sat/vbyte:
```
 deposits count total fee (satoshis) sat/vbyte
              1                  201         1
              2                  292         1
              3                  384         1
              4                  475         1
              5                  566         1
```

> If we assume we have a 10 P2WSH deposit transaction of size `1022 vBytes`, we will propose `1022*52 = 53144 sat` as a total fee to include it in the next block. If it turns out that half of the actual transaction are P2SH deposits, the transaction size will be around `1770 bytes`. We used `53144 sat` as the total fee which gives `53144/1770 = 30 sat/vbyte` as the actual fee. That means the transaction will be included in 4-5 blocks. This still works and won't cause any serious problems despite the transaction will be included a bit later.

This is ~40% decrease compared to the value we should propose:
```
53144 / 1022 = 52 sat/vbyte
53144 / 1770 = 30 sat/vbyte
```

Agreed, it still works in the current market conditions but the % change is quite significant and we can not guarantee this will work correctly a few months from today. 

I will open an issue to track mid-term and long-term work but in the meantime... what's off with the Blockstream's estimation?

pdyraga: I am not running integration tests, I am running `./keep-client coordinator estimate-deposits-sweep-fee` on a binary that I have previously built with `make mainnet` pointing to `electrum.blockstream.info:60001` in the config.

lukasz-zimnoch: `electrum.blockstream.info:60001` is testnet. Use `electrum.blockstream.info:50001` for mainnet. Compare https://1209k.com/bitcoin-eye/ele.php?chain=btc and https://1209k.com/bitcoin-eye/ele.php?chain=tbtc

pkg/bitcoin/electrum/electrum_integration_test.go
```diff
@@ -345,6 +345,24 @@ func TestGetTransactionsForPublicKeyHash_Integration(t *testing.T) {
 	}
 }
 
+func TestEstimateSatPerVByteFee_Integration(t *testing.T) {
```
pdyraga, line 348: This test has a similar problem as [this one](https://github.com/keep-network/keep-core/pull/3578#discussion_r1200310615). I do not consider it blocking this PR but we should investigate what is happening and it it is not a quick fix, at least leave a TODO. 

I altered the host name in `fulcrum ssl` test to something that can't be resolved and while all other tests fail nicely with `electrum_integration_test.go:369: failed to initialize electrum client: [retry timeout [1m0s] exceeded; most recent error: [dial tcp: lookup blackie.c3-softfff.com: no such host]]` , this one gets stuck:

```
=== RUN   TestEstimateSatPerVByteFee_Integration/fulcrum_ssl
panic: test timed out after 10m0s

goroutine 429 [running]:
testing.(*M).startAlarm.func1()
        /opt/homebrew/Cellar/go@1.18/1.18.5/libexec/src/testing/testing.go:2029 +0x8c
created by time.goFunc
        /opt/homebrew/Cellar/go@1.18/1.18.5/libexec/src/time/sleep.go:176 +0x3c

goroutine 1 [chan receive, 2 minutes]:
testing.(*T).Run(0x14000555520, {0x102e1e49b?, 0x1c114841bbb5b?}, 0x103160588)
```

It is not a problem as long as `EstimateSatPerVByteFee` is not used by an automated tool but we should be aware of the limitation and find what is the reason.

pkg/coordinator/sweep.go
pdyraga: 🤦🏻

pkg/bitcoin/electrum/electrum_integration_test.go
lukasz-zimnoch: See 9044c33e

Issue #3580, opened 2023-05-22
Make the scheme (protocol) part of the Electrum server URL


pdyraga:
See [this](https://github.com/keep-network/keep-core/pull/3576#discussion_r1198881633) discussion.

Our config for `[bitcoin.Electrum]` is not consistent with the config for `[ethereum]`.

Compare:

https://github.com/keep-network/keep-core/blob/51127b93140c795409116b42948f5fe0dad6f181/configs/config.toml.SAMPLE#L3-L4

https://github.com/keep-network/keep-core/blob/51127b93140c795409116b42948f5fe0dad6f181/configs/config.toml.SAMPLE#L49-L56

Scheme (`wss://`,`ssl://`, `tcp://`) should [be a part of the URL](https://www.rfc-editor.org/rfc/rfc1738#section-5) just like we have it for `[ethereum]`. If we want to resolve the protocol later, this should be an auxiliary function on the `Config` type parsing the protocol from the URL and not a separate field:

https://github.com/keep-network/keep-core/blob/51127b93140c795409116b42948f5fe0dad6f181/pkg/bitcoin/electrum/config.go#L27-L28

If we do this change, then we can return just an URL from `readElectrumUrls` [without bothering](https://github.com/keep-network/keep-core/blob/cf266a693b5ad922699039e0fbc15451027c43d3/config/electrum.go#L19-L22) about tuples or introducing other types.

Please make sure to update:
- [x] docs.threshold.network (https://app.gitbook.com/o/R2meumXNNad4y1B10iL7/s/WosjlL4zUGUMlcMfuSAp/~/changes/216), 
- [x] infrastructure setup (https://github.com/keep-network/keep-core/pull/3604),
- [x] config samples (https://github.com/keep-network/keep-core/pull/3582).

PR #3582, opened 2023-05-22
Support Electrum WebSocket connections 

nkuba:
In this PR we switch to the latest version of the go-electrum client that supports WebSocket connections to Electrum servers (a1b9133b54c35e8ca453b9f448a0a26da3499315, 4d4f3bdb9246daded1eb6e60de209741b603b64d).
Refs: https://github.com/keep-network/go-electrum/pull/2).

We also modify the way Electrum server details are provided in the config to include the Protocol inside the URL property. It is expected that `--electrum.url` will be provided in the following format: `scheme://host:port`. A separate `--electrum.protocol` config is no longer necessary. (71f3618930bb2ec3ca4f8415858243fc8134db56)
Refs: https://github.com/keep-network/keep-core/issues/3580

When establishing SSL connection the client will no longer skip the certificate verification (`InsecureSkipVerify`). Some of the servers we were previously using in tests have self-signed or improperly configured certificates, but the Blockstream's server that we defined in `./config/_electrum_urls` works fine. (f673aa97e163db22b9f9780126eee73c7fcc97c3)

We updated integration tests to reflect the changes described above. To run them execute: `go test -v -tags=integration ./pkg/bitcoin/electrum`.


go.mod
```diff
@@ -8,7 +8,7 @@ replace (
 	// here:
 	github.com/btcsuite/btcd => github.com/btcsuite/btcd v0.22.3
 	github.com/btcsuite/btcd/v2 => github.com/btcsuite/btcd v0.23.4
-	github.com/checksum0/go-electrum => github.com/keep-network/go-electrum v0.0.0-20230516144526-567f2c1de4bc
+	github.com/checksum0/go-electrum => github.com/keep-network/go-electrum v0.0.0-20230519144954-5d10dfb01e18
```
lukasz-zimnoch, line 11: Let's remember about updating to the latest commit from https://github.com/keep-network/go-electrum/pull/2

config/electrum.go
```diff
@@ -3,11 +3,10 @@ package config
...
-	"github.com/keep-network/keep-core/pkg/bitcoin"
-	"github.com/keep-network/keep-core/pkg/bitcoin/electrum"
 	"math/rand"
-	neturl "net/url"
 	"strings"
+
+	"github.com/keep-network/keep-core/pkg/bitcoin"
 )
 
 //go:embed _electrum_urls/*
```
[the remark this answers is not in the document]
nkuba: We will handle it in a separate PR as we still confirming the list.

go.mod
nkuba: cd684e655146a095b26d19870985b3387d257d9b

lukasz-zimnoch (approved):

Issue #3579, opened 2023-05-22
Rename `--goerli` flag to `--testnet` flag, remove `--mainnet` flag


pdyraga:
See [this](https://github.com/keep-network/keep-core/pull/3576#discussion_r1198874503) discussion.

As someone not that familiar with the client code, I would be confused about `--goerli` flag.  This flag clearly connects the client to Goerli Ethereum testnet, but how would I select the Bitcoin testnet? 

Eventually, we should have one client binary where I select `--testnet` for Sepolia (will replace Goerli) + Bitcoin testnet, and `--developer` for local Ethereum and regtest. Mainnet Ethereum + mainnet Bitcoin would be the default one. Renaming `--goerli` to `--testnet` is the first step. The next step is figuring out the build process (out of scope currently) so that we build one client binary instead of testnet-specific and mainnet-specific versions (out of scope of this PR).

To be done when working on this issue:
- [x] Rename `--goerli` flag to `--testnet` (https://github.com/keep-network/keep-core/pull/3583) https://github.com/keep-network/keep-core/blob/cf266a693b5ad922699039e0fbc15451027c43d3/cmd/flags.go#L86
- [x] Remove `--mainnet` flag https://github.com/keep-network/keep-core/blob/cf266a693b5ad922699039e0fbc15451027c43d3/cmd/flags.go#L78-L79
- [x] Update infrastructure config files (https://github.com/keep-network/keep-core/pull/3604)
- [x] Update config samples and documentation (https://app.gitbook.com/o/R2meumXNNad4y1B10iL7/s/WosjlL4zUGUMlcMfuSAp/~/changes/216)

nkuba:
>     * [ ]  Remove `--mainnet` flag https://github.com/keep-network/keep-core/blob/cf266a693b5ad922699039e0fbc15451027c43d3/cmd/flags.go#L78-L79

We decided to mark the flag deprecated for backward compatibility and remove it in the future. Please see https://github.com/keep-network/keep-core/pull/3583

PR #3583, opened 2023-05-23
Replace flag --goerli with --testnet

nkuba:
We define a general `--testnet` flag for the client that replaces the `--goerli` flag. 
The flag determines the threshold network against which the client runs and is used to resolve Ethereum and Bitcoin networks.

After we added Bitcoin support the `--goerli` flag was confusing to use, as it relates to Ethereum only.

To run the client against the mainnet network no flag is required. To maintain backward compatibility the `--mainnet` flag remained available but it was marked as deprecated. We will remove the `--mainnet` flag in the future.

Refs: https://github.com/keep-network/keep-core/issues/3579


lukasz-zimnoch (approved):
LGTM!

Let's remember about updating relevant parts of `docs.threshold.network` and infrastructure setup.

nkuba:
> Let's remember about updating relevant parts of `docs.threshold.network` and infrastructure setup.

Yes, we track the remaining parts in https://github.com/keep-network/keep-core/issues/3579.
Docs were already updated in https://app.gitbook.com/o/R2meumXNNad4y1B10iL7/s/WosjlL4zUGUMlcMfuSAp/~/changes/216

lukasz-zimnoch (approved):

PR #3593, opened 2023-05-30
Electrum servers list updates

nkuba:
### Blockstream's Servers Keep Alive Interval (ba416cedb989b0eebc8768ec7fc56de662647803)
Blockstream's servers timeout session after 60 seconds of inactivity which
is much shorter than the expected 600 seconds (see: [Electrum Protocol Docs](https://electrumx-spesmilo.readthedocs.io/en/latest/environment.html#envvar-SESSION_TIMEOUT)).
To workaround connection drops and logs pollution with warnings, we reduce the
`KeepAliveInterval` for Blockstream's servers to less than 60 seconds.
 
### Integration Tests for Embedded Electrum Servers (a288918fc657a779ec09b0b8b14a4bd1a80c4468)
We add an integration test to check the client can establish a connection to the embedded servers.

To run it execute:
```
go test -v -tags=integration ./config -run TestEmbeddedElectrumServerConnect_Integration
```

### Add Electrum Servers (e677fc312c87febe1080d69db3c11e5883dd6847)

We add more servers to the Electrum Servers list.

### DIFF
--- a/config/_electrum_urls/mainnet
+++ b/config/_electrum_urls/mainnet
@@ -1,1 +1,3 @@
+wss://electrumx-server.tbtc.network:8443
+wss://electrum.boar.network:2083
 ssl://electrum.blockstream.info:50002
--- a/config/_electrum_urls/mainnet
+++ b/config/_electrum_urls/mainnet
@@ -1,3 +1,4 @@
 wss://electrumx-server.tbtc.network:8443
 wss://electrum.boar.network:2083
+wss://bitcoin.threshold.p2p.org:50004
 ssl://electrum.blockstream.info:50002
--- a/config/electrum.go
+++ b/config/electrum.go
@@ -38,7 +38,7 @@ func readElectrumUrls(network bitcoin.Network) (
 // resolveElectrum checks if Electrum is already configured. If the Electrum URL
 // is empty it reads the Electrum configs from the embedded list for the given
 // network and picks up one randomly.
-func (c *Config) resolveElectrum() error {
+func (c *Config) resolveElectrum(rng *rand.Rand) error {
 	network := c.Bitcoin.Network
 
 	// Return if Electrum is already set.
@@ -70,7 +70,7 @@ func (c *Config) resolveElectrum() error {
 
 	// #nosec G404 (insecure random number source (rand))
 	// Picking up an Electrum server does not require secure randomness.
-	selectedURL := urls[rand.Intn(len(urls))]
+	selectedURL := urls[rng.Intn(len(urls))]
 
 	// Set only the URL in the original config. Other fields may be already set,
 	// and we don't want to override them.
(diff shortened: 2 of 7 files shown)


config/_electrum_urls/mainnet
```diff
@@ -1 +1,3 @@
-ssl://electrum.blockstream.info:50002
\ No newline at end of file
+wss://electrumx-server.tbtc.network:8443
+wss://electrum.boar.network:2083
+ssl://electrum.blockstream.info:50002
```
pdyraga, line 3: P2P set up their Electrum server. It is available at:
wss://bitcoin.threshold.p2p.org:50004
ssl://bitcoin.threshold.p2p.org:50002

nkuba: It wasn't available the last time I checked. I'll test it again.

config/electrum.go
```diff
@@ -63,11 +70,20 @@ func (c *Config) resolveElectrum() error {
...
 	c.Bitcoin.Electrum.URL = selectedURL
 
+	// Blockstream's servers timeout session after 60 seconds of inactivity which
+	// is much shorter than expected 600 seconds. To workaround connection drops
+	// and logs pollution with warning we reduce the KeepAliveInterval for the
+	// Blockstream's servers to less than 60 seconds.
+	if c.Bitcoin.Electrum.KeepAliveInterval == 0 &&
+		strings.Contains(selectedURL, "electrum.blockstream.info") {
+		c.Bitcoin.Electrum.KeepAliveInterval = blockstreamKeepAliveInterval
+	}
```
pdyraga, line 86: This is good for now but ultimately, I'd rather replace `config/_electrum_urls/mainnet` file with `config/electrum/mainnet` directory and separate config files for each electrum server.

config/_electrum_urls/mainnet
```diff
@@ -1 +1,4 @@
-ssl://electrum.blockstream.info:50002
\ No newline at end of file
+wss://electrumx-server.tbtc.network:8443
+wss://electrum.boar.network:2083
+wss://bitcoin.threshold.p2p.org:50004
+ssl://electrum.blockstream.info:50002
```
pdyraga, line 4: Should we include all those servers in tests executed in `/pkg/bitcoin/electrum/electrum_integration_test.go`? 

Checking if the connection works is fine but checking if they can execute e.g. `GetLatestBlockHeight` is even better.

nkuba: Yes, we should have a way to validate the servers, it's tracked in https://github.com/keep-network/keep-core/issues/3594.

pdyraga:
Tested locally and works as advertised. All mainnet servers are randomly selected and I no longer see warnings for Blockstream server. To test locally, I had to add a small change: #3596 so that I can see to which Electrum server I am actually connected.

pdyraga (approved):

PR #3596, opened 2023-05-31
Print Electrum URL in the logs

pdyraga:
Unless specified in the config, mainnet client binary selects the Electrum server from one of the preconfigured nodes. It is nice to know which server was selected, even for debugging purposes. Additionally, we should log both Bitcoin and Ethereum networks used by the client. Electrum server URL is printed to the logs only if it has been auto-selected. This approach is consistent with not logging Ethereum server URL at all given we expect it is passed in the configuration.

I snaked one more change in this PR: fixed the number of accounts requirement in `scripts/install.sh`. Given the recent changes in tbtc-v2 project, at least 11 accounts are required in the local Ethereum client keystore to deploy and initialize contracts.

The `v1Redeemer` [set in tbtc-v2 `hardhat.config.ts`](https://github.com/keep-network/tbtc-v2/blob/7f48d6362f8bfc9a2931c4087e5eba28efee5370/solidity/hardhat.config.ts#L227) uses index 10 and indexes start from 0.

Example:
```
2023-06-05T08:54:24.185+0200    INFO    keep-config     config/config.go:95     using [mainnet] Ethereum network and [mainnet] Bitcoin network
2023-06-05T08:54:24.187+0200    INFO    keep-config     config/electrum.go:68   Auto-selecting Electrum server: [ssl://electrum.blockstream.info:50002]
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -92,6 +92,12 @@ func (c *Config) resolveNetworks(flagSet *pflag.FlagSet) (network.Type, error) {
 	c.Ethereum.Network = clientNetwork.Ethereum()
 	c.Bitcoin.Network = clientNetwork.Bitcoin()
 
+	logger.Infof(
+		"using [%v] Ethereum network and [%v] Bitcoin network",
+		c.Ethereum.Network,
+		c.Bitcoin.Network,
+	)
+
 	return clientNetwork, err
 }
 
--- a/config/electrum.go
+++ b/config/electrum.go
@@ -65,6 +65,8 @@ func (c *Config) resolveElectrum() error {
 	// Picking up an Electrum server does not require secure randomness.
 	selectedURL := urls[rand.Intn(len(urls))]
 
+	logger.Infof("Auto-selecting Electrum server: [%v]", selectedURL)
+
 	// Set only the URL in the original config. Other fields may be already set,
 	// and we don't want to override them.
 	c.Bitcoin.Electrum.URL = selectedURL
--- a/config/electrum.go
+++ b/config/electrum.go
@@ -72,7 +72,7 @@ func (c *Config) resolveElectrum(rng *rand.Rand) error {
 	// Picking up an Electrum server does not require secure randomness.
 	selectedURL := urls[rng.Intn(len(urls))]
 
-	logger.Infof("Auto-selecting Electrum server: [%v]", selectedURL)
+	logger.Infof("auto-selecting Electrum server: [%v]", selectedURL)
 
 	// Set only the URL in the original config. Other fields may be already set,
 	// and we don't want to override them.
--- a/pkg/bitcoin/electrum/electrum.go
+++ b/pkg/bitcoin/electrum/electrum.go
@@ -679,7 +679,8 @@ func (c *Connection) verifyServer() error {
 	}
 
 	logger.Infof(
-		"connected to electrum server [version: [%s], protocol: [%s]]",
+		"connected to [%s] electrum server [version: [%s], protocol: [%s]]",
+		c.config.URL,
 		server.version,
 		server.protocol,
 	)
--- a/pkg/bitcoin/electrum/electrum.go
+++ b/pkg/bitcoin/electrum/electrum.go
@@ -679,8 +679,7 @@ func (c *Connection) verifyServer() error {
 	}
 
 	logger.Infof(
-		"connected to [%s] electrum server [version: [%s], protocol: [%s]]",
-		c.config.URL,
+		"connected to electrum server [version: [%s], protocol: [%s]]",
 		server.version,
 		server.protocol,
 	)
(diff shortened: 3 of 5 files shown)


pkg/bitcoin/electrum/electrum.go
```diff
@@ -679,7 +679,8 @@ func (c *Connection) verifyServer() error {
 	}
 
 	logger.Infof(
-		"connected to electrum server [version: [%s], protocol: [%s]]",
+		"connected to [%s] electrum server [version: [%s], protocol: [%s]]",
+		c.config.URL,
```
nkuba, line 683: I'm not sure we should print it in all the cases. What if the URL is configured by the user and contains authorization data? Maybe we could print the server in the function where we draw it when it's missing in the config?

pdyraga: Good point. Unfortunately, printing the server in the function where we draw it gives a weird order of entries when starting the client:
```
2023-06-02T16:24:16.833+0200    INFO    keep-config     config/electrum.go:68   Auto-selecting Electrum server: [ssl://electrum.blockstream.info:50002]
2023-06-02T16:24:16.834+0200    INFO    keep-cmd        cmd/start.go:62 Starting the client against [mainnet] ethereum network...
```

We can do it given it is better to print it this way than not print at all but I'd like to explore the possibility of using a regex to remove possible credentials first.

nkuba: You may want to look at the url library to parse the string and cut off the credentials.

pdyraga: I took a bit different approach, see https://github.com/keep-network/keep-core/pull/3596/commits/604f2a0355b272508b96b434189325a08de29ecb:
1. We only print Electrum server URL when it is auto-selected. This is consistent with the approach we use for Ethereum: we are not printing the URL to the logs because we expect the Ethereum server URL is set in the config. Printing the auto-selected URL is logged in the config logger because the auto-selection happens in the config.
2. We now print both Ethereum and Bitcoin networks used. This is also happening on the config level because those networks are resolved on the config level.

pdyraga: Notes for future-us:

1. Redacting with `url` package is nice but it only redacts the password. I'd prefer to redact the username as well.
2. I do not think Fulcrum/ElectrumX allows passing username+password in the URL. This would be only secure on TLS-encrypted connections and I did not find anything about it in the config. If we ever see someone like Infura/Alchemy for Bitcoin, they will probably require to pass the project ID as a part of the path and allow only TLS-encrypted connections, just like Infura/Alchemy. In any case, we shouldn't output this URL to logs and that's why we are not outputting Ethereum server URL into logs.

tomaszslabon (dismissed):

config/config.go
```diff
@@ -94,6 +94,12 @@ func (c *Config) resolveNetworks(flagSet *pflag.FlagSet) (network.Type, error) {
 	c.Ethereum.Network = clientNetwork.Ethereum()
 	c.Bitcoin.Network = clientNetwork.Bitcoin()
 
+	logger.Infof(
+		"using [%v] Ethereum network and [%v] Bitcoin network",
+		c.Ethereum.Network,
+		c.Bitcoin.Network,
+	)
```
nkuba, line 101: This check is based on the flags declared by the user in config. Could we add two checks?
1. If Ethereum is testnet, then Bitcoin also has to be testnet, If Ethereum is mainneg then Bitcoin also has to be mainnet.
2. That the actual Ethereum and Electrum servers the node is connected to supports the desired networks?
We have a similar solution in the Electrum TS library: https://github.com/keep-network/tbtc-v2/blob/7f48d6362f8bfc9a2931c4087e5eba28efee5370/typescript/src/electrum.ts#L208-L224

config/electrum.go
```diff
@@ -72,6 +72,8 @@ func (c *Config) resolveElectrum(rng *rand.Rand) error {
 	// Picking up an Electrum server does not require secure randomness.
 	selectedURL := urls[rng.Intn(len(urls))]
 
+	logger.Infof("Auto-selecting Electrum server: [%v]", selectedURL)
```
nkuba, line 75:
```suggestion
	logger.Infof("auto-selecting Electrum server: [%v]", selectedURL)
```

config/config.go
nkuba: This doesn't have to be handled in this PR, as the scope of this PR was significantly reduced to just printing general information.

config/electrum.go
pdyraga: https://github.com/keep-network/keep-core/pull/3596/commits/b06c66d69b75996d12ab1461187dec851b1ebc3b

config/config.go
pdyraga: Can you please open an issue to track this work?

nkuba: https://github.com/keep-network/keep-core/issues/3602

nkuba (approved):

PR #3607, opened 2023-06-06
Redemptions prototype: Assemble BTC redemption transaction

lukasz-zimnoch:
Refs #3609 

Here we introduce the `assembleRedemptionTransaction` function that allows constructing an unsigned Bitcoin redemption transaction based on provided input arguments.

### DIFF
--- a/pkg/tbtc/redemption.go
+++ b/pkg/tbtc/redemption.go
@@ -77,8 +77,7 @@ func withRedemptionTotalFee(totalFee int64) redemptionFeeDistributionFn {
 // assembleRedemptionTransaction constructs an unsigned redemption Bitcoin
 // transaction.
 //
-// Regarding input arguments, the walletMainUtxo parameter is mandatory and
-// must be set accordingly. The requests slice must contain at least one element.
+// Regarding input arguments, the requests slice must contain at least one element.
 // The fee shares applied to specific requests according to the provided
 // feeDistribution function are not validated in any way so must be chosen with
 // respect to the system limitations. The shape argument is optional - if not


pkg/tbtc/redemption.go
```diff
@@ -0,0 +1,201 @@
...
+	// fixed and leverages some SPV proof cost optimizations made in the Bridge
+	// implementation.
+	RedemptionChangeFirst RedemptionTransactionShape = iota
+	// RedemptionChangeLast is a shape where the change output is the last one
+	// in the transaction output vector.
+	RedemptionChangeLast
+)
+
+// RedemptionRequest represents a tBTC redemption request.
+type RedemptionRequest struct {
```
pdyraga, line 28: Some of the fields, for example `TxMaxFee` and `RequestedAt`, are currently unused but I take the plan was to reflect the [on-chain structure](https://github.com/keep-network/tbtc-v2/blob/66a04c618e140230d25ea0d784ed75e6ba3030a2/solidity/contracts/bridge/Redemption.sol#L137) 1:1 because we'll use those fields in the redemption action anyway. Is this correct assumption?

pkg/tbtc/redemption.go
```diff
@@ -0,0 +1,201 @@
...
+
+		return feeShares
+	}
+}
+
+// assembleRedemptionTransaction constructs an unsigned redemption Bitcoin
+// transaction.
+//
+// Regarding input arguments, the walletMainUtxo parameter is mandatory and
+// must be set accordingly. The requests slice must contain at least one element.
```
pdyraga, line 81: All parameters but `shape` are mandatory, right?

pkg/internal/tbtctest/tbtctest.go
```diff
@@ -20,6 +20,36 @@
 //   one input (a P2WSH deposit) was swept into a P2WPKH main UTXO.
 //   For reference see:
 //   https://live.blockcypher.com/btc-testnet/tx/9efc9d555233e12e06378a35a7b988d54f7043b5c3156adc79c7af0a0fd6f1a0
+//
```
pdyraga, line 23: Could be worth noting all redemption test scenarios have a single P2WPKH input.

pkg/tbtc/redemption_test.go
```diff
@@ -0,0 +1,137 @@
...
+		},
+	}
+
+	for testName, test := range tests {
+		t.Run(testName, func(t *testing.T) {
+			requests := make([]*RedemptionRequest, test.requestsCount)
+
+			feeShares := withRedemptionTotalFee(test.totalFee)(requests)
+
+			if !reflect.DeepEqual(test.expectedFeeShares, feeShares) {
```
pdyraga, line 126: Can we use `deep.Equal` instead? Fewer linting complaints.

pdyraga (commented):
Some minor comments. Badass work.

pkg/tbtc/redemption.go
lukasz-zimnoch: Yeah, correct.

pkg/tbtc/redemption.go
lukasz-zimnoch: Yeah. Added  that info to the fucntion's docstring: 55ffd641

pkg/internal/tbtctest/tbtctest.go
lukasz-zimnoch: See 378ff634

pkg/tbtc/redemption_test.go
lukasz-zimnoch: See 2d4d0415

pkg/tbtc/redemption.go
pdyraga: What do you think about removing this part:

> Regarding input arguments, the walletMainUtxo parameter is mandatory and must be set accordingly.

This is what caused by confusion. We say that `walletMainUtxo` is mandatory, then we describe requests slice, fee shares, and fee distribution without mentioning they are mandatory as well.

With the change from https://github.com/keep-network/keep-core/commit/55ffd641ba3aabe99818150640fea379f342041d, I think we can remove the mentioned sentence and this will be all less confusing.

lukasz-zimnoch: 1bd586a7

pdyraga (approved):

PR #3618, opened 2023-06-07
Update dependencies to `solidity-contracts` and `random-beacon`

michalinacienciala:
We have published new `@threshold-network/solidity-contract` and `@keep-network/random-beacon` packages. Those packages no longer include the `prepare-dependencies.sh` script from
`@threshold-network/solidity-contracts` which was causing random failures during `yarn install`. As in some CI jobs we install the dependencies based on the lockfile (with the `--frozen-lockfile` flag), we need to update the dependencies in the lockfile so that the new, improved packages would be used.

Refs:
https://github.com/threshold-network/solidity-contracts/issues/142
https://github.com/threshold-network/token-dashboard/pull/533
https://github.com/keep-network/coverage-pools/pull/233
https://github.com/keep-network/tbtc-v2/pull/631


michalinacienciala:
I approve @dimpar's changes (one small correction needed to be done afterwards, but we should be good now).

Issue #3612, opened 2023-06-07
Consider removing public Blockstream server from the list of auto-selected servers


pdyraga:
> "This server is a chaos monkey in our system."

@lukasz-zimnoch 🧙🏻‍♂️ 

During the execution of the first sweep, some operators reported a problem with the Electrum connection leading to an error and the client exiting the signing loop:

```
2023-06-07T09:11:41.239Z	WARN	keep-electrum	electrum/electrum.go:795	connection to electrum server is down; reconnecting...
2023-06-07T09:11:41.390Z	INFO	keep-electrum	electrum/electrum.go:800	reconnected to electrum server
(...)
2023-06-07T09:16:42.561Z	WARN	keep-electrum	electrum/electrum.go:795	connection to electrum server is down; reconnecting...
2023-06-07T09:16:42.712Z	INFO	keep-electrum	electrum/electrum.go:800	reconnected to electrum server
(...)
2023-06-07T09:21:43.889Z	WARN	keep-electrum	electrum/electrum.go:795	connection to electrum server is down; reconnecting...
2023-06-07T09:21:44.038Z	INFO	keep-electrum	electrum/electrum.go:800	reconnected to electrum server
(...)
2023-06-07T09:31:18.132Z	WARN	keep-electrum	electrum/electrum.go:795	connection to electrum server is down; reconnecting...
2023-06-07T09:31:18.269Z	INFO	keep-electrum	electrum/electrum.go:800	reconnected to electrum server
(...)
2023-06-07T09:32:44.983Z	ERROR	keep-tbtc	tbtc/wallet.go:116	action execution terminated with error: [error while ensuring wallet state is synced between BTC and host chain: [cannot get mempool: [cannot get P2PKH mempool items for public key hash [0xb0ba76edfe18e81365bddd1d46511a57a4ff8dce]: [failed to get mempool for script [0x76a914b0ba76edfe18e81365bddd1d46511a57a4ff8dce88ac]: [retry timeout [2m0s] exceeded; most recent error: [request failed: [use of closed network connection]]]]]]]	{"wallet": "0x04b0a483e97dfbb15e88ecbc2ef8f7e37776dd713eed54089f87704a4fbae0442aa3ff5a0ec7f4d52bf4680b5b9b70be9accb64970df11b96abb61d48be7d8db8b", "action": "DepositSweep"}
```

[...]

PR #3620, opened 2023-06-08
Remove public Blockstream server from the list of auto-selected servers

nkuba:
The blockstream's electrums servers seems very unstable and causes a lot of noise in the logs, even after our optimizations we did in ba416cedb989b0eebc8768ec7fc56de662647803.

Closes: https://github.com/keep-network/keep-core/issues/3612


pdyraga (approved):
Bye bye.

PR #3630, opened 2023-06-13
MaintainerProxy bindings

tomaszslabon:
This PR adds binding generation for the `MaintainerProxy` contract.
The bindings were generated using the `make mainnet` command.


pkg/chain/ethereum/ecdsa/gen/abi/WalletRegistry.go
```diff
@@ -28,6 +28,20 @@ var (
 	_ = event.NewSubscription
 )
 
+// EcdsaDkgData is an auto generated low-level Go binding around an user-defined struct.
```
pdyraga, line 31: The regenerated contract bindings include dApp-only functions. The contracts should be generated based on the mainnet bindings so you need to run `make mainnet`. The default environment is `development`:

https://github.com/keep-network/keep-core/blob/86e7644fbe7de9fc801437f8a70844546b20c21a/Makefile#L1-L5

With the contracts regenerated based on the mainnet artifacts we should see no changes in other bindings that the ones for `MaintainerProxy` contract.

pdyraga (changes requested):

pkg/chain/ethereum/ecdsa/gen/abi/WalletRegistry.go
tomaszslabon: Indeed, after running `make mainnet` there are only changes related to `MaintainerProxy`.
Done in 36c0c90b775a81a366f20c8e40a396ab4c10af92.

pkg/chain/ethereum/tbtc/gen/Makefile
```diff
@@ -20,9 +21,28 @@ required_contracts := Bridge LightRelay LightRelayMaintainerProxy WalletCoordina
...
+	$(if $(filter $(type),MaintainerProxy),$(call fix_maintainer_proxy_contract_collision))
+endef
+define fix_maintainer_proxy_contract_collision
+	@perl -pi -e s,BitcoinTxUTXO,BitcoinTxUTXO2,g ./contract/MaintainerProxy.go
+	@perl -pi -e s,BitcoinTxProof,BitcoinTxProof2,g ./contract/MaintainerProxy.go
+	@perl -pi -e s,BitcoinTxInfo,BitcoinTxInfo3,g ./contract/MaintainerProxy.go
+	@perl -pi -e s,BitcoinTxUTXO,BitcoinTxUTXO2,g ./cmd/MaintainerProxy.go
+	@perl -pi -e s,BitcoinTxProof,BitcoinTxProof2,g ./cmd/MaintainerProxy.go
+	@perl -pi -e s,BitcoinTxInfo,BitcoinTxInfo3,g ./cmd/MaintainerProxy.go
+endef
```
pdyraga, line 46: I don't understand why we need this piece. Can you elaborate, please?

Contract bindings are generated based on the ABI. If we fixed the ABI in `after_abi_hook`, why the additional work?

lukasz-zimnoch: I was about to ask the same question 😀

tomaszslabon: If I comment this function and run `make mainnet`, I'm getting compilation errors like this:
```
pkg/chain/ethereum/tbtc/gen/contract/MaintainerProxy.go:730:3: cannot use arg_mainUtxo (variable of type "github.com/keep-network/keep-core/pkg/chain/ethereum/tbtc/gen/abi".BitcoinTxUTXO) as type "github.com/keep-network/keep-core/pkg/chain/ethereum/tbtc/gen/abi".BitcoinTxUTXO2 in argument to mp.contract.NotifyMovingFundsBelowDust
```

It seems the generated contract bindings still contain the wrong type (`BitcoinTxUTXO` instead of `BitcoinTxUTXO2`):
```
func (mp *MaintainerProxy) NotifyMovingFundsBelowDust(
	arg_walletPubKeyHash [20]byte,
	arg_mainUtxo abi.BitcoinTxUTXO,

	transactionOptions ...chainutil.TransactionOptions,
) (*types.Transaction, error) {
....
}
```

pdyraga: Let me re-check my understanding of the issue.

`after_abi_hook` renames structures that were incorrectly re-declared in the same Go package given this bug: https://github.com/ethereum/go-ethereum/issues/24627.

This is the version on `main`: 

https://github.com/keep-network/keep-core/blob/78da87c16749ae0b13478b3041c15f1ea1100e92/pkg/chain/ethereum/tbtc/gen/Makefile#L24-L26

My understanding of the bug, is that both the generated `tbtc/gen/abi/Bridge.go` and `tbtc/gen/abi/WalletCoordinator.go` declares `BitcoinTxInfo` struct. To make the Go compiler happy, we rename `BitcoinTxInfo` to `BitcoinTxInfo2` in `tbtc/gen/abi/WalletCoordinator.go`.

Those two Go files are generated with `@go run github.com/ethereum/go-ethereum/cmd/abigen`.

Now, we add another contract that also uses `BitcoinTx.Info` in Solidity so it will have `BitcoinTxInfo` re-declared in the Go code given the issue from https://github.com/ethereum/go-ethereum/issues/24627. We do the same trick as in `main`. We extend the `after_abi_hook` and do some renames:

https://github.com/keep-network/keep-core/blob/36c0c90b775a81a366f20c8e40a396ab4c10af92/pkg/chain/ethereum/tbtc/gen/Makefile#L26-L33

This makes sense to me up to this point.

But this is not enough! If we try to build the project, we will encounter errors like:
```
pkg/chain/ethereum/tbtc/gen/contract/MaintainerProxy.go:730:3: cannot use arg_mainUtxo (variable of type "[github.com/keep-network/keep-core/pkg/chain/ethereum/tbtc/gen/abi](http://github.com/keep-network/keep-core/pkg/chain/ethereum/tbtc/gen/abi)".BitcoinTxUTXO) as type "[github.com/keep-network/keep-core/pkg/chain/ethereum/tbtc/gen/abi](http://github.com/keep-network/keep-core/pkg/chain/ethereum/tbtc/gen/abi)".BitcoinTxUTXO2 in argument to mp.contract.NotifyMovingFundsBelowDust
```

Looking at the generated code at the problematic line:

```
transaction, err := mp.contract.NotifyMovingFundsBelowDust(
  transactorOptions,
  arg_walletPubKeyHash,
  arg_mainUtxo,
)
```

`mp.contract.NotifyFundsBelowDust` is declared in the `tbtc/gen/abi/MaintainerProxy.go` and expects `BitcoinTxUTXO2`.

`arg_mainUtxo` from `MaintainerProxy.go`  has an incorrect type though: `arg_mainUtxo abi.BitcoinTxUTXO,`

OK, that makes sense. Just like `tbtc/gen/abi/MaintainerProxy.go` this file is generated based on `tbtc/gen/abi/MaintainerProxy.abi` and in the `*.abi` file we don't have the rename.

In other words, the `after_contract_hook` is partially a result of https://github.com/ethereum/go-ethereum/issues/24627 and partially how our Go contract bindings work. The fix from https://github.com/ethereum/go-ethereum/pull/24924 will change the generated `tbtc/gen/abi/*.go` files but we'll have to adjust the `@go run github.com/keep-network/keep-common/tools/generators/ethereum` logic to follow the same naming rules as the `*.abi` file will remain unchanged.

Does it make sense?

tomaszslabon: @pdyraga Yes, it makes sense.

lukasz-zimnoch: Makes sense to me as well!

pdyraga: @tomaszslabon Can you please add a comment to `after_contract_hook` referencing https://github.com/keep-network/keep-common/issues/117 ?

tomaszslabon: Done in 800fddab0fe71758f09efb9a982a6a59da0ab25c.

pdyraga (approved):

Issue #3608, opened 2023-06-06
Improve Ethereum and Bitcoin metrics and diagnostics


pdyraga:
The client exposes the Ethereum connectivity metric which is helpful to diagnose some problems with the connection but it is not enough to say if the client is up to date with the global Ethereum state (it has access to the chain tip). 

For Bitcoin, we do not have the Electrum connectivity metric at all.

We could:
- [x] Add Ethereum diagnostic reading the most recent Ethereum block number.
- [x] Add Electrum connectivity metric.
- [x] Add Electrum diagnostic reading the most recent Bitcoin block number.
- [x] Update Grafana dashboard at status.threshold.network to include Ethereum and Bitcoin connectivity metrics for nodes.
```[tasklist]
### Tasks
- [ ] https://github.com/keep-network/keep-core/pull/3637
- [ ] https://github.com/keep-network/keep-core/pull/3648
- [ ] https://github.com/keep-network/keep-core/pull/3705
- [ ] https://github.com/keep-network/keep-core/pull/3708
```

pdyraga:
@tomaszslabon Please take care of updating Grafana once this version is deployed to testnet/mainnet.

tomaszslabon:
In #3705 we updated the test Grafana dashboard with BTC Connectivity and ETH Connectivity.
The updated dashboard can be seen here: https://monitoring.test.threshold.network.
Only the test dashboard was updated.

Things to consider in the future:
1) Update the production Grafana dashboard with BTC Connectivity and ETH Connectivity in the similar manner.
    (must be done by somebody with edit privileges cc @pdyraga, @nkuba).
2) Discuss if we should display Latest ETH block and Latest BTC block in the dashboard.
    If we want to display them, we need to move these two measurements from the `\diagnostics` endpoint  into `\metrics`   so that it can be picked by Prometheus.

PR #3637, opened 2023-06-15
Adding Ethereum latest block number diagnostic

dimpar:
Refs https://github.com/keep-network/keep-core/issues/3608

Exposing the latest block number diagnostic can be helpful to diagnose some of the problems with a node connectivity and determine if a node is catching up with the global Ethereum state.

Example output:
```
{
  "chain_info": {
    "latest_eth_block_number": 3525
  },
  "client_info": {
    "chain_address": "0x407190f04d838Ec47dad4C0e0D2a9c49d7737479",
    "network_id": "16Uiu2HAm4WrMWq9Gc7WsymsqgrwbaD1KeW2eH8eY63DTZHXbRsuE",
    "revision": "8c444b4b9",
    "version": "v2.0.0-m3-70-g8c444b4b9"
  },
  "connected_peers": null
}
```

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -211,11 +211,6 @@ func initializeClientInfo(
 		config.ClientInfo.EthereumMetricsTick,
 	)
 
-	registry.ObserveEthBlockNumber(
-		blockCounter,
-		config.ClientInfo.EthereumMetricsTick,
-	)
-
 	registry.RegisterMetricClientInfo(build.Version)
 
 	registry.RegisterConnectedPeersSource(netProvider, signing)
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -214,6 +214,7 @@ func initializeClientInfo(
 	registry.RegisterMetricClientInfo(build.Version)
 
 	registry.RegisterConnectedPeersSource(netProvider, signing)
+
 	registry.RegisterClientInfoSource(
 		netProvider,
 		signing,
@@ -221,7 +222,7 @@ func initializeClientInfo(
 		build.Revision,
 	)
 
-	registry.RegisterChainInfo(blockCounter)
+	registry.RegisterChainInfoSource(blockCounter)
 
 	logger.Infof(
 		"enabled client info endpoint on port [%v]",
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -211,6 +211,11 @@ func initializeClientInfo(
 		config.ClientInfo.EthereumMetricsTick,
 	)
 
+	registry.ObserveEthBlockNumber(
+		blockCounter,
+		config.ClientInfo.EthereumMetricsTick,
+	)
+
 	registry.RegisterMetricClientInfo(build.Version)
 
 	registry.RegisterConnectedPeersSource(netProvider, signing)
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -226,6 +226,8 @@ func initializeClientInfo(
 		build.Revision,
 	)
 
+	registry.RegisterChainInfo(blockCounter)
+
 	logger.Infof(
 		"enabled client info endpoint on port [%v]",
 		config.ClientInfo.Port,
--- a/pkg/clientinfo/diagnostics.go
+++ b/pkg/clientinfo/diagnostics.go
@@ -52,7 +52,7 @@ func (r *Registry) RegisterConnectedPeersSource(
 		for peerNetworkID, multiaddrs := range connectedPeersAddrInfo {
 			peerPublicKey, err := connectionManager.GetPeerPublicKey(peerNetworkID)
 			if err != nil {
-				logger.Error("error on getting peer public key: [%v]", err)
+				logger.Errorf("error on getting peer public key: [%v]", err)
 				continue
 			}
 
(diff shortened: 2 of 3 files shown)


pdyraga:
General question: why is the Ethereum chain tip block number a metric and not a diagnostic? 

Metrics are useful for time series, measurable, and changing over time. The number of nodes connected is a good metric. The connectivity status (0/1) to some external service like Ethereum or Electrum is a good metric. The number of pre-params in the pool is a good metric. The values can go up and down over time and they are measurable: N peers, N parameters, yes/no status.

I don't think chain time block number is a good metric, I would rather make it a diagnostic. It is always increasing and is not really measurable. We can think that "there are N blocks in the chain" but we are mostly interested in a non-measurable block number of the tip.

dimpar:
> General question: why is the Ethereum chain tip block number a metric and not a diagnostic?
Metrics are useful for time series, measurable, and changing over time. The number of nodes connected is a good metric. The connectivity status (0/1) to some external service like Ethereum or Electrum is a good metric. The number of pre-params in the pool is a good metric. The values can go up and down over time and they are measurable: N peers, N parameters, yes/no status.
I don't think chain time block number is a good metric, I would rather make it a diagnostic. It is always increasing and is not really measurable. We can think that "there are N blocks in the chain" but we are mostly interested in a non-measurable block number of the tip.

Agree with what you said. I put it under metrics because codewise it's very close to Ethereum connectivity status and required very minimal change. But I think that Eth connectivity also doesn't exactly fit under metrics then. The status is binary, 0 or 1, connection problems or not and I'd rather see it under client's diagnostics. It won't be changed like pre-params where the range is way bigger. We can certainly leave it alone cause most likely some API calls are already in place and use it. Just giving you my reasoning.

For the latest block number, I will move it to diagnostics.

cmd/start.go
```diff
@@ -221,6 +221,8 @@ func initializeClientInfo(
 		build.Revision,
 	)
 
+	registry.RegisterChainInfo(blockCounter)
```
pdyraga, line 224: Can we rename it to `RegisterChainInfoSource` to keep it consistent with `RegisterConnectedpeersSource` and `RegisterClientInfoSource`? 

Let's also take an opportunity and add an empty line between the calls.

pkg/clientinfo/diagnostics.go
```diff
@@ -118,6 +124,32 @@ func (r *Registry) RegisterClientInfoSource(
...
 }
 
+// RegisterChainInfo registers the diagnostics source providing
+// information about chains.
+func (r *Registry) RegisterChainInfo(
+	blockCounter chain.BlockCounter,
+) {
+	r.RegisterDiagnosticSource("chain_info", func() string {
+		currentBlock, err := blockCounter.CurrentBlock()
+
```
pdyraga, line 134: Let's drop the empty line.

pkg/clientinfo/diagnostics.go
```diff
@@ -118,6 +124,32 @@ func (r *Registry) RegisterClientInfoSource(
...
+		if err != nil {
+			return ""
```
pdyraga, line 136: Let's log the error with `logger.Errorf`.

pdyraga: I have also noticed we have a bunch of `logger.Error` instead of `logger.Errorf` in this file. Can we please fix them?

pkg/clientinfo/diagnostics.go
```diff
@@ -118,6 +124,32 @@ func (r *Registry) RegisterClientInfoSource(
...
+		}
+
+		chainInfo := Chain{
+			EthBlockNumber: currentBlock,
+		}
+
+		bytes, err := json.Marshal(chainInfo)
+		if err != nil {
+			logger.Error("error on serializing chain info to JSON: [%v]", err)
```
pdyraga, line 145: Shouldn't this be `Errorf`?

cmd/start.go
dimpar: Sure [0493b72](https://github.com/keep-network/keep-core/pull/3637/commits/0493b72b3f8e4c41684037eaa534beef7433474c)

pkg/clientinfo/diagnostics.go
dimpar: [0493b72](https://github.com/keep-network/keep-core/pull/3637/commits/0493b72b3f8e4c41684037eaa534beef7433474c)

pkg/clientinfo/diagnostics.go
dimpar: Sure [0493b72](https://github.com/keep-network/keep-core/pull/3637/commits/0493b72b3f8e4c41684037eaa534beef7433474c)

pkg/clientinfo/diagnostics.go
dimpar: Replaced with Errorf [0493b72](https://github.com/keep-network/keep-core/pull/3637/commits/0493b72b3f8e4c41684037eaa534beef7433474c)

pdyraga:
Tested locally, works as advertised!

```
❯ curl -D - localhost:8081/diagnostics
HTTP/1.1 200 OK
Date: Tue, 20 Jun 2023 13:07:22 GMT
Content-Length: 323
Content-Type: text/plain; charset=utf-8

{
  "chain_info": {
    "latest_eth_block_number": 741
  },
  "client_info": {
    "chain_address": "0x146748a2b46B99ee1470b587bc9812Ea59b79597",
    "network_id": "16Uiu2HAm2GxAmGNcQ6Aq8GMEYsYCnuZqbWLqCKD5fGWciwvUMtoj",
    "revision": "0493b72b3",
    "version": "v2.0.0-m3-73-g0493b72b3"
  },
  "connected_peers": null
}
```

Issue #3609, opened 2023-06-06
Client-side support for redemptions


pdyraga:
The following items need to be implemented so that the client can perform a redemption:
- [x] Implement code to assemble Bitcoin redemption transaction. (https://github.com/keep-network/keep-core/pull/3607)
- [x] Add `WalletCoordinator` code for redemption request. (https://github.com/keep-network/tbtc-v2/pull/633)
- [ ] Implement client redemption action reacting on events from `WalletCoordinator`, assembling the transaction, triggering signing, and broadcasting signed Bitcoin transaction. (https://github.com/keep-network/keep-core/pull/3639, https://github.com/keep-network/keep-core/pull/3647)

For the `WalletCoordinator` part, we need to think if any off-chain validation is needed. It may happen the only thing we need to validate are timeouts and if the redemption request exists. The `Bridge` is just taking tBTC from someone and storing the redemption address on-chain. No manipulation is possible on the coordinator's side.

Issue #3611, opened 2023-06-07
Couldn't find unmarshaler for type ... errors when sweeping


pdyraga:
The following errors were logged on some clients when the first mainnet sweep was being executed:
```
(...)
2023-06-07T09:30:43.530Z	INFO	keep-tbtc	tbtc/deposit_sweep.go:225	deposit [5/5] - fetching deposit's extra data	{"wallet": "0x04b0a483e97dfbb15e88ecbc2ef8f7e37776dd713eed54089f87704a4fbae0442aa3ff5a0ec7f4d52bf4680b5b9b70be9accb64970df11b96abb61d48be7d8db8b", "action": "DepositSweep", "startBlock": 17427619, "step": "validateProposal"}
2023-06-07T09:30:43.828Z	INFO	keep-tbtc	tbtc/deposit_sweep.go:292	calling chain for proposal validation	{"wallet": "0x04b0a483e97dfbb15e88ecbc2ef8f7e37776dd713eed54089f87704a4fbae0442aa3ff5a0ec7f4d52bf4680b5b9b70be9accb64970df11b96abb61d48be7d8db8b", "action": "DepositSweep", "startBlock": 17427619, "step": "validateProposal"}
2023-06-07T09:30:43.949Z	INFO	keep-tbtc	tbtc/deposit_sweep.go:299	deposit sweep proposal is valid	{"wallet": "0x04b0a483e97dfbb15e88ecbc2ef8f7e37776dd713eed54089f87704a4fbae0442aa3ff5a0ec7f4d52bf4680b5b9b70be9accb64970df11b96abb61d48be7d8db8b", "action": "DepositSweep", "startBlock": 17427619, "step": "validateProposal"}
2023-06-07T09:30:46.031Z	INFO	keep-libp2p	libp2p/libp2p.go:242	number of connected peers: [147]
2023-06-07T09:30:48.294Z	ERROR	keep-libp2p	libp2p/channel.go:286	couldn't find unmarshaler for type [protocol_announcer/announcement_message]
2023-06-07T09:30:48.294Z	ERROR	keep-libp2p	libp2p/channel.go:286	couldn't find unmarshaler for type [protocol_announcer/announcement_message]
2023-06-07T09:30:48.295Z	ERROR	keep-libp2p	libp2p/channel.go:286	couldn't find unmarshaler for type [protocol_announcer/announcement_message]
(...)
[...]

PR #3647, opened 2023-06-21
Redemption prototype: Validate proposal and broadcast Bitcoin transaction

lukasz-zimnoch:
Closes https://github.com/keep-network/keep-core/issues/3609

Here we add the remaining part necessary to enable redemption support on the client side. Specific changes include:

### Extracting common components used across wallet actions

During the work on this PR, we identified that some code can be detached from the existing `depositSweepAction` and live as an independent being that can be reused across different wallet actions. Those components are:
- The `walletTransactionExecutor` which is a thin layer encapsulating some logic necessary to sign and broadcast wallet Bitcoin transactions
- The `EnsureWalletSyncedBetweenChains` function which allows to make sure the wallet is synced between the Bitcoin and host chain before starting an action

Thanks to that change, we avoid code duplication across wallet actions.

### Implementation of the `redemptionAction`

We introduce a dedicated `redemptionAction` that encapsulates everything that must be done to perform a successful redemption. Its structure is very similar to the existing `depositSweepAction`. It starts by making the proposal validation, then constructs, signs, and broadcasts the redemption transaction over the Bitcoin network.

### DIFF
--- a/pkg/bitcoin/bitcoin.go
+++ b/pkg/bitcoin/bitcoin.go
@@ -45,6 +45,18 @@ func readCompactSizeUint(varLenData []byte) (CompactSizeUint, int, error) {
 	return CompactSizeUint(csu), wire.VarIntSerializeSize(csu), nil
 }
 
+// writeCompactSizeUint writes the provided CompactSizeUint into a
+// byte slice.
+func writeCompactSizeUint(csu CompactSizeUint) ([]byte, error) {
+	var buffer bytes.Buffer
+	err := wire.WriteVarInt(&buffer, 0, uint64(csu))
+	if err != nil {
+		return nil, err
+	}
+
+	return buffer.Bytes(), nil
+}
+
 // ByteOrder represents the byte order used by the Bitcoin byte arrays. The
 // Bitcoin ecosystem is not totally consistent in this regard and different
 // byte orders are used depending on the purpose.
--- a/pkg/bitcoin/bitcoin.go
+++ b/pkg/bitcoin/bitcoin.go
@@ -23,6 +23,11 @@ import (
 // >= 0x10000 && <= 0xffffffff             |   5   | 0xfe followed by the number as uint32
 // >= 0x100000000 && <= 0xffffffffffffffff |   9   | 0xff followed by the number as uint64
 //
+// Worth noting, the encoded number value is represented using the little-endian
+// byte order. For example, to convert the compact size uint 0xfd0302, the
+// 0xfd prefix must be skipped and the 0x0302 must be reversed to 0x0203 and
+// then converted to a decimal number 515.
+//
 // For reference, see:
 // https://developer.bitcoin.org/reference/transactions.html#compactsize-unsigned-integers
 type CompactSizeUint uint64
--- a/pkg/tbtc/chain_test.go
+++ b/pkg/tbtc/chain_test.go
@@ -59,6 +59,12 @@ type localChain struct {
 	depositSweepProposalValidationsMutex sync.Mutex
 	depositSweepProposalValidations      map[[32]byte]bool
 
+	pendingRedemptionRequestsMutex sync.Mutex
+	pendingRedemptionRequests      map[[32]byte]*RedemptionRequest
+
+	redemptionProposalValidationsMutex sync.Mutex
+	redemptionProposalValidations      map[[32]byte]bool
+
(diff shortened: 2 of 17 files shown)


pkg/tbtc/chain_test.go
```diff
@@ -59,6 +59,12 @@ type localChain struct {
 	depositSweepProposalValidationsMutex sync.Mutex
 	depositSweepProposalValidations      map[[32]byte]bool
 
+	pendingRedemptionRequestsMutex sync.Mutex
+	pendingRedemptionRequests      map[[32]byte]*RedemptionRequest
+
+	redemptionProposalValidationsMutex sync.Mutex
```
pdyraga, line 65: Non-blocking but I think we should start using a single mutex per chain test class. This is escalating with every PR.

pkg/bitcoin/bitcoin.go
```diff
@@ -40,6 +45,18 @@ func readCompactSizeUint(varLenData []byte) (CompactSizeUint, int, error) {
 	return CompactSizeUint(csu), wire.VarIntSerializeSize(csu), nil
 }
 
+// writeCompactSizeUint writes the provided CompactSizeUint into a
+// byte slice.
+func writeCompactSizeUint(csu CompactSizeUint) ([]byte, error) {
+	var buffer bytes.Buffer
+	err := wire.WriteVarInt(&buffer, 0, uint64(csu))
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah...

pkg/tbtc/redemption_test.go
```diff
@@ -10,6 +13,168 @@ import (
...
+
+			// Choose an arbitrary start block and expiration time.
+			proposalProcessingStartBlock := uint64(100)
+			proposalExpiresAt := now.Add(4 * time.Hour)
+
+			// Simulate the on-chain proposal validation passes with success.
+			err = hostChain.setRedemptionProposalValidationResult(
+				proposal,
+				true,
+			)
```
pdyraga, line 86: For a followup PR: as a minimum, develop a test that ensures we do not proceed with signing when the validation fails.

pkg/tbtc/wallet.go
```diff
@@ -260,6 +432,92 @@ func DetermineWalletMainUtxo(
...
+//   Such a transactions indicate a possible state change in the future
+//   but their outcome cannot be determined at this stage so, the wallet
+//   should not perform new actions at the moment.
+func EnsureWalletSyncedBetweenChains(
+	walletPublicKeyHash [20]byte,
+	walletMainUtxo *bitcoin.UnspentTransactionOutput,
+	btcChain bitcoin.Chain,
+) error {
+	// Take the recent transactions history for the wallet.
+	history, err := btcChain.GetTransactionsForPublicKeyHash(walletPublicKeyHash, 5)
```
pdyraga, line 455: I know it's just a code we moved but... why do we take 5 transactions and not just one?

pkg/tbtc/node.go
```diff
@@ -470,8 +470,81 @@ func (n *node) handleRedemptionProposal(
 	startBlock uint64,
 	delayBlocks uint64,
 ) {
-	// TODO: Implementation.
-	panic("not implemented yet")
+	wallet, ok := n.walletRegistry.getWalletByPublicKeyHash(
```
pdyraga, line 473: I am afraid we may have the same problems as in sweeping with unregistered unmarshalers: https://github.com/keep-network/keep-core/issues/3611.

I don't think it's blocking this PR but it's something we should address both for sweeps and redemptions before doing the user acceptance tests.

pkg/tbtc/wallet.go
lukasz-zimnoch: Just in case. This function tries to be universal and must find the actual main UTXO even if it doesn't come from the recent transaction.

lukasz-zimnoch: Just in case :D

pkg/tbtc/node.go
lukasz-zimnoch: Yeah, sure!

pkg/tbtc/chain_test.go
lukasz-zimnoch: Sure, we can do it in a follow-up PR.

pdyraga (approved):

Issue #3615, opened 2023-06-07
Automated SPV maintainer for submitting sweep and redemption proofs


pdyraga:
We should implement an automated maintainer for SPV proofs: both redemption and sweep. This should be a part of the client's maintainer module, activated/deactivated with a config flag.

Just in case, CLI for submitting redemption proofs should also be available.

When working on this issue please ensure the approach is consistent with the coordinator maintainer module from #3614.
```[tasklist]
### Pull Requests
- [ ] https://github.com/keep-network/keep-core/pull/3646
- [ ] https://github.com/keep-network/keep-core/pull/3649
- [ ] https://github.com/keep-network/keep-core/pull/3673
- [ ] https://github.com/keep-network/keep-core/pull/3669
- [ ] https://github.com/keep-network/keep-core/pull/3670
- [ ] https://github.com/keep-network/keep-core/pull/3667
- [ ] https://github.com/keep-network/keep-core/pull/3672
- [ ] https://github.com/keep-network/keep-core/pull/3675
- [ ] https://github.com/keep-network/keep-core/pull/3676
- [ ] https://github.com/keep-network/keep-core/pull/3680
```

PR #3649, opened 2023-06-21
SPV deposit sweep maintainer logic

tomaszslabon:
Refs: https://github.com/keep-network/keep-core/issues/3615.
This PR adds implementation of deposit sweep proofs in SPV maintainer.
The SPV maintainer periodically checks if there are unproven deposit sweep transactions.
If there are such transactions, it assembles the proof and submits it.

Example command:
```
KEEP_ETHEREUM_PASSWORD="password" ./keep-client maintainer --spv --spv.historyDepth 200000 --config maintainer.toml 
```


pkg/maintainer/spv/chain.go
```diff
@@ -27,7 +27,30 @@ type Chain interface {
 		fundingOutputIndex uint32,
 	) (*tbtc.DepositChainRequest, error)
 
+	// TODO: Most likely merge it with `GetDepositRequest` function.
+	// Deposits gets the on-chain deposit request for the given funding
+	// transaction hash and output index. If the deposit does not exist it
+	// returns a zero-filled structure without an error.
```
pdyraga, line 33: Agreed, those functions are mostly identical right now.

pkg/maintainer/spv/spv.go
```diff
@@ -59,8 +63,271 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
 func (sm *spvMaintainer) maintainSpv(ctx context.Context) error {
 	logger.Infof("Maintaining SPV proof...")
```
pdyraga, line 64: I would drop this line given we log earlier:

https://github.com/keep-network/keep-core/blob/8030414dee6ea771e69b21278163e8aee1cb6319/pkg/maintainer/spv/spv.go#L40

pkg/maintainer/spv/spv.go
```diff
@@ -59,8 +63,271 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+	// a list of unique wallet public key hashes.
+	walletPublicKeyHashes := uniqueWalletPublicKeyHashes(
+		depositSweepTransactionProposals,
+	)
+
+	unprovenDepositSweepTransactions := []*bitcoin.Transaction{}
+
+	for _, walletPublicKeyHash := range walletPublicKeyHashes {
+		// TODO: Should we check the wallet's state before attempting to submit
+		//       the deposit sweep proof?
```
pdyraga, line 167: Hmm... why? I can't think of a scenario when we should wait with providing SPV proof but of course, I could miss something. In my opinion, it should happen as soon as possible - if the wallet was late with signing, providing SPV proof right away will allow the next action to be performed.

pkg/maintainer/spv/spv.go
```diff
@@ -59,8 +63,271 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+		// TODO: Think what the limit of transactions retrieved should be.
+		walletTransactions, err := sm.btcChain.GetTransactionsForPublicKeyHash(
+			walletPublicKeyHash,
+			5,
```
pdyraga, line 171: In theory, the wallet should not execute the next action if the previous one was not yet completed. I think checking 5 as a default is fine, but I would also allow this value to be passed as a parameter.

pdyraga: Oh, wait. There is a chance the wallet performed e.g. 20 redemptions after sweeping. This function fetches all Bitcoin transactions no matter the type.

Assuming we execute sweep every 48 hours and redemption every 3 hours, it seems that the default value should be closer to 20.

pkg/maintainer/spv/spv.go
```diff
@@ -59,8 +63,271 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+			// UTXO among the transaction's inputs since only one input may have
+			// such a structure that the calculated hash will match the wallet's
+			// main UTXO hash stored on-chain.
+			isMainUtxo, err := sm.isInputCurrentWalletsMainUTXO(
+				fundingTransactionHash,
+				fundingOutpointIndex,
+				walletPublicKeyHash,
+			)
+			if err != nil {
+				return false, fmt.Errorf("failed to check is input is the main UTXO")
```
pdyraga, line 248:
```suggestion
				return false, fmt.Errorf("failed to check if input is the main UTXO")
```

pkg/maintainer/spv/spv.go
tomaszslabon: We could filter out wallets that are not `Live` or `MovingFunds` for performance reasons.
If the wallet is `Closed`, we could skip it immediately.

pkg/maintainer/spv/spv.go
```diff
@@ -59,8 +63,271 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+		Outpoint: &bitcoin.TransactionOutpoint{
+			TransactionHash: fundingTxHash,
+			OutputIndex:     fundingOutputIndex,
+		},
+		Value: fundingOutputValue,
+	})
+
+	// Get the wallet and check if its main UTXO matches the calculated hash.
+	// TODO: `GetWallet` may return an error if the wallet is not found.
+	//       Should we return an error in such case or continue?
```
pdyraga, line 304: IMO, we should return an error. Most likely, this is something bad on the infrastructure level (e.g. connection error to Ethereum node). Less likely, this is some terrible bug in the code.

pkg/maintainer/spv/spv.go
pdyraga: That's a good point, I thought this is about the wallet lock state in the `WalletCoordinator` contract. If this is easy, I think we should do it. This is not an issue now when all wallets are Live but we'll get to the point when this check will matter.

pkg/maintainer/spv/spv.go
```diff
@@ -59,8 +63,271 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+	depositSweepTransactions, err := sm.getUnprovenDepositSweepTransactions()
+	if err != nil {
+		return fmt.Errorf(
+			"failed to get unproven deposit sweep transactions: [%v]",
+			err,
+		)
+	}
+
+	// TODO: Consider handling a situation in which the block headers in the
+	//       proof span multiple Bitcoin difficulty epochs.
```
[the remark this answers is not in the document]
tomaszslabon: And if it failed on-chain, the maintainer would keep trying to submit it anyway on each round of processing.
It would lead to unnecessary burning of funds from the `ReimbursementPool`.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 32150aba63c2cea884d25fa231532958169bdcb1.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 28e3faa5fa4eda8ca3f2f451f60ac5889aa2cd40.

pkg/maintainer/spv/spv.go
tomaszslabon: Agree. Left the code as is (we return an error if wallet not found) and removed the TODO in ae31a234b0f5ba60571f63479c158d633c15a4eb.

pkg/maintainer/spv/spv.go
```diff
@@ -59,8 +63,271 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
 
 	return nil
 }
+
+func (sm *spvMaintainer) getUnprovenDepositSweepTransactions() (
+	[]*bitcoin.Transaction,
+	error,
+) {
+	// TODO: Limit how far in the past we are looking for the events.
+	//       Possibly store latest checked height in memory or file.
```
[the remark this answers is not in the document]
tomaszslabon: Added the `HistoryDepth` parameter in d2e4466b6f5a17f3c8854c280472c2110771cf1e.
The default value for it is `40320`, the approximate number of Ethereum blocks in one week:
7 * 24 * 3600 / 15 = 40320

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 45bb022c94708fe000003e3ea54112c81f967cab.
The default value for the parameter is `20`.

pkg/maintainer/spv/chain.go
tomaszslabon: As joining `GetDepositRequest` and `Deposits` modifies some code unrelated to this PR I decided to handle it in a small separate PR: https://github.com/keep-network/keep-core/pull/3654.
Once the PR is merged to main, I will remove the `Deposits` function.

pkg/maintainer/spv/spv.go
tomaszslabon: Added a check for wallet state in 8abdeecdac4f3cc85fce44b60968bab6d68b7d94.

pkg/chain/ethereum/tbtc.go
```diff
@@ -1126,6 +1186,37 @@ func (tc *TbtcChain) GetDepositRequest(
 	}, nil
 }
 
+func (tc *TbtcChain) Deposits(
```
pdyraga, line 1189: Note to self: this should be removed after #3654.

pkg/maintainer/spv/config.go
```diff
@@ -9,8 +9,23 @@ type Config struct {
...
+	// HistoryDepth is the number of blocks to look back from the current block
+	// when searching for past deposit sweep proposal submitted events.
+	// This depth determines how far into the past the system will consider
+	// events for processing.
+	HistoryDepth uint64
+
+	// TransactionLimit sets the maximum number of confirmed transactions
+	// returned when getting transactions for a public key hash. For example,
+	// if set to `20`, only the latest twenty transactions will be returned.
+	TransactionLimit int
```
pdyraga, line 21: Let's add some extra context for future-us and other contributors. The values of those two parameters are crucial. My proposition:

```suggestion
	// HistoryDepth is the number of blocks to look back from the current block
	// when searching for past deposit sweep proposal submitted events.
	// To find Bitcoin transactions for which the SPV proof should be submitted,
	// the maintainer first inspects the sweep proposal events from the wallet
	// coordinator contract. This depth determines how far into the past the
	// system will consider events for processing. This value must not be too
	// high so that the event lookup is efficient. At the same time, this value
	// can not be too low to make sure all performed and not yet proven sweeps
	// can be found.
	HistoryDepth uint64

	// TransactionLimit sets the maximum number of confirmed transactions
	// returned when getting transactions for a public key hash. Once the
	// maintainer establishes the list of proposed sweeps based on the sweep
	// proposal events from the wallet coordinator contract, it needs to check
	// Bitcoin transactions executed by the wallet. Then, it tries to find the
	// sweep among those transactions. For example, if set to `20`, only the
	// latest twenty transactions will be returned. This value must not be too
	// high so that the transaction lookup is efficient. At the same time, this
	// value can not be too low to make sure the performed sweep can be found in
	// case the wallet decided to execute some other Bitcoin transaction after
	// the yet-not-proven sweep.
	TransactionLimit int
```

pkg/maintainer/spv/spv.go
```diff
@@ -1,22 +1,63 @@
...
+	// back from the current block when searching for past deposit sweep
+	// proposal submitted events. The value is the approximate number of
+	// Ethereum blocks in a week.
+	spvDefaultHistoryDepth = 40320
+
+	// Default value for the limit of transactions returned for a given wallet
+	// public key hash. The value is based on the frequency of how often deposit
+	// sweep and redemption transaction will happen. Deposit sweep transactions
+	// are assumed to happen every 48h. Redemption transactions are assumed to
+	// happen every 3h.
```
pdyraga, line 29: We could also mention this small yet important wallet implementation detail.

One could argue we shouldn't mention wallet implementation details in the maintainer docs and there is a truth in it but... this is quite a crucial piece of knowledge and I think it's worth mentioning here in case we need to make some changes in the SPV maintainer logic in the future.

```suggestion
	// Default value for the limit of transactions returned for a given wallet
	// public key hash. The value is based on the frequency of how often deposit
	// sweep and redemption transaction will happen. Deposit sweep transactions
	// are assumed to happen every 48h. Redemption transactions are assumed to
	// happen every 3h. The wallet should refuse any proposals from the
	// coordinator if the previously executed Bitcoin transaction was not proved
	// to the Bridge yet so in theory, the value of 1 should be enough. We make
	// it a bit higher - better to be safe than sorry.
```

pkg/maintainer/spv/spv.go
```diff
@@ -57,10 +98,302 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
 }
 
 func (sm *spvMaintainer) maintainSpv(ctx context.Context) error {
```
pdyraga, line 100: I really like how the control flow is organized in those functions. The `for` loops, the `select`s, and how the functions are divided. It makes a really logical structure that is easy to follow. 👌🏻

pkg/maintainer/spv/spv.go
```diff
@@ -57,10 +98,302 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+	for _, walletPublicKeyHash := range walletPublicKeyHashes {
+		wallet, err := sm.chain.GetWallet(walletPublicKeyHash)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get wallet: [%v]", err)
+		}
+
+		// Skip if the wallet is not `Live` or `MovingFunds`.
+		if wallet.State != 1 && wallet.State != 2 {
+			continue
+		}
```
pdyraga, line 233: The values of `1` and `2` are Solidity/EVM implementation details. I think we should modify `WalletChainData` structure to return `State` not as `uint8` but as `WalletState` enum we will introduce. We have `WalletActionType` and `parseWalletActionType` in the Ethereum implementation. We should introduce identical enum for `WalletState` and have an identical parsing function. I don't think we should necessarily reuse the same code given `WalletActionType` comes from the `WalletCoordinator` contract and it is another type.

pkg/chain/ethereum/tbtc.go
tomaszslabon: Removed in cbd754abb9a1718e495209884eb4714db6e5ad68.

pkg/maintainer/spv/config.go
tomaszslabon: Done in 72058add43edb65b5c2176770366dcc99ced7685.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 72058add43edb65b5c2176770366dcc99ced7685.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 2386f5ba3d386d0db1144f88dc4b76631dc9901b.
Unfortunately, I had to prefix each state constant:
```
const (
	StateUnknown WalletState = iota
	StateLive
	StateMovingFunds
	StateClosing
	StateClosed
	StateTerminated
)
``` 
with `State...`, because we already have `MovingFunds` constant in `WalletActionType`.

pkg/maintainer/spv/spv.go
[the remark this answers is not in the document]
tomaszslabon: I'll do the TODO in a follow-up PR.

pkg/maintainer/spv/spv.go
pdyraga: I propose that we prefix all types of `WalletActionType` with `Action*` in a separate PR to keep it consistent.

pkg/maintainer/spv/spv.go
```diff
@@ -1,35 +1,83 @@
...
 )
 
 var logger = log.Logger("keep-maintainer-spv")
 
+const (
+	// Default value for history depth which is the number of blocks to look
+	// back from the current block when searching for past deposit sweep
+	// proposal submitted events. The value is the approximate number of
+	// Ethereum blocks in a week.
+	spvDefaultHistoryDepth = 40320
```
pdyraga, line 24: Shouldn't it be 50400 assuming 12s block time? 🤔

pkg/maintainer/spv/spv.go
```diff
@@ -57,10 +105,348 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+
+		// Deserialize the first block header and verify its difficulty.
+		rawBlockHeader := [bitcoin.BlockHeaderByteLength]byte{}
+		copy(
+			rawBlockHeader[:],
+			proof.BitcoinHeaders[:bitcoin.BlockHeaderByteLength],
+		)
+		firstBlockHeader := bitcoin.BlockHeader{}
+		firstBlockHeader.Deserialize(rawBlockHeader)
+		firstBlockHeaderDifficulty := firstBlockHeader.Difficulty()
```
pdyraga, line 182: What do you think about moving this piece of code to a function exposed by `bitcoin.SpvProof`? We'd use it here as `proof.FirstBlockHeaderDifficulty()`, for example.

pkg/maintainer/spv/spv.go
```diff
@@ -57,10 +105,348 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+			// header in the proof does not match the current or the
+			// previous difficulty seen on-chain. The reason for this is
+			// most likely that the proof consists of block headers from the
+			// beginning of a new Bitcoin difficulty epoch and the epoch is
+			// not proven on-chain yet. In that case the transaction will be
+			// proven in the future. The other reason could be that the
+			// transaction is older than the last two Bitcoin difficulty
+			// epoch. In that case the transaction will soon leave the
+			// sliding window of recent transactions.
+			continue
```
pdyraga, line 205: I think we should log a warning in this case. The second case is pretty dangerous. 

We should also make sure those warnings are included in our monitoring rules cc @lukasz-zimnoch.

pkg/maintainer/spv/spv.go
```diff
@@ -57,10 +105,348 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+		wallet, err := sm.spvChain.GetWallet(walletPublicKeyHash)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get wallet: [%v]", err)
+		}
+
+		if wallet.State != tbtc.StateLive &&
+			wallet.State != tbtc.StateMovingFunds {
+			// The wallet can only submit deposit sweep proofs if it's `Live` or
+			// `MovingFunds`. If the state is different skip it.
+			continue
```
pdyraga, line 285: Let's add an info-level log for this case.

pkg/maintainer/spv/spv.go
```diff
@@ -57,10 +105,348 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+			return fmt.Errorf(
+				"failed to get transaction confirmations: [%v]",
+				err,
+			)
+		}
+
+		if accumulatedConfirmations < requiredConfirmations {
+			// Skip the transaction as it has not accumulated enough
+			// confirmations. It will be proven later.
+			continue
```
pdyraga, line 162: Let's add an info-level log for this case.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in a separate PR https://github.com/keep-network/keep-core/pull/3662.

pkg/maintainer/spv/spv.go
tomaszslabon: It should. For some reason I assumed one Ethereum block is about `15s`, but it is indeed close to `12s` (at least in recent months).
Done in 004af961764e3fa51e63aecfa0485b9a32e1990e.

pkg/chain/ethereum/bitcoin_difficulty.go
```diff
@@ -224,3 +225,19 @@ func (bdc *BitcoinDifficultyChain) CurrentEpoch() (uint64, error) {
...
+
+// GetCurrentAndPrevEpochDifficulty returns the difficulties of the current
+// and previous Bitcoin epochs.
+func (bdc *BitcoinDifficultyChain) GetCurrentAndPrevEpochDifficulty() (
+	*big.Int, *big.Int, error,
+) {
+	difficulties, err := bdc.lightRelay.GetCurrentAndPrevEpochDifficulty()
+	if err != nil {
+		return nil, nil, fmt.Errorf(
+			"failed to get epoch difficulties: [%v]",
```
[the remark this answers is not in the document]
tomaszslabon: Done in e08ef679f7f3ac2f32982ed52917c4dc1fd5e024.

pkg/maintainer/spv/spv.go
tomaszslabon: It makes sense. Done in 7cd1de0b8dbe6cc5ed8db4c8854ad2f46484d89e.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 132ac7d666c8964b47a71255b3c50334f240fe22.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 132ac7d666c8964b47a71255b3c50334f240fe22.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 132ac7d666c8964b47a71255b3c50334f240fe22.

pkg/maintainer/spv/spv.go
```diff
@@ -49,18 +57,384 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
 
 func (sm *spvMaintainer) maintainSpv(ctx context.Context) error {
-	logger.Infof("Maintaining SPV proof...")
+	for {
+		if err := sm.proveDepositSweepTransactions(); err != nil {
+			return fmt.Errorf(
+				"error while proving deposit sweep transactions: [%v]",
+				err,
+			)
+		}
```
pdyraga, line 74: Let's align it with how we log the activity for the wallet coordinator maintainer:

```suggestion
		logger.Infof("starting deposit sweep proof task execution...")

		if err := sm.proveDepositSweepTransactions(); err != nil {
			return fmt.Errorf(
				"error while proving deposit sweep transactions: [%v]",
				err,
			)
		}

		logger.Infof(
			"deposit sweep proof task run completed; next run in [%s]",
			sm.config.IdleBackOffTime,
		)
```

Compare:

https://github.com/keep-network/keep-core/blob/fc1bbf3974f4562f65d74208f11c976fc04fe0d7/pkg/maintainer/wallet/wallet.go#L69-L96

pkg/maintainer/spv/spv.go
```diff
@@ -49,18 +57,384 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+}
+
+func (sm *spvMaintainer) proveDepositSweepTransactions() error {
+	depositSweepTransactions, err := sm.getUnprovenDepositSweepTransactions()
+	logger.Infof(
+		"Found [%d] unproven deposit sweep transaction(s)",
```
pdyraga, line 90: We should start log messages from a lower-case:

```suggestion
		"found [%d] unproven deposit sweep transaction(s)",
```

Let's align all other log messages in this file.

pkg/maintainer/spv/spv.go
```diff
@@ -49,18 +57,384 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
-	// TODO: Implement. For now, just wait.
-	<-ctx.Done()
+		if accumulatedConfirmations < requiredConfirmations {
+			// Skip the transaction as it has not accumulated enough
+			// confirmations. It will be proven later.
+			logger.Infof(
+				"Skipped proving deposit sweep transaction [%s] due to "+
+					"transaction not having enough confirmations yet",
+				transactionHashStr,
+			)
```
pdyraga, line 148: Let's be more specific:

```suggestion
			// Skip the transaction as it has not accumulated enough
			// confirmations. It will be proven later.
			logger.Infof(
				"Skipped proving deposit sweep transaction [%s]; transaction "+
					"has [%v/%v] confirmations",
				transactionHashStr,
				accumulatedConfirmations,
				requiredConfirmations,
			)
```

pkg/maintainer/spv/spv.go
```diff
@@ -49,18 +57,384 @@ func (sm *spvMaintainer) startControlLoop(ctx context.Context) {
...
+			// not proven on-chain yet. In that case the transaction will be
+			// proven in the future. The other reason could be that the
+			// transaction is older than the last two Bitcoin difficulty
+			// epoch. In that case the transaction will soon leave the
+			// sliding window of recent transactions.
+			logger.Warnf(
+				"Skipped proving deposit sweep transaction [%s] due to "+
+					"difficulties mismatch between proof and relay",
+				transactionHashStr,
+			)
```
pdyraga, line 179: Let's be more specific:

```suggestion
			logger.Warnf(
				"skipped proving deposit sweep transaction [%s]; found "+
					"difficulties mismatch between proof and relay: "+
					"proof first header difficulty: [%v], "+
					"previous relay difficulty: [%v], "+
					"current relay difficultry: [%v]",
				transactionHashStr,
				firstBlockHeaderDifficulty,
				previousDifficulty,
				currentDifficulty,
			)
```

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 61e77e4760e35d764dd85340f0251116f7ccc49f.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 69922dff13522866329c1ca88f804e025ecefb2c.

pdyraga (approved):

pdyraga:
We agreed to implement unit tets for this code in a separate PR so that we can work on the remaining SPV maintainer logic in parallel. 

The remaining work is captured on the tasks list in https://github.com/keep-network/keep-core/issues/3615.

I tested this code locally and it submitted deposit sweep SPV proof successfully.

pkg/maintainer/spv/spv.go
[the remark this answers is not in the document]
tomaszslabon: The followup PR: https://github.com/keep-network/keep-core/pull/3667.

Issue #3632, opened 2023-06-13
Separate chain interface for the coordinator package


pdyraga:
> I know it became a common practice recently and `tbtc.Chain` exposes functions for `pkg/coordinator` but I think it's not a proper approach regarding Go interfaces. The `tbtc.Chain` was supposed to define chain expectations just for the `pkg/tbtc`. If we start reusing it from another package, we are introducing new dependencies that are probably unnecessary. Additionally, the `tbtc.Chain` exposes methods that are not used by `pkg/tbtc` which causes confusion and the local chain implementation (`tbtc.localChain`) has to implement all those unused methods as well.
> 
> I think the right approach is to make `pkg/coordinator` define its own `coordinator.Chain` interface with a subset of methods relevant just to the coordinator (just like the `pkg/maintainer`). I know this may cause some duplications but they are low-cost as we would duplicate interfaces, not implementations. In return, we would receive fewer dependencies between packages and smaller, more precise interfaces.

Posted by @lukasz-zimnoch [here](https://github.com/keep-network/keep-core/pull/3605#discussion_r1226716126) and further discussed [here](https://github.com/keep-network/keep-core/pull/3626#discussion_r1227829389). 

Let's make a separate chain interface for the `coordinator` package and clean up chain interface in `tbtc` package. This approach helps to avoid import cycles but also makes test structs such as the one in `pkg/coordinator/tbtc_chain_test.go` way shorter as we only need to implement functions essential to the given package.

PR #3650, opened 2023-06-26
Separate chain interface for the coordinator package

nkuba:
In this PR we separate `coordinator.Chain` interface definition from the `tbtc.Chain` interface. This change makes the interfaces cleaner as each package requires only the functions that it is actually using.

Closes: https://github.com/keep-network/keep-core/issues/3632


pkg/tbtc/chain.go
```diff
@@ -359,6 +298,7 @@ type WalletCoordinatorChain interface {
 	// proposal events according to the provided filter or unfiltered if the
 	// filter is nil. Returned events are sorted by the block number in the
 	// ascending order, i.e. the latest event is at the end of the slice.
+	// TODO: This function is not used, consider removing it.
 	PastDepositSweepProposalSubmittedEvents(
```
nkuba, line 302: @lukasz-zimnoch Can we safely remove it? This function is not used, but there is some existing local chain implementation, which I'm not sure I can remove.

pkg/coordinator/chain.go
```diff
@@ -1,17 +1,69 @@
...
-	// TODO: Change to something more specific once https://github.com/keep-network/keep-core/issues/3632
-	// is handled.
-	tbtc.Chain
+	// GetDepositRequest gets the on-chain deposit request for the given
+	// funding transaction hash and output index. Returns an error if the
+	// deposit was not found.
+	GetDepositRequest(
+		fundingTxHash bitcoin.Hash,
+		fundingOutputIndex uint32,
+	) (*tbtc.DepositChainRequest, error)
```
nkuba, line 34: As for the functions, it's clear that we want to define them separately for each package, even if it means duplication. But what about the types the functions are expecting or returning?
I don't think duplication is the right way. Alternatively to the current solution where we just import the types from the `tbtc` package, we could alias them, but it doesn't make much sense.

I'm curious about your opinion @pdyraga @lukasz-zimnoch.

lukasz-zimnoch: In my opinion, we should look at the packages as some bounded contexts of the same domain:
- `pkg/tbtc` is the context corresponding to the tbtc wallet
- `pkg/coordinator` is the context of a tbtc wallet coordinator

That said, both should define domain objects (i.e. types) and functions specific to their contexts. However, some of the domain may be shared between those two contexts and this is the case we observe here. I think we have two options:
1. Declare the shared objects in one of the contexts
2. Extract the shared objects to a separate place

I think option 2 is worse in our case as it will introduce unnecessary confusion. In my opinion, we should aim for option 1 as it is more "natural" for our codebase. 

The only question is: "Which package should hold the shared objects?". In our case, the `pkg/tbtc` is kind of a core package while the `pkg/coordinator` is kind of an overlay specialized for the coordination part. If we agree about that, this implies that all shared objects should live in the `pkg/tbtc` package.

pdyraga: I would be fine with the duplication for now but I think the problem with the duplication is that those structs are separate types and the chain implementation struct would have to redeclare the same function multiple times. For example:
```
func GetDepositRequest(...) coordinator.DepositChainRequest // pkg/coordinator chain interface
func GetDepositRequest(...) spv.DepositChainRequest // pkg/maintainer/spv chain interface
func GetDepositRequest(...) wallet.DepositChainRequest // pkg/maintainer/wallet chain interface
```

The current solution is the lesser evil but I think we can do better. Instead of creating a dependency to `tbtc` package, we could have a separate package with the common type declarations. 

```
pkg/chain/types/types.go

pkg/tbtc/chain.go                           // imports pkg/chain/types
pkg/coordinator/chain.go                    // imports pkg/chain/types
pkg/maintainer/wallet/chain.go              // imports pkg/chain/types
pkg/maintainer/spv/chain.go                 // imports pkg/chain/types
```

pkg/tbtc/chain.go
lukasz-zimnoch: Yeah I think we can remove it.

pkg/tbtc/chain.go
```diff
@@ -497,4 +451,6 @@ type Chain interface {
 	DistributedKeyGenerationChain
 	BridgeChain
 	WalletCoordinatorChain
+	ValidateDepositSweepProposalChain
+	ValidateRedemptionProposalChain
```
lukasz-zimnoch, line 455: So far, the structure of interfaces in the `chain.go` file reflected the contract structure we have on the chain. I think we should stick to that approach and avoid creating interfaces like `ValidateDepositSweepProposalChain`. If we go that way, we would soon have problems with defining the right boundaries between those interfaces. For example, `ValidateDepositSweepProposalChain` declares `GetPendingRedemptionRequest`. What if I need this function for a purpose other than proposal validation?

pkg/coordinator/chain.go
lukasz-zimnoch: As an alternative, we could also use a similar pattern that we use in `pkg/tecdsa`, i.e.:
```
pkg
    tbtc
        wallet           // holds the current contents of pkg/tbtc
        coordinator // holds the current contents of pkg/coordinator
        maintainer   // holds the current contents of pkg/maintainer
        chain.go      // holds interfaces and objects shared by the above packages
```

pdyraga: I think we intentionally decided to flatten packages compared to `pkg/tecdsa` and treat `pkg/tbtc` as the client-side wallet implementation. I don't think the alternative is bad but I also like the fact the packages are currently more flat compared to `pkg/tecdsa`.

lukasz-zimnoch: Yeah, that's true and I'm also strongly for the flat packages approach. I just think that if we want to extract the shared part somewhere, we should rather use this pattern as it fits more our domain-oriented package structure than the proposed `/pkg/chain/tbtc/types/types.go` which feels a bit artificial.

lukasz-zimnoch: Anyway, I think the best we can do for now is just keep the shared objects in `pkg/tbtc` and decouple the interfaces. This still makes `pkg/coordinator` dependent on `pkg/tbtc` but we achieve more specialized interfaces and shorter mocks. 

The dependency on `pkg/tbtc` objects is not the end of the world. Note that the existence of that dependency may indicate that the boundaries between our packages are not optimal but it's really hard to achieve the right shape on the first try and we should rather seek gradual improvements over time instead of making a revolution now 🤷

pdyraga: > Note that the existence of that dependency may indicate that the boundaries between our packages are not optimal but it's really hard to achieve the right shape on the first try and we should rather seek gradual improvements over time instead of making a revolution now 🤷

I wouldn't be too hard on ourselves. I like the current setup of our packages. It is really clean and separates boundaries well. `pkg/tbtc` is the core responsibility of tBTC client: being a wallet, responding to on-chain requests, and signing stuff. `pkg/coordinator` and `pkg/maintainer` could as well live outside of `keep-network/keep-core` but for a ton of other valid reasons we decided to keep them together in one binary. This is fine. The core of the logic takes the central seat in `pkg/tbtc` and all the auxiliary stuff sits out of the center of the room in case we ever need to move it outside.

Now, there are some common types both for the core logic and the auxiliary logic. `DepositChainRequest` is one of them. It is needed by the core wallet signing logic and the auxiliary maintainer logic. If `pkg/tbtc` is about the wallet signing logic, the auxiliary maintainer logic shouldn't import it. `pkg/tbtc/types` is a completely separate package from the Go perspective. 

I do agree it feels a bit artificial from the domain perspective but I think it does not feel artificial from the Go side and there are some examples of using this pattern in `go-ethereum` as well: [`core/types`](https://github.com/ethereum/go-ethereum/tree/d73eb87979021e63cb5867319c8ef07a9ba8a7d2/core/types) or [`beacon/types`](https://github.com/ethereum/go-ethereum/tree/d73eb87979021e63cb5867319c8ef07a9ba8a7d2/beacon/types).

My slight preference is on `pkg/tbtc/types` package but I could as well have the import to `pkg/tbtc`. I would avoid duplication given the problems in the implementation structure and I would avoid moving the `maintainer` and `coordinator` packages to `pkt/tbtc` because I really like the flat structure and the current way of thinking about the responsibilities of our packages.

pkg/tbtc/chain.go
pdyraga: > For example, `ValidateDepositSweepProposalChain` declares `GetPendingRedemptionRequest`. What if I need this function for a purpose other than proposal validation?

This is the case for #3649 currently in progress.

I do agree with this comment. I see it was done to avoid the duplication here:

https://github.com/keep-network/keep-core/blob/7eeea1f9e3a5aae0463244731e87cf5adf57b294/pkg/coordinator/chain.go#L112-L113

but I think having the duplication of function declarations is a better choice in this particular case. I am not a huge fan of importing types from `pkg/tbtc` as you already know from the other comment but importing interfaces makes it look even worse for me. 😉

pkg/coordinator/chain.go
lukasz-zimnoch: > If pkg/tbtc is about the wallet signing logic, the auxiliary maintainer logic shouldn't import it. pkg/tbtc/types is a completely separate package from the Go perspective.

Agreed but in fact, `pkg/tbtc` is a little bit broader than we assume here. It also contains some logic that is useful for auxiliary packages, for example `tbtc.ValidateDepositSweepProposal`. That means that even if we extract types to `pkg/tbtc/types`, we would still have the dependency due to some shared logic. This is why I'm leaning towards leaving shared types in `pkg/tbtc` and do not complicate the situation with an additional package.

pkg/maintainer/wallet/chain_test.go
```diff
@@ -0,0 +1,175 @@
...
+	walletAction   tbtc.WalletActionType
+}
+
+type localChain struct {
+	mutex sync.Mutex
+
+	walletLocks map[[20]byte]*walletLock
+}
+
+func newLocalChain() *localChain {
```
pdyraga, line 25: I realized we have a neat pattern we agreed on early in [`pkg/sortition`](https://github.com/keep-network/keep-core/tree/main/pkg/sortition) where we define the local chain implementation in `pkg/internal/local`. This way, `_test.go` is only for actual tests and we can unit-test the local chain implementation as well. I know we were not very consistent in using this pattern, but this refactor feels like a good opportunity to align at least this package.

cc @lukasz-zimnoch @tomaszslabon

lukasz-zimnoch: This pattern can cause import cycles. Suppose we implement `tbtc.Chain` in `pkg/tbtc/internal/local.go`, then the structure declared in this internal package must refer to `tbtc` chain types used by `tbtc.Chain`. On the other hand, the unit tests living in `tbtc` package must create an instance of `local.Chain`. Here is why this pattern is used only in `pkg/sorititon` where `sorition.Chain` does not declare any chain types.

pdyraga: ... and that's why I think we should have [the types declared in a separate package](https://github.com/keep-network/keep-core/pull/3650#discussion_r1242336911) 😁

pkg/coordinator/chain.go
lukasz-zimnoch: In order to elaborate more on why I'm a bit against `pkg/tbtc/types`:
- It's a bit too generic and does not give a clear answer about what should live inside and what not
- It is nested within `pkg/tbtc` which will make a wrong dependency direction: `pkg/tbtc` -> `pkg/tbtc/types`. In Go, we typically make sub-packages depending on the outer packages (a lot of examples in Go std lib but also ours `pkg/net`, `pkg/tecdsa`, and so on).

If we really insist on extracting, I would rather go with creating `pkg/tbtccore` and put there all core types and shared logic that does not belong to any specific domain context.

pkg/tbtc/chain.go
nkuba: Alternative approach: 70bd4497a02fd50ade14bf471b78d0365e990055

pkg/coordinator/chain.go
pdyraga: > * It is nested within `pkg/tbtc` which will make a wrong dependency direction: `pkg/tbtc` -> `pkg/tbtc/types`. In Go, we typically make sub-packages depending on the outer packages (a lot of examples in Go std lib but also ours `pkg/net`, `pkg/tecdsa`, and so on).

Yes, we would have to bend this rule. It is not ideal but we wouldn't be the only ones. For example, the `Block` structure is defined in `github.com/ethereum/go-ethereum/core/types` package and it is imported by `go-ethereum/core/blockchain_insert.go` [here](https://github.com/ethereum/go-ethereum/blob/d73eb87979021e63cb5867319c8ef07a9ba8a7d2/core/blockchain_insert.go#L24). This is exactly our use case. As long as we do it only for the `types` package, I would be personally fine with it. 

> * It's a bit too generic and does not give a clear answer about what should live inside and what not

It is less generic than `tbtccore` :) I see the reasoning behind the `tbtccore` but I am afraid this will sooner or later evolve into a common bag of stuff and create even more complicates mesh of dependencies. I think it's nothing wrong if `coordinator` or `maintainer` packages imports logic from `tbtc`. It is clear, they perform the same validation as the wallet as in the mentioned `tbtc.ValidateDepositSweepProposal` example and in my opinion, this is a really nice feature of the current code organization.

My slight preference is for a separate types package with good documentation of why it's there and what types should be there. This would give us more flexibility and avoid import cycles. But I am also fine with leaving it as-is with imports from `tbtc`.

pkg/maintainer/wallet/chain_test.go
nkuba: > I know we were not very consistent in using this pattern, but this refactor feels like a good opportunity to align at least this package.

IMHO if agree on a specific pattern we should refactor and apply it to the whole keep-core project. Currently, we have multiple patterns that we use randomly in different packages. When we create a new package we're not sure which pattern to follow, and we add even more inconsistencies.

pkg/coordinator/chain.go
nkuba: Unfortunately, it seems that there is no perfect solution.

If we decide to declare the types in a common types package what types should go there? All the types used by chain interfaces across all the packages? We have types that are common to `pkg/tbtc` and `pkg/coordinator` and specific to the `pkg/tbtc` only. Which do we want to move to the common types?

Separating types from the chain interfaces declaration may be confusing, as in V2 we decided to declare the chain interfaces closer to the actual usage in packages.

Importing types defined in `pkg/tbtc` package into the `pkg/coordinator` doesn't seem perfect either. The `pkg/coordinator` package uses `pkg/tbt` types, but also declares its' own types (e.g. `NewWalletRegisteredEvent`).

We cannot duplicate the types either, as the functions defined in different packages will require the same type, e.g. `ValidateRedemptionProposal` defined in both `pkg/coordinator` and `pkg/tbtc` must use the same `RedemptionProposal` type.

I lean towards the current approach where types are defined in `pkg/tbtc` which is considered a broader and core package and we just import them in `pkg/coordinator`. This isn't perfect but it doesn't require an aggressive refactor. 

We could add another issue for the future to find the perfect solution for types definition.

pkg/tbtc/chain.go
nkuba: da06660aefb76b9a0da07a385e1607bd74ff1dbe

pkg/maintainer/wallet/chain_test.go
nkuba: Since it doesn't look like we have a consistent idea of the approach we want to follow, and it seems out-of-scope of the current PR and https://github.com/keep-network/keep-core/issues/3632, I propose to handle a refactor of all local chain implementations in a separate PR (along with other changes like one mutex for the local chain you proposed recently).

pkg/tbtc/chain.go
pdyraga: This approach is fine for me but I'd like to hear from @lukasz-zimnoch if it does not break anything for the automation part.

pkg/coordinator/chain.go
pdyraga: Captured it here: https://github.com/keep-network/keep-core/issues/3652

pdyraga (approved):

pkg/tbtc/chain.go
lukasz-zimnoch: Looks good!

Issue #3614, opened 2023-06-07
Automate deposit and redemption WalletCoordinator actions


pdyraga:
The [`WalletCoordinator` contract](https://etherscan.io/address/0x64EA4b84e2BdfD313428b96658260E495a420093) is a synchronization point for the network clients. They observe events emitted from this contract and execute the given action previously performing all the required validations of the proposal.

Deposit sweep proposals are currently submitted manually by the dev team. This is possible because sweeps are not as time-sensitive as redemptions but still, we consider it a temporary measure until the proper automation is implemented. 

This issue aims at automating both deposit sweep proposals and redemption proposals in the `WalletCoordinator`. This code should be a part of the existing maintainer module in the client, activated/deactivated based on the config flag.

The logic should be simple for the start: check every 3 hours if there are any redemptions requested. If yes, check if no sweeping is in progress and request redemption of max N deposits. N should be adjustable in the `WalletCoordinator` contract and set based on the gas cost of submitting redemption SPV proof for the given N. 

For the start, sweeps should be proposed every 48 hours if there are no redemptions pending. Given we have optimistic minting, sweeps are not time-sensitive.

When working on this issue please ensure the approach is consistent with SPV proof maintainer module from #3615.

The solution should integrate with redemptions requests implementation from https://github.com/keep-network/keep-core/issues/3609.
```[tasklist]
### Pull Requests
- [ ] https://github.com/keep-network/keep-core/pull/3626
- [ ] https://github.com/keep-network/keep-core/pull/3666
- [ ] https://github.com/keep-network/keep-core/pull/3631
- [ ] https://github.com/keep-network/keep-core/pull/3635
- [ ] https://github.com/keep-network/keep-core/pull/3640
- [ ] https://github.com/keep-network/keep-core/pull/3651
- [ ] https://github.com/keep-network/keep-core/pull/3659
- [ ] https://github.com/keep-network/keep-core/pull/3660
- [ ] https://github.com/keep-network/keep-core/pull/3661
- [ ] https://github.com/keep-network/keep-core/pull/3663
- [ ] https://github.com/keep-network/keep-core/pull/3668
```

nkuba:
> For the start, sweeps should be proposed every 48 hours if there are no redemptions pending. Given we have optimistic minting, sweeps are not time-sensitive.

Would it be worth defining a max number of deposits to sweep in the `WalletCoordinator` contract?

PR #3651, opened 2023-06-27
Estimate redemption transaction fee during proposal

lukasz-zimnoch:
Refs https://github.com/keep-network/keep-core/issues/3614

Here we add a function allowing us to estimate the redemption transaction total fee (`coordinator.EstimateRedemptionFee`) and we integrate it with the existing `coordinator.ProposeRedemption` function.

### DIFF
--- a/pkg/coordinator/redemptions.go
+++ b/pkg/coordinator/redemptions.go
@@ -13,8 +13,7 @@ import (
 )
 
 type redemptionEntry struct {
-	walletPublicKeyHash [20]byte
-
+	walletPublicKeyHash  [20]byte
 	redemptionKey        string
 	redeemerOutputScript bitcoin.Script
 	requestedAt          time.Time
@@ -142,6 +141,7 @@ func FindPendingRedemptions(
 // ProposeRedemption handles redemption proposal submission.
 func ProposeRedemption(
 	chain Chain,
+	btcChain bitcoin.Chain,
 	walletPublicKeyHash [20]byte,
 	fee int64,
 	redeemersOutputScripts []bitcoin.Script,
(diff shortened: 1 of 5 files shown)


pkg/coordinator/redemptions.go
```diff
@@ -318,3 +334,43 @@ redemptionRequestedLoop:
...
+		case bitcoin.P2PKHScript:
+			sizeEstimator.AddPublicKeyHashOutputs(1, false)
+		case bitcoin.P2WPKHScript:
+			sizeEstimator.AddPublicKeyHashOutputs(1, true)
+		case bitcoin.P2SHScript:
+			sizeEstimator.AddScriptHashOutputs(1, false)
+		case bitcoin.P2WSHScript:
+			sizeEstimator.AddScriptHashOutputs(1, true)
+		default:
+			panic("non-standard redeemer output script type")
```
pdyraga, line 359: I am not sure if we should trust the contract validation that much. We can have an upgrade and change the on-chain logic. I think it is OK to assume the script has to be one of the four types listed before but if it is not I would rather return an error, log it, and decide not to propose a redemption instead of killing the process. 

At some point in the future

lukasz-zimnoch: The main purpose of this panic is to signal that something changed on the contract side and requires our attention. I'm afraid that returning an error and just logging it will not be visible enough. WDYT?

pdyraga: I think the client should panic only if there is no way out. Here there is a way out - we can decide not to propose a redemption and eventually, we can even skip the given script when proposing the redemption. Agreed about calling the attention but I would address it somewhere else, e.g. have a script type for every redemption request at tbtcscan.com

lukasz-zimnoch: See dc866e45

pdyraga (approved):

PR #3654, opened 2023-06-27
Modified retrieval of deposit request data

tomaszslabon:
This PR modifies the `GetDepositRequest` function by adding a boolean value to the returned values.
The returned values represent: 
- deposit request which is non-nil only when the deposit request was found,
- boolean value which is true if the deposit request was found, false otherwise,
- error which is non-nil only when the function execution failed. It will be nil if the deposit request was not found, but the function execution succeeded.

This can be useful when we want to check if a deposit exists, but we do not want to issue an error when it does not
exist, i.e. we don't consider it an error.

### DIFF
--- a/pkg/maintainer/spv/chain.go
+++ b/pkg/maintainer/spv/chain.go
@@ -20,8 +20,9 @@ type Chain interface {
 	) error
 
 	// GetDepositRequest gets the on-chain deposit request for the given
-	// funding transaction hash and output index. Returns an error if the
-	// deposit was not found.
+	// funding transaction hash and output index. If the deposit was not found
+	// a deposit object will still be returned, but the `RevealedAt` timestamp
+	// will be set to zero.
 	GetDepositRequest(
 		fundingTxHash bitcoin.Hash,
 		fundingOutputIndex uint32,
--- a/pkg/maintainer/spv/chain.go
+++ b/pkg/maintainer/spv/chain.go
@@ -20,13 +20,18 @@ type Chain interface {
 	) error
 
 	// GetDepositRequest gets the on-chain deposit request for the given
-	// funding transaction hash and output index. If the deposit was not found
-	// a deposit object will still be returned, but the `RevealedAt` timestamp
-	// will be set to zero.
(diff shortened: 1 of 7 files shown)


pkg/maintainer/spv/chain.go
```diff
@@ -20,8 +20,9 @@ type Chain interface {
 	) error
 
 	// GetDepositRequest gets the on-chain deposit request for the given
-	// funding transaction hash and output index. Returns an error if the
-	// deposit was not found.
+	// funding transaction hash and output index. If the deposit was not found
+	// a deposit object will still be returned, but the `RevealedAt` timestamp
+	// will be set to zero.
```
tomaszslabon, line 25: Alternatively, we could add an additional parameter `errorOnNotFound` to the `GetDepositRequest` function which when set to true would cause the function to return an error if deposit not found.

pdyraga: My preference would be to return three values from `GetDepositRequest`: 
- `*tbtc.DepositChainRequest` that would be non-nil only when the given request exists,
- `bool` that would be true if the deposit exists, `false` otherwise,
- `error` that would be non-nil only in case of connectivity-related issues and such.

I could also see the proposition with `errorOnNotFound` working but my preference is for the `bool` version. I personally don't like having a conditional flow based on errors if those errors may represent normal situations and a valid system state.

I would avoid returning a non-nil pointer with empty fields. In my opinion, this is misleading because the default value for a pointer is `nil`. 

Also, please note we have constructions like this and we need to clean it up as a part of the work on this PR:

https://github.com/keep-network/keep-core/blob/e8a572023c58c6f16ec7a97b1c60f7e964462594/pkg/coordinator/redemptions.go#L248

tomaszslabon: Done in 2b827ca2010e1daea0bc0541d3b725977945b08a.
The piece of code you posted deals with redemption requests and not deposit requests:
```
case errors.Is(err, tbtc.ErrPendingRedemptionRequestNotFound): 
```
This PR should not affect it.

pdyraga (approved):

Issue #3586, opened 2023-05-23
Electrum integration tests hang for 10 minutes in case wrong host is provided


lukasz-zimnoch:
Observed in:
- https://github.com/keep-network/keep-core/pull/3581#discussion_r1202195003
- https://github.com/keep-network/keep-core/pull/3578#discussion_r1200310615

Steps to recreate:
1. Go to `pkg/bitcoin/electrum/electrum_integration_test.go`
2. Replace "blackie.c3-soft.com:57006" at line 72 with something that can't be resolved, e.g. "blackie.c3-softtt.com:57006"
3. Run `go test github.com/keep-network/keep-core/pkg/bitcoin/electrum -v -tags=integration ./...` 

In some cases, one of the integration tests hangs for 10 minutes and fail with:
```
panic: test timed out after 10m0s

goroutine 429 [running]:
testing.(*M).startAlarm.func1()
        /opt/homebrew/Cellar/go@1.18/1.18.5/libexec/src/testing/testing.go:2029 +0x8c
created by time.goFunc
        /opt/homebrew/Cellar/go@1.18/1.18.5/libexec/src/time/sleep.go:176 +0x3c

goroutine 1 [chan receive, 2 minutes]:
testing.(*T).Run(0x14000555520, {0x102e1e49b?, 0x1c114841bbb5b?}, 0x103160588)
```
while in this case, all tests should always fail with:
```
failed to initialize electrum client: [retry timeout [1m0s] exceeded; most recent error: [dial tcp: lookup blackie.c3-softfff.com: no such host]]
```

The above problem may indicate some problems with the Electrum implementation connection retry logic.
```[tasklist]
### Tasks
- [ ] https://github.com/keep-network/keep-core/pull/3599
```

nkuba:
Go tests are executed with the default timeout of 10m. If the suite doesn't finish execution within this time the tests are aborted with panic.

In case of electrum connection problems for any of the servers, the executor has 1m for retries before it fails the particular test. If one of the servers doesn't work, given the number of tests the suite defines the execution won't be completed within 10 minutes timeout. So we see the panics.

I'll increase the timeout for integration tests to not panic if any of the servers is non-responsive.
I'll add parallel test execution to reduce the time required for the whole suite to execute.

nkuba:
Actually, we already increased the timeout for the integration test to 20m in https://github.com/keep-network/keep-core/pull/3599.

[...]

PR #3656, opened 2023-06-28
Run electrum integration tests in parallel

nkuba:
To reduce the time required for the electrum integration tests execution we run the tests in parallel.

Refs: https://github.com/keep-network/keep-core/issues/3586


pdyraga:
Tested locally with `go test -v -tags=integration ./pkg/bitcoin/electrum -count=1`. The tests are executing about ~110s now and they were executing ~470s before.

pdyraga (approved):

PR #3657, opened 2023-06-28
Split contracts API doc into multiple files

michalinacienciala:
Having all the contracts documented in one common file turned out to be hard to render by the GitBook. We're switching to documenting each contract file in a separate Markdown file.
As the generated files will be much smaller now and GitBook has its own file `ON THIS PAGE` section, we don't need to generate Table of Contents. The one thing that we also change as part of this PR is `rsyncDelete` setting - we'll have it set to `true`, in order to delete documentation of contracts which get removed from the repo. As we're not working directly on a `main` branch, this isn't dangerous (we'll see all the deletions in the PR diff).

Refs:
https://github.com/threshold-network/solidity-contracts/pull/149
https://github.com/keep-network/tbtc-v2/pull/640
https://github.com/threshold-network/threshold/pull/36

### DIFF
--- a/.github/workflows/contracts-ecdsa-docs.yml
+++ b/.github/workflows/contracts-ecdsa-docs.yml
@@ -46,6 +46,7 @@ jobs:
     with:
       projectDir: /solidity/ecdsa
       publish: false
+      addTOC: false
       commentPR: true
       exportAsGHArtifacts: true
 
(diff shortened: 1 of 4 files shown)


michalinacienciala:
Docs updated in https://github.com/threshold-network/threshold/pull/36.

pdyraga (approved):

PR #3658, opened 2023-06-28
Block header structure update

tomaszslabon:
This PR adds new functions to the `BlockHeader` structure:
- `Deserialize`
- `Target`
- `Difficulty`

These functions will be needed to verify if the difficulties of block headers in the SPV proof are consistent with the difficulties stored on-chain. They will be useful to detect a situation when the relay has not been updated yet with the current Bitcoin difficulties, but the proof already have block headers with those current difficulties.

### DIFF
--- a/pkg/bitcoin/block.go
+++ b/pkg/bitcoin/block.go
@@ -3,6 +3,8 @@ package bitcoin
 import (
 	"encoding/binary"
 	"math/big"
+
+	"github.com/btcsuite/btcd/blockchain"
 )
 
 // BlockHeaderByteLength is the byte length of a serialized block header.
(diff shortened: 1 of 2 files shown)


pdyraga:
Can you please update the PR description to explain why do we need the new functions?

pkg/bitcoin/block.go
```diff
@@ -59,6 +62,39 @@ func (bh *BlockHeader) Serialize() [BlockHeaderByteLength]byte {
...
+	// PreviousBlockHeaderHash
+	copy(
+		bh.PreviousBlockHeaderHash[:],
+		rawBlockHeader[offset:offset+HashByteLength],
+	)
+	offset += HashByteLength
+
+	// MerkleRootHash
+	copy(bh.MerkleRootHash[:], rawBlockHeader[offset:offset+HashByteLength])
+	offset += HashByteLength
```
pdyraga, line 84: Would it make sense to align the `Serialize` function to do `offset += HashByteLength` instead of `offset += len(bh.PreviousBlockHeaderHash)` and `offset += len(bh.MerkleRootHash)`? 

I like the consistency and if we are not sure that `HashByteLength` is future-proof then we need to think about modifying the `Serialize` function.

pkg/bitcoin/block.go
```diff
@@ -69,3 +105,70 @@ func (bh *BlockHeader) Hash() Hash {
 	//       3. Construct the Hash instance appropriately.
 	return Hash{}
 }
+
+// Target calculates the difficulty target of a block header.
+func (bh *BlockHeader) Target() *big.Int {
+	// A serialized 80-byte block header stores the `bits` value as a 4-byte
+	// little-endian hexadecimal value in a slot including bytes 73, 74, 75, and
```
pdyraga, line 112: Why *hexadecimal*? If something is stored as bytes, it's just bytes.

pkg/bitcoin/block.go
```diff
@@ -69,3 +105,70 @@ func (bh *BlockHeader) Hash() Hash {
...
+	// 76. This function's input argument is expected to be a numerical
```
pdyraga, line 113: Which function's ?

pkg/bitcoin/block.go
```diff
@@ -69,3 +105,70 @@ func (bh *BlockHeader) Hash() Hash {
...
+	// - 0x1b0404cb >>> 24 = 0x0000001b
+	// - 0x0000001b & 0xff = 0x1b
+	// - 0x1b - 3 = 24 (decimal)
+	//
+	// To extract the mantissa, we just need to take the last three bytes:
+	// - 0x1b0404cb & 0xffffff = 0x0404cb = 263371 (decimal)
+	//
+	// The final difficulty can be computed as mantissa * 256^exponent:
+	// - 263371 * 256^24 =
+	// 1653206561150525499452195696179626311675293455763937233695932416 (decimal)
```
pdyraga, line 139: In `bitcoin.go` we have `readCompactSizeUint` that uses `wire.ReadVarInt` to parse bytes into `CompactSizeUint`. If we already rely on `btcsuite/btcd` for parsing, we could eliminate most of the complexity from this function and in the function body call [`difficulty.CompactToBig`](https://github.com/btcsuite/btcd/blob/f9cbff0d819c951d20b85714cf34d7f7cc0a44b7/blockchain/difficulty.go#L62).

The same comment applies to `Difficulty()` function. In the body of that function we could call [`difficulty.CalcWork`](https://github.com/btcsuite/btcd/blob/f9cbff0d819c951d20b85714cf34d7f7cc0a44b7/blockchain/difficulty.go#L142).

I would avoid creating a separate type for this "compact" representation. We already have `type CompactSizeUint uint64` in the `bitcoin` package representing [Bitcoin CompactSize unsigned integers](https://developer.bitcoin.org/reference/transactions.html#compactsize-unsigned-integers). `CompactSizeUint` is not the same as the "compact" format of storing [the target nBits](https://developer.bitcoin.org/reference/block_chain.html#target-nbits). It makes sense to have `Target()` and `Difficulty()` functions as we have now and do not allow anything about the other "compact" representation to leak out of them.

Unit tests should stay as they are now, we should have the base coverage even if we call the library function.

pkg/bitcoin/block.go
tomaszslabon: It makes sense. Done in bfea8c6cebe61384628c03aef341e03843b49cae.

pkg/bitcoin/block.go
tomaszslabon: Since this function was updated and uses the `btcsuite` library now I removed this detailed explanation of how the 
target is calculated. I only added a simple and short description for this function.

pkg/bitcoin/block.go
tomaszslabon: I replaced this lengthy explanation of how target is calculated with a short function description.
We relay on the `btcsuite` library now.

pkg/bitcoin/block.go
tomaszslabon: I used the `CompactToBig()` function for calculating the target.
However, we cannot use `CalcWork()` to calculate difficulty. We need to calculate difficulty the difficulty the same way 
we do in the [Solidity library](https://github.com/keep-network/bitcoin-spv/blob/8f4c751cc1fdd3e1f0c7ad40f6e4d2024a65d19f/contracts/BTCUtils.sol#L339).

tomaszslabon:
> Can you please update the PR description to explain why do we need the new functions?

I updated the PR description and explained why we are adding these functions.

pdyraga (approved):

PR #3660, opened 2023-06-28
Invoke automated redemptions periodically by maintainer

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3614
Depends on: https://github.com/keep-network/keep-core/pull/3659

Here we add automation for the redemption proposal submission logic. Every 3 hours, the maintainer will check for pending redemptions starting from the oldest wallet and will propose them for redemption by submitting a proposal to the `WalletCoordinator` contract.

### DIFF
--- a/pkg/maintainer/wallet/redemptions.go
+++ b/pkg/maintainer/wallet/redemptions.go
@@ -270,27 +270,55 @@ func getPendingRedemptions(
 	// TODO: We should consider narrowing the block range in filter the fetch
 	//       only events that are newer than `now - redemptionRequestTimeout`.
 
-	redemptionsRequested, err := chain.PastRedemptionRequestedEvents(filter)
+	events, err := chain.PastRedemptionRequestedEvents(filter)
 	if err != nil {
 		return nil, fmt.Errorf(
 			"failed to get past redemption requested events: [%w]",
 			err,
 		)
 	}
-	logger.Infof("found %d RedemptionRequested events", len(redemptionsRequested))
 
 	// Take the oldest first.
-	sort.SliceStable(redemptionsRequested, func(i, j int) bool {
-		return redemptionsRequested[i].BlockNumber < redemptionsRequested[j].BlockNumber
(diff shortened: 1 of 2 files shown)


pkg/maintainer/wallet/redemptions.go
```diff
@@ -11,6 +12,43 @@ import (
 	"time"
 )
 
+func (wm *walletMaintainer) runRedemptionTask(ctx context.Context) error {
+	redemptionMaxSize, err := wm.chain.GetRedemptionMaxSize()
+	if err != nil {
+		return fmt.Errorf("failed to get deposit sweep max size: [%w]", err)
```
pdyraga, line 18:
```suggestion
		return fmt.Errorf("failed to get redemption max size: [%w]", err)
```

pkg/maintainer/wallet/redemptions.go
```diff
@@ -11,6 +12,43 @@ import (
...
+	}
+
+	walletPublicKeyHash, redeemersOutputScripts, err := FindPendingRedemptions(
+		wm.chain,
+		[20]byte{},
+		redemptionMaxSize,
+	)
```
pdyraga, line 25: `FindPendingRedemptions` in the current form stops searching when it finds a wallet with a pending redemption. I don't think this is enough. A malicious redeemer can easily make redemption requests to other wallets time out. I think we should - at a minimum - introduce a parameter that will limit the number of wallets that can process redemptions at the same time and not stop searching in `FindPendingRedemptions` until we reach the limit.

Maybe it's already on your radar for this PR, just making sure.

lukasz-zimnoch: Yep, this is captured in the task list of https://github.com/keep-network/keep-core/issues/3614

pkg/maintainer/wallet/redemptions.go
lukasz-zimnoch: See 14bca05a

pdyraga (approved):

PR #3663, opened 2023-07-03
Maintainer optimizations around redemptions

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3614

Here we introduce a bunch of optimizations to the wallet maintainer. All of them are related to redemptions. I recommend reviewing commit by commit. Here is a brief description of the changes.

### Improve past `RedemptionRequested` events filtering

The wallet maintainer seeks past `RedemptionRequested` events in order to propose redemptions. So far, the events lookup was not constrained to any block range which could be problematic for several Ethereum providers. This can be improved as the wallet maintainer is only interested in pending requests that are not timed out yet. That means there is no sense to fetch events that were emitted before `now - redemptionRequestTimeout` block. We reflect that by narrowing the block range of the filter thus making the call more lightweight.

### Submit multiple redemption proposals for different wallets

Redemptions are time-sensitive so we must make sure all of them are handled in a reasonable time. So far, the maintainer iterated from the oldest wallet and proposed redemptions for the first wallet having pending redemptions. We can do it better and propose redemptions for multiple wallets at the same time. However, given the fact signing is computationally heavy and a lot of operator nodes have shares in multiple wallets, we need to set a reasonable limit on the maximum number of wallets performing redemptions at the same time. We achieve that by introducing the `RedemptionWalletsLimit` configurable parameter.

### Add an ability to limit the proposed redemptions value

We want to limit the risk for initial mainnet redemptions done shortly after the redemptions release. To achieve that, we need a way to limit the total amount of a single redemption proposal to a reasonable value. We address that indirectly, by introducing the `RedemptionRequestAmountLimit` for the wallet maintainer. This parameter guarantees that the redemption proposals submitted by the maintainer will consist of redemption requests whose amounts are below a certain satoshi value.
Combined with `WalletCoordinator.redemptionMaxSize` on-chain parameter, we obtain a tool to cap the total value of submitted redemption proposals to safe levels. All redemption requests exceeding the `RedemptionRequestAmountLimit` can be handled manually through the `maintainer-cli` tool after confirming everything is all right.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -342,6 +342,13 @@ func initMaintainerFlags(command *cobra.Command, cfg *config.Config) {
 		"The time interval in which pending redemptions requests are checked.",
 	)
 
+	command.Flags().Uint16Var(
+		&cfg.Maintainer.WalletCoordination.RedemptionWalletsLimit,
+		"walletCoordination.redemptionWalletsLimit",
+		wallet.DefaultRedemptionWalletsLimit,
+		"Limits the number of wallets that can receive a redemption proposal in the same time",
+	)
+
 	command.Flags().DurationVar(
 		&cfg.Maintainer.WalletCoordination.DepositSweepInterval,
 		"walletCoordination.depositSweepInterval",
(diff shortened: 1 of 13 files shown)


pdyraga (dismissed):

pdyraga:
The code looks good to me but there are merge conflicts that need to be resolved before I am able to merge @lukasz-zimnoch

pdyraga (approved):

PR #3668, opened 2023-07-04
Unit tests for wallet maintainer redemptions

lukasz-zimnoch:
Closes: https://github.com/keep-network/keep-core/issues/3614

Here we add missing unit tests for the wallet maintainer redemption code.

### DIFF
--- a/pkg/maintainer/wallet/redemptions_test.go
+++ b/pkg/maintainer/wallet/redemptions_test.go
@@ -2,9 +2,12 @@ package wallet_test
 
 import (
 	"encoding/hex"
+	"github.com/go-test/deep"
 	"github.com/keep-network/keep-core/internal/testutils"
 	"github.com/keep-network/keep-core/pkg/bitcoin"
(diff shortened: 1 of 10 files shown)


pkg/maintainer/wallet/redemptions_test.go
```diff
@@ -37,3 +41,185 @@ func TestEstimateRedemptionFee(t *testing.T) {
...
+			},
+		},
+		"fee estimated": {
+			fee: 0, // trigger fee estimation
+			expectedProposal: &tbtc.RedemptionProposal{
+				WalletPublicKeyHash:    walletPublicKeyHash,
+				RedeemersOutputScripts: redeemersOutputScripts,
+				RedemptionTxFee:        big.NewInt(4300),
+			},
+		},
```
pdyraga, line 178: We could also add a test case proving the proposal will not pass if the validation failed.

pdyraga (approved):

PR #3669, opened 2023-07-05
Prepare SPV maintainer for redemptions

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3615

The current code of SPV maintainer is strongly tied to deposit sweeps. In order to implement redemption proofs in an efficient and clean way, we need to make them more generic and extract common parts that can be reused by different proof types in order to avoid duplication. Here we aim to achieve this. I recommend reviewing commit by commit to understand all changes being part of this refactoring.

PR #3670, opened 2023-07-05
Support for redemption SPV proofs

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3615
Depends on: https://github.com/keep-network/keep-core/pull/3669

This changeset adds redemption support to the SPV maintainer.

### DIFF
--- a/pkg/maintainer/spv/redemptions.go
+++ b/pkg/maintainer/spv/redemptions.go
@@ -1,0 +1,29 @@
+package spv
+
+import "github.com/keep-network/keep-core/pkg/bitcoin"
+
+// SubmitRedemptionProof prepares redemption proof for the given transaction
+// and submits it to the on-chain contract. If the number of required
+// confirmations is `0`, an error is returned.
+//
+// TODO: Expose this function through the maintainer-cli tool.
+func SubmitRedemptionProof(
+	transactionHash bitcoin.Hash,
+	requiredConfirmations uint,
+	btcChain bitcoin.Chain,
+	spvChain Chain,
+) error {
+	panic("not implemented yet")
+}
+
+func getUnprovenRedemptionTransactions(
+	historyDepth uint64,
+	transactionLimit int,
+	btcChain bitcoin.Chain,
+	spvChain Chain,
+) (
+	[]*bitcoin.Transaction,
+	error,
+) {
+	panic("not implemented yet")
+}
--- a/pkg/maintainer/spv/redemptions.go
+++ b/pkg/maintainer/spv/redemptions.go
@@ -2,7 +2,6 @@ package spv
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"github.com/keep-network/keep-core/pkg/bitcoin"
 	"github.com/keep-network/keep-core/pkg/tbtc"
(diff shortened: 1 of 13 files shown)


pkg/maintainer/spv/redemptions.go
```diff
@@ -0,0 +1,312 @@
...
+			} else {
+				return false, fmt.Errorf(
+					"failed to get pending redemption request: [%w]",
+					err,
+				)
+			}
+		}
+	}
+
+	return true, nil
```
tomaszslabon, line 294: I would add a check that at least one output represented a redemption, i.e. the change output was not the only output. We check this condition in the Bridge.

pkg/maintainer/spv/redemptions.go
```diff
@@ -0,0 +1,312 @@
...
+		// request.
+		_, err := spvChain.GetPendingRedemptionRequest(
+			walletPublicKeyHash,
+			output.PublicKeyScript,
+		)
+		if err != nil {
+			if errors.Is(err, tbtc.ErrPendingRedemptionRequestNotFound) {
+				// This output is neither a change nor a pending request.
+				// That means this is not a redemption transaction.
+				return false, nil
```
tomaszslabon, line 284: Wouldn't it be better to refactor the `GetPendingRedemptionRequest` in the similar manner as we did with GetDepositRequest:
```
	// GetDepositRequest gets the on-chain deposit request for the given
	// funding transaction hash and output index.The returned values represent:
	// - deposit request which is non-nil only when the deposit request was
	//   found,
	// - boolean value which is true if the deposit request was found, false
	//   otherwise,
	// - error which is non-nil only when the function execution failed. It will
	//   be nil if the deposit request was not found, but the function execution
	//   succeeded.
	GetDepositRequest(
		fundingTxHash bitcoin.Hash,
		fundingOutputIndex uint32,
	) (*tbtc.DepositChainRequest, bool, error)
```
If we added the boolean value indicating whether or not the pending redemption was found, we wouldn't have to
rely on error comparison. The fact that the pending redemption was not found is not really an error.

pkg/maintainer/spv/redemptions.go
```diff
@@ -0,0 +1,312 @@
...
+			}
+
+			if isChange {
+				changeFound = true
+				continue
+			}
+		}
+
+		// If the given output is not a change, it must be a pending redemption
+		// request.
```
tomaszslabon, line 275: According to [some checks in Bridge](https://github.com/keep-network/tbtc-v2/blob/716bf066c2cae0ccf8fe16b71ac43dac2bb7b708/solidity/contracts/bridge/Redemption.sol#L925) the output of a redemption transaction can also be a timed-out redemption.
If the wallet made a redemption transaction handling such timed-out redemption requests we should probably handle it.

pkg/maintainer/spv/redemptions.go
lukasz-zimnoch: Good point! See c372c0ce

pkg/maintainer/spv/redemptions.go
lukasz-zimnoch: I think we should not submit such proofs automatically. This is an abnormal situation that should not happen and this condition lives in the Bridge just in case. It will be safer if the maintainer fails and a human can assess the situation and submit the proof manually if needed.

tomaszslabon (approved):

PR #3675, opened 2023-07-07
Unit tests for SPV redemption maintainer logic

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3615

This PR adds unit tests to the SPV redemption maintainer logic.

### DIFF
--- a/pkg/maintainer/spv/bitcoin_chain_test.go
+++ b/pkg/maintainer/spv/bitcoin_chain_test.go
@@ -1,25 +1,38 @@
 package spv
 
 import (
+	"bytes"
+	"fmt"
 	"github.com/keep-network/keep-core/pkg/bitcoin"
 	"sync"
 )
 
-//lint:ignore U1000 Ignore unused type temporarily.
 type localBitcoinChain struct {
 	mutex sync.Mutex
+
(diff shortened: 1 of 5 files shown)


pkg/maintainer/spv/bitcoin_chain_test.go
```diff
@@ -1,25 +1,38 @@
...
+	"fmt"
 	"github.com/keep-network/keep-core/pkg/bitcoin"
 	"sync"
 )
 
-//lint:ignore U1000 Ignore unused type temporarily.
 type localBitcoinChain struct {
 	mutex sync.Mutex
+
+	transactions []*bitcoin.Transaction
```
tomaszslabon, line 13: I would consider storing transactions in a map, just like we did in the [bitcoin](https://github.com/keep-network/keep-core/blob/db145b629c3b4ec2677bc4457478439c685a2eff/pkg/bitcoin/chain_test.go#L10).
The reason for this is that some tests will require adding transactions and using a map for this ensures we will never add the same transaction twice.

lukasz-zimnoch: `BroadcastTransaction` prevents against adding the same transaction twice.

PR #3678, opened 2023-07-10
Use non-privileged ports for ElectrumX

nkuba:
### Use non-privileged ports for ElectrumX

The TCP/IP port numbers below 1024 are considered privileged ports. Normal users and processes are not allowed to use them for various security reasons.

We are running our container with non-root and after recent upgrades in the GCP cluster (Jul 7th), the ElectrumX instances stopped working due to failures in binding TCP and SSL ports:
```
ERROR:SessionManager:TCP server failed to listen on all_interfaces:80: [Errno 13] error while attempting to bind on address ('0.0.0.0', 80): permission denied
ERROR:SessionManager:SSL server failed to listen on all_interfaces:443: [Errno 13] error while attempting to bind on address ('0.0.0.0', 443): permission denied
INFO:SessionManager:WS server listening on all_interfaces:8080
INFO:SessionManager:WSS server listening on all_interfaces:8443
```

We updated the ports to use the defaults described in ElectrumX documentation: `50001-50004`.

The service is still exposed on ports `80`, `443`, `8080`, `8443`.

### Define podManagementPolicy: Parallel
    
The default podManagementPolicy is `OrderedReady`, which waits with spinning up subsequent replicas until the previous replica is ready.
In our use case, it's inaccurate, as we want to have three independent replicas. If the first one is broken, we still want to have a chance to
have another working.
    
Read more: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies

---

Changes were already introduced to the clusters.

### DIFF
--- a/infrastructure/kube/templates/bitcoin/electrumx/electrumx-statefulset.yaml
+++ b/infrastructure/kube/templates/bitcoin/electrumx/electrumx-statefulset.yaml
@@ -6,6 +6,7 @@ metadata:
 spec:
   replicas: 1
   serviceName: electrumx
+  podManagementPolicy: Parallel
   template:
     spec:
       securityContext:
--- a/infrastructure/kube/templates/bitcoin/electrumx/electrumx-statefulset.yaml
+++ b/infrastructure/kube/templates/bitcoin/electrumx/electrumx-statefulset.yaml
@@ -48,7 +48,7 @@ spec:
             - name: SSL_KEYFILE
               value: /mnt/electrum/cert/tls.key
             - name: SERVICES
-              value: tcp://:80,ssl://:443,ws://:8080,wss://:8443,rpc://0.0.0.0:8000
+              value: tcp://:50001,ssl://:50002,ws://:50003,wss://:50004,rpc://0.0.0.0:8000
             - name: COST_SOFT_LIMIT
               value: "0"
             - name: COST_HARD_LIMIT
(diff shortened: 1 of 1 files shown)


infrastructure/kube/templates/bitcoin/electrumx/electrumx-statefulset.yaml
```diff
@@ -48,7 +49,7 @@ spec:
             - name: SSL_KEYFILE
               value: /mnt/electrum/cert/tls.key
             - name: SERVICES
-              value: tcp://:80,ssl://:443,ws://:8080,wss://:8443,rpc://0.0.0.0:8000
+              value: tcp://:50001,ssl://:50002,ws://:50003,wss://:50004,rpc://0.0.0.0:8000
```
pdyraga, line 52: Do we need to update the readiness probe config? Is it something that requires yaml file update?

nkuba: No, we don't have to update them. We defined names for the ports:
```yaml
          ports:
            - name: tcp
              containerPort: 50001
```
So we can reference the ports by name:
```yaml
          readinessProbe:
            tcpSocket:
              port: tcp
```
Same we do in the [service (`targetPort`)](https://github.com/keep-network/keep-core/blob/11485daad086fa88fe9cc4b2c12c2aaa6a1b96d7/infrastructure/kube/templates/bitcoin/electrumx/electrumx-service.yaml#L13-L24):
```yaml
    - name: tcp
      port: 80
      targetPort: tcp
```

pdyraga: Nice!

pdyraga (approved):

PR #3680, opened 2023-07-11
Return accumulated confirmations

tomaszslabon:
#Depends on https://github.com/keep-network/keep-core/pull/3673.
This is a follow-up PR that handles a [review comment](https://github.com/keep-network/keep-core/pull/3667#discussion_r1253063238) from another PR.
Getting the number of accumulated confirmations of a transaction was moved into the `getProofInfo` function.

### DIFF
--- a/pkg/maintainer/spv/spv.go
+++ b/pkg/maintainer/spv/spv.go
@@ -421,7 +421,7 @@ func getProofInfo(
 	// The other case could be that the transaction is older than the last two
 	// Bitcoin difficulty epochs. In that case the transaction will soon leave
 	// the sliding window of recent transactions.
-	return false, accumulatedConfirmations, 0, nil
+	return false, 0, 0, nil
 }
 
 // walletEvent is a type constraint representing wallet-related chain events.
--- a/pkg/maintainer/spv/spv.go
+++ b/pkg/maintainer/spv/spv.go
@@ -5,9 +5,10 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
-	"github.com/keep-network/keep-core/pkg/tbtc"
 	"math/big"
 	"time"
+
+	"github.com/keep-network/keep-core/pkg/tbtc"
 
 	"github.com/ipfs/go-log/v2"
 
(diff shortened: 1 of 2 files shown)


pdyraga:
Does it close #3615 ?

tomaszslabon:
> Does it close #3615 ?

I believe so. Unless we have something else to do, like updating deployments?

pkg/maintainer/spv/spv.go
```diff
@@ -165,17 +166,10 @@ func (sm *spvMaintainer) proveTransactions(
...
-		)
-		if err != nil {
-			return fmt.Errorf(
-				"failed to get transaction confirmations: [%v]",
-				err,
-			)
-		}
-
-		isProofWithinRelayRange, requiredConfirmations, err := getProofInfo(
+		isProofWithinRelayRange,
```
lukasz-zimnoch, line 169: Can we format this call like this:
```
isProofWithinRelayRange, accumulatedConfirmations, requiredConfirmations, err := getProofInfo(
    transaction.Hash(),
    sm.btcChain,
    sm.spvChain,
    sm.btcDiffChain,
)
```
Even if it exceeds the maximum line length it should be more readable than the current form. WDYT?

pkg/maintainer/spv/spv.go
```diff
@@ -430,7 +424,7 @@ func getProofInfo(
 	// The other case could be that the transaction is older than the last two
 	// Bitcoin difficulty epochs. In that case the transaction will soon leave
 	// the sliding window of recent transactions.
-	return false, 0, nil
+	return false, accumulatedConfirmations, 0, nil
```
lukasz-zimnoch, line 427: Shouldn't we return `0` instead of `accumulatedConfirmations` in this case?

tomaszslabon: I also thought about, but in my opinion if we ended up here, it's not really an error.
It means that the transaction SPV proof goes into an epoch that is not proven in the relay yet.
I'm ok with returning `0`, though. It will not affect the flow.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in e982bd06521a6293bde6ae37ddf100836abd7b00.

pkg/maintainer/spv/spv.go
tomaszslabon: Done in 31c0ce911678b2254f63e7d1ea00b345e0d25a2d.

lukasz-zimnoch (approved):

PR #3690, opened 2023-07-19
Remove deprecation notice for --mainnet flag

nkuba:
We planned to remove the `--mainnet` flag, as it seemed unnecessary, as the client running without any network flag runs for mainnet by default. But from an infrastructure administrator perspective, the flag may be actually useful when running the application on multiple environments. It's much easier to maintain a common template where the administrator will have to change the value of the flag rather than set the flag for some environments and remove it for mainnet.

When running the application for **mainnet** an administrator will have two possibilities:
- add `--mainnet` flag,
- don't define any network flag.

When running the application for **testnet** an administrator will have to add `--testnet` flag.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -77,12 +77,10 @@ func initGlobalConfigFlags(cmd *cobra.Command, configFilePath *string) {
 // start command to run the client against test networks. Only one flag
 // from this set is allowed.
 func initGlobalNetworkFlags(cmd *cobra.Command) {
-	// TODO: Remove the mainnet flag.
-	// DEPRECATED
 	cmd.PersistentFlags().Bool(
 		network.Mainnet.String(),
 		false,
-		"Mainnet network (DEPRECATED)",
+		"Mainnet network",
 	)
 
 	cmd.PersistentFlags().Bool(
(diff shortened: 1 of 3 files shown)


cmd/flags.go
```diff
@@ -77,12 +77,10 @@ func initGlobalConfigFlags(cmd *cobra.Command, configFilePath *string) {
 // start command to run the client against test networks. Only one flag
 // from this set is allowed.
 func initGlobalNetworkFlags(cmd *cobra.Command) {
-	// TODO: Remove the mainnet flag.
-	// DEPRECATED
 	cmd.PersistentFlags().Bool(
 		network.Mainnet.String(),
 		false,
-		"Mainnet network (DEPRECATED)",
+		"Mainnet network",
```
dimpar, line 83: On line 99 we have a comment about removing the mainnet flag. Can you pls also remove it?

dimpar:
Left just one minor comment. Looks good to me.

cmd/flags.go
lukasz-zimnoch: See https://github.com/keep-network/keep-core/pull/3690/commits/ca8bdb802d7f2e04ddafa1047de60d28b6776dea

dimpar (approved):

PR #3700, opened 2023-08-21
Set unmarshallers early during signing

tomaszslabon:
#Refs https://github.com/keep-network/keep-core/issues/3611.
This PR handles problems with unmarshallers not found errors during signing:
```
eep-libp2p	libp2p/channel.go:286	couldn't find unmarshaler for type [protocol_announcer/announcement_message]
keep-libp2p	libp2p/channel.go:286	couldn't find unmarshaler for type [tbtc/signing_done_message]
```
The solution is is to register the marshaller for those messages early, before the signing process begins.

### DIFF
--- a/pkg/protocol/announcer/announcer.go
+++ b/pkg/protocol/announcer/announcer.go
@@ -71,6 +71,12 @@ type Announcer struct {
 	membershipValidator *group.MembershipValidator
 }
 
+func RegisterUnmarshaller(channel net.BroadcastChannel) {
+	channel.SetUnmarshaler(func() net.TaggedUnmarshaler {
+		return &announcementMessage{}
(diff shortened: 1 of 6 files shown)


dimpar:
I run the clients just fine, but I don't think I can hit the early signing stage before the sweep proposal on my local. System tests passed. Code looks good to me as well. Leaving the final approval to @lukasz-zimnoch or @pdyraga since they are more familiar with the flow.

dimpar (approved):

dimpar:
It was decided that the full flow will be validated on the testnet with the keep clients that have this change. Otherwise, it is very tricky to hit the early signing stage before the sweep on the local env.

Issue #3699, opened 2023-08-18
Silent panic - concurrent write to websocket


thevops:
Hello,

I had a problem with my tBTC node. It failed silently but didn't stop completely. Peers/bootstraps connections worked, but Prometheus monitoring (https://monitoring.threshold.network/grafana/public-dashboards/1a09fa3a621c4837988b36f2d6ae6e24?orgId=0) lost him for some time (until I noticed it by viewing the above Prometheus).
I also noticed that endpoint `/metrics` worked after the failure, but `/diagnostics` did not (infinite loading).
In that case, the app should exit with an error code, then my system would start it again (I use containers).

I had to manually restart the node to be discovered by the monitoring to be qualified for rewards.

### Logs

```
2023-08-17 03:21:21.933 {"level":"info","ts":"2023-08-17T01:21:21.933Z","logger":"keep-libp2p","caller":"libp2p/libp2p.go:242","msg":"number of connected peers: [144]"}
2023-08-17 03:20:21.934 {"level":"info","ts":"2023-08-17T01:20:21.933Z","logger":"keep-libp2p","caller":"libp2p/libp2p.go:242","msg":"number of connected peers: [144]"}
2023-08-17 03:19:21.934 {"level":"info","ts":"2023-08-17T01:19:21.934Z","logger":"keep-libp2p","caller":"libp2p/libp2p.go:242","msg":"number of connected peers: [144]"}
2023-08-17 03:18:21.933 {"level":"info","ts":"2023-08-17T01:18:21.933Z","logger":"keep-libp2p","caller":"libp2p/libp2p.go:242","msg":"number of connected peers: [144]"}
2023-08-17 03:17:33.277 	/usr/local/go/src/net/http/server.go:3071 +0x4db
2023-08-17 03:17:33.277 created by net/http.(*Server).Serve
2023-08-17 03:17:33.277 	/usr/local/go/src/net/http/server.go:1966 +0x5d7
2023-08-17 03:17:33.277 net/http.(*conn).serve(0xc00151b0e0, {0x1d6fb38, 0xc013641020})
2023-08-17 03:17:33.276 	/usr/local/go/src/net/http/server.go:2916 +0x43b
2023-08-17 03:17:33.276 net/http.serverHandler.ServeHTTP({0xc001849200?}, {0x1d6f010, 0xc0017162a0}, 0xc001865b00)
2023-08-17 03:17:33.276 	/usr/local/go/src/net/http/server.go:2462 +0x149
[...]

PR #3701, opened 2023-08-24
Wrapping request client calls with a read/write mutex

dimpar:
Refs https://github.com/keep-network/keep-core/issues/3699

Depends on https://github.com/keep-network/keep-core/pull/3704

It happens that the keep-core client panics when interacting with the `go-electrum` library. We noticed a "concurrent write to websocket connection" error was thrown when calling `GetLatestBlockHeight()` function. The stack trace leads to gorilla/websocket/WriteMessage which is called from the `go-electrum` library. The latest block height call was already wrapped with a read mutex, but we should also wrap it with the write mutex to prevent such concurrent errors.


dimpar:
I wasn't able to recreate the `concurrent write to websocket connection` error in the integration tests. They were timing out.

tomaszslabon (approved):

pkg/bitcoin/electrum/electrum.go
```diff
@@ -771,9 +771,9 @@ func requestWithRetry[K interface{}](
 			requestCtx, requestCancel := context.WithTimeout(ctx, c.config.RequestTimeout)
 			defer requestCancel()
 
-			c.clientMutex.RLock()
```
lukasz-zimnoch, line 774: Nitpick: If we no longer use read locks, we can change the `clientMutex` type to `sync.Mutex`.

pkg/bitcoin/electrum/electrum.go
```diff
@@ -771,9 +771,9 @@ func requestWithRetry[K interface{}](
...
+			c.clientMutex.Lock()
```
lukasz-zimnoch, line 774: This effectively creates a semaphore that allows just one request at the same. In theory, this can affect performance if multiple goroutines try to make multiple requests simultaneously. This, in turn, can cause problems in higher layers where some timeouts may be hit unexpectedly. Have we done any benchmarks against real Electrum servers to measure the impact of that change performance-wise?

pkg/bitcoin/electrum/electrum_integration_test.go
```diff
@@ -232,6 +233,43 @@ func TestGetTransactionConfirmations_Negative_Integration(t *testing.T) {
 	})
 }
 
+func TestGetLatestBlockHeightConcurrently_Integration(t *testing.T) {
```
lukasz-zimnoch, line 236: Checking my reasoning: This test asserts that we no longer see the panic described in https://github.com/keep-network/keep-core/issues/3699 when multiple goroutines call the function but we expect occasional failures as described in https://github.com/checksum0/go-electrum/issues/10. Is that correct?

dimpar: Correct, this fixes https://github.com/keep-network/keep-core/issues/3699, but at the same time this test made https://github.com/checksum0/go-electrum/issues/10 issue happen way more frequently because it uses multiple goroutines.

pkg/bitcoin/electrum/electrum.go
dimpar: I run the integration tests and system tests, no performace-wise tests. I'm not sure if we have any benchmarks/performace focused tests (please correct me if I'm wrong). Yes, it might affect the performance but I guess only when we expect _a lot of_ write requests at the same time. Not too familiar with the flows to answer this off the top of my head. FWIW we had a Read mutex already, and I'd expect we read more than we write based on the fact that we have way more getter functions and this particular issue came from `GetLatestBlockHeight`.

The closest to a real-world testing would be including this fix in a release candidate -> deploy to a tesnet and see if we hit any timeout issues. Open for suggestions regarding the performace-wise testing.

lukasz-zimnoch: Agree regarding the `client` instance itself - we rarely reconnect (thus replace this instance) and mostly use it to perform network calls to the Electrum server. But, my whole point was actually about the performance of those network calls. So far, with the read mutex, multiple client goroutines could use the `client` instance to make multiple network calls at the same time. Now, only one client goroutine can do so. Others have to wait for their turn. That means the last goroutines in the queue may wait too long if their predecessors stumble upon longer response times from the Electrum server (or expect other network problems). This can cause cascading failures as those client goroutines may have their own timeouts. This is why I'm raising the performance concern.

That said, I'm wondering whether we can limit the impact of this fix. You said (and this is my reasoning as well), the problem is related only to `GetLatestBlockHeight`, specifically `client.SubscribeHeaders` is problematic. Can't we just wrap the `client.SubscribeHeaders` call with a mutex instead of wrapping the whole `requestWithRetry` function? This way, only one function is affected and we can easily focus on it during real-world tests. WDYT?

dimpar: Hm, I'd also like to limit the impact of this fix, but limiting this mutex only for `GetLatestBlockHeight` might not be enough and we can observe the same issue in other getters. The thing is that most of the getters hit the `s.transport.SendMessage(bytes)` at some point and if the transport layer is set to `ws` then we hit the following code `https://github.com/keep-network/go-electrum/blob/master/electrum/transport_ws.go#L91` and here we _might_ have another occurrence of this issue. I suspect that we observe the problem only for `GetLatestBlockHeight` because this function is being hit more frequently than the other getters in a certain flow. Note that this `concurrent write to websocket connection` error happen not very frequently, but when it does it affects the client uptime.
So having said that, I guess there's a tradeoff we need to make. Limiting the scope of this mutex has a lower impact on performance, but the error might pop up in a different function. Leaving the fix for a wider scope might affect the performance, but no error would occur. The consequence of this error was lowering the uptime bar to 50% because it affected the rewards distribution. Obviously, stakers wouldn't mind a client being slightly slower, but not losing the rewards. Also, I agree with your point that we want to avoid timeout issues that _might_ be a concern, but we don't know if it is. I also suspect that our client had some slack for such cases like this and the timeouts are not balancing on the edge. I'd propose releasing a client with a wider scope mutex fix and seeing if we hit any timeouts in a testing env. Thoughts?

pkg/bitcoin/electrum/electrum.go
dimpar: Sure, [c0e9d25](https://github.com/keep-network/keep-core/pull/3701/commits/c0e9d25fc1868266d6afef35fe26253b05d8c9b2). Let's see what we agree on the discussion below. If we limit the scope of the fix, I'll revert this change. If we leave the scope wider, I'll leave it.

pkg/bitcoin/electrum/electrum.go
lukasz-zimnoch: > The thing is that most of the getters hit the s.transport.SendMessage(bytes) at some point and if the transport layer is set to ws then we hit the following code https://github.com/keep-network/go-electrum/blob/master/electrum/transport_ws.go#L91 and here we might have another occurrence of this issue

Ok, this is the missing piece I was not aware of. Thanks for clarifying that!

>  I'd propose releasing a client with a wider scope mutex fix and seeing if we hit any timeouts in a testing env.

That makes sense now but there is one thing I would add in this case: better observability for those calls. It may happen that we won't detect anything on testnet but this problem will manifest in the future and we won't have an easy way to figure out whether it is actually caused by throttled calls to the Electrum server. We can, for example, simply wrap all `electrum.Connection` methods with a logger that will display `starting [method_name] call to Electrum server` at the beginning and then `completed [method_name] to Electrum server with [success/error] after [time]` at the end. This will be really helpful to have such info logs next to timeout errors if they even occur. That's just a proposition so leaving it up to you but I think it will greatly facilitate future debugging.

pkg/bitcoin/electrum/electrum_integration_test.go
dimpar: That sounds good to me.

pkg/bitcoin/electrum/electrum.go
dimpar: Yes, good idea. Mind if I open this in a separate PR? This one is for fixing an existing error, another PR will be around improving observability. Both will be included in the next release.

pkg/bitcoin/electrum/electrum_integration_test.go
dimpar: Here [9882119](https://github.com/keep-network/keep-core/pull/3701/commits/98821197ac36afa1b5b3f538a6a7a9162c22f437)

pkg/bitcoin/electrum/electrum.go
lukasz-zimnoch: Sure! Leaving this discussion open for future reference.

lukasz-zimnoch (approved):

PR #3705, opened 2023-09-04
Updated Grafana test dashboard 

tomaszslabon:
#Refs: https://github.com/keep-network/keep-core/issues/3608

This PR updates test dashboards (both public and non-public) with new metrics: BTC connectivity and ETH connectivity.
The updated test dashboard can be seen here: https://public.monitoring.test.threshold.network.


dimpar (approved):
Tomasz and I looked at the dashboard and played with value mappings as well. `0/null/na` is assigned to `false`. `1` is assigned to `true`. The dashboard showed the expected results. Looks good to me.

PR #3706, opened 2023-09-05
Support deployment on Sepolia

michalinacienciala:
The Görli testnet currently used by Threshold/Keep for development purposes is planned to become deprecated with the end of year 2023. The testnet that is planned to replace it is called
[Holešky](https://github.com/eth-clients/holesky), however it's not yet available - it's planned it will become widely accessible on Oct 1, 2023 ([source](https://everstake.one/blog/new-ethereum-testnet-holesky-all-you-need-to-know-now)). Switching our infrastructure to support new testnet is quite time consuming, so moving directly from Görli to Holešky may be quite risky, especially if there would be some delays in the date of Holešky genesis (not meeting the planned timelines is not a rare occurrence in the Ethereum space). As a solution, we decided to switch first to another testnet that is currently live - Sepolia. This testnet's EOL is planned for 2026, which gives us plenty of time to move to Holešky before Sepolia gets deprecated.

Refs:
https://github.com/threshold-network/solidity-contracts/issues/150
https://github.com/keep-network/ci/pull/48
https://github.com/threshold-network/solidity-contracts/pull/151
https://github.com/keep-network/tbtc-v2/pull/691
https://github.com/threshold-network/token-dashboard/pull/605

- [ ] Update `TESTNET_ETH_CONTRACT_OWNER_PRIVATE_KEY` so that it is prefixed with `0x`

### DIFF
--- a/.github/workflows/contracts-ecdsa.yml
+++ b/.github/workflows/contracts-ecdsa.yml
@@ -28,7 +28,7 @@ on:
     inputs:
       environment:
         description: "Environment (network) for workflow execution, e.g. `sepolia`"
-        required: false
+        required: true
       upstream_builds:
         description: "Upstream builds"
         required: false
--- a/.github/workflows/contracts-ecdsa.yml
+++ b/.github/workflows/contracts-ecdsa.yml
@@ -27,7 +27,7 @@ on:
   workflow_dispatch:
     inputs:
       environment:
-        description: "Environment (network) for workflow execution, e.g. `goerli`"
+        description: "Environment (network) for workflow execution, e.g. `sepolia`"
         required: false
       upstream_builds:
         description: "Upstream builds"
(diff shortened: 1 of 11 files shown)


.github/workflows/contracts-ecdsa.yml
```diff
@@ -252,9 +252,24 @@ jobs:
 
       - name: Deploy contracts
         env:
-          CHAIN_API_URL: ${{ secrets.GOERLI_ETH_HOSTNAME_HTTP }}
-          ACCOUNTS_PRIVATE_KEYS: ${{ secrets.GOERLI_ETH_CONTRACT_OWNER_PRIVATE_KEY }}
-          ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
+          # Use fake ternary expressions to decide which credentials to use,
+          # depending on chosen environment. Note: if `GOERLI...` credentials
+          # are empty, the expressions will be evaluated to the `SEPOLIA...`
+          # ones.
```
lukasz-zimnoch, line 258: Same question as in https://github.com/threshold-network/solidity-contracts/pull/151#discussion_r1324225132

.github/workflows/contracts-random-beacon.yml
```diff
@@ -334,9 +349,24 @@ jobs:
...
+          # Use fake ternary expressions to decide which credentials to use,
+          # depending on chosen environment. Note: if `GOERLI...` credentials
+          # are empty, the expressions will be evaluated to the `SEPOLIA...`
+          # ones.
+          CHAIN_API_URL: |
+            ${{ inputs.github.event.inputs.environment == 'goerli'
+              && secrets.GOERLI_ETH_HOSTNAME_HTTP
+              || secrets.SEPOLIA_ETH_HOSTNAME_HTTP }}
+          # TODO: Shouldn't we use `ACCOUNTS_PRIVATE_KEYS` here instead of the
+          # two below envs?
```
lukasz-zimnoch, line 361: Could make sense as the deployer is usually the first account. Also, even if we don't use `ACCOUNTS_PRIVATE_KEYS` having two separate envs for Goerli and Sepolia feels a bit redundant as those will likely be the same account.

.github/workflows/contracts-ecdsa.yml
michalinacienciala: Addressed (074739e858aa1eb7a2897097bbe4cc9b7ac51da4 and 63d086d5394e83ba014bcddfbd083580de4183a8)

michalinacienciala:
Ready for review, but I don't want to merge this PR yet before I check the deployment, so leaving as draft for now.

.github/workflows/contracts-random-beacon.yml
michalinacienciala: Decided to use it, see https://github.com/keep-network/keep-core/pull/3706/commits/31367734ef7257213faa88b1aa53b372de27ca20.

lukasz-zimnoch (approved):

PR #3715, opened 2023-09-14
Logs for electrum connection

dimpar:
Refs the discussion https://github.com/keep-network/keep-core/pull/3701#discussion_r1324571809

Electrum client calls sometimes are prone to timeouts or other issues.
For debugging purposes we wrap electrum client calls with a logger and measure the time it takes to complete a call.

### DIFF
--- a/pkg/bitcoin/electrum/electrum.go
+++ b/pkg/bitcoin/electrum/electrum.go
@@ -82,9 +82,6 @@ func (c *Connection) GetTransaction(
 ) (*bitcoin.Transaction, error) {
 	txID := transactionHash.Hex(bitcoin.ReversedByteOrder)
 
-	logger.Infof("starting [%s] call to Electrum server", "GetRawTransaction")
-
-	startTime := time.Now()
 	rawTransaction, err := requestWithRetry(
 		c,
 		func(ctx context.Context, client *electrum.Client) (string, error) {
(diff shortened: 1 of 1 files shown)


pkg/bitcoin/electrum/electrum.go
```diff
@@ -94,14 +93,21 @@ func (c *Connection) GetTransaction(
 			// See: https://github.com/Blockstream/electrs/pull/36
 			return client.GetRawTransaction(ctx, txID)
 		})
+	elapsed := time.Since(startTime)
 	if err != nil {
 		return nil, fmt.Errorf(
-			"failed to get raw transaction with ID [%s]: [%w]",
+			"failed to get raw transaction with ID [%s] after [%s]: [%w]",
```
lukasz-zimnoch, line 99: What do you think about doing it all directly in `requestWithRetry`? I think this can look as follows:
```
func requestWithRetry[K interface{}](
	c *Connection,
	requestFn func(ctx context.Context, client *electrum.Client) (K, error),
	requestName string,
) (K, error) {
	startTime := time.Now()
	logger.Infof("starting [%s] request to Electrum server", requestName)

	var result K

	err := wrappers.DoWithDefaultRetry(
		c.parentCtx,
		c.config.RequestRetryTimeout,
		func(ctx context.Context) error {
			if err := c.reconnectIfShutdown(); err != nil {
				return err
			}

			requestCtx, requestCancel := context.WithTimeout(ctx, c.config.RequestTimeout)
			defer requestCancel()

			c.clientMutex.Lock()
			r, err := requestFn(requestCtx, c.client)
			c.clientMutex.Unlock()

			if err != nil {
				return fmt.Errorf("request failed: [%w]", err)
			}

			result = r
			return nil
		})

	solveRequestOutcome := func(err error) string {
		if err != nil {
			return fmt.Sprintf("error: [%v]", err)
		}
		return "success"
	}
	
	logger.Infof("[%s] request to Electrum server completed with [%s] after [%s]",
		requestName,
		solveRequestOutcome(err),
		time.Since(startTime),
	)

	return result, err
}
```

lukasz-zimnoch (approved):

PR #3716, opened 2023-09-14
Use `threshold-network/tss-lib` fork

lukasz-zimnoch:
Here we replace the original `bnb-chain/tss` library with our own `threshold-network/tss-lib` fork.


dimpar (approved):
lgtm

PR #3719, opened 2023-09-20
Wallet maintainer: Resolve deposit script type at the very end

lukasz-zimnoch:
The script type resolution requires access to transaction data stored on the Bitcoin blockchain. If the given transaction hasn't actually happened but was revealed as deposit to the `Bridge`, the maintainer will not be able to resolve the deposit script type. So far, an error was returned in that case and the maintainer stopped the entire deposit sweep task. This cannot happen as the maintainer can be blocked on deposit sweeps for a long time. Instead, the maintainer should log an error and leave the deposit type unresolved.

### DIFF
--- a/pkg/maintainer/wallet/deposit_sweep.go
+++ b/pkg/maintainer/wallet/deposit_sweep.go
@@ -117,19 +117,6 @@ func FindDeposits(
 
 		logger.Debugf("getting details of deposit %d/%d", i+1, len(depositRevealedEvents))
 
-		depositTransaction, err := btcChain.GetTransaction(
-			event.FundingTxHash,
-		)
-		if err != nil {
-			return result, fmt.Errorf(
-				"failed to get deposit transaction: [%w]",
-				err,
-			)
-		}
-
-		publicKeyScript := depositTransaction.Outputs[event.FundingOutputIndex].PublicKeyScript
-		scriptType := bitcoin.GetScriptType(publicKeyScript)
-
 		depositKey := chain.BuildDepositKey(event.FundingTxHash, event.FundingOutputIndex)
 
 		depositRequest, found, err := chain.GetDepositRequest(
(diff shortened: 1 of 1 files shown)


pkg/maintainer/wallet/deposit_sweep.go
```diff
@@ -172,6 +159,20 @@ func FindDeposits(
...
 
+		var scriptType bitcoin.ScriptType
+		depositTransaction, err := btcChain.GetTransaction(
+			event.FundingTxHash,
+		)
+		if err != nil {
+			logger.Errorf(
+				"failed to get deposit transaction data: [%v]",
+				err,
+			)
```
tomaszslabon, line 170: We don't return on `GetTransaction` error to allow to find other deposits (deposits for which `GetTransaction` succeeded), right?

pkg/maintainer/wallet/deposit_sweep.go
```diff
@@ -172,6 +159,20 @@ func FindDeposits(
 			continue
 		}
 
+		var scriptType bitcoin.ScriptType
```
tomaszslabon, line 162: If there is a `GetTransaction` error `scriptType` will not be assigned any value, is it expected?
Shouldn't we call `continue` for deposits with `GetTransaction` errors?

pkg/maintainer/wallet/deposit_sweep.go
lukasz-zimnoch: Yep exactly. We can't return the error as it will nuke the entire process for all deposits.

pkg/maintainer/wallet/deposit_sweep.go
lukasz-zimnoch: It is expected. We may want to show that transaction for `list-deposits` CLI command anyway. Useful if we want to determine which deposit transactions are actually fakes.

tomaszslabon (approved):

PR #3723, opened 2023-10-02
Hardened wallet actions

lukasz-zimnoch:
Here we introduce some improvements regarding two mechanisms important for wallet actions execution:

### Harden the wallet main UTXO lookup mechanism

The goal of that mechanism is to determine the plain-text wallet main UTXO whose hash is registered in the `Bridge` contract. The plain-text version is necessary to construct wallet transactions on Bitcoin.

The current version of the main UTXO lookup mechanism looks just at the last five confirmed transactions targeting the wallet public key hash to determine the plain text main UTXO of the wallet. This mechanism is not ideal as it doesn't recognize the difference between true wallet transactions and arbitrary transfers to the wallet public key hash that can be made by anyone. That means it is enough to craft several arbitrary transfers to block the main UTXO lookup and prevent the given wallet from performing actions requested by the wallet coordinator.

To address that problem, we are improving the mechanism to take the full transaction history into account. To make it efficient, we are taking just transaction hashes first and fetching full transaction data only for the latest transactions, where the chance to find the wallet UTXO is the highest.

### Harden the wallet sync check mechanism

The goal of this mechanism is to ensure the previous wallet transaction on Bitcoin chain was properly proved to the Bridge contract. This must be ensured before initiating new wallet transactions in order to maintain proper Bitcoin transaction ordering enforced by the Bridge contract. (see https://github.com/keep-network/keep-core/pull/3559 for further reference)

The current version of this mechanism was a naive implementation that checked whether the wallet main UTXO comes from the latest Bitcoin transaction or, if there was no main UTXO, the wallet doesn't have a transaction history. Additionally, this implementation required the mempool to be empty for both cases. This logic is prone to spam transactions sending funds to wallet addresses arbitrarily. Such spam transactions can cause the wallet to abandon all actions proposed by the coordinator.

Here we fix that by using a more sophisticated mechanism:

For wallets having a registered main UTXO, it is enough to check whether their registered UTXO is still among the confirmed unspent outputs from the Bitcoin network standpoint. In order to do that check, we are leveraging ElectrumX `listunspent` method that returns outputs not used as inputs by any (either confirmed or mempool) transaction. If a wallet uses their main UTXO to produce another transaction, `listunspent` will not show it and `EnsureWalletSyncedBetweenChain` will detect this state drift preventing to start another action.

For fresh wallets which don't have main UTXO yet, the situation is more complicated. In that case, we are additionally taking mempool UTXOs into account. If there are no UTXOs at all, that implies the wallet has not produced any (either confirmed or mempool) Bitcoin transaction so far. If some UTXOs targets the wallet, we need to check whether they are spam or actually result of proper wallet transaction. We do this by checking the first input of each transaction. Very first transactions of wallets are always deposit sweeps and all their inputs must point to revealed deposit. If the first input refers to a deposit in that case, that means the wallet already produced their first transaction on Bitcoin and no other action should be taken until the corresponding SPV proof is submitted to the Bridge. Otherwise, such a transaction is spam. If all transactions are spam, the wallet can safely start the given action.

### DIFF
--- a/pkg/tbtc/wallet.go
+++ b/pkg/tbtc/wallet.go
@@ -384,11 +384,6 @@ func (w *wallet) String() string {
 // currently registered in the Bridge on-chain contract. The returned
 // main UTXO can be nil if the wallet does not have a main UTXO registered
 // in the Bridge at the moment.
-//
-// WARNING: THIS FUNCTION CANNOT DETERMINE THE MAIN UTXO IF IT COMES FROM A
-// BITCOIN TRANSACTION THAT IS NOT ONE OF THE LATEST FIVE TRANSACTIONS
-// TARGETING THE GIVEN WALLET PUBLIC KEY HASH. HOWEVER, SUCH A CASE IS
-// VERY UNLIKELY.
 func DetermineWalletMainUtxo(
 	walletPublicKeyHash [20]byte,
 	bridgeChain BridgeChain,
@@ -412,9 +407,12 @@ func DetermineWalletMainUtxo(
 	// to the second last BTC transaction. In theory, such a gap between
 	// the actual latest BTC transaction and the registered main UTXO in
 	// the Bridge may be even wider. To cover the worst possible cases, we
-	// always take the five latest transactions made by the wallet for
-	// consideration.
-	transactions, err := btcChain.GetTransactionsForPublicKeyHash(walletPublicKeyHash, 5)
+	// must rely on the full transaction history. Due to performance reasons,
+	// we are first taking just the transactions hashes (fast call) and then
+	// fetch full transaction data (time-consuming calls) starting from
+	// the most recent transactions as there is a high chance the main UTXO
+	// comes from there.
+	txHashes, err := btcChain.GetTxHashesForPublicKeyHash(walletPublicKeyHash)
 	if err != nil {
 		return nil, fmt.Errorf("cannot get transactions history for wallet: [%v]", err)
 	}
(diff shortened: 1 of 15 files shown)


pkg/tbtc/wallet.go
```diff
@@ -464,89 +471,127 @@ func DetermineWalletMainUtxo(
...
-	// no pending wallet transactions exist in the mempool. That way,
-	// we are handling a plenty of corner cases like transactions races
-	// that could potentially lead to fraudulent transactions and funds loss.
-	mempool, err := btcChain.GetMempoolForPublicKeyHash(walletPublicKeyHash)
-	if err != nil {
-		return fmt.Errorf("cannot get mempool: [%v]", err)
-	}
+		allUtxos := append(confirmedUtxos, mempoolUtxos...)
+		if len(allUtxos) == 0 {
+			// Wallet have not produced any transactions - we are good.
```
tomaszslabon, line 538:
```suggestion
			// Wallet has not produced any transactions - we are good.
```

pkg/tbtc/wallet.go
```diff
@@ -464,89 +471,127 @@ func DetermineWalletMainUtxo(
...
+			return nil
+		}
 
-	if len(mempool) != 0 {
-		return fmt.Errorf("unconfirmed transactions exist in the mempool")
-	}
+		for _, utxo := range allUtxos {
+			// We know that valid first transaction of the wallet always
+			// have just one output. Any utxos with output index other
```
tomaszslabon, line 544:
```suggestion
			// has just one output. Any utxos with output index other
```

tomaszslabon (approved):

pkg/tbtc/wallet.go
```diff
@@ -464,89 +471,127 @@ func DetermineWalletMainUtxo(
...
-		// the Bridge yet.
-		if len(history) != 0 {
-			return fmt.Errorf(
-				"wallet main UTXO doesn't exist but there are BTC " +
-					"transactions produced by the wallet",
-			)
+		// Otherwise, the wallet is a fresh one and requires special
+		// treatment. We need to minimize the chance the wallet is
+		// currently doing their first Bitcoin transaction but, in the same
+		// time, we cannot just assume their transaction history must be
```
tomaszslabon, line 526: "in the same time" = > "at the same time"

pkg/tbtc/wallet.go
```diff
@@ -464,89 +471,127 @@ func DetermineWalletMainUtxo(
...
+					"cannot get deposit request for hash [%s] "+
+						"and output index [%v]: [%v]",
+					input.Outpoint.TransactionHash.String(),
+					input.Outpoint.OutputIndex,
+					err,
+				)
+			}
+
+			if isDeposit {
+				// If that's the case, the wallet was already done their
```
tomaszslabon, line 581: " the wallet was already done" => "the wallet has already done"

PR #3744, opened 2023-11-24
RFC 12 Implementation: Coordination layer groundwork

lukasz-zimnoch:
Refs: https://github.com/keep-network/tbtc-v2/issues/737

Here we present the first part of the changes meant to implement [RFC 12: Decentralized wallet coordination](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-12.adoc) in the tBTC wallet client. This pull request focuses on the groundwork necessary to run the off-chain coordination layer. 

### Coordination layer orchestration

The node orchestrates the coordination layer upon startup. Specifically, it runs two separate goroutines:
- Coordination window watch
- Coordination result processor

Both goroutines are steered by the node's root context and communicate through a dedicated channel.

### Coordination window watch

The coordination window watch goroutine is responsible for detecting coordination windows that occur every `900` blocks. Once a window is detected, it runs the window handler function and passes the window data as an argument. 
The window watch process guarantees that the window handler is called only once for the given coordination window. Moreover, it also guarantees that the window handler is called for all coordination windows. This is achieved by calling each handler in a separate goroutine so the watch loop does not block for long and the chance of missing a coordination window signal is negligible.

The window handler function triggers the window processing logic. Specific steps of that logic are:
1. Determine the list of wallets controlled by the given node
2. Take a coordination executor for each wallet
3. Use the coordination executors to run the coordination procedure for each wallet (in parallel)
4. Collect coordination results and push them to the processing channel

### Coordination executor

The coordination executor is a component that is responsible for running the coordination procedure for the given wallet and coordination window. It is designed to encapsulate the logic of the procedure (coordination seed, communication, and so on). It also ensures that only one instance of the procedure is executed at a time. The executor is also responsible for assembling the coordination procedure's result and reporting all coordination faults detected during execution.

The design of the coordination executor is inspired by the existing signing and DKG executor. It attempts to fit the coordination procedure's logic into the existing codebase in an elegant way.

### Coordination result processor

The coordination result processor goroutine listens for incoming coordination results and triggers the result handler function in a separate goroutine to ensure all results are processed independently.

The result handler function triggers the result processing logic. Specific steps of that logic are:
1. Record coordination faults reported in the result
2. Detect the type of the action proposal being part of the result
3. Fire the appropriate proposal handler

The proposal handlers are part of the existing codebase. They are responsible for the orchestration and execution of the proposed wallet actions.

### Intersection with the existing chain-based coordination mechanism

The presented groundwork was built alongside the existing chain-based coordination mechanism. Some initial integration steps around data types were done. The existing mechanism will be gradually removed and replaced in the follow-up pull requests.

### Next steps

The next steps (coarse-grained) on the way towards RFC 12 implementation are:

- Implement coordination procedure logic (i.e. implement the `coordinationExecutor.coordinate` method)
- Finalize coordination result processing (i.e. implement the `processCoordinationResult` function and refactor `node`'s handlers appropriately)
- Remove the existing chain-based mechanism (i.e. detach `WalletCoordinator`'s events handlers and remove unnecessary code from `chain.go`)
- Modify the SPV maintainter to not rely on `WalletCoordinator`'s events during unproven transactions lookup

### DIFF
--- a/pkg/tbtc/coordination.go
+++ b/pkg/tbtc/coordination.go
@@ -3,6 +3,7 @@ package tbtc
 import (
 	"context"
 	"fmt"
+	"github.com/keep-network/keep-core/pkg/chain"
 	"github.com/keep-network/keep-core/pkg/generator"
 	"github.com/keep-network/keep-core/pkg/net"
 	"github.com/keep-network/keep-core/pkg/protocol/group"
@@ -103,16 +104,95 @@ func watchCoordinationWindows(
 	}
 }
 
+// CoordinationFaultType represents a type of the coordination fault.
+type CoordinationFaultType uint8
+
+const (
+	// FaultUnknown is a fault type used when the fault type is unknown.
+	FaultUnknown CoordinationFaultType = iota
+	// FaultLeaderIdleness is a fault type used when the leader was idle, i.e.
+	// missed their turn to propose a wallet action.
+	FaultLeaderIdleness
+	// FaultLeaderMistake is a fault type used when the leader's proposal
+	// turned out to be invalid.
+	FaultLeaderMistake
+	// FaultLeaderImpersonation is a fault type used when the leader was
+	// impersonated by another operator who raised their own proposal.
+	FaultLeaderImpersonation
+)
+
+func (cft CoordinationFaultType) String() string {
+	switch cft {
+	case FaultUnknown:
+		return "Unknown"
+	case FaultLeaderIdleness:
+		return "LeaderIdleness"
+	case FaultLeaderMistake:
+		return "FaultLeaderMistake"
+	case FaultLeaderImpersonation:
+		return "LeaderImpersonation"
+	default:
+		panic("unknown coordination fault type")
+	}
+}
+
+// coordinationFault represents a single coordination fault.
+type coordinationFault struct {
+	// culprit is the address of the operator that is responsible for the fault.
+	culprit chain.Address
+	// faultType is the type of the fault.
+	faultType CoordinationFaultType
+}
+
+func (cf *coordinationFault) String() string {
+	return fmt.Sprintf(
+		"operator [%s], fault [%s]",
+		cf.culprit,
+		cf.faultType,
+	)
+}
+
+// coordinationProposal represents a single action proposal for the given wallet.
+type coordinationProposal interface {
+	// actionType returns the specific type of the walletAction being subject
+	// of this proposal.
+	actionType() WalletActionType
+	// validityBlocks returns the number of blocks for which the proposal is
+	// valid.
+	validityBlocks() uint64
+}
+
+// noopProposal is a proposal that does not propose any action.
+type noopProposal struct{}
+
+func (np *noopProposal) actionType() WalletActionType {
+	return ActionNoop
+}
+
+func (np *noopProposal) validityBlocks() uint64 {
+	// Panic to make sure that the proposal is not processed by the node.
+	panic("noop proposal does not have validity blocks")
+}
+
 // coordinationResult represents the result of the coordination procedure
 // executed for the given wallet in the given coordination window.
 type coordinationResult struct {
-	actionType WalletActionType
-	//lint:ignore U1000 This function will be used in the future.
-	proposal interface{}
+	wallet   wallet
+	window   *coordinationWindow
+	leader   chain.Address
+	proposal coordinationProposal
+	faults   []*coordinationFault
 }
 
 func (cr *coordinationResult) String() string {
-	return "" // TODO: Implementation.
+	return fmt.Sprintf(
+		"wallet [%s], window [%v], leader [%s], proposal [%s], faults [%s]",
+		&cr.wallet,
+		cr.window.coordinationBlock,
+		cr.leader,
+		cr.proposal.actionType(),
+		cr.faults,
+	)
 }
 
 // coordinationExecutor is responsible for executing the coordination
@@ -143,7 +223,6 @@ func newCoordinationExecutor(
 	}
 }
 
-//lint:ignore U1000 This function will be used in the future.
 // wallet returns the wallet this executor is responsible for.
 func (ce *coordinationExecutor) wallet() wallet {
 	// All signers belong to one wallet. Take that wallet from the
@@ -161,8 +240,18 @@ func (ce *coordinationExecutor) coordinate(
 	}
 	defer ce.lock.Release(1)
 
-	// TODO: Implement coordination logic. Determine how to handle window's
-	//       context.
+	// TODO: Implement coordination logic. Remember about:
+	//       - Setting up the right context
+	//       - Using the protocol latch
+	//       - Using the membership validator
+	//       Example result:
+	result := &coordinationResult{
+		wallet:   ce.wallet(),
+		window:   window,
+		leader:   ce.wallet().signingGroupOperators[0],
+		proposal: &noopProposal{},
+		faults:   nil,
+	}
 
-	return nil, nil
+	return result, nil
 }
--- a/pkg/tbtc/coordination.go
+++ b/pkg/tbtc/coordination.go
@@ -86,7 +86,7 @@ func watchCoordinationWindows(
 	for {
 		select {
 		case block := <-blocksChan:
-			if block % coordinationFrequencyBlocks == 0 {
+			if block%coordinationFrequencyBlocks == 0 {
 				// Make sure the current window is not the same as the last one.
 				// There is no guarantee that the block channel will not emit
 				// the same block again.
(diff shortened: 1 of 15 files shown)


pkg/tbtc/coordination_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	testutils.AssertIntsEqual(
+		t,
+		"end block",
+		1000,
+		int(window.endBlock()),
+	)
+}
+
+func TestCoordinationWindow_IsAfterActivePhase(t *testing.T) {
```
pdyraga, line 33: Shouldn't it be:

```suggestion
func TestCoordinationWindow_IsAfter(t *testing.T) {
```

?

lukasz-zimnoch: Yeah! Copilot tricked me 🤦

pkg/tbtc/coordination.go
```diff
@@ -0,0 +1,257 @@
...
+) (*coordinationResult, error) {
+	if lockAcquired := ce.lock.TryAcquire(1); !lockAcquired {
+		return nil, errCoordinationExecutorBusy
+	}
+	defer ce.lock.Release(1)
+
+	// TODO: Implement coordination logic. Remember about:
+	//       - Setting up the right context
+	//       - Using the protocol latch
+	//       - Using the membership validator
```
pdyraga, line 246: It is something I am not yet fully following. My understanding of this function is that it is doing the work of the centralized coordinator. For example, it checks if the sweeping is necessary or if redemption is necessary and then proposes it. 

If so, it does not feel like the right place to use the protocol latch or membership validator. Shouldn't this happen later in the coordinator result processor that will fire the appropriate proposal handler?

pkg/tbtc/node.go
```diff
@@ -552,6 +643,165 @@ func (n *node) handleRedemptionProposal(
...
+	// This check is actually redundant. We know the node controls some
+	// wallet signers as we just got the wallet from the registry.
+	// However, we are doing it just in case. The API contract of
+	// getWalletsPublicKeys and/or getCoordinationExecutor may change one day.
+	if !ok {
+		procedureLogger.Infof("node does not control signers of this wallet")
+		return nil, false
+	}
+
+	result, err := executor.coordinate(window)
```
pdyraga, line 772: Where will the code check if the current client is the leader for this window?

pkg/tbtc/node_test.go
```diff
@@ -134,6 +136,269 @@ func TestNode_GetSigningExecutor(t *testing.T) {
...
+	}
+
+	// Mock the coordination procedure execution. Return predefined results
+	// on specific coordination windows.
+	executeCoordinationProcedureFn := func(
+		_ *node,
+		window *coordinationWindow,
+		walletPublicKey *ecdsa.PublicKey,
+	) (*coordinationResult, bool) {
+		if signer.wallet.publicKey.Equal(walletPublicKey) {
```
pdyraga, line 295: I don't understand why do we need this check. In `runCoordinationLayer`, `executeCoordinationProcedureFn` is executed only if the current client has a member in the signing group.

pkg/tbtc/deposit_sweep.go
```diff
@@ -47,6 +54,26 @@ const (
...
+	DepositsRevealBlocks []*big.Int
+}
+
+func (dsp *DepositSweepProposal) actionType() WalletActionType {
+	return ActionDepositSweep
+}
+
+func (dsp *DepositSweepProposal) validityBlocks() uint64 {
+	return depositSweepProposalValidityBlocks
+}
```
pdyraga, line 75: This is something that should be proposed by the leader and validated by all followers to double-check they operate on the same value before the protocol execution starts.

pkg/tbtc/coordination.go
lukasz-zimnoch: Sure, I'm aware that the full picture may not be completely visible yet so, let me explain this.

> My understanding of this function is that it is doing the work of the centralized coordinator. For example, it checks if the sweeping is necessary or if redemption is necessary and then proposes it.

This is true but that's just the part of the job. Here is the pseudocode of the `coordinationExecutor.coordinate` function:
```
func (ce *coordinationExecutor) coordinate(
	window *coordinationWindow,
) (*coordinationResult, error) {
    	ce.acquireLock()
        ce.lockProtocolLatch()
        
        seed := ce.coordinationSeed(window)

        ctx := withCancelOnBlock(context.Background, window.activePhaseEndBlock())

        if isLeader(seed) {
            proposal := c.leaderRoutine() // This is the please where the leader checks if sweeping is necessary, etc.
            message := ce.coordinationMessage(proposal)
            ce.broadcastChannel.Send(ctx, message)
            return ce.coordinationResult(proposal)
        } else {
            var proposalChan
            ce.broadcastChannel.Recv(ctx, func(message net.Message) {
		if ce.membershipValidator.IsValidMembership(message) {
                     proposalChan <-message.proposal
                 }
	    })
            return ce.coordinationResult(<-proposalChan)
        }
}
```

pkg/tbtc/node.go
lukasz-zimnoch: Inside `executor.coordinate`, see: https://github.com/keep-network/keep-core/pull/3744#discussion_r1405813388

pkg/tbtc/node_test.go
lukasz-zimnoch: What if we have an error in `runCoordinationLayer` that will run `executeCoordinationProcedureFn` for all wallets? This check is here to ensure that `runCoordinationLayer` runs the procedure only for the expected wallets.

pkg/tbtc/deposit_sweep.go
lukasz-zimnoch: I can't agree. The design proposed in RFC 12 assumes the leader does not propose action execution timing. The only responsibility of the leader is assembling an action proposal. The execution start block is always the last block of the window and the proposal's validity (so, execution timeout) is fixed. The less the leader is supposed to propose, the better. This reduces complexity and room for manipulation. Do you see any particular reason to let the leader propose the validity?

pdyraga: My comment was not clear enough. If `validityBlocks` is a function of `DepositSweepProposal`, one can assume - and that was my assumption as well - that the validity blocks will be eventually sent over the wire like the rest of the proposal fields and that the receiver will check if it is in some expected range. This could in theory give some better flexibility for upgrades and let to adjust the time based on the current situation in the network. However, I don't think this is necessary at this stage. It feels like an additional complexity that we may never need. The version you described is better, but then I would remove the `validityBlocks()` from `DepositSweepProposal` and had it defined as a constant somewhere else. The same for `validityBlocks()` for all other proposals.

pkg/tbtc/node_test.go
pdyraga: So this check is only present in the unit test so that we can catch the problem and have the assertion failed? We will not have this check in the non-test code?

pkg/tbtc/coordination.go
pdyraga: Alright, so this is leader and follower logic. This makes sense. Would be good to make it crystal-clear in the `coordinate` function docs.

pkg/tbtc/deposit_sweep.go
pdyraga: >  I don't think that's a big problem if we don't include it in the serialized proposal version sent over the wire.

If we don't do that, let's just make it crystal-clear in the docs for `validityBlocks()` to avoid confusion and don't let the future-us have a great idea of making it a marshaled/unmarshalled field.

lukasz-zimnoch:
> Did a high-level review, looks good to me. Leaving it in the hands of @tomaszslabon.

Thanks for reviewing @pdyraga !

pkg/tbtc/node.go
```diff
@@ -552,6 +643,165 @@ func (n *node) handleRedemptionProposal(
...
+}
+
+// processCoordinationResult processes the given coordination result.
+func processCoordinationResult(node *node, result *coordinationResult) {
+	logger.Infof("processing coordination result [%s]", result)
+
+	// TODO: Record coordination faults.
+
+	// TODO: Detect proposal type and run the appropriate handler.
+	switch result.proposal.actionType() {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, we need to write a handler for that, e.g. `node.handleMovingFundsProposal`. That handler should orchestrate an implementation of `walletAction` for moving funds that will submit the commitment, assemble and sign the Bitcoin transaction, and so on.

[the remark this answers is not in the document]
tomaszslabon: And in that handle, we will use the `result *coordinationResult` so that we know which wallet we are dealing with?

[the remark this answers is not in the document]
lukasz-zimnoch: Yep! I will show that for existing actions, in one of the follow-up PRs.

tomaszslabon (approved):

PR #3756, opened 2023-12-14
Require ECDSA DKG result challenger to be an EOA

lukasz-zimnoch:
The `challengeDkgResult` function uses several `try-catch` blocks as part of its business logic. However, the EVM has a call stack depth limit equal to 1024. A third-party contract can leverage this limitation and force the `try-catch`-ed calls to revert unconditionally, by using recursion and letting those calls be executed at depth 1025. In such a case, the control flow is passed to the `catch` clauses which may lead to undesired side effects like invalidation of a proper DKG result. To address that problem, we are adding a requirement that `challengeDkgResult` can only be called by an EOA. This prevents third-party contracts from calling `challengeDkgResult`.

### DIFF
--- a/solidity/ecdsa/contracts/WalletRegistry.sol
+++ b/solidity/ecdsa/contracts/WalletRegistry.sol
@@ -800,6 +800,9 @@ contract WalletRegistry is
     ///      requires an extra amount of gas to be left at the end of the
     ///      execution.
     function challengeDkgResult(DKG.Result calldata dkgResult) external {
+        // solhint-disable-next-line avoid-tx-origin
+        require(msg.sender == tx.origin, "Not EOA");
+
         (
             bytes32 maliciousDkgResultHash,
             uint32 maliciousDkgResultSubmitterId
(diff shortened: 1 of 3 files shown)


solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -800,6 +800,9 @@ contract WalletRegistry is
     ///      requires an extra amount of gas to be left at the end of the
     ///      execution.
     function challengeDkgResult(DKG.Result calldata dkgResult) external {
+        // solhint-disable-next-line avoid-tx-origin
+        require(msg.sender == tx.origin, "Not EOA");
```
pdyraga, line 804: Could be worth adding this requirement to the function docs.

pdyraga (approved):
Let's not forget to apply the same fix to the Random Beacon contract.

lukasz-zimnoch:
> Let's not forget to apply the same fix to the Random Beacon contract.

Created an issue to capture that work: https://github.com/keep-network/keep-core/issues/3758

PR #3760, opened 2024-01-08
Compilation error fix due to duplicated generated types

tomaszslabon:
This PR fixes a compilation problem due to duplicated types that were generated by the `abigen`.
The problem is due to an error in the version of `abigen` that we use for contract code generation. 
See https://github.com/keep-network/keep-common/issues/117 for explanation.

### DIFF
--- a/pkg/chain/ethereum/tbtc/gen/Makefile
+++ b/pkg/chain/ethereum/tbtc/gen/Makefile
@@ -25,6 +25,7 @@ define after_abi_hook
 endef
 define fix_wallet_proposal_validator_collision
 	@perl -pi -e s,BitcoinTxInfo,BitcoinTxInfo2,g ./abi/WalletProposalValidator.go
+	@perl -pi -e s,BitcoinTxUTXO,BitcoinTxUTXO3,g ./abi/WalletProposalValidator.go
 endef
 define fix_maintainer_proxy_collision
 	@perl -pi -e s,BitcoinTxUTXO,BitcoinTxUTXO2,g ./abi/MaintainerProxy.go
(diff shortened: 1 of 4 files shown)


pkg/chain/ethereum/tbtc/gen/Makefile
```diff
@@ -35,8 +36,13 @@ endef
 # See explanation in https://github.com/keep-network/keep-common/issues/117.
 define after_contract_hook
 	$(eval type := $(1))
+	$(if $(filter $(type),WalletProposalValidator),$(call fix_wallet_proposal_validator_contract_collision))
 	$(if $(filter $(type),MaintainerProxy),$(call fix_maintainer_proxy_contract_collision))
 endef
+define fix_wallet_proposal_validator_contract_collision
```
tomaszslabon, line 42: In addition to updating `fix_wallet_proposal_validator_collision` a new function had to be added `fix_wallet_proposal_validator_contract_collision` which updates the generated `/contract/WalletProposalValidator.go` and `and /cmd/WalletProposalValidator.go`
The contract and cmd files did not have to updated this way when we only had a collision in `BitcoinTx.Info` as this type is wrapped with another type.
This is not true in case of `BitcoinTx.UTXO` which is passed directly into one of the contract's function.

lukasz-zimnoch (commented):
Looks good but shouldn't we re-generate contract bindings as well?

tomaszslabon:
> Looks good but shouldn't we re-generate contract bindings as well?

See 10b46e4a4885959b4d87828d99c593851797f2d7.
I ran `make` and committed all the files related to `WalletProposalValidator`.

lukasz-zimnoch (approved):

PR #3762, opened 2024-01-15
SPV maintainer: Add coinbase data to submitted SPV proofs

lukasz-zimnoch:
Refs: https://github.com/keep-network/tbtc-v2/security/advisories/GHSA-wg2x-rv86-mmpx

The SPV proof verifier used by the `Bridge` contract has been changed. It now expects the coinbase's preimage and merkle proof to be submitted as part of the proven transaction's SPV proof. Here we adhere to this requirement by adjusting the SPV maintainer accordingly.


pkg/bitcoin/electrum/electrum.go
```diff
@@ -519,6 +519,40 @@ func (c *Connection) getConfirmedScriptHistory(
 	return confirmedItems, nil
 }
```
tomaszslabon, line 521: Could you please add some brief description of this function?

lukasz-zimnoch: Please see cb79db16

tomaszslabon (approved):

Issue #3770, opened 2024-02-05
Reduce overhead around DKG


lukasz-zimnoch:
The extension of the beta operators set made it clear that the current DKG process is resource-heavy and problematic for operators. The current retry scheme is inefficient, a huge amount of network traffic is generated, and logs are often clunky. This task aims to improve the situation.
```[tasklist]
### Tasks
- [ ] https://github.com/keep-network/keep-core/pull/3771
- [ ] https://github.com/keep-network/keep-core/pull/3773
- [ ] https://github.com/keep-network/keep-core/pull/3776
- [ ] https://github.com/keep-network/keep-core/pull/3777
```

PR #3776, opened 2024-02-07
Single-shot DKG

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3770
Depends on: https://github.com/keep-network/keep-core/pull/3775

### DIFF
--- a/pkg/tbtc/dkg.go
+++ b/pkg/tbtc/dkg.go
@@ -28,7 +28,7 @@ const (
 	// is used to calculate the submission delay period that should be respected
 	// by the given member to avoid all members submitting the same DKG result
 	// at the same time.
-	dkgResultSubmissionDelayStepBlocks = 15
+	dkgResultSubmissionDelayStepBlocks = 3
 	// dkgResultApprovalDelayStepBlocks determines the delay step in blocks
 	// that is used to calculate the approval delay period that should be
 	// respected by the given member to avoid all members approving the same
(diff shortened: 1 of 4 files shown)


pkg/tbtc/dkg.go
```diff
@@ -28,7 +28,7 @@ const (
 	// is used to calculate the submission delay period that should be respected
 	// by the given member to avoid all members submitting the same DKG result
 	// at the same time.
-	dkgResultSubmissionDelayStepBlocks = 15
+	dkgResultSubmissionDelayStepBlocks = 3
```
pdyraga, line 31: I am concerned this is too low. The gas price bump happens after [one minute by default](https://github.com/keep-network/keep-common/blob/f89349433020ed33f9cff2eafc07e97dbeef32aa/pkg/chain/ethereum/ethutil/mining_waiter.go#L19) so if the initial estimation was not successful, there will be not enough time to do even one price bump.

pdyraga (commented):
Looks good to me, left just one comment.

pkg/tbtc/dkg.go
lukasz-zimnoch: I tested that yesterday on Sepolia where gas prices were quite high and volatile. Haven't observed any problem so I think we should be good. It is important to reduce DKG timeout to the minimum and this factor strongly commits to that value. We will monitor the situation on mainnet and increase that if necessary.

lukasz-zimnoch: Worth noting that the price bump will be actually done. The only downside is that the next member may front-run with their own transaction. This may lead to a collision sometimes but is not harmful to the protocol.

PR #3788, opened 2024-02-22
Adjustments to the wallet proposal generator

lukasz-zimnoch:
Refs: https://github.com/keep-network/tbtc-v2/issues/781
Depends on: https://github.com/keep-network/tbtc-v2/pull/788 (the `client-build-test-publish` job will become green once 788 is merged and a new `tbtc-v2@development` package with `RedemptionWatchtower` artifact will be published to npm)

The optimistic redemption upgrade introduces a veto mechanism (https://github.com/keep-network/tbtc-v2/pull/788)  that enforces a processing delay on each redemption request. The exact delay value depends on the number of objections raised against the given redemption request. The `WalletProposalValidator` contract has been modified to include validation of that delay factor. The first change ed9c4ae7289e1609150762c1733298f2268be1ec introduces the same for the redemption proposal generator. This ensures the generator issues proposals that conform to the on-chain validation rules and coordination windows are not being wasted.

By the way, we are taking an opportunity to optimize the deposit sweep proposal generation. The `WalletProposalValidator` contract enforces the minimum age of a deposit that can become part of the proposal. The proposal generator
was missing this check and was often producing invalid proposals that were violating the on-chain minimum deposit age rule. We fix that in 55be487d43873e64718dffccd4352e82d7cd5f3f.


pkg/chain/ethereum/tbtc.go
```diff
@@ -194,13 +195,43 @@ func newTbtcChain(
...
+	redemptionWatchtowerAddress, err := bridge.GetRedemptionWatchtower()
+	if err != nil {
+		return nil, fmt.Errorf(
+			"failed to get RedemptionWatchtower address from Bridge: [%v]",
+			err,
+		)
+	}
+
+	var redemptionWatchtower *tbtccontract.RedemptionWatchtower
+	if redemptionWatchtowerAddress != [20]byte{} {
```
tomaszslabon, line 207: We are allowing the `RedemptionWatchtower` not to be set, unlike other contracts which must always be set.
Could you please add a brief explanation why we treat the `RedemptionWatchtower` differently?

lukasz-zimnoch: Sure! Please see: dcdcdec0

tomaszslabon (approved):

PR #3791, opened 2024-03-07
Don't add comments about API docs previews to PRs

michalinacienciala:
Previously we've been adding comments to PRs modifying Solidity contracts about available previews of the API documentation. As each push to the PR branch results in one such comment, PRs with a lot of pushes get quite busy with a lot of the similar comments. In the future it would be good to modify our commenting mechanism so that it would just leave one comment per PR and update it on each push. Until we do that we're turning off the PR commenting, to increase readability.

Ref:
https://github.com/keep-network/tbtc-v2/pull/793
https://github.com/threshold-network/solidity-contracts/pull/165


lukasz-zimnoch (approved):
Thank you for this change!

PR #3794, opened 2024-03-13
Additional check when ensuring wallet synchronised between chains

tomaszslabon:
#Depends on https://github.com/keep-network/keep-core/pull/3793.
This PR adds an additional check when ensuring the wallet is synchronised between chain.
The new check involves verifying if any UTXO assigned to the wallet comes from the moved funds sweep transaction.
We only checked if the UTXO comes from a deposit sweep transaction, but it is also possible a moved funds sweep could be the first transaction coming from a new wallet. 

Additionally, the `GetMovedFundsSweepRequest` function was updated to follow the same pattern as functions for getting deposit sweep requests or pending redemption requests.


pkg/maintainer/spv/moved_funds_sweep.go
```diff
@@ -275,6 +275,12 @@ func isUnprovenMovedFundsSweepTransaction(
 		)
 	}
 
+	// Check is it's a moved funds sweep request at all.
```
lukasz-zimnoch, line 278:
```suggestion
	// Check if it's a moved funds sweep request at all.
```

pkg/tbtc/wallet.go
```diff
@@ -617,11 +624,40 @@ func EnsureWalletSyncedBetweenChains(
...
+						"and output index [%v]: [%v]",
+					input.Outpoint.TransactionHash.String(),
+					input.Outpoint.OutputIndex,
+					err,
+				)
+			}
+
+			if isRequest {
+				// If that's the case, the wallet has already created a moved
+				// funds sweep their first Bitcoin transaction and the Bridge
```
lukasz-zimnoch, line 656:
```suggestion
				// funds sweep as their first Bitcoin transaction and the Bridge
```

lukasz-zimnoch (commented):
Excellent that you caught the need to make the presented modification! Eagle eye 👁️ 

Just two nitpick comments and I'm happy to merge!

pkg/tbtc/wallet.go
tomaszslabon: Done in c5768c558ccac5d71684a8be05eff7abf489e26b.

pkg/maintainer/spv/moved_funds_sweep.go
tomaszslabon: Done in c5768c558ccac5d71684a8be05eff7abf489e26b.

lukasz-zimnoch (approved):

Issue #3796, opened 2024-03-18
Report wallet heartbeat failure


lukasz-zimnoch:
The heartbeat mechanism is implemented but their failures are not reported anywhere. A heartbeat failure is one of the contract-level triggers for the moving funds process hence, it's an important missing piece of the entire wallet lifecycle. The goal of this task is adding heartbeat failure reporting to the client. This consists of two subtasks:
- Making sure unstaking wallet operators do not take part in heartbeat execution (see [this TODO](https://github.com/keep-network/keep-core/blob/ee0ed3ddfa4fcaa912870be17be4f09f6d9343e7/pkg/tbtc/heartbeat.go#L89)). This is an important mechanism that would lead to wallet moving funds in case the majority of wallet operators want to unstake.
- Trigger heartbeat notification failure using [`WalletRegistry.notifyOperatorInactivity`](https://github.com/keep-network/keep-core/blob/ee0ed3ddfa4fcaa912870be17be4f09f6d9343e7/solidity/ecdsa/contracts/WalletRegistry.sol#L875C14-L875C38) function after `N` consecutive heartbeat failures (value of `N` to be determined after a discussion)
```[tasklist]
### Tasks
- [ ] https://github.com/keep-network/keep-core/pull/3800
```

PR #3800, opened 2024-04-08
Heartbeats: Notify operator inactivity on failure

tomaszslabon:
#Refs: https://github.com/keep-network/keep-core/issues/3796.
This PR enables notifying of operator inactivity failure after multiple failed heartbeats.
The inactivity notification is executed after three consecutive heartbeat failures.
A heartbeat is considered as failed if there  was an error during message signing execution or the number of active signers was below a certain threshold.
The operator inactivity notification is performed by `inactivityClaimExecutor` and is similar to the DKG result submission process.


pkg/tbtc/node.go
```diff
@@ -65,6 +66,24 @@ type node struct {
 	// dkgExecutor MUST NOT be used outside this struct.
 	dkgExecutor *dkgExecutor
 
+	heartbeatFailureCountersMutex sync.Mutex
+	// heartbeatFailureCounters holds counters keeping track of consecutive
+	// heartbeat failures. Each wallet has a separate counter. The key used in
+	// the map is the uncompressed public key (with 04 prefix) of the wallet.
+	heartbeatFailureCounters map[string]*uint
```
lukasz-zimnoch, line 73: What do you think about introducing a specialized structure `heartbeatFailureCounter` and passing it to heartbeat actions instead of bare `*uint`? The structure would look as follows:

```
type heartbeatFailureCounter struct {
 mutex sync.Mutex
 counters map[string]uint // uncompressed public key (with 04 prefix) of the wallet as key
}

func (hfc *heartbeatFailureCounter) increment(walletPublicKey string)
func (hfc *heartbeatFailureCounter) reset(walletPublicKey string)
func (hfc *heartbeatFailureCounter) get(walletPublicKey string)
```

The main reason is that dealing with bare `*uint` is error-prone. We always need to dereference the pointer in order to update the underlying value. It's not so hard to make a runtime-level mistake here while doing some changes in the future. Second reason is extensibility. If we ever need to persist the counter to survive client restarts, we have everything in one place and the task will be much easier.

WDYT?

pkg/tbtc/inactivity.go
```diff
@@ -0,0 +1,263 @@
...
+
+const (
+	// inactivityClaimSubmissionDelayStepBlocks determines the delay step in blocks
+	// that is used to calculate the submission delay period that should be respected
+	// by the given member to avoid all members submitting the same inactivity claim
+	// at the same time.
+	inactivityClaimSubmissionDelayStepBlocks = 3
+	// inactivityClaimMaximumSubmissionBlocks determines the maximum block
+	// duration of inactivity claim submission procedure.
+	inactivityClaimMaximumSubmissionBlocks = 60
```
lukasz-zimnoch, line 30: What's the reasoning behind 60?

pkg/tbtc/heartbeat.go
```diff
@@ -111,16 +138,72 @@ func (ha *heartbeatAction) execute() error {
...
+			"leaving without notifying about operator inactivity; current "+
+				"heartbeat failure count is [%d]",
+			*ha.failureCounter,
+		)
+		return nil
+	}
+
+	// The value of consecutive heartbeat failures exceeds the threshold.
+	// Proceed with operator inactivity notification.
+	err = ha.inactivityClaimExecutor.publishClaim(
```
lukasz-zimnoch, line 191: Do we know the duration of `publishClaim`? Asking because:
- `heartbeatProposalValidityBlocks` is `300` blocks so ~1h
- `heartbeatRequestTimeoutSafetyMarginBlocks` is `25` blocks so ~5m

That said, signing can take up to ~55m (see `heartbeatCtx`) and leave no time for `publishClaim`. I think we need to adjust the above parameters to include `publishClaim` in the action time.

pkg/tbtc/heartbeat.go
```diff
@@ -111,16 +138,72 @@ func (ha *heartbeatAction) execute() error {
...
+		// Leave the list of inactive operators empty even if some operators
+		// were inactive during signing heartbeat. The inactive operators could
+		// simply be in the process of unstaking and therefore should not be
+		// punished.
+		[]group.MemberIndex{},
+		true,
+		messageToSign,
+		signingEndBlock,
```
lukasz-zimnoch, line 199: If signing returns an error, `signingEndBlock` will be `0`. I think `publishClaim` does not handle this case.

lukasz-zimnoch (commented):
First part of comments. More to come later.

pkg/tbtc/heartbeat.go
```diff
@@ -111,16 +138,72 @@ func (ha *heartbeatAction) execute() error {
...
 	if err != nil {
-		return fmt.Errorf("cannot sign heartbeat message: [%v]", err)
+		logger.Infof("error while generating heartbeat signature: [%v]", err)
+	} else {
+		logger.Infof(
+			"not enough active operators during signing; required [%d]: "+
+				"actual [%d]",
+			activeOperatorsCount,
+			heartbeatSigningMinimumActiveOperators,
+		)
```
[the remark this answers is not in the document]
tomaszslabon: Done in ae51571c43bc4d40bb8ffa0a17b9ce223266c4d2.

pkg/tbtc/heartbeat.go
```diff
@@ -69,29 +81,44 @@ func newHeartbeatAction(
...
-	// TODO: When implementing the moving funds action we should make sure
-	// heartbeats are not executed by unstaking clients.
+	// Do not execute the heartbeat action if the operator is unstaking.
+	isUnstaking, err := ha.chain.IsOperatorUnstaking()
+	if err != nil {
+		return fmt.Errorf("failed to check if the operator is unstaking")
+	}
+
+	if isUnstaking {
+		logger.Warn(
```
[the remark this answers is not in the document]
tomaszslabon: Good catch. Done in 2f3f7df5a6692e0f5bd773ef5ec530369ba6f7c6.

pkg/tbtc/heartbeat.go
```diff
@@ -111,16 +138,72 @@ func (ha *heartbeatAction) execute() error {
...
+	*ha.failureCounter++
+
+	// If the number of consecutive heartbeat failures does not exceed the
+	// threshold do not notify about operator inactivity.
+	if *ha.failureCounter < heartbeatConsecutiveFailureThreshold {
+		logger.Infof(
+			"leaving without notifying about operator inactivity; current "+
+				"heartbeat failure count is [%d]",
+			*ha.failureCounter,
+		)
```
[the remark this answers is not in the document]
tomaszslabon: Done in 2f3f7df5a6692e0f5bd773ef5ec530369ba6f7c6.

pkg/tbtc/node.go
tomaszslabon: Done in 2695f771739aa688487a89ce29122765342709de.
I wanted to keep it simple and didn't validate the string argument in each function.
We could in theory validate the length of the string argument, but it would make using the counter unwieldy - we would have to check for error every time we use the counter.
Perhaps it would be better to have a structure `heartbeatFailureCounter` for each wallet:
```
type heartbeatFailureCounter struct {
   counter int
}

func (hfc *heartbeatFailureCounter) increment()
func (hfc *heartbeatFailureCounter) reset()
func (hfc *heartbeatFailureCounter) get()
```
and store the map of these objects in a map in `node`?
WDYT?

pkg/tbtc/heartbeat.go
tomaszslabon: Good catch!
If we cannot rely on `signingEndBlock` returned from signing then perhaps we could use `startBlock` or `expiryBlock` stored into the `heartbeatAction`?

Couldn't `expiryBlock` be used as the timeout block for inactivity submission?

pkg/protocol/inactivity/states.go
```diff
@@ -0,0 +1,209 @@
...
+
+func (css *claimSigningState) CanTransition() bool {
+	// Require the number of received signatures to be at least the honest
+	// threshold. Unlike in the case of DKG, we cannot expect all the members to
+	// participate in signing as we know we are dealing with some problem
+	// arising from operator inactivity.
+	// TODO: Consider passing the number of required signatures from the code
+	//       that launched the inactivity operator execution.
+	messagingDone := len(receivedMessages[*claimSignatureMessage](css.BaseAsyncState)) >=
+		css.member.group.HonestThreshold()
```
lukasz-zimnoch, line 90: Shouldn't we expect `>= css.member.group.HonestThreshold() - 1`?

pkg/tbtc/heartbeat_failure.go
```diff
@@ -0,0 +1,41 @@
+package tbtc
+
+import (
+	"sync"
+)
+
+// heartbeatFailureCounter holds counters keeping track of consecutive
+// heartbeat failures. Each wallet has a separate counter. The key used in
+// the map is the uncompressed public key (with 04 prefix) of the wallet.
+type heartbeatFailureCounter struct {
```
lukasz-zimnoch, line 10: Let's move it to the `heartbeat.go` file. No need to create a separate file for one structure.

pkg/sortition/chain.go
```diff
@@ -51,6 +51,10 @@ type Chain interface {
 	// or not.
 	IsEligibleForRewards() (bool, error)
 
+	// IsOperatorUnstaking checks if the operator is unstaking. It returns true
+	// if the operator has deauthorized their entire stake, false otherwise.
+	IsOperatorUnstaking() (bool, error)
```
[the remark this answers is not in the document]
lukasz-zimnoch: Ok got it! However, `sortition.Chain` interface defines chain expectations of the `sortition` package. It's imported by `tbtc.Chain` because `tbtc` package needs the same. In this particular case, `IsOperatorUnstaking` is not needed by `sortition` package so it should be placed in `tbtc` package. That said, I think introducing a new `tbtc.StakingChain` interface makes perfect sense.

pkg/tbtc/node.go
```diff
@@ -405,6 +424,89 @@ func (n *node) getCoordinationExecutor(
...
+		// This is not an error because the node simply does not control
+		// the given wallet.
+		return nil, false, nil
+	}
+
+	// All signers belong to one wallet. Take that wallet from the first signer.
+	wallet := signers[0].wallet
+
+	channelName := fmt.Sprintf(
+		"%s-%s",
```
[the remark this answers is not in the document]
tomaszslabon: Done in 78960d24708a6606d455edd1027a01afd61fd068.

pkg/protocol/inactivity/message.go
```diff
@@ -0,0 +1,42 @@
+package inactivity
+
+import (
+	"github.com/keep-network/keep-core/pkg/protocol/group"
+)
+
+const messageTypePrefix = "tecdsa_inactivity/"
+
+// message holds common traits of all signing protocol messages.
```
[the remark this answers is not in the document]
tomaszslabon: Done in 78960d24708a6606d455edd1027a01afd61fd068.

pkg/protocol/inactivity/message.go
```diff
@@ -0,0 +1,42 @@
+package inactivity
+
+import (
+	"github.com/keep-network/keep-core/pkg/protocol/group"
+)
+
+const messageTypePrefix = "tecdsa_inactivity/"
```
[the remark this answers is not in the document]
tomaszslabon: Done in 78960d24708a6606d455edd1027a01afd61fd068.

pkg/protocol/inactivity/states.go
```diff
@@ -0,0 +1,209 @@
...
+	return nil
+}
+
+func (css *claimSigningState) CanTransition() bool {
+	// Require the number of received signatures to be at least the honest
+	// threshold. Unlike in the case of DKG, we cannot expect all the members to
+	// participate in signing as we know we are dealing with some problem
+	// arising from operator inactivity.
+	// TODO: Consider passing the number of required signatures from the code
+	//       that launched the inactivity operator execution.
```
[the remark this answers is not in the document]
tomaszslabon: Removed the TODO in e9e6b0fa9df8e703d669947aff4bfc21cdf9ba0f.

pkg/tbtc/heartbeat.go
lukasz-zimnoch: I think the easiest way is to make `inactivityClaimExecutor.publishClaim` context-driven. 

Then, in `heartbeatAction.execute` we need to split `heatbeatCtx` into two:
- `heartbeatSigningCtx` that will steer the signing executor
- `heartbeatClaimCtx` that will control the inactivity claim executor

The `heartbeatSigningCtx` should expire in the middle of `heartbeatProposalValidityBlocks`. The `heartbeatClaimCtx` at the end. Both should maintain a safety margin of ~5 minutes (25 blocks). In that configuration we may need to extend `heartbeatProposalValidityBlocks` to e.g. 600 blocks (see https://github.com/keep-network/keep-core/pull/3800#discussion_r1590995784) but leaving that decision to you as I'm not sure what is the execution time of the inactivity claim.

pkg/protocol/inactivity/states.go
lukasz-zimnoch: Yes, and it will always add their own signature on top. That means we need `threshold - 1` of 3rd party signatures to meet the required threshold.

pkg/tbtc/inactivity.go
```diff
@@ -0,0 +1,263 @@
...
+		groupParameters:     groupParameters,
+		protocolLatch:       protocolLatch,
+		waitForBlockFn:      waitForBlockFn,
+	}
+}
+
+func (ice *inactivityClaimExecutor) publishClaim(
+	inactiveMembersIndexes []group.MemberIndex,
+	heartbeatFailed bool,
+	message *big.Int,
```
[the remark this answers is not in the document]
tomaszslabon: Done in c0ec7f7416bec6d6495bf2c3ba9b06ffdeaaef52.

pkg/tbtc/heartbeat_failure.go
tomaszslabon: Done in acd49ee1ec5326a0134c1554ea6c3f16be04dd9c.

pkg/protocol/inactivity/claim.go
```diff
@@ -0,0 +1,55 @@
...
+// ClaimSignatureHashByteSize.
+func ClaimSignatureHashFromBytes(bytes []byte) (ClaimSignatureHash, error) {
+	var hash ClaimSignatureHash
+
+	if len(bytes) != ClaimSignatureHashByteSize {
+		return hash, fmt.Errorf(
+			"bytes length is not equal %v", ClaimSignatureHashByteSize,
+		)
+	}
+	copy(hash[:], bytes[:])
```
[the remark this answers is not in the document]
tomaszslabon: Done in 6fb3bb7db3bf28fb2225929d09b07cf59927db39.

pkg/tbtc/inactivity.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+	}
+
+	return walletMemberIDs, nil
+}
+
+func (ice *inactivityClaimExecutor) publish(
+	ctx context.Context,
+	inactivityLogger log.StandardLogger,
+	seed *big.Int,
```
lukasz-zimnoch, line 230: I think this should be `sessionID`

pkg/tbtc/inactivity_submit.go
```diff
@@ -0,0 +1,209 @@
...
+		chain: chain,
+	}
+}
+
+func (ics *inactivityClaimSigner) SignClaim(claim *inactivity.Claim) (
+	*inactivity.SignedClaim,
+	error,
+) {
+	if claim == nil {
+		return nil, fmt.Errorf("result is nil")
```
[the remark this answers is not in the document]
tomaszslabon: Done in e25efaa96deb4512b86b369fc1e2f67e857bae10.

pkg/tbtc/inactivity.go
lukasz-zimnoch, line 30: Remember that if `inactivityClaimMaximumSubmissionBlocks` is lesser than `groupSize * inactivityClaimSubmissionDelayStepBlocks` only part of the group will have a chance to submit before the timeout.

pkg/tbtc/inactivity_submit.go
```diff
@@ -0,0 +1,209 @@
...
+		// Someone who was ahead of us in the queue submitted the claim. Giving up.
+		ics.inactivityLogger.Infof(
+			"[member:%v] inactivity claim already submitted; "+
+				"aborting inactivity claim on-chain submission",
+			memberIndex,
+		)
+		return nil
+	}
+
+	inactivityClaim, err := ics.chain.AssembleInactivityClaim(
```
lukasz-zimnoch, line 145: We already have `claim` in this context so let's add the `chain*` prefix to make it clear what we assemble here.
```suggestion
	chainClaim, err := ics.chain.AssembleInactivityClaim(
```

pkg/protocol/inactivity/inactivity.go
```diff
@@ -0,0 +1,88 @@
...
+type ClaimSubmitter interface {
+	SubmitClaim(
+		ctx context.Context,
+		memberIndex group.MemberIndex,
+		claim *Claim,
+		signatures map[group.MemberIndex][]byte,
+	) error
+}
+
+func Publish(
```
lukasz-zimnoch, line 37: Let's be more specific:
```suggestion
func PublishClaim(
```

pkg/protocol/inactivity/inactivity.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/keep-network/keep-core/pkg/net"
+	"github.com/keep-network/keep-core/pkg/protocol/group"
+	"github.com/keep-network/keep-core/pkg/protocol/state"
+)
+
+// SignedClaim represents information pertaining to the process of signing
+// an inactivity claim: the public key used during signing, the resulting
+// signature and the hash of the inactivity claim that was used during signing.
+type SignedClaim struct {
```
lukasz-zimnoch, line 17: What about:
```suggestion
type SignedClaimHash struct {
```

I think this name better reflects the purpose. This is not a full claim object.

pkg/protocol/inactivity/claim.go
```diff
@@ -0,0 +1,55 @@
...
+	})
+
+	return sortedIndexes
+}
+
+const ClaimSignatureHashByteSize = 32
+
+// ClaimSignatureHash is a signature hash of the inactivity claim. The hashing
+// algorithm used depends on the client code.
+type ClaimSignatureHash [ClaimSignatureHashByteSize]byte
```
lukasz-zimnoch, line 39: Let's simplify it to just:
```suggestion
type ClaimHash [ClaimSignatureHashByteSize]byte
```

pkg/protocol/inactivity/claim.go
```diff
@@ -0,0 +1,55 @@
...
+	copy(sortedIndexes, c.InactiveMembersIndexes)
+
+	sort.Slice(sortedIndexes, func(i, j int) bool {
+		return sortedIndexes[i] < sortedIndexes[j]
+	})
+
+	return sortedIndexes
+}
+
+const ClaimSignatureHashByteSize = 32
```
lukasz-zimnoch, line 35:
```suggestion
const ClaimHashByteSize = 32
```

pkg/protocol/inactivity/claim.go
```diff
@@ -0,0 +1,55 @@
...
+
+// ClaimSignatureHash is a signature hash of the inactivity claim. The hashing
+// algorithm used depends on the client code.
+type ClaimSignatureHash [ClaimSignatureHashByteSize]byte
+
+// ClaimSignatureHashFromBytes converts bytes slice to ClaimSignatureHash.
+// It requires provided bytes slice size to be exactly
+// ClaimSignatureHashByteSize.
+func ClaimSignatureHashFromBytes(bytes []byte) (ClaimSignatureHash, error) {
```
lukasz-zimnoch, line 44:
```suggestion
func ClaimHashFromBytes(bytes []byte) (ClaimSignatureHash, error) {
```

pkg/tbtc/chain.go
```diff
@@ -120,6 +121,58 @@ type DistributedKeyGenerationChain interface {
...
+// after a submitted inactivity claim lands on the chain.
+type InactivityClaimedEvent struct {
+	WalletID    [32]byte
+	Nonce       *big.Int
+	Notifier    chain.Address
+	BlockNumber uint64
+}
+
+// InactivityChainClaim represents an inactivity claim submitted to the chain.
+type InactivityChainClaim struct {
```
lukasz-zimnoch, line 134: I think we are good to name it just:
```suggestion
type InactivityClaim struct {
```

pkg/protocol/inactivity/claim.go
```diff
@@ -0,0 +1,55 @@
...
+	"crypto/ecdsa"
+	"fmt"
+	"math/big"
+	"sort"
+
+	"github.com/keep-network/keep-core/pkg/protocol/group"
+)
+
+// Claim represents an inactivity claim.
+type Claim struct {
```
lukasz-zimnoch, line 13: Assuming we have `tbtc.InactivityClaim` (see another comment) that represents the full claim type as defined on-chain (data + signatures), I think we should rename this one and emphasize the fact that this is just part of the claim used to build the signature. This way we will also avoid ambiguity between `tbtc.InactivityClaim` and `inactivity.Claim`. What do you think about:

```suggestion
type ClaimPreimage struct {
```

This will explicitly mark the fact that this structure holds data necessary to build the `ClaimHash` that is signed as part of the inactivity protocol.

pkg/protocol/inactivity/claim.go
lukasz-zimnoch, line 1: I would consider moving the content of this file to `inactivity.go` to have all main package types together. This will allow readers to understand the package's API in an easier way.

pkg/tbtc/chain.go
```diff
@@ -120,6 +121,58 @@ type DistributedKeyGenerationChain interface {
...
+	// SubmitInactivityClaim submits the inactivity claim to the chain.
+	SubmitInactivityClaim(
+		claim *InactivityChainClaim,
+		nonce *big.Int,
+		groupMembers []uint32,
+	) error
+
+	// CalculateInactivityClaimSignatureHash calculates hash for the given
+	// inactivity claim.
+	CalculateInactivityClaimSignatureHash(
```
lukasz-zimnoch, line 167:
```suggestion
	CalculateInactivityClaimHash(
```

pkg/tbtc/inactivity.go
```diff
@@ -0,0 +1,263 @@
...
+			walletMemberIDs = append(walletMemberIDs, fetchedOperatorID)
+		} else {
+			walletMemberIDs = append(walletMemberIDs, operatorID)
+		}
+	}
+
+	return walletMemberIDs, nil
+}
+
+func (ice *inactivityClaimExecutor) publish(
```
lukasz-zimnoch, line 227: Right now the `inactivityClaimExecutor` exposes `publishClaim` and `publish` functions. This is ambiguous for external clients. Maybe we can rename them as follows:
- `publishClaim` -> `claimInactivity`. This will follow the pattern of all other executors, e.g. `signingExecutor.sign` or `coordinationExecutor.coordinate`
- `publish` -> `publishInactivityClaim`. Here we go down into the details of "claiming inactivity". Within this function, we execute `inactivity.PublishClaim` which triggers the protocol.

pkg/tbtc/chain.go
tomaszslabon: Done in 0e3b9e5a5e7bfe2b010ecf77509bdfe5536713a2.

pkg/protocol/inactivity/claim.go
tomaszslabon: Done in 0e3b9e5a5e7bfe2b010ecf77509bdfe5536713a2.

pkg/tbtc/chain.go
tomaszslabon: Done in 0e3b9e5a5e7bfe2b010ecf77509bdfe5536713a2.

pkg/protocol/inactivity/claim.go
tomaszslabon: Done in 0e3b9e5a5e7bfe2b010ecf77509bdfe5536713a2.

pkg/protocol/inactivity/claim.go
tomaszslabon: Done in 0e3b9e5a5e7bfe2b010ecf77509bdfe5536713a2.

pkg/protocol/inactivity/claim.go
tomaszslabon: Done in 0e3b9e5a5e7bfe2b010ecf77509bdfe5536713a2.

pkg/protocol/inactivity/inactivity.go
tomaszslabon: Done in 0e3b9e5a5e7bfe2b010ecf77509bdfe5536713a2.

pkg/protocol/inactivity/inactivity.go
tomaszslabon: Done in 0e3b9e5a5e7bfe2b010ecf77509bdfe5536713a2.

pkg/tbtc/inactivity_submit.go
tomaszslabon: Done in 0e3b9e5a5e7bfe2b010ecf77509bdfe5536713a2.

pkg/tbtc/inactivity_submit.go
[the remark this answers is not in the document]
tomaszslabon: Done in 4105cd1c1416d1c0395a8fe021f42e200506ffda.

pkg/protocol/inactivity/claim.go
tomaszslabon: Done in 4105cd1c1416d1c0395a8fe021f42e200506ffda.

pkg/tbtc/inactivity.go
tomaszslabon: Done in b2fcba0ad3fd07397312b04c820de1c7dcdf7daf.

pkg/tbtc/inactivity.go
tomaszslabon: Done in b2fcba0ad3fd07397312b04c820de1c7dcdf7daf.

pkg/chain/ethereum/tbtc.go
```diff
@@ -993,6 +994,196 @@ func (tc *TbtcChain) DKGParameters() (*tbtc.DKGParameters, error) {
...
+			"could not convert signatures to chain format: [%v]",
+			err,
+		)
+	}
+
+	// Sort inactiveMembersIndices slice in ascending order as expected by the
+	// on-chain contract.
+	sort.Slice(inactiveMembersIndices[:], func(i, j int) bool {
+		return inactiveMembersIndices[i] < inactiveMembersIndices[j]
+	})
```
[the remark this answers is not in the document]
tomaszslabon: Done in 3ff579ac1f9118f6a2f6f824549570f93ea69800.

pkg/tbtc/inactivity.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(ice.signers))
+
+	for _, currentSigner := range ice.signers {
+		ice.protocolLatch.Lock()
+		defer ice.protocolLatch.Unlock()
+
+		defer wg.Done()
```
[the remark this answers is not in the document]
tomaszslabon: Good catch! Done in e76d60d7bda9782f12b71cfd6f91fbafa8c07682.

pkg/tbtc/inactivity.go
tomaszslabon: As discussed we decided to make the synchronization context-driven. See 9ad6371c9f14856f97900d95e627f05ed29dc344.

pkg/tbtc/heartbeat.go
tomaszslabon: Done in 9ad6371c9f14856f97900d95e627f05ed29dc344.

pkg/tbtc/inactivity.go
tomaszslabon: As discussed we decided to make the synchronization context-driven. See 9ad6371c9f14856f97900d95e627f05ed29dc344.

pkg/tbtc/heartbeat.go
tomaszslabon: As discussed we decided to make the synchronization context-driven. See 9ad6371c9f14856f97900d95e627f05ed29dc344.

pkg/tbtc/heartbeat.go
```diff
@@ -2,27 +2,43 @@ package tbtc
...
+	// this is the worst-case time for a wallet heartbeat during which the
+	// wallet is busy and cannot take another actions. It includes the total
+	// duration need to perform both both signing the heartbeat message and
+	// optionally notifying about operator inactivity if the heartbeat failed.
+	// The value of 600 blocks is roughly 2 hours, assuming 12 seconds per block.
+	heartbeatTotalProposalValidityBlocks = 600
+	// heartbeatInactivityNotificationValidityBlocks determines the duration
+	// that needs to be preserved for the optional notification about operator
+	// inactivity that follows a failed heartbeat signing.
+	heartbeatInactivityNotificationValidityBlocks = 300
```
lukasz-zimnoch, line 28: Let's be consistent:
```suggestion
	heartbeatInactivityClaimValidityBlocks = 300
```

pkg/tbtc/heartbeat.go
```diff
@@ -2,27 +2,43 @@ package tbtc
...
-	// safety margin that must be preserved between the signing timeout
-	// and the timeout of the entire heartbeat action. This safety
-	// margin prevents against the case where signing completes too late and
-	// another action has been already requested by the coordinator.
+	// safety margin that must be preserved between the timeout of operator
+	// inactivity notification and the timeout of the entire heartbeat action.
+	// This safety margin prevents against the case where signing completes too
+	// late and another action has been already requested by the coordinator.
 	// The value of 25 blocks is roughly 5 minutes, assuming 12 seconds per block.
 	heartbeatRequestTimeoutSafetyMarginBlocks = 25
```
lukasz-zimnoch, line 35: Let's simplify this a bit:
```suggestion
	heartbeatTimeoutSafetyMarginBlocks = 25
```

pkg/tbtc/heartbeat.go
```diff
@@ -2,27 +2,43 @@ package tbtc
...
-	// validity time expressed in blocks. In other words, this is the worst-case
-	// time for a wallet heartbeat during which the wallet is busy and cannot
-	// take another actions. The value of 300 blocks is roughly 1 hour, assuming
-	// 12 seconds per block.
-	heartbeatProposalValidityBlocks = 300
+	// heartbeatTotalProposalValidityBlocks determines the total wallet
+	// heartbeat proposal validity time expressed in blocks. In other words,
+	// this is the worst-case time for a wallet heartbeat during which the
+	// wallet is busy and cannot take another actions. It includes the total
+	// duration need to perform both both signing the heartbeat message and
```
lukasz-zimnoch, line 21: Typo:
```suggestion
	// duration need to perform both signing the heartbeat message and
```

pkg/tbtc/inactivity.go
```diff
@@ -0,0 +1,451 @@
...
+		broadcastChannel:    broadcastChannel,
+		membershipValidator: membershipValidator,
+		groupParameters:     groupParameters,
+		protocolLatch:       protocolLatch,
+		waitForBlockFn:      waitForBlockFn,
+	}
+}
+
+func (ice *inactivityClaimExecutor) claimInactivity(
+	parentCtx context.Context,
```
lukasz-zimnoch, line 70: Let's simplify it and rename the child context (see next comment):
```suggestion
	ctx context.Context,
```

pkg/tbtc/inactivity.go
```diff
@@ -0,0 +1,451 @@
...
+
+			defer wg.Done()
+
+			execLogger.Info(
+				"[member:%v] starting inactivity claim publishing",
+				signer.signingGroupMemberIndex,
+			)
+
+			ctx, cancelCtx := context.WithCancel(parentCtx)
+			defer cancelCtx()
```
lukasz-zimnoch, line 133:
```suggestion
			signerCtx, cancelSignerCtx := context.WithCancel(ctx)
			defer cancelSignerCtx()
```

lukasz-zimnoch (commented):
Code looks good to me 👌 Some last nitpick comments around naming:

pkg/tbtc/heartbeat.go
tomaszslabon: Done in 7137d4f01fb7a99807664d61f4d62b8407d3dee1.

pkg/tbtc/heartbeat.go
tomaszslabon: Done in 7137d4f01fb7a99807664d61f4d62b8407d3dee1.

pkg/tbtc/heartbeat.go
tomaszslabon: Done in 7137d4f01fb7a99807664d61f4d62b8407d3dee1.

lukasz-zimnoch (approved):

pkg/tbtc/heartbeat.go
```diff
@@ -69,29 +101,54 @@ func newHeartbeatAction(
...
+			err,
+		)
+	}
+
+	if isUnstaking {
+		ha.logger.Warn(
+			"quitting the heartbeat action without signing because the " +
+				"operator is unstaking",
+		)
+		return nil
```
tomaszslabon, line 139: @lukasz-zimnoch Should we consider it an error is operator is unstaking? Or leave `nil` as the returned value?

lukasz-zimnoch: Returning `nil` is totally fine. Unstaking operator is not an erroneous case.

lukasz-zimnoch (approved):

Issue #3797, opened 2024-03-18
Archive key material on wallet closure


lukasz-zimnoch:
Wallet operators should listen to `Bridge`'s `WalletClosed` events and perform necessary operations upon wallet closure. The most important thing is archiving the key material so the given wallet is no longer considered active by the operator's node. This is important in the context of the new RFC 12 coordination mechanism as closed wallets' operators should not waste network bandwidth for coordination messages. 
```[tasklist]
### Tasks
- [ ] https://github.com/keep-network/keep-core/pull/3814
```

PR #3814, opened 2024-05-16
Archive key material on wallet closure

tomaszslabon:
#Refs: https://github.com/keep-network/keep-core/issues/3797.
This PR handles archiving key material on wallet closures.
Archiving the wallet data happens in two situations:
- when the `WalletClosed` event is received
- on node start, if there are closed wallets in the node's wallet registry.

Archiving wallet's key material causes the wallet data to be moved to the `archive` directory and the wallet is removed
form the registry's cache..
Such wallets will no longer participate in the RFC-12 coordination.

### DIFF
--- a/pkg/chain/ethereum/tbtc_test.go
+++ b/pkg/chain/ethereum/tbtc_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 
 	"github.com/keep-network/keep-core/internal/testutils"
+	"github.com/keep-network/keep-core/pkg/chain/local_v1"
 	"github.com/keep-network/keep-core/pkg/protocol/group"
 )
 
@@ -277,6 +278,49 @@ func TestCalculateInactivityClaimHash(t *testing.T) {
 		expectedHash,
 		hex.EncodeToString(hash[:]),
 	)
+}
+
+func TestCalculateWalletID(t *testing.T) {
+	hexToByte32 := func(hexStr string) [32]byte {
+		if len(hexStr) != 64 {
+			t.Fatal("hex string length incorrect")
+		}
+
+		decoded, err := hex.DecodeString(hexStr)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		var result [32]byte
+		copy(result[:], decoded)
+
+		return result
+	}
+
+	xBytes := hexToByte32(
+		"9a0544440cc47779235ccb76d669590c2cd20c7e431f97e17a1093faf03291c4",
+	)
+
+	yBytes := hexToByte32(
+		"73e661a208a8a565ca1e384059bd2ff7ff6886df081ff1229250099d388c83df",
+	)
+
+	walletPublicKey := &ecdsa.PublicKey{
+		Curve: local_v1.DefaultCurve,
+		X:     new(big.Int).SetBytes(xBytes[:]),
+		Y:     new(big.Int).SetBytes(yBytes[:]),
+	}
+
+	actualWalletID, err := calculateWalletID(walletPublicKey)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expectedWalletID := hexToByte32(
+		"a6602e554b8cf7c23538fd040e4ff3520ec680e5e5ce9a075259e613a3e5aa79",
+	)
+
+	testutils.AssertBytesEqual(t, expectedWalletID[:], actualWalletID[:])
 }
 
 func TestParseDkgResultValidationOutcome(t *testing.T) {
--- a/pkg/tbtc/node.go
+++ b/pkg/tbtc/node.go
@@ -124,7 +124,13 @@ func newNode(
 	proposalGenerator CoordinationProposalGenerator,
 	config Config,
 ) (*node, error) {
-	walletRegistry := newWalletRegistry(keyStorePersistance)
+	walletRegistry, err := newWalletRegistry(
+		keyStorePersistance,
+		chain.CalculateWalletID,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create wallet registry: [%v]", err)
+	}
 
 	latch := generator.NewProtocolLatch()
 	scheduler.RegisterProtocol(latch)
@@ -146,7 +152,7 @@ func newNode(
 
 	// Archive any wallets that might have been closed or terminated while the
 	// client was turned off.
-	err := node.archiveClosedWallets()
+	err = node.archiveClosedWallets()
 	if err != nil {
 		return nil, fmt.Errorf("cannot archive closed wallets: [%v]", err)
 	}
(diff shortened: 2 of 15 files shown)


pkg/tbtc/node.go
```diff
@@ -1075,6 +1089,172 @@ func processCoordinationResult(node *node, result *coordinationResult) {
...
+	if err != nil {
+		return fmt.Errorf("error getting block counter [%v]", err)
+	}
+
+	currentBlock, err := blockCounter.CurrentBlock()
+	if err != nil {
+		return fmt.Errorf("error getting current block [%v]", err)
+	}
+
+	confirmationBlock := currentBlock + walletClosureConfirmationBlocks
```
lukasz-zimnoch, line 1152: I understand the purpose but:
- This will block client startup for 32 blocks (~6 min) which is bad
- I'm not sure if we actually need this check here. There is a very low probability that a client would restart roughly at the same time a wallet closure transaction gets reorganized and rejected by the chain. A wallet closure is quite a rare event. A reorg that would evict it from the chain is even rarer as this would require another transaction entering just before the closure transaction and changing the state in a way that would revert the closure. Also, even if such a scenario occurs, it would affect only the restarting client so the impact is limited. Last but not least, this is easily recoverable as we can always release a patch to unarchive the key material.

I propose to remove the confirmation check from here and simplify this code. Of course, the confirmation of the trigger based on `WalletClosed` events should stay. This is a different case with a higher probability and potential impact (all nodes are listening for those events at the same time).

pkg/tbtc/registry.go
```diff
@@ -156,6 +156,48 @@ func (wr *walletRegistry) getWalletByPublicKeyHash(
...
+
+	var walletPublicKey *ecdsa.PublicKey
+
+	for _, value := range wr.walletCache {
+		if value.walletPublicKeyHash == walletPublicKeyHash {
+			// All signers belong to one wallet. Take the wallet public key from
+			//  the first signer.
+			walletPublicKey = value.signers[0].wallet.publicKey
+		}
+	}
```
lukasz-zimnoch, line 176: We can use `wr.getWalletByPublicKeyHash` and extract the public key from the returned `wallet` instance. Note that it also returns a boolean flag which can be used to check whether the node controls that wallet.

pkg/tbtc/node.go
```diff
@@ -1075,6 +1089,172 @@ func processCoordinationResult(node *node, result *coordinationResult) {
 	}
 }
 
+// archiveClosedWallets archives closed or terminated wallets.
+func (n *node) archiveClosedWallets() error {
+	getClosedWallets := func(walletPublicKeyHashes [][20]byte) (
+		closedWallets [][20]byte,
+	) {
+		for _, walletPublicKeyHash := range walletPublicKeyHashes {
+			walletChainData, err := n.chain.GetWallet(walletPublicKeyHash)
```
lukasz-zimnoch, line 1098: Once we have the `walletID` cached (see another comment), I think we can call `n.chain.IsWalletRegistered` instead of inspecting the wallet state in the `Bridge`. This way we have the same condition for startup and event-based archiving.

pkg/tbtc/registry.go
```diff
@@ -194,6 +236,21 @@ func (ws *walletStorage) saveSigner(signer *signer) error {
 	return nil
 }
 
+// archiveWallet archives the given wallet data in the underlying persistence
+// layer of the walletStorage.
+func (ws *walletStorage) archiveWallet(walletStoragePath string) error {
```
lukasz-zimnoch, line 241: Let's be consistent:
```suggestion
func (ws *walletStorage) archiveWallet(walletStorageKey string) error {
```

tomaszslabon: Done in 740721253cad39bc9d76c872ddfdc55a98304518.

pkg/tbtc/registry.go
tomaszslabon: I initially wanted to do that, but the problem is that both the new function (`wr.archiveWallet`) and `wr.getWalletByPublicKeyHash` lock the same mutex.
Therefore calling `wr.getWalletByPublicKeyHash` inside `wr.archiveWallet` will cause a deadlock.

The solution could be writing a new function `wr.getWalletByPublicKeyHashImpl` and use it in both `wr.getWalletByPublicKeyHash` and `wr.archiveWallet`. Or we could simply leave it as is.
WDYT @lukasz-zimnoch

pkg/tbtc/node.go
tomaszslabon: Makes sense.
Done in 442905743cf198b3ee3a602ad180951898432acf.

pkg/tbtc/registry.go
lukasz-zimnoch: Yeah, you are right! Let's leave as is.

pkg/tbtc/registry.go
```diff
@@ -156,6 +156,48 @@ func (wr *walletRegistry) getWalletByPublicKeyHash(
...
+	wr.mutex.Lock()
+	defer wr.mutex.Unlock()
+
+	var walletPublicKey *ecdsa.PublicKey
+
+	for _, value := range wr.walletCache {
+		if value.walletPublicKeyHash == walletPublicKeyHash {
+			// All signers belong to one wallet. Take the wallet public key from
+			//  the first signer.
+			walletPublicKey = value.signers[0].wallet.publicKey
```
lukasz-zimnoch, line 174: Shouldn't we `break` the loop once `walletPublicKey` is set?

pkg/tbtc/node.go
```diff
@@ -1075,6 +1089,144 @@ func processCoordinationResult(node *node, result *coordinationResult) {
 	}
 }
 
+// archiveClosedWallets archives closed or terminated wallets.
+func (n *node) archiveClosedWallets() error {
```
lukasz-zimnoch, line 1093: I think we can simplify and optimize the `archiveClosedWallets` function by removing the `getClosedWallets` inner function and using a single `for` loop. What do you think?

pkg/tbtc/registry.go
tomaszslabon: Good catch.
Done in ce85a89584bdb6a6a337f65ead9f32c999d220a0.

pkg/tbtc/node.go
tomaszslabon: Done in 88f3bf6b3041e6f24f8b1cbcee6d17de4d6151ec.

pkg/tbtc/node.go
tomaszslabon: Done in 95c56bb097cd97d7d22d717ded4226e6a5755788.
Although I had to use `n.chain.CalculateWalletID(walletPublicKey)` instead of the stored `walletID`, as the walletID is stored in wallet cache and is not returned from `getWalletByID` or `getWalletByPublicKeyHash`.

pkg/chain/ethereum/tbtc_test.go
```diff
@@ -279,6 +280,49 @@ func TestCalculateInactivityClaimHash(t *testing.T) {
...
+		return result
+	}
+
+	xBytes := hexToByte32(
+		"9a0544440cc47779235ccb76d669590c2cd20c7e431f97e17a1093faf03291c4",
+	)
+
+	yBytes := hexToByte32(
+		"73e661a208a8a565ca1e384059bd2ff7ff6886df081ff1229250099d388c83df",
+	)
```
lukasz-zimnoch, line 306: Are we using an existing wallet PK here? If not, can we use one from `tbtcscan` to make sure this logic is correct?

pkg/tbtc/node.go
```diff
@@ -1075,6 +1095,130 @@ func processCoordinationResult(node *node, result *coordinationResult) {
...
+	if err != nil {
+		return fmt.Errorf("failed to archive the wallet: [%v]", err)
+	}
+
+	logger.Infof(
+		"Successfully archived wallet with wallet ID [0x%x] and public key "+
+			"hash [0x%x]",
+		walletID,
+		walletPublicKeyHash,
+	)
```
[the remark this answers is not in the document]
tomaszslabon: Done in c35ddaf26e051911c15125234fca89e7f17cef32.

pkg/chain/ethereum/tbtc_test.go
tomaszslabon: I copied this test data from some existing `WalletRegistry` contract test:
https://github.com/keep-network/keep-core/blob/ea48ec5f13465461f232b01e92f9a10bf0fab492/solidity/ecdsa/test/WalletRegistry.Wallets.test.ts#L29.

lukasz-zimnoch: Ok, makes sense!

lukasz-zimnoch (dismissed):

lukasz-zimnoch (approved):

PR #3824, opened 2025-02-18
fix-Update lp-tokens.js

Hack666r:


lionakhnazarov:
Closing as stale. lp-tokens.js belongs to the legacy dashboard/v1 surface, which is being retired (#3947) and extracted (#3998). No longer applicable to current main.

PR #3826, opened 2025-03-05
Allowlist for Wallet Registry

pdyraga:
The allowlist contract replaces the Threshold `TokenStaking` contract and is as an outcome of TIP-092 and TIP-100 governance decisions. Staking tokens is no longer required to operate nodes. Beta stakers are selected by the DAO and operate the network based on the allowlist maintained by the DAO. The contract will be integrated with the `WalletRegistry` and replace calls to `TokenStaking`.

I have been experimenting with various approaches, and the most extreme one was to remove most of the `EcdsaAuthorization` logic as well as all `TokenStaking.seize` calls. This would have cascading effects on tBTC Bridge contracts as they rely on `WalletRegistry.seize`. That would also require implementing weight decrease delays in the `Allowlist,` so essentially doing work that is already done in `WalletRegistry`. Considering the pros and cons, I decided on the least invasive option. The `WalletRegistry` still thinks in terms of stake authorization, but everything is based on the staking provider's weight as set in the `Allowlist`, and weight decrease delays are enforced by the existing mechanism in `EcdsaAuthorization`. The `seize` function does nothing except of emitting an event about detecting beta staker misbehavior.

# To be done

## Deployment script

We need to capture all existing beta stakers along with their current authorizations and initialize the `Allowlist` contract. We can do it by either replicating the existing weights or giving them all the same weight.

## Integrate with `WalletRegistry`

There are two approaches to achieve it. The first one is to get rid of all references to `TokenStaking` from tests and update them to work with `Allowlist`. Another approach is to let them work with `TokenStaking` but introduce another integration test for those two contracts. In this option, we could use in `WalletRegistry` something like:

```
    modifier onlyStakingContract() {
        address _allowlist = address(allowlist);
        require(
            // If the allowlist is set, accept calls only from the allowlist.
            // This is post-TIP-98 scenario. If the allowlist is not set, accept
            // calls only from the staking contract. This is pre-TIP-98 scenario.
            (_allowlist != address(0) && msg.sender == _allowlist) ||
                (_allowlist == address(0) && msg.sender == address(staking)),
            "Caller is not the staking contract"
        );
        _;
    }

    /// @notice Initializes V2 version of the WalletRegistry operating with the
    ///         Allowlist contract, as a result of TIP-098 and TIP-100 governance
    ///         decisions.
    function initializeV2(address _allowlist) external reinitializer(2) {
        allowlist = Allowlist(_allowlist);
    }

    /// @dev Provides the expected IStaking reference. If the allowlist is set,
    ///      it acts as the staking contract. If it is not set, the TokenStaking
    ///      acts as the staking contract.
    function _staking() internal returns (IStaking) {
        if (address(allowlist) != address(0)) {
            return IStaking(allowlist);
        }

        return staking;
    }
```

Note that the `WalletRegistry` is close to the maximum allowed contract size and - surprise! - adding the logic above makes it exceed the allowed size. This could potentially be alleviated by removing some of the functionality. For example, in the `challengeDkgResult` function we have a try catch as well as a call to `dkg.requireChallengeExtraGas()`. This could potentially be eliminated as a no-op `seize` in `Allowlist` is guaranteed to always succeed. Also, post [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702), the `require(msg.sender == tx.origin, "Not EOA")` check is no longer guaranteed to work as expected.

### DIFF
--- a/solidity/ecdsa/contracts/Allowlist.sol
+++ b/solidity/ecdsa/contracts/Allowlist.sol
@@ -1,0 +1,206 @@
+// SPDX-License-Identifier: GPL-3.0-only
+//
+// ▓▓▌ ▓▓ ▐▓▓ ▓▓▓▓▓▓▓▓▓▓▌▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▄
+// ▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▌▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
+//   ▓▓▓▓▓▓    ▓▓▓▓▓▓▓▀    ▐▓▓▓▓▓▓    ▐▓▓▓▓▓   ▓▓▓▓▓▓     ▓▓▓▓▓   ▐▓▓▓▓▓▌   ▐▓▓▓▓▓▓
+//   ▓▓▓▓▓▓▄▄▓▓▓▓▓▓▓▀      ▐▓▓▓▓▓▓▄▄▄▄         ▓▓▓▓▓▓▄▄▄▄         ▐▓▓▓▓▓▌   ▐▓▓▓▓▓▓
+//   ▓▓▓▓▓▓▓▓▓▓▓▓▓▀        ▐▓▓▓▓▓▓▓▓▓▓         ▓▓▓▓▓▓▓▓▓▓         ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
+//   ▓▓▓▓▓▓▀▀▓▓▓▓▓▓▄       ▐▓▓▓▓▓▓▀▀▀▀         ▓▓▓▓▓▓▀▀▀▀         ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▀
+//   ▓▓▓▓▓▓   ▀▓▓▓▓▓▓▄     ▐▓▓▓▓▓▓     ▓▓▓▓▓   ▓▓▓▓▓▓     ▓▓▓▓▓   ▐▓▓▓▓▓▌
+// ▓▓▓▓▓▓▓▓▓▓ █▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+// ▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+//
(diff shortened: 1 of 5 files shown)


solidity/ecdsa/contracts/Allowlist.sol
```diff
@@ -0,0 +1,228 @@
...
+            revert ZeroWeight();
+        }
+
+        StakingProviderInfo storage info = stakingProviders[stakingProvider];
+
+        if (info.weight != 0) {
+            revert StakingProviderAlreadyAdded();
+        }
+
+        emit StakingProviderAdded(stakingProvider, weight);
```
jose-blockchain: if makes sense for you, consider moving the emit after state changes for consistency with checks-effects-interactions pattern

solidity/ecdsa/contracts/Allowlist.sol
```diff
@@ -0,0 +1,228 @@
...
+        info.weight = weight;
+        walletRegistry.authorizationIncreased(stakingProvider, 0, weight);
+    }
+
+    /// @notice Allows the governance to request weight decrease for the given
+    ///         staking provider. The change does not take the effect immediately
+    ///         as it has to be approved by the WalletRegistry contract based on
+    ///         decrease delays required. Overwrites pending weight decrease
+    ///         request for the given staking provider. Reverts if the staking
+    ///         provider is now known or if the proposed new weight is higher
```
jose-blockchain: typo: `provider is now known` -> `provider is not known`

solidity/ecdsa/contracts/Allowlist.sol
```diff
@@ -0,0 +1,228 @@
...
+    {
+        return stakingProviders[stakingProvider].weight;
+    }
+
+    /// @notice No-op stake seize operation. After TIP-092 tokens are not staked
+    ///         so there is nothing to seize from.
+    /// @dev The function signature maintains compatibility with Threshold
+    ///      TokenStaking contract to minimize the TIP-092 impact on the
+    ///      WalletRegistry contract.
+    function seize(
```
jose-blockchain: if applicable, this function has no access control and anyone can emit these events. Should at minimum validate `msg.sender == address(walletRegistry)` or implement some access restriction. Otherwise, noise events could fill monitoring systems.

solidity/ecdsa/deploy/17_upgrade_wallet_registry_v2.ts
```diff
@@ -0,0 +1,316 @@
...
+    "upgradeAndCall",
+    [walletRegistryDeployment.address, newImplementation.address, initializeV2Data]
+  )
+
+  // Get ProxyAdmin owner
+  const proxyAdmin = await ethers.getContractAt(
+    ["function owner() view returns (address)"],
+    proxyAdminAddress
+  )
+  const proxyAdminOwner = await proxyAdmin.owner()
```
jose-blockchain: maybe for mainnet deployments, consider adding a sanity check that `proxyAdminOwner` is actually the timelock contract address, not an EOA

solidity/ecdsa/test/WalletRegistry.DualMode.test.ts
```diff
@@ -0,0 +1,576 @@
...
+          toAmount,
+          fromAmount
+        )
+      ).to.be.revertedWithCustomError(walletRegistry, "CallerNotStakingContract")
+
+      await ethers.provider.send("hardhat_stopImpersonatingAccount", [stakingContract.address])
+
+      // This ensures authorization source consistency (no mixing)
+    })
+  })
```
jose-blockchain: just test quality minor comment, the impersonation tests are good, but consider adding a test for what happens when both allowlist AND staking are somehow set to the same address (edge case for storage corruption scenarios)

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -243,25 +253,85 @@ contract WalletRegistry is
         address notifier
     );
 
+    // Custom Errors
+
+    // Authorization Errors
+
+    /// @notice Raised when caller is not the staking contract or allowlist contract.
+    error CallerNotStakingContract();
```
jose-blockchain: now with allow list might be better to change name to `CallerNotAuthorizationSource` or similar

PR #3909, opened 2026-03-25
Remove bootstrap nodes and replace with operator peers

lrsaturnino:
## Problem

The network relies on centrally-managed bootstrap nodes for initial peer discovery. Their embedded public keys bypass firewall `IsRecognized()` staking checks via the AllowList, meaning an unstaked or slashed embedded peer retains permanent network access. Operators hardcoding bootstrap addresses in `--network.peers` will lose connectivity when bootstrap infrastructure is decommissioned.

## Solution

- Replace bootstrap node entries with operator peers (`/dns4/` or `/ip4/` format)
- Decouple firewall AllowList from discovery peers — pass `firewall.EmptyAllowList` so all peers pass `IsRecognized()` staking checks
- Remove dead `ExtractPeersPublicKeys` function
- Deprecate `--network.bootstrap` flag with runtime warning
- Rename `connected_bootstrap_count` metric to `connected_wellknown_peers_count`

## Tests

- `TestValidate_EmptyAllowList_RecognizedPeerAccepted` — recognized peer passes via IsRecognized path
- `TestValidate_EmptyAllowList_UnrecognizedPeerRejected` — unrecognized peer rejected with no AllowList bypass
- `TestValidate_EmptyAllowList_PreviouslyAllowlistedPeerMustPassIsRecognized` — previously allowlisted peer no longer bypasses checks
- `TestResolvePeers` — updated expectations for new operator peer entries (mainnet and testnet)
- `TestNetworkBootstrapFlagDescription_ContainsDeprecationNotice` — flag description includes deprecation text
- `TestIsBootstrap` — returns correct boolean value
- `TestConnectedWellknownPeersCountMetricName` — metric constant has correct value
- `TestObserveConnectedWellknownPeersCount_Callable` — renamed function exists and executes without panic

## Summary by CodeRabbit

* **Deprecations**
  * The `--network.bootstrap` CLI flag is now deprecated and scheduled for removal in v3.0.0.

* **Chores**
  * Updated peer endpoint configurations for mainnet and testnet networks.
  * Enhanced network handshake timeout handling to improve connection reliability.
  * Renamed connected-peer metrics from bootstrap-focused to well-known peers tracking.


lionakhnazarov (dismissed):
lgtm

piotr-roslaniec (dismissed):

piotr-roslaniec (approved):

pdyraga:
I strongly advise against this change.

Bootstrap nodes in the tBTC client are not just an address book. They work in a relay mode, improving the dissemination of network messages. Given the specificity of the tECDSA algorithm used by tBTC nodes, the delivery of messages on time is essential, and messages *are* dropped by nodes when CPU load is high executing tECDSA steps. This is how libp2p floodsub works, and this is how we designed message delivery buffers in the client. Another symptom that is clearly visible in metrics is the number of connections being dropped when executing CPU-intensive tECDSA steps. 

Bootstrap nodes work on a high-availability infrastructure, do not participate in tECDSA, and provide a stable entrance point to the network. If several operator nodes lose their in-memory address books for any reason (could be, for example, a high CPU load leading to a restart or panic in the algorithm), they will find HA bootstrap immediately and recover the peer list in a matter of seconds. 

Eliminating bootstrap nodes should be preceded by changing the network layer design, probably moving to gossip mode, and performing load tests on a larger network resembling mainnet conditions. We made an attempt at this architecture change in the past, and it was not a trivial endeavour.

nkuba:
I agree with @pdyraga here.

The current network architecture has been battle-tested over a long period of time. It went through months of iterative testing on testnet and then further validation in production. This setup is not accidental, it reflects a lot of learnings around reliability and behavior under load.

Bootstrap nodes seem to play a role beyond just peer discovery, especially when it comes to message propagation and overall network stability during heavy tECDSA workloads. Removing them without a proven alternative introduces a real risk.

I strongly recommend against changing such a critical part of the system without thorough validation. At minimum, this should be backed by extensive testing and observation in controlled environments and then gradual rollout with production monitoring.

It’s also worth noting that the testnet is no longer maintained and has effectively been shut down, which makes it even harder to properly validate changes like this before exposing them to mainnet conditions.

In short, I’m in favor of the security improvements, but removing bootstrap infrastructure at this stage feels premature without stronger evidence and testing.

lrsaturnino:
Thanks @pdyraga, @nkuba — I take this seriously and agree it shouldn't be a hard cutover. Let me be precise about scope, share what's been validated so far, and lay out how I'd roll it out.

**Scope.** This retires the dedicated `bst-*.tbtc.boar.network` bootstrap hosts and leans on the staying-operator nodes for discovery. Boar's operator node stays a well-known peer; what's being removed is the dedicated, always-on, non-tECDSA bootstrap role.

**The tradeoff** — it's the one you're flagging: the bootstrap hosts are HA and don't run tECDSA, so they never restart under signing load and give a stable re-entry point; the operator peers that now carry discovery do run tECDSA. We're trading a dedicated non-tECDSA anchor for redundancy across the ~20 operator peers. That's a real change, which is why I'd stage and monitor it rather than cut over.

**Validation completed so far** (controlled environment — Bitcoin testnet4 + Sepolia L1, 3-operator cohort, client `v2.5.3-rc.1` @ `ac7e62e`):

| Scenario | Result |
|---|---|
| Zero-bootstrap cold start, rolling restart, partial outage (S-1–S-3) | PASS — mesh converges and recovers to full peering with **no bootstrap node present** |
| 24h stability (17 ticks) | PASS — peer count stable, **0 panic/fatal**, flat CPU ~40% / mem ~272 MB, no version drift |
| DKG (S-5a) | PASS on-chain — 2 wallets, M8 ≤5 (2 and 1 attempts); wallet #1 operator-log capture was the one PARTIAL sub-gate, covered by wallet #2 |
| Signing + sweep + SPV + mint (S-5b) | PASS — full deposit→sweep→Sepolia mint, incl. strict `e2e-deposit.ts` + on-chain TBTC mint receipt |
| DKG + signing under CPU stress (S-5c) | PASS — `stress-ng` ~70% + a 5h run; DKG no retry elevation; Redemption + DepositSweep signing both complete with SPV proofs |
| Stressed vs unloaded (S-5d) | DKG retries **Δ0**; signing **+37s** (Redemption) / **+50s** (DepositSweep) — no material degradation |
| Deprecation warning (S-4) & rollback rehearsal | PASS — `--network.bootstrap` warns; image swap + rollback recovers full mesh in ~10s |

These were validated on a Sepolia **dev deployment** (not the canonical Threshold Sepolia contracts) plus Bitcoin **testnet4** — reproducible below:

| Artifact | Deployment | Block | Reference (tx / event) |
|---|---|---|---|
| DKG `WalletCreated` (3 wallets) | Sepolia **dev** — `WalletRegistry 0x872d27463e7FE6aC843d7F7AE5f401BC5547c51B` | 10833304 / 10840352 / 10846422 | on-chain events |
| Deposit reveal | Sepolia **dev** — `Bridge 0xd48ae449332a75bc8a28678345d82379bf767656` | 10839195 | `0x84c2c749dcc20104a18579c997dcab3442241ca78e4beeb591865b80bb8668a5` |
| Mint settlement (TBTC `Transfer(0x0→…)`) | Sepolia **dev** — `Bridge 0xd48ae449332a75bc8a28678345d82379bf767656` | 10850702 | `0x24c9228b507fccd918766cb3b41df34a4a5db5530b3869244c6e7a4749c6bc15` |
| BTC sweep | Bitcoin **testnet4** | — | `4fd7f08a57b7bede9c7f385802262aee9bcbfa3ee6bfe64bb1ad5ac462eb9b0b` (mempool.space/testnet4) |

Scope: **3-operator, group-size-3** cohort — it validates functional correctness and behavior under CPU stress at small scale; it does **not** reproduce mainnet TSS message volume (~100-of-N groups). A few items remain open (a re-staking rerun, refreshing the stale Sepolia verification pointer, and restoring SPV diagnostics coverage). That gap is exactly why the mainnet step is staged and monitored rather than released directly.

**Rollout:**

- **Staged, reversible:** canary on a couple of operator nodes first, then widen in waves. I'd keep the `bst-*` bootstrap hosts running until the canary proves operator-peer discovery holds — so if discovery or message delivery degrades, we re-add the bootstrap entries and roll back before going wider. The hosts are decommissioned only once the wider rollout is stable. No single fleet-wide cutover.
- **Message-delivery monitoring:** you're right that floodsub drops messages when a peer's outbound queue backs up under tECDSA CPU. The current monitoring watches connection counts, not delivery — I'll add a message-drop/round-volume metric so we have production signal on exactly that before any bootstrap goes offline.
- **Relay:** reading the client, the only `topic.Relay()` forwarder is the Random Beacon's (opt-in, off by default), and `--network.bootstrap` mode skips beacon/tBTC init so it subscribes to no application topic — so in that mode the `bst-*` hosts wouldn't relay tBTC traffic at the app layer. You'd know exactly how they're deployed, so correct me if they run as full subscribing nodes.
- **Longer term** I agree the deeper answer is the network-layer change you mentioned (we're on floodsub, not gossipsub); I'd scope that separately.

lrsaturnino:
Thanks @nkuba, and agreed on the security improvements. On testnet — you're right the public testnet is effectively gone; the validation above ran on a privately-operated testnet4 cohort against Sepolia. We won't overstate it (group size 3, three operator nodes run by one party, so it doesn't reproduce mainnet TSS volume) — which is exactly why we're leaning on the staged mainnet rollout with production monitoring as the real validation vehicle rather than presenting testnet as proof of mainnet safety.

lionakhnazarov (dismissed):

piotr-roslaniec (dismissed):

piotr-roslaniec (approved):

PR #3910, opened 2026-03-25
fix(ci): run Go tests across all packages

piotr-roslaniec:
## Summary

- CI has been silently running **0 tests** because `gotestsum -- -timeout 15m` (without `./...`) only tests the root package, which has no test files
- Fix: add explicit `./...` so all subpackages are tested
- Also fix `peers_test.go` placeholder hostnames to match actual `config/_peers/testnet` values — this test failure was hidden by the above bug

## Root cause

`gotestsum`'s default `./...` package pattern only applies when **no args** are passed after `--`. With `-- -timeout 15m`, gotestsum forwards args directly to `go test`, which defaults to `.` (current directory only). CI log confirms: `DONE 0 tests in 8.107s`.

## Test plan

- [ ] CI should now run all Go tests (expect 100+ tests instead of 0)
- [ ] `TestResolvePeers/sepolia_network` should pass with corrected hostnames


lrsaturnino (approved):
LGTM

PR #3918, opened 2026-04-01
fix(spv): align deposit sweep proof length with Bridge accumulated difficulty

lionakhnazarov:
Updates the SPV maintainer’s getProofInfo logic so the number of Bitcoin headers in the SPV proof matches what Bridge.evaluateProofDifficulty requires: sum of per-header difficulty ≥ requestedDifficulty × txProofDifficultyFactor, with the first header matching relay current or previous epoch difficulty. Adds relay-window checks so the proof range stays in the previous epoch, current epoch, or spans previous→current only.

This fixes failures like Insufficient accumulated difficulty in header chain on chains where per-block work varies (e.g. testnet4 min-difficulty streaks), where a fixed block count × epoch-average math was wrong.

Changes
pkg/maintainer/spv/spv.go: implement proofRangeWithinRelayWindow; rewrite getProofInfo to walk headers from proof start, sum header.Difficulty(), resolve requestedDiff from first header vs relay difficulties.
pkg/maintainer/spv/spv_test.go + bitcoin_chain_test.go: populate mock headers with realistic Bits/difficulty; fix setCurrentAndPrevEpochDifficulty argument order; update expectations.
(if included) .github/workflows/contracts-ecdsa.yml: resolve accidental merge conflict markers in on: triggers.


pkg/maintainer/spv/spv.go
```diff
@@ -330,135 +365,85 @@ func getProofInfo(
...
-		return true, accumulatedConfirmations, uint(txProofDifficultyFactor.Uint64()), nil
+	firstHeaderDiff := firstHeader.Difficulty()
+
+	var requestedDiff *big.Int
+	switch {
+	case firstHeaderDiff.Cmp(currentEpochDifficulty) == 0:
+		requestedDiff = currentEpochDifficulty
+	case firstHeaderDiff.Cmp(previousEpochDifficulty) == 0:
+		requestedDiff = previousEpochDifficulty
+	default:
```
lrsaturnino, line 400: The `default` branch here silently returns `false`, which makes the caller log "proof goes outside the previous and current difficulty epochs" — but the real issue is the tx block was mined at min-difficulty. 

On testnet4 this will cause silent retries with a misleading log. Could you add a `logger.Warnf` distinguishing this case? Something like "transaction block difficulty %s matches neither current (%s) nor previous (%s) epoch — may be unprovable if mined in a min-difficulty block."

pkg/maintainer/spv/spv_test.go
lrsaturnino, line 1: Missing a test for the scenario that motivated this whole PR — a tx block at min-difficulty where `firstHeaderDiff` matches neither epoch difficulty. Would be good to add a case with `difficultyAtBlock(proofStartBlock) = big.NewInt(1)` while epoch difficulties are much higher, expecting `isProofWithinRelayRange=false`.

lrsaturnino (commented):
Nice one! Just a couple of nits...

PR #3927, opened 2026-04-03
fix(deps): address Semgrep supply chain security findings

piotr-roslaniec:
## Summary

Addresses Semgrep supply chain findings in keep-core across Go modules and JavaScript lock files.

## Go modules

The go-ethereum upgrade (CVE-2026-22862, CVE-2026-22868, CVE-2026-26314) was attempted but reverted -- the new version requires CGO support not available in the Alpine-based CI environment. This will be addressed in a separate PR.

**Known limitations (require separate PRs):**
- **go-ethereum CVE-2026-22862, CVE-2026-22868, CVE-2026-26314**: Upgrade to v1.17.2 reverted due to CGO/Alpine blocker. Separate PR needed with Dockerfile changes.
- **btcd CVE-2024-38365**: Upgrading btcd to v0.24.0+ requires removing the `replace` workaround that redirects `btcd` and `btcd/v2` to compatible versions for transitive deps using the old btcec API. The existing `replace => v0.22.3` predates the v0.23 rewrite where this bug was introduced.
- **tss-lib GHSA-h24c-6p6p-m3vx**: Already mitigated -- `replace` directive uses `github.com/threshold-network/tss-lib` fork instead of the vulnerable `bnb-chain/tss-lib`.

## JavaScript

| Package | Fix | CVEs/MALs |
|---|---|---|
| `scrypt-shim@0.1.0` (ecdsa, random-beacon) | Direct yarn.lock pin to npm security placeholder `0.0.1-security` | MAL-2022-5972 |
| `get-func-name@2.0.0` (ecdsa, random-beacon) | Direct yarn.lock pin to `2.0.2` (yarn projects -- `overrides` field is npm-only and ignored by yarn) | CVE-2023-43646 |
| `get-func-name@2.0.0` (solidity-v1, token-stakedrop, dashboard) | npm `overrides` `>=2.0.2` | CVE-2023-43646 |
| `http-cache-semantics@4.0.x` | npm `overrides` `>=4.1.1` in dashboard, solidity-v1, token-stakedrop | CVE-2022-25881 |
| `axios@0.21.2` (dashboard) | Direct upgrade to `^1.8.2` | CVE-2026-25639 |
| `terser@4.6.3` (dashboard) | npm `overrides` `>=4.8.1` | CVE-2022-25858 |
| `decompress@4.2.0` (dashboard) | npm `overrides` `>=4.2.1` | CVE-2020-12265 |
| `bsock@0.1.9` (token-stakedrop) | `overrides` `>=0.1.10` for non-bundled path; **bundled copy inside bcoin fork cannot be overridden** | CVE-2023-50475 |

## Test plan

- [ ] Go build passes (`go build ./...`)
- [ ] CI passes
- [ ] Semgrep scan shows reduced findings


solidity/ecdsa/package.json
```diff
@@ -79,5 +79,8 @@
   },
   "engines": {
     "node": ">= 14.0.0"
+  },
+  "overrides": {
+    "get-func-name": "^2.0.2"
```
lrsaturnino, line 84: This `overrides` block only affects `package-lock.json`. Since this project also has a `yarn.lock` and CI runs `yarn install --frozen-lockfile`, the override is silently ignored — `get-func-name@2.0.0` (CVE-2023-43646) will still be resolved.

Adding a `resolutions` field would cover the yarn path, but I noticed commit 58f8656 removed a previous `resolutions` entry because it broke `--frozen-lockfile`. So the fix likely needs a `yarn.lock` regeneration step as well.

lrsaturnino (changes requested):
The npm `overrides` field in `solidity/ecdsa/package.json:83` is ignored by Yarn Classic (1.22.x) — Yarn requires `resolutions` instead. All three yarn.lock files still resolve `get-func-name@2.0.0` (CVE-2023-43646):

- `solidity/ecdsa/yarn.lock:6396` — has `overrides` in package.json but yarn ignores it
- `solidity/random-beacon/yarn.lock:5791` — no override or resolution attempted
- `token-stakedrop/merkle-distributor/yarn.lock:3516` — not addressed by this PR

The surgical yarn.lock pattern already used for scrypt-shim works here too — pin `get-func-name` to `2.0.2` with resolved URL `https://registry.yarnpkg.com/get-func-name/-/get-func-name-2.0.2.tgz#0d7cf20cd13fda808669ffa88f4ffc7a3943fc41` and integrity `sha512-8vXOvuE167CtIc3OyItco7N/dpRtBbYOsPsXCz7X/PMnlGjYjSGuZJgM1Y7mmew7BKf9BqvLX2tnOVy1BBUsxQ==`. Alternatively, add `"resolutions": { "get-func-name": "^2.0.2" }` to the relevant package.json files and regenerate the lockfiles.

Without this, Semgrep will keep flagging the same finding this PR aims to fix.

piotr-roslaniec:
Fixed \`get-func-name\` in the two yarn projects where we have direct push access:

- \`solidity/ecdsa/yarn.lock\` -- surgical pin to \`2.0.2\` (same pattern as \`scrypt-shim\`)
- \`solidity/random-beacon/yarn.lock\` -- same

For \`token-stakedrop/merkle-distributor/yarn.lock\`: this lives inside the \`keep-network/merkle-distributor\` submodule where we don't have push access.

**Follow-up plan:** create \`threshold-network/merkle-distributor\` as a maintained copy of \`keep-network/merkle-distributor\`, apply the \`get-func-name\` fix there, and update the submodule URL in \`.gitmodules\` (and the keep-core submodule reference) to point to the new org-owned copy. This gives us long-term control over submodule security fixes without depending on external repo access.

lrsaturnino (approved):
LGTM

PR #3932, opened 2026-04-08
Feat/testnet4 deployment support

lionakhnazarov:
# feat: Bitcoin testnet4 defaults + Sepolia Hardhat/deploy tooling

## Summary

This PR wires the Go client to **Bitcoin testnet4** when using the **Sepolia** Ethereum mapping, adds an **embedded Electrum URL list** for testnet4, and updates **ECDSA / random-beacon Hardhat** config and **deploy scripts** so Sepolia deployments and redeploys work reliably with local artifacts and deployer accounts.
It fixes several failures that were blocking branch validation, including Yarn/Corepack mismatches in Docker and docs workflows, Git/Yarn checkout issues in GitHub Actions, stale random-beacon deploy exports used by ECDSA tests, and flaky/incompatible Go integration tests caused by provider rate limits, Electrum timeouts, and local block watcher race conditions.

## Client (Go)

- Adds `bitcoin.Testnet4` and extends `Network.String()` so embedded Electrum resolution can use `_electrum_urls/testnet4`.
- Maps **Sepolia** → **Bitcoin testnet4** in `config/network` (replacing the legacy testnet mapping).
- Adds `config/_electrum_urls/testnet4` with a default Electrum server URL and extends `config/electrum_test.go` accordingly.

## Solidity / Hardhat

- **ECDSA `hardhat.config.ts`**: makes contract verification optional (`@nomicfoundation/hardhat-verify`), adjusts Sepolia **named accounts** to use index `0` (works with `ACCOUNTS_PRIVATE_KEYS`), trims `external` deployment paths for Sepolia to avoid RPC issues with stale `transactionHash` in npm artifacts, and relaxes typing where needed.
- **Deploy scripts** (`02_deploy_dkg_validator`, `03_deploy_wallet_registry`): supports forced compile via `THRESHOLD_FORCE_DKG_COMPILE`, and sets `skipIfAlreadyDeployed: false` for DKG validator where bytecode must be refreshed after Solidity changes.
- **Random beacon**: updates `export.json`, `hardhat.config.ts`, `05_approve_random_beacon_in_token_staking.ts`, and **yarn.lock** (large lockfile refresh).

## Other

- **CI**: trivial newline-only edits at EOF in `contracts-ecdsa.yml` and `contracts-random-beacon.yml`.
- **`config/peers_test.go`**: adds an extra expected bootstrap peer for the testnet test case (consider splitting out if this PR should stay scoped only to testnet4/Sepolia tooling).
- Fixed GitHub Actions Yarn setup by isolating Git config for Yarn-based git dependency fetches.
- Updated Solidity Docker builds for `ecdsa` and `random-beacon` to include `.yarnrc.yml` and use Corepack with `yarn@4.8.1`.
- Replaced upstream docs workflow usage with a local reusable workflow that installs dependencies with Corepack/Yarn 4.
- Added a bundled `random-beacon` export deploy path for ECDSA so CI does not depend on gitignored sibling exports or stale npm exports.
- Fixed the exported `05_approve_random_beacon_in_token_staking.js` flow to skip networks where `TokenStaking` does not expose `approveApplication`.
- Updated `WalletRegistryGovernance` test fixture setup to load the full deployment fixture so `T` and other external deployments are available.
- Improved Ethereum integration tests by avoiding fragile latest-block decoding and skipping transient provider rate-limit failures.
- Improved Electrum integration tests by skipping transient transport timeout failures from public backends.
- Fixed `local_v1` block counter watcher shutdown behavior and increased watcher buffering to reduce flaky coordination-layer test failures.

## How to test

- `go test ./config/... ./pkg/bitcoin/...`
- From `solidity/ecdsa` / `solidity/random-beacon`: `yarn install` and deploy or dry-run against Sepolia as you normally do for this stack.

## Notes / follow-ups

- Review whether the Electrum URL and deployer/account choices in config are appropriate for a public merge (or replace with placeholders and document env overrides).
- Consider moving **`peers_test.go`** and **pure lockfile/export churn** to separate commits or PRs if reviewers want a minimal diff.
- Regenerate **OpenZeppelin / deployment JSON** locally and commit if your workflow expects those artifacts in-repo.


config/_electrum_urls/testnet4
```diff
@@ -0,0 +1 @@
+ssl://134.199.227.217:50002
```
lrsaturnino: This is a single bare IPv4 over `ssl://` with no DNS and no fallback, which is a step back from the `mainnet` and `testnet` files that use DNS hostnames. Using a hostname both simplifies TLS cert validation and lets you rotate the server without a client-side change — with a bare IP, a VPS outage or IP rotation takes every default-configured testnet4 node offline at once. Happy with any of: swapping to a DNS hostname with a valid TLS cert, adding at least one fallback entry, or shipping a placeholder plus env-override documentation (you already flag this as a follow-up in the PR body). Small nit while we're here — the file is also missing the trailing newline that the other URL files have.

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -479,7 +479,9 @@ contract WalletRegistry is
     function withdrawRewards(address stakingProvider) external {
         address operator = stakingProviderToOperator(stakingProvider);
         if (operator == address(0)) revert UnknownOperator();
-        (, address beneficiary, ) = staking.rolesOf(stakingProvider);
+        (, address beneficiary, ) = _currentAuthorizationSource().rolesOf(
```
lrsaturnino, line 482: This reroutes the reward beneficiary through `_currentAuthorizationSource()` rather than `staking` directly, which means the allowlist when `allowlist != address(0)` and `staking` otherwise. On mainnet today `allowlist()` is non-zero — I checked the live proxy and it returns `0x5E1b4E41ea2eeA6aC7656431CCDC7B204a88386f`, and that contract exposes `rolesOf(address)(address,address,address)`. So once this implementation is redeployed and the proxy is repointed, beneficiary resolution under `withdrawRewards` actually does change on mainnet for any staker where `owner ≠ beneficiary` delegation is in play. Would feel much better if this change landed as its own PR with isolated audit visibility rather than inside the testnet4 work, ideally with a Solidity test that exercises `withdrawRewards()` under an `allowlist != address(0)` state.

solidity/ecdsa/deploy/02_deploy_dkg_validator.ts
```diff
@@ -5,22 +5,23 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
...
 
+  // skipIfAlreadyDeployed: false — so hardhat-deploy compares on-chain bytecode to the
+  // current artifact. With true, an existing deployments/*.json skips deploy entirely
+  // even when Solidity was patched (e.g. groupSize 100 → 3), leaving stale contracts.
   const EcdsaDkgValidator = await deployments.deploy("EcdsaDkgValidator", {
     from: deployer,
     args: [EcdsaSortitionPool.address],
     log: true,
     waitConfirmations: 1,
+    skipIfAlreadyDeployed: false,
```
lrsaturnino: This isn't gated by network, so a future `hardhat deploy --network mainnet` with any bytecode drift — different solc version, optimizer runs, metadata hash — will happily deploy a new `EcdsaDkgValidator` and overwrite `deployments/mainnet/EcdsaDkgValidator.json`. `WalletRegistry.dkgValidator` is constructor-set and won't auto-repoint, so runtime isn't broken immediately, but the repo's deployment record silently desyncs from the active on-chain validator and any later governance step that reads the artifact would end up pointing at the new (possibly untested) contract. Something like `skipIfAlreadyDeployed: hre.network.name === "mainnet"`, or gating via `hre.network.tags.allowRedeploy`, would keep the dev-flow convenience while protecting mainnet. The `THRESHOLD_FORCE_DKG_COMPILE` env flag above is a separate lever and doesn't actually gate the redeploy itself.

solidity/ecdsa/hardhat.config.ts
```diff
@@ -1,8 +1,11 @@
-import "@nomicfoundation/hardhat-verify"
+// import "@nomicfoundation/hardhat-verify" // Optional; comment out if not installed
```
lrsaturnino: The only nudge is that commenting the import out globally also disables verification-via-Hardhat on mainnet going forward, not just on Sepolia. If bumping `@nomicfoundation/hardhat-verify` to a version that targets Etherscan API v2 is feasible, that would restore the tool everywhere; otherwise a quick TODO next to the commented line pinning the expected version would at least leave a breadcrumb for whoever comes back to this.

lrsaturnino (commented):
Hey Lev, nice work on the testnet4 side — the core bits (enum variant, embedded URL, Sepolia remap) look clean and focused. Left a small set of inline comments on the infrastructure pieces that came along; these are the ones I'd flag as worth addressing before merge rather than general suggestions.

One thing worth surfacing at the top is the `WalletRegistry.withdrawRewards` change. I poked at the live mainnet state and the proxy at `0x46d52E41C2F300BC82217Ce22b920c34995204eb` currently returns a non-zero `allowlist()` (`0x5E1b4E41ea2eeA6aC7656431CCDC7B204a88386f`), and that contract does implement `rolesOf`. So once this implementation is redeployed and the proxy repointed, the switch to `_currentAuthorizationSource()` does change beneficiary resolution on mainnet for any staker where `owner ≠ beneficiary` delegation is in play — it's not quite the no-op it might look like. Worth calling out explicitly in the PR description, and I'd really like to see that contract change land as its own PR with a test exercising the allowlist-set path.

solidity/ecdsa/hardhat.config.ts
lionakhnazarov: TODO + version targets are in place until a Hardhat 3 migration is scheduled.

solidity/ecdsa/contracts/WalletRegistry.sol
lionakhnazarov: addressed in #3945

solidity/ecdsa/contracts/WalletRegistry.sol
```diff
@@ -479,7 +479,9 @@ contract WalletRegistry is
...
+            stakingProvider
+        );
```
lrsaturnino, line 481: Body now routes beneficiary lookup through `_currentAuthorizationSource()` (allowlist when set), but the docstring above (lines 452–478) still says `NOT MIGRATED: Beneficiary lookup remains on TokenStaking` and explicitly notes the two sources return different beneficiaries for delegated setups. A mainnet spot-check confirms the divergence for at least one live staking provider.

`withdrawRewards` returns 0 today per TIP-092/100 so the immediate impact is bounded, but the new bytecode ships on the next redeploy. Could we either revert this one-liner, or update the docstring and split it into its own commit? It currently lives inside one titled as a test refactor, which makes the contract change easy to miss.

solidity/ecdsa/hardhat.config.ts
```diff
@@ -1,8 +1,14 @@
-import "@nomicfoundation/hardhat-verify"
+// TODO: Re-enable after Hardhat 3 upgrade — use `@nomicfoundation/hardhat-verify` ^3.x (Etherscan API v2).
+// `^2.1.3` is the last Hardhat-2–compatible line; without the plugin loaded, `hardhat verify` is unavailable on
+// mainnet and all testnets, not only Sepolia.
+// import "@nomicfoundation/hardhat-verify"
```
lrsaturnino: This turns off `hardhat verify` package-wide (the TODO acknowledges it — not only Sepolia). Downstream, `deploy/03_deploy_wallet_registry.ts:62-88` wraps the verify call in a `try/catch` that swallows the error, while the sibling `deploy/02_deploy_dkg_validator.ts:30-32` leaves it bare — so a mainnet redeploy under this config would either succeed with unverified bytecode or throw uncaught, depending on which script runs.

Bumping to `@nomicfoundation/hardhat-verify@^2.1.3` (the last Hardhat-2-compatible line) keeps the plugin available, or gating the commented state on an env var keeps the default verification-on. Either way, aligning the try/catch pattern between `02` and `03` would avoid the divergence.

solidity/random-beacon/.yarn/install-state.gz
piotr-roslaniec, line 1: Should we remove this file?

piotr-roslaniec (changes requested):
Can we split this PR into a series of stacked PRs? That would help a lot with reviewing

lionakhnazarov:
This PR is superseded by a stacked series for easier review:

1. #3999 — CI/Yarn4 tooling foundation  
2. #4000 — Solidity deploy/runtime logic  
3. #4001 — Generated Sepolia deployment artifacts  
4. #4002 — Go testnet4/runtime integration changes

Please review the stack above instead of this combined PR.

PR #3935, opened 2026-04-09
fix: implement audit findings from consolidated review

piotr-roslaniec:
## Summary

Implements 15 fixes from the consolidated audit of PR #3933 (covenant signer + fault isolation). These address findings from 5 review passes (initial triage, 2 multi-agent lens reviews, PR reviewer feedback, validation review).

### Fixes included

**P1 - High**
- **A2**: Use `canonicaljson.Marshal` for handoff payload hash (cross-language determinism)
- **A3/A23**: Improve wallet registry error messages, elevate log to Error, add WalletChainData godoc
- **A4**: Extract Submit critical section into `createOrDedup` helper (5 unlock points -> defer)
- **A5**: Deterministic tiebreaker when both timestamps unparseable (lexicographic RequestID)
- **A6**: Restrict healthz auth bypass to GET method only
- **A22**: Poison route keys from skipped jobs to preserve dedupe semantics

**P2 - Medium**
- **A7**: Document AuthToken CLI flag /proc visibility risk
- **A12/A27**: Cancel service context on init failure + add SIGINT/SIGTERM signal handling
- **A15**: Document advisory flock limitations and storage requirements
- **A16**: Add aggregate load summary with skip count
- **A24**: Remove superseded job from byRequestID on dedup replacement
- **A25**: Rename misleading test after resilient loading change
- **A29**: Use `errors.Is()` for errJobNotFound comparison

---

### Pre-mainnet (must address before mainnet deployment)

**A1: Default `requireApprovalTrustRoots` to true in production**
- **File:** `pkg/covenantsigner/server.go:70-75`
- **Risk:** When `signerApprovalVerifier` is nil, the service accepts requests without cryptographic signer approval verification. The `requireApprovalTrustRoots` config flag exists as mitigation but defaults to false.
- **Action:** Default `requireApprovalTrustRoots` to true in production configs, or fail startup when port is non-zero and no verifier is available.
- **Effort:** Low

**A8: Bitcoin regtest integration tests for witness scripts**
- **File:** `pkg/tbtc/covenant_signer.go` (design)
- **Risk:** Go unit tests cannot enforce `cleanstack`, signature encoding quirks, or opcode limits. A script passing unit tests could be rejected by the Bitcoin network, locking funds.
- **Action:** Introduce `btcd` or Bitcoin Core regtest integration tests that compile witness scripts, sign transactions, and broadcast on a local regtest node.
- **Effort:** High

### Post-merge (code quality and maintainability)

**A9: Split monolithic validation.go (1922 lines)**
- **File:** `pkg/covenantsigner/validation.go`
- **Issue:** Mixes input parsing, crypto verification, normalization, and commitment computation in a single file.
- **Action:** Split into focused files: `validation_quote.go`, `validation_approval.go`, `validation_template.go`.
- **Effort:** High

**A10: Deduplicate UTXO resolution logic**
- **File:** `pkg/tbtc/covenant_signer.go:512-628`
- **Issue:** `resolveSelfV1ActiveUtxo` and `resolveQcV1ActiveUtxo` are near-identical (~60 lines each).
- **Action:** Extract shared `resolveActiveUtxo(request, witnessScript, templateName)` helper.
- **Effort:** Medium

**A11: Deduplicate trust root normalization**
- **File:** `pkg/covenantsigner/validation.go:628-724`
- **Issue:** `normalizeDepositorTrustRoots` and `normalizeCustodianTrustRoots` are structurally identical.
- **Action:** Unify via generics or shared inner function.
- **Effort:** Low

**A13: strictUnmarshal trailing token rejection**
- **File:** `pkg/covenantsigner/validation.go:71-75`
- **Issue:** Inconsistent with server-level `decodeJSON` which rejects trailing JSON tokens.
- **Action:** Add trailing-token rejection or document single-document context constraint.
- **Effort:** Low

**A14: Document Engine interface contract**
- **File:** `pkg/covenantsigner/engine.go`
- **Issue:** No documentation of what nil Transition means (OnSubmit: default to Pending; OnPoll: no-op).
- **Action:** Add godoc specifying the contract and allowed return states.
- **Effort:** Low

**A26: Rate limiting on submit endpoint**
- **File:** `pkg/covenantsigner/server.go:345`
- **Issue:** Authenticated callers can spawn unbounded concurrent 5-minute signing operations. Mitigated by auth and routeRequestID dedup but no per-client throttle exists.
- **Action:** Add token-bucket rate limiting per auth token (e.g., `golang.org/x/time/rate`, ~5 req/min).
- **Effort:** Medium

**A28: Replace reflect.DeepEqual for Handoff comparison**
- **File:** `pkg/covenantsigner/service.go:201`
- **Issue:** `sameJobRevision` uses `reflect.DeepEqual` for `map[string]any` Handoff comparison. Correct but potentially slow at scale.
- **Action:** Define a concrete `HandoffData` struct with an `Equal` method. Acceptable as-is for single-digit RPM throughput.
- **Effort:** Low

**A30: Document qcV1SignerHandoff schema**
- **File:** `pkg/tbtc/covenant_signer.go:35`
- **Issue:** The `Kind` constant serves as a version identifier but the downstream contract for handoff artifacts is implicit.
- **Action:** Add doc comment noting this struct defines the downstream API schema.
- **Effort:** Low

### Type-safety follow-up (from A3)

**A3 (type-level): Add `HasRegistryData` to WalletChainData**
- **File:** `pkg/tbtc/chain.go:418`
- **Issue:** Zero `[32]byte` for MembersIDsHash means "unavailable" but this is implicit. Downstream guards work but future code could silently use the zero value.
- **Action:** Add `HasRegistryData bool` field or use `*[32]byte` for MembersIDsHash to make unavailability explicit at the type level.
- **Effort:** Medium (touches many callers)

---

## Pre-existing test failure

`TestSubmitHandlerPreservesContextValues` fails on the base branch (`fix/review-findings`). Confirmed not introduced by these changes -- the test expects request context values to propagate through the service context detachment, which is not the current design.

## Test plan
- [x] `go build ./pkg/covenantsigner/...` and `go build ./pkg/tbtc/...` compile cleanly
- [x] All covenantsigner tests pass (except pre-existing `TestSubmitHandlerPreservesContextValues`)
- [x] All relevant tbtc tests pass (GetWallet fault isolation, signer approval, payload hash)
- [x] Pinned hash test (`TestComputeQcV1SignerHandoffPayloadHash_DeterministicKeyOrdering`) still passes after canonicaljson switch
- [ ] CI green

### DIFF
--- a/pkg/chain/ethereum/tbtc.go
+++ b/pkg/chain/ethereum/tbtc.go
@@ -1478,8 +1478,10 @@ func (tc *TbtcChain) GetWallet(
 
 	walletRegistryWallet, err := tc.walletRegistry.GetWallet(wallet.EcdsaWalletID)
 	if err != nil {
-		logger.Warnf(
-			"cannot get wallet registry data for wallet [0x%x]: [%v]",
+		logger.Errorf(
+			"wallet registry unavailable for wallet [0x%x]; "+
+				"MembersIDsHash will be zero -- signer approval "+
+				"operations will fail until the registry recovers: [%v]",
 			wallet.EcdsaWalletID,
 			err,
 		)
--- a/pkg/covenantsigner/config.go
+++ b/pkg/covenantsigner/config.go
@@ -10,7 +10,9 @@ type Config struct {
 	// binds to. Empty defaults to loopback-only.
 	ListenAddress string
 	// AuthToken enables static Bearer authentication for signer endpoints.
-	// Non-loopback binds must set this.
+	// Non-loopback binds must set this. Prefer environment variables or
+	// config files over CLI flags to avoid exposing the token in
+	// /proc/PID/cmdline.
 	AuthToken string
 	// EnableSelfV1 exposes the self_v1 signer HTTP routes. Keep this disabled
 	// for a qc_v1-first launch unless self_v1 has cleared its own go-live gate.
--- a/pkg/covenantsigner/server.go
+++ b/pkg/covenantsigner/server.go
@@ -128,7 +128,7 @@ func Initialize(
 		return nil, false, fmt.Errorf("failed to bind covenant signer port [%d]: %w", config.Port, err)
 	}
 
-	go func() {
+	go func() { // #nosec G118 -- parent ctx is already cancelled; shutdown needs a fresh deadline
 		<-ctx.Done()
 
 		// Cancel the service context so in-flight threshold signing
@@ -136,7 +136,7 @@ func Initialize(
 		cancelService()
 
 		shutdownCtx, cancelShutdown := context.WithTimeout(
-			context.Background(), // #nosec G118 -- parent ctx is already cancelled; shutdown needs a fresh deadline
+			context.Background(),
 			5*time.Second,
 		)
 		defer cancelShutdown()
--- a/pkg/covenantsigner/server.go
+++ b/pkg/covenantsigner/server.go
@@ -10,8 +10,10 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os/signal"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/ipfs/go-log/v2"
@@ -124,11 +126,25 @@ func Initialize(
 
 	listener, err := net.Listen("tcp", server.httpServer.Addr)
 	if err != nil {
+		cancelService()
 		return nil, false, fmt.Errorf("failed to bind covenant signer port [%d]: %w", config.Port, err)
 	}
 
+	// Listen for both the parent context cancellation and OS signals so
+	// that in-flight signing operations are cancelled promptly on any
+	// shutdown path, including SIGINT/SIGTERM.
+	signalCtx, stopSignal := signal.NotifyContext(
+		context.Background(),
+		syscall.SIGINT,
+		syscall.SIGTERM,
+	)
+
 	go func() {
-		<-ctx.Done()
+		select {
+		case <-ctx.Done():
+		case <-signalCtx.Done():
+		}
+		stopSignal()
 
 		// Cancel the service context so in-flight threshold signing
 		// operations observe shutdown and terminate promptly.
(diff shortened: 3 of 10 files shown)


mswilkison (commented):
Findings

- High: `pkg/covenantsigner/server.go:136` now installs its own `signal.NotifyContext` for `SIGINT`/`SIGTERM`, but the normal node start path still builds the process root context from `context.Background()` in `cmd/start.go:145` and then waits forever on `<-ctx.Done()` in `cmd/start.go:276`. With covenant signer enabled, the first `SIGINT`/`SIGTERM` will now be consumed by the signer's internal handler, which only cancels `serviceCtx` and shuts down the signer HTTP server. The rest of the node keeps running instead of exiting on the first signal. Before this change, the process terminated normally. This signal handling needs to live at the top level and cancel the shared root context, rather than being added inside `covenantsigner.Initialize`.

- Medium: The new poisoned-route safeguard can mask a valid surviving job. In `pkg/covenantsigner/store.go:240`, every malformed file with a recoverable route key marks that route as poisoned, and `pkg/covenantsigner/store.go:371` returns `errPoisonedRouteKey` before checking `byRouteKey`. That means if startup sees both a valid job and a malformed stale sibling for the same route key, retries by `routeRequestId` stop deduping to the valid job and instead fail with "manual recovery required." A realistic path to that state already exists when a replacement leaves the old file behind after delete failure. Poisoning should only apply when no valid winner for that route survives load, or be cleared once a valid job is loaded for the same key.

lrsaturnino (approved):
LGTM

lrsaturnino:
Dropped the poison-route mechanism (A22) in 52a3f2d in favor of hard-fail on corrupt persisted job files, aligning with the approach in PR #3933.

The poison-route had two issues: (1) an ordering bug where a valid job iterated before its malformed sibling left the route permanently poisoned despite the valid job being loaded, and (2) when the corrupt file was unrecoverable (binary garbage), no poison was set at all, silently losing dedupe protection. Hard-fail is the safer default for a pre-mainnet signing service — ambiguous persisted state should block startup rather than risk duplicate signing jobs.

The rest of this PR (createOrDedup extraction, healthz auth bypass, canonicaljson, error messages, test fixes, signal handler removal) is unaffected.

PR #3938, opened 2026-04-10
fix: implement deep-audit findings for covenant signer

piotr-roslaniec:
## Summary

Implements 8 fixes from the deep audit of the covenant signer feature (20 agent passes across 5 analysis areas + cross-provider verification with Gemini and GPT models).

### P1 Fixes

- **A31**: Auto-derive `DataDir` from `storage.Dir` when unset and signer port is non-zero. Without this, the exclusive file lock was silently disabled, allowing two processes to corrupt the job store concurrently.
- **A32**: Replace `context.Background()` with `signal.NotifyContext(SIGINT, SIGTERM)` in `cmd/start.go`. The shutdown goroutine (HTTP drain, cancel signing ops, release file lock) was dead code -- it blocked on `ctx.Done()` which never fired.
- **A33**: Unconditionally evict stale `byRequestID` entries on route-key replacement in `Put()`, and add route-key holder verification in `loadPollJob`. Previously, a failed file delete left an orphan that a stale Poll could use to seize the route key from a newer job.

### P2 Fixes

- **A34**: Clear poisoned route keys on successful `Put()`. Poison was self-reinforcing at runtime because `Submit` hit `GetByRouteRequest` (which returns `errPoisonedRouteKey`) before reaching `Put()`.
- **A36**: Normalize `migrationPlanQuoteTrustRoots` KeyID at startup, matching the pattern for depositor/custodian roots. Config with accidental whitespace in KeyID caused immediate request rejection.
- **A38**: Release file lock if trust root normalization fails after `NewStore` succeeds, using a deferred cleanup on named error return.

### P3 Fixes

- **A35**: Trim whitespace from `destination.ReservationID` in `normalizeMigrationDestination`, matching `normalizeMigrationPlanQuote`. Asymmetric trim caused spurious rejection for whitespace-padded IDs.
- **A37**: Enforce low-S normalization on signer approval certificate threshold signature. Without this, a certificate holder could submit the high-S variant to create a duplicate job (doubled resource consumption, no fund loss).

### Audit methodology

- 20 agent runs (4 passes x 5 areas: coverage, contract verification, threat model, property testing, integration seams)
- Cross-provider verification: Gemini 3 Pro Preview (security), Gemini 2.5 Pro (integration), GPT-5.2 (state layer)
- Multi-agent consolidated review for final severity calibration
- All 7 Bitcoin P2WSH spending paths verified correct via script execution traces
- No P0 (fund-loss) vulnerabilities found

## Test plan
- [x] `go build ./cmd/... ./pkg/covenantsigner/... ./pkg/tbtc/...` compiles cleanly
- [x] `go test -race ./pkg/covenantsigner/...` passes
- [x] Relevant tbtc tests pass (GetWallet, signer approval, payload hash)
- [ ] CI green


pkg/covenantsigner/store.go
```diff
@@ -349,6 +355,7 @@ func (s *Store) Put(job *Job) error {
 
 	s.byRequestID[job.RequestID] = cloned
 	s.byRouteKey[key] = job.RequestID
+	delete(s.poisonedRoutes, key)
```
lrsaturnino: `s.poisonedRoutes` isn't a declared field on `Store` — the struct definition (lines 18–25 of this file) declares only `handle`, `mutex`, `lockFile`, `byRequestID`, and `byRouteKey`, and no `errPoisonedRouteKey` sentinel or poison-tracking infrastructure exists anywhere in the PR. Repo-wide `git grep -n 'poisonedRoutes\|errPoisonedRouteKey'` on `origin/fix/deep-audit-findings` returns only this one line.

Empirically from the head ref:
- `go build ./pkg/covenantsigner/...` → exits 1 with `pkg/covenantsigner/store.go:358:11: s.poisonedRoutes undefined (type *Store has no field or method poisonedRoutes)`
- `go build ./pkg/tbtc/...` and `go build ./cmd/...` transitively fail (both import `pkg/covenantsigner`)
- `go test ./pkg/covenantsigner/...` → build-failed
- `go vet ./pkg/covenantsigner/...` → same undefined-field error

On the base ref (`origin/fix/review-findings`) all of the above succeed.

Commit `f54d4e4d4` (the A34 fix) is a single-line insertion of this `delete(...)` call with no accompanying struct-field declaration or producer-side poison-setting code — looks like a rebase may have dropped the foundation commit.

Flagging only, not a change request.

lrsaturnino (commented):
Flagging a compile-breaking reference found during a deep review of this PR — see the inline comment on `pkg/covenantsigner/store.go:358`. Not requesting changes, just bringing it to attention.

lrsaturnino (commented):
Just a double-check: Is this PR intended to merge with `fix/review-findings` or `main`?

piotr-roslaniec:
@lrsaturnino It should target `main` since the previous branch was already merged. This PR is probably out of date, I will look into updating it. Thanks for the review

lrsaturnino (approved):

PR #3866, opened 2026-02-19
feat(tbtc/node): FROST/ROAST Go node — distributed DKG, interactive signing loop, Taproot wallet lifecycle

mswilkison:
## What

This is the Go node/coordinator side of the tBTC FROST/Schnorr migration. It teaches keep-core to run threshold-Schnorr (Taproot) wallets end to end: a real **distributed FROST DKG** with the full on-chain submit → challenge → approve result lifecycle, an **interactive ROAST signing pipeline wired into the production signing loop**, and Taproot-native funds movement (deposit sweep, redemption, moving funds) with the matching Bitcoin, SPV-maintainer, and proposal-generator plumbing. Moving tBTC wallets from tECDSA to FROST gives key-path-only P2TR wallet outputs and, via ROAST, signing that makes progress despite unresponsive or misbehaving signers while producing attributable blame evidence.

The cryptographic engine lives in the companion Rust signer (`libfrost_tbtc`, #4005) and is linked over a cgo FFI bridge behind build tags; a default (untagged) build compiles to today's legacy tECDSA behavior with the FROST code inert. Interactive signing (the node driving per-attempt commit → sign → aggregate rounds through the ROAST coordinator) is implemented but **opt-in**: it is gated behind default-off env switches, so the shipped default remains the transitional one-shot ("coarse") path until the engine audit clears — see **Activation state** below. The FFI bridge requires signer FFI contract **major 3 exactly, minor ≥ 3**, failing closed on any mismatch. Wallet key generation is a distributed FROST DKG (round 1, sealed round-2 transport, finalize, engine-persisted key packages); there is no trusted-dealer path.

The node addresses FROST wallets by canonical `bytes32` wallet ID (with legacy 20-byte pubkey-hash compatibility mapping), builds and signs BIP-341 key-path spends for wallet transactions, and coordinates with the `FrostWalletRegistry` / `FrostDkgValidator` contracts from the tbtc-v2 side (#971).

## Scope

**ROAST coordination (`pkg/frost/roast` — pure Go, no cgo)**
- Coordinator state machine with deterministic, seed-derived participant selection (seed bridge, shuffle corpus + seed-vector fixtures) and retry adapter into the existing retry loop.
- Authenticated wire messages (protobuf: signing package, share submission, evidence) with membership pre-auth, sender binding, and pre-allocation byte caps on unmarshal.
[...]

PR #3962, opened 2026-05-22
docs(rfc): RFC-21 ROAST coordinator, retry, and transition evidence

mswilkison:
## Summary

Adds **RFC-21** as the design doc that scopes the M4 (transition
evidence) and M7 (ROAST-aware retry) findings from the independent
review of #3866 into a single layered design and a phased,
PR-sized implementation plan.

This PR is **doc-only**. It introduces no behaviour change. Subsequent
implementation PRs reference RFC-21 in their descriptions.

Stacked on #3961.

## Why one design, not two

M4 and M7 share the same notion of *attempt context* and *transition
evidence*:

- Fixing M4 alone produces evidence that no consumer reads.
- Fixing M7 alone gives the consumer nothing to drive retry decisions on.

The RFC treats them as one design split into linear phases.

## Phasing

- **Phase 0** -- this RFC.
- **Phase 1** -- `AttemptContext` type + canonical hash; protocol
  messages carry attempt-context binding (optional during migration).
- **Phase 2** -- receiver overflow tracking (M4 layer A) plumbed
  through the three `select { default }` drop sites, default no-op.
- **Phase 3** -- coordinator state machine: `BeginAttempt`,
  `RecordEvidence`, `NextAttempt`. Deterministic
  `(AttemptContext, TransitionEvidence) -> AttemptContext` map.
- **Phase 4** -- wire receiver to coordinator behind
  `frost_roast_retry` build tag.
- **Phase 5** -- retry adapter +
  `EvaluateRoastRetryForSigning`; migrate first call site behind
  the build tag with readiness-gate guard.
- **Phase 6** -- migrate remaining call sites; delete the
  byte-identical-to-tECDSA shuffle once unused.
- **Phase 7** -- flip the readiness manifest to `present` once Phase
  6 ships and integration tests run against a real testnet (only
  then; no early flip).

## Open questions called out explicitly

The RFC lists four open design questions that need cross-team
review before Phase 3 lands:

1. Cross-process coordinator agreement -- gossip topic choice.
2. Persistence across signer restart.
3. FFI surface (Rust signer error-code style; follows the L5
   pattern from #425 / #3961).
4. Backward-compat horizon for the `AttemptContextHash` field.

## Out of scope

- DKG retry (separate RFC).
- Bitcoin transaction-builder changes.
- Operator UX changes (CLI, dashboards) -- land alongside Phase 5/6.
- Cross-domain ROAST between keep-core and tbtc-signer.

## Test plan

- [ ] Reviewer reads RFC end-to-end.
- [ ] Reviewer flags any phase that should be split further or
  reordered before Phase 1 begins.
- [ ] Reviewer answers the four open questions or marks them
  defer-to-Phase-3.

No code change in this PR, so no CI test run is meaningful beyond
asciidoc rendering.

### DIFF
--- a/docs/rfc/rfc-21-roast-coordinator-retry-and-transition-evidence.adoc
+++ b/docs/rfc/rfc-21-roast-coordinator-retry-and-transition-evidence.adoc
@@ -170,7 +170,7 @@ type AttemptContext struct {
     AttemptNumber  uint
     IncludedSet    []group.MemberIndex
     ExcludedSet    []group.MemberIndex
-    AttemptSeed    int64
+    AttemptSeed    [32]byte
 }
 
 func (AttemptContext) Hash() [32]byte
@@ -181,6 +181,22 @@ message emitted in the attempt -- contribution messages already carry a
 `SessionIDValue`; we extend them with an `AttemptContextHash` field so
 the receiver can reject stale-attempt messages structurally instead of
 relying on session ID alone.
+
+`AttemptSeed` is widened from `int64` to `[32]byte` and *must* be
+derived from inputs the group already agrees on -- specifically:
+
+[source,go]
+----
+AttemptSeed = SHA256(
+    DkgGroupPublicKey || SessionID || MessageDigest,
+)
+----
+
+This binding prevents a malicious coordinator from picking a seed that
+shapes the included set in its favour. The seed is a pure function of
+session inputs; it is never chosen, only derived. Any signer can
+recompute it from the session header and verify the coordinator's
+participant selection.
 
 === Layer A: Receiver transition evidence (M4)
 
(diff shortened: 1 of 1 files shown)


mswilkison:
## Second-pass review feedback folded in

Thanks for the substantive review. Five concrete recommendations
addressed in commit f2f1e99c5; no phase or scope changes.

### Changes mapped to recommendations

| Recommendation | Where | Resolution |
|---|---|---|
| Bind `AttemptSeed` to agreed-upon inputs so a coordinator cannot manipulate participant selection | Shared types section | Widened to `[32]byte`; specified `AttemptSeed = SHA256(DkgGroupPublicKey \|\| SessionID \|\| MessageDigest)`. Pure function of session inputs, never chosen. |
| Resolve cross-process evidence agreement before Phase 3 | Open question 1 | Elevated *signed-evidence gossip on a dedicated topic* as the recommended path entering Phase 3. Open question remains open pending protocol-owner sign-off; recommendation now explicit. Spelled out why deterministic `NextAttempt` without input agreement produces divergent outputs. |
| Per-category quotas in the evidence recorder | Layer A | Replaced single ring with `categoryQuota` struct: overflow=8, reject=8, conflict=4, silence=1. Peer cannot spam one category to mask another. Bound is now `O(\|IncludedSet\| * sum(quotas))`. |
| Define exclusion thresholds concretely | Layer B | Specified as constants: `overflowExclusionThreshold = 4`, `rejectExclusionThreshold = 1`, `silenceParkingThreshold = 1`. Explained why constants (no runtime drift) beat parameters. |
| Persistence sketch via gossip rather than on-disk state | Open question 2 | Added `SyncState` gossip message sketch for Phase 5+: restarting node broadcasts `(LastKnownAttemptContextHash, KeyGroupID)`, peers reply with their current attempt + signed attestations. Gossiped attestations *are* the persistent record. |

### Still open

Both open questions stay open. The recommendations above are the
preferred direction *entering* Phase 3 design review; they are not
yet decisions. Phase 3 will not start until protocol owners confirm
or redirect.

RFC now 498 lines (was 419). Diff is +97/-18.

PR #3997, opened 2026-05-24
chore(ethereum/test): read mainnet RPC URL from env, drop hardcoded key

piotr-roslaniec:
## Summary

- `pkg/chain/ethereum/ethereum_integration_test.go` contained a hardcoded `https://mainnet.infura.io/v3/<key>` URL with an embedded Infura API key. The key is in git history and on the public mirror, so it must be treated as **compromised and rotated** on the Infura side independently of this PR.
- This change makes the test read the JSON-RPC endpoint from a `KEEP_TEST_ETHEREUM_URL` env variable instead, and skip when the variable is not set. That removes the leak from the source tree and avoids relying on a single shared credential.

## Notes

- This test is already failing on `main` for a separate reason: mainnet blocks now contain EIP-7702 type-4 transactions that `go-ethereum v1.13.x` cannot decode (`cannot get current block: [transaction type not supported]`). That's tracked separately and requires a `go-ethereum` bump.
- The skip behaviour here means CI will no longer spuriously fail when no URL is configured.

## Follow-ups

- [ ] Rotate the leaked Infura key on the provider side
- [ ] Wire `KEEP_TEST_ETHEREUM_URL` into the relevant CI workflow if/when integration tests should run there
- [ ] Bump `go-ethereum` to a version with EIP-7702 support to actually unblock the test against current mainnet

## Test plan

- [x] `go test -tags=integration -c -o /dev/null ./pkg/chain/ethereum/` compiles
- [x] Test is skipped (not failed) when `KEEP_TEST_ETHEREUM_URL` is unset

## Summary by CodeRabbit

* **Documentation**
  * Updated Ethereum integration test instructions to identify the previously included provider URL/key as compromised.
  * Added guidance to rotate the provider key before reuse.
  * Clarified that the test must run against Ethereum mainnet, not testnets, because it validates known mainnet block numbers.


lrsaturnino:
2 findings addressed — 1 fixed, 1 requires manual follow-up.

**Fixed**
- PR was unmergeable against current `main`; conflict-resolution risk was overstated. Merged `origin/main` (`6f87e20b9`) and resolved the single real conflict in `pkg/chain/ethereum/ethereum_integration_test.go` by keeping main's more complete fix (env-var-driven skip logic plus the 429-retry skip helper, untouched by the conflict) and folding in this branch's value-add: a doc comment noting the RPC URL must point to mainnet and that the previously hardcoded key must be treated as compromised. Build, `go vet`, and `gofmt` pass; branch is now mergeable with zero remaining conflicts. Commit `243e6f239`.

**Unresolved — needs manual follow-up**
- Infura key rotation was never completed; the leaked key is still live. The code-level fix is complete: no hardcoded key remains anywhere in the tree, the test reads `ETHEREUM_MAINNET_RPC_URL` from the environment, and the doc comment flags the prior key as compromised. What remains — rotating/revoking the key in Infura's dashboard — is a provider-side action outside this repo. Critical severity; needs a maintainer with Infura dashboard access to action directly.

lrsaturnino (approved):
1 finding fixed and pushed; 1 requires manual follow-up (see comment above).

PR #3999, opened 2026-05-25
chore(ci): align Yarn 4 workflows and solidity workspace tooling

lionakhnazarov:
## Stack Context
This is **PR 1/4** in the reorganization stack.
Base: `main`
Next: #4000

### 1) CI + workflow execution model
- Added reusable Git/Yarn setup action: `.github/actions/setup-git-for-yarn/action.yml`.
- Added local reusable docs workflow: `.github/workflows/reusable-solidity-docs.yml`.
- Updated Solidity workflows to use the new setup and Yarn/Corepack-compatible behavior:
  - `.github/workflows/contracts-ecdsa.yml`
  - `.github/workflows/contracts-random-beacon.yml`
  - `.github/workflows/contracts-ecdsa-docs.yml`
  - `.github/workflows/contracts-random-beacon-docs.yml`
  - `.github/workflows/npm-ecdsa.yml`
  - `.github/workflows/npm-random-beacon.yml`

### 2) Solidity workspace tooling alignment
- Added Yarn 4 config files:
  - `solidity/ecdsa/.yarnrc.yml`
  - `solidity/random-beacon/.yarnrc.yml`
- Updated build/tooling ignore and container files:
  - `solidity/ecdsa/.dockerignore`, `.eslintignore`, `.gitignore`, `.prettierignore`, `Dockerfile`
  - `solidity/random-beacon/.dockerignore`, `.gitignore`, `.prettierignore`, `Dockerfile`

### 3) Dependency metadata and lockfiles
- Updated manifests/config:
  - `solidity/ecdsa/package.json`, `solidity/ecdsa/tsconfig.json`
  - `solidity/random-beacon/package.json`, `solidity/random-beacon/tsconfig.json`
- Refreshed lockfiles:
  - `solidity/ecdsa/yarn.lock`
  - `solidity/random-beacon/yarn.lock`

## Out of Scope
- Solidity deploy logic and contract behavior updates.
- Generated Sepolia deployment artifacts.
- Go/Testnet4 runtime changes.

## Test Plan
- [ ] Workflow YAML is valid and jobs trigger.
- [ ] Solidity workspaces install with Yarn/Corepack in CI-compatible mode.

## Summary by CodeRabbit

* **New Features**
  * Added a reusable workflow to generate and optionally publish Solidity API documentation with configurable publish and artifact options.

* **Chores**
  * Introduced shared CI actions to standardize Node/Yarn dependency installation, git handling for yarn, and caching across workflows.
[...]

PR #4015, opened 2026-06-04
chore(ci): apply review fixes on top of #3999

piotr-roslaniec:
## Summary

Follow-up to #3999. Applies the multi-agent / validate-findings review
recommendations that were confirmed valid on top of that PR's Yarn 4
migration. Keeps #3999 surgical and lets each fix stand on its own.

## Commits

1. `fix(ci): scope GIT_CONFIG vars to wrapper subshell only` (C1)
   `setup-git-for-yarn` was exporting `GIT_CONFIG_GLOBAL=/dev/null` to
   `$GITHUB_ENV`, breaking downstream `git config --global` calls (e.g.
   docs publish `user.email/name`). Scope GIT_CONFIG vars inside the
   wrapper subshell only.
2. `refactor(ci): extract Yarn install into composite action` (S1)
   ~22-line install bash was duplicated 13 times across workflows.
   New `.github/actions/install-yarn-deps` action takes `working-directory`
   and centralises the install incantation. Net diff: -351 / +42.
3. `ci: restore Yarn dependency caching via actions/cache` (C4)
   `cache: "yarn"` was removed from `setup-node` with no replacement,
   forcing every CI job to re-download deps. Add `actions/cache@v4`
   keyed on yarn.lock for `.yarn/cache`, `.yarn/install-state.gz`,
   and `node_modules`.
4. `fix(ci): re-enable Yarn lockfile integrity checks` + `fix(ci): keep HARDENED_MODE disabled, documented` (Sec1)
   Removed `YARN_CHECKSUM_BEHAVIOR=ignore` (lockfile checksums verified
   locally as consistent). Kept `YARN_ENABLE_HARDENED_MODE=0` with a
   documented reason: Yarn 4 auto-enables hardened mode in public PR
   context, and ECDSA's lockfile contains a legitimate npm-to-git remap
   for `ethereumjs-abi` (the npm 0.6.8 release was published broken;
   the git pin is canonical). Hardened mode rejects all such remaps
   with `YN0078`, regardless of intent.
5. `fix(workflows): replace unmaintained hub with gh pr create` (C6)
   `hub` was removed from `ubuntu-22.04` runner images in 2024 and is
   not present on `ubuntu-24.04`. Replace with the preinstalled `gh`.
6. `fix(workflows): use gh credential helper instead of token-in-URL` (Sec3)
   `git clone https://user:$TOKEN@github.com/...` persisted the token
   in `remote.origin.url`. Configure `gh auth setup-git` instead;
   token stays in env, out of `.git/config`.
7. `ci(workflows): pin upload-artifact to v4` (Sec4)
   `actions/upload-artifact@master` was unpinned. Pin to `@v4`.
8. `ci(workflows): bump checkout and setup-node to v4 in new docs workflow` (C5)
   Surgical version bump on the newly-introduced
   `reusable-solidity-docs.yml` only; pre-existing v3 usage elsewhere
   left alone.

## Notes

- Random Beacon `build-and-test` / `deployment-dry-run` and ECDSA
  `build-and-test` failures with errors like
  `Property 'pushNotificationReward' does not exist on type 'TokenStaking'`
  and `No method named "approveApplication"` are **pre-existing** in
  #3999 (verified against its prior commit's run log). They are not
  introduced by this PR and are out of scope here.
- `setup-node` is left on v3 in `contracts-*.yml` and `npm-*.yml` to
  match repo convention. Bumping there can be a follow-up.

## Test plan

- [ ] CI install/lint/slither/docs jobs green
- [ ] Sec3 / Sec4 / C5 / C6 only affect docs workflow (not exercised on
      normal PRs unless `publish == true`); verify by inspection

### DIFF
--- a/.github/actions/install-yarn-deps/action.yml
+++ b/.github/actions/install-yarn-deps/action.yml
@@ -1,0 +1,37 @@
+name: Install Yarn dependencies (Yarn 4 / Corepack)
+description: >
+  Enable Corepack, set up the Git-for-Yarn wrapper, and run
+  `yarn install --immutable` in the given workspace. Centralises the
+  install incantation so each workflow does not need to re-export PATH,
+  GIT_CONFIG_*, or pin the Corepack-managed Yarn version.
+inputs:
+  working-directory:
+    description: "Workspace directory containing package.json and yarn.lock."
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Set up Git wrapper for Yarn git dependencies
+      uses: ./.github/actions/setup-git-for-yarn
+
+    - name: Install dependencies
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      run: |
+        set -euo pipefail
+        export PATH="${RUNNER_TEMP:-/tmp}/git-clean-bin:$PATH"
+        export YARN_ENABLE_HARDENED_MODE=0
+        export YARN_CHECKSUM_BEHAVIOR=ignore
+        corepack enable
+        yarn --version
+        unset_args=()
+        while IFS='=' read -r key _; do
+          case "$key" in
+            GIT_CONFIG_*) unset_args+=("-u" "$key") ;;
+          esac
+        done < <(env)
+        env "${unset_args[@]}" \
+          GIT_CONFIG_NOSYSTEM=1 \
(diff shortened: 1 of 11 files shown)


lrsaturnino (approved):
LGTM

PR #4030, opened 2026-06-11
spec(frost/roast): normative coordinator-seed derivation (RFC-21 Annex A) + cross-language conformance vectors

mswilkison:
Stacked on #3866 (base: `feat/frost-schnorr-migration-scaffold`). Implements item 3 of the review feedback (duplicated, divergent protocol constants) — Go half; the Rust half is the paired PR stacked on #4005.

## Problem

The coordinator-shuffle seed derivation exists twice, in two languages, on two branches, with no single source of truth — and the two copies disagree (flagged in #4026):

| | seed | attempt numbering |
|---|---|---|
| Go RFC-21 layer | `fold(SHA256(KeyGroup ‖ SessionID ‖ MessageDigest))` | 0-based |
| Rust engine validation | `int64_be(MessageDigest[0..8])` (legacy `signingAttemptSeed` convention) | 1-based wire |

At Phase-7 wiring, every Go-derived attempt context would fail the Rust engine's strict-mode validation — a network-fracturing liveness failure that property tests on either side cannot catch.

## What this PR does (Go half)

1. **RFC-21 Annex A (normative)** — single normative definition of the derivation: inputs (including the exact `KeyGroupBytes` definition for `FrostTBTCSignerV1` material — the UTF-8 bytes of the hex key-group handle, treated opaquely), the 0-based composition with the two's-complement-wrapping addition, the `wire = AttemptNumber + 1` FFI mapping, and the accepted non-goals (unframed concatenation, first-8-byte fold, grindability bounds) with rationale. The Go derivation is adopted as normative: it binds key group + session + digest rather than the digest alone, and the live `pkg/tbtc` signing loop's legacy convention is explicitly documented as the thing Phase 7 migrates *from*.

2. **Generated conformance vectors** — `pkg/frost/roast/testdata/coordinator_seed_vectors.json`: ten end-to-end vectors (folded seed int64 + selected coordinator) covering attempts 0/1/3/5/7, sparse and production-size (n=100) member sets, opaque key-group handles, and negative folded seeds. Regenerated from the deterministic input matrix via `ROAST_SEED_VECTORS_REGEN=1 go test -run TestRegenerateCoordinatorSeedVectors` — generation-from-spec rather than hand-pinning, per the review.

3. **Conformance test** — `TestCoordinatorSeedDerivation_ConformanceVectors` pins `DeriveAttemptSeed → foldAttemptSeed → SelectCoordinator` end to end against the file, asserts the wire-mapping invariant on every vector, and requires at least one negative-seed pin so an unsigned-integer port cannot pass.

The paired Rust PR switches the engine to this derivation (subtracting 1 from the wire attempt number before composition) and consumes a byte-identical copy of the vector file, so either side drifting fails its own CI rather than fracturing coordinator agreement in a mixed deployment.

No behavior change on the Go side — it was already normative-conformant; this PR makes that the *specified* behavior and pins it.

## Tests

`go test ./pkg/frost/...` passes; vectors verified present with 7 negative-seed pins out of 10.


mswilkison:
Annex A amended (latest commit): `MessageDigest` is now specified unambiguously as the **raw signing message big-endian left-padded to 32 bytes** (keep-core's `messageDigestFromBigInt` output; >32 significant bytes rejected) — never a transcript hash of the message. This closes the input-layer ambiguity behind the divergence Codex caught on the paired Rust PR #4031, where the engine seeded from its internal `SHA256(message)` transcript digest. The engine-side contract is pinned by `start_sign_round_accepts_go_derived_attempt_context_in_strict_mode` on #4031.

PR #4039, opened 2026-06-12
fix(spv): compute required proof headers from actual header difficulties

lionakhnazarov:
# fix(spv): compute required proof headers from actual header difficulties

## Problem

The SPV maintainer computed the number of block headers for a proof assuming every header carries the relay epoch difficulty: with `txProofDifficultyFactor = 1` it always assembled a **single-header proof**. On testnet4 (BIP94), sweep transactions are regularly mined in minimum-difficulty (DIFF1) blocks, so the proof contained only a DIFF1 header. The Bridge skips DIFF1 headers when binding to the relay difficulty, finds no decisive header, and reverts with:

```
execution reverted: Not at current or previous difficulty
```

Because the error is deterministic, the maintainer crash-looped on the same transaction every cycle (`error while maintaining SPV: ... restarting maintainer`), and the affected wallet's main UTXO could never be registered on the host chain — blocking all subsequent sweeps for that wallet.

## Fix

`getProofInfo` now mirrors the Bridge's `BitcoinTx.determineRequestedDifficulty` / `evaluateProofDifficulty` logic instead of assuming a fixed per-header difficulty:

- walk headers forward from the proof start block,
- skip leading minimum-difficulty headers, but only when both relay epoch difficulties are above minimum (so test/dev relays at difficulty 1 keep working),
- bind the requested difficulty to the first decisive header matching the relay's **current or previous** epoch difficulty; if it matches neither, skip the transaction (the Bridge would revert),
- accumulate headers (DIFF1 headers still contribute their work) until total observed difficulty reaches `requestedDifficulty × txProofDifficultyFactor`,
- bound the walk by `maxProofHeaders = 144` and report "wait for more confirmations" when the chain tip is reached first.

The old epoch-boundary block-count arithmetic is removed; epoch spans are handled implicitly by summing actual header difficulties.

## Testing

- `TestGetProofInfo` rewritten for the new semantics: current/previous-epoch proofs, epoch-spanning proofs with difficulty drop/raise, leading DIFF1 headers, relay at minimum difficulty, decisive-header mismatch, DIFF1 run exceeding the header bound, and chain tip reached early. Headers are generated with real compact-bits encoding (`blockHeaderWithDifficulty`).
- `go test ./pkg/maintainer/spv` passes.
- Verified live on the Sepolia/testnet4 stack: a sweep stuck for 10 days (`876510fd...62c8`, mined in a DIFF1 block at height 137542) was proven on the first cycle after deploying this fix — `submitDepositSweepProof` tx `0xfeeefad32c7116727949483edfb1e38f8117d923521fd3467886a05226b82577`, status 1 — and the maintainer completed its round cleanly.

## Notes

- Requires a Bridge with DIFF1-aware `determineRequestedDifficulty` (deployed via the DIFF1 library upgrade). This changes maintainer behavior only for transactions whose proof would begin with minimum-difficulty (DIFF1) headers: those are now skipped rather than the maintainer deterministically failing on them, and a leading DIFF1 run exceeding `maxProofHeaders` is reported with a distinct log and metric. Transactions whose proof begins with a real-difficulty header are handled as before.

### DIFF
--- a/pkg/maintainer/spv/bitcoin_chain_test.go
+++ b/pkg/maintainer/spv/bitcoin_chain_test.go
@@ -3,10 +3,42 @@ package spv
 import (
 	"bytes"
 	"fmt"
+	"math/big"
 	"sync"
 
+	"github.com/btcsuite/btcd/blockchain"
 	"github.com/keep-network/keep-core/pkg/bitcoin"
 )
+
+// populateBlockHeaders adds headers for [fromHeight, toHeight] inclusive using
+// difficultyAt(height) for each block's Bits-derived difficulty.
+func populateBlockHeaders(
+	lbc *localBitcoinChain,
+	fromHeight, toHeight uint,
+	difficultyAt func(uint) *big.Int,
+) error {
+	for h := fromHeight; h <= toHeight; h++ {
+		header := blockHeaderWithDifficulty(difficultyAt(h))
+		if err := lbc.addBlockHeader(h, header); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// blockHeaderWithDifficulty returns a header whose Difficulty() matches the
+// given value (within Bitcoin compact encoding precision). Powers of two and
+// small values round-trip exactly.
+func blockHeaderWithDifficulty(difficulty *big.Int) *bitcoin.BlockHeader {
+	maxTarget := new(big.Int)
+	maxTarget.SetString(
+		"ffff0000000000000000000000000000000000000000000000000000",
+		16,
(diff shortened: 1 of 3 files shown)


lrsaturnino (approved):

piotr-roslaniec:
Follow-up: a multi-agent review of this change surfaced some edge-case, observability, and test-quality findings. Since this PR is already merged, they are addressed in a separate follow-up PR: #4186 (`fix(spv): harden DIFF1 proof-header computation`). The core algorithm here is unchanged.

PR #4041, opened 2026-06-12
feat(frost/signing): adopt the tbtc-signer init-time config FFI

mswilkison:
Go-host adoption of `frost_tbtc_init_signer_config` — the companion to keep-core#4037 (which added the init-time config FFI to the Rust signer on the mirror branch). Hosts can now install the signer's operational configuration once at startup instead of exporting ~40 `TBTC_SIGNER_*` process environment variables.

## What this adds

- **`InstallNativeTBTCSignerConfig(configJSON)`** — passes operator JSON through verbatim; the Rust signer owns the schema and all validation (unknown fields rejected, enforcement-gated policy combinations validated at install with rollback, environment ignored wholesale once installed, idempotent re-install by fingerprint). Go stays a transport: no schema duplication, no drift surface.
- **`TBTC_SIGNER_INIT_CONFIG_PATH`** — when set, the JSON file is installed during native FROST engine registration, *before* any other signer call. Setting the path is an explicit demand for config-mode operation, so every failure — unreadable file, validation rejection, or a loaded signer library that predates the symbol — **fails registration closed**. Unset, registration proceeds on the environment-fallback path exactly as today (zero behavior change for existing deployments). Net ops effect: one pointer variable replaces forty value variables; secrets stay on the dedicated env/command key-provider channel.

## Cross-branch compatibility (why this is safe to land now)

The bridge resolves every symbol per call via `dlsym(RTLD_DEFAULT, …)` with graceful degradation — the established pattern in this file. A signer library **without** the new symbol returns the existing `ErrNativeCryptographyUnavailable` classification rather than failing the link, so:

- path unset + old library → status quo;
- path unset + new library → status quo (env fallback, with the signer's own one-time production warning suggesting the init FFI);
- path set + new library → config installed, logged with fingerprint/key-count/idempotency;
- path set + old library → registration fails closed with a precise error (operator demanded config mode; the deployment is wrong).

## Build-shape verification

Three configurations verified: default tags (stub + unit test asserting the unavailable classification), `frost_native` alone (the registration wiring compiles against the stub), and `frost_native && frost_tbtc_signer && cgo` (real cgo bridge, compile + vet — running bridge tests requires the built signer library, which lives on the mirror branch until promotion). `gofmt`, `go vet`, and the full `pkg/frost` test suite are clean under default tags.

## Out of scope

keep-client TOML config plumbing (a first-class config section feeding this) — deliberately left for the team's config-schema conventions; the file-pointer wiring gives ops a complete adoption path today without touching the client config surface.

### DIFF
--- a/pkg/frost/signing/native_ffi_primitive_transitional_frost_native.go
+++ b/pkg/frost/signing/native_ffi_primitive_transitional_frost_native.go
@@ -44,9 +44,14 @@ func defaultNativeExecutionFFISigningPrimitiveProviderForBuild() (
 // file. The operator setting the path is an explicit demand for config-mode
 // operation, so every failure - unreadable file, validation rejection, or a
 // loaded signer library that predates frost_tbtc_init_signer_config - fails
-// the native engine registration closed. With the path unset this is a
-// no-op and the signer reads TBTC_SIGNER_* from the process environment
-// (the transitional path).
+// the FROST-native engine registration. Precisely: the process keeps
+// running on the legacy bridge with FROST operations reporting unavailable
+// (the registration layer's deliberate safe-by-default posture - it never
+// crashes the binary), a misconfigured signer can therefore never execute
+// FROST operations, and the failure is logged at error level here with the
+// config-mode context in addition to the registration layer's generic
+// warning. With the path unset this is a no-op and the signer reads
+// TBTC_SIGNER_* from the process environment (the transitional path).
 func installConfiguredTBTCSignerInitConfig() error {
 	configPath := strings.TrimSpace(os.Getenv(TBTCSignerInitConfigPathEnv))
 	if configPath == "" {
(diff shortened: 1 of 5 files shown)


mswilkison:
**Review follow-ups addressed in `8cb12a02f`** (Codex and Gemini passes were clean; my own findings):

1. **Failure loudness + precise semantics.** I traced the set-path failure end to end: provider error → generic `Warnf` in the registration layer → `LastNativeRegistrationError()` → process continues on the legacy bridge. The security property always held (a misconfigured signer cannot execute FROST operations), but one generic warn line is a weak signal for an operator who explicitly demanded config mode. `installConfiguredTBTCSignerInitConfig` now logs at **error** level with config-mode context on every failure path, and its doc states the exact degradation: registration fails closed, the binary keeps running on the legacy bridge with FROST unavailable — it does not exit.
2. **Config-file permissions note** on `TBTCSignerInitConfigPathEnv`: the file may carry the `state_key_command` execution spec, so restrict it to the operator account (e.g. `0600`).

**Open question for the team (deliberately not changed here):** should a *set-path* install failure be **process-fatal** in production deployments? Today's posture follows the registration layer's explicit "never crash the binary" contract — rejected config means legacy-bridge operation with FROST unavailable plus an error log. There's a reasonable argument that an operator who demanded config mode would rather the process exit than run degraded. Flipping that changes the registration contract, so it's a team call; happy to implement either way.

Also re-verified after the change: all four build shapes compile (default, `frost_native`, full cgo bridge, and bridge tags with `CGO_ENABLED=0`), default-tag tests pass uncached.

mswilkison:
**Resolving the open team question** ("should set-path install failure be process-fatal in production?"):

**Decided (2026-06-12, MacLane): yes — fatal, and broader than the question as posed.** Whenever `TBTC_SIGNER_INIT_CONFIG_PATH` is set, any state in which the FROST-native engine does not come up is process-fatal in **every profile and build flavor** — covering the config-install leg, the engine-registration leg after a successful install, and a binary built without `frost_native`. Env-fallback mode (path unset) keeps this PR's safe-by-default degrade posture unchanged.

Why fatal: this branch ships to production only when FROST is a production duty, so "running but FROST-dead" is the dangerous silent state — it erodes FROST wallet fault budgets invisibly, while threshold redundancy is designed to absorb loud, full, bounded outages. Why not profile-conditional: an unreadable config file cannot reveal its profile, and production-by-omission means absence must be treated as production — path-set is the only non-circular trigger. It also means testnet rehearses exactly the semantics production will have.

Implementation: #4045 (scaffold). Decision record + operational consequences (canary config pushes node-by-node; attestation rotation becomes enforced-by-restart-failure): #4048 (gates-doc Decision Log entry 7 + runbook prerequisite 7, mirror).

PR #4043, opened 2026-06-12
docs(tbtc/signer): record 2026-06-12 architecture decisions in gates doc

mswilkison:
Records MacLane's decisions on the follow-up checklist's open questions: audit = hard gate for ECDSA retirement; sidecar over in-process cgo; script-tree vs timelock explicitly open; item 7 deferred with a binding evidence-retention condition; item 6 scheduled as the first Phase 7 item (with the architectural reason: transitional shares are bound to the full included set at StartSignRound, so first-t-responsive requires the interactive exchange that is Phase 7 itself).

### DIFF
--- a/pkg/tbtc/signer/docs/roast-phase-5-security-rollout-gates.md
+++ b/pkg/tbtc/signer/docs/roast-phase-5-security-rollout-gates.md
@@ -115,6 +115,26 @@ architecture questions:
    responsive subset is known - the interactive two-round exchange
    that IS Phase 7's core. Pulling it earlier would implement the
    interactive path without its Go-side consumer.
+6. **Transitional deterministic-nonce path: committed for DELETION.**
+   The path is already production-gated (production signing is
+   interactive-FROST-only with OS randomness), so it serves
+   dev/staging only - while its nonce safety rests on the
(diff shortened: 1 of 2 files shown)


mswilkison:
**Codex + Gemini P1 (both identical) — valid, fixed in `6b74e5d39`.** The parenthetical asserted evidence retention "is implemented in the Go RFC-21 layer", but the base branch only detects the conflict and returns ErrSnapshotConflict, dropping the envelope — retention lives in the still-open PR #4044. Since item 7's deferral is *conditioned* on retention existing, that wording created exactly the false sense of diagnosability they describe.

Reworded so the decision log is honest about sequencing: the deferral is now **explicitly contingent** on retention landing; retention is attributed to PR #4044 (scaffold branch); and the base layer's drop-the-envelope behavior until that merges is stated plainly, so the deferral does not read as already in force. Merge ordering is now self-documenting — if this doc lands before #4044, it correctly describes the gap rather than papering over it.

The reviewers' other verifications (Phase 7 / t-of-included rationale, the nonce-path freeze, audit-gate phrasing, recovery-leaf openness) came back clean — no changes there.

PR #4044, opened 2026-06-12
feat(frost/roast): retain signed equivocation evidence at detection points

mswilkison:
Implements the binding retention condition from today's decision log (PR #4043): proof-carrying blame (follow-up item 7) is deferred until production, **provided** telemetry/logging retain enough signed bytes to diagnose whether targeted equivocation is occurring — otherwise the revisit condition lacks data.

## What this adds

`EquivocationEvidence` events carrying the **exact signed snapshot envelopes** (wire bytes verbatim — the #4040 format makes these available at every detection point) for the three detections that exist today:

- `snapshot_conflict` — a sender re-submits a *different* signed snapshot for the same attempt to the coordinator. Both envelopes are retained; two operator-signed bodies from the same sender for the same attempt are self-incriminating, which is exactly the substrate item 7's wire format will formalize.
- `own_snapshot_mutated_in_bundle` — a bundle carries this member's snapshot with a signature that differs from what it submitted (both envelopes retained).
- `own_snapshot_missing_from_bundle` — censorship detection (self envelope retained).

Each event is logged in full (these are rare, and the bytes are the diagnosis) and forwarded to a process-wide observer hook following the repo's existing single-observer telemetry pattern, so hosts can persist evidence into their telemetry stack. Emission is purely additive on the existing error paths — encode failures degrade to nil fields with a log line, never perturbing the protocol path.

## Deliberately out of scope

Cross-member comparison (a receiver checking a bundle's snapshot for sender X against X's direct broadcast) — that's item 7 proper. These are the detection points that exist today, instrumented so the production deferral is honest.

Tests pin byte-exact envelope retention for all three kinds, and that an idempotent identical re-submission emits nothing. `go build ./...`, vet, gofmt clean; full frost suite green.

### DIFF
--- a/pkg/frost/roast/coordinator_state.go
+++ b/pkg/frost/roast/coordinator_state.go
@@ -358,6 +358,13 @@ func (c *inMemoryCoordinator) RecordEvidence(
 		}
 		if !bytes.Equal(existingBytes, newBytes) ||
 			!bytes.Equal(existing.OperatorSignature, snapshot.OperatorSignature) {
+			emitEquivocationEvidence(EquivocationEvidence{
+				Kind:                EquivocationKindSnapshotConflict,
+				AttemptContextHash:  append([]byte(nil), snapshot.AttemptContextHash...),
+				Sender:              snapshot.SenderID(),
+				ExistingEnvelope:    snapshotEnvelopeForEvidence(existing),
+				ConflictingEnvelope: snapshotEnvelopeForEvidence(snapshot),
+			})
 			return ErrSnapshotConflict
 		}
 		// Identical re-submission: idempotent no-op.
--- a/pkg/frost/roast/coordinator_state.go
+++ b/pkg/frost/roast/coordinator_state.go
@@ -326,6 +326,18 @@ func (c *inMemoryCoordinator) RecordEvidence(
 		return fmt.Errorf("coordinator: %w", err)
 	}
 
+	// Emit any equivocation evidence AFTER c.mu is released: a registered
+	// observer is host telemetry (possibly a blocking write) and must not
+	// run while the coordinator state mutex is held. The evidence value is
+	// fully materialized (bytes copied) under the lock; emission only reads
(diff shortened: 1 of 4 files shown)


mswilkison:
**Self-review finding fixed in `156cda945`:** the `snapshot_conflict` emit ran while the coordinator's `c.mu` was held, so a registered observer (host telemetry, possibly blocking) would have stalled every concurrent `RecordEvidence`/`AggregateBundle` on that coordinator. The evidence value is now materialized under the lock (bytes copied as before) into a local, and a deferred closure — registered before the unlock defer so it runs *after* it — emits once `c.mu` is released. The other two emit sites were already lock-free. Verified under `go test -race`; the byte-exact retention tests still pass, confirming the deferred emit fires with the correct envelopes.

mswilkison:
**Both Codex re-review P2s fixed in `5fa1c19cd`** — both valid:

1. **Observer panic could escape the protocol path** — directly contradicting `emitEquivocationEvidence`'s own "never fails" doc. The observer call is now wrapped in recover-and-log, so a panicking telemetry observer can't abort `RecordEvidence`/`verifyOwnObservationsPresent`. Regression test: a panicking observer still yields `ErrSnapshotConflict` from the protocol path.

2. **Evidence bytes aliased the `Marshal()` cache** — and this is precisely the must-not-mutate contract this stack pinned in the #4040 envelope work; my own evidence code then violated it by handing the cached `wireEnvelope` slice to an external observer. `snapshotEnvelopeForEvidence` now returns a defensive copy, so an observer that retains and mutates the bytes can't corrupt the signed bytes used by later bundle aggregation. Regression test: mutating the returned evidence envelope leaves the snapshot's cached `Marshal` output intact.

Both verified under `go test -race`. Good catch on #2 especially — it's the exact cache-aliasing hazard from two PRs ago, which I should have caught when I wrote `snapshotEnvelopeForEvidence`.

PR #4082, opened 2026-06-18
feat(frost): stable member-scoped ROAST session key + transition record (7.3 PR2a)

mswilkison:
## Summary

RFC-21 **Phase 7.3, PR2a** — the first half of the cross-attempt ROAST work. Makes ROAST retry **actually consult the previous attempt's transition state** instead of silently always falling back to the legacy shuffle, by introducing a stable, member-scoped session key and a unified transition record. Design per the Codex + Gemini consult.

## The trap (verified against the worktree)

`signingSessionID` embeds the attempt number, so the orchestration cleanup stored the transition bundle under an **attempt-specific** key while the selector looked it up by `fmt.Sprintf("%v", srl.message)` — the keys never met, so the selector always fell back to legacy. Even fixing the format, attempt N+1's selector needs attempt N's state, so the key must be **stable across attempts**. The consult surfaced two more (both verified):
- the orchestration cleanup **cleared the handle** (`ClearCurrentAttemptHandleForSession`) before the selector could use it;
- the selector passed a **`nil` `dkgGroupPublicKey`** to `NextAttempt`, which derives a context `NewActiveRoastAttempt` then rejects (seed mismatch).

## Fix

- **Stable session id:** add `roastSessionID(message, root, startBlock)` (no attempt number) beside the attempt-specific `signingSessionID`; thread it via `signing.Request.RoastSessionID` → `NativeExecutionFFISigningRequest` and into `AttemptContext.SessionID`. The coarse/legacy path keeps the attempt-specific `SessionID` for its replay isolation (**split at the request level, unified inside ROAST/interactive**, per the consult).
- **Unified record:** replace the bundle registry with one `RoastTransitionRecord{Bundle, PreviousHandle, PreviousContext, DkgGroupPublicKey}` keyed by **`(roastSessionID, memberIndex)`**. It subsumes the key fix, the cleared-handle trap (the record carries the handle), the nil-dkgkey trap (carries the key), and multi-seat keying (consistent with #4081). Cleanup populates it under the stable `ctx.SessionID`; the **handle registry stays keyed by the attempt-specific `request.SessionID`** so the coarse receive-loop's binding validation + snapshot submission are unaffected (this separation is deliberate — keying the handle registry by the stable id would silently disable those coarse-path lookups).
- **Selector:** `signingParticipantSelector.Select` gains a `memberIndex` param; the ROAST selector reads `RoastTransitionForSession(roastSessionID, member)` and feeds `PreviousHandle` / `Bundle` / `DkgGroupPublicKey` to `EvaluateRoastRetryForSigning`.

## Scope (staged)

PR2a delivers the **coordinator/local-bundle (and single-node) retry path** — a record is populated only for the seat that produces the bundle locally (the elected coordinator). Non-coordinator seats receive the bundle via **PR2b**'s snapshot/transition bus exchange, which also wires the blame/evidence bridge (`ClassifyCandidateCulprits` via the existing `EngineRound2ShareVerifier`, `CoordinatorPackageProofs`, proof-of-attendance snapshots). Production interactive signing stays gated until both land.

## Tests

- Registry: round-trip, **member-scoping** (two seats don't collide; a member with no record reads absent), overwrite, clear, nil-bundle-ignored, TTL eviction, default-build no-op.
- Orchestration cleanup: produces a record (with handle + DKG key) when the elected seat, none when non-elected, swallows `AggregateBundle` errors.
- Selector: falls back to legacy when no record / registry empty / **no record for this member**; consumes the record (calls `NextAttempt` against `PreviousHandle` with the record's DKG key) when present.

All four tag combinations build/vet/test green (default / `frost_native` / `+frost_roast_retry` / `+frost_tbtc_signer cgo`); full pkg/tbtc suite green both builds.

### DIFF
--- a/pkg/frost/signing/attempt_context_from_request.go
+++ b/pkg/frost/signing/attempt_context_from_request.go
@@ -121,8 +121,19 @@ func BuildAttemptContextFromRequest(
 	}
 	attemptNumber := uint32(request.Attempt.Number - 1)
 
+	// Prefer the STABLE ROAST session id so ctx.SessionID -- and everything
+	// keyed off it (the orchestration handle + transition-record registries,
+	// the selector lookup, the interactive engine session) -- is stable across
+	// attempts; the per-attempt SessionID would make the next attempt's selector
+	// unable to find the previous attempt's transition record. Fall back to
+	// SessionID when the caller does not drive ROAST orchestration.
+	roastSessionID := request.RoastSessionID
+	if roastSessionID == "" {
+		roastSessionID = request.SessionID
+	}
+
 	ctx, err := attempt.NewAttemptContextWithParking(
-		request.SessionID,
+		roastSessionID,
 		keyGroupID,
 		dkgPub,
 		digest,
--- a/pkg/frost/signing/native_ffi_executor_adapter.go
+++ b/pkg/frost/signing/native_ffi_executor_adapter.go
@@ -16,6 +16,7 @@ import (
 type NativeExecutionFFISigningRequest struct {
 	Message             *big.Int
 	SessionID           string
+	RoastSessionID      string
 	MemberIndex         group.MemberIndex
 	GroupSize           int
 	DishonestThreshold  int
@@ -89,6 +90,7 @@ func (nefea *nativeExecutionFFIExecutorAdapter) Execute(
 	ffiRequest := &NativeExecutionFFISigningRequest{
 		Message:             request.Message,
 		SessionID:           request.SessionID,
+		RoastSessionID:      request.RoastSessionID,
 		MemberIndex:         request.MemberIndex,
 		GroupSize:           request.GroupSize,
 		DishonestThreshold:  request.DishonestThreshold,
--- a/pkg/frost/signing/request.go
+++ b/pkg/frost/signing/request.go
@@ -11,9 +11,19 @@ import (
 
 // Request carries execution input for a FROST signing backend.
 type Request struct {
-	Message     *big.Int
-	SessionID   string
-	MemberIndex group.MemberIndex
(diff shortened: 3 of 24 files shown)


mswilkison:
## Update — re-scoped after review (head `3d1e9c0df`)

Two external review rounds, both useful; PR is now the **data foundation only**.

**Round 1 (Codex; Gemini approved):** 2 valid folds — (P1) the interactive drive bound `request.SessionID` into `NewActiveRoastAttempt`, but `ctx.SessionID` is now the stable id, so every *gated* interactive attempt would have hard-failed before round 1 → now binds to `attemptCtx.SessionID`; (P2) the `(session, member)` record persisted after consumption → consume-on-read.

**Round 2 (Codex):** 2 valid issues that **re-scoped the PR** — (P1) records are produced only locally by the elected coordinator and aren't distributed, so a record-*consuming* selector would diverge (coordinator takes ROAST, peers fall back to legacy → split `IncludedSet`s); (P2) the member-level `NextAttempt` result was squeezed through the legacy *address*-based selector, losing per-member precision under multi-seat / partial readiness.

**Resolution:** the selector's *consumption* of records can't be separated from PR2b's distribution (P1) + member-level integration (P2). So this PR now keeps the **session-key fix + `RoastTransitionRecord` registry + cleanup production + the `Select` interface widening (`memberIndex`)** — and the **ROAST selector delegates to legacy** (no consumption → no divergence). PR2b adds the snapshot/transition **bus exchange** + **member-level** selection (fixing P1+P2 together) + the blame/evidence bridge.

Net: this PR is the non-fracturing foundation; the genuinely-coupled selector activation moves to PR2b where distribution makes it correct.

Issue #18, opened 2017-12-13
Bundle or remove dynamic dependencies


mhluongo:
As found in #17, dynamic dependencies required by the BLS implementation make builds and end-user deploys a PITA.

In particular, https://github.com/dfinity/go-dfinity-crypto/ depends on https://github.com/dfinity/bn, which is intended to be installed or included pre-built as a dynamic lib.

We need to either get to a single static executable, or package the client up with any dependencies in a platform-specific way for at least macOS and reasonable Debians.

Shadowfiend:
Having lost the dependency on bn and go-dfinity-crypto, this is no longer an issue :tada:

Issue #15, opened 2017-12-10
Pin all docker images


mhluongo:
We'll need something other than https://github.com/thomasWeise/docker-texlive and should pin `circle/node:latest` in CI

lionakhnazarov:
Closing as stale. Legacy infra/CI/local-dev task that no longer matches current tooling (Yarn 4, updated workflows, v2 stack) on main.

PR #4109, opened 2026-06-23
Security release/candidate 1

lionakhnazarov:
## Summary

Integration candidate for the coordinated security release. This description
tracks the current PR head and distinguishes implemented behavior from the
release contract that remains to be built and approved.

| Identity | Current value |
|---|---|
| PR head | `d7bf8c0753f3aac574f8d20d93c75268f350b389` |
| PR base | `main@a7ac8989b662d51d8a94fa28f1ac226058d5d6cc` |
| tss-lib | `github.com/threshold-network/tss-lib@v0.0.0-20260729021955-d847ce003019` (`d847ce003019`) |
| keep-common | `v1.7.1-tlabs.1` |
| go-ethereum | `v1.17.3` |
| Go | `1.25.7`, toolchain `go1.25.10` |

The candidate is **not a mainnet-ready artifact**.
`participation.MainnetCutoverBlock` is deliberately `0`, and mainnet startup
fails closed until a reviewed release commit supplies the real cutover block.

## Implemented state at this head

- The participation gate uses an authoritative chain clock for live
  admission/completion decisions.
- A process-local permit selects an immutable legacy/security-v2 strategy from
  the caller-supplied ceremony start-block number and retains it for that
  in-process permit's lifetime.
- Cutover diagnostics and release-manifest checks are present.
- Canonical block-hash/incarnation binding and durable restart inheritance are
  **not** present; those remain R03/R04 work.
- The `d847ce003019` nominal-legacy tss-lib branch does **not** reproduce
  PRIOR's ZK/ZKV transcript equations. Complete PRIOR compatibility and the
  exact mixed-fleet qualification remain R01 blockers.
- The broader authenticated tBTC coordination-envelope rejection boundary and
  the final independent release-evidence chain are not complete.

## Intended cutover contract — not a present-readiness claim

- Every production process and eligible seat runs the exact R1 artifact before
  reviewed cutover block `C`; the permitted residual at `C` is zero PRIOR
  processes and zero PRIOR-eligible seats.
- A verified canonical work incarnation selects its mode once. Work anchored
  below `C` may retain repaired PRIOR-compatible behavior until terminal; work
  anchored at or after `C` uses security-v2.
- Retries and restarts inherit a durably sealed mode. Current height, wallet
  age, delivery time, peer claims, and dependency version are not selectors.
- A sealed R1 legacy completion may cross `C` under its canonical commit fence.
  It does not authorize a PRIOR process, unsealed legacy session, post-`C`
  legacy penalty commit, or downgrade retry.
- One ceremony is homogeneous; mixed-mode contributions fail closed rather
  than negotiating or falling back.

## Breaking-change summary

| ID | Change | Current limitation / required handling |
|---|---|---|
| BC-1..BC-5 | security-v2 GG20 transcript/session binding and typed fixed-width session IDs | Current nominal-legacy ZK/ZKV is not PRIOR-compatible; R01 and durable R03/R04 sealing remain required |
| BC-6..BC-8 | mode-specific ECDH/hash-to-curve behavior and Go API changes | Rebuild external integrators and carry one explicit ceremony strategy |
| BC-9 | bootstrap removal and embedded well-known peers | Remove obsolete bootstrap config and update peer-count monitoring |
| BC-10 | RandomBeacon reentrancy guard in a directly deployed, non-proxy contract | Activate only through a separately reviewed deployment/address cutover |
| OV-1 | temporary `clientInfo.port = 9601` compatibility default | Pin an explicit value; expose the unauthenticated endpoint only over a trusted path |
| OV-2 | `connected_bootstrap_count` renamed to `connected_wellknown_peers_count` | Update dashboards and alerts |
| OV-3 | `--network.bootstrap=true` deprecated | Remove when convenient |

## Direct landing stack

1. [lionakhnazarov/keep-core#17](https://github.com/lionakhnazarov/keep-core/pull/17)
   corrects release documentation and repairs the required PR fuzz execution
   path while retaining ClusterFuzzLite build validation. Exact head:
   `05dac3e0b4703c459d5dfafb38ee261a358b8850`.
2. [lionakhnazarov/keep-core#18](https://github.com/lionakhnazarov/keep-core/pull/18)
   adds the non-authoritative, permanently blocked R00 inventory. It is
   cumulative and must reduce to its two R00 commits after #17 lands. Exact
   head: `27768989b609f51b14781a845fec5c920e234bd2`.
3. [lionakhnazarov/keep-core#19](https://github.com/lionakhnazarov/keep-core/pull/19)
   adds only the reviewed, provisional, non-gating canonical-encoding
   foundation. It is cumulative and must reduce to its single R00A commit
   after #17 and #18 land. Exact head:
   `1775696e87d5feca58c22ece23dc63a225440f46`.

None of these PRs is merged by this description update. They must land in
#17, #18, #19 order, with owner review and hosted checks repeated after each
base change. Prefer merge commits. An exact-head fast-forward is allowed only
when the then-current owner base is already an ancestor of the exact PR head;
after a merge-commit landing of a prior slice, that will normally be false for
the cumulative successor. Never squash, rebase-merge, reset, or force-update
the owner branch to simulate a fast-forward. Immediately before every merge,
refetch and require the expected owner base, exact PR head, reviewed reduced
diff, and current-base hosted-check context to be unchanged; stop, re-vet, and
rerun on any drift.
The draft R00B policy remains local and is intentionally excluded. R00A is not
complete: #19 excludes the registry, ledger, signatures, trust, authority,
fixtures, CLI, and cross-language implementation.

## Validation state

- The current upstream head's required client, contract, lint, vet, scan, and
  integration checks are green; the sole red required check is the existing
  ClusterFuzzLite `PR (address)` job.
- Fresh upstream run
  [32481030115](https://github.com/threshold-network/keep-core/actions/runs/32481030115)
  failed in libFuzzer's `TracePC::ClearInlineCounters` before the Go fuzz
  callback executed, on a different target and the empty seed. The evidence
  supports an intermittent native-Go/libFuzzer startup failure; explicit
  `language: go` is a necessary action-input correction but is not claimed as
  the cause or remedy.
- Landing PR #17 retains the pinned ClusterFuzzLite build action and replaces
  only the required PR execution path with native Go ASAN fuzzing over all 41
  registered targets. Its first hosted run exposed cold fuzz-instrumented
  compilation inside the target deadline; the settled head precompiles one
  real fuzz binary for each of 11 packages before starting phase timing.
- A later cumulative hosted run passed 12 native targets, then Go 1.25.10
  emitted its own `context deadline exceeded` after the requested eight-second
  interval completed on `FuzzAct1MessageUnmarshal`. This matches
  [golang/go#75804](https://github.com/golang/go/issues/75804); the upstream
  correction in [golang/go#79199](https://github.com/golang/go/pull/79199) is
  newer than the pinned Go 1.25 line. The final runner permits one retry only
  for the complete EOF-anchored harness signature and continues to fail on
  every extra diagnostic, incomplete interval, zero-execution interval,
  nested finding, crash, sanitizer report, signal, other status, or repeat.
- Exact Linux/amd64 Go 1.25.10 evidence on final Stage 1 head
  `05dac3e0b4703c459d5dfafb38ee261a358b8850` passed 11/11 fuzz-instrumented
  ASAN precompiles and 41/41 eight-second targets, including the previously
  affected target. Its Linux contract suite, deadline/process cleanup,
  crash-persistence, failure-propagation, and adversarial guards also passed;
  independent final review found no P0/P1/P2 issue.
- Earlier #17/#18 hosted runs stopped while fetching modules from
  `proxy.golang.org` with HTTP/2 `INTERNAL_ERROR` before native timed fuzzing;
  those are infrastructure failures, not source validation. The older #19
  `PR (address)` run is the completed-interval Go coordinator race described
  above.
- Fresh exact-head `target-sync` is green on #17, #18, and #19. Current
  `PR (address)` / Client runs are
  [#17 fuzz 33105613931](https://github.com/lionakhnazarov/keep-core/actions/runs/33105613931) /
  [client 33105613904](https://github.com/lionakhnazarov/keep-core/actions/runs/33105613904),
  [#18 fuzz 33105667389](https://github.com/lionakhnazarov/keep-core/actions/runs/33105667389) /
  [client 33105667351](https://github.com/lionakhnazarov/keep-core/actions/runs/33105667351),
  and [#19 fuzz 33105703470](https://github.com/lionakhnazarov/keep-core/actions/runs/33105703470) /
  [client 33105703460](https://github.com/lionakhnazarov/keep-core/actions/runs/33105703460).
  All three exact-head `PR (address)` runs are green with 41 retained CFL
  target builds, 11 native-Go package precompiles, and 41/41 terminal
  native-target passes. #19 exercised the
  bounded mitigation once: `FuzzPointsAccusationsMessageUnmarshal` completed
  8.06 seconds and 899 executions, emitted only the exact Go harness deadline
  signature, then passed its sole retry. All three exact-head Client workflows
  are green through downstream integration. All 12 newest-head workflow runs
  across the three PRs are terminal success; expected conditional jobs are
  skipped. #18 and #19 must still rerun against each newly landed predecessor
  base before their own merge.
- A fresh local `go test -count=1 -timeout=15m ./...`, repository-wide
  `go vet ./...`, R00 race tests, and 31-anchor immutable-source verification
  passed on 2026-08-27.
- Earlier Sepolia smoke results remain historical characterization; they do not
  prove the final source/build/policy identity.

## Current release blockers

- [x] Fresh green hosted `PR (address)` on the exact #17, #18, and #19 heads
- [x] Terminal green relevant Client checks on the exact #17, #18, and #19 heads
- [ ] Owner review and sequential merge of #17, #18, and #19, with checks
      rerun after each owner-base change
- [ ] Predicate-complete R00 reproductions, including the complete R00-18 trace
- [ ] Complete R00A registry/ledger codecs, outer resource limits, external
      golden corpora, and cross-language parity
- [ ] Approved signature, trust-root, authenticated monotonic-authority, and
      interoperability contract; synthetic fixtures are not authority
- [ ] Protocol-owner and security-owner approval over the canonical R00B policy
      identity; no signatures or approvals are inferred here
- [ ] R01 byte-exact PRIOR TSS compatibility and exact-binary fleet matrix
- [ ] R03 canonical incarnation binding and R04 crash-safe durable ledger
- [ ] Downstream product/recovery mechanisms and independently derived
      qualification/final evidence bound to the exact promotable OCI digest
- [ ] Reviewed nonzero mainnet cutover block `C` and release commit
- [ ] Required human review approval

The external scenario runner and DigitalOcean evidence production remain in
`tlabs-xyz/pr4109-do-infra`; they are not vendored into this product/release
DAG.

## Review focus

- [ ] Present-tense implementation claims match code; intended-contract claims
      remain visibly conditional on their prerequisite nodes
- [ ] Canonical work identity, reorg/restart handling, and effect persistence
      fail closed
- [ ] Source, policy, build inputs, executable bytes, OCI manifests, platform,
      and evidence roots remain bound end to end
- [ ] Evidence producers cannot decide their own acceptance result
- [ ] Release docs, manifests, metrics, and runtime behavior agree

## Related

- Security epic: [tlabs-xyz/keep-core-security#41](https://github.com/tlabs-xyz/keep-core-security/pull/41)
- Hardened dependency: [threshold-network/tss-lib](https://github.com/threshold-network/tss-lib) at `d847ce003019`
- External test/evidence project: [tlabs-xyz/pr4109-do-infra](https://github.com/tlabs-xyz/pr4109-do-infra)


lrsaturnino:
The PR description says head is `dd2c87360` with go-ethereum bumped to v1.17.3 (+ Go 1.25), but the current branch tip `ec5dbd1b3` doesn't have those changes. Looks like they got dropped during the July 12 rebase.

Worth double-checking `go.mod` before this merges — go-ethereum v1.13.15 has an open CVE advisory.

lionakhnazarov:
> The PR description says head is `dd2c87360` with go-ethereum bumped to v1.17.3 (+ Go 1.25), but the current branch tip `ec5dbd1b3` doesn't have those changes. Looks like they got dropped during the July 12 rebase.
> 
> Worth double-checking `go.mod` before this merges — go-ethereum v1.13.15 has an open CVE advisory.

thanks Leo! Fix applied

lrsaturnino (dismissed):
LGTM

mswilkison:
Two `go.mod` observations from a dependency sweep. The tss-lib upgrade itself looks right — the bump to `86bd1a375cc0` is the current fork head, and both mandatory caller obligations from the protocol break are wired in (`SetSessionNonce` 21×, `fullBytesLen` 11× in the diff), which is the part that would have failed closed if missed.

### 1. The `x/crypto` bump lands just below the advisory floor

This PR goes `0.32.0 → 0.47.0`. The seven open critical advisories against `x/crypto` (`CVE-2026-39830`, `-39831`, `-39832`, `-39833`, `-39834`, `-42508`, `-46595`) are all **first patched in `0.52.0`**, so they stay open after this merges.

There is also a separate open PR #4142 doing `0.32.0 → 0.52.0`. These two will conflict on `go.mod`, and whichever lands second should win at `>= 0.52.0`.

Worth keeping in proportion: all seven are in `ssh` / `ssh/agent`, and `go list -deps ./...` shows **zero `x/crypto/ssh` packages** in the build — keep-core only pulls `sha3`, `blake2b`, `chacha20`, `curve25519`, `scrypt` and friends. So this is closing advisory noise rather than real exposure. Raising it only because a release framed as a security release probably wants the alerts actually cleared.

### 2. `btcd` is untouched, and the `replace` is a *downgrade*

Not changed by this PR, but it sits in the same file and the contrast with the above is sharp:

```
go.mod:24  github.com/btcsuite/btcd v0.23.2                              // require
go.mod:13  github.com/btcsuite/btcd => github.com/btcsuite/btcd v0.22.3  // replace
```

The `replace` pins **below** the `require`, and `v0.26.2` is current. OSV reports 6 advisories against `v0.22.3`:

| advisory | summary | fixed in |
| --- | --- | --- |
| `GO-2022-1098` | DoS in message decoding | **0.23.2** |
| `GHSA-2chg-86hq-7w38` | mishandles witness size checking | **0.23.2** |
| `GO-2024-2818` | consensus failures | 0.24.0 |
| `GHSA-3jgf-r68h-xfqm` | consensus failures | 0.24.0 |
| `GO-2024-3189` | consensus failure | 0.24.2-beta.rc1 |
| `GHSA-27vh-h6mc-q6g8` | incorrect re-implementation of `FindAndDelete` | 0.24.2-beta.rc1 |

The first two are already fixed at the version this repo *requires* — the `replace` reintroduces them.

Unlike the `x/crypto` case, this code **is** in the build: `go list -deps ./...` returns 16 `btcsuite/btcd` packages including `blockchain`, `blockchain/indexers`, `btcec`, `btcjson` and `btcutil`. Consensus-failure advisories in a Bitcoin bridge seem worth either fixing here or recording why the downgrade is deliberate — right now the `replace` carries no comment, unlike the `keep-common` and `protobuf` ones above it.

mswilkison:
Opened a stacked PR against `security-release/candidate-1` with the fix for item 2: **lionakhnazarov/keep-core#14**. (Item 1, the `x/crypto` floor, is left alone — it belongs with #4142.)

The short version of what the `replace` was actually doing there: it existed **only** to keep the pre-split `btcd/btcec` package alive. btcd v0.23 moved `btcec` into its own module, so the whole btcd module was held one minor behind to keep one crypto package resolvable — and the Bitcoin consensus, wire, and script code came along for the ride. The fix serves that one package from an in-tree copy of the v0.22.3 `btcec` sources (gofmt-only delta, `VENDOR.md` documents provenance and a `diff -r` recipe) and moves the module itself to **v0.25.0**, which clears all six advisories.

Two things worth flagging from doing it:

**The `btcd/v2` alias was worse than untouched.** It only isolates self-contained packages — an aliased package importing a sibling does so by plain path, which resolved back to **v0.22.3**. So the "patched" alias was linking vulnerable wire/txscript underneath. It's deleted along with the downgrade.

**v0.26.2 isn't reachable.** v0.26 split `wire`/`txscript`/`chaincfg` into `/v2` modules and dropped the plain paths, which the go-electrum and tss-lib forks import with strings we don't control. v0.25.0 is the newest classic-layout release; since everything is fixed by v0.24.2-beta.rc1, that's a layout ceiling and not a security one.

One thing the upgrade itself introduced, which I'd flag even though it's now fixed in the PR: **btcd v0.24.2+ made the transaction decoder panic instead of erroring on oversized input.** It slices every script of a transaction out of a single 4 MiB buffer and bounds each script only against `maxWitnessItemSize`, never against what's left:

```
panic: runtime error: slice bounds out of range [:900000] with capacity 594304
  btcd@v0.25.0/wire/msgtx.go:1041  readScriptBuf
  pkg/bitcoin/transaction.go:155   (*Transaction).Deserialize
```

Reproduced against the worktree — 4 × 900 KB scripts decode fine, 5 panics. Both decoders being replaced (v0.22.3 and the v0.23.4 alias) handle the same bytes cleanly, so this would have been a regression on a path fed by the untrusted Electrum server: `decodeTransaction` runs *before* `GetTransaction`'s txid check, `GetTransactionConfirmations` has no hash check at all, and no `recover()` covers the tbtc signing paths. Upstream fixed it only in the `wire/v2` module, i.e. inside the v0.26 layout we can't reach. The PR caps deserialization at 4,000,000 bytes — above the max block weight, so no consensus-valid transaction is affected — with a regression test.

Also: the Dockerfile resolved the module graph from `go.mod`/`go.sum` before copying any source, so the new directory `replace` broke the image build while local builds passed. Fixed in the same PR.

Full `go test ./...` green, and the residuals (archived `btcutil` still pulled in by tss-lib's `base58`, leftover `/go.mod`-only hashes in `go.sum`) are written up in the PR description.

mswilkison:
Correction to my comment above: I pointed at a PR stacked on this branch, but that was the wrong home for it. Closed in favour of **#4219, against `main`**.

The reason: the btcd downgrade is *pre-existing on `main`* — identical `replace` lines, identical inert `v0.23.2` require. It isn't something this PR introduced (as I noted originally) and it isn't security-release-specific, so it shouldn't ride under the release candidate. On `main` it also gets CI and normal review. This branch picks it up on its next rebase; the `go.mod` conflict there is mechanical, since this branch's dependency set is much newer and the collisions are in surrounding lines rather than the btcd ones.

The substance is unchanged from what I described above, plus one thing worth repeating because it bites on any btcd bump, not just this one: **v0.24.2+ made the transaction decoder panic rather than error on oversized input.** It slices every script of a transaction out of a single 4 MiB buffer and bounds each script only against `maxWitnessItemSize`, never against what's left. Reproduced: 4 × 900 KB scripts decode, 5 panic. Reachable from the untrusted Electrum backend ahead of `GetTransaction`'s txid check, with no `recover()` on the tbtc signing paths. #4219 caps deserialization at 4,000,000 bytes — above max block weight, so nothing consensus-valid is affected.

piotr-roslaniec:
Checking in on this security release candidate — it conflicts with `main` across 15 files (`pkg/tbtc/node.go`, `pkg/tbtc/dkg.go`, `pkg/bitcoin/transaction_builder.go`, `pkg/maintainer/spv/spv.go`, `go.mod`/`go.sum`, among others). Given the security-sensitive scope, we'd rather you drive the rebase on your own timeline than have us touch it. Since this PR is from your fork, we can't push a rebase to it ourselves either way — only you can push to `lionakhnazarov/keep-core`. Let us know if there's a target window, or if it'd help to split this into smaller pieces.

lionakhnazarov:
> Checking in on this security release candidate — it conflicts with `main` across 15 files (`pkg/tbtc/node.go`, `pkg/tbtc/dkg.go`, `pkg/bitcoin/transaction_builder.go`, `pkg/maintainer/spv/spv.go`, `go.mod`/`go.sum`, among others). Given the security-sensitive scope, we'd rather you drive the rebase on your own timeline than have us touch it. Since this PR is from your fork, we can't push a rebase to it ourselves either way — only you can push to `lionakhnazarov/keep-core`. Let us know if there's a target window, or if it'd help to split this into smaller pieces.

@piotr-roslaniec Thanks for the ping.

I’ve synced security-release/candidate-1 with current main (merge commit on the PR tip) so the conflict set should be cleared for re-review. I’m aiming to keep this as a single PR and finish remaining review/CI follow-ups. I’d prefer not to split unless review becomes blocked on size/scope. The cutover/participation changes are tightly coupled; splitting would likely create awkward intermediate states

lrsaturnino:
Transparency note for @lionakhnazarov: I accidentally pushed 591b8797c directly to security-release/candidate-1 while setting up to open a PR for the transaction-monitor lifecycle-launch fix (a local clone had push.default=upstream, which redirected the push to the tracked ref instead of creating a new branch). I reverted it immediately with a forward commit (d2a95eed0) that restores the branch tree to its exact prior state without rewriting shared history, then reopened the same verified patch properly at #15 from my fork. Sorry for the disruption.

lrsaturnino:
Closing this one.

The implementation has moved to a T Labs-owned repository, so this branch is now retained as source of record rather than an edit base. It will be superseded by a replacement PR once that work is reviewed and ready, instead of being merged from here.

Nothing is lost — the full 456-commit history and the four stacked hardening branches are archived intact, verified tree-identical at both ends.

cc @lionakhnazarov @piotr-roslaniec @mswilkison

PR #3882, opened 2026-03-09
feat(tbtc): covenant signer final project branch

mswilkison:
## Summary

This PR introduces the **covenant signer** — a Go HTTP service that signs Bitcoin PSBTs (Partially Signed Bitcoin Transactions) using each operator's existing tECDSA wallet share from DKG. It is consumed by the off-chain covenant tooling (in `tbtc-v2-ac`) when a covenant migration or recovery operation needs an operator-side signature on a constructed PSBT.

**Repo footprint:** ~9.4K LoC of Go in `pkg/covenantsigner/` plus integration glue across `pkg/tbtc/`, `pkg/chain/ethereum/`, `pkg/bitcoin/`, `cmd/`, and `config/`. 48 files changed, +13,574 / −30.

## Audit positioning

This PR is the **secondary review target** for the Trail of Bits PSBT covenant audit. The primary review covers the on-chain Solidity contracts in `tlabs-xyz/tbtc-v2-ac`. The current SOW miscalls this PR an "off-chain attester / watchdog service" — that wording is incorrect:

- This PR is the **covenant signer**, not a watchdog. The signer has **no `ATTESTER_ROLE` and no `WATCHER_ROLE`** on-chain.
- The on-chain `WatchdogEnforcer.sol` (the actual permissionless watchdog) is already merged on `tbtc-v2-ac/dev`. It lives in `tbtc-v2-ac`, not here.
- The off-chain `ac-watchdog-service` (Node.js) lives in `tlabs-xyz/ac-watchdog-service`, not here.

## Where to start reading

Recommended file order:

1. `pkg/covenantsigner/doc.go` — package overview
2. `pkg/covenantsigner/DEPLOYMENT.md` — deployment topology, dedup scope
3. `pkg/covenantsigner/server.go` — HTTP, bearer-token auth, timeouts
4. `pkg/covenantsigner/service.go` — Submit/Poll, idempotency, request digest
5. `pkg/covenantsigner/validation.go` — request validation, fuzz target
6. `pkg/covenantsigner/types.go` — request/response shapes, `expiresAt`
7. `pkg/covenantsigner/store.go` — in-memory + on-disk persistence
8. `pkg/covenantsigner/engine.go` — engine interface
9. `pkg/tbtc/covenant_signer.go` — concrete engine: BTC confirmations, PSBT signing, tECDSA share lookup
10. `pkg/internal/canonicaljson/marshal.go` — canonical JSON

[...]

PR #4126, opened 2026-07-01
Trail of Bits tBTC v2 AC Extension — security remediation

lrsaturnino:
Applies the remediation for the Trail of Bits "tBTC v2 AC Extension" security assessment. Every in-scope finding whose target code lives in this repository is fixed in its own commit, tagged with its public report ID (`TOB-TBTCACEXT-<n>`) so each fix maps 1:1 to a finding, with regression test(s) added alongside each fix.

## Base / stacking

This branch was created off the frozen review commit `235e8fd9`, an ancestor of `feat/psbt-covenant-final-project-pr` (the reviewed PR, #3882). The PR therefore targets that branch so the diff is exactly the remediation (17 commits) rather than PR #3882's own changes. Because the feature branch has since advanced past the frozen commit (it now carries the deep-audit fixes in #3938), expect to reconcile against those later commits when this lands. Retarget to `main` once #3882 lands.

## Findings remediated (17)

- High: 1 (fail closed on covenant signing until the tBTC Bridge covenant fraud-defense is deployed), 2 (derive the P2WSH output script for the migration destination so deposits stay revealable), 27 (accumulate the withdrawn counter on revoked escrow withdrawals to stop repeated draining)
- Medium: 5 (guard the `doneSigners` map on every access in the signing-done check), 6 (synchronize `BackoffStrategy` tick counters against concurrent retransmission ticks), 22 (scan past address spam when discovering SPV proof targets), 36 (do not archive wallets still inside the DKG approval window), 56 (deduplicate DKG results only after handling completes), 57 (deduplicate wallet closures only after archival succeeds), 65 (use a flat job directory so covenant-signer jobs persist across restarts), 87 (deauthorize archived and non-live wallets in covenant signing)
- Low: 10 (read `TxMaxFee` from its own field in past redemption events), 26 (reject signing-done messages from members not in the current attempt), 72 (deduplicate the `DKGStarted` callback only after its handler completes)
- Informational: 11 (count the third operator's seats in the triplet-eligibility filter), 70 (reject non-canonical high-S signer approval certificates), 92 (fail production signer startup when approval roots are missing)

Finding 1's complete remediation is a bridge-side change in the tBTC Bridge (`Fraud.defeatFraudChallenge`) that lives in the `tbtc-v2` repository; the keep-core side fails closed until an operator confirms that defense is deployed.

## Post-review hardening (2026-07-13)

A follow-up review of the initial remediation surfaced residual issues in two already-fixed findings, plus several cleanup items. Each is folded into its originating finding's commit rather than appended separately, so the one-commit-per-finding mapping still holds:

- **Finding 22** — SPV proof discovery is now bounded. A persisted, per-wallet circular cursor (`pkg/maintainer/spv/scanner.go`) caps each maintainer tick to `TransactionScanLimit` (default 100, new `--spv.transactionScanLimit` flag) newly-examined transactions instead of re-walking a wallet's entire confirmed history every tick, while the circular design still guarantees spam can never permanently evict a real protocol transaction from eventual discovery.
- **Finding 5** — `signingDoneCheck` now constructs a fresh instance per retry attempt (via a factory) instead of reusing one instance across attempts, with explicit teardown on the early-continue paths in the retry loop. This closes a residual data race and cross-attempt message-corruption risk on the retry path that the original mutex-guarding fix didn't fully cover, verified by a new regression test exercising the failed-attempt-to-retry interleaving under `-race`.
- Cleanup: extracted a shared P2WSH-derivation helper in `covenant_signer.go`, removed a dead nil-map guard in `deduplicator.go`, and de-duplicated a test mock in `bitcoin_chain_test.go`.

## Verification

`go build ./...` passes. The affected-package tests are green (`pkg/tecdsa/retry`, `pkg/net/retransmission`, `pkg/covenantsigner`, `pkg/maintainer/spv`, and the covenant-signer, signing-done, and wallet-lifecycle regression tests in `pkg/tbtc`). Each finding commit carries its own regression test(s) — race-detector tests for the concurrency findings, handler-level transient-failure and redelivery tests for the deduplication findings, and value-level tests for the accounting and filtering fixes. The full module test suite and `-race` runs on the changed packages are green, including CI's `client-race` job commands.


lionakhnazarov (commented):
## Review summary

**Verdict: approve with minor follow-ups.**

This is a well-structured remediation: 17 ToB findings, each in its own commit (`TOB-TBTCACEXT-<n>`) with regression test(s). CI is green (including `client-race`). I ran the affected packages locally (`pkg/tbtc`, `pkg/covenantsigner`, `pkg/maintainer/spv`, `pkg/net/retransmission`, `pkg/tecdsa/retry`) — all pass.

### Strengths
- Concurrency fixes (5, 6) use mutex + race-detector tests.
- Deduplication redesign (56, 57, 72): `claim` / `confirm` / `abort` correctly allows redelivery retry after transient failures.
- High-severity fixes (1, 2, 27, 22, 36, 87) look correct and are well-tested.
- Handler extractions improve testability without changing behavior.

### Before merge into the feature branch
1. Add operator docs for `bridgeCovenantFraudDefenseConfirmed` (config sample + rollout order: bridge deploy first, then flip flag).
2. Link the matching `tbtc-v2` PR for Finding 1 in the PR description.
3. After #3882 lands, rebase/retarget to `main` and re-run CI.
4. Watch for overlap with security-release work (#4109) on deduplicator patterns when both trains converge.

### Stacking note
The merge commit rebases remediation onto a newer #3882 head (post-#3938). Reconcile any overlap with #3938 dedup/concurrency fixes when integrating.

lionakhnazarov (dismissed):
LGTM

PR #4134, opened 2026-07-07
test(frost/roast): real-under-failure ROAST retry + blame tests (dropout, invalid-share, equivocation, overflow, cross-node fracture)

mswilkison:
## What

Adds **real-under-failure** ROAST retry/blame tests. Prior coverage was disjoint: the real-cgo interactive e2es are happy-path only, and the retry/parking/blame machinery (`next_attempt.go`, the round-2 collector, the overflow primitive, the cross-node selector guard) is unit-tested with fakes — nothing wired the real component + an induced fault + its retry/blame outcome together. Each test asserts the induced fault was **actually reached** (not merely that the outcome happened), so a synthetic bundle can't make it pass vacuously.

## The tests

Over the real `pkg/net` transport against the real cgo FROST engine (`pkg/frost/signing`, `frost_native frost_tbtc_signer cgo`):

- **`..._dropout_retry_...`** (#1) — a *selected* signer withholds its round-2 share → the elected coordinator (the aggregator) starves on a real collect-shares timeout → `NextAttempt` **transiently parks** the silent seat → the reshuffled subset aggregates a real BIP-340 signature and reaches `Succeeded`. Asserts the coordinator starves *specifically* at share collection AND the target actually reached round 2 and produced the share it withheld.
- **`..._invalid_share_exclusion_...`** (#2) — a selected signer submits a structurally valid but cryptographically **wrong** round-2 share → the real aggregate fails with a typed share-verification error naming the culprit → an f+1 reject quorum **permanently excludes** it → the surviving subset finalizes for real.
- **`..._share_conflict_equivocation_...`** (#3b) — a selected signer runs round 2, produces a genuine FROST share, then broadcasts **two body-different signed share submissions** (its real share + a re-signed copy with a mutated `signature_share`), both binding the authoritative package so both are *accepted* shares. An honest `Round2Collector` flags the second as `EquivocationKindShareConflict` (`ErrShareConflict`) and the process-wide observer receives the culprit-naming evidence. Asserts the two envelopes are distinct, same-package-bound accepted shares (a genuine double-sign, not a divergent share).
- **`..._coordinator_equivocation_...`** (#3a) — the elected coordinator distributes **two VALID coordinator-signed packages** (distinct taproot roots) for the same attempt → `verifiedCoordinatorEquivocations` bypasses the f+1 gate and forces **instant permanent exclusion**, even when the two proofs are split across two observers. Uses a **real secp256k1 operator-key Signer/Verifier** (not the NoOp pair), with a non-vacuous negative control: one authentic package + one whose signature is corrupted (rejected by the real verifier) is only one distinct authentic body and does **not** exclude — proving the verifier is genuinely cryptographic.
- **`..._overflow_park_...`** (#3c) — a member floods a receive loop faster than it drains, genuinely overflowing the bounded inbound channel via the real `enqueueOrRecordOverflow` primitive → the real recorded overflow evidence, carried by an f+1 quorum, drives a **transient park** (not exclusion). Asserts the channel actually rejected enqueues and the recorder recorded overflow, that the park is transient (a following attempt reinstates the flooder), and that the f+1 quorum is genuinely enforced (a single accuser does not park). The flooder is itself a bundle sender, so silence-parking can't account for the park.

In `pkg/tbtc` (`frost_native frost_roast_retry`, pure-Go):

- **`..._cross_node_fracture_...`** (#4) — two nodes with divergent selection state would broadcast **different NextAttempt sets** (a ROAST-consumed set vs a legacy shuffle) — the fracture class that splits the group. The test proves the divergence is real (registered ROAST-active node selects the full transition set; a would-be-legacy node trims to the honest threshold), then proves the **fail-closed guard** collapses it: the SAME node-B `Select` call that returned a concrete legacy set with the registry empty fails closed once ROAST is active, and node A fails closed on its missing expected transition. Neither emits a set, so they can't converge on divergent ones. Each fail-closed decision is repeated to show it is deterministic per-seat.

## CI coverage

- #1/#2/#3a/#3b/#3c run under **frost-cgo-integration** (`frost_native frost_tbtc_signer frost_roast_retry`, linked libfrost_tbtc).
- #4 runs under the **client** `frost_native frost_roast_retry` tag set over `./pkg/tbtc/...`.

All six are green locally, stable across repeated runs, with no regression to the existing real-cgo suite.

### DIFF
--- a/pkg/frost/signing/roast_runner_real_cgo_coordinator_equivocation_frost_native_test.go
+++ b/pkg/frost/signing/roast_runner_real_cgo_coordinator_equivocation_frost_native_test.go
@@ -157,7 +157,7 @@ func TestRealCgoInteractiveSigning_CoordinatorEquivocationForcesInstantPermanent
 	// bodies differ via the taproot root (0 vs 32 bytes) - signing the same body
 	// twice would give a different ECDSA signature but the SAME BodyHash, which is
 	// NOT an equivocation.
-	pkgA := newSignedPkg(nil)                                             // key-path spend
+	pkgA := newSignedPkg(nil)                                                       // key-path spend
 	pkgB := newSignedPkg(bytes.Repeat([]byte{0xab}, roast.TaprootMerkleRootLength)) // script-path spend
 
 	// ---- FAULT REACHED: a genuine equivocation, not a synthetic pass. ----
--- a/pkg/frost/signing/roast_runner_real_cgo_coordinator_equivocation_frost_native_test.go
+++ b/pkg/frost/signing/roast_runner_real_cgo_coordinator_equivocation_frost_native_test.go
@@ -1,0 +1,291 @@
+//go:build frost_native && frost_tbtc_signer && cgo
+
+package signing
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"sort"
+	"testing"
+
+	"github.com/keep-network/keep-core/pkg/chain"
+	"github.com/keep-network/keep-core/pkg/chain/local_v1"
+	"github.com/keep-network/keep-core/pkg/frost/roast"
+	"github.com/keep-network/keep-core/pkg/frost/roast/attempt"
+	"github.com/keep-network/keep-core/pkg/operator"
+	"github.com/keep-network/keep-core/pkg/protocol/group"
+)
+
+// This test closes the fourth real-crypto-under-failure gap: a COORDINATOR
+// EQUIVOCATION - the elected coordinator distributing two VALID, coordinator-
+// signed signing packages with the same attempt context but different bytes -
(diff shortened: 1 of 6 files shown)


mswilkison:
Good catch — fixed in the latest push.

The test now asserts the withheld-share path is actually exercised, before it drives the retry:
- **coordinator** must fail *specifically* at `collect shares` (proving it built + broadcast the package, produced its own share, then starved on the missing one) — not a generic early error;
- **target** must have reached round 2 and produced the share it withheld. Since only the target's *outbound* share is dropped, it still receives the coordinator's share and aggregates locally off the shared engine; that local success (`sigLen=64`, deterministic across `-count=3`) is the proof it produced — and the bus withheld — a genuine round-2 share. A pre-round-2 target failure (undelivered package / engine error before `InteractiveRound2`) now fails the test.

PR #4164, opened 2026-07-16
fix(spv): allow proofs longer than 144 headers

mswilkison:
## Summary

- remove the client-only 144-header cap from the SPV proof walk
- keep walking through mined headers until the proof has enough work or reaches the Bitcoin chain tip
- skip only headers whose decoded target exactly matches the Bridge DIFF1 target
- reject non-positive header targets before calculating difficulty

## Regression coverage

- 144 DIFF1 headers followed by a decisive difficulty-16 header
- non-DIFF1 bits 0x1d00aaaa, whose integer difficulty rounds to one, remain decisive
- zero-target compact bits return a proof error instead of panicking

## Testing

- go test -count=1 ./pkg/maintainer/spv
- go test ./pkg/maintainer/...

Stacked on #4039. The temporary base branch mirrors #4039 at 832c1168cedaf7841c4dc76d92b2913c54f61a84; retarget this PR to main after #4039 merges.

### DIFF
--- a/pkg/maintainer/spv/spv.go
+++ b/pkg/maintainer/spv/spv.go
@@ -20,9 +20,6 @@ import (
 
 var logger = log.Logger("keep-maintainer-spv")
 
-// The length of the Bitcoin difficulty epoch in blocks.
-const difficultyEpochLength = 2016
-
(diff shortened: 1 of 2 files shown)


lionakhnazarov (approved):
LGTM

PR #4165, opened 2026-07-16
refactor(spv): isolate btcd script classification

### LINKED ISSUE #3677: Package `pkg/maintainer/spv` should not depend on `btcd` directly
The `btcd` package is a 3rd party dependency meant to be used only within `pkg/bitcoin` package which isolates it from the client code. However, `pkg/maintainer/spv` depends on `btcd` directly which makes it leak beyond `pkg/bitcoin`. We should fix that and ensure `btcd` is not used directly by any other package except `pkg/bitcoin`.

mswilkison:
## Summary

- replace pkg/maintainer/spv's direct txscript classification with the existing bitcoin.GetScriptType abstraction
- preserve P2PKH, P2WPKH, P2SH, and P2WSH handling
- remove the last direct btcd import from pkg/maintainer/spv

## Testing

- go test ./pkg/bitcoin
- go test ./pkg/maintainer/spv
- go test -count=1 ./pkg/maintainer/...

Fixes #3677

## Summary by CodeRabbit

* **Refactor**
  * Updated transaction input classification to use the application’s Bitcoin script type handling.
  * Preserved existing detection and validation behavior for main and deposit inputs.
  * Removed an unnecessary underlying script-processing dependency.

### DIFF
--- a/pkg/maintainer/spv/deposit_sweep.go
+++ b/pkg/maintainer/spv/deposit_sweep.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/keep-network/keep-core/pkg/tbtc"
 
-	"github.com/btcsuite/btcd/txscript"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/keep-network/keep-core/pkg/bitcoin"
 	"github.com/keep-network/keep-core/pkg/chain"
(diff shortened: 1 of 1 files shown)


mswilkison:
CI note: I reran the non-required client-integration-test after its initial failure. The rerun failed for the same shared external-infrastructure reason as PR #4167: all 30 failures are fulcrum_tcp cases receiving `dial tcp 193.30.123.70:50001: connect: connection refused`; every other package passed. This endpoint is not touched by this patch, and the matching result on an unrelated PR confirms this is not a regression here. Rerun job: https://github.com/threshold-network/keep-core/actions/runs/29531102639/job/87752372643

piotr-roslaniec (approved):

PR #4169, opened 2026-07-17
feat(covenantsigner): wallet-signed EIP-712 approvals + cooperative REDEEM/RENEW

piotr-roslaniec:
## Summary

Enables **wallet-signed** (`eth_signTypedData_v4`) covenant artifact approvals and adds
the cooperative **REDEEM** / **RENEW** action paths to the keep-core covenant signer.

This is the keep-core prerequisite (section A of the VBA EIP-712 change-list) that must land
before this covenant work goes live: today the signer verifies a **bare EIP-712 struct
hash** (no `0x1901`, no domain separator) against the depositor's **secp256k1** key, which
no connected ETH wallet can produce. This PR domain-wraps the digest, binds the depositor
approval to a pinned ETH identity via `ecrecover`, and builds out the redeem/renew
transaction path.

Stacked on top of #3882 (`feat/psbt-covenant-final-project-pr`) — the only branch carrying
`pkg/covenantsigner/`. Tracked by tlabs-xyz/vba-dashboard#172:
https://github.com/tlabs-xyz/vba-dashboard/issues/172

## What changed

**A1 - Domain-wrap the digest (EIP-712 v2, off-chain / salt-based).**
The approval digest is now `keccak256(0x1901 ‖ domainSeparator ‖ structHash)` with domain
`EIP712Domain(string name,string version,uint256 chainId,bytes32 salt)` (no
`verifyingContract`, since verification is off-chain). `name = "tBTC Covenant Artifact
Approval"`, `version = "2"`, `chainId` + `salt` from signer config (default salt =
`keccak256("tBTC Covenant Artifact Approval Domain v2")`). Domain params are threaded from
config → validation options → digest, including the exported `ComputeArtifactApprovalDigest`
used by `pkg/tbtc`.

**A2 - ETH identity binding.** Optional `EthAddress` on `DepositorTrustRoot` (keyed by
route/reserve/network). When pinned, the depositor approval is verified via
`ecrecover`-and-compare (65-byte `r‖s‖v`, `v` in {27,28} or {0,1}, low-S enforced);
otherwise it falls back to the secp256k1 script key. Custodian role stays secp256k1.

**A3 - Hard cut to `approvalVersion = 2`** (v1 rejected). Nothing is deployed yet, so no
transition window is needed.

**A4 - Vectors.** Regenerated the v2 approval vectors and added a **real-wallet vector**
whose digest is cross-verified **byte-identical against the independent `eth_account`
(0.13.7) EIP-712 encoder** — genuine `eth_signTypedData_v4` compatibility, not a
self-referential check. A startup guard rejects a config that pins depositor ETH addresses
while `eip712ChainId == 0`.

**A5 - Cooperative REDEEM / RENEW path.** New `CovenantAction` discriminator
(MIGRATION / REDEEM / RENEW; empty defaults to MIGRATION), `RedeemDestinationReservation` /
`RenewDestinationReservation`, redeem/renew commitment schemas (SHA-256 over canonical JSON,
mirroring migration), action-dispatched validation, and an action branch in
`buildCovenantTransactionBuilder`. The built transaction is proven to pay exactly the
committed output — the transaction plan's `destinationValueSats` is bound to the committed
output value.

Commitment schemas (normative contract for the client repos to mirror):
- **redeem**: `{reserve, epoch, route, revealer, vault, network, outputScriptHash, outputValueSats}`
- **renew**: `{reserve, epoch, route, revealer, vault, network, nextCovenantScriptHash, nextMaturityHeight, outputValueSats}`

## Client-side contract (must be mirrored byte-for-byte)

- digest = `keccak256(0x1901 ‖ domainSeparator ‖ structHash)`.
- Struct: `ArtifactApproval(uint8 approvalVersion,bytes32 route,bytes32 scriptTemplateId,bytes32 destinationCommitmentHash,bytes32 planCommitmentHash)`, `approvalVersion = 2`.
- `route` / `scriptTemplateId` are `bytes32 = keccak256(identifier string)` — the dApp must
  pre-hash them; a wallet encoder given the raw bytes32 reproduces the digest.

## Verification

- `go test ./pkg/covenantsigner/...` — 156 passed
- `go test ./pkg/tbtc/...` — 275 passed (full suite)
- `go build ./...` — green; `gofmt` clean
- Headline flow proven end-to-end: a wallet-signed (ETH-identity) cooperative REDEEM is
  accepted through `Service.Submit`, and the signed transaction pays exactly the committed
  redeem/renew output (scriptPubKey + value).

## Out of scope

The section-B/C client sync (covenant-manager, dashboard `engine.ts`, verification kit,
co-sign service, dual-connection wallet UX) lives in other repos. The pre-existing migration
`depositScript`-as-output gap is flagged, not fixed here — the redeem/renew paths sidestep it
by carrying an explicit output scriptPubKey.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -332,6 +332,12 @@ func initCovenantSignerFlags(cmd *cobra.Command, cfg *config.Config) {
 		covenantsigner.Config{}.AuthToken,
 		"Covenant signer provider static Bearer auth token. Required for non-loopback binds; prefer config file or env var over CLI in production.",
 	)
+	cmd.Flags().StringVar(
+		&cfg.CovenantSigner.AdminAuthToken,
+		"covenantSigner.adminAuthToken",
+		covenantsigner.Config{}.AdminAuthToken,
+		"Covenant signer certificate-issuer admin route static Bearer auth token. Distinct from covenantSigner.authToken -- never falls back to it. Required for non-loopback binds; prefer config file or env var over CLI in production.",
+	)
 	cmd.Flags().BoolVar(
 		&cfg.CovenantSigner.EnableSelfV1,
 		"covenantSigner.enableSelfV1",
--- a/cmd/flags_test.go
+++ b/cmd/flags_test.go
@@ -212,6 +212,13 @@ var cmdFlagsTests = map[string]struct {
 		expectedValueFromFlag: "secret-token",
 		defaultValue:          "",
 	},
+	"covenantSigner.adminAuthToken": {
+		readValueFunc:         func(c *config.Config) interface{} { return c.CovenantSigner.AdminAuthToken },
+		flagName:              "--covenantSigner.adminAuthToken",
+		flagValue:             "secret-admin-token",
+		expectedValueFromFlag: "secret-admin-token",
+		defaultValue:          "",
+	},
 	"covenantSigner.enableSelfV1": {
 		readValueFunc:         func(c *config.Config) interface{} { return c.CovenantSigner.EnableSelfV1 },
 		flagName:              "--covenantSigner.enableSelfV1",
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -101,6 +101,8 @@ type startDeps struct {
 		perfMetrics *clientinfo.PerformanceMetrics,
 		minActiveOutpointConfirmations uint,
 		bridgeCovenantFraudDefenseConfirmed bool,
+		eip712ChainID uint64,
+		eip712Salt [32]byte,
 	) (covenantsigner.Engine, error)
 	initializeSigner func(
 		ctx context.Context,
@@ -242,6 +244,13 @@ func startWithDeps(cmd *cobra.Command, deps startDeps) error {
 			btcChain,
 		)
 
+		eip712Salt, err := covenantsigner.ResolveEIP712DomainSalt(
+			clientConfig.CovenantSigner.EIP712Salt,
+		)
+		if err != nil {
+			return fmt.Errorf("error resolving covenant signer EIP-712 salt: [%v]", err)
+		}
+
 		covenantSignerEngine, err := deps.initializeTbtc(
 			ctx,
 			tbtcChain,
@@ -256,6 +265,8 @@ func startWithDeps(cmd *cobra.Command, deps startDeps) error {
 			perfMetrics, // Pass the existing performance metrics instance to avoid duplicate registrations
 			clientConfig.CovenantSigner.MinActiveOutpointConfirmations,
 			clientConfig.CovenantSigner.BridgeCovenantFraudDefenseConfirmed,
+			clientConfig.CovenantSigner.EIP712ChainID,
+			eip712Salt,
 		)
 		if err != nil {
 			return fmt.Errorf("error initializing TBTC: [%v]", err)
--- a/pkg/covenantsigner/approval_vectors_generator_test.go
+++ b/pkg/covenantsigner/approval_vectors_generator_test.go
@@ -1,0 +1,135 @@
+package covenantsigner
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"os"
+	"strings"
+	"testing"
+)
+
+const approvalContractVectorsPath = "testdata/covenant_recovery_approval_vectors_v2.json"
+
+// TestGenerateCovenantRecoveryApprovalVectorsV2 regenerates the canonical
+// approval-contract vectors from the same production code the runtime uses:
+// artifactApprovalDigest for the v2 domain-wrapped approval digest and
+// requestDigest (which normalizes first) for the request digest. Every protocol
+// change that moves either digest — a new approval or certificate version, a
+// changed EIP-712 domain, a new normalization rule — therefore has a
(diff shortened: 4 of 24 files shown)


piotr-roslaniec:
## Security-review follow-ups (deferred from this PR)

The hardening findings addressed in `d3afaa7f7` are self-contained. Three further findings were intentionally **not** included here because each reaches beyond the signer or conflicts with a recent decision. Documenting them so they aren't lost:

### 1. REDEEM/RENEW carry no freshness/authority bound — deferred (needs decision)
MIGRATION is gated by an ed25519 `migrationPlanQuote` with an `expiresAt` check (`validation_quote.go`), which is skipped for REDEEM/RENEW (`validation.go:353`). `ArtifactApprovalPayload` has no expiry field, so a REDEEM/RENEW depositor signature is long-lived and replayable within the epoch. Mitigation present: `activeOutpointTxid/Vout` is bound inside `planCommitmentHash`, pinning the input UTXO.

**Why deferred:** the "cryptographically bind and enforce signer-approval expiry" feature was reverted twice recently (`af236a25a`, `2faf6e764`). Re-introducing a freshness field reverses that deliberate decision and is a protocol change requiring coordinated client + orchestrator updates. Needs a design call before implementation.

**Options:** add an expiry field to the approval payload; require a parallel REDEEM/RENEW plan-quote; or explicitly document that epoch-scoped UTXO binding satisfies the threat model and why.

### 2. `MigrationTransactionPlan` name leaks into REDEEM/RENEW — deferred (cross-repo)
REDEEM payout and RENEW re-lock values bind to `MigrationTransactionPlan.DestinationValueSats`; the struct, fields, and JSON key `migrationTransactionPlan` all say "migration" while serving three actions. Every mirroring client (covenant-manager / dashboard / verification-kit) must send a field literally named "migration" to redeem/renew.

**Why deferred:** a back-compat input alias alone doesn't fix the response side (the server still emits `migrationTransactionPlan`), so client-facing confusion persists until the three client repos are updated together. Belongs in a coordinated cross-repo rename (e.g. to `covenantTransactionPlan`) rather than a signer hardening commit.

### 3. Empty `depositorTrustRoots` falls back to request-supplied key — already addressed
When `depositorTrustRoots` is empty and structured signing is enabled, the depositor key is caller-supplied and only a warning is logged (`server.go:110-118`).

**Status:** this is already enforceable via the existing `requireApprovalTrustRoots` config flag — `validateRequiredApprovalTrustRoots` (`server.go:190`) hard-errors on missing trust roots when the flag is set. The warn-only default is the intentional opt-in posture. Recommend documenting the flag rather than changing the default to a hard error (which would break trustless dev/test deployments).

---

Suggested next steps: track (1) and (2) as separate issues; treat (3) as resolved via config documentation.

lrsaturnino:
2 issues found, both fixed and pushed.

**Fixed: REDEEM/RENEW requests unconditionally rejected when `migrationPlanQuoteTrustRoots` is configured (Major)**
`normalizeRouteSubmitRequest` in `pkg/covenantsigner/validation.go` called `normalizeMigrationPlanQuote` unconditionally, while `validateCommonRequest` already gated that same call on `request.ResolvedAction() == CovenantActionMigration`. Any service configured with migration plan quote trust roots therefore rejected every REDEEM and RENEW submission with `request.migrationPlanQuote is required when migrationPlanQuoteTrustRoots are configured`, even though those actions don't use a migration plan quote. Fixed by applying the same action gate in `normalizeRouteSubmitRequest`, defaulting the normalized quote to nil for REDEEM/RENEW. Commit `91c99ba4f549c155158304c4347edfc145d1ddac`.

**Fixed: wallet-signed (ETH-identity) approval accepted at Submit is rejected at every subsequent Poll (Major)**
`Service.Poll` always set `policyIndependentDigest: true` and never passed `depositorTrustRoots`, so `validateCommonRequest`'s depositor-ETH-address resolution gate was unreachable on the poll path — wallet-signed (ETH ecrecover) approvals fell back to a secp256k1 check that could never verify them, failing every poll after a successful submit. Fixed by resolving and persisting the depositor's ETH address on the Job at Submit time and threading that pinned value into Poll's re-validation via a new `validationOptions.pinnedDepositorEthAddress` field, instead of relying on live trust-root resolution. Commit `db4f6bff55c0916e004fdf16353c0e43bf85bbc5`.

Both fixes are covered by new regression tests (added to `redeem_renew_test.go` and a submit-then-poll test respectively), which fail before the fix and pass after. Full package test suite (including `-race`), `go vet`, `gofmt`, `staticcheck`, and `go build ./...` all pass, along with builds of external consumers (`pkg/tbtc`, `cmd`, `config`).

No unresolved findings.

lrsaturnino (dismissed):
2 findings fixed and pushed; none require follow-up.

lrsaturnino:
Reconciled and tested against #4126 @ b5e11660b1b79543a242d82bc9757c7e5b8e93e4 on 2026-07-30T20:53:12Z.

lrsaturnino (dismissed):
LGTM

piotr-roslaniec:
@lrsaturnino Not merging this yet because it's based on audit remediation branch and I don't want to pollute it

Issue #4171, opened 2026-07-17
Deposit sweep can broadcast at the 1 sat/vByte relay floor and get stuck (no fee margin; sweeps are non-RBF with no fee-bumping)


mswilkison:
## Summary

A live tBTC mainnet wallet has been unable to sweep deposits for ~8h because its main sweep transaction was broadcast at **1 sat/vByte** (the relay floor) and is stuck unconfirmed in the mempool. Because sweeps are **non-RBF** and there is **no fee-bumping / stuck-transaction recovery**, the wallet is fully jammed: the deposits and main UTXO are locked in the unconfirmed tx, so every subsequent sweep attempt fails and eligible deposits pile up behind it.

This is two related gaps: (1) a sweep can be broadcast at the 1 sat/vByte floor with no safety margin, and (2) once stuck, there is no way to bump it.

## Observed incident (mainnet)

- **Wallet:** `0xb4745d903066a2ea7d4ffd75cb282fd9074c0347` (Live)
- **Stuck sweep tx:** [`9037bb979aef05a388414517a768e6a9370f6e12b2176796312ce1b34ca234d0`](https://mempool.space/tx/9037bb979aef05a388414517a768e6a9370f6e12b2176796312ce1b34ca234d0)
  - 3 inputs (main UTXO + two deposits) → 1 output (~861 BTC consolidated)
  - fee: **309 sat over 308 vB = 1.0 sat/vByte**
  - **nSequence = 0xffffffff on all inputs → not RBF-replaceable**
  - unconfirmed ~8h; network was uncongested the entire time (economy ~2, fastest ~5 sat/vByte)
- **Effect:** the two swept deposits show `sweptAt == 0` on the Bridge (no SPV proof possible until confirmation); a further ~12 eligible deposits cannot be swept because the main UTXO is locked in the stuck tx. Wallet `deposit_sweep_failed` increments fleet-wide every coordination window.

## Root cause in code

1. **Fee is floored at 1 sat/vByte with no margin.** `pkg/bitcoin/electrum/electrum.go`, `convertBtcKbToSatVByte`:

   ```go
   // Make sure the minimum returned sat/vbyte fee is always 1.
   satPerVByte = math.Max(satPerVByte, 1)
   ```

[...]

PR #4172, opened 2026-07-17
Enforce a minimum fee rate for deposit sweep transactions

mswilkison:
Fixes part **A** of #4171: a deposit sweep could be broadcast at the ~1 sat/vByte relay floor (a low-but-valid oracle estimate in an uncongested mempool), and since sweeps are not RBF-enabled it could get stuck and jam the wallet. Observed live on mainnet.

## Change

`estimateDepositsSweepFee`:

- **Errors** (instead of silently lowering the fee) if the raw estimate exceeds the Bridge maximum (uneconomical), or if the minimum safe fee (`minSweepTxSatPerVByteFee`, 5 sat/vByte) exceeds the maximum — so a sweep is never broadcast below the intended floor.
- Otherwise applies `max(floor, ceil(rate × 1.25))`: a 25% buffer over the oracle estimate to keep a margin during the estimate-to-broadcast delay and stay adaptive under congestion (per the #4171 reference design), floored at 5 sat/vByte, bounded above by the Bridge maximum.

The floor constant documents that it (and the buffer) is a stopgap for the current non-RBF path and should be revisited when RBF lands; the P2SH-vsize interaction is noted at the floor site.

## Tests

Unit test asserts a low estimate is floored to 5, an estimate above the floor is buffered by 25%, and a Bridge maximum below the floor returns an error (message-pinned to the floor branch). The deposit-sweep scenario testdata is updated for the buffered fee. `go test ./pkg/tbtcpg/...` passes.

## Follow-ups

- **#4179** — applies the same floor to redemptions, moving funds, and moved funds sweeps via a shared helper (stacked on this PR).
- **#4180** — follower-side soft check that a leader's proposed sweep fee is not below the floor (log-only; full enforcement belongs on-chain in `WalletProposalValidator`). Stacked on #4179.
- **Part B of #4171 (RBF + fee-bumping)** — the durable recovery fix; a stuck sweep should be replaceable rather than jamming the wallet. Larger change, targeted at FROST/ROAST. The static floor + buffer here is a stopgap and should be revisited when RBF lands.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved deposit sweep fee estimation to avoid underpriced, non-RBF-able sweeps stalling in the mempool.
  - Introduced a minimum safe fee-rate floor and updated failure behavior when safe fees can’t fit within the Bridge maximum.
  - Added a 25% buffered fee-rate (rounded up), enforced minimum/maximum caps, and updated informational sat/vByte reporting.
  - Enhanced proposal-time warning logs when minimum-safe-fee behavior prevents fee selection.
- **Tests**
  - Added coverage for minimum-floor, buffer, and maximum-cap fee behaviors, including rounding and error cases.
  - Updated the expected sweep fee in the proposal scenario test data.

### DIFF
--- a/pkg/tbtcpg/deposit_sweep.go
+++ b/pkg/tbtcpg/deposit_sweep.go
@@ -31,6 +31,14 @@ const depositScriptByteSize = 126
 // safely above the relay floor while remaining far below the Bridge's
 // per-deposit maximum fee. The value is intentionally conservative and could be
 // made configurable; see threshold-network/keep-core#4171.
+//
+// NOTE: this static floor and the 25% buffer applied below are a stopgap for
+// the current fire-and-forget, non-RBF sweep path: because a stuck sweep cannot
+// be fee-bumped, the fee must be right on the first broadcast. Once RBF /
+// fee-bumping lands (Part B, tracked in #4171) the safety net shifts to
+// monitor-and-bump, and this policy should be revisited rather than carried
+// forward unchanged: the defensive buffer can be dropped and the floor relaxed
+// toward the live estimate, keeping only a small relay-propagation minimum.
 const minSweepTxSatPerVByteFee = 5
 
 // DepositSweepLookBackBlocks is the look-back period in blocks used
@@ -677,6 +685,13 @@ func estimateDepositsSweepFee(
 	// (see threshold-network/keep-core#4171), then enforce the minimum floor and
 	// bound the result by the Bridge maximum (which the floor cannot exceed, per
 	// the check above).
+	//
+	// Caveat: transactionSize assumes all deposit inputs are witness (P2WSH), per
+	// this function's doc comment. A sweep that includes legacy P2SH deposits has
+	// a larger on-wire vsize than estimated here, so the effective on-wire rate
+	// can land slightly below the floor for such (rare) sweeps. It still dominates
+	// the 1 sat/vByte relay floor this fix targets; a fully accurate floor would
+	// require deposit-type-aware sizing.
 	rate := totalFee / transactionSize
 	rate = (rate*5 + 3) / 4 // ceil(rate * 1.25)
 	if rate < minSweepTxSatPerVByteFee {
--- a/pkg/tbtcpg/deposit_sweep.go
+++ b/pkg/tbtcpg/deposit_sweep.go
@@ -650,19 +650,29 @@ func estimateDepositsSweepFee(
 		return 0, 0, fmt.Errorf("cannot estimate transaction fee: [%v]", err)
 	}
 
-	// Clamp the estimated fee up to a safe minimum so the sweep is never
-	// broadcast near the 1 sat/vByte relay floor, which can leave it stuck in
-	// the mempool and jam the wallet (see minSweepTxSatPerVByteFee). The
-	// Bridge's maximum-fee check below still bounds the result from above.
-	if minTotalFee := minSweepTxSatPerVByteFee * transactionSize; totalFee < minTotalFee {
-		totalFee = minTotalFee
-	}
-
 	// Compute the maximum possible total fee for the entire sweep transaction.
 	totalMaxFee := uint64(depositsCount) * perDepositMaxFee
 
+	// A genuine fee estimate that exceeds the Bridge maximum is an error (the
+	// sweep is not economical to perform). This is checked before the minimum
+	// below is applied so that raising the fee to the minimum can never itself
+	// trigger this error.
 	if uint64(totalFee) > totalMaxFee {
 		return 0, 0, fmt.Errorf("estimated fee exceeds the maximum fee")
+	}
+
+	// Clamp the estimated fee up to a safe minimum so the sweep is never
+	// broadcast near the 1 sat/vByte relay floor, which can leave it stuck in
+	// the mempool and jam the wallet (see minSweepTxSatPerVByteFee). The minimum
+	// is itself bounded by the Bridge maximum above, so it can never exceed the
+	// cap. In practice the minimum is orders of magnitude below the cap; the
+	// bound only guards against future parameter changes.
+	minTotalFee := minSweepTxSatPerVByteFee * transactionSize
+	if uint64(minTotalFee) > totalMaxFee {
+		minTotalFee = int64(totalMaxFee)
+	}
+	if totalFee < minTotalFee {
+		totalFee = minTotalFee
 	}
 
 	// Compute the actual sat/vbyte fee for informational purposes.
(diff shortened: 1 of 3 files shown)


pkg/tbtcpg/deposit_sweep.go
```diff
@@ -641,10 +653,28 @@ func estimateDepositsSweepFee(
...
 
+	// Clamp the estimated fee up to a safe minimum so the sweep is never
+	// broadcast near the 1 sat/vByte relay floor, which can leave it stuck in
+	// the mempool and jam the wallet (see minSweepTxSatPerVByteFee). The minimum
+	// is itself bounded by the Bridge maximum above, so it can never exceed the
+	// cap. In practice the minimum is orders of magnitude below the cap; the
+	// bound only guards against future parameter changes.
+	minTotalFee := minSweepTxSatPerVByteFee * transactionSize
+	if uint64(minTotalFee) > totalMaxFee {
+		minTotalFee = int64(totalMaxFee)
```
lrsaturnino: This silently lowers the fee floor to the cap. When `totalMaxFee` is below `5 × transactionSize`, `minTotalFee` becomes the cap here, so the sweep can still go out between 1 and 5 sat/vByte — below the floor this PR is meant to enforce. It should return an error instead, the same way the check above (line 660) does.

pkg/tbtcpg/deposit_sweep.go
```diff
@@ -641,10 +653,28 @@ func estimateDepositsSweepFee(
...
 		return 0, 0, fmt.Errorf("estimated fee exceeds the maximum fee")
 	}
 
+	// Clamp the estimated fee up to a safe minimum so the sweep is never
+	// broadcast near the 1 sat/vByte relay floor, which can leave it stuck in
+	// the mempool and jam the wallet (see minSweepTxSatPerVByteFee). The minimum
+	// is itself bounded by the Bridge maximum above, so it can never exceed the
+	// cap. In practice the minimum is orders of magnitude below the cap; the
+	// bound only guards against future parameter changes.
+	minTotalFee := minSweepTxSatPerVByteFee * transactionSize
```
lrsaturnino: This is a flat `max(5, rawRate)` floor — once the oracle rate hits 5 sat/vByte or higher, it adds no protection above the raw estimate. The #4171 spec recommends `max(5, ceil(rawRate × 1.25))` instead, so there's a margin above the oracle estimate during the estimate-to-broadcast delay, and the fee stays adaptive under congestion. Worth adopting here if the team is treating that spec as the reference design.

pkg/tbtcpg/deposit_sweep_fee_test.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"testing"
+
+	"github.com/keep-network/keep-core/pkg/tbtcpg"
+)
+
+// TestEstimateDepositsSweepFee_MinimumFloor verifies that an unusably low fee
+// estimate is raised to the minimum sweep fee rate (see minSweepTxSatPerVByteFee,
+// currently 5 sat/vByte), while a healthy estimate is left unchanged.
+func TestEstimateDepositsSweepFee_MinimumFloor(t *testing.T) {
```
lrsaturnino: Both cases here (raw 1→5, raw 20→20) use a cap of 100,000 sats, far above any tested fee, so neither one exercises the cap-regression path. Please add a case where the cap sits below `5 × transactionSize` and assert it returns an error, not a lowered fee.

lrsaturnino:
This PR only covers deposit sweeps. `EstimateRedemptionFee` (redemptions.go:503), `EstimateMovingFundsFee` (moving_funds.go:659), and `EstimateMovedFundsSweepFee` (moved_funds_sweep.go:414) still return the raw oracle fee with no minimum at all. Recommend a follow-up PR applying the same floor to all three — this one is already scoped to deposit sweeps only, so it's not a blocker for merging this PR itself.

lrsaturnino:
Two small things in the description are out of date with the diff: it says the floor is applied "before" the Bridge maximum-fee check, but the code checks the cap first (line 660) and floors after (line 670). It also says the low-fee test was "flagged rather than included," but `deposit_sweep_fee_test.go` is already in this PR and covers that case. Worth a quick update so reviewers aren't working from a stale description.

lrsaturnino:
This closes the fee-floor gap on the generation side ([`estimateDepositsSweepFee`](https://github.com/threshold-network/keep-core/blob/f630e8df429c8f0db7f05319eb03b96ec8217af1/pkg/tbtcpg/deposit_sweep.go#L629-L693)), but nothing enforces it on the follower-validation side. The Go-side [`ValidateDepositSweepProposal`](https://github.com/threshold-network/keep-core/blob/f630e8df429c8f0db7f05319eb03b96ec8217af1/pkg/tbtc/deposit_sweep.go#L449-L458) does no independent fee check — it delegates to the chain validator, which on Ethereum calls the deployed [`WalletProposalValidator`](https://github.com/threshold-network/tbtc-v2/blob/b16c7b300d1b8d15a89fe3ab915f4d0e841c1022/solidity/contracts/bridge/WalletProposalValidator.sol#L359-L383). That contract only requires `sweepTxFee > 0` plus the existing upper bound — no minimum. A follower accepts the leader's proposed fee and signs without re-running the estimator ([coordination.go#L728-L815](https://github.com/threshold-network/keep-core/blob/f630e8df429c8f0db7f05319eb03b96ec8217af1/pkg/tbtc/coordination.go#L728-L815), [deposit_sweep.go#L196-L239](https://github.com/threshold-network/keep-core/blob/f630e8df429c8f0db7f05319eb03b96ec8217af1/pkg/tbtc/deposit_sweep.go#L196-L239)), so an unpatched or misbehaving leader can still get a sweep at the ~1 sat/vByte client-side floor signed — the #4171 scenario — as long as it's the one proposing.

lrsaturnino:
The fee floor/buffer this PR adds is computed off a P2WSH-only deposit-input estimate — `EstimateDepositsSweepFee` always calls `AddScriptHashInputs(depositsCount, depositScriptByteSize, true)` (isWitness=true), and the function's own doc comment already notes that real sweeps can contain legacy P2SH deposits, which "may be underestimated in some rare cases" (`pkg/tbtcpg/deposit_sweep.go:561-563`, `638`).

That pre-existing underestimation interacts directly with the new floor: since the floor/buffer is applied to the *estimated* vsize rather than the actual on-wire vsize, a sweep with a meaningful share of P2SH deposits can still be assembled at an effective rate well below the intended 5 sat/vByte.

piotr-roslaniec (dismissed):

lrsaturnino (dismissed):
LGTM

piotr-roslaniec (approved):

PR #4177, opened 2026-07-19
fix: quiet noisy WARN logs, add join-failure observability

lrsaturnino:
## Problem

keep-core's release pipeline carries several long-standing, non-blocking log-noise and observability gaps around three specific areas. The periodic well-known-peer connection supervisor logged a binary "all bootstrap attempts failed" warning whenever every currently-disconnected peer failed to dial, even on a node connected to the large majority of its well-known peers, so the message routinely misrepresented a healthy node as isolated. The Bitcoin RPC health check logged a warning on every occurrence of a known-benign "invalid params" response from Electrum servers that don't support a particular parameter, flooding logs without signaling anything actionable. And the inbound connection handshake path silently swallowed the specific cause of every join failure into a single undifferentiated counter, with no log output at production levels to distinguish a timeout from a connection reset from a firewall rejection from a genuine on-chain RPC error.

## Solution

The well-known-peer connection log now reports the live unreachable/total ratio measured after all dials complete, with warning severity reserved for the case where the node ends a round connected to zero well-known peers; anything less severe than total isolation logs at info. The Bitcoin health check downgrades the specific known-benign Electrum error to debug level while keeping every other RPC error at warning, and tracks the benign case through a new counter so it stays observable without spamming logs. The inbound (and, for logging purposes, outbound) connection path now classifies every failure into one of five low-cardinality reasons, exposes a matching per-reason counter alongside the existing aggregate, and logs the remote address, peer ID, and reason on every failure branch. A new counter isolates the on-chain firewall-validation call volume from cache hits, and a Grafana panel plus a Prometheus alert cover the join-request metric family, with the alert gated on a failure-rate burst combined with peer loss or coordination degradation rather than the raw failure ratio alone, since a high ratio can be expected, correct behavior.

## Tests

Each change has new or extended unit test coverage in the package it touches: connectivity-ratio and severity behavior for the well-known-peer log across full, partial, and zero connectivity; the benign-versus-genuine Bitcoin RPC error distinction; failure classification and per-reason counter wiring for the inbound/outbound connection path; and the on-chain-check counter's cache-hit-versus-live-call distinction. gofmt, go vet, go build, staticcheck, gosec, and the full Go test suite all pass against these changes (verified locally against the same commands the CI workflow runs); the two pre-existing findings the broader repo carries outside these files were confirmed unrelated to this change and are unaffected by it.

## Summary by CodeRabbit

* **New Features**
  * Added detailed network join request metrics, including success and failure reasons.
  * Added Bitcoin RPC health metrics for known benign header errors.
  * Improved connection-failure logging with clearer classifications and rate limiting.
  * Enhanced bootstrap connectivity reporting for isolated and partially connected nodes.

* **Monitoring**
  * Added Grafana visualizations for network join requests.
  * Added warnings for bursts of join failures alongside connectivity degradation.

* **Bug Fixes**
  * Known benign Bitcoin header errors are now logged at debug level while health status remains accurate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,4 +68,14 @@ tmp/
 out/
 data/
 logs/
-storage/
+storage/.ralph/
+__pycache__/
+.venv/
+venv/
+.pytest_cache/
+.mypy_cache/
+.ruff_cache/
+target/
+dist/
+build/
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -68,7 +68,8 @@ tmp/
 out/
 data/
 logs/
-storage/.ralph/
+storage/
+.ralph/
 __pycache__/
 .venv/
 venv/
--- a/.gitignore
+++ b/.gitignore
@@ -78,5 +78,4 @@ venv/
 .ruff_cache/
 target/
 dist/
-build/
 .DS_Store
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -104,6 +104,10 @@ func start(cmd *cobra.Command) error {
 		}); ok {
 			setter.SetMetricsRecorder(perfMetrics)
 		}
+
+		// Wire performance metrics into firewall validation so live
+		// on-chain IsRecognized calls are counted.
+		firewall.SetMetricsRecorder(perfMetrics)
 	}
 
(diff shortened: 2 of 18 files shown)


lionakhnazarov (approved):
LGTM

PR #4178, opened 2026-07-20
test(covenantsigner): add missing signer defensive-property coverage

piotr-roslaniec:
## What

Adds five regression tests closing genuinely-missing defensive-property coverage for the
covenant signer. Stacked on `feat/psbt-covenant-final-project-pr` (#3882); tests only, no
production changes.

## Why

A coverage review of the signer's defensive properties (auth, body cap, timeouts, replay,
reorg guard, approval-certificate validation) started from a ~27-item wishlist but found that
**~22 already exist** in the suite. Rather than duplicate them, this PR adds only the handful
that were actually absent, plus one boundary complement.

## Tests added

**`pkg/tbtc/signer_approval_certificate_test.go`**
- `TestVerifySignerApprovalCertificateRejectsUnsupportedVersion` — the `certificateVersion == 1`
  hard gate (`signer_approval_certificate.go:230`) had no test; rejects both `0` and `2`.
- `TestVerifySignerApprovalCertificateRejectsUnsupportedSignatureAlgorithm` — the
  `tecdsa-secp256k1` algorithm gate (`:233`) had no test.

**`pkg/tbtc/covenant_signer_test.go`**
- `TestEnsureActiveOutpointFinalityPropagatesProviderError` — the finality guard's error path
  (`covenant_signer.go:614`) was untested; existing tests only cover known confirmation counts
  above/below the threshold. Asserts fail-closed when confirmations cannot be determined.

**`pkg/covenantsigner/server_test.go`**
- `TestServerAcceptsBodyExactlyAtLimit` — complements the existing `maxRequestBodyBytes+1`
  rejection in `TestServerBoundaryErrorMatrix` by pinning the inclusive boundary (a body of
  exactly `maxRequestBodyBytes` must pass the size guard).

**`pkg/covenantsigner/store_lock_test.go`**
- `TestNewServiceReleasesLockOnInitFailure` — asserts `NewService` releases the store file lock
  when a later init step fails (`service.go:135-142` defer), so a signer that failed to start
  once can restart against the same data dir. Only the store-level `Close()` release was tested.

Dropped from the original list as **already covered**: the non-loopback-without-token startup
guard (already asserted inside `TestInitializeRejectsInvalidOrUnavailablePort`), Poll expiry
checks, confirmation thresholds/boundaries, dedup, lock contention, digest/signer-set/high-S/DER
cert rejections, trust-root pinning, quote expiry, oversized-body, and healthz-exempt.

## Verification

- `go test ./pkg/tbtc/ ./pkg/covenantsigner/` — pass.
- `go test -race ./pkg/covenantsigner/` — pass.
- `gofmt -l` clean, `go vet` clean.
- Note: `go test -race ./pkg/tbtc/` trips a **pre-existing** data race in
  `TestSigningExecutor_SignBatch` (reproduces on this branch's base tip `af236a25a` with none of
  these changes; unrelated to the covenant signer). Flagging, not fixing here.

## Open questions for @lrsaturnino

Two items surfaced during the review that need an owner decision; they gate two further tests
not included here:

1. **Signer-approval expiry binding.** The three tip commits (`e572ea3a8` / `a882cf142` /
   `ccbc4c3ac`, "cryptographically bind + enforce signer-approval expiry") were reverted.
   `SignerApprovalCertificate.EndBlock` is still checked at Poll (`service.go:275-290`), but the
   ECDSA signature covers `approvalDigest` only — `EndBlock` is not inside the signed digest, so
   its value is caller-settable. Is the binding intended to re-land, or is Poll-time check-only
   the accepted design? (Determines whether we add a `TestEndBlockIsBoundToSignature`.)

2. **`chainId == 0` guard on the EIP-712 leg.** On the follow-up `feat/covenant-eip712-eth-leg`
   (#4169), `EIP712ChainID` is folded into the EIP-712 domain separator, but nothing rejects
   `EIP712ChainID == 0` at startup — a misconfig fails soft (no real `eth_signTypedData_v4`
   signature is ever produced under chainId 0, so verification just never matches). Intentional,
   or should we add a startup guard + test?

### DIFF
--- a/pkg/covenantsigner/server_test.go
+++ b/pkg/covenantsigner/server_test.go
@@ -1080,3 +1080,64 @@ func TestSubmitHandlerPreservesServiceContextValues(t *testing.T) {
 		)
 	}
 }
+
+// TestServerAcceptsBodyExactlyAtLimit asserts the request-body size guard is an
+// inclusive cap: a body of exactly maxRequestBodyBytes must pass, since
+// http.MaxBytesReader only rejects bodies strictly larger than the limit. This
+// complements the maxRequestBodyBytes+1 rejection in TestServerBoundaryErrorMatrix
+// and pins the exact boundary so an off-by-one in the cap would be caught.
+func TestServerAcceptsBodyExactlyAtLimit(t *testing.T) {
+	handle := newMemoryHandle()
+	service, err := NewService(handle, &scriptedEngine{
+		submit: func(*Job) (*Transition, error) {
+			return &Transition{State: JobStatePending, Detail: "queued"}, nil
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	server := httptest.NewServer(newHandler(service, context.Background(), "test-token", true))
+	defer server.Close()
+
+	// Build a well-formed submit envelope padded to exactly maxRequestBodyBytes
+	// by stretching the facadeRequestId string field.
+	prefix := `{"routeRequestId":"ors_exact","stage":"SIGNER_COORDINATION","request":{"facadeRequestId":"`
+	suffix := `"}}`
+	pad := maxRequestBodyBytes - len(prefix) - len(suffix)
+	if pad <= 0 {
+		t.Fatalf("test assumption broken: prefix+suffix already exceed the body limit")
+	}
+	body := []byte(prefix + strings.Repeat("a", pad) + suffix)
+	if len(body) != maxRequestBodyBytes {
+		t.Fatalf("expected body of exactly %d bytes, got %d", maxRequestBodyBytes, len(body))
+	}
+
+	request, err := http.NewRequest(
+		http.MethodPost,
+		server.URL+"/v1/self_v1/signer/requests",
+		bytes.NewReader(body),
+	)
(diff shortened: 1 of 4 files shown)


lrsaturnino:
This is a clean, test-only PR — four `_test.go` files, +214/-0, no production code changed. All three claimed coverage gaps are genuine, and all five new tests exercise the code paths they claim.

Two non-blocking suggestions:
- `server_test.go:939` — the new boundary test pins the cap at exactly N bytes, but the existing oversized-body case is actually N+92 bytes, not N+1 as its comment implies, leaving a 91-byte band untested in the loosen-the-cap direction.
- `server_test.go:986` — the rejection assertion matches a message shared by two different decode-failure paths; tying it to the typed `SignerSubmitInput` struct (as the neighboring test already does via `mustJSON`) would make future schema drift fail at compile time instead of producing a misleading diagnostic.

lrsaturnino (approved):
Test-only change (four `_test.go` files, +214/-0), no production code touched, no behavioral regressions. Two non-blocking suggestions noted in comments.

lrsaturnino (approved):
LGTM

PR #4179, opened 2026-07-20
Apply the safe minimum fee floor to all wallet transactions

mswilkison:
Follows up on #4172 (lrsaturnino review): the fee floor there covers deposit sweeps only. `EstimateRedemptionFee`, `EstimateMovingFundsFee`, and `EstimateMovedFundsSweepFee` still returned the raw oracle fee, so a redemption, moving-funds, or moved-funds-sweep transaction could be broadcast at the ~1 sat/vByte relay floor and jam the wallet the same way a sweep can.

> **Stacked on #4172** (based on branch `fix/deposit-sweep-min-fee`), so this PR's diff shows only its incremental change: the shared helper plus its application to the three other fee types. Merge #4172 first.

## Change

- Extracts the 25% buffer + minimum floor + Bridge-max bound into a shared `applyWalletTxFeeFloor` helper and a shared `minWalletTxSatPerVByteFee` constant (`pkg/tbtcpg/fee.go`).
- Refactors deposit-sweep fee estimation onto the helper (no behavior change; identical math).
- Applies the helper to redemptions, moving funds, and moved funds sweeps.
- `EstimateRedemptionFee` now takes the redemption tx max total fee so the floor can be bounded by it; the caller fetches it from `GetRedemptionParameters`.

## Note on redemption scope

Redemption fee estimation previously deferred the fee cap entirely to on-chain validation. To apply a client-side floor safely, it now also bounds the fee by the max total fee (clamp, not reject) so the floor cannot produce a fee the Bridge would reject, and errors if the raw estimate already exceeds the cap (consistent with the other transaction types) rather than broadcasting an underpriced transaction. The per-request max fee is still validated on-chain.

## Tests

- New `TestApplyWalletTxFeeFloor` covers buffer / floor / cap-clamp / floor-over-cap error / invalid-vsize.
- Redemption, moving-funds, and moved-funds-sweep fee and propose tests updated for the buffered values.
- `go test ./pkg/tbtcpg/...` passes; `go build ./...` and `go vet` clean.

## Follow-ups still open from the #4172 review

- Follower-side / on-chain enforcement of the floor (a leader can still propose a low fee; validators do not re-check the minimum) — being handled separately because it is rollout-sensitive.

### DIFF
--- a/pkg/tbtcpg/deposit_sweep.go
+++ b/pkg/tbtcpg/deposit_sweep.go
@@ -21,26 +21,6 @@ import (
 // This will ensure that deposit sweep transaction fees are not underestimated.
 const depositScriptByteSize = 126
 
-// minSweepTxSatPerVByteFee is the minimum fee rate, in sat/vByte, applied to
-// deposit sweep transactions. A fee oracle can return an unusably low estimate
-// (down to the 1 sat/vByte relay floor enforced by the Electrum client) in an
-// uncongested mempool. Because a sweep consolidates significant wallet value
-// and is not RBF-enabled, it cannot be replaced once broadcast, so a floor-rate
-// sweep can get stuck in the mempool and jam the wallet: no new sweep can be
-// built while the previous one is unconfirmed. This minimum keeps the sweep fee
-// safely above the relay floor while remaining far below the Bridge's
-// per-deposit maximum fee. The value is intentionally conservative and could be
-// made configurable; see threshold-network/keep-core#4171.
-//
-// NOTE: this static floor and the 25% buffer applied below are a stopgap for
-// the current fire-and-forget, non-RBF sweep path: because a stuck sweep cannot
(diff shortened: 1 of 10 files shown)


piotr-roslaniec:
Closing in favor of #4192, which carries this work forward rebased cleanly onto current `main` and folds in fixes from a multi-agent review of the change:

- restored the fee-floor truncation invariant dropped during the shared-helper extraction,
- corrected the `estimatedFee`/`txVsize` unit docs,
- added a redemption diagnostic WARN when a floored per-request fee share would exceed the per-request `TxMaxFee`, and
- added test coverage for the redemption raw-estimate-above-cap error and the moving-funds / moved-funds-sweep floor-clamp paths.

#4192 is on a `threshold-network` branch so it can be iterated on directly. Thanks @mswilkison for the original implementation.

PR #4180, opened 2026-07-20
Add follower-side soft check for below-floor sweep fees

mswilkison:
Follows up on #4172 / #4179 (lrsaturnino review): the fee floor is enforced on the fee-*generation* side, but nothing enforces it on the follower/validation side. `ValidateDepositSweepProposal` delegates to the chain validator, and the on-chain `WalletProposalValidator` only bounds the sweep fee from above (plus `fee > 0`) — no minimum. So a misbehaving or unpatched coordination *leader* can propose a sweep at the ~1 sat/vByte relay floor and get patched followers to sign it — the #4171 jam scenario.

> **Stacked on #4179 → #4172** (based on branch `feat/floor-all-wallet-tx-fees`), so this PR's diff shows only its incremental change: the follower-side soft check in `pkg/tbtc/deposit_sweep.go`. Merge #4172 then #4179 first.

## Change

`ValidateDepositSweepProposal` now recomputes the safe minimum sweep fee (same vsize estimate + floor as the generator) and logs a warning if the proposed fee is below it.

**This is intentionally log-only, not a rejection.** Rejecting a below-floor proposal here would, during a mixed-version rollout, split signers (patched nodes reject, unpatched nodes sign) and could stall signing below threshold. This PR gives operators *detection* of an underpricing leader without a liveness risk.

## Why not hard enforcement here?

Hard enforcement of a fee minimum belongs where all signers apply the same rule with no version skew — **on-chain in `WalletProposalValidator` (tbtc-v2 repo)**, or behind a coordinated all-nodes keep-client upgrade. This PR is the safe keep-client-side increment; the on-chain change is the tracked follow-up.

## Notes

- The floor constant and deposit-script size are mirrored into `pkg/tbtc` (with keep-in-sync comments) because `pkg/tbtcpg` imports `pkg/tbtc`, so this package cannot import the canonical values without a dependency cycle.
- The check is log-only and its arithmetic mirrors the (tested) generator estimator, so no bespoke test is added — a dedicated test would require standing up the full `ValidateDepositSweepProposal` mock harness purely to assert a log line.

## Summary by CodeRabbit

* **Improvements**
  * Improved transaction fee estimation for deposits, redemptions, moved funds, and moving funds.
  * Added minimum safety floors and buffering to help prevent underpriced transactions.
  * Fee estimates now respect configured maximum limits and report an error when safe fees cannot fit within those limits.
  * Added validation warnings for unusually low proposed deposit sweep fees without rejecting valid proposals.
* **Tests**
  * Expanded coverage for fee floors, buffers, maximum limits, and error conditions.

### DIFF
--- a/pkg/tbtc/deposit_sweep.go
+++ b/pkg/tbtc/deposit_sweep.go
@@ -49,6 +49,16 @@ const (
 	// the transaction is known on the Bitcoin chain. This delay is needed
 	// as spreading the transaction over the Bitcoin network takes time.
 	depositSweepBroadcastCheckDelay = 1 * time.Minute
+	// minSweepTxSatPerVByteFee mirrors tbtcpg.minWalletTxSatPerVByteFee, the safe
+	// minimum sweep fee rate. It is duplicated here because pkg/tbtcpg imports
+	// pkg/tbtc, so this package cannot import the canonical constant without a
+	// dependency cycle; keep the two in sync. It backs a follower-side soft
+	// (log-only) check that the leader's proposed sweep fee is not below the
+	// floor (see threshold-network/keep-core#4171).
+	minSweepTxSatPerVByteFee = 5
+	// depositScriptByteSize mirrors tbtcpg.depositScriptByteSize, the worst-case
+	// deposit script size used to estimate the sweep transaction virtual size.
+	depositScriptByteSize = 126
 )
 
 // DepositSweepProposal represents a deposit sweep proposal issued by a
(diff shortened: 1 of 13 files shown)


piotr-roslaniec:
Superseded by #4194. This branch's head lived on a fork that drifted onto an old `main` and went `CONFLICTING`. #4194 is the same work (follower-side soft check + review follow-ups) rebased clean onto current `main`. Continuing there.

PR #4170, opened 2026-07-17
fix(electrum): replace dead fulcrum tcp integration test server

lionakhnazarov:
## Summary

- replace the dead `fulcrum tcp` public test server used by the Electrum integration test suite
- `v22019051929289916.bestsrv.de:50001` refuses connections, failing every test that connects through it (`TestConnect_Integration`, `TestGetTransaction_Integration`, `TestGetLatestBlockHeight_Integration`, etc.) via `newTestConnection`'s `t.Fatal` on connect failure
- this has been failing `client-integration-test` on `main` for at least the last 5 nightly runs, unrelated to any recent feature work

## Fix

Swapped the dead hostname for `testnet.aranguren.org:51001`, a currently-live public testnet3 Fulcrum server (confirmed via a `server.version` JSON-RPC handshake), keeping the same "fulcrum tcp" category/transport this test config represents.

## Testing

- `go build -tags integration ./pkg/bitcoin/electrum/...`
- `go test -tags integration ./pkg/bitcoin/electrum/... -run 'TestConnect_Integration/fulcrum_tcp$' -v` → PASS against the new endpoint

## Summary by CodeRabbit

* **Tests**
  * Updated the Electrum test environment to use a new testnet server endpoint.
  * No changes were made to application behavior or test coverage.

PR #4181, opened 2026-07-20
fix(electrum): set Network on all integration test configs

lrsaturnino:
## Problem

The nightly and per-PR `client-integration-test` job keeps failing on `TestEstimateSatPerVByteFee_Integration/embedded/testnet4/ssl://mempool.space:40002` with `errNo: -32603, errMsg: cannot estimate fee for 1008 blocks`. mempool.space's testnet4 Electrum endpoint intermittently has no fee data and rejects every confirmation target in the fallback ladder, so the failure comes and goes between runs; it hit four of the last six nightly runs on `main`, including the run right after #4170 removed the dead fulcrum server, and is the last remaining breaker of this job.

The client already handles this exact situation: when every target fails with a fee-oracle error, `EstimateSatPerVByteFee` falls back to a fixed low feerate, but only when `Config.Network` identifies a test network. The integration harness was supposed to provide that via an `init` that copied each test config's network into its Electrum client config. Go runs same-file `init` functions in source order, and that propagation `init` sits above the `init` that registers the embedded servers, so it runs first: the four static configs get their `Network` set while every `embedded/*` entry, including the testnet4 one, connects with `Network` left at its zero value (`bitcoin.Unknown`), which fails closed like mainnet and surfaces the oracle error instead of falling back.

## Solution

Set `Network` explicitly in every `clientConfig` literal, both in the static entries and in the embedded-server registration loop, mirroring production where the network is injected during config resolution. The propagation `init` is removed so registration order can no longer silently strand an entry without a network, and a comment on `testConfigs` records the constraint. With the network populated, an all-targets oracle failure on testnet4 resolves to the designed low-fee fallback and the test's minimum-fee assertion passes deterministically, while mainnet entries keep today's fail-closed behavior.

## Tests

`gofmt` is clean and `go vet` passes with and without the `integration` tag, and the package unit tests pass. The full `TestEstimateSatPerVByteFee_Integration` suite was run locally against all live servers: all six subtests pass, including the previously failing testnet4 endpoint, which resolved through the network-gated looser-target path. The fallback branch itself is already unit-covered for testnet4 in `electrum_test.go`; this change makes the integration harness actually reach it.

## Summary by CodeRabbit

* **Tests**
  * Improved integration test configuration by explicitly specifying the Bitcoin network for each Electrum server.
  * Updated embedded server test setup and comments to ensure network-specific behavior is validated consistently.

### DIFF
--- a/pkg/bitcoin/electrum/electrum_integration_test.go
+++ b/pkg/bitcoin/electrum/electrum_integration_test.go
@@ -42,25 +42,21 @@ type testConfig struct {
 	network      bitcoin.Network
 }
 
-// init propagates each test config's Bitcoin network into its Electrum
-// connection config. In production the network is injected during config
-// resolution; mirroring that here ensures the integration tests exercise the
-// same network-gated behavior (e.g. the low-fee estimate fallback) instead of
-// leaving Config.Network at its zero value (bitcoin.Unknown), which would
-// disable the fallback.
-func init() {
-	for key, tc := range testConfigs {
-		tc.clientConfig.Network = tc.network
-		testConfigs[key] = tc
-	}
-}
-
 // Servers details were taken from a public Electrum servers list published
 // at https://1209k.com/bitcoin-eye/ele.php?chain=tbtc.
+//
+// Every clientConfig must set Network explicitly, mirroring production config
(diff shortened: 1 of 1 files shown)


lionakhnazarov (approved):
approve

PR #4185, opened 2026-07-22
chore: low-risk code-quality cleanup across tBTC/beacon node

piotr-roslaniec:
## Summary

A focused, low-risk code-quality pass across the tBTC/beacon node. Almost all
changes are behavior-preserving cleanups (dedup, named types, dead-code removal,
error wrapping, naming/doc/comment consistency). Two are small, beneficial
functional fixes, called out under **Behavior changes** below. Net **-40 lines**.

Each touched package builds, vets, and passes its test suite (including the full
`pkg/tbtc` suite).

## Changes

- **clientinfo:** split the 240-line `registerAllMetrics` into per-type helpers
  (counters / wallet-actions / histograms / gauges), documenting the two-phase
  map-populate-then-observe concurrency invariant once per helper; remove
  field-group comments that restated field names; fix a stale ticker comment.
- **tbtc:** remove the dead `coordinationFailed` flag (both set-true branches
  return early, so the end guard was always taken); introduce a named
  `DepositKey` type replacing the anonymous struct shared across tbtc/tbtcpg/
  ethereum; extract a shared `movingFundsSafetyMarginChain` interface; switch
  `ParseWalletActionType` on iota constants; collapse three identical
  frequency-window guards; wrap the final-signing-group resolution error.
- **tbtcpg / protocol:** preserve the real error cause in
  `EstimateDepositsSweepFee` and the sync machine (was dropping / formatting the
  wrong value); rename `fnLogger` -> `taskLogger` to match convention; fix two
  interface doc comments to start with the method name.
- **spv:** add and register deposit-sweep proof-submission metric constants
  (mirroring redemptions) and replace raw metric-name strings; drop the
  `getGlobalMetricsRecorder` passthrough wrapper; trim restating variable
  comments.
- **style:** normalize the minority `marshalling` filenames to the majority
  `marshaling` spelling (git mv, no code changes); correct the `tools.go`
  comment to accurately describe the pinned modules as build-time-only
  dependencies (they are direct `go.mod` requires, not indirect).

## Behavior changes

Two changes alter runtime behavior; both are intentional fixes rather than pure
cleanups:

- **`EstimateDepositsSweepFee` now reports the real error cause.** The previous
  message formatted the zero-value `sweepMaxSize` (`cannot get sweep max size:
  [0]`), dropping the actual error; it now wraps `err`.
- **SPV deposit-sweep proof-submission metrics are now observable.** The
  `deposit_sweep_proof_submissions_{total,success_total,failed_total}` counters
  were previously incremented but never registered as observers, so they were
  never exported. Registering them in `registerCounterMetrics` makes them
  observable for the first time.

## Scope / follow-ups

This PR intentionally covers only safe-to-moderate cleanups. Deliberately left
out (larger blast radius, need dedicated review):

- exported/interface **signature changes** (e.g. the 11-value
  `GetMovingFundsParameters`),
- **architectural** refactors (splitting the ethereum adapter / `node.go`
  monoliths, `pkg/tbtc` package layout, metrics DI),
- **dependency** migrations (dual go-log, deprecated addr-util),
- **new test** coverage for untested entry points,
- crypto-path cross-package dedup and concurrency/error-flow behavior changes.

**One accepted exception to the above:** `DepositSweepProposal.DepositsKeys`
(see the **tbtc** bullet under Changes) changes element type from an
anonymous struct to the new named `DepositKey` type. Every in-repo consumer
was updated, so `go build ./...` is clean, but this is a real
source-compatibility break for any code outside this module that constructs
a `DepositSweepProposal` directly from the old anonymous-struct literal —
that code will fail to compile against the new type. This is flagged on the
`DepositKey` godoc as a heads-up for downstream consumers. It's a trivial,
low-risk adaptation and doesn't need dedicated review, but it is a signature
change, not merely an internal cleanup, so it doesn't fit the "deliberately
left out" bucket above without this caveat.

## Testing

- `go build ./...` clean
- `go test` green for every touched package (clientinfo, tbtc, tbtcpg,
  protocol, maintainer/spv, beacon/{registry,dkg,dkg/result}, protocol/inactivity)

## Summary by CodeRabbit

* **New Features**
  * Added protobuf-based serialization for threshold signing, membership, inactivity claims, and DKG result signatures.
  * Added clearer deposit identification in deposit sweep proposals.
  * Added deposit-sweep proof submission metrics for total, successful, and failed submissions.

* **Bug Fixes**
  * Improved validation and error reporting for malformed data and fee calculations.
  * Refined deposit and redemption logging, error propagation, and metrics recording.

* **Tests**
  * Added deterministic round-trip and fuzz testing for serialization.
  * Added coverage for deposit-sweep metrics and fee-calculation errors.


lrsaturnino:
Updated the PR description's "Scope / follow-ups" section: it previously listed exported/interface signature changes as entirely out of scope for this PR, but `DepositSweepProposal.DepositsKeys` (via the new `DepositKey` type) is itself a minor exported-type change — its element type moved from an anonymous struct to a named one, which breaks source compatibility for any external code that constructs `DepositSweepProposal` directly with the old struct literal. No in-repo consumer is affected and `go build ./...` stays clean, so this doesn't need to block the PR, but the description should reflect it rather than claim no exported changes occurred. Also added a godoc note on `DepositKey` itself flagging this for anyone building against this package.

lrsaturnino:
Two issues from review are fixed and pushed:

**Exported API change not reflected in the PR description** — `DepositSweepProposal.DepositsKeys` moved from an anonymous struct to the named `DepositKey` type, which is a source-compatibility break for any external code building `DepositSweepProposal` values, but the PR description didn't call this out as an accepted exception. Added a godoc note on `DepositKey` in `pkg/tbtc/deposit_sweep.go` documenting the break so downstream consumers aren't surprised on upgrade. (29dc2815b)

**Missing test coverage for the two behavior fixes** — neither the corrected error-wrapping in `GetDepositSweepMaxSize` nor the three new deposit-sweep proof-submission counters had test coverage; the existing fee-estimation table never exercised the `depositsCount == 0` branch that reaches the error-wrapping code, and the counters weren't referenced by any test. Added `LocalChain.SetDepositSweepMaxSizeError` plus a `depositsCount: 0` case to `pkg/tbtcpg/deposit_sweep_fee_test.go` asserting the wrapped error preserves its real cause, and added `TestDepositSweepProofSubmissionCountersRegistered` to `pkg/clientinfo/performance_test.go`, mirroring the existing counter-registration test pattern. (29bbc17e0)

No outstanding issues from this pass.

lrsaturnino (dismissed):
2 findings fixed and pushed; no outstanding issues requiring follow-up.

PR #4188, opened 2026-07-23
refactor(chain): return named parameter structs from tBTC getters

piotr-roslaniec:
## Summary

The on-chain tBTC parameter getters returned wide positional tuples that
callers destructured with long blank-identifier lists. For example:

```go
_, _, _, _, _, _, _, sweepTxMaxTotalFee, _, _, _, err := chain.GetMovingFundsParameters()
```

`GetMovingFundsParameters` returned 11 values plus `err`. Because many of the
values share a type, a positional slip in one of these lists compiles cleanly
and fails silently at runtime.

This replaces the wide tuples with named structs in `pkg/tbtc`, so callers
access fields by name:

```go
movingFundsParameters, err := chain.GetMovingFundsParameters()
... movingFundsParameters.SweepTxMaxTotalFee ...
```

## Changes

- New `pkg/tbtc/parameters.go` defining `MovingFundsParameters`,
  `WalletParameters`, `DepositParameters`, `RedemptionParameters`.
- `GetMovingFundsParameters` (on `tbtc.BridgeChain`, embedded by `tbtcpg.Chain`)
  and the three others (on `tbtcpg.Chain`) now return the corresponding struct.
- All structs live in `pkg/tbtc`, which both the ethereum adapter and `tbtcpg`
  already import, so no new dependency edges are introduced.
- Updated the ethereum adapter implementations, three inline-interface
  declarations in `pkg/tbtc/moving_funds.go`, both local test chains (which get
  simpler — their internal struct aliases are dropped), and all call sites.

Net: the diff removes more than it adds; the change is a simplification.

## Notes for review

- Each of the call-site conversions was mapped position-by-position against the
  original tuple order, and the `tbtcpg` tests discriminate a wrong-but-
  same-typed field pick (e.g. `moved_funds_sweep` sets `TxMaxTotalFee` to 0 but
  `SweepTxMaxTotalFee` to a non-zero value).
- `pkg/tbtc`'s test-only `localChain.GetWalletParameters` is left as the old
  wide-tuple `panic("unsupported")` stub. It satisfies no `tbtc` interface (only
  `tbtcpg.Chain` declares `GetWalletParameters`), so it is dead in the `tbtc`
  package and was left untouched to keep the change surgical.

## Verification

- `go build ./...` and `go vet ./...` clean.
- `pkg/tbtc`, `pkg/tbtcpg`, and `pkg/chain/ethereum` test suites pass.

## Summary by CodeRabbit

* **New Features**
  * Added standardized parameter structures for deposits, redemptions, wallets, and moving funds.
  * Consolidated configuration retrieval into structured results, making fee limits, thresholds, timeouts, and balance settings easier to access.

* **Refactor**
  * Updated related transaction, sweep, and wallet workflows to use the consolidated parameter structures without changing their existing behavior.

### DIFF
--- a/pkg/chain/ethereum/tbtc.go
+++ b/pkg/chain/ethereum/tbtc.go
@@ -1597,25 +1597,18 @@ func (tc *TbtcChain) BuildRedemptionKey(
 	return buildRedemptionKey(walletPublicKeyHash, redeemerOutputScript)
 }
 
-func (tc *TbtcChain) GetDepositParameters() (
-	dustThreshold uint64,
-	treasuryFeeDivisor uint64,
-	txMaxFee uint64,
-	revealAheadPeriod uint32,
-	err error,
-) {
-	parameters, callErr := tc.bridge.DepositParameters()
-	if callErr != nil {
-		err = callErr
-		return
+func (tc *TbtcChain) GetDepositParameters() (tbtc.DepositParameters, error) {
+	parameters, err := tc.bridge.DepositParameters()
+	if err != nil {
+		return tbtc.DepositParameters{}, err
 	}
 
-	dustThreshold = parameters.DepositDustThreshold
-	treasuryFeeDivisor = parameters.DepositTreasuryFeeDivisor
-	txMaxFee = parameters.DepositTxMaxFee
-	revealAheadPeriod = parameters.DepositRevealAheadPeriod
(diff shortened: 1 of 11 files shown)


lrsaturnino (approved):
No blocking issues found.

PR #4190, opened 2026-07-23
feat(maintainer): record SPV proof-submission metrics

piotr-roslaniec:
## What

Wires a `clientinfo` metrics registry + `PerformanceMetrics` into the `maintainer` command and threads a real metrics recorder through `maintainer.Initialize` into the SPV maintainer, so deposit-sweep and redemption proof-submission counters are actually recorded in production.

## Why

The SPV proof submitters accept an optional metrics recorder, but the maintainer boot path never built one, so the recorder was always `nil` and the counters were dead in production. `maintainer` is the only command that runs the SPV maintainer, so these metrics were never observable.

## Changes

- `cmd/maintainer.go`: build the client-info registry + `PerformanceMetrics` (new `initializeMaintainerMetrics` helper), register baseline client-info + Ethereum sources, and pass the recorder to `maintainer.Initialize`. The recorder is derived from the client-info registry: when `clientInfo.port=0`, `clientinfo.Initialize` reports not-configured and the recorder stays a genuine `nil` interface (no typed-nil trap), disabling metrics recording.
- `config/category.go`: add `ClientInfo` to `MaintainerCategories` so the client-info port is parsed for the maintainer command. Because `clientInfo.port` defaults to `9601` (the same shared flag the node `start` command uses), **the maintainer now serves the client-info metrics/diagnostics endpoint on `:9601` by default**, consistent with `start`. Operators who do not want the endpoint can set `clientInfo.port=0`. See the operator impact note below.
- `pkg/maintainer/maintainer.go`: thread `spv.MetricsRecorder` through `Initialize` into `spv.Initialize`.
- `pkg/maintainer/spv`: add an exported `MetricsRecorder` interface; store it on `spvMaintainer`; extend the `transactionProofSubmitter` type; forward the recorder (instead of hardcoded `nil`) at the deposit-sweep and redemption submitters. Moving-funds and moved-funds-sweep submitters accept the recorder to satisfy the type but are not yet instrumented (no counters defined for them).
- `cmd/maintainercli.go`: the two manual one-off proof-submit subcommands pass `nil` (no registry in that path).
- `docs/performance-metrics.adoc`: document the six SPV proof-submission counters that this change makes observable in production.

## Release note / operator impact

The `maintainer` command now exposes the client-info metrics/diagnostics HTTP endpoint on port `9601` (all interfaces) **by default**, matching the behavior of the node `start` command. The endpoint serves system/process metrics, client version, Ethereum connectivity/latest-block info, and the SPV proof-submission counters. It does **not** expose the Ethereum operator key or peer/operator identity. Operators who do not want the maintainer to open this endpoint should set `clientInfo.port=0` in the `ClientInfo` config section (or via `--clientInfo.port=0`).

## Tests

- `TestSubmitDepositSweepProof` and `TestSubmitRedemptionProof` thread a real recorder and assert the attempt/success/failure counters fire on the success path.
- `TestSubmitDepositSweepProofRecordsFailureMetrics` / `TestSubmitRedemptionProofRecordsFailureMetrics` cover the attempt+failure counters on the early zero-confirmations reject path.
- `TestSubmitDepositSweepProofRecordsAssembleFailureMetrics` / `TestSubmitRedemptionProofRecordsAssembleFailureMetrics` cover the failed-counter increment on the SPV-proof-assembly error branch.
- `TestSubmitDepositSweepProofRecordsParseFailureMetrics` / `TestSubmitRedemptionProofRecordsParseFailureMetrics` cover the failed-counter increment on the transaction-input parse-error branch.
- `TestSubmitDepositSweepProofRecordsSubmitFailureMetrics` / `TestSubmitRedemptionProofRecordsSubmitFailureMetrics` cover the failed-counter increment when the on-chain submit itself errors, via an injectable error field on the `localChain` test double.
- `TestInitializeMaintainerMetricsDisabledWhenPortUnset` guards the metrics on/off gate: a `0` client-info port yields a nil recorder.
- `TestInitializeMaintainerMetricsEnabledWhenPortSet` covers the enabled path: a real `*clientinfo.PerformanceMetrics` is built and wired through on a live ephemeral port, guarding against a regression that would silently disable production SPV metrics.
- `TestDepositSweepProofSubmissionCountersRegistered` / `TestSpvProofSkipCountersRegistered` assert the deposit-sweep and SPV proof-skip counters are registered upfront so they appear in `/metrics` before any increment, instead of being silently dropped by the lazy-create-without-register path.

## Base / stacking

Based on `chore/code-quality-followups` (#4187), which targets `dev`. Merge #4187 first and retarget this PR to `dev`, preserving the dependent branches. The recorder seam supplied by #4187 is used here. Follow-ups #4290 and #4291 add health/failure telemetry and alert configuration. Refs #3664.

## Refresh validation

- Rebased onto the current `chore/code-quality-followups` (#4187) tip after that
  branch's own review-fix commit reworked overlapping code in `pkg/clientinfo`,
  `pkg/maintainer/spv`, `pkg/tbtc`, and `pkg/tbtcpg`. Conflicts were resolved
  commit-by-commit and the final tree verified line-for-line against this
  branch's pre-rebase content: only base's own independent changes remain
  (the `estimateCappedFee` relocation to `pkg/tbtcpg/fee.go`, extra test
  coverage in `pkg/beacon` and `pkg/maintainer/spv`), with every constant,
  registration, and metrics-recording call this PR introduces intact.
- `cmd`, `config`, `pkg/clientinfo`, and `pkg/maintainer/...` tests pass after
  integration with the refreshed foundation.
- The newer relay-range/header-limit skip metrics from `dev` are preserved and
  receive the instance recorder.
- Full repo `go build ./...`, `gofmt -l`, and `go test ./...` pass (89
  packages, 1763 tests). `go vet ./...` passes except the pre-existing,
  unrelated protobuf lock-copy warning in `pkg/tecdsa/signing/protocol.go:758`.
- No live deployment or receiver validation was performed.
- CI is green on the rebased head: [Client, Solidity Random Beacon docs, and
  Solidity ECDSA docs all passed](https://github.com/threshold-network/keep-core/actions/runs/34140029518),
  including integration tests (the `electrs-esplora` RPC timeout on the
  previous head has cleared).

### DIFF
--- a/cmd/maintainer.go
+++ b/cmd/maintainer.go
@@ -77,7 +77,7 @@ func maintainers(cmd *cobra.Command, args []string) error {
 		)
 	}
 
-	metricsRecorder := initializeMaintainerMetrics(ctx, blockCounter, btcChain)
+	metricsRecorder := initializeMaintainerMetrics(ctx, blockCounter, tbtcChain, btcChain)
 
 	maintainer.Initialize(
 		ctx,
@@ -99,6 +99,7 @@ func maintainers(cmd *cobra.Command, args []string) error {
 func initializeMaintainerMetrics(
 	ctx context.Context,
 	blockCounter chain.BlockCounter,
+	ethRPC clientinfo.EthereumRPC,
 	btcChain bitcoin.Chain,
 ) spv.MetricsRecorder {
 	registry, isConfigured := clientinfo.Initialize(
@@ -121,7 +122,7 @@ func initializeMaintainerMetrics(
 	registry.ObserveBtcConnectivity(btcChain, clientConfig.ClientInfo.BitcoinMetricsTick)
 	registry.RegisterBtcChainInfoSource(btcChain)
 	healthChecker := clientinfo.NewRPCHealthChecker(
-		registry, blockCounter, btcChain, clientConfig.ClientInfo.RPCHealthCheckInterval,
+		registry, ethRPC, btcChain, clientConfig.ClientInfo.RPCHealthCheckInterval,
 	)
 	// An unavailable RPC must not delay starting the maintainer's control loop.
 	go healthChecker.Start(ctx)
--- a/cmd/maintainer.go
+++ b/cmd/maintainer.go
@@ -101,7 +101,7 @@ func initializeMaintainerMetrics(
 ) spv.MetricsRecorder {
 	registry, isConfigured := clientinfo.Initialize(
 		ctx,
-		clientConfig.ClientInfo.Port,
+		clientConfig.ClientInfo,
 	)
 	if !isConfigured {
 		logger.Infof("client info endpoint not configured")
--- a/cmd/maintainer.go
+++ b/cmd/maintainer.go
@@ -6,10 +6,14 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/keep-network/keep-core/build"
 	"github.com/keep-network/keep-core/config"
 	"github.com/keep-network/keep-core/pkg/bitcoin/electrum"
+	"github.com/keep-network/keep-core/pkg/chain"
 	"github.com/keep-network/keep-core/pkg/chain/ethereum"
+	"github.com/keep-network/keep-core/pkg/clientinfo"
 	"github.com/keep-network/keep-core/pkg/maintainer"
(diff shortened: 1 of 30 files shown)


lionakhnazarov (dismissed):
lgtm

PR #4192, opened 2026-07-23
Apply the safe minimum fee floor to all wallet transactions

piotr-roslaniec:
## What

Applies the safe minimum fee-rate floor (5 sat/vByte) plus a 25% buffer to
**all** wallet Bitcoin transactions - deposit sweeps, redemptions, moving funds,
and moved-funds sweeps - via the shared `applyWalletTxFeeFloor` helper in
`pkg/tbtcpg/fee.go`. A fee oracle can return an unusably low estimate in an
uncongested mempool; because these transactions are not RBF-enabled, an
underpriced one can get stuck and jam the wallet. The floor keeps the fee safely
above the relay floor while staying far below the Bridge maximum.

This carries forward the work from #4179 (opened from a fork that could not be
iterated on) rebased cleanly onto current `main`, and folds in the fixes from a
multi-agent review of that change.

## Review fixes included

- **Restore the truncation invariant** dropped when the deposit-sweep fee logic
  was extracted into the shared helper: the 25% buffer is applied to the
  truncated per-vByte rate, which is lossless only because `EstimateFee` returns
  an exact multiple of the vsize. Documented so a future change to that contract
  is caught rather than silently under-pricing.
- **Correct the `estimatedFee`/`txVsize` unit description** (satoshis and
  vBytes, not "sat/vByte").
- **Redemption diagnostic:** emit a distinct WARN in `ProposeRedemption` when the
  floored per-request fee share would exceed the per-request maximum fee, so
  operators can distinguish this from a generic on-chain validation failure. Fee
  behavior is unchanged; the per-request cap remains enforced by on-chain
  validation.
- **Test coverage** for previously untested branches: the redemption
  raw-estimate-above-cap error, and the minimum-floor clamp on the moving-funds
  and moved-funds-sweep paths.

## Notes

- `minWalletTxSatPerVByteFee` and the buffer are a stopgap for the current
  non-RBF, fire-and-forget wallet-transaction path; once fee-bumping lands
  (#4171) this policy should be revisited.
- Follower-side enforcement of the minimum fee is tracked separately (#4180).

## Testing

`gofmt`, `go build`, `go vet`, `staticcheck`, and `go test ./pkg/tbtcpg/...`
(91 tests) all pass on current `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Applied consistent wallet safe-minimum fee buffering across moved-funds, moving-funds, deposit sweeps, and redemptions.
  * Stopped fee estimation/proposal when the safe minimum would exceed the configured maximum total fee.
  * Improved redemption handling by enforcing a maximum total fee bound early and emitting clearer warnings when per-request fees are likely to be rejected.
* **Tests**
  * Updated and expanded fee-floor, buffering, cap, and invalid-input coverage across moving-funds, moved-funds, deposit sweep, and redemption scenarios.

### DIFF
--- a/pkg/tbtcpg/deposit_sweep.go
+++ b/pkg/tbtcpg/deposit_sweep.go
@@ -21,26 +21,6 @@ import (
 // This will ensure that deposit sweep transaction fees are not underestimated.
 const depositScriptByteSize = 126
 
-// minSweepTxSatPerVByteFee is the minimum fee rate, in sat/vByte, applied to
-// deposit sweep transactions. A fee oracle can return an unusably low estimate
-// (down to the 1 sat/vByte relay floor enforced by the Electrum client) in an
-// uncongested mempool. Because a sweep consolidates significant wallet value
-// and is not RBF-enabled, it cannot be replaced once broadcast, so a floor-rate
-// sweep can get stuck in the mempool and jam the wallet: no new sweep can be
-// built while the previous one is unconfirmed. This minimum keeps the sweep fee
-// safely above the relay floor while remaining far below the Bridge's
-// per-deposit maximum fee. The value is intentionally conservative and could be
-// made configurable; see threshold-network/keep-core#4171.
-//
-// NOTE: this static floor and the 25% buffer applied below are a stopgap for
-// the current fire-and-forget, non-RBF sweep path: because a stuck sweep cannot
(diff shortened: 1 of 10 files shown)


lrsaturnino (approved):
No blocking issues found.

PR #4194, opened 2026-07-23
Add follower-side soft check for below-floor sweep fees

piotr-roslaniec:
## Summary

Adds a follower-side soft (log-only) check that the leader's proposed deposit
sweep fee is not below the safe minimum. The on-chain `WalletProposalValidator`
only bounds the sweep fee from above, so a misbehaving or unpatched leader can
propose a fee at the ~1 sat/vByte relay floor that would otherwise be signed -
the same underpricing that jams the wallet (threshold-network/keep-core#4171).
Each node recomputes the safe minimum and logs a warning when the proposal is
below it.

The check is intentionally **log-only, not a rejection**: rejecting a
below-floor proposal would, during a mixed-version rollout, split signers
(patched nodes reject, unpatched nodes sign) and could stall signing. Hard
enforcement belongs on-chain in the `WalletProposalValidator` or behind a
coordinated all-nodes upgrade.

## Contents

This branch is rebased on current `main` and, in addition to the soft check,
carries the safe-minimum sweep-fee floor applied to all wallet transactions
(the `feat/floor-all-wallet-tx-fees` work, also open as #4179 / #4192). The
floor commit is bundled here so the branch builds and passes CI against `main`;
if the floor lands separately first, this branch should be rebased to drop it.

## Review follow-ups addressed

Incorporates the multi-agent review of the soft check:

- Warn when `SweepTxFee` is `nil` instead of silently skipping the check; a
  missing fee gets its own distinct log line.
- Compare the proposed fee with `big.Int.Cmp` instead of `Int64()`, which is
  undefined above `MaxInt64`.
- Replace the literal-pinned constant drift guard with a direct comparison of
  the (now exported) `pkg/tbtc` mirrors against the canonical `pkg/tbtcpg`
  constants, so drift is caught regardless of which side changes. A
  literal-based guard could be defeated by updating `tbtcpg` and the literal
  together while forgetting the mirror; comparing the live values closes that
  gap.

## Supersedes

Replaces #4180, whose head lived on a fork branch that had drifted onto an old
`main` and become `CONFLICTING`. This PR is the same work rebased clean onto
current `main`.

## Summary by CodeRabbit

- **Improvements**
  - Enhanced minimum-fee enforcement across wallet, moving-funds, and redemption transactions, including a safety buffer and caps for maximum fees.
  - Deposit sweep proposals now emit warnings when the provided sweep fee is missing or below the recommended minimum.
  - Fee sizing now uses a consistent worst-case estimate to keep fee calculations aligned.
- **Bug Fixes**
  - Updated fee estimation to better prevent proposals that could otherwise be constructed with unsafe or capped-out fee values.
  - Improved validation behavior for edge cases such as invalid fee sizes or nil fee inputs.

### DIFF
--- a/pkg/tbtc/deposit_sweep.go
+++ b/pkg/tbtc/deposit_sweep.go
@@ -50,14 +50,15 @@ const (
 	// the transaction is known on the Bitcoin chain. This delay is needed
 	// as spreading the transaction over the Bitcoin network takes time.
 	depositSweepBroadcastCheckDelay = 1 * time.Minute
-	// minSweepTxSatPerVByteFee mirrors tbtcpg.minWalletTxSatPerVByteFee, the safe
+	// minSweepTxSatPerVByteFee mirrors tbtcpg.MinWalletTxSatPerVByteFee, the safe
 	// minimum sweep fee rate. It is duplicated here because pkg/tbtcpg imports
 	// pkg/tbtc, so this package cannot import the canonical constant without a
-	// dependency cycle; keep the two in sync. It backs a follower-side soft
-	// (log-only) check that the leader's proposed sweep fee is not below the
-	// floor (see threshold-network/keep-core#4171).
+	// dependency cycle; keep the two in sync (guarded by TestSweepFeeConstants
+	// MirrorTbtcpg). It backs a follower-side soft (log-only) check that the
+	// leader's proposed sweep fee is not below the floor (see
+	// threshold-network/keep-core#4171).
 	minSweepTxSatPerVByteFee = 5
-	// depositScriptByteSize mirrors tbtcpg.depositScriptByteSize, the worst-case
+	// depositScriptByteSize mirrors tbtcpg.DepositScriptByteSize, the worst-case
 	// deposit script size used to estimate the sweep transaction virtual size.
 	depositScriptByteSize = 126
 )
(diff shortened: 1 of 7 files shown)


lrsaturnino:
2 issues found, both fixed and pushed.

**1. Zero test coverage for the follower-side soft-check (Major)** — the follower-side sweep-fee soft check in `ValidateDepositSweepProposal` (`pkg/tbtc/deposit_sweep.go`) had no test exercising it. Added `TestValidateDepositSweepProposal_SweepFeeSoftCheck` in `pkg/tbtc/deposit_sweep_test.go`, covering: fee below the floor (warns, no error), fee at/above the floor (no warning), and a nil fee (warns; only reachable via a test/mock caller since the real on-chain validation path already rejects nil fees earlier). Fixed in 9404365.

**2. Fee-floor/buffer logic ignored the on-chain per-request redemption fee-share cap (Major)** — `ProposeRedemption` (`pkg/tbtcpg/redemptions.go`) fetched `GetRedemptionParameters()` but discarded the per-request cap (`txMaxFee`), bounding the buffered fee estimate only by the flat total cap (`txMaxTotalFee`). The Bridge/WalletProposalValidator enforces both caps independently, so a tighter per-request cap could still cause a rejection. Fixed by bounding the estimate's ceiling by `min(txMaxTotalFee, txMaxFee * requestsCount)` and corrected the docstrings accordingly. Added a test case reproducing the tighter-per-request-cap scenario. Fixed in cca60b9.

No unresolved issues. `go build ./...`, `go vet`, `gofmt`, `go test ./pkg/tbtc/... ./pkg/tbtcpg/...`, staticcheck, and gosec all pass locally on both commits.

lrsaturnino (dismissed):
2 findings fixed and pushed; none require follow-up.

PR #4195, opened 2026-07-23
test(storage): cover disk storage initialization

piotr-roslaniec:
## What

Adds a test suite for `pkg/storage`, which previously had **0% coverage**. Raises package coverage to **~88%**.

The tests focus on the behavior that gives the package its reason to exist: separating **sensitive keystore data** (whose loss is "a serious protocol violation" per the code) from **restart-recoverable work data** into two sibling directories, and rooting each persistence handle under the correct one.

Covered:
- `Initialize` creates the `keystore` and `work` directory layout under the configured root.
- Restart idempotency (`Initialize` re-run against existing directories does not error).
- Error path when the storage root's parent is missing (`EnsureDirectoryExists` uses single-level `os.Mkdir`).
- Key store vs. work persistence each land under the correct root and **never leak into the other's directory** (this is the assertion that would catch a broken separation).
- Nested-target error path (missing intermediate parent).

## Verification

`go test ./pkg/storage/` passes (6 tests); `gofmt -l`, `go vet`, and `staticcheck -SA1019` (the `client-lint` CI gate) all clean.

## Note for reviewers: the test file is force-added

`.gitignore` has an unanchored `storage/` rule (added in 2026, intended for the client's **runtime** storage directory) that also matches the `pkg/storage/` **source** package. `storage.go` predates that rule (added 2022) and is grandfathered in as tracked, so it wasn't caught; a brand-new file here is. The file was added with `git add -f` and is now tracked normally (CI checks out tracked files regardless of `.gitignore`).

I kept this PR surgical and did **not** change the shared ignore rule. If you'd prefer a permanent fix, a one-line follow-up would anchor the rule to the runtime dir (`storage/` -> `/storage/`) or add an exception (`!/pkg/storage/`). Happy to open that separately.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for storage directory initialization and persistence setup.
  * Verified creation of keystore and work directories, repeated initialization, and correct directory placement.
  * Added validation for expected errors when required parent or intermediate directories are missing.

### DIFF
--- a/pkg/storage/storage_test.go
+++ b/pkg/storage/storage_test.go
@@ -1,0 +1,134 @@
+package storage
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+const testPassword = "grzegorz-brzeczyszczykiewicz"
+
+// assertIsDir fails the test unless path exists and is a directory.
+func assertIsDir(t *testing.T, path string) {
+	t.Helper()
+	info, err := os.Stat(path)
+	if err != nil {
+		t.Fatalf("expected directory [%s] to exist: [%v]", path, err)
+	}
+	if !info.IsDir() {
+		t.Fatalf("expected [%s] to be a directory", path)
+	}
+}
+
+// assertDoesNotExist fails the test unless path is absent.
+func assertDoesNotExist(t *testing.T, path string) {
+	t.Helper()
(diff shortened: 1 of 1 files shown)


lrsaturnino (approved):
No blocking issues found.

PR #4196, opened 2026-07-23
test(entry): cover signature-share validation and message type

piotr-roslaniec:
## What

`pkg/beacon/entry` previously had only marshaling round-trip tests (~14% coverage). This adds unit tests for the logic that can be exercised without a live broadcast channel or relay chain, raising coverage to **~27%**.

Covered:
- **`extractAndValidateShare`** — the gatekeeper that accepts or rejects a signature share received from another group member. A member that accepted a bad share would contribute to a corrupt relay entry, so all three rejection paths (unparseable share bytes, unknown sender, share that does not verify) are tested alongside the accept path, built on real `bls` sign/verify.
- **`SignatureShareMessage`** value object (`NewSignatureShareMessage`, `SenderID`) and its wire **`Type`** identifier, which must stay stable for cross-client network compatibility.
- `Unmarshal` rejecting a member index that overflows the uint8 range (the documented uint8/uint32 wire invariant).

## Not covered (by design)

`SignAndSubmit` (the signing orchestration loop), `broadcastShare`, `completeSignature`, and relay entry submission drive a broadcast channel and the beacon chain. Those are exercised by the beacon integration tests, not unit tests; forcing unit-level mocks there would be brittle and low-value.

## Verification

`go test ./pkg/beacon/entry/` passes; `gofmt -l`, `go vet`, and `staticcheck -SA1019` clean.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for validating signature shares, including malformed, missing, and invalid signatures.
  * Added coverage for signature share message creation, type identification, and rejection of overflowing sender identifiers.

### DIFF
--- a/pkg/beacon/entry/entry_test.go
+++ b/pkg/beacon/entry/entry_test.go
@@ -1,0 +1,74 @@
+package entry
+
+import (
+	"math/big"
+	"testing"
+
+	bn256 "github.com/ethereum/go-ethereum/crypto/bn256/cloudflare"
+
+	"github.com/keep-network/keep-core/pkg/bls"
+	"github.com/keep-network/keep-core/pkg/protocol/group"
+)
+
+// extractAndValidateShare is the gatekeeper that decides whether a signature
+// share received from another group member is trustworthy. A member that
+// accepted a bad share would contribute to a corrupt relay entry, so each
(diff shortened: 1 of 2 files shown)


lrsaturnino (approved):
No blocking issues found.

PR #4197, opened 2026-07-24
test: cover sweep-fee floor boundaries and add below-floor metric

piotr-roslaniec:
## Summary

Follow-up to #4194, addressing the confirmed, actionable findings from a multi-agent review of that PR. Stacked on `feat/follower-sweep-fee-soft-check`; rebase onto `main` once #4194 merges.

Scope is deliberately narrow: test coverage for the untested soft check and fee-floor boundaries, plus one small observability add. No behavior of the leader-side floor changes.

### Follower soft check: tests + metric (was untested — review's lead item)
- Extract the below-floor decision from `ValidateDepositSweepProposal` into a pure `checkSweepFeeFloor` helper, and unit-test it directly (just below / exactly at / above the safe minimum, plus missing and zero fee) instead of scraping logger output. The check previously had **zero behavioral tests** — only a constant-drift guard — so a flipped `Cmp` or a dropped branch would pass CI silently.
- Emit a `deposit_sweep_fee_below_floor_total` counter on the follower path when a proposal is below the floor. The check was log-only; a metric lets operators alert on underpriced proposals during a mixed-version rollout instead of grepping node logs. Reuses the existing follower `metricsRecorder`; `ValidateDepositSweepProposal`'s signature is unchanged and the node still signs the proposal.

### Redemption fee: cap boundary tests
- `EstimateRedemptionFee` gains coverage for two boundary outcomes of the safe-minimum floor:
  - buffered fee above the Bridge max is bounded down to the cap (not an error);
  - a raw estimate already above the cap errors before the floor is applied.
- The second is intended behavior (documented in `redemptions.go`), but it trades liveness (redemption not attempted this round) for not broadcasting a fee the Bridge would reject. The test pins that tradeoff so it is a visible, regression-safe decision.

## Reviewed but intentionally deferred (not fixed here)

- **Follower compares against the static `5*vsize` floor, not the buffered estimate.** Catching "above floor but underpaid" proposals would require a live fee-oracle query inside proposal validation — the wrong layer, and it introduces non-determinism into validation. Hard enforcement belongs on-chain in the `WalletProposalValidator`.
- **Soft check covers only 1 of the 4 floored flows** (deposit sweeps, not redemptions / moving funds / moved-funds sweeps). Kept as-is per the decision to keep-and-improve the deposit-sweep check; flagged here so the asymmetry is a visible decision rather than an oversight.
- **P2SH deposit vsize underestimation** (both leader and follower use `isWitness=true`) — already documented as a code caveat in `fee.go`.
- **Integer-truncation before the 25% buffer** in `applyWalletTxFeeFloor` — a no-op for current callers (`EstimateFee` returns exact multiples); safe to defer.
- **Nil-`SweepTxFee` branch** — defensive; a nil fee fails at ABI packing before the soft check runs, so it is not reachable in production.
- **Mirror constants** (`tbtc` <-> `tbtcpg`) — already guarded by the drift test added in #4194.

An overflow concern raised during review was **rejected** as logically unreachable: the `int64(maxTotalFee)` cast runs only inside `uint64(totalFee) > maxTotalFee`, which for an `int64 totalFee` implies `maxTotalFee < MaxInt64`, so the cast cannot wrap.

## Testing

- `go test ./pkg/tbtcpg/` — pass
- `go test ./pkg/tbtc/ -run 'TestDepositSweepAction|TestCheckSweepFeeFloor|TestSweepFeeConstantsMirrorTbtcpg'` — pass
- `go vet` / `gofmt` clean on both packages

### DIFF
--- a/pkg/clientinfo/performance.go
+++ b/pkg/clientinfo/performance.go
@@ -137,6 +137,10 @@ func (pm *PerformanceMetrics) registerCounterMetrics() {
 		MetricSigningSuccessTotal,
 		MetricSigningFailedTotal,
 		MetricSigningTimeoutsTotal,
+		MetricDepositSweepExecutionsTotal,
+		MetricDepositSweepExecutionsSuccessTotal,
+		MetricDepositSweepExecutionsFailedTotal,
+		MetricDepositSweepFeeBelowFloorTotal,
 		MetricRedemptionExecutionsTotal,
 		MetricRedemptionExecutionsSuccessTotal,
 		MetricRedemptionExecutionsFailedTotal,
@@ -276,6 +280,8 @@ func (pm *PerformanceMetrics) registerHistogramMetrics() {
 	durationMetrics := []string{
 		MetricDKGDurationSeconds,
 		MetricSigningDurationSeconds,
+		MetricDepositSweepExecutionDurationSeconds,
+		MetricDepositSweepTxSigningDurationSeconds,
 		MetricRedemptionActionDurationSeconds,
 		MetricWalletActionDurationSeconds,
 		MetricCoordinationDurationSeconds,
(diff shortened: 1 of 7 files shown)


lrsaturnino:
1 issue found and fixed.

**Fixed**

- **`deposit_sweep_fee_below_floor_total` counter never exposed on `/metrics`** (Major) — `IncrementCounter`/`RecordDuration` only ever wrote into the internal metrics map; a name only reaches the `/metrics` endpoint if it was also passed to `registerAllMetrics()` at startup. This counter (and 5 pre-existing sibling metrics in the same function: `deposit_sweep_executions_total`, `deposit_sweep_executions_success_total`, `deposit_sweep_executions_failed_total`, `deposit_sweep_execution_duration_seconds`, `deposit_sweep_tx_signing_duration_seconds`) used raw string literals and was never registered, so none of the six ever appeared in a scrape. Added named `clientinfo.Metric*` constants and registered all six in `registerAllMetrics()`, matching the existing pattern in `pkg/tbtc/redemption.go` and `pkg/maintainer/spv/redemptions.go`, and added a registration test mirroring the existing join-failure-counters test. Commit f5a284f7c.

No unresolved findings.

lrsaturnino (dismissed):
1 finding fixed and pushed; none require follow-up.

PR #4201, opened 2026-07-26
chore(ci): move the solidity toolchain off end-of-life Node to 22 LTS

mswilkison:
**Nine files, runtime pins only.** No dependency changes.

This previously carried all of #4109's commits and showed a 257-file diff — enough
that CodeRabbit declined to review it, and enough that merging it would have landed
the security release as a side effect. It has been rebased onto `main` and now
contains a single commit of its own.

Rebasing rather than retargeting at #4109 was the right move on two counts: that
branch lives on a fork (`lionakhnazarov/keep-core`), so GitHub cannot base a PR on
it; and there is no dependency to preserve — #4201's nine files are disjoint from
all 249 of #4109's. Merging in either order is conflict-free, verified by
simulating both.

> **The branch name is wrong.** `chore/ci-node-24-lts` says 24; this goes to
> **22**. Renaming would break the PR, so it stands — the title and the diff are
> the accurate ones, and the reason is below.

## The problem

All five workflows that build and test the contracts pin **Node 18.15.0**, which
reached end of life **2025-04-30** and receives no further security patches:

`contracts-ecdsa`, `contracts-random-beacon`, `npm-ecdsa`, `npm-random-beacon`,
`reusable-solidity-docs`

`engines.node` in both solidity manifests and both contract Dockerfiles say the
same.

## Why 22 and not 24

On support alone 24 is the better target — current LTS, EOL 2028-04-30, against
22's 2027-04-30. **This repo cannot reach it yet.**

Both solidity packages pin `hardhat 2.10.0`, which resolves `undici 5.19.1`.
Hardhat 2.x on the undici 5 line cannot download a compiler on Node 24: npm's
undici and Node 24's bundled undici share a global-dispatcher symbol, so
Hardhat's solc download reaches Node's internal dispatcher, which rejects
`maxRedirections`. Measured on this branch, `solidity/random-beacon`, cold
compiler cache:

```
Node v24.11.1   hardhat compile --force
                Error HH502: Couldn't download compiler versions list.
```

**A local run does not catch this.** HH502 is on the download path, so once a
compiler is cached the run succeeds. CI runners always start cold, which is
exactly where it fires. An earlier revision of this description reported a clean
Node 24 run and concluded 24 was safe; that run had a warm cache.

The fix is `hardhat >= 2.26`, and that is blocked on something larger:

- `@defi-wonderland/smock` (`^2.0.7`, resolved 2.0.7 in both packages) breaks at
  hardhat **2.20.0** — `Sandbox.create` — and is archived upstream, so no
  forward version is coming
- Node 24 needs hardhat **>= 2.26**

Disjoint ranges, so **smock has to go before Node 24 is reachable**. That is the
same chain tbtc-v2 worked through in threshold-network/tbtc-v2#1062 → #1063 →
#1064 — a contract-backed mock replacing smock, then hardhat 2.29 and Node 24
bumped together in one commit — and it is why that repo is on 24 and this one is
not. It has already paid for it.

So this is the interim step: off an end-of-life runtime now, onto the highest
version the current dependencies actually support. 24 follows the smock removal.

**Worth a separate issue:** `@defi-wonderland/smock` is pinned as `^2.0.7`. The
lockfile holds 2.0.7 today, but the range floats to 2.4.1 on any regeneration —
silently, into an archived package. Same trap tbtc-v2 hit.

## Verified

No dependency changes; runtime pins only.

`solidity/random-beacon`, full suite on this branch, via the package's own
`test` script (`hardhat typechain && hardhat check-accounts-count &&
USE_EXTERNAL_DEPLOY=true TEST_USE_STUBS_BEACON=true hardhat test`):

```
Node v18.20.8   934 passing, 6 pending, 0 failing
Node v22.23.1   934 passing, 6 pending, 0 failing
```

Identical. Node 24 is deliberately not quoted as passing — see above.

An earlier revision of this description quoted `168 passing, 12 failing` with
the 12 being `No deployment found for: T`. That does not reproduce: run through
the package's own script, which sets `USE_EXTERNAL_DEPLOY=true`, the external
deployment resolves and the suite is clean on both runtimes.

## `infrastructure/` deliberately untouched

It contains the worst offender in the repo (a **`node:11`** init container, EOL
2019, carrying 7 open Dependabot alerts), but **nothing there is built by CI**:

- `client.yml` lists `infrastructure/**` under `paths-ignore`
- the only build contexts across all workflows are `.`, `./solidity/ecdsa`,
  `./solidity/random-beacon`
- the init container image is pulled pre-built from
  `gcr.io/keep-dev-fe24/initcontainer-provision-keep-client-ethereum`, and is
  referenced only by `keep-dev` and `keep-test` — `keep-prd` has no keep-client
  manifests at all
- its `package-lock.json` has not changed since 2023-04-17

So those alerts are against a build this repo never performs. That wants a
decision about whether the image is still deployed — delete it, or move its build
into CI so the alerts mean something — rather than a version bump to a file
nothing builds. Same for `docker/ethereum/geth-node` (geth **v1.9.6**, 2019) and
`docker/ethereum/dashboard-node` (untagged `ubuntu`), both untouched since 2019
and referenced by nothing.

## Summary by CodeRabbit

* **Chores**
  * Updated contract development, testing, documentation, deployment, and publishing workflows to use Node.js 22.
  * Updated container environments to Node.js 22.
  * Raised the minimum required Node.js version to 22 for the Solidity ECDSA and random beacon packages.

### DIFF
--- a/.github/workflows/contracts-ecdsa.yml
+++ b/.github/workflows/contracts-ecdsa.yml
@@ -55,7 +55,7 @@ jobs:
           # Using fixed version, because 18.16 was sometimes causing issues with
           # artifacts generation during `hardhat compile` - see
           # https://github.com/NomicFoundation/hardhat/issues/3877
-          node-version: "18.15.0"
+          node-version: "22.23.1"
 
       - uses: ./.github/actions/install-yarn-deps
         with:
@@ -81,7 +81,7 @@ jobs:
           # Using fixed version, because 18.16 was sometimes causing issues with
           # artifacts generation during `hardhat compile` - see
           # https://github.com/NomicFoundation/hardhat/issues/3877
-          node-version: "18.15.0"
+          node-version: "22.23.1"
 
       - uses: actions/setup-python@v4
         with:
@@ -123,7 +123,7 @@ jobs:
           # Using fixed version, because 18.16 was sometimes causing issues with
           # artifacts generation during `hardhat compile` - see
           # https://github.com/NomicFoundation/hardhat/issues/3877
-          node-version: "18.15.0"
+          node-version: "22.23.1"
 
       - uses: ./.github/actions/install-yarn-deps
         with:
@@ -150,7 +150,7 @@ jobs:
           # Using fixed version, because 18.16 was sometimes causing issues with
           # artifacts generation during `hardhat compile` - see
           # https://github.com/NomicFoundation/hardhat/issues/3877
-          node-version: "18.15.0"
+          node-version: "22.23.1"
 
       - uses: ./.github/actions/install-yarn-deps
         with:
@@ -183,7 +183,7 @@ jobs:
           # Using fixed version, because 18.16 was sometimes causing issues with
           # artifacts generation during `hardhat compile` - see
           # https://github.com/NomicFoundation/hardhat/issues/3877
-          node-version: "18.15.0"
+          node-version: "22.23.1"
           registry-url: "https://registry.npmjs.org"
 
       - uses: ./.github/actions/install-yarn-deps
@@ -283,7 +283,7 @@ jobs:
           # Using fixed version, because 18.16 was sometimes causing issues with
           # artifacts generation during `hardhat compile` - see
           # https://github.com/NomicFoundation/hardhat/issues/3877
-          node-version: "18.15.0"
+          node-version: "22.23.1"
           registry-url: "https://registry.npmjs.org"
 
       - uses: ./.github/actions/install-yarn-deps
--- a/.github/workflows/contracts-random-beacon.yml
+++ b/.github/workflows/contracts-random-beacon.yml
@@ -55,7 +55,7 @@ jobs:
           # Using fixed version, because 18.16 was sometimes causing issues with
           # artifacts generation during `hardhat compile` - see
           # https://github.com/NomicFoundation/hardhat/issues/3877
-          node-version: "18.15.0"
+          node-version: "22.23.1"
 
       - uses: ./.github/actions/install-yarn-deps
         with:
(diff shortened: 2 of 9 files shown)


mswilkison:
Thanks — taking one of the two.

**Pinned both Dockerfiles to `node:22.23.1-alpine`.** Fair catch, and squarely in
scope: this PR's whole point is pinning a runtime, and leaving the images on a
floating `22-alpine` while CI pins `22.23.1` reproduces the drift it is meant to
remove. Pushed.

**Not taking the non-root change in this PR.** The finding is legitimate, but it
is a hardening change rather than a runtime pin, and as described it would break
the build:

- `corepack enable && corepack prepare yarn@4.8.1 --activate` writes under
  `/usr/local/lib/node_modules`, so a `USER` set before it fails
- set after it, `WORK_DIR` and everything `COPY . ./` put there is still
  root-owned, and `npx hardhat` writes `artifacts/`, `cache/` and `typechain/`
  into it — so it needs a `chown -R` too

That is a real change needing a real image build to validate, and it would put a
security-relevant edit inside a PR whose reviewability was the reason it was just
rebased from 257 files down to 9. Worth its own PR, where it can be tested.

Context for whoever picks it up: these two images are ephemeral build tooling
whose entrypoint is `npx hardhat`, not a network-facing service, so the exposure
is narrower than the label suggests. The repo has sharper container problems
already catalogued — a `node:11` init container from 2019 with open Dependabot
alerts, and `docker/ethereum/geth-node` on geth v1.9.6 — none of which CI builds.
Both are described at the bottom of this PR's description.

piotr-roslaniec (approved):
Self-approve as repo admin to advance the smock/hardhat/Node-24 chain. The 9-file diff is runtime pins only (Node 18 -> 22.23.1) and is consistent with the author's description. CI is green. Note: #4205 supersedes this PR per the author's own body in #4205; the conflict between 4201 and 4205 (same 9 files, 22.23.1 vs 24.11.1) will be resolved in the next decision.

PR #4202, opened 2026-07-27
build(solidity): raise hardhat to 2.19.5 and pin smock, in both packages

mswilkison:
**Dependencies only. No test file is touched.** First step in removing
`@defi-wonderland/smock`, which upstream has archived with this notice:

> ⚠️ **DEPRECATED – DO NOT USE**
>
> This repository is no longer maintained and is **deprecated**.
>
> It may contain **outdated, insecure, or vulnerable code** and should **not**
> be used in production or as a dependency in any project.
>
> The repository is retained solely for historical reference. No support,
> updates, or security patches will be provided.

npm carries the matching deprecation, the repo is archived, and the last push
was 2025-05-20. That is the reason to remove it. The toolchain ceiling below is
not the motivation — it is what makes removal unavoidable rather than
schedulable, since no smock version reaches the hardhat that Node 24 needs.

Independent of #4201 — different regions of the same `package.json` files, and
the two merge cleanly in either order (verified).

## Three coupled bumps, in both packages

```
hardhat                 ^2.10.0  ->  2.19.5
@defi-wonderland/smock   ^2.0.7  ->  2.3.4
hardhat-gas-reporter     ^1.0.8  ->  ^2.3.0
```

None of them works alone, which is why they land together.

**hardhat first.** The contract-backed mock that will replace smock configures
itself with transactions, and a transaction mines a block. Keeping that from
moving the chain clock needs `allowBlocksWithSameTimestamp` — and that option
**does not exist in 2.10.0**. Checked against the published tarballs: absent in
2.10.0, present by 2.14.0. 2.19.5 is the last version smock still works on, so
it is the one version with both properties.

**smock second, because 2.0.7 does not survive that bump.** On hardhat 2.19.5 it
dies before any test runs:

```
TypeError: vm.on is not a function
  at new ObservableVM (@defi-wonderland/smock/src/observable-vm.ts:19:35)
  at Sandbox.create   (@defi-wonderland/smock/src/sandbox.ts:82:12)
```

2.3.4 is the version that works there, and it is **pinned exactly rather than
left as a range**. `^2.0.7` resolves to **2.4.1** today, so the lockfile was one
regeneration away from silently jumping four minor versions into an archived
package.

2.4.1 is deliberately not the target. It advertises `hardhat >=2.21.0` support
and still fails — differently — on a real plugin stack:

```
TypeError: provider.init is not a function
  at getHardhatBaseProvider (@defi-wonderland/smock/src/utils/hardhat.ts:32:18)
  at Sandbox.create         (@defi-wonderland/smock/src/sandbox.ts:117:55)
```

That is what makes the smock removal unavoidable rather than optional: there is
no smock version that reaches hardhat 2.26+, and hardhat 2.26+ is what Node 24
requires.

**hardhat-gas-reporter third.** 1.x bundles `eth-gas-reporter`, which the mocha
inside hardhat 2.19.5 refuses to load on Node 22 —
`ERR_MOCHA_INVALID_REPORTER`, raised before a single test executes. Only
`random-beacon` hits it, because only `random-beacon` enables the reporter, but
both packages move together to keep them aligned. This matters now rather than
later because #4201 puts CI on Node 22.

## Verified

Whole suite, each package, on both runtimes:

| | before (hh 2.10, smock 2.0.7) | after |
| --- | --- | --- |
| `random-beacon` | 926 passing / 6 pending / 0 failing | **926 / 6 / 0** |
| `ecdsa` | 650 passing / 44 pending / 0 failing | **650 / 44 / 0** |

Identical on Node 18.20.8 and Node 22.23.1.

## Where this is going

1. this PR — dependencies only
2. replace smock with a contract-backed mock (`MockContract.sol` +
   `test/helpers/mock.ts`), portable from threshold-network/tbtc-v2#1062
3. hardhat 2.19.5 -> 2.29 and Node 22 -> 24, which only becomes possible once
   smock is gone

Roughly 26 `smock.fake` sites across 20 files, most of them in `ecdsa`. The one
`smock.mock` call, at `WalletRegistry.RandomBeacon.test.ts:236`, only reads
`.address` and never touches `setVariable`, so it is a plain factory deploy and
needs no mock at all.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive programmable contract-mocking coverage for return values, reverts, call recording, argument matching, resets, static calls, and transaction behavior.
  * Enabled previously skipped slashing failure scenarios and expanded event assertions.
  * Improved timestamp-sensitive test reliability and gas measurement consistency.
  * Standardized mock behavior and assertions across ECDSA and Random Beacon test suites.

* **Chores**
  * Added support for Yarn local-state files to be excluded from version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,6 @@ venv/
 target/
 dist/
 .DS_Store
+
+# Yarn 4 local state (not for zero-installs)
+.yarn/install-state.gz
--- a/solidity/ecdsa/contracts/test/IMockTarget.sol
+++ b/solidity/ecdsa/contracts/test/IMockTarget.sol
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+pragma solidity ^0.8.0;
+
+/// @notice Test-only interface exercising the shapes `MockContract` has to
+///         answer: a view function reached by STATICCALL, a state-changing
+///         function whose calls are asserted on, a function returning nothing,
+///         a struct return and a multi-value return.
+interface IMockTarget {
+    struct Info {
+        address owner;
+        uint64 createdAt;
+        bool active;
+    }
+
+    function doThing(address who, uint256 amount) external returns (bool);
+
+    function noReturn(uint256 value) external;
+
+    function readValue(uint256 key) external view returns (uint256);
+
+    function readInfo(uint256 key) external view returns (Info memory);
+
+    function readPair(uint256 key) external view returns (uint32, uint32);
+}
--- a/solidity/ecdsa/contracts/test/MockContract.sol
+++ b/solidity/ecdsa/contracts/test/MockContract.sol
@@ -1,0 +1,518 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+pragma solidity ^0.8.0;
+
+/// @notice Test-only programmable mock. Answers any call according to a table
+///         set up from the test, and records the calls it receives.
+///
+///         This is the contract half of the replacement for the archived
+///         defi-wonderland smock package. smock worked by reaching into
+///         Hardhat's provider internals, which is why it broke on Hardhat
+///         >= 2.20 and why it cannot be carried forward. Everything here is
+///         ordinary EVM state, so it survives any Hardhat, ethers, viem or
+///         Foundry change.
+///
+///         The model is deliberately Foundry's `vm.mockCall`: a
+///         calldata-to-returndata table plus call recording. If the suite ever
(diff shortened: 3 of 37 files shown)


piotr-roslaniec (approved):
Self-approve as repo admin. Dependency-only bump (hardhat 2.10.0 -> 2.19.5, smock 2.0.7 -> 2.3.4, hardhat-gas-reporter 1.0.8 -> 2.3.0). No test/source changes. Author's body documents why this exact combination is required to keep allowBlocksWithSameTimestamp while smock still works. CI is green.

PR #4204, opened 2026-07-27
build(solidity): raise hardhat to 2.29, which Node 24 needs

mswilkison:
**Stacked on #4203.** The last dependency step: hardhat 2.19.5 → **2.29.0** in
both packages.

smock capped this at 2.19.5. It is gone as of #4203, so the ceiling lifts —
and 2.26+ is what Node 24 requires. Below it, Hardhat's solc download fails on
Node 24 with `HH502`, because npm's undici and Node 24's bundled undici share a
global-dispatcher symbol.

## Verified

Whole suite, both packages, both runtimes:

| | Node 22 | Node 24 |
| --- | --- | --- |
| `ecdsa` | 673 / 44 pending / 0 | **673 / 44 / 0** |
| `random-beacon` | 955 / 0 pending / 0 | **955 / 0 / 0** |

Identical to the pre-bump results, so nothing regressed — and it is safe to
land before or after the Node bump in #4201, in either order.

## The other thing 2.29 brings

| | before | after |
| --- | --- | --- |
| `random-beacon` | 3m | **23s** |
| `ecdsa` | 2m | **31s** |

That is EDR, the Rust execution engine, backported into the 2.x line.

## Two tests needed adjusting

`challengeDkgResult` has an inline gas check, and two tests deliberately
underfund the call to trigger it. hardhat 2.29 defaults to a newer hardfork,
where EIP-7623's intrinsic calldata floor is enforced *before* execution:

```
ProviderError: Transaction requires gas floor of 161720 but got limit of 100000
```

The node rejected the transaction outright, so the contract's own check never
ran — the tests had stopped testing the thing they name. The limits move to
just above each floor (`100000 → 170000`, `200000 → 220000`), still far short
of what the call needs to complete, so both revert inside the contract as
intended.

The alternative was pinning the hardfork back to 2.19.5's default. Rejected:
mainnet is past it, and running the suite against an older EVM than production
is a worse trade than adjusting two gas limits.

## After this

Node 22 → 24 in the five workflows, two Dockerfiles and two `engines` fields —
a small follow-up, and the reason the whole chain existed. #4201 already moved
those from 18 to 22; this makes 24 reachable for the first time.

## Summary by CodeRabbit

* **Chores**
  * Updated the Hardhat development tooling to version 2.29.0 for the ECDSA and random beacon packages.

* **Tests**
  * Updated gas-limit scenarios to accurately validate insufficient-gas behavior under current transaction gas rules.
  * Clarified test coverage for contract-level gas checks.

### DIFF
--- a/solidity/ecdsa/package.json
+++ b/solidity/ecdsa/package.json
@@ -55,7 +55,7 @@
     "eslint": "^7.32.0",
     "ethers": "^5.5.3",
     "fs-extra": "^11.2.0",
-    "hardhat": "2.19.5",
+    "hardhat": "2.29.0",
     "hardhat-contract-sizer": "^2.3.0",
     "hardhat-dependency-compiler": "^1.1.2",
     "hardhat-deploy": "^0.11.11",
--- a/solidity/ecdsa/test/WalletRegistry.CustomErrors.test.ts
+++ b/solidity/ecdsa/test/WalletRegistry.CustomErrors.test.ts
@@ -493,7 +493,9 @@ describe("WalletRegistry - Custom Errors", () => {
         // The error will be triggered by the inline gas check at the end of challengeDkgResult
         await expect(
           walletRegistry.connect(unauthorized).challengeDkgResult(dkgResult, {
-            gasLimit: 100000, // Intentionally low gas
+            // Above EIP-7623's intrinsic calldata floor, which hardhat 2.29
+            // enforces before execution, and far below what the call needs.
+            gasLimit: 170000,
           })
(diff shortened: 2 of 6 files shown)


piotr-roslaniec (approved):
Self-approve as repo admin. hardhat 2.19.5 -> 2.29.0 in both packages. 6 files, 664 +/- 2046. yarn.lock updates remove @chainsafe/* (replaced by @ethereumjs/*). Two test gas-limit adjustments (100000->170000, 200000->220000) explained in-line with EIP-7623 floor reasoning.

PR #4219, opened 2026-07-27
fix(deps): lift btcd off the vulnerable v0.22.3 downgrade pin

mswilkison:
## Summary

`go.mod` replaces the whole `btcd` module with **v0.22.3** - *below* the `v0.23.2` the module itself requires - which keeps three already-fixed security advisories open against the Bitcoin consensus, wire, and script packages the bridge actually compiles. (Three vulnerabilities, six identifiers counting the GHSA aliases.)

```
go.mod  github.com/btcsuite/btcd v0.23.2                              // require
go.mod  github.com/btcsuite/btcd => github.com/btcsuite/btcd v0.22.3  // replace — wins
```

This is long-standing on `main` and not specific to any release branch. It surfaced during a dependency sweep of #4109; that PR's integration branch inherits it, and this fix is deliberately targeted here rather than stacked on the release candidate so it gets CI and normal review.

Unlike an unused-code advisory, this code **is** in the build — `go list -deps ./...` returns 16 `btcsuite/btcd` packages including `blockchain`, `blockchain/indexers`, `btcjson`, `btcutil`, `txscript`, and `wire`.

| Advisory | Package | Fixed in | Status before | After |
|---|---|---|---|---|
| GO-2022-1098 / GHSA-2chg-86hq-7w38 — DoS in message decoding, witness size checking | `wire` | 0.23.2 | **open** (below the repo's own require) | fixed |
| GO-2024-2818 / GHSA-3jgf-r68h-xfqm — consensus failures | `blockchain`, `txscript` | 0.24.0 | **open** | fixed |
| GO-2024-3189 / GHSA-27vh-h6mc-q6g8 — incorrect `FindAndDelete` re-implementation | `txscript` | 0.24.2-beta.rc1 | **open** | fixed |

## Why the downgrade was there, and why it can go

btcd v0.23 extracted `btcd/btcec` into the separate module `btcd/btcec/v2`, and later versions don't ship the pre-split package at all. Two consumers still import the pre-split path and can't simply be edited: the pinned **tss-lib fork** (`crypto/ckd`, `tss/curve`) and **first-party key-handling code** (`pkg/crypto/ephemeral`, `pkg/crypto/secp256k1`, `pkg/net/...`, `pkg/chain/local_v1`, `pkg/internal/pbutils`), which relies on v1 `btcec` types being aliases of the `crypto/ecdsa` types.

So one crypto package was holding the entire module — consensus code included - a minor version back.

This PR serves *just that package* from an in-tree copy of the btcd v0.22.3 `btcec` sources (`third_party/btcsuite/btcec`) and moves the module itself to **v0.24.2**. The vendored copy is byte-identical to upstream apart from a mechanical `gofmt` pass; `VENDOR.md` records provenance, a reproducible `diff -r` verification recipe, and the exit path (delete the directory once tss-lib migrates to `btcec/v2`). The upstream btcec test suite ships with it and passes.

### Why v0.24.2 and not v0.26.2

**v0.24.2 is the lowest version that fixes all three advisories** (the last one lands in `0.24.2-beta.rc1`), and going higher isn't currently possible:

- **v0.25.0 requires `btcec/v2` v2.3.5**, and `btcec/v2` **v2.3.4** changed `ecdsa.SignCompact` from `([]byte, error)` to `[]byte`. go-ethereum v1.13.15 calls the two-value form in `crypto/signature_nocgo.go`, so pulling `btcec/v2` that far forward breaks **every cgo-less build** - including the cross-compiled release binaries that `make release` produces. Raising btcd past v0.24.2 requires upgrading go-ethereum first, which is out of scope here.
- **v0.26** additionally restructured btcd: `wire`, `txscript`, `chaincfg`, `chainhash`, and `btcutil` moved to separate `/v2` modules and the plain-path packages no longer exist. The go-electrum and tss-lib forks import the plain paths with import strings we don't control, so v0.26.x isn't reachable without forking both.

So this is a compatibility ceiling, not a security gap - v0.24.2 is advisory-clean. Both constraints are recorded in the `go.mod` comment so the next person to touch this pin doesn't have to rediscover them.

### The `btcd/v2` alias goes too

`replace github.com/btcsuite/btcd/v2 => github.com/btcsuite/btcd v0.23.4` only isolates *self-contained* packages. Any aliased package that imports a sibling btcd package does so by plain path, which resolved back to the **v0.22.3** module - so the "patched" alias was linking vulnerable `wire`/`txscript` underneath it. With the downgrade gone the alias has no purpose, and the two `pkg/bitcoin/electrum` files using it return to plain paths.

## First-party fallout

- `txscript.NewTxSigHashes` takes a `PrevOutputFetcher` since v0.23 and (as of v0.24.2) requires the previous outputs of all inputs. `TransactionBuilder` now registers each input's locking script and value in a `MultiPrevOutFetcher` as inputs are added - it already fetched exactly that data for its sighash args. `ComputeSignatureHashes` also asserts the set is complete and returns an error, rather than letting a future gap reach the panicking path.
- `mempool.GetTxVirtualSize` / `btcutil.Hash160` / `btcutil.NewTx` move from the archived `btcsuite/btcutil` to `btcd/btcutil` (same semantics).

### Deserialization cap and recover - required by the version jump

btcd v0.24.2 rewrote `MsgTx.btcDecode` to slice every script of a transaction out of **one fixed-size 4 MiB buffer** (`scriptSlabSize = 1 << 22`), advancing it after each read. `readScriptBuf` bounds each individual script against `maxWitnessItemSize` but never against the *remaining* buffer, so once the cumulative script length of a single transaction passes 4 MiB the next slice **panics** instead of returning an error:

```
panic: runtime error: slice bounds out of range [:900000] with capacity 594304
  btcd@v0.24.2/wire/msgtx.go:1041  readScriptBuf
  ...
  pkg/bitcoin/transaction.go       (*Transaction).Deserialize
```

Reproduced directly: 4 × 900 KB output scripts decode cleanly, 5 panic. Both decoders this PR replaces (v0.22.3 and the v0.23.4 alias) allocate per script and handle the same bytes fine, so this would be a **regression introduced by the upgrade** - on a path fed by the untrusted Electrum backend. `decodeTransaction` runs *before* `GetTransaction`'s txid check, `GetTransactionConfirmations` performs no hash check at all, and no `recover()` covers the tbtc signing paths, so it would terminate the node. Upstream fixed it only in the split-out `wire/v2` module, part of the v0.26 layout the plain-path forks make unreachable.

The fix is two layers, and **only the combination closes the panic**: a `MaxTransactionByteLength = 4_000_000` cap applied in `bitcoin.Transaction.Deserialize` and `electrum.decodeTransaction` (a transaction's weight is `3*base_size + total_size` and cannot exceed the 4,000,000 WU maximum block weight, so a consensus-valid transaction is always strictly shorter than the cap - nothing legitimate is rejected), **plus** a `recover()` that turns the slice-bounds panic that can still fire below the cap into a plain error. Two regression tests (`TestTransaction_DeserializeDeclaredNotDelivered` and `TestDecodeTransactionDeclaredNotDelivered`, both added in `626538d2`) construct a 194,378-byte (189.8 KiB, 4.86% of the cap) transaction that panics in unpatched btcd and confirm the `recover()` returns the error. The cap alone is *not* sufficient - without the `recover()`, payloads of that size terminate the process. The cap bounds grossly oversized input; the `recover()` is what actually turns the under-cap panic into an error.

A subsequent commit also pre-checks `len(rawTx)/2` against `MaxTransactionByteLength` in `electrum.decodeTransaction` *before* `hex.DecodeString` allocates, so a malicious multi-GB hex string can't pay the full allocation before the guard runs.

Note: the `recover()` covers slice-bounds panics from `readScriptBuf` only. It does **not** cover fatal runtime errors such as out-of-memory from attacker-declared varint counts that btcd's wire decoder honors before reading backing bytes. Closing that class of input requires either a pre-decode budget check on declared input/output counts or a future btcd release; both are out of scope here.

The v0.22.3-to-v0.24.2 jump also changes btcd's script-decoding internals beyond the slab panic. Two material changes worth knowing for future triage:

- The shared script free-list ceiling rises from 6.4 MB globally (12,500 × 512 B entries) to roughly 524 MB globally (125 × 4 MiB slabs) - a ~82× increase, not ~8000× (the earlier draft conflated bytes-per-entry with bytes-per-slab). Per-decode worst case *becomes* one 4 MiB slab (v0.22.3 allocated per script); only the global retention grows.
- The per-script cap tightens from 32 MiB (`MaxMessagePayload`) to 4,000,000 (`maxWitnessItemSize`) - strictly safer, no legitimate transaction is affected.

Neither is a security issue; both are documented in the corresponding `MaxTransactionByteLength` doc comment and in `VENDOR.md`.

### Docker

`third_party/` is copied before `RUN go mod download`. The directory `replace` target must exist for the module graph to resolve, so the image build would otherwise fail at that step even though local builds pass.

## Verification

- `go build ./...`, `gofmt -l .` clean; full `go test ./...` green.
- `CGO_ENABLED=0 go build` for `darwin/amd64`, `darwin/arm64`, `linux/amd64`, `linux/arm64`, `windows/amd64` - the cgo-less cross-compile path `make release` uses, which selects go-ethereum's `signature_nocgo.go` and is where the `btcec/v2` incompatibility above shows up.
- `go vet ./...` reports only the pre-existing `pkg/tecdsa/signing/protocol.go` copylock warning, byte-identical on unmodified `main`.
- `TestTransactionBuilder_Signing` byte-vector fixtures pass: DER signatures, sighashes, and compressed public keys are **byte-identical** across the version jump. The signing path still uses v1 btcec (now vendored, byte-identical), so its implicit BIP-62 low-S normalization is preserved.
- Vendored-copy identity verified via the `VENDOR.md` recipe: only the four added files (`go.mod`, `go.sum`, `LICENSE`, `VENDOR.md`) differ from gofmt'd upstream `btcec@v0.22.3`.
- `docker build --target build-sources` passes, including the `go mod download` step that fails without the `third_party` COPY.
- Panic reproduced before the fix and confirmed to return a clean error after.

## Known residuals

- The archived `github.com/btcsuite/btcutil` (2019) is still linked, because the tss-lib fork's `crypto/ckd` imports its `base58` package - the same unchangeable-import-string constraint that motivates the vendoring. One pure codec package, no advisories against it; first-party code no longer touches it. MVS now selects an ~20-months-older `btcsuite/btcutil` pseudo-version than `main` did (the archived `v1.0.3-0.20201208143702` direct require is gone, replaced by the tss-lib fork's `v0.0.0-20190425235716-9e5f4b9a998d`); the only compiled package (`base58`) is functionally identical between the two commits, so this is benign.
- `go.sum` keeps `/go.mod`-only hash lines for two old btcd pseudo-versions reached through transitive requires that MVS overrides. No `h1:` hashes, so no source is compiled; the vulnerable `v0.22.3`/`v0.23.4` full hashes are gone. Naive graph-level SCA tools may still surface them.
- The vendored `btcec` sits at a module path OSV doesn't index, so a hypothetical *future* advisory against pre-split btcec code wouldn't be flagged automatically. None of the current three are in `btcec`. The vendored dir is also invisible to root-module `go test ./...` and has no automated advisory monitor; byte-identity relies on a human running the `VENDOR.md` recipe.
- The versionless directory `replace` means external modules importing keep-core packages can't resolve the btcec requirement. Not a new blast radius: `main` already carries the same unresolvable-with-replace pattern for `google.golang.org/protobuf/dev`, which downstream consumers cannot resolve either. The btcec pseudo-version is new on this branch, but the underlying pattern is pre-existing.

## Test plan

- [x] `go build ./...` / `go vet ./...` / `gofmt -l .`
- [x] `go test ./...` (full suite)
- [x] Vendored btcec upstream test suite, from the main module and standalone
- [x] `VENDOR.md` byte-identity verification recipe
- [x] `docker build --target build-sources`
- [ ] CI green on this branch

## Note for #4109

The security-release integration branch carries the same downgrade. Once this lands, its next rebase picks the fix up; the `go.mod` resolution there is mechanical (that branch has a much newer dependency set, so the conflict is in surrounding lines rather than the btcd ones).

## Summary by CodeRabbit

* **Bug Fixes**
  * Bitcoin transaction decoding now rejects oversized data and returns errors instead of crashing on malformed input.
  * Signature-hash generation now reports missing previous transaction outputs.
  * Updated Bitcoin cryptography support improves compatibility and dependency resolution.

* **Security**
  * Added secp256k1 key, signature, encryption, and decryption capabilities with validation safeguards.

* **Tests**
  * Expanded coverage for transaction handling, cryptographic operations, key parsing, signatures, and error scenarios.


mswilkison:
CI is green (21/21). One correction to the version choice above, worth recording since it constrains any future btcd bump.

The first push used **btcd v0.25.0** and `client-build-test-publish` failed:

```
signature_nocgo.go:85:14: assignment mismatch: 2 variables but btc_ecdsa.SignCompact returns 1 value
```

v0.25.0 requires `btcec/v2` v2.3.5, and **v2.3.4 changed `ecdsa.SignCompact` from `([]byte, error)` to `[]byte`**. go-ethereum v1.13.15 calls the two-value form in `crypto/signature_nocgo.go` — the file selected when cgo is disabled, i.e. exactly the cross-compiled release binaries `make release` produces. Nothing on the cgo path is affected, which is why it passed local builds, `go vet`, the full test suite, and a `docker build --target build-sources` (the cross-compile lives in the later `build-bins` stage).

Now pinned to **v0.24.2**, which requires `btcec/v2` v2.1.3 so MVS leaves it at v2.2.0. **All three advisories remain cleared** — v0.24.2 is the release that fixes the last of them (GO-2024-3189, in `0.24.2-beta.rc1`).

That makes the pin exact in both directions: it can't go lower without reopening an advisory, or higher without upgrading go-ethereum first. Both constraints are in the `go.mod` comment. Verified with `CGO_ENABLED=0 go build` across `darwin/amd64`, `darwin/arm64`, `linux/amd64`, `linux/arm64`, `windows/amd64` — and confirmed the old pin reproduces the CI error under the same check, so it's a real regression test rather than an assumption.

Note for whoever rebases `security-release/candidate-1`: that branch carries go-ethereum v1.17.3, so it tolerates v0.25.0 and could go higher — but v0.24.2 is advisory-clean there too, so the merge needs no adjustment.

lrsaturnino:
Reviewed the btcd downgrade pin and found one issue, now fixed:

**MaxTransactionByteLength cap does not close the panic it was added for (Critical)** — the cap only bounds a transaction's total wire length, but the underlying panic is triggered by the cumulative *declared* script/witness length within a single transaction exceeding btcd's shared, fixed-size 4,194,304-byte script buffer. A payload that delivers roughly 194 KB of real script data and then merely declares, without delivering, one more large item still panics — well under the byte-length cap. Fixed by wrapping both untrusted-input decode paths (`bitcoin.Transaction.Deserialize` and `electrum.decodeTransaction`) in a `recover()` that converts any panic from the underlying btcd decoder into a returned error, correcting the `MaxTransactionByteLength` doc comment (it implied the length check alone closed this panic), and adding regression tests in both packages covering the minimal reproduction. Commit 626538d27ed3c31c8a7a2913d8ff3c51baa49acc.

No other issues found.

lrsaturnino (dismissed):
1 finding fixed and pushed; no items requiring manual follow-up.

piotr-roslaniec (approved):

PR #4226, opened 2026-08-06
fix(tbtc): anchor the emergency rekey, and correct what the certified floor bounds

mswilkison:
Six fixes for the FROST native signer anchor path. Stacked on #4199 rather than folded into it, so that PR stays mergeable as-is.

None of these change the Rust signer: `frost_tbtc_trigger_emergency_rekey` already exists at the pinned ref, so the ABI and `ci/frost-signer-pin.env` are untouched.

## 1. Route the emergency rekey through the state anchor (fdbfdd28a)

The engine has always exported the kill-switch trigger, but **nothing in Go called it** — a repo-wide grep for the symbol returns nothing. With no Go caller, the only way to arm the switch is to stop the node and mutate its durable store directly. That write never passes through the anchored operation path, so it is never compare-and-swapped onto the anchor stream. On restart the anchor sees local and remote agreeing, and an operator who restores the pre-rekey state file first has erased a kill switch aimed at them with no evidence anywhere in the stream.

The fix routes the call through `callBuildTaggedTBTCSignerOperation`, so the durable write is CAS-ed onto the anchor stream before the call returns. That is what makes a later erasure detectable.

Two properties that look like omissions but are deliberate, and are documented at the entrypoint:

- **Not admission-gated.** The operation wrapper takes no capacity reservation, and admission refuses all work once headroom reaches the rotation floor (256) while the barrier keeps admitting until the certified window is genuinely exhausted. A kill switch that capacity accounting can veto is not a kill switch, so it runs unreserved in that band.
- **The barrier stays mandatory.** A poisoned anchor or an exhausted window fails the trigger. Every barrier refusal predicate is operation-independent, so a state in which this trigger is refused is one in which the node already refuses every signature-producing call — the switch is redundant there, not defeated, and its residual is availability rather than authority.

The exported call is single-flight by contract: the engine treats an armed event as immutable and no export clears it.

**No runtime behaviour changes from this PR.** The kill switch remains unreachable from any operator-facing CLI; the anchored path is landed so any trigger built on it will be certified. The remaining options for that trigger surface each need a decision the code cannot supply: what names the target session, whether the switch is per-wallet or node-wide, and what the recovery path is for an accidental trigger given the event is immutable. The #4222 exposure stays open until those answers exist. Note that an out-of-process CLI is specifically not a viable path here: with the running node holding the store's exclusive `flock` it cannot write, and with the node stopped the write is unanchored — the same bug restated — but that constraint is one input to the design, not the whole reason the surface is deferred.

Quarantine, fault scores, and `finalize_request_fingerprint` have the same missing-wiring shape but **no runtime writers in the Rust crate at all** — only test-support writers. A Go call into a nonexistent writer is not a fix, so those need a Rust-side writer first and are filed separately.

## 2. Correct what the anchor certified floor bounds (eb6692c71)

`docs/development/frost-anchor-rotation.adoc` said the floor is "checked at startup against pins that the running node cannot influence." That holds against a **compromised signer process**, which cannot forge the offline authority's signature. It does not hold against the **operator**: every artifact naming the floor — certificate chain, init config, activation manifest — is a local file the operator owns, and the certificate sequence and digest are only ever compared against the init config. The activation manifest is a signed envelope read from disk, not fetched from chain.

The doc now says that, and adds what genuinely bounds an operator-side downgrade:

1. The bundle moves in lockstep — every rotation must change the manifest hash and advance the manifest sequence and service epoch by exactly one, so a downgrade means reverting the whole bundle, not swapping one file.
2. The durable trust journal is append-only and its head must equal the configured pin, so a downgrade also needs a pre-rotation signer-store snapshot.
3. The anchor service can retire the previous binding — every signed request already carries the current `bindingHash` and `activationManifestSequence`.

Only (3) is a real control, and it lives outside this repository, so it becomes a pre-activation checklist item alongside a decision on where a monotonic floor is published if operator-side rollback is in scope.

**No code change closes this gap**, and the PR does not pretend otherwise: every candidate in-repo pin is downstream of an operator-owned file. Putting the floor *reference* into the activation manifest is not merely awkward but unsatisfiable — `manifestHash` feeds `bindingHash` feeds the event root, which *is* the floor, so the manifest would have to commit to a hash of itself. A plain minimum-sequence integer would dodge that circularity but is security-null here, because an old certificate is only presentable together with its own old manifest, which would carry the old minimum.

Worth noting for whoever picks up the checklist item: the signed activation-handshake attestation already exports `trustCertificateSequence`, `anchorServiceEpoch` and the certified floor revision/generation, so an external monitor retaining the maximum ever seen per operator detects a downgrade today with no protocol change.

## 3. Diagnose a durable trust head ahead of the configured pin (eaae1fcd8)

A durable trust head ahead of the configured pin now says the artifacts are older than the store, rather than surfacing later as an opaque endpoint-identity mismatch. The comment states explicitly that this is **not** a bound on operator rollback — reverting the bundle and the store together leaves the two equal and fires nothing — so a later reader does not mistake it for a fix.

## 4. Publish restartable headroom on every acknowledged commit (cd83062ad)

The restartable headroom mirror and the rotation warning were previously published only on a full readiness reconciliation, so a steadily signing node and a stalled node looked identical at the scrape until the next reconciliation. Since pre-sign authorization caches readiness per finality window, an entire authorized batch could burn window invisibly inside one window. Every successful compare-and-swap already holds an authenticated tip, so the publication moves onto the commit path and the rotation warning follows the seat count that travelled with the previous publication (seat count only changes on DKG and retirement, both of which force a full reconciliation).

## 5. Correct the runbook cadence and seat-scaling framing (c07b47578)

`docs/development/frost-anchor-rotation.adoc` carried stale workload numbers in several places. The corrected table derives each cell from the fault-free range `k+3..3k+2` generations per input: at 4 seats the worst case is 294 generations (was 588 — the retracted figure), the range at the 5-seat expected holder is 13–26 sweeps per epoch (was 6–13), and at 20 seats 2–7 sweeps (was 1–2). The seat-sizing directive was also rewritten: the original framing assumed equal-weighted signers and recommended sizing against the 5-seat average; this PR removes that assumption, names the sortition as stake-weighted with replacement, and reframes the high-seat rows as the upper tail of the distribution reachable by operators with concentrated stake.

## 6. Measure real anchor burn rate (69a24d045)

Every workflow that reaches the shared `reserve()` path now increments its own counter, so the published burn-rate ratio has a single well-defined unit. The single `admitted_total` counter is replaced by three: `admitted_presign_input_total`, `admitted_presign_relay_gate_total`, and `admitted_dkg_total`. The pre-sign reservation counts the per-input work; the relay-gate reservation counts the up-front authorization; DKG counts the per-local-seat reservation. The reset helper, var-block comment, and registration function doc were updated to match.

## Verification

```
go build ./...
go vet ./pkg/tbtc/... ./pkg/frost/signing/...
go test -run 'FrostNativeSignerAnchor' ./pkg/tbtc/
go test -run 'EmergencyRekey' ./pkg/frost/signing/                                    # default build: stub fails closed
CGO_ENABLED=1 go test -tags "frost_native frost_tbtc_signer" -run 'EmergencyRekey' ./pkg/frost/signing/
```

New tests cover the payload builder (empty/whitespace reason, over-long and malformed session IDs), the response decoder (six rejection cases, including the engine retargeting a per-signing session to its wallet session), the default-build stub failing closed, the rotation warning gauge value in both healthy and near-floor scenarios, the observer threading warning and seat count through to the accessors, the consumption and admission counters advancing on commit and staying flat on the skip-CAS path, the new cgo-tagged wiring and barrier-gating for the trigger, and both the honest and rejected paths of the new trust-head assertion.

### DIFF
--- a/docs/development/frost-anchor-rotation.adoc
+++ b/docs/development/frost-anchor-rotation.adoc
@@ -47,8 +47,8 @@ unconsumed window at once - so no seat count a wallet can award is
 excluded from signing. What is still finite is the number of inputs a
 node can actually sign between two rotations: a fault-free 21-input
 sweep consumes roughly `21 * (3 * localSeats + 2)` revisions and one to
-two generations per call, so a 4-seat node gets *six to thirteen
-full-size sweeps per anchor epoch* and a 20-seat node one to two. The
+two generations per call, so a 4-seat node gets *13-26 sweeps
+per anchor epoch* and a 20-seat node 2-7 sweeps. The
 derivation is in <<when-rotation-is-required>>.
 
 == What the anchor is, in operator terms
@@ -276,10 +276,10 @@ charged against that bound rather than against real consumption:
 |===
 | Local seats | Calls per 21-input sweep | Revisions consumed | Generations consumed
 
-| 1 | 105 | 105 | 105-210
-| 2 | 168 | 168 | 168-336
-| 3 | 231 | 231 | 231-462
-| 4 | 294 | 294 | 294-588
+| 1 | 105 | 105 | 84-105
+| 2 | 168 | 168 | 105-168
+| 3 | 231 | 231 | 126-231
+| 4 | 294 | 294 | 147-294
 |===
 
 === Cadence
@@ -304,9 +304,9 @@ full-size sweeps one at a time:
 | 2  | 105-168   | 22-36 |
 | 3  | 126-231   | 16-30 |
 | 4  | 147-294   | 13-26 |
-| 5  | 168-357   | 10-22 | the expected holder under equal weighting
-| 20 | 483-1302  | 2-7   | only reachable if the operator set shrinks to ~5
-| 100| 2163-6342 | 0-1   | not reachable under equal weighting; see below
+| 5  | 168-357   | 10-22 | the average mainnet holder
+| 20 | 483-1302  | 2-7   |
+| 100| 2163-6342 | 0-1   | the protocol maximum, reachable only when one operator holds all the stake
 |===
 
 The 100-seat row is the honest edge of the model. Such a node is
@@ -314,13 +314,18 @@ The 100-seat row is the honest edge of the model. Such a node is
 consume more window than one epoch has, so it may be refused part way
 through and need a rotation to finish. That is a throughput limit an
 offline ceremony fixes, not a permanent exclusion from the signing
-threshold, which is what the old ceiling was.
+threshold, which is what the old ceiling was. It is also not a
+configuration any real wallet produces: one operator holding all 100
+seats of a 100-seat wallet is the protocol maximum, not a sortition
+result.
 
-IMPORTANT: Seat count is `groupSize / operatorCount` because signers are
-equal-weighted, so the expected holder is around 5 seats and the high-seat
-rows are reachable only if the operator set shrinks a long way. Size
-against the 5-seat row, and treat 20 and 100 as sensitivity checks on
-operator-set size rather than as a stake-concentration tail.
+IMPORTANT: Signers are stake-weighted, so seat count tracks the operator's
+stake share. The expected holder's seat count tracks the operator's
+stake share, and the high-seat rows are reachable in the upper tail of
+the seat-count distribution - in practice, by operators with concentrated
+stake. Size against the row that matches your network's realistic stake
+concentration, and treat 20 and 100 as sensitivity checks on that
+concentration rather than as the expected case.
 
 NOTE: Two earlier sets of figures in this document are wrong and should
 not be used. An operability assessment put these at "1-2 at 4 seats,
@@ -539,11 +544,6 @@ answers.
 
 === Gaps in this area
 
-* *No metric.* Nothing anchor-related is registered with `clientinfo`,
-  so `/metrics` has no headroom gauge and no rotation warning. The only
-  scrapeable trace of a refusal is the generic
-  `wallet_action_failed_total` counter, which cannot distinguish a
-  headroom refusal from any other action failure.
 * *No log line.* Neither the readiness snapshot nor the anchor binding
   logs headroom. The first log evidence is the refusal itself, by which
   time the node is already turning work away.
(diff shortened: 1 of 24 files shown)


mswilkison:
Two further commits, both from the analysis in #4222 and both required under every candidate design there.

**`eaae1fcd8` — pin the store fingerprint in the descendant validator.** The Go and Rust anchor-trust descendant validators disagreed about which references are admissible, in both directions: Rust alone refused a checkpoint store-fingerprint change across generations, Go alone required the certified floor to be revision 1 of its service epoch. Neither gap is reachable today, but the more permissive tree would accept a chain the other refuses on *every store open*, with no truncation or rebase path back — a fail-closed brick of a signer store. This adds Rust's rule here; the mirrored genesis-floor rule lands in **#4227**, stacked on #4198. Union, not intersection.

**`cd83062ad` — publish headroom on acknowledgement, and export the rotation warning.** The headroom mirror refreshed only on successful readiness reconciliation, so a node signing steadily and a node that had stalled were indistinguishable at the scrape — and because pre-sign authorization caches readiness per finalized point on a branch that deliberately publishes nothing, an entire authorized batch could burn window with the gauge unchanged throughout. Publication now happens at the acknowledgement install, which every successful CAS already passes through with an authenticated tip in hand; that single choke point covers the ordinary commit, the startup local-ahead commit, and read-recovery.

The rotation warning gets its own gauge. It was computed but reachable only via readiness snapshots and the activation handshake — a hand-assembled loopback challenge that hard-errors once headroom reaches zero, which is exactly when an operator needs it. It can't be derived at scrape time from the two headroom gauges, because the workload term needs the node's largest local seat count, so the seat count now travels with the pair; that also lets the commit path recompute the warning without an inventory of its own.

A partial computation leaves the previous reading standing rather than publishing half a pair — the two dimensions are only meaningful together, and a stale pair is more honest than a half-fresh one.

`ci/frost-signer-pin.env` is deliberately **not** bumped. The pin must not point at an unmerged side branch; it should move once #4227 merges into #4198.

Verified: `go build ./...`, `go vet`, full `pkg/tbtc` and `pkg/frost/signing` suites. New tests cover the mirrored fingerprint rejection, the rotation-warning gauge (including that an unpublished mirror reads 0 exactly as the headroom gauges do, so alerts must still qualify on `headroom_observations_total`), and that an acknowledged tip refreshes the mirror while an unauthenticated one leaves it untouched. The registered-source-count pin test caught the new gauge on the first run, which is the test working as intended.

mswilkison:
Follow-up DR work is available in #4241. It is intentionally stacked on this
PR rather than appended here, so the anchor-integrity review remains focused.
Its native ABI/store dependency is #4240.

PR #4227, opened 2026-08-06
fix(tbtc/signer): require a genesis floor in the anchor descendant validator

mswilkison:
Stacked on #4198. Pairs with the mirrored Go-side change in #4226.

## The divergence

The Go and Rust anchor-trust descendant validators disagree about which references are admissible, **in both directions**:

| check | Go | Rust |
|---|---|---|
| certified floor must be revision 1 of its service epoch | ✅ | ❌ (this PR) |
| checkpoint store fingerprint fixed across generations | ❌ (#4226) | ✅ |

Neither gap is reachable today. Bootstrap and rotation endpoints both force a revision-1 `To` reference, so every call site already passes one — the full suite passes unchanged with the check added, which is the evidence for that claim rather than an assumption about it.

## Why fix a latent gap

A divergence between the trees is not a cosmetic asymmetry. Whichever tree is more permissive would accept a certificate chain the other refuses — and the refusal happens on **every store open**, since the Rust trust journal is replayed and re-validated fail-closed each time. There is no truncation, rebase, or compaction path back from that state.

So the failure mode is a deterministic brick of a permissioned signer store, recoverable only by re-provisioning onto a fresh store identity — which abandons every DKG key package bound to the old one. Fail-closed is the safe direction, and it is still a dead store. The two validators should not be able to drift apart again.

## The fix

Assert the **union** of the two rule sets rather than the intersection. Do not relax Go: its floor-revision rule is load-bearing at startup, where the floor arrives as external data from the Rust readback and the check is genuine defense in depth.

This PR adds Go's rule to Rust. The mirrored fingerprint check lands on the Go side in #4226.

Validation tightening only — no wire format, encoding, or ABI change, so no ABI revision. **`ci/frost-signer-pin.env` is deliberately not bumped here**: the pin must not point at an unmerged side branch. It should move to this commit once this merges into `codex/signer-store-identity-abi`.

## Verification

```
cargo fmt -- --check
cargo clippy --all-targets -- -D warnings
TBTC_SIGNER_STATE_PATH=/tmp/... cargo test     # 369 passed, 0 failed
```

Two new tests pin both halves of the parity so a future edit to either validator has to confront the other: one asserts a non-genesis floor is refused while a revision-1 floor still admits the same descendant (so the rule rejects the floor shape, not ordinary descendants), and one asserts a store-fingerprint change across generations is refused.

### DIFF
--- a/pkg/tbtc/signer/README.md
+++ b/pkg/tbtc/signer/README.md
@@ -472,6 +472,15 @@ storage guarantees for that hardware-level failure boundary.
     firewall enabled and restrict `TBTC_SIGNER_POLICY_ALLOWED_SCRIPT_CLASSES`
     to the intended output classes, such as `p2tr`.
 - Transcript accountability / quarantine config:
+  - **Runtime fault recording is not implemented.** The engine enforces a
+    quarantine set and fault scores wherever they are read, but nothing at
+    runtime ever writes them: they are only loaded from persisted state. So
+    enabling `TBTC_SIGNER_ENABLE_AUTO_QUARANTINE` validates its configuration
+    and then does nothing - no operator is ever scored, and no operator is ever
+    automatically quarantined. Enforcement of a quarantine set that is already
+    present in the store does work, and the failure direction is safe (no
+    operator can be falsely quarantined), but do not rely on automatic
+    exclusion. Operator removal is out-of-band today.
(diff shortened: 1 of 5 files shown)


mswilkison:
Follow-up DR work is available in #4240. It is intentionally stacked on this
PR rather than appended here, so this validator-parity review remains focused.
The companion Go orchestration/cutover PR is #4241.

PR #4327, opened 2026-09-10
Merge keep-common pkg/ into pkg/keepcommon/, bump go-ethereum to v1.17.3

piotr-roslaniec:
## Summary

Folds `threshold-network/keep-common`'s `pkg/*` tree into `keep-core` as an
in-tree subtree at `pkg/keepcommon/`, with full commit history preserved,
and bumps `keep-core`'s `github.com/ethereum/go-ethereum` dependency from
`v1.13.15` to `v1.17.3` to resolve a version mismatch this merge surfaced.

### (a) Subtree merge: `keep-common/pkg/*` → `keep-core/pkg/keepcommon/*`

Performed with `git subtree split` (in a `keep-common` clone) + `git subtree
add` (into this `keep-core` clone), **not** `--squash`, so the full 448-commit
history of `keep-common`'s `pkg/` tree is preserved as a second parent of the
merge commit (`d09f3f56c`). `git log --oneline <merge-commit>^2` shows all
448 commits; per-file `git blame` inside `pkg/keepcommon/*` resolves through
this history normally.

Every keep-common package landed uniformly under `pkg/keepcommon/<name>`
(mirroring keep-common's own internal layout, e.g.
`pkg/keepcommon/chain/ethereum/ethutil`), rather than flattening the
non-colliding packages into `keep-core/pkg/<name>` directly. Rationale below.

Import paths were then rewritten 1:1 across the whole tree:
```
github.com/keep-network/keep-common/pkg/  ->  github.com/keep-network/keep-core/pkg/keepcommon/
```
83 files changed (67 outside `pkg/keepcommon`, plus 16 files *inside* the
newly-merged subtree that reference sibling keep-common packages by their
old external import path). Zero logic changes — every diff outside the
subtree-add commit is an import-path string substitution only.

### (b) Package-name collisions and how they were resolved

Three of keep-common's twelve top-level `pkg/` packages collide by name with
packages `keep-core` already has:

| keep-common package | Collides with | Nature |
|---|---|---|
| `pkg/chain/ethereum` | `keep-core/pkg/chain/ethereum` | Same path, unrelated content — keep-common's version is a small low-level Ethereum utility package (balance monitor, block counter, nonce manager); keep-core's own is the large production chain implementation + generated ABI/contract bindings. keep-core's `pkg/chain/ethereum/ethereum.go` already imports keep-common's version today, aliased `commonEthereum`. |
| `pkg/clientinfo` | `keep-core/pkg/clientinfo` | Same path, same package name, overlapping filenames (`clientinfo.go`, `diagnostics.go`, `metrics.go`, `metrics_test.go` exist in both). keep-core's version is a thin façade around keep-common's lower-level Prometheus primitives. |
| `pkg/subscription` | `keep-core/pkg/subscription` | Same path, identical filenames (`subscription.go`, `subscription_test.go`). |

**Resolution:** rather than case-by-case renaming of only the 3 colliding
packages, the entire keep-common-origin tree was nested uniformly under one
namespace, `pkg/keepcommon/`. This eliminates all collisions in one uniform
move, keeps the merge boundary a trivially auditable 1:1 import-path
substitution with zero per-package special-casing, and carries zero risk of
a silent file overwrite. No Go `package` declarations were touched — only
import path strings — so e.g. `pkg/keepcommon/clientinfo` still declares
`package clientinfo` and call sites still read `clientinfo.X` unaliased.

This is a real design decision a flatter placement (individually renaming
just the 3 colliding packages, landing the other 9 directly under
`keep-core/pkg/<name>`) was also viable and arguably more idiomatic
keep-core style; happy to revisit in review if the uniform-nesting default
isn't the preferred shape.

### (c) go-ethereum bump: v1.13.15 → v1.17.3

keep-common's HEAD (post `threshold-network/keep-common#31`, the
govulncheck-driven dependency bump) pins `go-ethereum v1.17.3`. keep-core
was still pinned to `v1.13.15`. Without bumping, `go test` fails to even
*compile* `pkg/keepcommon/chain/ethereum/ethutil`:

```
pkg/keepcommon/chain/ethereum/ethutil/rate_limiter_test.go:485:14: undefined: ethereum.TransactionReceiptsQuery
```

`ethereum.TransactionReceiptsQuery` only exists in go-ethereum >= v1.17.x;
the mock in that test file implements
`SubscribeTransactionReceipts(ctx, *ethereum.TransactionReceiptsQuery, ...)`.
No *production* code was affected (`go build ./...` was already clean
without the bump) — this was a test-only compile failure, but a real one.

Bumping resolves it outright: `go mod tidy` pulled forward go-ethereum's own
transitive deps (`spf13/cobra` v1.5.0→v1.8.1, `golang.org/x/crypto`
v0.32.0→v0.47.0, `golang.org/x/tools`, `bits-and-blooms/bitset`,
`consensys/gnark-crypto`, etc. — see `go.mod`/`go.sum` diff). The
`btcsuite/btcd/btcec` local replace directive (pinned specifically because
of past btcd/go-ethereum version interplay, see the comment above the
`replace` block) needed **no change** — `go build ./...` and the full
`go test ./...` suite both pass unmodified against the bumped go-ethereum.

#### Verification (full `go test ./...`, not just merged-in packages)

- `go build ./...` — **100% clean**, zero output.
- `go test ./...` run 1: every package passed except
  `pkg/tecdsa/dkg`, which failed 8 TSS-round subtests with
  `"TSS round one outgoing message was not generated on time"` — a
  wall-clock-timeout-based test failure.
- `go test ./...` run 2 (same commit, no code changes): **100% clean**,
  including `pkg/tecdsa/dkg` (216s, all green).
- Isolated re-runs of the run-1 failures (`go test ./pkg/tecdsa/dkg/...
  -run TestTssRound...`): all of `TestTssRoundTwo`, `TestTssRoundThree` and
  every subtest that failed in run 1 passed in isolation. A *different*
  subtest (`TestTssRoundOne`) flaked once on that isolated pass, then passed
  cleanly (7.35s) when re-run alone again.
- Host load during this session was heavy (`load average: 19.91, 28.17,
  16.53` on a 16-core box, multiple sibling agents running concurrent
  builds/tests) — the same signature as the pre-existing,
  confirmed-unrelated flakes this repo already has
  (`pkg/net/retransmission`'s `TestOnTickTimeTicker`,
  `pkg/generator`'s `TestCheckProtocols_NoProtocols`, both of which passed
  cleanly in both full runs here).
- `pkg/tecdsa/dkg` does not import `go-ethereum` at all — its only
  keepcommon-side dependency is `pkg/keepcommon/persistence`, unaffected by
  the version bump. Combined with "different subtest fails each run, all
  pass in isolation, unrelated import surface," this is a pre-existing,
  load-induced timing flake in the TSS-round tests, **not a regression from
  the go-ethereum bump**.
- `go vet ./...` surfaces one pre-existing issue in
  `pkg/tecdsa/signing/protocol.go:737` (lock-value copy from
  `bnb-chain/tss-lib`'s `common` package) — confirmed present, byte-for-byte
  unchanged, at `keep-core`'s pre-merge HEAD (`a7ac8989b`); untouched by
  this PR and out of scope here.
- Integration-tagged tests (`ethereum_integration_test.go`,
  `electrum_integration_test.go`, requiring live external RPC endpoints)
  were not run — same as keep-common's own CI and this repo's default
  `go test ./...`, they're excluded from the default build set and need
  live infra this scratch environment doesn't have.

## Follow-up (not in this PR)

Once this PR merges, `threshold-network/keep-common`'s own CI
(`.github/workflows/client.yml`, `.github/workflows/release.yml`,
`.github/dependabot.yml`) becomes meaningless and should be retired as a
**separate, follow-up** change — intentionally **not done here**.
keep-common's CI must stay defended (its govulncheck/gosec/lint/dependabot
workflows keep running) until this merge PR actually lands in `keep-core`,
in case this PR needs further changes before merge.


piotr-roslaniec (commented):
Still in progress

